PR #1533, opened 2013-09-29
Horizontal listview

eulerreich:
Added ability to switch orientation of listview to horizontal. I've only tested in my own app and it works fine. The only caveat, as with vertical listview, is that if size_hints of the views are not None in the corresponding orientation, then populate() might generate inconsistent sizes as you scroll down the list. Because of this, I forced size_hint to be (None, None) in the case of the labels for item_strings.

Hopefully this is compatible with geoff's branch.


akshayaurora:
please rebase and re-open

PR #1614, opened 2013-11-17
Right-to-left labels support

QGully:
Added right to left  word ordering support for the label core and the
label widget. Used for hebrew, arabic etc. to resolve incorrect word
ordering (and resulting line breaks) when rendering RTL text.
Essentially, we scan the words in reverse order (to ensure proper ordering of 'split' lines in case the label has a width constraint and is rendered over several lines), and reverse the glyphs to handle the reversed ordering of letters in these languages. Tested using hebrew and english.

### DIFF
--- a/kivy/lang.py
+++ b/kivy/lang.py
@@ -891,7 +891,7 @@ class ParserRule(object):
             value = self.properties[name].co_value
             if type(value) is CodeType:
                 value = None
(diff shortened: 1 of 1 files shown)


P0oOOOo0YA:
Hey Guys, Is this feature integrated into Kivy or not?

PR #1807, opened 2014-01-22
Touchring cursor

tshirtman:
Forgot to do a PR for this, should be functional, tests welcome.

### DIFF
--- a/kivy/modules/touchring.py
+++ b/kivy/modules/touchring.py
@@ -16,6 +16,7 @@ Configuration
     `alpha`: float, defaults to 1.
         Opacity of the image.
     `show_cursor`: boolean, default to False
+        .. versionadded:: 1.8.0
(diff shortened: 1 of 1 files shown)


kivy/modules/touchring.py
```diff
@@ -15,6 +15,15 @@
         Scale of the image.
     `alpha`: float, defaults to 1.
         Opacity of the image.
+    `show_cursor`: boolean, default to False
```
kenwaldek: one quick question to use it in a python script i need to add the code to my python program to be able to see the mouse gestures on my rpi

KeyWeeUsr: @kenwaldek this isn't really a good place to discuss it :D The PR is old and github is for issues, so instead - read [Community Guidelines](https://github.com/kivy/kivy/wiki/Community-Guidelines) and come visit our channels :)

PR #2214, opened 2014-05-27
Added handlers for non-multitouch touchscreen events

jesseanderson:
I found that the hidinput handlers did not handle the events that my touchscreen for Raspberry Pi was dispatching. These events are similar to the multitouch events already in place, but much simpler because they are for a simple resistive touchscreen. These changes allow these events to be handled where most appropriate.  Thanks!

### DIFF
--- a/kivy/input/providers/hidinput.py
+++ b/kivy/input/providers/hidinput.py
@@ -115,6 +115,11 @@ else:
     MSC_MAX = 0x07
     MSC_CNT = (MSC_MAX + 1)
 
+    # Events for stmpe touchscreen
+    ABS_X = 0x00
+    ABS_Y = 0x01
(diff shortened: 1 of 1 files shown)


akshayaurora:
There have been a lot of changes for touch input since this pr was opened.  Could you please retest if this is still a issue with latest master?

If it is could you please rebase.

Sry for the delay. One major issue why pr's like these don't get included asap is lack of appropriate hardware to test. I guess we could just make sure this does not break existing hardware  interaction.

jesseanderson:
Ah.  For the time being I don't have the hardware to update and recheck this.  When the opportunity arises I will come back to this.  In the meantime I think it makes sense for me to close this since I don't think there's any immediate action I can take.  Thanks for the response.

Issue #1619, opened 2013-11-19
Feature Request: right to left support for TextInput


nim4n:
##

cryptoone:
we need this to be resolved

PR #2411, opened 2014-08-20
basic bidi support

kived:
Support bi-directional text using [pyfribidi](http://pyfribidi.sourceforge.net/) or [python-bidi](https://pypi.python.org/pypi/python-bidi), if available. This just handles bidi text, and does not do any reshaping (I tested with Hebrew).

bidi support is added to core `LabelBase` and exposed via properties in `Label` and `TextInput`, and should work with any widget subclassing those (I tested `Button` as well).

Also adds a config section `i18n` and setting `bidi`, which controls the default value for the core label setting.

This will help with #1619.

### DIFF
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -216,6 +216,17 @@ Available configuration tokens
     Check the specific module's documentation for a list of accepted
     arguments.
 
+:i18n:
+
+    `bidi`: int, 0 or 1
+        Support bi-directional text by default on all widgets subclassing
+        :class:`~kivy.uix.label.Label` or
+        :class:`~kivy.uix.textinput.TextInput`, if supported. Support can
+        still be toggled for an individual widget, regardless of this setting.
+
+.. versionchanged:: 1.9.0
+    `i18n` section added.
+
 .. versionchanged:: 1.8.0
     `systemanddock` and `systemandmulti` has been added as possible values for
     `keyboard_mode` in the kivy section. `exit_on_escape` has been added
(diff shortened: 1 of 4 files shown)


sajadbanooie:
the only reason im not using kivy is it doesn't support Persian language

ibtisamN:
Since kivy use SDL for drawing a text, [here](https://bugzilla-attachments.libsdl.org/attachment.cgi?id=2354) is new patch that support RTL languages for SDL_TTF.
I have applied the patch and run unicode_textinput example and I got  rtl_language
But another problem is now raises with cursor position in RTL text with SDL_TTF. The cursor deals with RTL text as LTR text, so it behave incorrectly. Anyone can help me to find out the Kivy code that I should modify to solve this bug.

Fahad-Alsaidi:
@ibtisamN have you checked this folder? 
https://github.com/kivy/kivy/tree/master/kivy/core/text

ibtisamN:
@Fahad-Alsaidi  yes I have checked that folder, but nothing was related to the cursor handling. I have looked this file https://github.com/kivy/kivy/blob/master/kivy/uix/textinput.py which handle cursor selection, movement ...etc. I found the code that set the cursor offset and it was coded  incorrectly because they assume fixed width for all characters which is fault since each character have different width. So now I'm working on adding new api to [libraqm ](https://github.com/HOST-Oman/libraqm) which will give us the correct position of the cursor.

engmms:
@ibtisamN,,How I can apply this patch

RevengeComing:
I suggest behave all texts as bi-directional. (at least in TextInputs )

For example in android sdk every texts are bi-directional, but the direction depends on what language do you start your text, if you start your text with LTR ( like english ) the text will be in left direction , but if you start it with RTL the text will be in right direction.

for example :
left_dir
right_dir

I can work on it ? and i guess i can fix cursor problem for RTL texts.

tito:
Now that we have experimental Pango2 support, i'll close this one. The general direction is going to supports more languages.

Issue #1440, opened 2013-08-13
How to set the default font?


logbird:
How to set the default font?
Each time you create a Label setting is very troublesome.
I preview kivy code, found in many parts of the fonts are fixed.
Can you put the default font in the config.ini file to set.

Thank You!

senchoi:
Font doesn’t apply when the widgets are different. Does anyone know how to change the default font for any text appears on kivy app?

Thimphu123:
You can import `from kivy.core.text import LabelBase, DEFAULT_FONT`
and add this command `LabelBase.register(DEFAULT_FONT, "path/to/font.ttf")`

Paalal:
You can change the default font in the <HOME_DIRECTORY>/.kivy/config.ini file. In this file you can find a setting called "default_font".

Put the fonts you want to use in the <HOME_DIRECTORY>/AppData/Local/Packages/PythonSoftwareFoundation.Python.3.11/LocalCache/local-packages/Python311/site-packages/kivy/data/fonts folder.

AlkisPis:
> You can import `from kivy.core.text import LabelBase, DEFAULT_FONT` and add this command `LabelBase.register(DEFAULT_FONT, "path/to/font.ttf")`

For me this didn't change anything. The default font seems to be always Roboto, which unfortunately cannot handle Arabic and Hebrew languages, which are used --surprisingly so-- in https://github.com/kivy/kivy/blob/master/examples/text/pango_demo.py.

AlkisPis:
> You can change the default font in the <HOME_DIRECTORY>/.kivy/config.ini file. In this file you can find a setting called "default_font".
> 
> Put the fonts you want to use in the <HOME_DIRECTORY>/AppData/Local/Packages/PythonSoftwareFoundation.Python.3.11/LocalCache/local-packages/Python311/site-packages/kivy/data/fonts folder.

This seems to be the only way. I tested it with Arabic and Hebrew text, setting **Arial.ttf** as default font (Windows).  
Yet, the font I used does not handle Chinese and Tibetan. I really wonder then what is the meaning of using a text demo (https://github.com/kivy/kivy/blob/master/examples/text/pango_demo.py) when the default Kivy font cannot handle those languages ...
(**DejaVuSans.ttf**, included in 'data/(https://github.com/kivy/kivy/blob/master/kivy/data/fonts can also handle Arabic and Hebrew, but not the other languages.)

ElliotGarbus:
@AlkisPis 
[...]

PR #2412, opened 2014-08-20
make fonts easier to load; add default font function

kived:
Quick font fixes, primarily to make it easier to support non-Latin languages.

1) Add a `set_default_font()` function to change the global default.
2) Make it easier to load fonts. Kivy includes DroidSans.ttf and DejaVuSans.ttf in the same folder (data/fonts). The default `font_name` is 'DroidSans', but setting the `font_name` to 'DejaVuSans' didn't work when it seems like it should.

Just give feedback for now, do not merge yet. This probably won't apply cleanly with #2411 due to adjacent hunks, and I will still need to add some docstrings.

### DIFF
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -53,6 +53,16 @@ FONT_BOLD = 2
 FONT_BOLDITALIC = 3
 
 
+def set_default_font(fontname):
+    global DEFAULT_FONT
+    DEFAULT_FONT = fontname
+
+    from kivy.uix.label import Label
+    Label.font_name.defaultvalue = fontname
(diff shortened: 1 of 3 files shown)


akshayaurora:
@kived could we revisit this in light of the recent changes to font loading ?

andong777:
Any progress on this PR?
I find it a little bit annoying to specify font name for widgets. It is a must for us eastern-asian developers to use unicode-supported font to display unicode chars.
Either setting global font in the code, or setting it in a property file #1440 will be good!

Zen-CODE:
This PR seems to be hanging around. Is it still relevant? I'm happy to test it if it's made merge-able, but otherwise can we close it?

PR #2537, opened 2014-10-01
Add tools/kviewer (auto-reload an on-disk kv file and displays the result)

tshirtman:
inspired by kivycatalog, but fresh codebase


dessant:
@tshirtman :pager:

tshirtman:
:blush:

tito:
Better than nothing.

PR #2567, opened 2014-10-09
Support for CompositeListItem selection initialization

Zen-CODE:
Currently, it only seems possible to initialize a CompositeListItem with the first item selected using the 'allow_empty_selection=False' parameter to the adapter. This PR adds support for specifying any amount of items to be selected by default.

This gist can be used to demonstrate and test:
https://gist.github.com/Zen-CODE/db169c77416c78da12e2

It seems the CompositeListItem needs to manually delegate the initial selection to it's children.
The 'select' and 'unselect' events seem to be fired only on initialization (tested via print statements in the functions), so this seems like an optimal solution. The forced subclassing of 'SelectableView' ensures that the children will all have 'select' and 'unselect' methods to invoke, so there should be no issue with missing methods.

### DIFF
--- a/kivy/uix/listview.py
+++ b/kivy/uix/listview.py
@@ -742,9 +742,13 @@ class CompositeListItem(SelectableView, BoxLayout):
 
     def select(self, *args):
         self.background_color = self.selected_color
+        for child in self.children:
+            child.select()
 
     def deselect(self, *args):
         self.background_color = self.deselected_color
+        for child in self.children:
+            child.deselect()
 
     def select_from_child(self, child, *args):
         for c in self.children:


dessant:
@Zen-CODE, @kived, what is the fate of this pr?

Zen-CODE:
Okay, I think I missed kiveds' point last time. The issue here is assuming the classes in 'cls_dicts' subclass SelectableView. If they don't and don't provide 'select' and 'deselect' methods, stuff will break. From the examples and my use case, I assumed the 'cls_dicts' were composed of 'SelectableView' classes/sub-classes, but that not explicitly stated so unfair to assume

I made the PR as it is very was convenient to have selection propagated automatically in these cases, and makes sense for automatically handling CompositeListItem selection. But it does make that assumption.

Its is convenient, and we could guard against errors by checking for the attribute first i.e.

    def select(self, *args):
        self.background_color = self.selected_color
        for child in self.children:
            selection = getattr(child, 'select', None):
            if selection in not None:
                selection()

But that could also change behavior if they, by coincidence have such a method or property. We could check that the children do inherit from SelectableView, but that's also does not feel good.

So, it comes down to the core decision:

1. if composite item should be composed of the SelectableView subclasses (the examples all do this, but who know what's out there is real life), then I'd say yes, this PR should be merged.

2. If composite items should be able to have children of any kind, then no, this should be refused or modified.

So:
@dessant or kived: Feel free to make a call and either close this or expand the discussion. In retrospect, I tend to side with kived: compatibility is probably foremost, so kill it :-)

dessant:
I stay away from everything related to listview :stuck_out_tongue_winking_eye: , the decision is yours guys.

Zen-CODE:
Yeah, the listview's are already on the dangerous side of trying to be too clever. Sometimes, trying to be too abstract makes it more complex to do and less flexible. 

So lets not make it more fragile. @kived. Thanks for the catch. Agreed. :-)

PR #2630, opened 2014-10-30
Models loaded in kivy are much smaller than opengl

### LINKED ISSUE #2629: Matrix Transform shrinks loaded models
The below program highlight the issue / difference between opengl and kivy perspective matrices which cause the shrinkage.

http://pastebin.com/zSxe8pPR - opengl
http://pastebin.com/N1Tfa3eY - kivy

Looking at the code here https://github.com/kivy/kivy/blob/master/kivy/graphics/transformation.pyx it looks like self.mat[10] and self.mat[14] need to be switched but i am no expert, should self.mat[14] even be hard coded when it seems to vary with gluperspective ?

I tried patching this myself switching the above value seems to solve the issue, i shall do some more tests to make sure it has not broken anything else.
Will changing this cause any other issue for existing applications ? if it does could we add a new method called gluPerspective or similar which matchs the opengl implementation.

The below screenshots outline the issue, the triangle is the same in both applications but some how much much smaller in kivy.

selection_148
renderer_149

Adjusted code produces this.

renderer_150

looking through the code perspective seems to be referenced a few times in the .c files for example context.c vbo.c instructions.c etc not sure it this is internal though and seperate from the python method, there seems to be no reference in the pyx files other than in transformations.pyx file where it is defined.

olymk2:
changes the perspective matrix to work like gluperspective
fixes #2629 

Reading up about this i noticed the values seemed to be the wrong way around, changing it fixes the tiny model issue and all the examples seem to still work, can someone confirm this and suggest a better fix if it will cause issues ?

### DIFF
--- a/kivy/graphics/transformation.pyx
+++ b/kivy/graphics/transformation.pyx
@@ -181,10 +181,10 @@ cdef class Matrix:
         self.mat[8]  = 0.0
         self.mat[9]  = 0.0
         self.mat[10] = (zFar + zNear) / (zNear - zFar)
-        self.mat[11] = (2 * zFar * zNear) / (zNear - zFar)
+        self.mat[11] = -1.0
         self.mat[12] = 0.0
         self.mat[13] = 0.0
-        self.mat[14] = -1.0
+        self.mat[14] = (2 * zFar * zNear) / (zNear - zFar)
         self.mat[15] = 0.0
 
     cpdef Matrix view_clip(Matrix self, double left, double right,


Kovak:
Ok I think this might be a dangerous change actually.

OpenGL work with column-major ordering matrices, and Kivy with row-major. This particular change happen to work because shifting the one value results in the transpose of the original matrix so that it is now in column-major the way @olymk2 expects. However, now this mean we are dealing with a single column-major and every other aspect of code base is row-major still.

Issue #2514, opened 2014-09-26
window_sdl2.screenshot uses pygame


dessant:
https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L147

akshayaurora:
I can't reproduce the issue using sdl2 from mercurial on a mac or a window machine.

tito:
same for me, not reproducable on my osx / linux / windows.

tito:
Closing, unless we have a way to reproduce.

PR #2661, opened 2014-11-09
Sdl2 screenshot leads to slanted image

inclement:
Not for merging right now!

I implemented this a while ago, and made the pr for easy reference. This code doesn't quite work, because the flipVert function doesn't flip the texture properly - there's an offest or something, and you get a slant instead. It's probably a small fix, and it works otherwise.

Fixes https://github.com/kivy/kivy/issues/2514.


akshayaurora:
I can't reproduce the issue using sdl2 from mercurial on a mac or a window machine.

PR #2692, opened 2014-11-20
handle data uris for images

kived:


akshayaurora:
Update: Most Major kivy providers for image now support loading from memory. Would be great to have this be refactered to take advantage of that.

akshayaurora:
@kived this is currently not able to merge, refactor, rebase and reopen.

PR #2729, opened 2014-12-07
don't require a text argument for CompositeListItems

kived:


dessant:
@kived, please rebase. :yellow_heart:

dessant:
:+1:

PR #2730, opened 2014-12-07
replace nbsp with space

kived:


akshayaurora:
This is now redundant with matham's pr.

PR #2743, opened 2014-12-13
Clipboard os specific

akshayaurora:
This is just a backup platform specific code for clipboard. This frees us from being too dependent on any one library like pygame or sdl2.

Windows: win_ctypes
Linux: dbusklipper
OSX: NSPasteboard(requires pyobjus: obligatory dog food post...)

ios: Not yet implemented
Android already had android_clipboard.

Tested to work on linux, windows and osx.
The only part left to include is the iOS clipboard implementation, that can be included later.


stuartmarsden:
I have added to this by making a linux clipboard provider which uses xsel so is not tied to KDE.

[Pull request](https://github.com/kivy/kivy/pull/3159/files#diff-f67d151f08b550cc8a89074f800b9292)

PR #2749, opened 2014-12-15
prevent flicker when using shader transitions

### LINKED ISSUE #2748: shader transitions flicker on start
When using shader transitions with `ScreenManager`, the screen sometimes flickers and shows the background behind the `ScreenManager`.

shader

The issue is easier to reproduce with a lower framerate (i.e. set maxfps = 5 in config.ini). With this test code, the white canvas sometimes flickers in the background at maxfps = 60, and is clearly visible every time with maxfps = 5:

````python
#import pydevd; pydevd.settrace('localhost', port=51514, suspend=False)

import kivy
kivy.require('1.8.1')

from kivy.app import App
from kivy.lang import Builder
from kivy.clock import Clock

#:import sm kivy.uix.screenmanager
ScreenManager:
    transition: sm.WipeTransition()
    canvas.before:
        Color:
            rgba: 1, 1, 1, 1
        Rectangle:
            pos: self.pos
            size: self.size
    Screen:
        name: 'first'
        BoxLayout:
            canvas.before:
                Color:
                    rgba: 1, .5, .5, 1
                Rectangle:
                    size: self.size
            Label:
                text: 'first'
            Button:
                text: 'button'
    Screen:
        name: 'second'
        BoxLayout:
            canvas.before:
                Color:
                    rgba: .5, .5, 1, 1
                Rectangle:
                    size: self.size
            Label:
                text: 'second'
            BoxLayout:
                orientation: 'vertical'
                Button:
                    text: 'button 1'
                Button:
                    text: 'button 2'
''')

class TestApp(App):
    def build(self):
        Clock.schedule_interval(lambda dt: setattr(root, 'current', 'first' if root.current == 'second' else 'second'), 1)
        return root

if __name__ == '__main__':
    TestApp().run()
````

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7034536-shader-transitions-flicker-on-start?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kived:
Keeps rendering the screen-out's canvas until the render context has actually rendered its first frame. Prevents flicker at the beginning of shader transitions.

Fixes #2748.

### DIFF
--- a/kivy/uix/screenmanager.py
+++ b/kivy/uix/screenmanager.py
@@ -487,6 +487,12 @@ class ShaderTransition(TransitionBase):
         self.render_ctx['t'] = 1.
         super(ShaderTransition, self).on_complete()
 
+    def _remove_out_canvas(self, *args):
+        if (self.screen_out
+                and self.screen_out.canvas in self.manager.canvas.children
+                and self.screen_out not in self.manager.children):
+            self.manager.canvas.remove(self.screen_out.canvas)
+
     def add_screen(self, screen):
         self.screen_in.pos = self.screen_out.pos
         self.screen_in.size = self.screen_out.size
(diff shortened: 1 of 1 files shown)


kived:
@tito After `screen_out` is removed from the `ScreenManager`'s children, its canvas is added back to the `ScreenManager`'s canvas. The final removal of the canvas is delayed until the transition renders its first frame, by use of a `Callback`. When the `Callback` runs, it removes itself from the instructions, so it won't be called again, and schedules removal of `screen_out`'s canvas.

PR #2756, opened 2014-12-18
fixed LoadIdentity constructor

IMDagger:
Fixed exception due to property expectation of dictionary instead of string. It seems that was a typo.


IMDagger:
Hello. Are there any updates about this pull request? This fix is very simple and still actual (fixed small typo).

tito:
replaced by #3750

PR #2769, opened 2014-12-22
Minor fix in slider.py. slider.value now set to slider.min at init

jmriviere:
While I was coding some software using sliders, a DivisionByZero exception made me realise that the slider's value was not set to slider.min at init but rather to 0. Simply adding a constructor to Slider and calling self.value = self.min fixed the issue

EDIT: the fix works when specifying the Slider in python code. Still trying to get it work with kv lang


akshayaurora:
@jmriviere :+1:

Issue #944, opened 2013-02-05
Add more transformation methods to Texture


CreamyCookie:
In the Texture class there's already a method to flip_vertical, I propose to add flip_horizontal and several rotation methods:

``` python
def rotate_90_degrees(self):
    self.tex_coords = self.tex_coords[2:] + self.tex_coords[:2]

def rotate_180_degrees(self):
    self.tex_coords = self.tex_coords[4:] + self.tex_coords[:4]

def rotate_270_degrees(self):
    self.tex_coords = self.tex_coords[-2:] + self.tex_coords[:-2]

ROTATE_FUNC = {0: lambda a: None, 90: rotate_90_degrees, 180: rotate_180_degrees, 270: rotate_270_degrees}

def rotate(self, angle):
    assert angle % 90 == 0, 'the angle must be a value that can be divided by 90 without remainder'

    ROTATE_FUNC[angle % 360](self)
```
##

PR #2770, opened 2014-12-22
Implementing #944 (Feature request: texture rotation)

victor-rene:
I didn't implement the rotate function, as described in #944. Something about it seems off.
While it makes sense to have a generic rotate function, I don't like introducing confusing behavior.
Better to keep each function separate and call them as needed.

### DIFF
--- a/kivy/graphics/texture.pxd
+++ b/kivy/graphics/texture.pxd
@@ -38,6 +38,9 @@ cdef class Texture:
 
     cpdef flip_vertical(self)
     cpdef flip_horizontal(self)
+    cpdef rotate_90_deg(self)
+    cpdef rotate_180_deg(self)
+    cpdef rotate_270_deg(self)
     cpdef get_region(self, x, y, width, height)
     cpdef bind(self)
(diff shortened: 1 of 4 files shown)


snapy666:
I agree with @victor-rene, despite having made the proposal for the general `rotate` function.

In praxis one would probably not use integers such as `(0, 90, 180, 270)`, but `(0, 1, 2, 3)` or, if the texture is an arrow, `('left', 'right', 'top', 'down')`. Also, writing such a mapping is trivial.

CreamyCookie:
@victor-rene Any news on this? Maybe you could modify this pull request to remove the generic rotate function..

CreamyCookie:
@inclement Have you thought about the naming?

I think adding a generic `rotate` function is a bad idea. (One can easily write such a mapping, or use an if-else-if-chain.)

Zen-CODE:
This issue was already addressed with this merge: https://github.com/kivy/kivy/pull/1996

Thanks anyway :-)

PR #2772, opened 2014-12-22
attempt to handle proxy servers in UrlRequest

kived:
Untested - I don't have a proxy server to test with. ;)


DavidA2014:
Confirmed that this works with Python 3.4.3 behind a proxy server. Thanks.

dessant:
@kived could you rebase?

dessant:
@kived :bouquet:

PR #2774, opened 2014-12-23
Fix cutoff last line when line_height is less than 1.0

### LINKED ISSUE #2502: Label with line_height < 0 are broken
Test case:

```
runTouchApp(Label(
    text="AAAApppp\nppppAAAAA",
    line_height=0.5,
    text_size=(None, None)))
```

Result:
image

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/4664096-label-with-line_height-0-are-broken?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

matham:
Fixes #2502.

Sorry guys for the delay, but this should fix #2502.

It doesn't completely fix it for when markup is True with more than one line, but I don't think it's quite possible to to fix there as well without a lot more effort. The reason being that imagine for the various lines we use different `line_height`. Then to find the total required height so that nothing is cut off, we'd need to keep track of the various leftovers that doesn't fit vertically on each line to get the maximum total. Because keep in mind that `line_height` just moves lines rather than actually rescaling. But for markup, I imagine the user has the option to just use a smaller font size rather than changing the line size.

### DIFF
--- a/kivy/core/text/markup.py
+++ b/kivy/core/text/markup.py
@@ -410,18 +410,20 @@ class MarkupLabel(MarkupLabelBase):
             psp = pph = 0
             for word in layout_line.words:
                 options = self.options = word.options
+                # the word height is not scaled by line_height, only lh was
+                wh = options['line_height'] * word.lh
                 # calculate sub/super script pos
                 if options['script'] == 'superscript':
                     script_pos = max(0, psp if psp else self.get_descent())
                     psp = script_pos
-                    pph = word.lh
+                    pph = wh
                 elif options['script'] == 'subscript':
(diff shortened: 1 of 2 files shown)


kived:
`CodeInput` doesn't work at all with this (tested in Showcase). No text renders at all, and every line looks to contain a single space and nothing else.

matham:
@kived, It is fixed, it was due to using an uninitialized variable.

PR #2782, opened 2014-12-23
bind listview adapter triggers on adapter change

### LINKED ISSUE #2781: ListView adapter bug
When the data backing a ListView changes, the ListView does not always update correctly.

Sample code here:   https://gist.github.com/billmcc64/a97bd1db128c87a91026#file-listviewproblem-py

uncomment line 25 to correct the problem.

The ListView constructor executes the method bind_triggers_to_view() on the adapter.

If the adapter changes after the constructor has been called (as in this example, or if the ListView is created via a .kv file), this method does not get called, and updates to the data backing the ListView are not displayed.

Not sure the best way to fix this.   The __set__ method in the ObjectProperty could be overriden for ListView to catch changes to the adapter.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/7221844-listview-adapter-bug?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kived:
Fixes #2781

### DIFF
--- a/kivy/uix/listview.py
+++ b/kivy/uix/listview.py
@@ -899,12 +899,18 @@ class ListView(AbstractView, EventDispatcher):
                   item_strings=self.item_strings_changed,
                   adapter=self._trigger_populate)
 
+        self._trigger_bind_adapter = Clock.create_trigger(
+            lambda dt: self.adapter.bind_triggers_to_view(
+                self._trigger_reset_populate),
+            -1)
+        self.bind(adapter=self._trigger_bind_adapter)
+
         # The bindings setup above sets self._trigger_populate() to fire
         # when the adapter changes, but we also need this binding for when
         # adapter.data and other possible triggers change for view updating.
         # We don't know that these are, so we ask the adapter to set up the
         # bindings back to the view updating function here.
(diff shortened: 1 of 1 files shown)


kivy/uix/listview.py
```diff
@@ -899,12 +899,18 @@ def __init__(self, **kwargs):
                   item_strings=self.item_strings_changed,
                   adapter=self._trigger_populate)
 
+        self._trigger_bind_adapter = Clock.create_trigger(
+            lambda dt: self.adapter.bind_triggers_to_view(
```
akshayaurora: you get a new function signature every time with lambda so it is essentially as good as not using a trigger.

kived: Every time what? This is in `__init__`, it will only be called once per widget. The resulting trigger will be a `ClockEvent` which will function the same regardless of whether the callback is a lambda or not.

akshayaurora: You are right, I didn't think of this clearly enough.

PR #2784, opened 2014-12-24
Added property name setters in Color __init__

inclement:
It's always mildly annoyed me that the Color instruction didn't support setting with kwargs in python, and there's even an old issue that references it, so this seems like a reasonable fix.


dessant:
:+1:  super duper

Issue #2795, opened 2014-12-27
Issue with DropDown() Widget Usage with ToggleButton()


cuspyprotocol:
Issue:
While the current code works for most cases, the current case for which I am using the `DropDown()` widget (a toggle button), when I press down, with the auto_dismiss parameter set to True, the drop down list is dismissed when it should not be dismissed.  The following addition fixes that issue.

Solution: 

See: Actual Fix to Issue #2795
##

akshayaurora:
@cuspyprotocol please start using git posting code as is won't help us review it,  there is a great tutorial on github for this https://help.github.com/articles/good-resources-for-learning-git-and-github/

PR #2796, opened 2014-12-27
Fix to Issue #2795

cuspyprotocol:
Fixes issue https://github.com/kivy/kivy/issues/2795 when I press down on a toggle button attached to a
DropDown() widget, with the auto_dismiss parameter set to True, the
drop down list is dismissed when it should not be dismissed.


akshayaurora:
Why should it not be dismissed when you click on the toggle?

cuspyprotocol:
Akshay,
       I think I uploaded this when I had the wrong idea about how the interactions were set up.  I would probably deny this merger for now until the issue discussed here https://github.com/kivy/kivy/issues/2795#issuecomment-69807670 is resolved.  Sorry about that.

PR #2799, opened 2014-12-27
Added joypad events and changed event handling complexity from O(n) to O(1)

snapy666:
Added joypad events and changed event handling complexity from O(n) to O(1) as described [here](http://stackoverflow.com/questions/11445226/better-optimization-technique-using-if-else-or-dictionary).

### DIFF
--- a/kivy/core/window/window_pygame.py
+++ b/kivy/core/window/window_pygame.py
@@ -108,6 +108,8 @@ class WindowPygame(WindowBase):
 
         # never stay with a None pos, application using w.center will be fired.
         self._pos = (0, 0)
+        
+        self.prepare_event_dict()
 
         # prepare keyboard
         repeat_delay = int(Config.get('kivy', 'keyboard_repeat_delay'))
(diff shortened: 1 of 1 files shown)


snapy666:
@Kovak @inclement Could you review this and, if it's okay, merge it? If you don't have time, how long will it normally take till it is done?

snapy666:
@dessant @Zen-CODE  Hello and sorry for the nudge :)

Can I get some feedback regarding this pull request and, if it's the case, why it is ignored? Did I miss something that was mentioned in the *Contributing section*? I'm fine with this not being used, but I would at least like to know what's wrong.

I've also realized that this change should probably be done in all `window*.py` files. (e.g. SDL2)
(Also, I don't need to be included in the eternal commit log.)

Zen-CODE:
@snappy66. Sorry, i don't enough technical knowledge on this area to yay or nay. ;-
(

Kovak:
@snapy666 first of all thank you very much for the contribution I definitely like the direction of it. I wanted to give some explanation (make excuses) for why we haven't gotten to this yet. First and foremost, most of the team is heavily involved in trying to roll up 1.9 for release, while we tackle the backlog it may take a bit for us to reach more current contributions. We have not forgotten you and we will get here eventually!

Secondly, we must ensure that these changes work correctly on every platform that has pygame as a provider. For this type of thing it means waiting until we have a couple confirmed people on every OS, and, in this case, with joypad hardware, report the changes working. A simple test application would probably help speed this process up. 

Thirdly, kivy really do not care so much for single provider features. Kivy strive to expose a common set of functionality available among every provider we have. Before joypad support can be brought into kivy we need to make sure we have it everywhere. Same for the performance changes with the dispatching of events, these should probably be made for SDL2 as well if necessary, and any other input provider we use. Especially as we are looking to replace pygame with sdl2 backend moving forward. 

Finally, technically we have 2 features here. One is the performance refactoring of event handling and the other is the joypad support. It would be conceptually easier and much more streamlined to separate this into 2 separate PR, this way we could probably more quickly tackle the event handling stuff as it does not require a joypad for testing. 

One again thank you for the contribution and the refactor of event handling was a good spot!

Kovak:
After some discussion with the other kivy core devs:

There is a concern over the performance implications of the re-factoring of event handling. The cost associated with the added function call over the current situation is most likely not in favor of this change. We would welcome profiling to investigate the situation here, but we are skeptical of the potential benefits. If such a change did prove to be favorable it would require a much larger overhaul throughout the code base to implement the same pattern everywhere the current approach is used. More data is necessary before making such a call.

We would welcome separating out the joypad section and implementing it in the current if then else statement as a separate PR

snapy666:
@Kovak Thank you! I agree with splitting this pull requests, although I'm not sure if I want or could implement this everywhere.

Regarding the concern: The function call is a constant cost, while the if-elif-else statement has a cost that scales with the number of events handled.

I have created such [**a test to measure the execution time**](https://gist.github.com/snapy666/1ad9c6810d51ec8d7c12). Here are the results: (You might need to first have a look at the test to understand what the `test_` names stand for.)

# Every key found in dict or if-elif-else
Name | Execution time in seconds
------------ | -------------
test_dict_get |              8.716617422000127
test_check_key_in_dict |     9.00435929500236
test_except_key_error |      6.264725733999512
test_default_return_dict |   10.186109538000892
test_defaultdict |           6.141208559998631
test_if_elif_else |          10.721702154998638


# 50 found / 50 not found
Name | Execution time in seconds
------------ | -------------
test_dict_get |              7.955001826001535
test_check_key_in_dict |     7.73699967200082
test_except_key_error |      11.682552223999664
test_default_return_dict |   14.72374195500015
test_defaultdict |           6.099408977999701
test_if_elif_else |          17.825281888999598


# No key found in dict or if-elif-else
Name | Execution time in seconds
------------ | -------------
test_dict_get |              7.479436647998227
test_check_key_in_dict |     6.2445404840000265
test_except_key_error |      16.10462998499861
test_default_return_dict |   18.835381493998284
test_defaultdict |           5.325450674001331
test_if_elif_else |          24.779503347999707

kived:
@snapy666 Yes, dict lookups are always faster than if/elif. But you're missing the most important point - the original code **did not call functions** while your code does. Function calls are expensive in Python, and it's unlikely that this is an actual speed up.

snapy666:
@kived Oh.. You're right. Thanks and sorry! :+1:  How could I not see that? I've rectified my post, but I think, the conclusion still pretty much holds true. But if you disagree, that's fine and I'd like to hear that, as I have yet (or always) a lot to learn.

Note that there's still a version of `if-elif-else`, which I've renamed to `test_if_elif_else_with_function_calls`, that calls functions, because it is cleaner code and would be preferred by me in other situations.

---

@Kovak Thank you! :+1:

It makes sense to separate this pull request. I'm already happy with the gamepad API in SDL2, which I didn't knew about that, when I created this. Therefore I won't create a pull request for that. Sorry!

Regarding profiling: I've created [**a test to measure the execution time**](https://gist.github.com/snapy666/1ad9c6810d51ec8d7c12). You might first need to have a look at the linked test to understand what the following `test_` names stand for.

## Every key found in dict or if-elif-else
Name | Execution time in seconds
------------ | -------------
test_dict_get |           7.952993699000217
test_check_key_in_dict |  8.647647624000456
test_except_key_error |   6.057729496999855
test_default_return_dict |9.819119497999964
test_defaultdict |        5.720464381000056
test_if_elif_else |       6.831272555999931
test_if_elif_else_with_function_calls |    10.540674192000552

## 50 found / 50 not found
Name | Execution time in seconds
------------ | -------------
test_dict_get |           7.912311480999961
test_check_key_in_dict |  8.01090596100039
test_except_key_error |   12.802067794000322
test_default_return_dict |14.195735510999839
test_defaultdict |        5.764159217999804
test_if_elif_else |       13.870207252999535
test_if_elif_else_with_function_calls |    17.866621209000186

## No key found in dict or if-elif-else
Name | Execution time in seconds
------------ | -------------
test_dict_get |           7.459581842999796
test_check_key_in_dict |  6.864278949999971
test_except_key_error |   17.304682987999513
test_default_return_dict |18.125663195000016
test_defaultdict |        5.079881313000442
test_if_elif_else |       20.7206846629997
test_if_elif_else_with_function_calls |    24.594715916000496

### Note regarding defaultdict
Using `defaultdict` would be okay as long as the number of events is limited — which is usually the case —, as every item access (`dict[key]`) for a `key` that isn't in the dictionary would then be inserted.

### Conclusion
Even though the `test_defaultdict` contains function calls, which are expensive in Python, the `defaultdict` is still more efficient than using a normal `if elif else` statement, even though it only consists of *5* `conditions. (The `if elif else` statement gets slower for every condition, while the function cost is constant.) The `window_pygame.py` has 14 conditions.

snapy666:
@kived Oh.. You're right. Thanks and sorry! :+1:  How could I not see that? I've rectified and improved my post, but I think, the conclusion still holds true. If you disagree, that's great and I'd like to hear why that is, as there's always something to learn.

Note that there's still a version of `if-elif-else` that calls a function for every key, because it is cleaner code and would be preferred by me in other situations. I've renamed it to `test_if_elif_else_with_function_calls`.

---

@Kovak Thank you! :+1:

It makes sense to separate this pull request. I'm already happy with the gamepad API in SDL2, which I didn't knew about, when I created this PR. Therefore, and because SDL2 will probably replace pygame, I won't create a pull request for gamepad support in pygame though. Sorry!

---

Regarding profiling: I've created [**tests to measure the execution time**](https://gist.github.com/snapy666/1ad9c6810d51ec8d7c12). You, the reader, might first need to have a look at the mentioned test to understand what the following `test_` names stand for.

For every setup, there is a version with 5 conditions in `if-elif-else` and one with 13 conditions. For reference, the `window_pygame.py` has 14 conditions.

They both have slightly different number of elements to test for, which is why, for example, `test_dict_get` with *5 conditions* is faster than `test_dict_get` with *13 conditions*.

# Every key found in dict or if-elif-else
## 5 conditions
Name | Execution time in seconds
------------ | -------------
test_dict_get |           7.952993699000217
test_check_key_in_dict |  8.647647624000456
test_except_key_error |   6.057729496999855
test_default_return_dict |9.819119497999964
test_defaultdict |        5.720464381000056
test_if_elif_else |       6.831272555999931
test_if_elif_else_with_function_calls |    10.540674192000552

## 13 conditions
Name | Execution time in seconds
------------ | -------------
test_dict_get | 9.449431921000723
test_check_key_in_dict | 10.698207212000852
test_except_key_error | 7.381142012000055
test_default_return_dict | 11.772208508000404
test_defaultdict | 6.93195446900063
test_if_elif_else | 13.499845717000426
test_if_elif_else_with_function_calls | 17.57994560000043


# 50 found / 50 not found
## 5 conditions
Name | Execution time in seconds
------------ | -------------
test_dict_get |           7.912311480999961
test_check_key_in_dict |  8.01090596100039
test_except_key_error |   12.802067794000322
test_default_return_dict |14.195735510999839
test_defaultdict |        5.764159217999804
test_if_elif_else |       13.870207252999535
test_if_elif_else_with_function_calls |    17.866621209000186

## 13 conditions
Name | Execution time in seconds
------------ | -------------
test_dict_get | 8.793422193000879
test_check_key_in_dict | 8.888842536998709
test_except_key_error | 13.854044087998773
test_default_return_dict | 15.674793932001194
test_defaultdict | 6.432961722000982
test_if_elif_else | 35.18961696000042
test_if_elif_else_with_function_calls | 39.2035331900006


# No key found in dict or if-elif-else
## 5 conditions
Name | Execution time in seconds
------------ | -------------
test_dict_get |           7.459581842999796
test_check_key_in_dict |  6.864278949999971
test_except_key_error |   17.304682987999513
test_default_return_dict |18.125663195000016
test_defaultdict |        5.079881313000442
test_if_elif_else |       20.7206846629997
test_if_elif_else_with_function_calls |    24.594715916000496

## 13 conditions
Name | Execution time in seconds
------------ | -------------
test_dict_get | 8.228163970999958
test_check_key_in_dict | 7.519528319999154
test_except_key_error | 18.427840266000203
test_default_return_dict | 19.96727259399995
test_defaultdict | 5.754181506999885
test_if_elif_else | 56.88675264700032
test_if_elif_else_with_function_calls | 60.5560651199994

### Note regarding defaultdict
Using `defaultdict` would be okay as long as the number of events is limited — which is usually the case —, as every item access (`dict[key]`) for a `key` that isn't in the dictionary would then be inserted.

### Conclusion
Some of these test options (especially `defaultdict`) are more efficient than using a `if elif else` *chain*, even though the `test_defaultdict` contains function calls, which have a high, but constant cost in Python. Probably obvious, but the more conditions are added to the *chain*, the slower it gets.

Using a dictionary not only buys time, it also buys cleaner code, as the handling for every event is in a separate function / method.

akshayaurora:
@snappy666 ok, this is surprising that the added function does not effect the result as much as the change from if to dict, I didn't not expect this.

Considering the result I guess moving to a dict + function makes sense. Just make sure to avoid functions inside functions, that makes things difficult to debug, read. Looking forward to your new pr.

akshayaurora:
@snappy666 I hope you don't mind I closed this pr as you agreed to separate the change for move to dict from the joypad changes. Please reopen their pr with the proposed changes. or a new one if you prefer that.

snapy666:
@akshayaurora Okay! :)

Regarding creating a new pull request (for now only [window_sdl2.py](https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L245)):
* By *"Just make sure to avoid functions inside functions"* did you mean, that the event handlers should be methods of `WindowSDL` instead of local functions?

* Since you both mentioned these concerns: @kived and @Kovak — Are okay with that too? I don't want to waste my time creating something that will not be used.
* If the event name is, for example, `mousemotion` — should the name be `handle_mousemotion` or `on_mousemotion`?
* Are you all okay with a `defaultdict` with the name `event_to_handler`?
* In the case of the following code — is there a reason to keep it? That would mean inserting empty functions into the dictionary for each event.

```python
    elif action in ('fingermotion', 'fingerdown', 'fingerup'):
        # for finger, pass the raw event to SDL motion event provider
        # XXX this is problematic. On OSX, it generates touches with 0,
        # 0 coordinates, at the same times as mouse. But it works.
        # We have a conflict of using either the mouse or the finger.
        # Right now, we have no mechanism that we could use to know
        # which is the preferred one for the application.
        #SDL2MotionEventProvider.q.appendleft(event)
        pass
```
* Should the local `key_swap` dictionary (defined [here](https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L360)) be a member of `WindowSDL` (or a constant)? I could just leave it like that, but that seems like a waste of time, because the dictionary would be re-created every time.

matham:
There are multiple issues with the optimization you proposed. First, though, your test code is needlessly complex and it's hard to know for sure what actually the result means. Here is some simpler test code: https://gist.github.com/matham/7a995c8f9eb7774ee797 showing that for at least 15 test conditions, if/elif is always faster than dict+func.

Among the reasons is that you used floats and strings, but the actual event tests are ints (I'm pretty sure) and for that an if statement is faster. In fact, if we replaced all the 

Overall, though, the test code only shows how dict+func compares to if/elif in the test code, you'd have to show that the actual code performs similarly because the actual code has a lot more stuff it is doing so all these optimizations might not even be visible. In particular, in my personal opinion, having a ton of defined functions with a extra layer of indirection which makes the code more complex would only be a justified change if there is a significantly demonstrable performance difference.

snapy666:
@matham [window_sdl2.py](https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L245) uses `str`ings.

These are my test results with [your code, slightly modified and updated to Python 3](https://gist.github.com/snapy666/83ef8263b03f40f167f0):

```
# number of keys: 15
dict+func    8.917926
if/elif      9.809378

# number of keys: 11
dict+func    8.854053
if/elif      9.827428

# number of keys: 9
dict+func    8.850366
if/elif      9.938295

# number of keys: 6
dict+func    9.078270
if/elif      10.301112
```

matham:
@snapy666 Running your code I get the following result:

```
('# number of keys:', 15)
dict+func    10.854189
if/elif      10.678702
()
('# number of keys:', 11)
dict+func    10.864527
if/elif      10.576125
()
('# number of keys:', 9)
dict+func    11.082304
if/elif      10.655297
()
('# number of keys:', 6)
dict+func    11.015001
if/elif      10.500294
()
```
This could be a py2/3 difference.

About why this might not be representative of the actual code. If your changes does not lead to a large perf change in the example, then in the real code you might not see any difference at all. The reason is that the real code does more things when handling the events, which is probably order or magnitudes slower than any effect due to using dict+func or if/elif. So all of this might not make any difference and hence why I said you'd need to show actual performance difference for this particular code to justify adding more complexity to the code.

snapy666:
> The reason is that the real code does more things when handling the events, which is probably order or magnitudes slower than any effect due to using dict+func or if/elif. So all of this might not make any difference [...]

Okay, but what you are implying by that is that it indeed has an effect, but it just might be a very small change, barely noticeable. I've never said that it would be an enormous speed improvement.

> [...] adding more complexity to the code.
The code complexity added is minimal, but the code is certainly cleaner, because there's no giant `if-elif-else` chain with lots of `elif action == 'eventname'`. Every event has it's method. Oh.. I just got an idea: By using a `dict`, you could replace parts of the event handling without changing the code (`if-elif-else`).

The time complexity is actually lower. Instead of `O(n)` we have about `O(1)` (constant time).

matham:
That's a good reason for doing it the way you suggested, if you can formalize it.

snapy666:
@matham Cool.
> if you can formalize it

What do I need to do exactly?

snapy666:
@matham Yeah :), and I didn't even think of this, when I created this PR. What do you mean with "formalize"?

matham:
@snapy666 I meant make a proper api for updating the dict. But this is not my particular neck of the woods so others may have a better idea about it.

snapy666:
@akshayaurora
> Just make sure to avoid functions inside functions

* Do you mean that the event handlers should be methods of `WindowSDL` and not of local functions, right?
* If the event name is, for example, `mousemotion` — should the name be `handle_mousemotion` or `on_mousemotion`? Should they be private?
* Are you okay with a `defaultdict` with the name `event_to_handler`?
* Should the local `key_swap` dictionary (defined [here](https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L360)) be a member of `WindowSDL` (or a constant)? I could just leave it like that, but that seems like a waste of time, because the dictionary is re-created every time.

@kived @Kovak 
If I create a new pull request for [window_sdl2.py](https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L245), would each of you accept it? I don't want to spend my time for something, if there's no chance of it being accepted.

Issue #2807, opened 2014-12-31
Sdl2 window crashes on scroll


inclement:
I tested only with a trackpad and trackpoint (not scroll wheel), but I think it's the scrolling that's the problem.

The traceback is

    Traceback (most recent call last):
      File "main.py", line 212, in <module>
        ShowcaseApp().run()
      File "/home/asandy/kivy/kivy/app.py", line 825, in run
        runTouchApp()
      File "/home/asandy/kivy/kivy/base.py", line 484, in runTouchApp
        EventLoop.window.mainloop()
      File "/home/asandy/kivy/kivy/core/window/window_sdl2.py", line 482, in mainloop
        self._mainloop()
      File "/home/asandy/kivy/kivy/core/window/window_sdl2.py", line 308, in _mainloop
        self._mouse_x, self._mouse_y, btn, self.modifiers)
    AttributeError: 'WindowSDL' object has no attribute '_mouse_x'

Edit: I haven't tested extensively, but it seems to stop happening after some action is performed - for instance, if I click and drag a scrollview *then* scroll with two fingers, the behaviour is not triggered. I guess this may be related to my sdl2 multitouch fix meaning code for mouse movement isn't run any more (where before it was on every movement even without touch), though it's a bug anyway.

Edit: I think I can confirm how to reproduce, you must scroll *before* any other touch event.

PR #2808, opened 2014-12-31
Sdl2 scroll fix

inclement:
This resolves https://github.com/kivy/kivy/issues/2807

The problem seems to occur because sdl2 treats the scroll event differently to a touch. In pygame, the _mouse_x and _mouse_y are set in the mouse down event handler, which occurs even if the 'button' is the scroll wheel, but with sdl2 the scroll event doesn't receive the mouse position as input.

I couldn't think of a better fix than setting it on every mouse move, though I also set it on every mouse down/up because I'm not clear if we're guaranteed to receive movement events for some devices that otherwise appear as mice. Maybe this isn't necessary at all.


akshayaurora:
@inclement :+1:

Issue #2791, opened 2014-12-25
KivyCatalog: Selectors crash


matham:
Open kivycatalog, open Selectors, click on dropdown menu, and click on any of the options - results:

```
[INFO              ] Kivy v1.9.0-dev
[INFO              ] [Python      ] v2.7.8 (default, Jun 30 2014, 16:03:49) [MSC v.1500 32 bit (Intel)]
[INFO              ] [Factory     ] 173 symbols loaded
[INFO              ] [Image       ] Providers: img_tex, img_dds, img_pygame, img_gif (img_sdl2, img_ffpyplayer, img_pil ignored)
[INFO              ] [Text        ] Provider: pygame
[INFO              ] [Window      ] Provider: pygame(['window_egl_rpi'] ignored)
libpng warning: iCCP: known incorrect sRGB profile
GLEW initialization succeeded
[INFO              ] [GL          ] OpenGL version <3.3.0>
[INFO              ] [GL          ] OpenGL vendor <NVIDIA Corporation>
[INFO              ] [GL          ] OpenGL renderer <GeForce 8600M GT/PCIe/SSE2>
[INFO              ] [GL          ] OpenGL parsed version: 3, 3
[INFO              ] [GL          ] Shading version <3.30 NVIDIA via Cg compiler>
[INFO              ] [GL          ] Texture max size <8192>
[INFO              ] [GL          ] Texture max units <32>
[INFO              ] [Window      ] virtual keyboard not allowed, single mode, not docked
[INFO              ] [Video       ] Provider: null(['video_pygst', 'video_ffmpeg', 'video_ffpyplayer', 'video_pyglet'] ignored)
[INFO              ] [GL          ] NPOT texture support is available
[ERROR             ] [filechooser ] win32file module is missing
[ERROR             ] [filechooser ] we cant check if a file is hidden or not
[INFO              ] [OSC         ] using <thread> for socket
[INFO              ] [Base        ] Start application main loop
libpng warning: iCCP: known incorrect sRGB profile
[INFO              ] [Context     ] Reloading graphics data...
[INFO              ] [Context     ] Reloading done in 0.1630s
[INFO              ] [Base        ] Leaving application in progress...
 Traceback (most recent call last):
   File "G:\Python\dev2\kivy\examples\demo\kivycatalog\main.py", line 166, in <module>
     KivyCatalogApp().run()
   File "g:\python\dev2\kivy\kivy\app.py", line 825, in run
     runTouchApp()
[...]

PR #2813, opened 2015-01-01
Kivycatalog fix

kived:
Fixes an issue with how KivyCatalog reloads when changing screens, which I think may have caused #2791. Also prevents `Spinner` from trying to reopen itself when selecting an item (on select changes is_open to False which calls dismiss which triggers on_dismiss which calls toggle_dropdown setting is_open back to True).


akshayaurora:
@kived :+1:

PR #2819, opened 2015-01-02
ProbeSysfs: ignore devices which look like they are used as pointers

kived:
Checks with xinput to find any devices which are used as pointers (mice), so they can be excluded from ProbeSysfs. This should solve issues with multitouch touchpads being used by multiple providers.


akshayaurora:
@kived :+1:

PR #2824, opened 2015-01-04
Removed 'this is experimental' tags from a couple of places

inclement:
We have this in a few places, mostly still justified, but I think these two should be removed.

Screenmanager is stable and used by many people (the main change I think is worth making is renaming switch_to or modifying its behaviour for sceens that are already added), and I've seen users discouraged by it being labelled this way.

Pause mode is similarly popular and seems to work fine. I'm not sure if the warning about crashes is still needed, but if so I think it should be instead of the experimental label.

### DIFF
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -264,11 +264,6 @@ Pause mode
 
 .. versionadded:: 1.1.0
 
-.. warning::
-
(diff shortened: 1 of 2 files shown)


dessant:
:+1:

PR #2825, opened 2015-01-05
Horizontal align for popup title

Cuuuurzel:
This will allow to center the text of the popup title.
popup-title-centered

Can be used just by passing 'title_align' to the Popup options.


dessant:
@Cuuuurzel good job. Could you also add the versionadded (1.9.0) tags?

PR #2830, opened 2015-01-07
Activate SDL2 compilation and usage by default

tito:
This PR includes:

- usage of OSX Frameworks by default instead of the default method with pkgconfig. This implies that the requirements (SDL2, SDL2_ttf, SDL2_mixer, SDL2_image, GStreamer) must be installed into /Libraries/Frameworks. This can be disabled with USE_OSX_FRAMEWORKS=0.
- generate a kivy/setupconfig.py, the python equivalent of the kivy/graphics/config.h / config.pxi
- use the setupconfig.USE_SDL2 to determine if pygame is allowed in the core provider (audio/text/window/image).


tito:
@akshayaurora about the #2818, did you have tested on mobile, or is it just a guess? Cause unless it's tested, i don't think it will be sufficient, and so, the PR could wait until the mobile is working too

Issue #2823, opened 2015-01-04
White square when loading zip Image on Windows (1.9.0-dev)


victor-rene:
Zip loading works in 1.8.0, must be related to #1730.
On a side note #2102 might also be related, although it's likely buggy in both versions. Needs confirmation, but it all comes down to reloading and refreshing of zipped images. I think some code must be written to trigger the update when the source isn't a static file but a collection.

EDIT. A zip file: https://drive.google.com/open?id=0B3-RtQdBRQiOMHJ4SFRodVZWNE0&authuser=0

tshirtman:
can you run the example?
```
python examples/widgets/sequenced_images/main.py
```
seems to work for me (linux)

victor-rene:
Running "python examples/widgets/sequenced_images/main.py". The fifth widget on the left bar is a white square on Windows (1.9.0-dev).

tshirtman:
actually works for me after `make force` on master, may be the result of a faulty compilation.

bionoid:
It seems like my system has the same problem as @victor-rene for sequenced_images example - fifth widget is white.

"Load gif" button works; both "zipped png's" and "zipped jpg's" produce a white square only. The remaining 4 buttons all crash the app with this:

```
     _animdelay = self.img._coreimage.anim_delay
 AttributeError: 'NoneType' object has no attribute 'anim_delay'
```

victor-rene:
@bionoid I wrote about the _coreimage problem yesterday in IRC, it fell into dead ears. I don't know if opening another issue is the right thing to do, but a user of AsyncImage was searching for a way to bypass caching (finally solved with setting nocache=True). Meanwhile I suggested self.img._coreimage.remove_from_cache() which caused the same error message AttributeError: 'NoneType' object has no attribute 'remove_from_cache'. I feel like something is going on here.

akshayaurora:
@victor-rene you have http://kivy.org/docs/api-kivy.uix.image.html#kivy.uix.image.Image.nocache and http://kivy.org/docs/api-kivy.uix.image.html#kivy.uix.image.Image.reload both mentioned in the docs, the issue with .zip is not reproducible for me on windows 8.1 with the latest master.

As tshirtman outlined doing `make force` to make sure you have everything compiled for master does not reproduce the issue. On irc you simply didn't respond the question when I asked whether you had done `make force`. Please give us more details on whether this is working with make force and which version of windows are you using.

victor-rene:
[...]

PR #2834, opened 2015-01-09
Fixes 2823, zip files not loaded by ImageLoaderPygame.

victor-rene:
Must be tested on all other platforms. Was developed with python 2.7.9, Windows 7 Home 64-bit.
See #2823 for details.


kived:
Works on Linux with pygame. :+1:

victor-rene:
Okay, I'm setting up a VM for SDL2. I don't know how much time it takes though, don't hold your breath.

Issue #2723, opened 2014-12-04
Label rendering glitch using Pygame / Unicode text / Custom font


tito:
Simple text case, if you render a label with "Hello " (note the space at the end), some font will display a box. The new changes in master replace the space before/after the text with unbreakable space unicode. But they don't replace back with standard space when rendering, which give us a beautiful box :)

This can be seen with a font like MrEavesSanOT-Reg.otf (http://www.myfonts.com/fonts/emigre/mr-eaves-sans-opentype/ot-reg/specs.html)

The current fixes i did in pygame is:

```
diff --git a/kivy/core/text/text_pygame.py b/kivy/core/text/text_pygame.py
index abbf2dc..7b7a58a 100644
--- a/kivy/core/text/text_pygame.py
+++ b/kivy/core/text/text_pygame.py
@@ -92,6 +92,7 @@ class LabelPygame(LabelBase):
         self._pygame_surface.fill((0, 0, 0, 0))
 
     def _render_text(self, text, x, y):
+        text = text.replace(u"\u00a0", " ")
         font = self._get_font()
         color = [c * 255 for c in self.options['color']]
         color[0], color[2] = color[2], color[0]
```

But i guess that was not the right place to put it. Any advice? @matham ?

kived:
So, I think we should either:

1) Merge my PR, close this issue, and open another issue to review this for 1.9.1.
2) Revert my previous PR which caused this issue, reopen the previous issue that PR fixed, and tag it 1.9.1.
3) Someone (@matham?) submit a replacement for my previous PR and merge it today.

matham:
I'll look into adding a variable called `internal_strip`, which will allow not stripping only internal spaces.

matham:
It's actually quite difficult to do this with markup, because it's not straight forward to figure out what is the before/after whitespace, since there could be markup text.

But also, I'm not sure that always stripping internal text is always correct. E.g. when `size_hint_x` is `None`, the user knows exactly how it'd layout each line. So I'm wondering whether we should not go back to how it was and instead fix the original problem.

matham:
This should now be fixed with #2840.

akshayaurora:
closed  by matham in #2840

Issue #2503, opened 2014-09-23
Label wrap are wrong sometimes


tito:
This wasn't an issue in 1.8 before, but it is now:

```python
from kivy.base import runTouchApp
from kivy.uix.label import Label
runTouchApp(Label(
    text="This is a big line of text hell yeah.",
    text_size=(100, None)))
```

Result:
![image](https://cloud.githubusercontent.com/assets/37904/4371418/41d063f4-4317-11e4-945d-309d92adc644.png)

matham:
Reopened to be fixed by #2840.

akshayaurora:
closed by #2840

PR #2840, opened 2015-01-10
Fix text stripping issues

matham:
This fixes both #2503 and #2723.

When the universe began, after my changes to text a while back, there was a new issue, where when `strip` was False, lines split because `size_hint_x`, was not `None` may have began with a space (#2503). @kived fixed it ( #2677) by always stripping internal spaces. This had a few issues, in particular, it didn't work for markup and it lead to #2723.

This pr, reverts  #2677 and instead provides an explicit option called `strip_reflow` which when `strip` is False, and `size_hint_x` is not None, will strip only the ends and starts of lines that are broken due to reflow. This will not strip lines that are not broken, because I believe if the user wanted to strip such lines, they should have selected `strip` to be True. Only lines that they could not know will be split due to size constraints will be stripped. I think this is the best of both worlds.

Please test to see if #2503 is still fixed and then merge :)

__Note__ #2502 should be merged first, because this branch starts after that. Only the last two commits are new for this issue.

### DIFF
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -550,23 +550,14 @@ class LabelBase(object):
                                     options['halign'][-1] == 'y')
         uw, uh = options['text_size'] = self._text_size
         text = self.text
-        if not strip:
-            # all text will be stripped by default. unicode NO-BREAK SPACE
-            # characters will be preserved, so we replace the leading and
-            # trailing spaces with \u00a0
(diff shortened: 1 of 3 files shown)


matham:
Actually, there's one thing I wasn't sure about. I only added `strip_reflow` to `Label` because I didn't think it's something a user would need to change. It's only added to the core options. Should I also add it to `Label`?

akshayaurora:
@matham good job!

PR #2843, opened 2015-01-12
Added test/test_utils.py. BUGFIXES: escape_markup, iterate (deprecated)

victor-rene:
This brings coverage for the utils.py module to 91%. Reason is Platform class is untestable except for the return values of _get_platform.

BUGFIX 1 (escape_markup)
Since '&' character is part of the escape sequence you must use it first otherwise, it gets added more then needed. Example:

    '[1]' -> '&amp;bl;1&amp;br;'
    instead of
    '[1]' -> '&bl;1&br;'

BUGFIX 2 (iterate)
You cannot reverse an iterator.

### DIFF
--- a/kivy/tests/test_utils.py
+++ b/kivy/tests/test_utils.py
@@ -1,0 +1,152 @@
+'''
+utils tests
+===========
+'''
+
+import unittest
+
+from kivy.utils import (boundary, escape_markup, format_bytes_to_human,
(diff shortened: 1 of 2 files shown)


akshayaurora:
@victor-rene dude awesome job !

victor-rene:
@akshayaurora  I found a way to mock os.environ, I will probably make another PR if it works and bring coverage to 100%

Issue #2861, opened 2015-01-16
Consider refactoring platform detection code.


victor-rene:
Not a real issue per see, but there are a few things that feels a bit off. I create the issue to discuss the changes. Follwing with PR if there are no objections.

Remove globals
---
My biggest gripe: https://github.com/kivy/kivy/blob/master/kivy/utils.py#L272-278
Those two globals should live inside the Platform class. As they are now, during testing I have to hard reset those to None with side effects on the following tests. It would make more sense to create a platform object and discard it without side effects. I did not find code that makes use of those two global flags.

    def test_Platform_android(self):
        with patch.dict('os.environ', {'ANDROID_ARGUMENT': ''}):
            # must hard reset globals
            # import kivy.utils
            # kivy.utils._platform_android = None
            # kivy.utils._platform_ios = None
            pf = Platform()
            self.assertTrue(pf == 'android')
        self.assertNotIn('ANDROID_ARGUMENT', os.environ)

Remove unused class var
---
A quick [query](https://github.com/kivy/kivy/search?utf8=%E2%9C%93&q=_platform) shows that Internal variable _platform does not need to exist and serves no purpose. I could be wrong but I doubt it.

Use of sys.platform
---
Also I should praise the use of _sys_platform instead of a direct call to sys.platform. :+1:

PR #2862, opened 2015-01-16
Updated test of kivy.utils to 100% (platform detection)

victor-rene:
Do not merge if you believe people still use _platform_android and _platform_ios instead of platform.
Details in #2861.


dessant:
Those alone shouldn't be a problem because they are private vars.

dessant:
:+1:

PR #2864, opened 2015-01-17
lang: fix_double include. closes #2821

### LINKED ISSUE #2821: Kv "include" directive with reload is loading twice
By looking at the code in the `lang.py`:

```python
                if ref in __KV_INCLUDES__:
                    if not force_load:
                        Logger.warn('WARNING: {0} has already been included!'
                                    .format(ref))
                        break
                    else:
                        Logger.debug('Reloading {0} because include was forced.'
                                    .format(ref))
                        Builder.unload_file(ref)
                        Builder.load_file(ref)
                if not os.path.isfile(ref):
                    raise ParserException(self, ln, 'Invalid or unknown file: '
                                                    '{0}'.format(ref))
                Logger.debug('Including file: {0}'.format(0))
                __KV_INCLUDES__.append(ref)
                Builder.load_file(ref)
```

If the file was already loaded, it will be unloaded once, but loaded twice after. I think that's an issue :)

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/7457491-kv-include-directive-with-reload-is-loading-twice?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

akshayaurora:


kivy/lang.py
```diff
@@ -1199,9 +1202,7 @@ def execute_directives(self):
                                     .format(ref))
                         Builder.unload_file(ref)
                         Builder.load_file(ref)
-                if not os.path.isfile(ref):
-                    raise ParserException(self, ln, 'Invalid or unknown file: '
-                                                    '{0}'.format(ref))
+                        break
```
dessant: Why do you break here?

PR #2866, opened 2015-01-17
image_overlay_play and image_loading attributes of VideoPlayer fixed

robertour:
The image_overlay_play and image_loading attributes of VideoPlayer were not working. 

The paths for the images in VideoPlayerPreview were hard coded. I created the respective attributes and associate them to the existen attributes in the VideoPlayer class with the definition of the on_image_overlay_play and on_image_loading attribute methods.


akshayaurora:
:+1:

PR #2874, opened 2015-01-21
Simple, easy way to use different pygments styles for the CodeInput widget.

LogicalDash:
CodeInput now has a property ```style_name```. Set it to some style that pygments has. Now the highlights will be in that style.


kivy/uix/codeinput.py
```diff
@@ -138,7 +148,9 @@ def _get_bbcode(self, ntext):
             # replace brackets with special chars that aren't highlighted
             # by pygment. can't use &bl; ... cause & is highlighted
             ntext = ntext.replace(u'[', u'\x01;').replace(u']', u'\x02;')
-            ntext = highlight(ntext, self.lexer, self.formatter)
+            ntext = highlight(ntext, self.lexer, BBCodeFormatter(
```
akshayaurora: doing this every time a piece of text needs rendering would be super expensive, instead of doing this, use on_style

```
def on_style(self, instance, value):
    self.formatter = BBCodeFormatter(style=styles.get_style_by_name(self.style_name))
```

This was you get the style changed only when the property changes. No need to instantiate it every time a text is being rendered.

LogicalDash:
I think I'm about done with this. I was thinking of adding properties to override parts of the style, but that turns out to be a pretty big project.

akshayaurora:
@LogicalDash :+1:

Issue #2882, opened 2015-01-23
Detection via pkgconfig does not work with launchpad.net (PPA)


thopiekar:
Just looked at our build whether the last changes on gstreamer worked in setup.py, but does not take any effects according to the latest buildlog:
https://launchpadlibrarian.net/190496775/buildlog_ubuntu-vivid-amd64.kivy_1.9.0-dev~testing0%2B201411161734-2438-testing45~ubuntu15.04.1_UPLOADING.txt.gz

thopiekar:
Looks like my code fortunately fixed that problem.
https://code.launchpad.net/~kivy-team/+archive/ubuntu/kivy-daily/+build/6745997

PR #2883, opened 2015-01-23
Print warning when executed shell command does not return code 0

thopiekar:
- Might be useful for https://github.com/kivy/kivy/issues/2882


tito:
Yes and no, as even if pkg-config doesn't return appropriate flag, it can be OK for the targetted architecture. (like not detecting gstreamer-1.0 is OK).

So i'm unsure about commit this as it is, i'm 200% users will think something is wrong by default.

thopiekar:
I see.. But do most of the users even care about those messages?
How often issues appear where users don't read the output and python really gives exact messages what the problem is.
At the end it will definitly help us to debug bugs and as you can see in my testing builds my changes make our setup.py detect SDL2 and gstreamer-1.0 on launchpad. Probably because of the .wait() or using the stdout.read() directly.

tito:
Ok. :)

PR #2886, opened 2015-01-23
Added miscellaneous examples folder and a first entry

inclement:
I think it would be useful to have a miscellaneous examples folder in which to put short example scripts that don't explicitly demonstrate a specific feature. This is partly because we make quite a lot of these and they're great resources but mostly get forgotten after a while, and partly because I'd like to have them contribute to an example gallery page on the website, as has been discussed before.

So...this PR adds a first random script I made to that folder.

### DIFF
--- a/examples/miscellaneous/two_panes.py
+++ b/examples/miscellaneous/two_panes.py
@@ -1,4 +1,4 @@
-''
+'''
 'Demonstrates using kv language to create some simple buttons and a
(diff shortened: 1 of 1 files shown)


dessant:
+1 for this. @inclement, could you fix the quotes on line 1?

inclement:
Good catch, updated.

dessant:
:ice_cream:

PR #2891, opened 2015-01-24
Change the default font from DroidSans to Roboto

dessant:
Besides the well-known advantages of Roboto, it also has better char coverage, we already use it for the italic files because DroidSans does not have italic versions.

Glyph counts:

```sh
# Current default fonts
898  DroidSans.ttf
898  DroidSans-Bold.ttf
949  DroidSans-Italic.ttf (old version of Roboto-Italic)
946  DroidSans-BoldItalic.ttf (old version of Roboto-BoldItalic)

# Latest version of Roboto
1072  Roboto-Regular.ttf
1072  Roboto-Bold.ttf
1073  Roboto-Italic.ttf
1073  Roboto-BoldItalic.ttf
```

Roboto files taken from http://www.google.com/design/spec/style/typography.html.


dessant:
@tito what do you think?

tito:
What about the size? But +1 anyway.

dessant:
@tito, there are slight size differences (first line is Roboto), but i don't see it as a problem.

screenshot from 2015-01-24 18 36 58

What happens with the droid files, do we keep them in the repo? How about DejaVuSans.ttf?

dessant:
Maybe we could also change our default text_size from 15sp to 14sp.
14sp is recommended by Google for button and body text.

screenshot from 2015-01-24 19 04 18

tito:
My bad, i was talking about size on the disk, as it supports more char, it might be increased a little bit. But i'm unsure.

By the way, i wonder if we could use the default android font instead of packing it into the apk... Might be an optimization to do one day :)

dessant:
@tito the 4 droid files are 547 kB, while the 4 new roboto files are 520 kB combined.

If this added space is a concern, we can either remove the droid files, or delete DejaVuSans.ttf, which is 581 kB. DejaVuSans was replaced as a default font years ago, we shouldn't ship it.

akshayaurora:
For using default android fonts,  we already have a fund in core.text get_system_fonts_dir https://github.com/kivy/kivy/blob/master/kivy/core/text/__init__.py#L160, we theoretically should just need to set "font_file_name"  and if it exists we would be able to use it.

akshayaurora:
@tito @dessant  just tested a apk with  font_name: 'ComingSoon' on android using kivy==master. It just works for us. No need for including a default font! \o/ :dancer:

tito:
the VKeyboard still reference the DejaVuSans. Ensure it works with Roboto (ie that all the glyph used in the default keyboard works)

dessant:
meh, it looks like dejavu stays

screenshot from 2015-01-25 01 03 46

Issue #2889, opened 2015-01-24
Making sphinx-build more failsave


thopiekar:
Also found this problem with ```sphinx-build -b latex```.
Wouldn't it be better to remove the ```-W``` option to make this process also failsave and add a command with ```-W``` to our test section in https://github.com/kivy/kivy/blob/master/Makefile ?
In my point of view ```-W``` is just interesting, when looking for "small" problems.
Additionally it will make these ugly lines at https://github.com/kivy/kivy/blob/pkg-debian-daily/rules#L32 unneeded.

Of course I will modify https://github.com/kivy/kivy/blob/master/Makefile if you like the idea.

```
reading sources... [ 99%] tutorials/firstwidget
reading sources... [ 99%] tutorials/pong
reading sources... [100%] user-guide


Warning, treated as error:
/build/buildd/kivy-1.9.0-dev~testing0+201501241054-2560/doc/sources/api-kivy.setupconfig.rst:1: ERROR: Document or section may not begin with a transition.

Makefile:82: recipe for target 'latex' failed
make[2]: *** [latex] Error 1
make[2]: Leaving directory '/build/buildd/kivy-1.9.0-dev~testing0+201501241054-2560/doc'
```
- https://launchpadlibrarian.net/195700247/buildlog_ubuntu-vivid-amd64.kivy_1.9.0-dev~testing0%2B201501241054-2560-testing48~ubuntu15.04.1_UPLOADING.txt.gz

PR #2894, opened 2015-01-24
Make doc build failsave

thopiekar:
* Includes fix for #2889
* Made autobuild.py more silent
* Moved make pdf to doc/Makefile and made it failsave as long as file is created
* Made use of some options making sphinx-build and pdflatex more silent and non-interactive (user might be confused to be prompted to make any inputs)
* ENDUSER_BUILD=no will skip all my changes (in case there is a purpose to have all these long outputs)


tito:
Looks good, thanks !

PR #2905, opened 2015-01-26
Disable input and postproc by omiting config file sections

derekstavis:
I don't know if its the most elegant way of doing this, but it definitely works for me.

### DIFF
--- a/kivy/input/__init__.py
+++ b/kivy/input/__init__.py
@@ -23,6 +23,11 @@ several default modules for :
 * Decreasing jittering
 * Decreasing the inaccuracy of touch on "bad" DIY hardware
 * Ignoring regions
+
+If your application don't use input or post processing you can remove the
+respective config file sections. If you remove just `[input]` section, it will
+disable `[postproc]` as a side-effect. Removing just `[postproc]` section will
+prevent only postproc modules to load.
 '''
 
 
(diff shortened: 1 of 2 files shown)


KeyWeeUsr:
Seems good to me if it leaves the default behavior untouched (just an option to disable).

kivy/input/__init__.py
```diff
@@ -23,14 +23,26 @@
 * Decreasing jittering
 * Decreasing the inaccuracy of touch on "bad" DIY hardware
 * Ignoring regions
+
+If your application don't use input or post processing you can remove the
```
KeyWeeUsr: Doesn't/won't/...

kivy/input/postproc/__init__.py
```diff
@@ -13,12 +13,17 @@
 from kivy.input.postproc.retaintouch import InputPostprocRetainTouch
 from kivy.input.postproc.dejitter import InputPostprocDejitter
 from kivy.input.postproc.calibration import InputPostprocCalibration
+from kivy.config import Config
+
+# Enable postproc if the section exists in config file
+_postproc_enabled = (len(Config.items('input')) > 0 and
+                     len(Config.items('postproc')) > 0)
```
tshirtman: len(x) > 0 -> any(x)

matham:
The Zen of python says `Explicit is better than implicit.`. With that change, removing a section implicitly disables it which I think is incorrect. There should be a way to explicitly disable them if they need to be disabled. Speaking of which, why do you want to disable it?

Also, they should be able to be disabled independently.

tshirtman:
Closing until a rational for this behavior is provided, as merging this could certainly have surprising effects (double touch postproc anyone?).

PR #2906, opened 2015-01-26
Don't break application on time changes

derekstavis:
Python's standard library `time.time` function relies on `gettimeofday`, a clock source sensitive to system/ntp time changes, causing Kivy applications to freeze when time goes backwards. 

This patch makes use of `clock_gettime` to retrieve a always increasing time via `CLOCK_MONOTONIC` or `CLOCK_MONOTONIC_RAW` (a Linux-specific clone) which aren't affected by system/ntp time changes.

### DIFF
--- a/kivy/clock.py
+++ b/kivy/clock.py
@@ -243,16 +243,46 @@ try:
     else:
         if platform == 'darwin':
             _libc = ctypes.CDLL('libc.dylib')
+            _default_time = time.time
         else:
-            _libc = ctypes.CDLL('libc.so')
+            from ctypes.util import find_library
+            _libc = ctypes.CDLL(find_library('c'), use_errno=True)
+
+            def _libc_clock_gettime_wrapper():
+                from os import strerror
+
+                class struct_tv(ctypes.Structure):
+                    _fields_ = [('tv_sec', ctypes.c_long),
+                                ('tv_usec', ctypes.c_long)]
+
+                _clock_gettime = _libc.clock_gettime
+                _clock_gettime.argtypes = [ctypes.c_long,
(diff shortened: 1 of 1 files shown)


gtcarlos:
:+1:

davidbranco:
That's what I'm talking about! Nice one @derekstavis

derekstavis:
I'm happy the patch helped you guys. Thanks for testing it :)

akshayaurora:
@derekstavis  thanks, :+1:

dessant:
@derekstavis, I've tried to reproduce a freeze on linux when setting the time back, with a sample app, but I couldn't do it. Would you share an example or at least tell us how te reproduce the bug? I would like to document it in a new issue.

I'm not sure if you're aware of it, but the pr was patched because of https://github.com/kivy/kivy/issues/3797, so now it mostly just falls back to the old implementation, I wasn't able to use it on debian and travis also failed with it, which is ubuntu.

@akshayaurora, you mentioned running it in a vm, could you tell us the distro and version you used?

derekstavis:
Hi @dessant. I have reproduced the ~~break~~ freeze sucessfuly on Kivy 1.9.0 by using `Clock.schedule_interval` to blink a digital clock separator. When using `date` command line utily to set date before current one (backwards) the separator stopped blinking.

derekstavis:
About the patch, yeah, it looks like Android's Bionic doesn't export `clock_gettime`. I have tested this PR in both OS X and Linux, and have been using it in production Linux even before the PR was opened here. I have debugged it using `strace` to make sure no `gettimeofday` call was being done.

PR #2908, opened 2015-01-26
Fix binding issues

matham:
This resolves two issues:

- When rebinding, we unbound too far and didn't account for it. It is fixed here not unbind more than needed. The following code demonstrates the issue:
```py
from kivy.lang import Builder
from kivy.properties import ObjectProperty, NumericProperty

class MyWidget(Widget):
    num = NumericProperty(0)
    prop = ObjectProperty(None, rebind=True)

class MyWidget2(Widget):
    p = ObjectProperty(None, rebind=True)

class MyWidget3(Widget):
    p2 = ObjectProperty(None, rebind=True)

kv = '''
MyWidget:
    num: self.prop.p.p2.num if self.prop and self.prop.p else 5.3
    on_num: print(self.num)
'''

w1 = MyWidget2(p=MyWidget3(p2=MyWidget(num=1)))
w2 = MyWidget3(p2=MyWidget(num=2))
root = Builder.load_string(kv)
root.prop = w1
w1.p.p2.num = 1.5
root.prop.p = w2
w2.p2.num = 2.5
```
On master it prints:
```
1
1.5
2
```

but on the pr it prints correctly:
```
1
1.5
2
2.5
```
- When a rule said e.g. `MyWidget.obj.prop` When storing `MyWidget` it'd try `getattr(f, 'proxy_ref', f)` where `f` was the `MyWidget` class. `proxy_ref` returned the property itself, because `MyWidget` is a class, which is wrong, so the code fixes to skip this on classes.

### DIFF
--- a/kivy/lang.py
+++ b/kivy/lang.py
@@ -1512,10 +1512,11 @@ def update_intermediates(base, keys, bound, s, fn, args, instance, value):
     del bound[s:]
 
     # find the first attr from which we need to start rebinding.
-    if len(bound):
-        f = bound[-1][0]
-    else:  # if it's the very first attr, we start with the base.
-        f = base
+    f = getattr(*bound[-1][:2])
+    if f is None:
+        fn(args, None, None)
+        return
+    s += 1
     append = bound.append
 
     # bind all attrs, except last to update_intermediates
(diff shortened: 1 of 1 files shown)


dessant:
@quanon ping

akshayaurora:
:+1:

PR #2911, opened 2015-01-26
Fix licensing and extras [2/2]

thopiekar:


dessant:
@thopiekar both docs and kivy build correctly on my part. :+1:

PR #2913, opened 2015-01-26
Updated to correct cython version

ohaz:
Was cython >= 0.15, now is cython==0.20.
A lot of people complained that they didn't manage to install kivy because they installed cython 0.21


dessant:
:+1:

dessant:
The problem is that buildbot was broken/not runnig for some time, so they will not see this until it gets fixed.

ohaz:
Can someone fix it? If it doesn't take too much time.

PR #2920, opened 2015-01-27
Proposed wrapping gl error

olymk2:
This is convenient because you can place ipdb in the new method or other commands, also removes duplication.


inclement:
Could you declare glDebugMessage as inline to avoid the cost of the function call?

@matham also suggested it would be better to replace the print with a debug message, though I'm not sure if that need hold up merging it since it was printing anyway.

Thanks for the contribution though, and sorry for the extreme delay in reviewing it.

olymk2:
no worries at least it getting looked at, by inline do you mean something like cdef inline glDebugMessage(GLenum ret) 

I am happy to replace the print statement but dont know kivy well enough to know what to replace it with.

Kovak:
This is something that's been in the back of my mind recently it would be nice to have more thorough error reporting in general. It would be great if we could have the error checking be a compile time addition like the way OGRE macros it. Not certain if we can achieve the same in cython. For reference:

https://bitbucket.org/sinbad/ogre/src/aa7a1ea160eddccaa362e51525a37fb53f1ff13f/RenderSystems/GLES2/include/OgreGLES2Prerequisites.h?at=default#cl-279

and an example of usage:

https://bitbucket.org/sinbad/ogre/src/aa7a1ea160eddccaa362e51525a37fb53f1ff13f/RenderSystems/GLES2/src/OgreGLES2HardwareVertexBuffer.cpp?at=default#cl-54

tshirtman:
@Kovak is the change you ask for trivial? (adding "inline" in the definition or something), if so, we could do it and merge, if that ease debugging of gl errors, that seems nice to have.

dessant:
This will need a rework after the recent gl changes.

@olymk2, @Kovak: please open a new pr.

PR #2927, opened 2015-01-27
Fix the default of border property (in PageLayout) in the documentation

robertour:
The documentation says that the default value for border is 0, but is 50.


kivy/uix/pagelayout.py
```diff
@@ -52,7 +52,7 @@ class PageLayout(Layout):
     the previous/next page when needed.
 
     :data:`border` is a :class:`~kivy.properties.NumericProperty` and
-    defaults to 0.
+    defaults to 50.
```
dessant: It could even be: defaults to 50dp.

akshayaurora:
@robertour  thanks.

PR #2934, opened 2015-01-28
Add both src and build paths to setup.py for generating setupconfig and co.

matham:
It seems pip still has a problem installing with the master zip file. Not sure what changed, because after my last fix I was able to install from a zip. After the fix, setupconfig and co. was generated in the build path, which is what gets installed to site-packages, rather than the temp location where it's being built (source location).

However, now it seems pip cannot install if these files are not also in the source path, so this changes to generate the files in both the build and source paths, if they are different.


kived:
:+1:

PR #2937, opened 2015-01-29
audio example new filetypes

merriam:
Add public domain music so audio example would test mp3 and ogg files
as well as wav.  Also, improve error messages.


merriam:
Folded into #2938 with fewer git issues.

PR #2948, opened 2015-01-30
Have unittest destroy its test file.

merriam:
test_font.py now deletes it unicode named .ttf file during tear-down


dessant:
:100:

PR #2960, opened 2015-01-31
Add new checks to kivy/tools/pep8.py

merriam:
Check for files ending in multiple blank lines (E389) and using two or
three spaces after a period in a comment.  Also fix a minor issue for
the —show-pep8 option.


dessant:
Awesome! Thanks a lot for this.

PR #2961, opened 2015-01-31
Fix spacing and long lines.  'make style' is now clean.

merriam:
Fix spacing, ending whitespace, and long lines in many files.   Most were issues of having two spaces after the
end of a sentence inside documentation.   Added _webdebugger.py to the ignore list.   'make style' now does have any warnings.


dessant:
@merriam thanks for taking the time to fix this. :dancer:

PR #2962, opened 2015-01-31
Video fixes

matham:
- Fixes an bad super call in uix/video which resulted in the incorrect super being called.
- Fixes a filepath issue with video test.
- Fixes ffpyplayer to use a daemon thread, otherwise kivy doesn't exit if you don't unload directly.
- Fixes ffpyplayer (img and video) to keep forwarding logs to kivy.


akshayaurora:
:+1:

PR #2973, opened 2015-02-01
Create and document examples/canvas/mesh_manipulation.py

merriam:
This code was from a Planet Kivy blog post I ran across.  URL is in the docstring below.

### DIFF
--- a/examples/canvas/mesh_manipulation.py
+++ b/examples/canvas/mesh_manipulation.py
@@ -79,6 +79,9 @@ class MeshBallApp(App):
         return Builder.load_string(kv)
 
     def update_points(self, *args):
+        """ replace self.mesh_points based on current slider positions.
+        Called continuously by a timer because this only sample code.
+        """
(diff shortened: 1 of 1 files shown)


dessant:
Awesome addition! Thanks for all your pulls, they are of great value.

examples/canvas/mesh_manipulation.py
```diff
@@ -0,0 +1,100 @@
...
+
+kv = '''
+BoxLayout:
+    Widget:
+        canvas:
+            Color:
+                rgba: 1, 1, 1, 1
+            Mesh:
+                vertices: app.mesh_points
+                indices: range(len(app.mesh_points) / 4)
```
dessant: This should be `// 4`, otherwise with Python 3 we get:
```python
TypeError: 'float' object cannot be interpreted as an integer
```

dessant:
Patched in https://github.com/kivy/kivy/pull/3008.

PR #2975, opened 2015-02-01
Modify Makefile's 'make style' to check entire tree

merriam:
The makefile targets 'style' and 'stylereport' had only checked
the kivy/ directory.  This allowed style errors to creep into
the examples/ directory.   Now 'style' will check all python
files from the kivy root directory, excluding those
explicitly ignored in kivy/tools/pep8checker/pep8kivy.py.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,10 +93,10 @@ html-embedded:
 	$(MAKE) -C doc html
 
 style:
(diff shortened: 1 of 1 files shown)


dessant:
:+1:

PR #2952, opened 2015-01-30
Add and pass tests for kivy/atlas.py

merriam:
Refactor kivy/atlas.py for testability.  Write tests.  Fix bugs until it passes tests.  Test coverage for atlas.py is now  89%.

PR #2977, opened 2015-02-02
Resubmit changes from merriam_branch_11, pr #2952

merriam:
This is the resubmit; not sure where the merge conflicts were but probably in the leading docstring.

These commits refactored kivy/atlas.py by moving the __main__ code into a function.  After adding some tests, I found there were a bunch of cases where the command line processing failed.  I fixed those.

### DIFF
--- a/kivy/atlas.py
+++ b/kivy/atlas.py
@@ -6,8 +6,8 @@ Atlas
 
 Atlas manages texture atlases: packing multiple textures into
 one. With it, you reduce the number of images loaded and speedup the
-application loading.  This module contains both the Atlas class and command line
-processing for creating an atlas from a set of individual PNG files.   The
+application loading. This module contains both the Atlas class and command line
+processing for creating an atlas from a set of individual PNG files. The
 command line section requires the Pillow library, or the defunct Python Imaging
 Library (PIL), to be installed.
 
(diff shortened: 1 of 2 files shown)


merriam:
This is 10 days old, fixes existing problems with atlas (try "python atlas.py -- --use-path --use-path --use-path" for a crash) and is on the second attempt of pull requests.   Can I help speed this one up?

kivy/atlas.py
```diff
@@ -388,30 +397,38 @@ def create(outname, filenames, size, padding=2, use_path=False):
         return outfn, meta
 
 
-if __name__ == '__main__':
-    """ Main line program. Process command line arguments
-    to make a new atlas. """
+def run(argv):
```
tito: Can you let the run() under the __main__ test ?
One of the experimentation i'm doing is to strip python files of non-reachable code when shipping to production. Everything in __main__ except the main entrypoint is never executed on mobile, and strip down.

Thanks :)

tito:
All the rest looks ok, +1

merriam:
Sure.  Go head and merge it and I'll put in another PR.

It's difficult to juggle the passing PRs.

merriam:
Sorry, just to be clear "Let the run under the main test" means" and "Everything in main except the main entrypoint is never executed on mobile, and strip down." means add a test case explicitly calling main, or breaking out the atlas create functions so that they are not available in this file, but in a different file?

merriam:
Why does this have a documentation tag?  Its functional changes?

tito:
I'll rephrase it: the run() method have no usage if the script is not running as the `__main__` script. Just don't move that part of the code outside the `__main__` test :)

merriam:
Putting code into __main__ makes it untestable, which is probably how all
the atlas errors crept in.   This program is meant to be run as main, at
least from platforms with
a command line.


How about making platform specific code explicit?

In tools:

def not_on_mobile(func):
    ''' decorator to mark function not to be included on mobile platforms
'''
    if is_mobile:
        return func

In atlas:
@not_on_mobile
def run() ......

In this way, no code space would be wasted on a mobile platform that has no
command line, but the code is still testable.   I don't yet know the kivy
test for 'is_mobile'.


Charles

merriam:
If you merge this, I'll put in a PR with the generalized strip() for run. -- Charles

dessant:
Closing this as it conflicts with master. Please resubmit with the changes mentioned by @tito.

PR #2985, opened 2015-02-02
Add to examples/audio sample files 

merriam:
Adding many file formats for a ten second, public domain piano piece.


merriam:
There are more changes missing.   Let me pull it all together and resubmit.

PR #2989, opened 2015-02-03
fix params for TypeError (multiple kw args) in clock.py line 374

Taper21:
Hi,

I made a PR to a fork(kived), but it got merged, so i forked this repo.
what i explained in the PR: https://github.com/kived/kivy/pull/1 
"Hi,
I posted a question on stackoverflow (http://stackoverflow.com/questions/27745932/python-kivy-async-storage) 2 days ago and you answered it.
I tried out my code with your branch and got a error message :

File "/home/sebastian/Dokumente/py-spielwiese/venv/local/lib/python2.7/site-packages/kivy/clock.py", line 374, in tick
ret = callback(self._dt)
TypeError: store_exists_async() got multiple values for keyword argument 'key'

The problem is that the first potional argument should be key and here it is self.dt, so i added dt as the first postional arg to every store_*_async method"

and i made an optional argument callback, like it is explained in the documentation: http://kivy.org/docs/api-kivy.storage.html

### DIFF
--- a/kivy/storage/__init__.py
+++ b/kivy/storage/__init__.py
@@ -67,7 +67,7 @@ asynchronous version.
 
 For example, the *get* method has a `callback` parameter. If set, the
 `callback` will be used to return the result to the user when available:
-the request will be asynchronous. If the `callback` is None, then the
+the request will be asynchronous.  If the `callback` is None, then the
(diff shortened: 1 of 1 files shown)


Taper21:
works in kivy 1.9.1

PR #2993, opened 2015-02-04
Fixed the get_pos method.

rastogiachyut:
Previously the get_pos method used to return pygame.mixer.music.get_pos() which does not work on mixer.sound objects. It has been changed to return (time_now - start_time) which is a float value. The time used is the one in kivy (from kivy.clock import _default_time).

### DIFF
--- a/kivy/core/audio/audio_pygame.py
+++ b/kivy/core/audio/audio_pygame.py
@@ -4,6 +4,7 @@ AudioPygame: implementation of Sound with Pygame
 
 __all__ = ('SoundPygame', )
 
+from kivy.clock import _default_time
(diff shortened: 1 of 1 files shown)


rastogiachyut:
If you feel the patch is inadequate or bad in some way, please do tell me, I'll improve on it.

tshirtman:
Hm, did you try it playing multiple sounds at the same time? i'm not user get_pos() will correctly work in this situation. And you are using a global, that your redefine locally in play(), i don't think that'll work at all.

rastogiachyut:
ok , how do I remove this PR.

PR #2994, opened 2015-02-04
Fixed the get_pos method(tested with two files).

rastogiachyut:
Previously the get_pos method used to return pygame.mixer.music.get_pos() which does not work on mixer.sound objects. It has been changed to return (time_now - self. start_time) which is a float value. The time used is the one in kivy (from kivy.clock import _default_time). This is the code I used to test it.   https://gist.github.com/rastogiachyut/aff9b2eeeba2e8c212c4

### DIFF
--- a/kivy/clock.py
+++ b/kivy/clock.py
@@ -644,7 +644,6 @@ class ClockBase(_ClockBase):
 ClockBase.time.__doc__ = '''Proxy method for time.time() or time.clock(), 
 whichever is more suitable for the running OS'''
     
-
 def mainthread(func):
     '''Decorator that will schedule the call of the function for the next
     available frame in the mainthread. It can be useful when you use
--- a/kivy/clock.py
+++ b/kivy/clock.py
@@ -211,7 +211,7 @@ __all__ = ('Clock', 'ClockBase', 'ClockEvent', 'mainthread')
 
 from sys import platform
 from os import environ
-from functools import wraps,partial
+from functools import wraps, partial
 from kivy.context import register_context
 from kivy.weakmethod import WeakMethod
 from kivy.config import Config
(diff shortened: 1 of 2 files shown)


matham:
Why are you using `_default_time` from `clock`? That is an internal method to clock and not publicly available, event for internal use. You should be using a clock from python's time module instead.

matham:
Also, your code does not seem to take into account the time jump that occurs when seeking.

rastogiachyut:
Ahh.. yes I could use the time.clock() method(on stackoverflow they said that it should be used for all things python as it gives good accuracy), but I also asked on the IRC what gives the best precision , and kived said different methods for different os and its already been done in clock, so I used it. I did not know doing so is bad, so if its a no go, I'll change it and use the time.clock().

matham:
Well, it's only bad because it's an internal method (i.e. prefaced by a underscore). As @kived suggested, if it's added to Clock as a proper method or properly exported in clock.py it is not a problem.

rastogiachyut:
ok , so I put this -> current_time = staticmethod(_default_time)
in clock and then use current_time instead of _default_time.

matham:
yes, that seems right, except maybe call it just time, e.g. `time = staticmethod(_default_time)`. Also, you should add some docs to it indicating what it does. I'm not 100% where these docs should go so that sphinx will read them.

rastogiachyut:
Since get_pos is a pygame implementation it gets to have comments. I've written docs for static method time but if there is a problem then I'll deal with it separately. I am submitting a PR for now, please pull if it meets standards. This is my first time doing this so thank you for your patience.

tito:
Is the issue you're trying to solve is only for android but also on desktop?

rastogiachyut:
Did not add #bug no. in the comments, will do that when I implement kived and matham's suggestions on using self.time.

PR #2997, opened 2015-02-05
Just a few small typos

DizzyDogg:


doc/sources/guide/graphics.rst
```diff
@@ -44,8 +44,8 @@ To add a canvas instruction to a widget, you use the canvas context:
 Context instructions
 --------------------
 
-Context instructions manipulate the opengl context, you can rotate, translate,
-and scale your canvas, attach a texture or change the drawing color, this one
+Context instructions manipulate the opengl context. You can rotate, translate,
+and scale your canvas. You can also attach a texture or change the drawing color. This one
```
akshayaurora: You are going beyond 80 columns, are you sure you have our commit hook installed? Goto kivy install dir and type.

```make hook```

This will install our commit hook that will warn you for all the pep8 violations at commit time.

akshayaurora:
ok, I guess this is small enough that we can fix it ourself, for future reference though you should take a look at our contributing guide http://kivy.org/docs/contribute.html. :+1:

PR #3007, opened 2015-02-08
Fix PageLayout indexing issues.

### LINKED ISSUE #2928: Behaviour of PageLayout might not be the expected.
I noticed that the PageLayout has an unexpected behaviour. The "Pages" in the middle get covered by the next page (the one to the page). To fix it, the programmer would have to make sure that there is an empty space (equal to the border property of PageLayout) on the middle pages, but not on the extremes. This makes me think that the current behaviour is wrong, and should be fixed in order to keep things centered. 

Here is an example to replicate

    <MyPageLayout>:
        Button:
            font_size: 26
            text: 'This is a very long description for the page 1 and centered'
        Button:
            font_size: 26
            text: 'This is a very long description for the page 2 not centered'
        Button:
            font_size: 26
            text: 'This is a very long description for the page 3 and centered'

And 3 screenshots. Notice how the second one is convered by the 3rd one.

page1
page2
page3

Thank you!

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8083665-behaviour-of-pagelayout-might-not-be-the-expected?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

matham:
This fixes #2928 by setting the correct width for first AND last page. It also fixes some page scrolling issues which is obvious if you run the following code and page right to left or left to right:

```py
from kivy.lang import Builder
from kivy.modules import inspector

PageLayout:
    Button:
        font_size: 26
        text: 'This is a very long description for the page 1 and centered'
        text_size: self.width, None
    Button:
        font_size: 26
        text: '1This is a very long description for the page 2 not centered'
        text_size: self.width, None
    Button:
        font_size: 26
        text: '2This is a very long description for the page 2 not centered'
        text_size: self.width, None
    Button:
        font_size: 26
        text: 'This is a very long description for the page 3 and centered'
        text_size: self.width, None
''')
inspector.create_inspector(Window, root)
runTouchApp(root)
```

@tshirtman can you verify that it's ok.

### DIFF
--- a/kivy/uix/pagelayout.py
+++ b/kivy/uix/pagelayout.py
@@ -75,34 +75,36 @@ class PageLayout(Layout):
             pos=self._trigger_layout)
 
     def do_layout(self, *largs):
-        l_children = len(self.children)
+        l_children = len(self.children) - 1
+        h = self.height
+        x_parent, y_parent = self.pos
+        p = self.page
+        border = self.border
+        right = self.right
         for i, c in enumerate(reversed(self.children)):
-            if i < l_children:
-                width = self.width - self.border
+            if not i or i == l_children:
+                width = self.width - border
             else:
-                width = self.width - 2 * self.border
+                width = self.width - 2 * border
 
-            if i == 0:
-                x = self.x
-
-            elif i < self.page:
-                x = self.x
-
-            elif i == self.page:
-                x = self.x + self.border
-
-            elif i == self.page + 1:
-                x = self.right - self.border
-
+            if i < p:
+                x = x_parent
+            elif i == p:
+                if not p:
+                    x = x_parent
+                else:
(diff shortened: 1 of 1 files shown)


kived:
There is still an issue that the first and last pages are wider than the middle pages. I think the pages should all be the same size, but use different border sizes on the first and last page.

So with the default 50dp:
    0dp first page 50dp
    25dp middle pages 25dp
    50dp last page 0dp

kived:
:+1:

PR #3013, opened 2015-02-08
Update clock.py

rastogiachyut:
changed all(4) of the _default_time uses in ClockBase to self.time , to make it faster

### DIFF
--- a/kivy/clock.py
+++ b/kivy/clock.py
@@ -49,7 +49,7 @@ to write a short function that does accept dt. For Example::
 .. note::
 
     You cannot unschedule an anonymous function unless you keep a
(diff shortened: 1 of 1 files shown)


rastogiachyut:
actually I had my reasons mixed up its not faster , actually slightly slower.

rastogiachyut:
this way we have a method we can override if needed , thats why it was recommended I think.

matham:
Weird, but it's not mergeable. If you fix it I'll merge it.

rastogiachyut:
ok, this looks like my clone of kivy time traveled a few commits back?
I pushed and pulled before submitting this.

PR #3014, opened 2015-02-08
core/text: allow others font extension to be loaded

tito:
Freetype itself supports much more than ttf: otf, woff, etc.
See the full list at http://www.freetype.org/freetype2/

It can prevent sdl2 issue like this when you load otf: 

```
 Exception AssertionError in 'kivy.core.text._text_sdl2._get_font' ignored
There are 0 joysticks available
```


akshayaurora:
:dancers:

PR #3015, opened 2015-02-09
Fix for issue plus some extras

thopiekar:
https://github.com/kivy/kivy/issues/3011


dessant:
Sorry for the late reply, i've just tested it and everything seems to be ok.
Great job. :+1:

thopiekar:
No problem ;)

PR #3021, opened 2015-02-11
change slider default padding to sp(16) to match sp(32) size of slider c...

denys-duchier:
...ursor.

this fixes problems with the colorpicker where cursors can seem stuck and
non-responsive when at extremal extent.


tito:
ty!

PR #3023, opened 2015-02-12
Enhance examples/audio to use multiple file types.

merriam:
This is a resubmit of PR#2987.   And PR#2985.  And PR#2938.

It updates examples/audio to not crash when failing to play a file, attempt to play additional formats, and have some samples of additional music to play.


tito:
I'm scared to show that. Most of the extensions wont work on mobile, as only .mp3 / .wav are save to use on android/ios.

merriam:
Sorry, I don't understand.

Most extensions currently aren't supported on any platform, with different
failures on each case.  For example, .ogg currently just 'clicks' my
speakers on os/x.   The idea is to have a way of seeing what happens.
It's not currently meant as an advertisement.

-- Charles

tito:
That's why there is a difference between an audio example that works and a audio example that partially works. The partially works = users will open issue because our example doesn't work, as we did already have for non-fully working example.

Your multiples tests files looks more something that we could have for a "testing" kivy app, to see if all the cases are working, or detect the one that aren't.

merriam:
How about some code to answer what extensions work on the current platform?

Adding to kivy.core.audio/__init__.py SoundLoader:

     @classmethod
     def extensions:
           ''' return list of extensions supported on current configuration"
           return list(set(itertools.chain([c.extensions() for c in
SoundLoader._classes])))

Then in examples/audio/main.py:
     extensions = SoundLoader.extensions()

This gives a nice boost to the audio api, though consumers of other media
will want something similar.  For the audio demo, only supported file types
for the current platform will be displayed.

Should I just put in a PR for these changes?

Charles

dessant:
I don't see the point of adding more audio files as far as the example is concerned. From the perspective of a user, it is just changing the filename in the audio path.

The current example already serves its purpose and shows how to play audio in a kivy app. The pr's benefits do not justify bloating the size of our examples folder, which is pulled on every pip install or otherwise.

I think it would be much more useful to have something similar as an unit test, in that case the benefits would be evident.

merriam:
@dessant  I'll trim the audio files to the ones I know we support somewhere.  Can you please look to be sure you have no additional objections?   I feel like I'm putting in a PR, get a nit, put in a PR, get another one.   In the mean time, the unchanged audio.py just crashes when it can't play a sound on a platform.

Outside of audio files, like FLAC, not playing on any platform, do you have any objections?

@tito  Should I just add 'extensions' to the core.audio API?  Should I be asking before making small changes to the API?

dessant:
@merriam as i said before, i don't see the point of this pr. The user wants to know how to play audio, which is already shown, the additional files are not relavant for the example and they are just bloat, even if you trim them down.

merriam:
OK, I removed the extra formats.  Meh.

This PR now handles:
- docstring and such so that audio can be in examples.
- removing crashes in audio
- including one longer wav file so that one can see that multiple sounds can play at once.

Please reopen and apply it.  No, I will not submit a new PR.

PR #3025, opened 2015-02-12
Makefile - add 'test-quiet' target

merriam:
Add a target to run nosetests without kivy log messages, test names, or the coverage report.  It just
shows dots and errors.


merriam:
Placed in a new PR to prevent further version skew.

PR #3026, opened 2015-02-12
kivy/logger.py - Document environment variables

merriam:


dessant:
I think linking [here](http://kivy.org/docs/guide/environment.html?highlight=environment#configuration) would be better than duplicating parts of the doc.

merriam:
Hadn't found that link when searching for it.   Now links to it and updated the docs in the configuration section to be more clear and consistent.

doc/sources/guide/config.rst
```diff
@@ -1,10 +1,22 @@
...
+There are three ways to configure Kivy and they can be used in combination.
+
+* Setting :ref:`environment variables` starting with **KIVY_**.  Environment variables can also be set by
+  a program before importing the kivy module.
+* Passing options on the command line.  Importing the kivy module parses command line
+  options, unless KIVY_NO_ARGS is set.  If the option "*--*" is found, parsing stops
+  and remaining options can be parsed the program.  Use *--help* for a list of command
+  line options.
+* Using the kivy configuration file, named `config.ini`.  This adheres
+   to the `standard INI <http://en.wikipedia.org/wiki/INI_file>`_ format.
```
kived: This line is over-indented by one space. Line continuations in a list should be indented to the same level as the beginning line (2 spaces in this case, not 3). I don't know how the doc builder will handle this, but at least GitHub's parsing (rich diff) shows that this could cause issues.

merriam:
Yes.  Thank you for catching that.  Fixed.

dessant:
Besides what @kived mentioned, there are other formatting issues (spaces, line length), please make a commit for those too, so we can merge the pr.

merriam:
These are going in with a new PR.

PR #3028, opened 2015-02-13
Rename (change case) of 3Drendering screenshot.

merriam:
Rename 3DRendering__main__py.png to 3Drendering__main__py.png.  The case failing to
match the examples/3drendering page is undetectable on OS/X (it preserves but ignores
case) but fails on website production.


merriam:
That it, the kivy.org website gallery doesn't have 3DRendering Rotating Monkey Head because the screenshot case was wrong.  This fixes the problem.

PR #3029, opened 2015-02-14
gallery - Avoid rewriting unchanged files.

merriam:
Rewriting unchanged files causes the ‘make html’ and sphinx to waste
time rebuilding every example page.   Added code to skip writing
unchanged files and associated UnitTest test case.

Diff got confused here, as I converted the existing test case from nose to UnitTest.
Now at MaxPR.


merriam:
Put in with new pull request as gallery.py changed again.

PR #3030, opened 2015-02-14
Method to retrieve the matrix

olymk2:
Add a new method which will return the value of the matrix which we can reference in python code

### DIFF
--- a/kivy/graphics/transformation.pyx
+++ b/kivy/graphics/transformation.pyx
@@ -59,6 +59,17 @@ cdef class Matrix:
     def __init__(self):
         self.identity()
 
+    def get(self):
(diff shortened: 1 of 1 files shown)


tito:
I wonder why we didn't have it in the first case :)
Can you add the documentation for it, with a versionadded tag to 1.9 ?

Thanks!

olymk2:
I am happy to but where, looking in kivy/doc/sources i dont see an obvious place to add any docs or existing docs for matrix ? is this a different repo ?

olymk2:
figured out you just wanted a doc string, added :)

dessant:
:+1:

PR #3031, opened 2015-02-14
document automatic dependencies some more.

denys-duchier:
Here is a little bit extra documentation on automatic dependencies, for the problem of self.a.b[self.c.d].e.f which was not at first obvious to me.  Perhaps others may find it helpful too.

### DIFF
--- a/kivy/lang.py
+++ b/kivy/lang.py
@@ -287,6 +287,30 @@ The Button text changes with the state of the button. By default, the button
 text will be 'Plop world', but when the button is being pressed, the text will
 change to 'Release me!'.
 
+More precisely, the kivy language parser detects all substrings of the form
+`X.a.b` where `X` is `self` or `root` or `app` or a known id, and `a` and `b`
+are properties: it then adds the appropriate dependencies to cause the
+the constraint to be reevaluated whenever something changes.  For example,
+this works exactly as expected::
+
+    <IndexedExample>:
+        beta: self.a.b[self.c.d]
+
+However, due to limitations in the parser which hopefully may be lifted in the
+future, the following doesn't work::
+
+    <BadExample>:
+        beta: self.a.b[self.c.d].e.f
+
+indeed the `.e.f` part is not recognized because it doesn't follow the expected
+pattern, and so, does not result in an appropriate dependency being setup.
+Instead, an intermediate property should be introduced to allow the following
+constraint::
+
+    <GoodExample>:
+        alpha: self.a.b[self.c.d]
+        beta:  self.alpha.e.f
+
 
 Graphical Instructions
 ----------------------


matham:
I'm slightly hesitating to merge this because it's stating something that I consider a deficiency in the kv parser as a if it was by design. 

The difference is that with the hope that in the future the kv language will be able to parse such things, we should not state it like you stated it, but rather by saying e.g. `However, due to limitations in the parser, the following doesn't work::`, implying one should not rely on this, because in the future it may be fixed to work.

denys-duchier:
it is NOT a defiiciency in the kv parser, but rather one of design.  Dependencies are determined statically rather than dynamically.  I don't wan't to start a pointless altercation about design; I'd rather document the current limitations.  If they are subequently removed, I'd be delighted to update the documentation accordingly.

matham:
Huh? The reason why I wanted you to mention that is because if we write it as is, it can be taken as a promise that this is how it should work. The result being that if we change it in the future to bind properly when someone writes `beta: self.a.b[self.c.d].e.f` we'd be breaking backward compat. But if we just say that is how it's working and indicting it may change in the future, such a change would not be breaking backward compat, or less so because it wouldn't be a promise that this is how kv works.

I'm not sure what you mean by `Dependencies are determined statically rather than dynamically.` but the reason that `beta: self.a.b[self.c.d].e.f` doesn't work is not because of any static eval of the props during binding at widget creation, but because the parser doesn't accept that as a correct pattern. If the parser accepted it as a correct pattern and passed it on, the builder could statically evaluate the value of `self.a.b[self.c.d].e.f`` during binding as well. Unless I misunderstood what you meant by static dependencies. Also, not that it is necessarily related to this, but some props accept a `rebind` parameter which would make it dynamic.

denys-duchier:
by statically, I mean: by analysis of the code ahead of time.  by dynamically, I mean: by recording at runtime what properties have actually been accessed.  currently, if certain dependencies are hidden from static analysis, whether through a limitation of pattern matching or because they are incurred through function calls, then they can never cause a reevaluation of the property.

Once again, I am not advocating a different way of doing things, but lack of documentation was a problem for me; and, it is only by puzzling over the code in kivy/lang.py that I finally understood why some of my properties were not working as I intuitively had expected.

Whether the current limitations are documented, or must be rediscovered by each user, cannot prevent said user from making the unwarranted assumption that these limitations will persist indefinitely.  But documentation will save time and puzzlement, and can always be removed later if the aforementioned limitations are eventually overcome (and in that case, the advice I proposed will still continue to work).

matham:
I agree, this PR is needed. I think you might have miss-understood me, I wasn't against the PR I just wanted you to change the wording a bit to make it clear that this behavior might change (hopefully) in the future. Anyway, I'll merge later whether or not you make the wording change, please do if you can :).

denys-duchier:
ah ok! done!

PR #3037, opened 2015-02-16
Make EventDispatcher forward kwargs when it doesn't use it.

### LINKED ISSUE #3036: EventDispatcher eats up all kwargs even if it did not process it.
If any any of the kwargs are property names, EventDispatcher sets those property to the kwargs values. However, if the key is not a property, it does not forward it with super, but just eats it up.

For example:

```
class MyColor(object):

    def __init__(self, color='blue', **kwargs):
        print 'color is {}'.format(color)

class Coloring(EventDispatcher, MyColor):
    pass

color = Coloring(color='green')
```
prints `color is blue`. rather than green.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8607842-eventdispatcher-eats-up-all-kwargs-even-if-it-did-not-process-it?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

matham:
Fixes #3036.

### DIFF
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -244,12 +244,14 @@ cdef class EventDispatcher(ObjectWithUid):
     def __init__(self, **kwargs):
         cdef basestring func, name, key
         cdef dict properties
-        # object.__init__ takes no parameters as of 2.6; passing kwargs
-        # triggers a DeprecationWarning or worse
-        super(EventDispatcher, self).__init__()
+        cdef list prop_args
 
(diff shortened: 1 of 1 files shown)


kived:
:+1:

PR #3039, opened 2015-02-16
Rename focused to focus and make focused an alias of focus.

matham:
Rename focused to focus and make focused an alias of focus.


akshayaurora:
:100:

PR #3040, opened 2015-02-16
Add a put method to Matrix

olymk2:
change get to return a numpy friendly format so we can do numpy.array(Matix().get())
add a put method to insert data back into the matrix also in a numpy format add in method to get and set individual values by index

### DIFF
--- a/kivy/graphics/transformation.pyx
+++ b/kivy/graphics/transformation.pyx
@@ -59,16 +59,59 @@ cdef class Matrix:
     def __init__(self):
         self.identity()
 
-    def get(self):
-        '''Retrieve the value of the current matrix.
+    def get(Matrix self):
+        '''Retrieve the value of the current matrix in numpy format.
+        for example m.get() will return 
+
+                [[1.000000, 0.000000, 0.000000, 0.000000],
+                [0.000000, 1.000000, 0.000000, 0.000000],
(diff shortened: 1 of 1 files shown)


Kovak:
thanks for adding the typing on the selfs, you can type the one in __init__ as well.

Can you also pull self.mat into the local namespace for get and set so that we aren't performing the . lookup 16 times.

olymk2:
you will have to explain how to pull in to the local namespace.

mat = self.mat does not work results in Cannot convert 'matrix_t' to Python object
matrix_t mat = self.mat also does not work with Syntax error in simple statement list

this is where i get lost because i am not familiar with cython.

kived:
@Kovak: FYI, typing isn't needed on the `self` parameter. Cython already knows what type it should be, and I can confirm that this works at least on 0.21.2. The generated C code will be identical whether `self` is typed or not. In fact, you can put `str self` and Cython will just ignore you. :)

Also, pulling a cdef attribute into the local namespace is wasteful. Accessing a cdef attribute is just an offset from a pointer and is far quicker than Python attribute access.

olymk2:
Just relised this would not merge cleanly so i have rebased so it will now :)

based on the comments above i dont think i need to change anything else at the moment ?

Kovak:
Sorry about that @olymk2 ignore my comments, I was basing the typing off of outdated Cython documentation it seems.

PR #3048, opened 2015-02-18
Add Graphics RoundedRectangle

hey-sancho:
Added a new vertex instruction - RoundedRectangle.
This allows to create rectangles with rounded corners on the canvas, just like Ellipse, Rectangle and such.
Dimensions for each corner can be specified by tuples for elliptical corners, or by numbers for circular corners.
Segments parameter defines number of segments for smoothness of all corners.
Imports and factory register allow to use RoundedRectangle both from code and .kv file.
Basic documentation and commentaries are provided.

Preview:
rounded_rect


inclement:
This looks really nice. Would you be able to add a simple example to the canvas examples folder? If you have it still, the app you used for the screenshot would be fine.

Thanks!

hey-sancho:
I've added a similar example, both with kvlang and python syntax.

hey-sancho:
Added possibility to set different number of segments for different corners by using list of integers (along with the single integer).
This can be useful in situations, where large figure has one small round corner and one big. To make big corner very round there should be lots of segments, while such amount is not needed for the small one.
Also it allows to cut some of the corners.

kivy/graphics/vertex_instructions.pyx
```diff
@@ -1091,3 +1091,312 @@ cdef class Ellipse(Rectangle):
...
+            float tangetial_factor, radial_factor, theta
+            list points
+
+        # convert to radians
+        angle_start *= 0.017453292519943295
+        angle_end *= 0.017453292519943295
+
+        # number of vertices for arc, including start & end
+        theta = (angle_end - angle_start) / segments
+        tangetial_factor = tan(theta)
```
inclement: Small typo -> should be tangential

inclement:
This looks good to me, I think it can be merged, though I've noted a small typo that would be nice to fix.

hey-sancho:
@inclement thanks for noticing, i've added a fix.

There's a small issue with current code though: 
Very small figures with `segments=1` and biggest radius (a rhombus-like figure) look ugly. Opposing corners on vertical and/or horizontal axes seem misaligned.
I have a fix for that, but it involves reducing maximum radius by one, so for a figure of size 100x100, maximum radius would be 49, instead of 50. Therefore, user can draw perfect tiny rhombuses, but slightly imperfect circular figures. However, with adequate radius and number of segments this is insignificant.

I wonder if all these features even must be a part of "rounded" rectangle, perhaps it should just "do one thing well", and leave rhombuses to `Mesh`?

akshayaurora:
@hey-sancho nice work :+1:

PR #3060, opened 2015-02-21
Fixed filechooser path incorrectly updated when going to parent directory

ODiogoSilva:
In the current version of FileChooser, the path attribute does not seem to properly update when selecting the parent ("../") directory. I often use a Label on top of the FileChooser widget showing the current path, and everytime I clicked on the "../" item to go to the parent directory, this would append the string "../" to the path. For example, if I wanted to go to the parent directory from my home ("/home/diogo"), I would get this on my label: "/home/diogo/../". This would also create errors when opening files with the "../" strings on the path. This issue can be replicated using a slightly modified version of the basic filechooser widget in kivy/uix/filechooser.py - https://gist.github.com/ODiogoSilva/ef7d78551c89e6c636d6

This commit seems to have fixed things for me, though I only tested it on Linux.

### DIFF
--- a/kivy/uix/filechooser.py
+++ b/kivy/uix/filechooser.py
@@ -263,7 +263,6 @@ class FileChooserController(FloatLayout):
             root is refreshed.
         `on_subentry_to_entry`: entry, parent
             Fired when a sub-entry is added to an existing entry.
-        `on_remove_subentry`: entry, parent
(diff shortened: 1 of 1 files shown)


akshayaurora:
@ODiogoSilva nice catch.

Issue #2871, opened 2015-01-20
List dependency versions in the install docs and future release notes


dessant:
The install docs should recommend safe dep versions, and they should be updated after every kivy release. This would save users from cases like this one https://github.com/kivy/kivy/issues/2794.

kived:
Fixed with #3061. Not sure why this issue didn't close automatically.

PR #3061, opened 2015-02-21
improve cython support messages and add unsupported versions

kived:


kived:
Modified installation documentation so that it always reflects the currently supported Cython versions - Fixes #2871.

Also adds a `fasthtml` target to the doc Makefile to easily build docs using multiple (four) cores, and automatically add `..` to the PYTHONPATH when building html.

Issue #2805, opened 2014-12-30
Cython 0.22 does not compile master


matham:
Using cython dev from github:

```shell
e:\MinGW\bin\gcc.exe -shared -s build\temp.win32-2.7\Release\kivy\graphics\trans
formation.o build\temp.win32-2.7\Release\kivy\graphics\transformation.def -Le:\P
ython27\libs -Le:\Python27\PCbuild -lopengl32 -lglew32 -lpython27 -lmsvcr90 -o G
:\Python\dev2\kivy\kivy\graphics\transformation.pyd -lm
cythoning kivy\graphics/vbo.pyx to kivy\graphics\vbo.c

 Error compiling Cython file:
 ------------------------------------------------------------
 ...
     cpdef release(self)
     cpdef get_pixel_color(self, int wx, int wy)

     cdef void create_fbo(self)
     cdef void delete_fbo(self)
     cdef void apply(self)
                   ^
 ------------------------------------------------------------

 kivy\graphics\fbo.pxd:27:19: Signature not compatible with previous declaration


 Error compiling Cython file:
 ------------------------------------------------------------
 ...
     cdef void push_states(self, list names) except *
     cdef void pop_state(self, str name) except *
     cdef void pop_states(self, list names) except *
     cdef void enter(self) except *
     cdef void leave(self) except *
     cdef void apply(self) except *
                   ^
 ------------------------------------------------------------

 kivy\graphics\instructions.pxd:123:19: Previous declaration is here
building 'kivy.graphics.vbo' extension
e:\MinGW\bin\gcc.exe -mdll -O -Wall -Ie:\Python27\include -Ie:\Python27\PC -c ki
vy\graphics\vbo.c -o build\temp.win32-2.7\Release\kivy\graphics\vbo.o
kivy\graphics\vbo.c:1:2: error: #error Do not use this file, it is the result of
 a failed Cython compilation.
 #error Do not use this file, it is the result of a failed Cython compilation.
  ^
 error: command 'e:\\MinGW\\bin\\gcc.exe' failed with exit status 1
make: *** [force] Error 1
```

I think they most have become stricter about keeping function prototypes consistent.

matham:
Reopening because this is now a issue with cython 0.22 on pip.

matham:
Here's the cython bug report of this issue: https://groups.google.com/forum/#!topic/cython-users/r3dQSfOKMjI

kived:
[...]

PR #3062, opened 2015-02-21
fix for cython 0.22

kived:


kived:
Fixes #2805

matham:
@kived I suspect that this will fix the issue, however I did not think this would be acceptable. The reason is that you're promoting __all__ apply methods that inherit from instruction to have an `except *`. However, Only RenderContext and below had one. When you fix it just for RenderContext and below, that's when I get the error.

Not only will it, I imagine, add slowness to all these methods when they might not need to watch out for any exception. But in fact, in the past when we might have gotten a printout of e.g. `graphics exception ignored....` when e.g. things go wrong by the graphics engine, now it'd actually crash instead. And I'm not sure that's a good change.

So I didn't think it was a good idea to change our pattern to accommodate this cython bug, unless there was a actual good reason for adding all these `except *`. IMHO it'd be better to light a candle under the cythod devs asses or fix it ourselves while refusing to build with 0.22 for now, then to do such a change and face the bug down the line...

kived:
Quite true. I thought about the option of changing them all to `cdef int xxx(xxx) except -1`, which wouldn't have the same performance implications (since it will only generate the call to `PyErr_Occurred` on a retval of -1), though it means the methods all need to return something (0). I don't know that it's really a good idea to allow exceptions to be ignored unless we're sure the exception should be ignored.

A few of the changes (shader.pyx) will probably be required either way - the pxd already declared them as `except *` and it looks like 0.22 requires the definition to use the exact same signature as the declaration. Kind of pedantic, but not really a bug and it's less likely that part will change.

If you want to take a look at #3061, that includes code to refuse to build on 0.22.

Issue #2940, opened 2015-01-29
Textinput handles are still touchable even when unfocused


tito:
1. Select the first textinput. Check visually where the handles are poping.
2. Unfocus the first textinput
3. Then try to focus the second textinput. It will not work where the handles of the first textinput where.

![screenshot_2015-01-29-16-14-56](https://cloud.githubusercontent.com/assets/37904/5960980/a269850c-a7d9-11e4-8207-dc9f2ef40628.png)

kived:
Fixed in #3063

PR #3063, opened 2015-02-22
move TextInput handles/bubble to window with transformation

### LINKED ISSUE #2939: TextInput Handles are covered behind other siblings in 1.9
Due to the changes in https://github.com/kivy/kivy/pull/2390. We now have a issue where the default behaviour of TextInput Handles is not working as expected. 

http://txzone.net/files/tmp/Screenshot_2015-01-29-16-14-56.png

This needs to be fixed to still use the old behaviour and enable this option using a parameter/property for when the user needs to put The TextInput inside a ScrollView or a Satter. . 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8130615-textinput-handles-are-covered-behind-other-siblings-in-1-9?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kived:
Fixes #2939. Moves rendering of the handles and bubble back to the Window, but provides the appropriate transformations. I set up RelativeLayout, Scatter and ScrollView to apply transformations; if other widgets cause similar transformations they will need to override _apply_transform().

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -26,6 +26,7 @@ from kivy.utils import platform, reify
 from kivy.context import get_current_context
 from kivy.uix.behaviors import FocusBehavior
 from kivy.setupconfig import USE_SDL2
+from kivy.graphics.transformation import Matrix
 
 # late import
 VKeyboard = None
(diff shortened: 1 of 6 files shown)


kived:
Added a commit which fixes #2940 as well.

akshayaurora:
@kived :+1:

Issue #3046, opened 2015-02-17
Uninformative exception raised if canvas follows children widgets in kv


matham:
Running the following code:

```
from kivy.lang import Builder
from kivy.app import runTouchApp

runTouchApp(Builder.load_string('''
Widget:
    Label:
        text: 'Hello'
    canvas:
        Color:
            rgb: 1, 1, 1
'''))
```

Will result in:
```
 Traceback (most recent call last):
   File "G:\Python\libs\Playground\src\playground5.py", line 11, in <module>
     '''))
   File "g:\python\dev2\kivy\kivy\lang.py", line 1804, in load_string
     self._apply_rule(widget, parser.root, parser.root)
   File "g:\python\dev2\kivy\kivy\lang.py", line 1915, in _apply_rule
     cls = Factory_get(cname)
   File "g:\python\dev2\kivy\kivy\factory.py", line 130, in __getattr__
     raise AttributeError
 AttributeError
```

Because you cannot have the canvas after any children widgets. However, the exception is completely uninformative.

PR #3068, opened 2015-02-22
Improved error when canvas instructions are added after child widgets in kv

inclement:
Resolves https://github.com/kivy/kivy/issues/3046

Maybe the Exception type should be more specific, not sure what's best.

### DIFF
--- a/kivy/lang.py
+++ b/kivy/lang.py
@@ -1910,11 +1910,12 @@ class BuilderBase(object):
         for crule in rule.children:
             cname = crule.name
 
+            if cname in ('canvas', 'canvas.before', 'canvas.after'):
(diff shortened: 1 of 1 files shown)


kivy/lang.py
```diff
@@ -1910,6 +1910,10 @@ def _apply_rule(self, widget, rule, rootrule, template_ctx=None):
         for crule in rule.children:
             cname = crule.name
 
+            if cname in ('canvas', 'canvas.before', 'canvas.after'):
+                raise Exception('canvas instructions added in kv must '
```
kived: This should use `ParserException`; i.e. `ParserException(crule.ctx, crule.line, 'canvas instructions....')`. If you change that then :+1: for merging!

PR #3070, opened 2015-02-22
Scale origin property

rastogiachyut:
added property origin and made changes to set_scale method in class scale.


inclement:
I'm unsure what's going on here, I just merged #3134 that does the same thing, but I'd have expected this to raise merge conflicts. Anyway, I'm closing this as no longer necessary, but thanks again for the other PR.

PR #3072, opened 2015-02-23
Fixed two bugs with incompatibilities with python3 in hidinput.py.

Nixon-:
First was on line 593. Second was 673.

First:  Return result of fcntl.ioctl needs to convert to string first.

Second: Simple, range (float) doesn't work. In py2 this would have worked since int/int is int, while py3 int/int is float

Found when trying to run Kivy App on Raspberry Pi with compiled Cython 0.20.1 and compiled Python 3.4


akshayaurora:
Thanks nice find! :+1:

PR #3074, opened 2015-02-23
more general css selectors for kv rules

denys-duchier:
I would like some feedback on a prototype for a pure extension to kv-lang introducing css-like contextual selectors for kv rules. 

The new contextual rule selectors are enclosed in `<< ... >>` which makes them entirely separate from existing selectors.  Here is an example selector:

```
 <<A B.p1.p2 > C *#i3>>
```

Such a rule applies to any (`*`) widget with `id=i3`, that is a descendant of a widget whose Python class inherits from `C`, which itself is an immediate child of a widget whose python class inherits from `B` and whose `cls` property has elements `p1`and `p2`, and which itself is a descendant of a widget whose Python
class inherits from `A`.

Since widgets get their properties constrained fairly late in the build process, contextual selection uses statically evaluable properties `id` and `cls` of rules whenever possible and as early as possible.

An advantage of this extension is that you can more easily modularize your design: you can provide a common pattern as a rule that is instantiated many time, and contextual rules to specialize that pattern for each instantiation.

Here is a (not very good) example:

```
ScreenManager:
    Screen:
        id: one
    Screen:
        id: two

<Screen>:
    id: screen
    BoxLayout:
        orientation: "vertical"
        id: screen-layout
        Label:
            id: banner
            size_hint: 1,None
            height: dp(50)
        Button:
            id: button1
            screen: screen

<<Screen#one>>:
    name: "un"

<<Screen#two>>:
    name: "deux"

<<Screen#one Label#banner>>:
    text: "Screen 1"

<<Screen#two Label#banner>>:
    text: "Screen 2"

<<Screen#one Button#button1>>:
    text: "To Screen 2"
    on_release:
        root.screen.manager.transition.direction = "left"
        root.screen.manager.current = "deux"

<<Screen#two Button#button1>>:
    text: "To Screen 1"
    on_release:
        root.screen.manager.transition.direction = "right"
        root.screen.manager.current = "un"
```

### DIFF
--- a/kivy/css_selectors.py
+++ b/kivy/css_selectors.py
@@ -192,8 +192,7 @@ class IdSelector(FilterSelectorBase):
     op = "#"
 
     def match(self, elt):
-        name = self.name
-        return name == elt.id
+        return self.name == elt.id
 
 
 class PrecedenceSelector(SelectorBase):
--- a/kivy/css_selectors.py
+++ b/kivy/css_selectors.py
@@ -22,6 +22,23 @@ late (too late for rule selection), class selectors and id selectors also
 consult the information stored in the rule responsible for creatng the widget.
 If that information can be evaluated statically, then it can influence rule
 selection (presence of an `id` or `cls` property).
+
+After rule selection has occurred, the selected rules will also be used to
+get more precise information about `id` and `cls` before they have taken effect
+on the widget.  In this fashion, the context for descendent widgets becomes
+more precise.
+
+If you don't use contextual rules, nothing changes.  However, if you do, then
+the rule match cache is deactivated; it probably won't have a measurable
+incidence on your application, but "caveat emptor!".
+
+A widget that is created from Python code, rather than in from a kv-lang
+specification, does not have a context in which to evaluate contextual rules
+that might or might not apply to it.  For this reason, the :class:`Widget`
+constructor also accepts a `css_ctx` keyword parameter whose value maybe
+either the intended :class:`Widget` parent or an explicit :class:`CSSCtx`
+instance describing the hierarchy into which the new widget will be
+inserted.
 '''
 import re
 
--- a/kivy/css_selectors.py
+++ b/kivy/css_selectors.py
@@ -14,6 +14,78 @@ of a widget whose python class inherits from `B` and whose `cls` property has
 elements `p1`and `p2`, and which itself is a descendant of a widget whose Python
 class inherits from `A`.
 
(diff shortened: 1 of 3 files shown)


tshirtman:
Wow, needs to test this, but i like the idea! :)

denys-duchier:
It is now possible to nest contextual rules.  A nested rule inherits as a prefix the selectors of those it is nested in.  Thus,  the contextual rules of the previous example can now also be written as follows:

```
<<Screen#one>>:
    name: "un"
    <<Label#banner>>:
        text: "Screen 1"
    <<Button#button1>>:
        text: "To Screen 2"
        on_release:
            root.screen.manager.transition.direction = "left"
            root.screen.manager.current = "deux"

<<Screen#two>>:
    name: "deux"
    <<Label#banner>>:
        text: "Screen 2"
    <<Button#button1>>:
        text: "To Screen 1"
        on_release:
            root.screen.manager.transition.direction = "right"
            root.screen.manager.current = "un"
```

dessant:
@tito, are you ok with merging?

denys-duchier:
hem guys, I was seeking **feedback** on this idea. I didn't think of it as ready for merging.  if you like the idea, then I'll go the extra mile to make it ready for merge. note that, when contextual rules are present in the app, caching is automatically turned off.  I have not measured how much of an impact this may have on performance.

tito:
Don't merge @dessant ^^. Theses implying new notion that need thinking first.

Initially, that's something i wanted to target as well, this is why the language have internally selectors for "cls", "id" in addition to the most used class-matching. 
But there is a lot of issues when i tried to go on this road that i simply stopped to work on it, like:
- saving all the callback to unbind them at some point
- ability to reparse/match the whole tree to unapply/reapply new rule
- children!! Some widget don't work with one, or multiple. Any changes to widgets can just be very bad as none of us design a widget to allow changes of the kv rules.

I guess this would be nice to have, if it's currently limited to properties changes.

The syntax could be plug on the current one, there is no need to create a second <<>>.

From now i'll try to think about it and review your initial code. Thanks a LOT of this :)

denys-duchier:
of course, the only reason for <<>> was to **not** interfere with the existing framework, precisely because this was a speculative extension.  I would be glad to get your input and make whatever modification would be beneficial.

tito:
Good initial idea, but the latest feedback wasn't enough to continue the discussion. And the PR itself cannot be merged. Anybody that have an interest can try to revive it!

PR #3080, opened 2015-02-25
screenmanager: screen's layout is fixed before on_enter is dispatched.

fumiya-kubota:
When `on_enter` is called, I hope size of the screen has been already fixed.

```python
from kivy.app import App
from kivy.uix.screenmanager import ScreenManager
from kivy.uix.screenmanager import Screen
from kivy.clock import Clock

class TestScreen(Screen):
    def __init__(self, **kw):
        super(TestScreen, self).__init__(**kw)

    def on_enter(self, *args):
        print 'on_enter screen size >', self.size
        Clock.schedule_interval(self.update, 1.0 / 60.0)

    def update(self, dt):
        print 'update on loop screen size >', self.size


class TestApp(App):
    def build(self):
        root = ScreenManager()
        return root

    def on_start(self):
        self.root.add_widget(TestScreen(name='scene'))

if __name__ == '__main__':
    TestApp().run()
```

```
[INFO   ] [Logger      ] Record log in /Users/sasau/.kivy/logs/kivy_15-02-25_7.txt
[INFO   ] [Kivy        ] v1.9.0-dev
[INFO   ] [Python      ] v2.7.6 (default, Sep  9 2014, 15:04:36) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)]
[INFO   ] [Factory     ] 173 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_imageio, img_dds, img_gif, img_pygame (img_pil, img_ffpyplayer ignored)
[INFO   ] [Window      ] Provider: pygame
[INFO   ] [GL          ] OpenGL version <2.1 INTEL-10.2.46>
[INFO   ] [GL          ] OpenGL vendor <Intel Inc.>
[INFO   ] [GL          ] OpenGL renderer <Intel Iris Pro OpenGL Engine>
[INFO   ] [GL          ] OpenGL parsed version: 2, 1
[INFO   ] [GL          ] Shading version <1.20>
[INFO   ] [GL          ] Texture max size <16384>
[INFO   ] [GL          ] Texture max units <16>
on_enter screen size > [1, 1]
update on loop screen size > [800, 600]
update on loop screen size > [800, 600]
update on loop screen size > [800, 600]
```

### DIFF
--- a/kivy/uix/screenmanager.py
+++ b/kivy/uix/screenmanager.py
@@ -909,8 +909,8 @@ class ScreenManager(FloatLayout):
             self.transition.screen_out = previous_screen
             self.transition.start(self)
         else:
-            screen.pos = self.pos
             self.real_add_widget(screen)
+            self.do_layout()
             screen.dispatch('on_pre_enter')
             screen.dispatch('on_enter')
 


kived:
Good catch, but I think it would be better to force layout rather than assuming the `Screen` should be the same size as the `ScreenManager`.

Instead of:
````python
else:
    screen.pos = self.pos
    screen.size = self.size
    self.real_add_widget(screen)
````

Try this:
````python
else:
    self.real_add_widget(screen)
    self.do_layout()
````

That way both `pos` and `size` will be set as expected, and I will gladly merge this PR. :)

**Comment by fumiya-kubota** [comment]
@kived 
Thank you for reviewing.
I revised this commit.

**Comment by kived** [comment]
:+1:

File: kivy/uix/screenmanager.py [hunk 1]
```diff
@@ -909,8 +909,8 @@ def on_current(self, instance, value):
             self.transition.screen_out = previous_screen
             self.transition.start(self)
         else:
-            screen.pos = self.pos
```
akshayaurora: We still need to set the pos,  ScreenManager inherits from a FloatLayout which does not manage the position of it's children unless the child uses pos_hint.

PR #3088, opened 2015-02-28
use WeakMethod for property bindings

### LINKED ISSUE #3087: property callbacks are bound with strong refs instead of weak refs
In the following test code, even after the widget is removed the callback keeps firing because the callback is bound with a strong ref:

````python
import kivy
kivy.require('1.8.0')

import gc
import objgraph

from kivy.app import App
from kivy.clock import Clock

class HoverWidget(Widget):
	def __init__(self, **kwargs):
		super(HoverWidget, self).__init__(**kwargs)
		EventLoop.ensure_window()
		EventLoop.window.bind(mouse_pos=self.update_pos)
	
	def update_pos(self, win, pos):
		print '%dx%d' % pos

class TestApp(App):
	def build(self):
		root = FloatLayout()
		root.add_widget(HoverWidget())
		Clock.schedule_once(self.remove_label, 4)
		return root
	
	def remove_label(self, dt):
		print 'removing widget'
		c = id(self.root.children[0])
		self.root.clear_widgets()
		gc.collect()
		objgraph.show_backrefs([objgraph.at(c)], max_depth=5, refcounts=True)

if __name__ == '__main__':
	TestApp().run()
````

Graph:
screenshot from 2015-02-28 16 35 23

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9058296-property-callbacks-are-bound-with-strong-refs-instead-of-weak-refs?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### PR DESCRIPTION
**Pull request by kived**
Fixes #3087

screenshot from 2015-02-28 16 35 27

### DIFF
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -435,7 +435,7 @@ cdef class EventDispatcher(ObjectWithUid):
                 observers.fast_bind(WeakMethod(value), None, None, 1)
             else:
                 ps = self.__storage[key]
-                ps.observers.bind(value)
+                ps.observers.fast_bind(WeakMethod(value), None, None, 1)
 
     def unbind(self, **kwargs):
         '''Unbind properties from callback functions with similar usage as
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -891,11 +891,11 @@ cdef class EventObservers:
 
         while callback is not None:
             if is_ref and not callback.is_ref:
-                cb_equal = callback == observer()
+                cb_equal = callback.func == observer()
             elif callback.is_ref and not is_ref:
-                cb_equal = callback() == observer
+                cb_equal = callback.func() == observer
             else:
-                cb_equal = callback == observer
+                cb_equal = callback.func == observer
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: kivy/_event.pyx [hunk 1]
```diff
@@ -894,12 +891,16 @@ cdef class EventObservers:
         cdef BoundCallback new_callback
 
         while callback is not None:
+            if is_ref and callback.is_ref:
+                f = callback.func()
+            else:
+                f = callback.func
             if (callback.lock != deleted and callback.largs is None and
-                callback.kwargs is None and callback.func == observer):
+                callback.kwargs is None and f == observer):
```
matham: I'm trying to think what would happen here, since `observer` might be a ref, but it is not derefed before comparison.

matham:
+1

PR #3090, opened 2015-03-02
fix osc thread management on windows

tshirtman:


kivy/lib/osc/oscAPI.py
```diff
@@ -80,29 +81,42 @@ def _queue_message(self, message):
...
+                if thread_id is not None:
+                    if id != thread_id:
+                        continue
+                thread = oscThreads[id]
+
+        q = thread.queue
+        h = thread.addressManager.handle
+
+        while q:
+            h(q.popleft())
```
tshirtman: i guess one lookup could be saved by doing p = thread.queue.popleft and using h(p()), since it could be a tightloop, maybe that's better?

PR #3095, opened 2015-03-03
Update widget.py

Paradisee:
Add ValueError

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -464,13 +464,15 @@ class Widget(WidgetBase):
         canvas = self.canvas.before if canvas == 'before' else \
             self.canvas.after if canvas == 'after' else self.canvas
 
-        if index == 0 or len(self.children) == 0:
+        if index > len(self.children):
+            raise ValueError
+        elif index == 0 or len(self.children) == 0:
             self.children.insert(0, widget)
             canvas.add(widget.canvas)
         else:
             canvas = self.canvas
             children = self.children
-            if index >= len(children):
+            if index == len(children):
                 index = len(children)
                 next_index = 0
             else:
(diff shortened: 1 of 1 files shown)


kivy/uix/widget.py
```diff
@@ -464,13 +464,15 @@ def add_widget(self, widget, index=0, canvas=None):
...
+            raise ValueError
+        elif index == 0 or len(self.children) == 0:
             self.children.insert(0, widget)
             canvas.add(widget.canvas)
         else:
             canvas = self.canvas
             children = self.children
-            if index >= len(children):
+            if index == len(children):
                 index = len(children)
```
denys-duchier: that seems redundant.

Paradisee:
This is not consistent for me, because when add a widget to a Layout and set an index greater than the length of the children's list, you can't absolutely make it work! Instead you can Raise an IndexError. That's my point.

denys-duchier:
I mean: look at lines 475,476

matham:
I don't think this is a good idea. Why restrict it? What do we gain by raising an exception rather than by assuming that the user just wanted to stick it at the end? Also, it breaks backward compat.

Finally, python's list.insert allows indices that are larger than the list, so I don't think we should treat this any different. Just try `a = [1]`, and `a.insert(10, 2)` and you'll see it doesn't raise an exception.

denys-duchier:
I agree.  I vote -1

Paradisee:
Then i could put an index = 100000, where the list contains only 10 items. Coherence.

matham:
@Paradisee yes, exactly, if you put 100,000 as index, the assumption is that you meant to just put it at the end of the list. That's really not a surprising assumption, especially given that that's how python list behaves. Anyway, it does break backward compat for no apparent benefit, so I'm closing for now.

PR #3096, opened 2015-03-04
default is inverted now, also the invertion function was wrong

fscz:
tested on raspberry pi 2 with kernel from github/raspberrypi/firmware


inclement:
Does this require appropriate hardware to test?

fscz:
That is the question. I am running Ubuntu Linux armhf branch on an RPI 2 with a USB mouse, using the evdev input driver in Kivy. If you connect a USB mouse via evdev to Kivy on any other Linux installation I would expect the same behaviour and result - but who knows?

akshayaurora:
@fscz thanks.

PR #3107, opened 2015-03-05
make icons scale properly in actionbar

denys-duchier:
When I used the actionbar on my note4, the icons appeared extremely tiny.  This patch fixes the issue for me while also making the icon-sizing kv-parametrable.

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -573,6 +573,7 @@
     color: self.color[:3] + [0 if (root.icon and not root.inside_group) else 1]
 
     Image:
+        allow_stretch: True
         opacity: 1 if (root.icon and not root.inside_group) else 0
         source: root.icon
         mipmap: root.mipmap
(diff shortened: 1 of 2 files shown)


kived:
This works on my Note 4 with the actionbar.py demo. :+1: But the sizing doesn't really work right overall. If the images are larger than the widget, then the width will be large yet the image will still be scaled down to the height parameter:

example

The image set for the icon:

4circles

I would suggest making the icons default to being square, by defaulting the new properties to 0 and using `self.height` instead:

````diff
diff --git a/kivy/data/style.kv b/kivy/data/style.kv
index 13d3d75..7fc58b0 100644
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -591,8 +591,6 @@
     size_hint_x: 1
     minimum_width: '100sp'
     important: True
-    previous_image_width: dp(prev_icon_image.texture_size[0])
-    app_icon_width: dp(app_icon_image.texture_size[0])
     BoxLayout:
         orientation: 'horizontal'
         pos: root.pos
@@ -603,14 +601,18 @@
             opacity: 1 if root.with_previous else 0
             allow_stretch: True
             size_hint_x: None
-            width: root.previous_image_width if root.with_previous else dp(8)
+            width:
+                (root.previous_image_width or self.height) \
+                if root.with_previous else dp(8)
             mipmap: root.mipmap
         Image:
             id: app_icon_image
             source: root.app_icon
             allow_stretch: True
             size_hint_x: None
-            width: root.app_icon_width if self.texture else self.height
+            width:
+                (root.app_icon_width or self.height) \
+                if self.texture else self.height
             mipmap: root.mipmap
         Widget:
             size_hint_x: None
diff --git a/kivy/uix/actionbar.py b/kivy/uix/actionbar.py
index 51a4454..66520b2 100644
--- a/kivy/uix/actionbar.py
+++ b/kivy/uix/actionbar.py
@@ -157,7 +157,7 @@ class ActionPrevious(ActionButton):
        'data/logo/kivy-icon-32.png'.
     '''
 
-    app_icon_width = NumericProperty(10)
+    app_icon_width = NumericProperty(0)
     '''Width of app_icon image.
     '''
 
@@ -169,7 +169,7 @@ class ActionPrevious(ActionButton):
        defaults to 'atlas://data/images/defaulttheme/previous_normal'.
     '''
 
-    previous_image_width = NumericProperty(10)
+    previous_image_width = NumericProperty(0)
     '''Width of previous_image image.
     '''
 
````

denys-duchier:
I see what you mean, but now I am (slightly) unhappy for different reasons. I'll have to mull over this some more.

denys-duchier:
I have updated the patch to remove the squareness assumption that was bugging me. now the icons are scaled if they are too tall; but you can have very wide icons.

kived:
Looks good! :+1:

PR #3114, opened 2015-03-08
The type of data is converted to bytes from str for packing

Sentient07:
According to documentation of format for struct.pack(fmt, val1,val2..) , the argument for ">s" must be a "bytes" object and not string (https://docs.python.org/3.2/library/struct.html#format-characters ).

### DIFF
--- a/kivy/lib/osc/OSC.py
+++ b/kivy/lib/osc/OSC.py
@@ -179,7 +179,7 @@ def OSCArgument(data):
 
     if isinstance(data, string_types):
         OSCstringLength = math.ceil((len(data)+1) / 4.0) * 4
-        binary = struct.pack(">%ds" % (OSCstringLength), data)
+        binary = struct.pack(">i%ds" % (OSCstringLength), bytes(data, 'utf-8')
         tag = "s"
     elif isinstance(data, float):
         binary = struct.pack(">f", data)


FeralBytes:
I think this would require a different behavior for Python3 vs Python2. bytes() in Python 2.7 is an alias for str() this is also true of 2.6 but only in more recent versions of 2.6. In the case of 2.5 there is no alias and so it would just fail.
In buildozer this difference of behavior was ahieved with this code:
```python
IS_PY3 = sys.version_info[0] >= 3
if IS_PY3:
    print('Do Py3 stuff')
else:
    print('Do Py2 stuff')
```
But I would still wait for another Kivy dev to weigh in.

kivy/lib/osc/OSC.py
```diff
@@ -179,7 +179,7 @@ def OSCArgument(data):
 
     if isinstance(data, string_types):
         OSCstringLength = math.ceil((len(data)+1) / 4.0) * 4
-        binary = struct.pack(">%ds" % (OSCstringLength), data)
+        binary = struct.pack(">i%ds" % (OSCstringLength), bytes(data, 'utf-8')
```
kived: Why do you add an int to the struct? Note that `OSCstringLength` is not being packed, it's just modifying the format string to set the data string length.

FeralBytes:
OK after some discussion and getting educated some. I believe the proper way to implement this PR would be to write it like so:
First make sure OSC imports:
```python
import kivy.compat.PY2
```
```python
if kivy.compat.PY2 is False:
    if isinstance(data, unicode): 
        data = data.encode('utf-8')
binary = struct.pack(">%ds" % (OSCstringLength), data)
```

tito:
Good to me. if anything is wrong, @tshirtman will be the first one to tell :p

RedXBeard:
I'm a bit late obviously but, on that line there is a missing parenthesis :s

tshirtman:
i was about to say that :D

tshirtman:
this was in fact still not fixed after 4b85a47, @FeralBytes way seems to work with python2 at least, but for some reason i tuio doesn't seem to work for me on py3, looking into it.

tshirtman:
Ok, there is a lot to fix in OSC.py and OscAPI.py for py3, it certainly didn't work before that, not a regression.

PR #3115, opened 2015-03-08
Handle CWOverrideRedirect different: Instead of using CWOverrideRedirect...

jegger:
... when no-border is set, try to hide (via wm-hints) the window decoration. Let it be possible to set CWOverrideRedirect via env-variables (for special cases).

@tito : I know it's a long time but do you remember why you set CWOverrideRedirect (back then) when border=False? (Any special cases which I might not see?)


tito:
@jegger i don't remember at all, as the current code was based on some others examples i found. Let's merge it.

PR #3116, opened 2015-03-08
 report.py sends report to https://gist.github.com/

Sentient07:
report.py sends the reports to https://gist.github.com/  with each report as individual file generating a single gist link


tito:
great :) :)

PR #3119, opened 2015-03-09
Fixes crash using latest pyjnius

emanuele-f:


tito:
@akshayaurora Do you know why toString was used, and not needed anymore?

emanuele-f:
I've noticed that line [line 66](https://github.com/kivy/kivy/blob/master/kivy/core/clipboard/clipboard_android.py#L66]) uses toString method as well.
I've found these relevant documentation links:

http://developer.android.com/reference/android/text/ClipboardManager.html - for the api <11 code chunck

http://developer.android.com/reference/android/content/ClipData.Item.html - for the api >= 11 code chunck

They both state getText() returns a CharSequence. Hope it helps.

tito:
Care, on the last pyjnius version, CharSequence are correctly decoded/encoded. It wasn't the case before. Maybe that's why?

tito:
Ie, java.lang.String was mapped to a python string, but not the CharSequence. It's now mapped to python string as well.

akshayaurora:
That's probably the reason, here. I distinctly remember getting errors using the earlier versions of pyjnius.

Issue #3120, opened 2015-03-09
Keyboard behavior in select_for_key_down elides over pauses, combining keys.


merriam:
The design of 'select_for_key_down()' in behaviors.py allows for multiple key strokes to be returned.  For example, running examples/widgets/compound_selection.py, typing "13" first sees an input of '1' and then aggregates it for a second input of '13'.   However pausing will randomly reset the typing button, so waiting a few seconds and typing '13' again returns aggregates of '131' and '1313'.   It should have reset and returned '1' and '13' as before.

The issue is calling time.clock() instead of time.time().  A fix can be found in https://github.com/merriam/kivy/commit/d735622ad5802d6d42a2a739ab83a7521a34e6d4, an
unmerged PR.

tito:
Hi @merriam,

I saw you answered to a numerous issues referencing your closed PR. Just to be sure, we never intented to reject your work, but submitting thousands of files with unrelated changes is not going to work. It is making the exchange harder, as we (as maintainer) can delay the merge until all the fixes are "ok", which can delay all the fixes that were ok as well.

Tiny atomic commit are the key for a single bug fix, or a bunch of commit on the same new features / refactoring, as long as it's the exact same subject.

On this specific issue, thanks again for finding it and providing a fix, you can resubmit a PR by doing this step:

- Fetch all the latest ref in master: `git fetch -a`
- Create a new branch: `git checkout --track -b fix-3120 origin/master`
- Pick only this commit and merge it into the new branch: `git cherry-pick d735622`
- Publish your branch on your repo: `git push origin fix-3120`
- Then, go back to github, and it should show you the possibility to create a PR from your newly fresh branch: fix-3120.

Thanks you :)

merriam:
Hi Mathieu,

Thank you for your kind note.   It does appear from previous exchanges that
the intent is indeed to reject any work I directly submit.  The larger
block of changes was submitted because every smaller block was nit-picked,
ignored, or randomly closed.   The nit-picks, usually about the number of
spaces after a period in comments, do not appear to be about the quality of
the submission, any errors, or any substantive issue.   In fact, each PR I
submit usually fixes these nits in other's accepted PRs.

I provide tiny, atomic commits.   I tried side-stepping the PR nit problem
by putting many commits in one PR.   As each uncommitted pull request is a
separate branch, I need to provide a reasonable limit to the number of open
and disputed Pull Requests.   I set that at about five or six.

[...]

PR #3122, opened 2015-03-10
Fixes 3120 Credit to @merriam

FeralBytes:
The keyboard input would clump characters together in a continuously
growing stream instead of timing out after a seond.  That is, typing
’12’, pausing, then typing ’34’ would register as a single string
‘1234’.   This showed up in the widgets/compound_selection.py example
and possible in issues with keyboard recognition problems.  The issue
is that time.clock() returns a vaguely defined ‘time on processor’
whereas time.time() is the number of seconds.

### DIFF
--- a/kivy/uix/behaviors.py
+++ b/kivy/uix/behaviors.py
@@ -38,7 +38,7 @@ from kivy.base import EventLoop
 from kivy.logger import Logger
 from functools import partial
 from weakref import ref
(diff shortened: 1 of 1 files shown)


FeralBytes:
Fixes [3120](https://github.com/kivy/kivy/issues/3120)
All credit to @merriam.

PR #3125, opened 2015-03-10
Original texture is displayed along with combined texture

Sentient07:


tito:
Good enough, thanks :)

PR #3128, opened 2015-03-12
introduce ActionGroup.dropdown_width property

denys-duchier:
This patch adds the dropdown_width property to ActionGroup and makes it easy to accommodate a dropdown list with wide items.  Full disclosure: it is already possible (but imho much less intuitive) to do something similar using the minimum_width property on the LAST item.

### DIFF
--- a/kivy/uix/actionbar.py
+++ b/kivy/uix/actionbar.py
@@ -265,6 +265,10 @@ class ActionGroup(ActionItem, Spinner):
        defaults to 'normal'.
     '''
 
+    dropdown_width = NumericProperty(0)
+    '''If non zero, provides the width for the associated DropDown.
+    '''
+
     def __init__(self, **kwargs):
         self.list_action_item = []
         self._list_overflow_items = []
@@ -307,7 +311,7 @@ class ActionGroup(ActionItem, Spinner):
         children = ddn.container.children
 
         if children:
-            ddn.width = max([self.width, children[0].minimum_width])
+            ddn.width = self.dropdown_width or max([self.width, children[0].minimum_width])
         else:
             ddn.width = self.width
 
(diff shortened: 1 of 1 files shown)


kivy/uix/actionbar.py
```diff
@@ -291,7 +295,7 @@ def _toggle_dropdown(self, *largs):
         children = ddn.container.children
 
         if children:
-            ddn.width = max([self.width, children[0].minimum_width])
+            ddn.width = self.dropdown_width or max([self.width, children[0].minimum_width])
```
FeralBytes: This always assigns what ever self.dropdown_width, which is the left side of the or statement. But if I don't set that, then my width will always be 0.

kivy/uix/actionbar.py
```diff
@@ -249,6 +249,10 @@ class ActionGroup(ActionItem, Spinner):
        defaults to 'normal'.
     '''
 
+    dropdown_width = NumericProperty(0)
+    '''If non zero, provides the width for the associated DropDown.
```
FeralBytes: In my opinion, your documentation is too minimal, especially for a new property.
Should be more like:
```python
'''If non zero, provides the width for the associated DropDown.
.. versionadded:: 1.9.0
:attr:`dropdown_width` is an :class:`~kivy.properties.NumericProperty` and defaults to 0.
'''
```

kivy/uix/actionbar.py
denys-duchier: it should definitely prefer dropdown_width. that was the whole point.

kivy/uix/actionbar.py
denys-duchier: I agree.  I even wanted to add a test and that's when I discovered that the graphics tests are messed up https://github.com/kivy/kivy/issues/3130

kived: `versionadded` should come after the property/default descriptions, but both changes are required before this PR can be merged.

KeyWeeUsr:
@denys-duchier Please resolve the conflict in actionbar.py and add the documentation properly (order: description, property, version). Seems to be good for merge. ^^

dessant:
Same as https://github.com/kivy/kivy/pull/3625#issuecomment-264732811.

Issue #3126, opened 2015-03-11
ScreenManager API inconsistency


rnixx:
Hi,

While working with ScreenManager i ran into an issue with switch_to which i consider an API inconsistency.

While it's documented, i think the behavior of switch_to with auto removing the previous displayed screen should only apply on demand. It should also be possible to switch to already added screens.

E.g.:

I have several screens in my manager. One of them is a Status Page which gets displayed on start and on resume, displaying the initialization progress, and if an error happened. All others screens are application workflow related -> and get changed with WipeTransition (or whatever).

Now it's desired, on resume, to use NoTransitions when switching to status page, to avoid weird UI look and feel. But since the status page is already a child of the manager, using switch_to with specified transition won't work any longer since status page is already contained.

Furthermore this construct https://github.com/kivy/kivy/blob/master/kivy/uix/screenmanager.py#L1019 seem to have zero benefit. In my understanding, using names as unique identifyers also imply beeing aware of the used names. Imho checking for the name and generating a substitute for it if name already exists (for another screen!) feels error prone and tries to be too clever at the wrong place.

From user POV i would expect switch_to behave somewhat like:
* Check if given screen already added.
* If not added yet, check screen name and raise exception if screen with name already exists, otherwise add to manager.
* switch to screen
* remove previous screen if remove_prev (or similar) kwarg is True

Regards,

Robert

rnixx:
https://github.com/kivy/kivy/blob/master/kivy/uix/screenmanager.py#L1030
This changes the options of manager.transition permanently if no custom transition given

rnixx:
I Implemented the proposed function here. It has a new name to avoid breaking of existing implementations depending on the other behavior.

https://github.com/rnixx/kivy/commit/bfad546b6992f9a3d7083cae426f3b1bdc9c86e4

rnixx:
digging around a little more -> https://github.com/kivy/kivy/blob/master/kivy/uix/screenmanager.py#L944 -> this statement is dead code. screen names must be unique.

PR #3133, opened 2015-03-13
Implement ScreenManager.display_screen function

rnixx:
Here's the pull request related to #3126

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,14 @@ python.exe
 dist
 htmlcov
 
+bin/
+include/
+lib/
+local/
+man/
+pip-selfcheck.json
+share/
+
 MANIFEST
 
 kivy/core/image/img_imageio.c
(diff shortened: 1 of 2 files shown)


inclement:
Sorry for the delay reviewing this. I guess the point is that this should be a better (and more flexible) replacement for switch_to?

It seems good (and we need such a thing), but I think a better solution would be to merge the two functions and have a default parameter option to remove the previous screen. That way the api does not become more cluttered. I'll see what others think.

dessant:
I second that, the new behavior should be integrated in switch_to.

rnixx:
@inclement - the idea behind using a new signature is not to break code depending on the existing behavior, deprecate switch_to in favor of display_page and remove switch_to after a while.

the remove flag is already implemented in display_page

rnixx:
@inclement the motivation behind the new function is documented in #3126

rnixx:
anyone else? how to handle this further? should i simply replace switch_to now?

inclement:
I do think this should be integrated with switch_to. Sorry for not replying earlier, we did discuss this a little on irc and had some disagreement about it - though I don't think anyone preferred the option of adding a new method name.

rnixx:
The related code can be found at https://gist.github.com/rnixx/2aab5e13e17c1582b6dd

Issue #2693, opened 2014-11-20
Scale() should have an origin property


sbrother:
Rotate() now supports an origin property, allowing rotations around an arbitrary origin. Scale does not, only supporting scaling around the lower left of the canvas. This means that to achieve control over scaling a widget, you need to wrap Scale in two Translate commands. It would be nice if Scale also had an origin property to avoid this.

rastogiachyut:
Working on this one now.

rastogiachyut:
Really working on it now.

rastogiachyut:
Stuck.I am adding origin property to scale exactly like it was done in rotate, it still does not work.
This( https://gist.github.com/rastogiachyut/afc7dd9d0526ac3c894f ) code does not work, this is my Scale class in context_instruction.pyx file( https://gist.github.com/rastogiachyut/1c82e1c9d23b3c11ef11 ) ,Why? Ive hit a dead end. Probably I am doing something stupid.

hey-sancho:
Here's my try at fixing this:
https://gist.github.com/hey-sancho/b131a6d02d9f15838beb

Tried to do the same way as in `Rotate` - everything uses `compute()` function, which translates and scales the object. Rewritten `set_scale()` to work via `compute()`, but i guess it could become deprecated as the `Rotate.set`.

@rastogiachyut, I don't know what result did you get, but my version makes your code look like this:
![screenshot - 26 02 15 - 18 27 31](https://cloud.githubusercontent.com/assets/1079585/6396144/7d6ba146-bde5-11e4-93f0-2f8a6b3c4958.png)

rastogiachyut:
@hey-sancho 
Sorry for the late reply.
I submited a PR for this a while but it did not get pulled.
here -> https://github.com/kivy/kivy/pull/3070
your fix looks right too.

rastogiachyut:
I don't know if your code works in all cases, mine does.
But yours looks better cause it makes a new function compute.
Well make a PR and let the devs decide.

rastogiachyut:
made a clean new pull request https://github.com/kivy/kivy/pull/3134 
please pull if satisfactory.

PR #3134, opened 2015-03-13
Added scale w.r.t origin property

rastogiachyut:
Added scale(issue #2693 ) with respect to origin property, similarly to how it was done in rotate. The set_scale method coresponds to compute method in rotate.

### DIFF
--- a/kivy/graphics/context_instructions.pxd
+++ b/kivy/graphics/context_instructions.pxd
@@ -65,6 +65,7 @@ cdef class Rotate(Transform):
     cdef void compute(self)
 
 cdef class Scale(Transform):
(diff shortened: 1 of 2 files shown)


rastogiachyut:
two files changed in one commit because if I did it in two there would have been a state in between where the code would be broken.

rastogiachyut:
two files changed in one commit because if I did it in two there would have been a state in between where the code would be broken.

inclement:
Looks good to me, thanks! I've also added a new example demonstrating this feature.

rastogiachyut:
Sorry for  #3070, should have closed that one myself.

Issue #1456, opened 2013-08-22
SDL2 - add new config options to the graphics section: maximized, minimized, hidden


voen:
There is no easy way to maximize a window from kivy.
It could be:

<pre><code>
#maximize: (0, 1)
#Default is 0.
from kivy.config import Config
Config.set('graphics', 'maximize', 1)
</pre></code>

UPDATE:
These options were already added as methods for the sdl2 window provider.
Pygame mostly lacks these flags, i think it should be best if we'd add this when sdl2 will be the default on desktop.

PR #3136, opened 2015-03-13
Add maximized, minimized and hidden config options to the graphics section

dessant:
This is a continuation of https://github.com/kivy/kivy/pull/2590 and should fix https://github.com/kivy/kivy/issues/1456.

For some reason `minimized` has no effect on Debian Jessie with Gnome, but `Window.minimize()` works regardless.

```python
Config.set('graphics', 'maximized', 1)
# Config.set('graphics', 'minimized', 1)
# Config.set('graphics', 'hidden', 1)

# from kivy.core.window import Window
# Window.minimize()

from kivy.app import App


class TestApp(App):
    def build(self):
        return Button()

if __name__ == '__main__':
    TestApp().run()
```

### DIFF
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -239,7 +239,7 @@ Available configuration tokens
     is required for value changes to take effect.
 
 .. versionchanged:: 1.9.0
-    `borderless`, `maximized`, `minimized` and `hidden` has been added
+    `borderless`, `maximized`, `minimized` and `hidden` have been added
     to the graphics section.
     The `fake` option of `fullscreen` in the graphics section has been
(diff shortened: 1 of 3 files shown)


inclement:
Looks good to me, but this can't be merged automatically any more, could you rebase?

dessant:
Sure, but we had a discussion with @matham and he suggested to have there three states in one option, `window_start` or  `window_state`. I see the logic behind it, but i also feel it might be a wrong approach because it is less explicit. @inclement what is your opinion?

dessant:
Closing in favor of https://github.com/kivy/kivy/pull/3200.

PR #3138, opened 2015-03-13
fix #3118: kwargs ignored if property set via kv

### LINKED ISSUE #3118: kwargs ignored if property set via kv
````python
import kivy
kivy.require('1.8.1')

from kivy.app import App
from kivy.lang import Builder

Builder.load_string('''
<Button>:
    text: 'bad'
''')

class TestApp(App):
    def build(self):
        return Button(text='good')

if __name__ == '__main__':
    TestApp().run()
````

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9367446-kwargs-ignored-if-property-set-via-kv?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

denys-duchier:


akshayaurora:
I think this should already be fixed by https://github.com/kivy/kivy/pull/3469.
@denys-duchier do you have a reproducible snippet we can use to test this issue?

dessant:
I can reproduce with https://github.com/kivy/kivy/issues/3118#issue-60371530

akshayaurora:
Thinking more about this. Events and properties are set after widget instantiations  through kv and kv > py in this regard. 

So kv properties are set after the initialization, this is normal and expected behavior since a long time. Is there a valid reason to reverse this behavior?

This is how I expect things to be and it is intuitive that things set in kv take precedence over widget initialization(widget is initialized then kv rules/props are set), the other way around will be highly unintuitive.

dessant:
Closing in favor of https://github.com/kivy/kivy/pull/3736.

Issue #3143, opened 2015-03-14
Screen already managed by another ScreenManager (kivy users)


denys-duchier:
as reported on kivy-users, the showcase demo crashes when using any of the ShaderTransitions.
##

denys-duchier:
see PR #3144

akshayaurora:
@denys-duchier I don't see the crash. Could you post a log?

denys-duchier:
the exception happens once you have moved to the next screen and try to get back to the previous screen.

denys-duchier:
Below are the modifs you need to exercise this bug; then run the showcase demo, click on the right-facing caret to go to the next screen, then click on the left-facing caret to go back to the previous screen, and that's when you get an exception:

``` diff
diff --git a/examples/demo/showcase/showcase.kv b/examples/demo/showcase/showcase.kv
index a81d062..76dacde 100644
--- a/examples/demo/showcase/showcase.kv
+++ b/examples/demo/showcase/showcase.kv
@@ -35,6 +35,7 @@
                 height: self.height if root.fullscreen else self.minimum_height
 
 
+#:import SM kivy.uix.screenmanager
 BoxLayout:
     orientation: 'vertical'
 
@@ -95,6 +96,7 @@ BoxLayout:
 
     ScreenManager:
         id: sm
+        transition: SM.FadeTransition()
         on_current_screen:
             spnr.text = args[1].name
             idx = app.screen_names.index(args[1].name)
```

To understand the problem, I instrumented, in class `ScreenManager`, the methods `add_widget`, `real_add_widget`, `remove_widget`, and `real_remove_widget`, and for each one I printed the argument widget and the stack trace.

On the **good** case (i.e. with `SlideTransition`), the following happens when you click on a caret::

    on_release > go_next_screen > switch_to >
        add_widget
        real_add_widget

    The animation runs,  then

    stop > on_complete >
        remove_widget
        remove_screen > real_remove_widget

On the **bad** case (i.e. with `FadeTransition`), the following happens::

    on_release > go_next_screen > switch_to >
        add_widget
        real_remove_widget

    The animation runs, then

    stop > on_complete >
        remove_screen > real_add_widget

Notice that the previous screen has not gone through `remove_widget`, therefore when you attempt to move back to it, the code goes into `add_widget` with a screen that was **not** removed; and that's the reason for the exception.

My patch modifies the bahaviour of `ShaderTransition` so that it conforms again to the logic used in `TransitionBase`.

[...]

PR #3144, opened 2015-03-14
fix screen logic fo ShaderTransition (bug #3143)

denys-duchier:


matham:
Merged in https://github.com/kivy/kivy/pull/6279.

PR #3153, opened 2015-03-17
core/audio: new pure-python/pyobjus AVFoundation audio player

tito:
This extend our audio provider with a native provider, and supports far
more format than the current sdl2 (mp3, ogg, wav).

The full list is available at
https://developer.apple.com/library/ios/documentation/MusicAudio/Concept
ual/CoreAudioOverview/SupportedAudioFormatsMacOSX/SupportedAudioFormatsM
acOSX.html


tito:
1.9 is out, let's merge this one :)

PR #3154, opened 2015-03-17
examples added

picibucor:
these examples explains how id's work.


dessant:
Thanks for the contribution. :+1: It will be reviewed soon after 1.9.0 is released.
As a general recommendation, don't use spaces in folder or file names, you could replace them with underscores.

inclement:
The content looks good to me, though it will indeed be necessary to use filenames without spaces. I think an appropriate directory might be examples/kv/ids - what do others think?

dessant:
@inclement +1.

picibucor:
I got stuck:
fail

examples/kv/ids/kv_and_py/kv_and_py.py
```diff
@@ -0,0 +1,33 @@
...
+kivy.require('1.8.0')
+
+
+class RootWidget(BoxLayout):
+
+    def first_function(self, status):
+        # print out the given parameter
+        print(status)
+        # check the status of the switch by referring on the id
+        if self.ids['my_switch'].active is True:
```
dessant: You could do `self.ids.my_switch.active` instead, so it conforms with the rest of the examples and docs.

dessant:
Ready for merging after that small fix. :+1:

picibucor:
@dessant : done

dessant:
Thanks a lot. :cake:

PR #3162, opened 2015-03-18
Docs add books section

hackebrot:
Hey there,

this PR adds a new section about Kivy books to the docs.

Although I don't think that it's strictly required, it may be a good starting point for beginners in addition to the tutorials that we already have.

Please let me know your thoughts and I'll be happy to change the PR accordingly. :smile: 

/CC @robertour @buchuki @mvasilkov


mvasilkov:
Wow, this is flattering :)

kived:
This would be a better fit in a new article on the [Kivy GitHub wiki](https://github.com/kivy/kivy/wiki/KivyBooks), so that authors can add their own books and Kivy devs don't have to be responsible for maintaining the list. It should also include a disclaimer that this is just a list of books which have been written about Kivy, and that inclusion in this list does not imply endorsement of the books by the Kivy Organization or any of its developers, and there is no guarantee regarding the accuracy of the information presented in the books.

After you've added the page to the wiki, please submit a new PR adding a link to the KivyBooks wiki in the docs.

hackebrot:
Ok cool, thank you for your feedback @kived :bow:

I agree that having some kind of disclaimer is required. I'll be happy to add an according wiki page soon.

PR #3164, opened 2015-03-20
Fixed iOS support for iPhone 6 plus, retina iOS devices, and start app i...

matafc:
...n portrait mode.

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -25,7 +25,7 @@ cdef class _WindowSDL2Storage:
         self.win_flags = SDL_WINDOW_OPENGL | SDL_WINDOW_SHOWN
 
         IF USE_IOS:
-            self.win_flags |= SDL_WINDOW_BORDERLESS | SDL_WINDOW_RESIZABLE | SDL_WINDOW_FULLSCREEN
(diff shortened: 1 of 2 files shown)


akshayaurora:
@matafc I was investigating in fixing the res on both retina mac books and on iOS on branch `retina` on kivy.

Just pull latest kivy and checkout retina branch.

This also fixes the size reported by window.size and uses SDL_GL_GetDrawableSize to get the actual internal size of the window.

This right now works on iOS tested on iPad Air, iPhone 6plus, iPad2 emulators.

The work on retina MB/MBP is still a WIP, we get the display as we like on the retina screen but moving from a retina to a non-retina screen is still being fixed, so is mouse pos.

There is no need to setup the orientation in setup window as in your pr and we need to use GL_DRAWABLE_SIZE for retina screens on desktop too.

Can you ping me on irc if you are available right now?

matafc:
Sorry, I don't use irc. The pull request contains the minimum changes to get ios projects in a presentable state again. Orientation was required, not because it was crashing, but it would start in landscape mode on ios (even when holding the phone in portrait mode), then switch to portrait mode. What I added fixed that.

OSX is not my build target, but when I run on my retina macbook locally(kivy main.py), it seems to use pygame instead of sdl2, which displays everything fine. 

Use what I changed if you like. I'm using my custom build of kivy to release because I already had to wait over 2 months for all the ios issues to be fixed.

Thanks for reviewing my pr so quickly.

akshayaurora:
@matafc it's ok I think I can manage my work on top of your pr, thanks a ton for this nice work!

PR #3165, opened 2015-03-20
fix scaling for border image

Kovak:
Currently if the value of borders greatly exceeds the size of the image on rendering we can get some very odd results:
With the default image:
1's size is 10, 10
2's size is 25, 25
however they appear roughly the same size. The issue appears even more obvious if we use a very large button graphic.

If the image is being displayed much smaller than its true size we want to find the normalized size of the borders values and then scale them by the actual size being displayed, however we never want to exceed the original value of borders.

example1

After changes:
example2

Test code:

```python
from kivy.app import App
from kivy.lang import Builder

kv = """
<Test>:
    Button:
        text: '1'
        size_hint: (None, None)
        pos: 50, 50
        size: 10, 10
    Button:
        text: '2'
        size_hint: (None, None)
        pos: 80, 45
        size: 25, 25
    Button:
        text: '1'
        size_hint: (None, None)
        pos: 150, 50
        background_normal: 'big_button.png'
        size: 400, 400
        border: [125, 125, 125, 125]
    Button:
        text: '2'
        size_hint: (None, None)
        background_normal: 'big_button.png'
        pos: 550, 350
        size: 200, 200
        border: [125, 125, 125, 125]

"""

Builder.load_string(kv)


class Test(FloatLayout):
        super(Test, self).__init__(*args, **kwargs)

class TestApp(App):
    def build(self):
        return Test()

if __name__ == '__main__':
    TestApp().run()
```
test large button:
big_button

### DIFF
--- a/kivy/graphics/vertex_instructions.pyx
+++ b/kivy/graphics/vertex_instructions.pyx
@@ -876,13 +876,17 @@ cdef class BorderImage(Rectangle):
         tb[2] = b2 / th * tch
         tb[3] = b3 / tw * tcw
 
-
+        cdef float sb0, sb1, sb2, sb3
+        sb0 = min((b0/th) * h, b0)
+        sb1 = min((b1/tw) * w, b1)
+        sb2 = min((b2/th) * h, b2)
+        sb3 = min((b3/tw) * w, b3)
         # horizontal and vertical sections
         cdef float hs[4]
         cdef float vs[4]
         hs[0] = x;            vs[0] = y
-        hs[1] = x + b3;       vs[1] = y + b0
-        hs[2] = x + w - b1;   vs[2] = y + h - b2
+        hs[1] = x + sb3;       vs[1] = y + sb0
+        hs[2] = x + w - sb1;   vs[2] = y + h - sb2
         hs[3] = x + w;        vs[3] = y + h
 
         cdef float ths[4]


luminosuslight:
This commit makes BorderImages with small borders appear larger than they should. The problem is that I don't know how to fix this properly. Maybe @Kovak knows a solution.

Here is a screenshot of the new "bug":
border_image_bug

Kovak:
Can you provide some more details on when this occurring? (Hopefully a code example) The changes I made will never make a Border appear larger than its sizing, only scaling if it is smaller than the provided sizing.

luminosuslight:
Ok, I was fooled by the transparent shadow: The total BorderImage area has the right size but inside the borders are too small.

Here is a short example:
```python
from kivy.app import App
from kivy.lang import Builder

Builder.load_string("""
<Test>:
    canvas:
        Color:
            rgb: 1, 1, 1
        Rectangle:
            pos: self.pos
            size: self.size
    Widget:
        pos: 200, 200
        size: 64, 64
        size_hint: None, None
        canvas:
            Color:
                rgb: 1, 1, 1
            BorderImage:
                pos: self.pos
                size: self.size
                source: "border32.png"
                border: 32, 32, 32, 32
""")


class TestApp(App):
    def build(self):
        return Test()
      
class Test(FloatLayout):
    pass

if __name__ in ('__android__', '__main__'):
    TestApp().run()
```
Image for the code above:
border32

The 32px border should fill the whole 64px wide widget like this:
expected_result

But instead it looks like this:
border_bug

Kovak:
This is precisely how the behavior was changed: If you provide a BorderImage of size larger than the actual widget you are drawing instead of  'consuming' all (or potentially more than all) the space with border we scale down now so that the expected ratios are preserved. 

Imagine for instance instead you tried to draw a 34 px border to your same 64 px image. We would now have a border image being drawn with the right border to the left of the left border and the left border to the right of the right border.

luminosuslight:
Ah, ok, I think I got the idea :-)
I just tested it with a border image that contains only the four edges ( -> 64x64px) and it works as expected.
I'm sorry if I confused you ;-)

Kovak:
No problem, this is definitely a secret 'break' in this change here. If you want to have your border image be respected in border size exactly it must be <= to the size of the element.

tshirtman:
I'm really wondering about that fix, sure, it makes things better for this corner case, but it break things that worked previously, and it's not always obvious how to fix them (i did cut my textures as explained up, but now i have size/position issues with my BorderImages, and have to find where it comes from), it's a nice feature, but it was more obvious before what the issue was when there was a size mismatch.

Maybe there should be a flag to activate this scaling.

PR #3171, opened 2015-03-21
clipboard_dbusklipper and clipboard_xel: return list not string

Daetalus:
the get_types in clipboard_dbusklipper and clipboard_xel return string, rewrite it to let it return list.This is a small contribution that prepare for GSoC 2015.


akshayaurora:
:+1:

PR #3179, opened 2015-03-22
improve TextInput functionality

kived:
This does a few things to improve the functionality and usability of `TextInput`:

1. Text is always unicode (best practice; also fixes issues with cursor positioning and utf8-encoded byte strings)
2. Adds `replace_crlf` property, defaults to `True`
3. Supports modifier keys:
    - Ctrl+Left/Right - move cursor one word to the left/right (password==False)
    - Ctrl+Up/Down - scroll one line up/down (multiline==True)
    - Ctrl+Home/End - go to beginning/end of text
    - Alt+Up/Down - shift line(s) up/down (multiline==True, readonly==False)
4. Triple-tap selects entire line, even when wrapped


akshayaurora:
@kived we maintain a unicode sandwich.
i.e text is always made sure to be unicode inside for our internal usage, then it is converted back to str in py2  for usage in other libs.

It is often desirable to pass the text directly to other libs which expect the input to be str on py2.

Unicode sandwich is also the suggested way/"Best Practice" for handling text in py2, 

The issues with py3 are not the same but it still does not solve the unicode issue completely so keeping the part of converting to unicode internally is also necessary on py3.

kivy/uix/textinput.py
```diff
@@ -835,7 +856,186 @@ def _set_undo_redo_bkspc(self, ol_index, new_index, substring, from_undo):
...
+
+        c = text[pos]
+        if c in ws:
+            mode = 'ws'
+        elif c == '_':
+            mode = 'us'
+        elif c in punct:
+            mode = 'punct'
+        elif c in lcase:
+            mode = 'camel'
```
akshayaurora: isn't this the same code as above, should we not move the common bits to a common function and not duplicate?

kivy/uix/textinput.py
```diff
@@ -835,7 +856,186 @@ def _set_undo_redo_bkspc(self, ol_index, new_index, substring, from_undo):
         #reset redo when undo is appended to
         self._redo = []
 
-    def do_cursor_movement(self, action):
+    def _move_word_left(self, index=None):
```
akshayaurora: Could you please rename it to _move_cursor_...? would be more clear while reading the code.

akshayaurora: In fact I thunk it should be possible to combine these two functions on to one `_move_cursor(movement_mode='word-left')`?

Where movement_mode can be one of `word-left` `word-right`, `line-up`, `line-down`, 'home', `end`?

kivy/uix/textinput.py
```diff
@@ -835,7 +856,186 @@ def _set_undo_redo_bkspc(self, ol_index, new_index, substring, from_undo):
...
+        if c in ws:
+            mode = 'ws'
+        elif c == '_':
+            mode = 'us'
+        elif c in punct:
+            mode = 'punct'
+        elif c not in ucase:
+            mode = 'camel'
+
+        while 0 < pos:
```
akshayaurora: I don't understand the need for going letter by letter for this, Can you explain why is this needed?

kived:
@akshayaurora A Unicode sandwich means all text inside the program is unicode. This would include the text in the TextInput. It should never be converted to bytes until it is being passed out of the program or to a lib which requires it. Also, without this change, `get_cursor_from_index()` doesn't work as soon as you have multi-byte codepoints in your text. But regardless, everything inside the TextInput should be unicode - it is up to the app developer to encode to bytes when necessary, or up to us when a provider or something else we are passing it to requires it.

As for the move functions, they are not the same. They apply rules slightly differently and in different orders. The reason it goes character by character is to provide a more useful implementation via a state machine. This doesn't just go to the next space - it's designed to mimic the functionality in PyCharm, not something like Chrome or Word. I'm not worried about performance, as it happens in response to user input and won't generally have to check many characters before finding a stopping point.

I can definitely rename the functions though. :)

akshayaurora:
If some function is converting it to bytes then we have issues cause last I
checked we were ensuring unicode inside,  maybe some commits got in that
were not ensuring we do this?

We should ensure that we do maintain the unicode sandwich.

The `text` property uses a getter which converts all the lines into one
continuous text, this is where we provide the text for consumption by other
libs and other widgets.

Whether we should be doing the conversion back to original type here or
leave it; is something we didn't yet have a proper discussion about, maybe
we should.

kived:
This is ready for review again. I've made some changes, mainly to make the functions use the lines list instead of compiling the text every time.

PR #3180, opened 2015-03-22
Cython22

DragonSA:
Add support for cython 0.22.

### DIFF
--- a/kivy/graphics/context_instructions.pxd
+++ b/kivy/graphics/context_instructions.pxd
@@ -16,39 +16,39 @@ cdef class PopState(ContextInstruction):
     pass
 
 cdef class LineWidth(ContextInstruction):
-    cdef void apply(self)
+    cdef void apply(self) except *
 
 cdef class Color(ContextInstruction):
-    cdef void apply(self)
+    cdef void apply(self) except *
 
 cdef class BindTexture(ContextInstruction):
     cdef int _index
     cdef object _source
     cdef Texture _texture
-    cdef void apply(self)
+    cdef void apply(self) except *
 
 
 cdef class LoadIdentity(ContextInstruction):
     pass
 
 cdef class PushMatrix(ContextInstruction):
-    cdef void apply(self)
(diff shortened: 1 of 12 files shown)


tito:
Good to me. @kived ?

akshayaurora:
Won't this result in a significant performance hit?

tito:
no idea, need to be tested.

kived:
I already have a PR which does the same thing. But using except * does cause a performance hit, as the interpreter will check for an exception after every call to those functions. We discussed switching the functions all to int except -1, which won't check for exceptions unless the function returns -1. Should be better, but it means adding return 0 to every apply function.

DragonSA:
Just a warning about this pull request: this patch (and others) on kivy 1.8.0 are causing a lot of failed unit tests (errors=2, failures=9).  This could be other issues with FreeBSD however I haven't isolated the problem (more difficult as FreeBSD ships with cython 0.22 - hence this patch).

DragonSA:
Never mind, the failures come from audio and font.  No issues with unit tests from graphics.  Apologies for the noise.

kived:
FYI - you can use pip to install a specific version of Cython.

tito:
Then +1 for adding return 0.

kived:
At this point, I'm leaning toward that option as well. The Cython devs haven't even acknowledged the bug report, so it may be better to take the slight performance hit to avoid waiting on them.

matham:
Yep, agree that the return 0/-1 is a good choice at this point. But not before the release.

matham:
We'll just have to make sure as well that exception previously eaten by cython will not be forwarded. E.g. textures to large etc, in the past those did not cause an exception because cython ignored them and just printed that.

inclement:
I did a PR to add 'except -1' instead, at https://github.com/kivy/kivy/pull/3283.

akshayaurora:
closing in favor of https://github.com/kivy/kivy/pull/3283

PR #3181, opened 2015-03-22
fix reported window size

### LINKED ISSUE #3177: [AnchorLayout][Regression] widgets are not positioned correctly until the window is resized
The bug is not present when using pygame, seen on Debian Jessie with Gnome 3.14.

screenshot from 2015-03-22 11 37 19

```python
Config.set('graphics', 'width', 300)
Config.set('graphics', 'height', 300)

from kivy.app import App
from kivy.lang import Builder

kv = """
AnchorLayout:
    Button:
        size_hint: None, None
"""


class TestApp(App):
    def build(self):

if __name__ == '__main__':
    TestApp().run()
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9865081-anchorlayout-regression-widgets-are-not-positioned-correctly-until-the-window-is-resized?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kived:
Fixes #3177

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -92,12 +92,11 @@ cdef class _WindowSDL2Storage:
         cdef SDL_DisplayMode mode
         SDL_GetWindowDisplayMode(self.win, &mode)
 
-        IF USE_IOS:
-            cdef int w, h 
-            SDL_GL_GetDrawableSize(self.win, &w, &h)       
-            mode.w = w
-            mode.h = h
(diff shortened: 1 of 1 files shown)


dessant:
superbe

PR #3185, opened 2015-03-24
Fix error in return types for clipboard xsel and dbusklipper

stuartmarsden:
Pull request [3171](https://github.com/kivy/kivy/pull/3171) was to fix the return for xsel and dbusklipper clipboards. It should return a list but using: 

`list('text/plain')`

actually returns ['t', 'e', 'x', 't', '/', 'p', 'l', 'a', 'i', 'n'] which is not what is expected. Better just to return:

`[u'text/plain']`


akshayaurora:
@stuartmarsden thanks, sry for the mess.

PR #3198, opened 2015-03-28
Original atlas textures are stored

CreamyCookie:
The textures (which contain the `textures`) of the original atlas images (the sprite sheets so to say) are now stored in `original_atlas_textures`.

### DIFF
--- a/kivy/atlas.py
+++ b/kivy/atlas.py
@@ -151,6 +151,13 @@ CoreImage = None
 class Atlas(EventDispatcher):
     '''Manage texture atlas. See module documentation for more information.
     '''
+    
+    '''List of original atlas textures (which contain the :attr:`textures`).
+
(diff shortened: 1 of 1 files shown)


kivy/atlas.py
```diff
@@ -201,12 +208,14 @@ def _load(self):
 
             # load the image
             ci = CoreImage(subfilename)
-
+            atlas_texture = ci.texture
+            original_atlas_textures.append(atlas_texture)
```
kived: Shouldn't this be **self.**original_atlas_textures?

kived:
What is the use case for this?

kivy/atlas.py
CreamyCookie: @kived Absolutely.

CreamyCookie:
@kived The use case is to apply a filter (for example `nearest`) to the whole Atlas.

PR #3203, opened 2015-03-30
Additional real world kivy language import examples

jtownley:
Goal: 
  Is to prevent people (like me) getting confused by the import syntax by adding 2 real world common examples to this process.

### DIFF
--- a/doc/sources/guide/lang.rst
+++ b/doc/sources/guide/lang.rst
@@ -80,7 +80,6 @@ To import something from python::
     #:import isdir os.path.isdir
     #:import np numpy
 
-
 Is equivalent to::
 
     from x.y import z as name
@@ -88,7 +87,6 @@ Is equivalent to::
     import numpy as np
 
 in python.
-
 
 To set a global value::
 
--- a/doc/sources/guide/lang.rst
+++ b/doc/sources/guide/lang.rst
@@ -84,6 +84,16 @@ Is equivalent to::
 
 in python.
 
+Additional examples::
+
+    #:import isdir os.path.isdir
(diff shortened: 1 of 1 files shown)


dessant:
Nice addition, although you could merge it with the original comparison, like this:

```
To import something from python::

    #:import name x.y.z
    #:import isdir os.path.isdir
    #:import np numpy

Is equivalent to::

    from x.y import z as name
    from os.path import isdir
    import numpy as np
```

jtownley:
I thought about that, but was worried about mixing theory with practice. Though I admit to being on the fence.

inclement:
import os.path.isdir

I don't think this is actually valid python, you can only import modules this way.

jtownley:
You are of course completely correct.

doc/sources/guide/lang.rst
```diff
@@ -77,13 +77,19 @@ There are two special syntax to define values for the whole Kv context:
 To import something from python::
 
     #:import name x.y.z
+    #:import isdir os.path.isdir
+    #:import np numpy
+
```
dessant: Please remove this newline and we can merge. ;)

doc/sources/guide/lang.rst
```diff
@@ -77,13 +77,19 @@ There are two special syntax to define values for the whole Kv context:
...
 
 Is equivalent to::
 
     from x.y import z as name
+    from os.path import isdir
+    import numpy as np
 
 in python.
 
+
```
dessant: same

dessant:
Thanks a bunch. :+1:

PR #3205, opened 2015-03-30
update rpi installation instructions

kived:


ivlevdenis:
Maybe need add comment about the recommendation of the maintainer of repository in the future not to use it because gstreamer packages already present in the mainly repository RPi.

PR #3209, opened 2015-04-01
core:Windows support for HIGHDPI on windows 8 plus

akshayaurora:


akshayaurora:
will re-open after redoing this.

PR #3214, opened 2015-04-02
Add KIVY_BINARY_DEPS and tests.

matham:
This adds KIVY_BINARY_DEPS env variable, which if present will add the contents of that path to kivy. This can be used when building e.g. a wheel where we need to add dlls (sdl, gstreamer, etc.) to the dist. One can then set the variable with the path containing the dlls.


setup.py
```diff
@@ -68,6 +69,18 @@ def pkgconfig(*packages, **kw):
     return kw
 
 
+def copy_files(src, dst):
```
tito: What about shutil.copytree ?

setup.py
```diff
@@ -259,6 +273,11 @@ def build_extensions(self):
             for e in self.extensions:
                 e.extra_link_args += ['-lm']
 
+        if kivy_binary_deps is not None and isdir(kivy_binary_deps):
```
dessant: `if kivy_binary_deps and isdir(kivy_binary_deps)`? 👅

setup.py
matham: @tito the reason is that copytree insists on the target directories must not exist, but here it definitely exists. There are similar issues if the subdirectories already exist. the copy_files method doesn't mind existing directories.

matham:
Ok, added to use setuptools only if env variable is set.

dessant:
We could document this var and possibly `KIVY_USE_SETUPTOOLS` in http://kivy.org/docs/guide/environment.html. Thanks for your packaging work (✿◠‿◠).

PR #3239, opened 2015-04-05
sdl2:Window: Introduce keycodes swaps for numpad keys

akshayaurora:
This should be merged into both master and stable.


tito:
+1

Issue #2831, opened 2015-01-07
Camera not working on android [$15]


nesergen:
When I start my programm on TabDevice the Camera widget shows the white screen. On DescktopComputer the Camera widget works perfect. 

My code:
```
class MyApp(App):
	
	def build(self):
		parent = Widget()

		layout = GridLayout(cols=1, rows =3, spacing=2, size=(800,800), halign="left")
		parent.add_widget(layout)

		cam = Camera(play= True)

		layout.add_widget(cam)

		return parent

if __name__ == '__main__':
	MyApp().run()
```

ErrorLog:
```
[INFO              ] Logger: Record log in /storage/emulated/0/kivy/experim/.kivy/logs/kivy_15-01-08_0.txt
[INFO              ] Kivy v1.8.0
[INFO              ] Factory: 157 symbols loaded
[DEBUG             ] Cache: register <kv.lang> with limit=None, timeout=Nones
[DEBUG             ] Cache: register <kv.image> with limit=None, timeout=60s
[DEBUG             ] Cache: register <kv.atlas> with limit=None, timeout=Nones
[WARNING           ] stderr: /data/data/org.kivy.pygame/files/lib/python2.7/site-packages/kivy/core/image/img_pygame.py:12: RuntimeWarning: import cdrom: No module named cdrom
[WARNING           ] stderr: (ImportError: No module named cdrom)
[INFO              ] Image: Providers: img_tex, img_dds, img_pygame, img_pil, img_gif 
[DEBUG             ] Cache: register <kv.texture> with limit=1000, timeout=60s
[DEBUG             ] Cache: register <kv.shader> with limit=1000, timeout=3600s
[INFO              ] Text: Provider: pygame
[DEBUG             ] Cache: register <kv.loader> with limit=500, timeout=60s
[INFO              ] Loader: using a thread pool of 2 workers
[DEBUG             ] Cache: register <textinput.label> with limit=None, timeout=60.0s
[DEBUG             ] Cache: register <textinput.width> with limit=None, timeout=60.0s
[DEBUG             ] Camera: Ignored <pygst> (import error)
[DEBUG             ] Camera: Ignored <opencv> (import error)
[CRITICAL          ] Camera: Unable to find any valuable Camera provider at all!
[DEBUG             ] App: Loading kv <./my.kv>
[DEBUG             ] App: kv <./my.kv> not found
[DEBUG             ] Window: Ignored <egl_rpi> (import error)
[...]

PR #3236, opened 2015-04-05
Android camera provider

jasonxunxu:
This is a fully working camera provider for Android. It is efficient and can achieve pretty high frame rate by receiving frames through a `SurfaceTexture` (based on [this snippet](https://groups.google.com/forum/#!msg/kivy-users/hiyAJqjhFjo/3-WLF4MJGpwJ)).

In addition, it provides an interface to obtain current frame data as a Numpy array (requiring OpenCV Python binding `cv2`), making it suitable for computer vision apps.

It fixes issue #2831.

PR #3250, opened 2015-04-07
Android camera provider

jasonxunxu:
This is a fully working camera provider for Android. It is efficient and can achieve pretty high frame rate by receiving frames through a `SurfaceTexture` (based on [this snippet](https://groups.google.com/forum/#!msg/kivy-users/hiyAJqjhFjo/3-WLF4MJGpwJ)).

In addition, it provides an interface to obtain current frame data as a Numpy array (requiring OpenCV Python binding `cv2`), making it suitable for computer vision apps.

It fixes issue #2831. 

This is a recreated PR after rebasing #3236.

### DIFF
--- a/kivy/core/camera/camera_android.py
+++ b/kivy/core/camera/camera_android.py
@@ -1,0 +1,83 @@
+from jnius import autoclass, PythonJavaClass, java_method
+from kivy.clock import Clock
+from kivy.graphics.texture import Texture
+from kivy.core.camera import CameraBase
+
+
+Camera = autoclass('android.hardware.Camera')
+SurfaceTexture = autoclass('android.graphics.SurfaceTexture')
+ImageFormat = autoclass('android.graphics.ImageFormat')
+
+
+class PreviewCallback(PythonJavaClass):
+    """
+    Interface used to get back the preview frame of the Android Camera
+    """
+    __javainterfaces__ = ('android.hardware.Camera$PreviewCallback', )
+
+    def __init__(self, callback):
+        super(PreviewCallback, self).__init__()
+        self.callback = callback
+
+    @java_method('([BLandroid/hardware/Camera;)V')
+    def onPreviewFrame(self, data, camera):
+        self.callback(data, camera)
+
+
+class CameraAndroid(CameraBase):
+    """
+    Implementation of CameraBase using Android API
+    """
+
+    def __init__(self, **kwargs):
+        self._android_camera = None
+        self._surface_texture = SurfaceTexture(-1)
+        self._preview_cb = PreviewCallback(self._on_preview_frame)
+        super(CameraAndroid, self).__init__(**kwargs)
+
+    def init_camera(self):
+        self._android_camera = Camera.open(self._index)
+        params = self._android_camera.getParameters()
+        width, height = self._resolution
+        params.setPreviewSize(width, height)
+        self._android_camera.setParameters(params)
+        #self._android_camera.setDisplayOrientation()
+
+        pf = params.getPreviewFormat()
+        assert(pf == ImageFormat.NV21)  # default format is NV21
+        bpp = ImageFormat.getBitsPerPixel(pf) / 8.
+        for k in range(2):  # double buffer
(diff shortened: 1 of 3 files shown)


kived:
I don't think you actually rebased this; GitHub says it cannot be automatically merged. Make sure you have pulled the latest changes from Kivy, not just your fork, and then rebase this branch off kivy/master - not the master of your fork. Then do `git push --force` to update the PR.

For convenience, I generally take the approach of making my fork's master branch mirror kivy/master. In my repo I have an 'upstream' remote, so from my master branch I just do `git fetch upstream` then `git reset --hard upstream/master`.

kived:
Thinking about it, your last branch was to merge into the 1.8.0 branch, which means this is probably based on 1.8.0 and not master, and there are thousands of commits in between. It may actually end up being easier to manually redo these changes to a copy of master, rather than trying to rebase.

jasonxunxu:
I actually rebased onto tag 1.8.0 which is on the master branch. I started the development before 1.9.0 was released so just wanted to start with a stable version then. The branch was rebased to tag 1.8.0 instead of current master tip in hope of 1.8.x based forks/branches being able to benefit from it as well.

I tested merging with current master, the only conflict is in `kivy/__init__.py` (where `kivy_options` specifies providers), and quite straightforward to resolve.

dessant:
Avoid basing your prs on 1.9.0 in the future, always use master, sometimes even with master we need to rebase or create a new pr, when conflicting stuff is merged before our code. Great pr btw, this feature will make a lot of users happy. ;)

jasonxunxu:
Got it, thanks for the clarification!

tshirtman:
Actually that wasn't hard to rebase, just had an expected conflict in __init__ (kivy_options), but i think i merged it ok. will do a PR from rebase_3250

jasonxunxu:
Since rebasing is so wanted I will rebase this PR once again onto master. Also I am making some improvements per @tito's [suggestion](https://github.com/kivy/kivy/pull/3236#discussion_r27855620) which will be committed shortly.

jasonxunxu:
Rebased to master, also added a new commit for rendering efficiency improvement as @tito suggested. It seems that merging this PR (instead of  #3255) is easier (thank you @tshirtman for helping though).

inclement:
I guess the reason this hasn't been merged is nobody has tested it? @thirtman, @dessant?

I'll try to look at it this weekend if that's the case. Sorry @jasonxunxu for the delay

jasonxunxu:
Wow, thank you @inclement for lending a hand!

kivy/core/camera/camera_android.py
```diff
@@ -0,0 +1,156 @@
...
+                return None
+            buf = self._buffer.tostring()
+            return buf
+
+    def decode_frame(self, buf):
+        """
+        Decode image data from grabbed frame
+        """
+        import numpy as np
+        from cv2 import cvtColor
```
akshayaurora: This makes numpy and cv2 as a essential requirement on android. Either we should look at replacing these or making bumpy and cv2 as a base requirement for `kivy` recipe in py4a. I assume these are used for performance reasons.

Would it be possible to replace these with another implementation that is equally fast but does not require loading these libs?

jasonxunxu:
It should be clarified that the normal operation of this provider (i.e. showing live stream from the camera) does **not** require OpenCV or Numpy. Those dependencies are only needed when you use the `read_frame` interface to fetch frame data (as a Numpy array) for some processing, in which case you most likely already have cv2 and numpy included in the project.

It would be nice if someone can contribute a fast and dependency-free method to do the needed format conversion, but again using `CameraAndroid` just as a regular camera provider does not make numpy and cv2 essential requirements on Android.

tshirtman:
Thanks for the clarification, i do think this dependency should be documented in the method's docstring.

jasonxunxu:
@tshirtman The dependencies are now documented. A few enhancements were added as well.

akshayaurora:
@jasonxunxu nice work! Really awesome contribution, thanks a ton.

PR #3257, opened 2015-04-09
Release TouchInput handles in wm_touch

azixMcAze:
TouchInput handles are not released after being used in wm_touch.
This will result in touches being completely ignored after some times (after ~1900 touches, in my tests) and the application no longer responding if a touch screen is the only input device


akshayaurora:
@azixMcAze thanks, nice find.

Issue #3256, opened 2015-04-09
RstDocument don't allow the right space for images


Farkal:
If you try to add images to RstDocument they will come on the text and not after.
On the realese **1.8.0** it was great, i have do some research but i didn't find what changed
```python
BoxLayout:
    RstDocument:
        text: "==================\nTitle\n==================\n\n\n\n.. image:: myimage.png\n\n"
        scroll_y: 0
```

PR #3258, opened 2015-04-09
Rigth space for images

Farkal:
Refers to my issue #3256

This correct the problem.


kived:
Unfortunately, this is not the right fix for the issue. The real issue is within the rst widgets. This change would affect many other uses of `AnchorLayout` than just this issue.

The root issue here is that the `AnchorLayout` created in the rst widgets was set to an initial height of 1. A binding was created that would update the height of the `AnchorLayout` if the image height changed, which made this work for remote (async) images, but for local images the height would already be set and so the binding would never fire.

PR #3259, opened 2015-04-09
Overhaul ScreenManager.switch_to

rnixx:
This is the replacement for #3133.


rnixx:
according to IRC discussion, switch_to behavior will not change. the proposed function in #3133 should go to a behavior mixin class.

rnixx:
The related code can be found at https://gist.github.com/rnixx/2aab5e13e17c1582b6dd

PR #3275, opened 2015-04-14
Add option for slider to be only sensitive on cursor

rdoursenaud:
This is useful for usecases where great value changes at once are
not desirable like sound and light control for example.

### DIFF
--- a/kivy/uix/slider.py
+++ b/kivy/uix/slider.py
@@ -24,6 +24,7 @@ from kivy.uix.widget import Widget
 from kivy.properties import (NumericProperty, AliasProperty, OptionProperty,
                              ReferenceListProperty, BoundedNumericProperty)
 from kivy.metrics import sp
+from kivy.graphics.vertex_instructions import Rectangle
 
 
 class Slider(Widget):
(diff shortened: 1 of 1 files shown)


rdoursenaud:
First time contributing to kivy and not yet familiar with the codebase.
Review and comments appreciated.

kivy/uix/slider.py
```diff
@@ -206,6 +213,17 @@ def on_touch_down(self, touch):
                     self.value = max(
                         self.min,
                         self.value - (self.max - self.min) / 20)
+        if self.sensitive == 'cursor':
+            for child in self.canvas.children:
+                if isinstance(child, Rectangle):
```
tito: My feeling are mixed here. That's the first time that we assume a certain graphical representation in order to act on the widget. And i think it's wrong.

The widget should expose the sensitive area/size, and the graphical representation (kv) must use it, or change it. But not the inverse, otherwise nobody can freely rewrite a kv rule without redoing the python code as well.

rdoursenaud: Thanks for taking the time to have a look at my PR!
I absolutely agree with you and I knew that this hack may raise some eyebrows but I didn't find any simple way of inferring the cursor area.
Is there already a reference implementation of widgets exposing their sensitive area(s) or should I get back to work and cook my own?

dessant:
The PR seems to be abandoned, please reopen if necessary.

rdoursenaud:
Hi @dessant
It's not abandoned, I'm trying to figure out how to achieve the desired behavior without resorting to that hack. But for now, I really don't know where to start…
If your have *any* insight, please share it with me.
Thanks.

dessant:
@rdoursenaud, the handle could be turned into a Widget so you can use the collision detection api for it and not rely on the graphical representation.

dessant:
The feature will be merged in https://github.com/kivy/kivy/pull/3275.

PR #3282, opened 2015-04-16
Fixes iOS density. 

tito:
This was a big mistake, leading to smaller element than the actual screen density, weird aliasing and more. Dunno why nobody reported it before (not even me.)

I didn't understand why my app on the ipad 3 (retina) was having the whole UI smaller than the equivalent on desktop. This was due to the * 0.75. The previous comment i put was a mistake, like my analysis. This must go away.


akshayaurora:
:dancer:

PR #3283, opened 2015-04-18
Changed 'except *' to 'except -1' for cython 0.22

inclement:
This is just a first simple commit, but it allows compilation and everything seems to work.

Not sure what other stuff is needed...I'm not sure how this will change exception propagation (if at all) where cython previously logged but ignored them.


kived:
:+1: for merge in 1.9.2

akshayaurora:
:+1:

PR #3285, opened 2015-04-20
Fix AttributeError in py3

LogicalDash:
The ``__repr__`` method should always return a string. It was returning a bytes object instead.

In py2 this made no difference. In py3 it USUALLY made no difference, but if you were using the inspector, and tried to inspect an object that contained a WeakProxy, the whole app would crash.


kived:
:+1:

Issue #3293, opened 2015-04-24
[Docs] Git Screencasts site referenced in Contributing Guide no longer available


kevgathuku:
The [Git Screencasts](http://gitcasts.com/) site referenced in the [Contributing guideline](https://github.com/kivy/kivy/blob/master/doc/sources/contribute.rst#L93) doesn't seem to be available anymore.
Maybe the link should now point to the videos on the official git website available at [Git - Videos](https://git-scm.com/videos)

akshayaurora:
closed via https://github.com/kevgathuku/kivy/commit/6688d7cf81250cf14079a89d12856627b54292b5

PR #3294, opened 2015-04-24
[Docs] Resove issue #3293. Update link to Git Screencasts

kevgathuku:
Update link from the now unavailable gitcasts.com to the videos on the official Git website.


kevgathuku:
:+1:

PR #3299, opened 2015-04-25
Variable id replaced with _id

AndersHogqvist:
This one major, can't use osc since id is used as a variable name.


tshirtman:
Good! But could be even better by giving a more explicit variable name, here i think `thread_id` or `osc_id` would be better, `id` was a bad name for more than one reason :).

akshayaurora:
@AndersHogqvist thanks!

PR #3303, opened 2015-04-26
Fix pyinstaller docs and hook for sdl2

### LINKED ISSUE #3272: Can't create package for windows with kivy 1.9 portable
I'm looking to port an existing kivy 1.8 project to kivy 1.9. I've just downloaded the portable version and have the application working.

However when packaging the app using pyinstaller and the instructions on http://kivy.org/docs/guide/packaging-windows.html the app packages, but on execution immediately fails with error:
```
  File "<string>", line 34, in <module>
ImportError: No module named pygame.pkgdata
```

I've tried using my old .spec file and generating a new one with exactly the same results.

I'm a bit mystified where this is coming from as pygame isn't imported anywhere in my application and I thought it had been replaced with sdl2 in kivy 1.9. I'm also confused that the application works when run directly.

Anyone come across this issue or can point me in the right direction?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11048742-can-t-create-package-for-windows-with-kivy-1-9-portable?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

matham:
Fixes #3272.

### DIFF
--- a/kivy/tools/packaging/pyinstaller_hooks/rt-hook-kivy.py
+++ b/kivy/tools/packaging/pyinstaller_hooks/rt-hook-kivy.py
@@ -29,15 +29,17 @@ environ['KIVY_EXTS_DIR'] = join(root, 'extensions')
 environ['KIVY_MODULES_DIR'] = join(root, 'modules')
 environ['KIVY_EMBED'] = '1'
 
+
 # Monkey-patch pygame to get around an issue with Pygame window icon and
 # PyInstaller 2.1. See kivy issue #1638
-# Uncomment the following to package pygame
-#import pygame.pkgdata
-#_original_getResource = pygame.pkgdata.getResource
-#
-#
-#def getResource(identifier, *args, **kwargs):
-#    if identifier == 'pygame_icon.tiff':
-#        raise IOError()
-#    return _original_getResource(identifier, *args, **kwargs)
-#pygame.pkgdata.getResource = getResource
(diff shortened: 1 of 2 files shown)


kivy/tools/packaging/pyinstaller_hooks/rt-hook-kivy.py
```diff
@@ -29,15 +29,17 @@
...
-#import pygame.pkgdata
-#_original_getResource = pygame.pkgdata.getResource
-#
-#
-#def getResource(identifier, *args, **kwargs):
-#    if identifier == 'pygame_icon.tiff':
-#        raise IOError()
-#    return _original_getResource(identifier, *args, **kwargs)
-#pygame.pkgdata.getResource = getResource
+def getResource(identifier, *args, **kwargs):
```
akshayaurora: @matham we don't provide pygame anymore, there is not reason to look to patch pygame at runtime and this does lead to a crash for a packaged app using just SDL2.

matham: How does this load to a crash at runtime? That's why it's under a
try/except block.

PR #3325, opened 2015-05-03
improve actionbar layout

kived:
Adds a `pack_width` alias property to `ActionItem`. This property is equal to the lesser of `minimum_width` and `width`. Previously, `ActionView` was assuming items would use more space than they actually do, because `minimum_width` was greater than the actual `width` of the item, leading to premature hiding of items.

Also modifies `ActionPrevious` so that the `Label` is not part of the `Button` area.

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -589,49 +589,68 @@
     temp_width: 0
     temp_height: 0
 
+<ActionPreviousButton@Button>:
+    background_normal: 'atlas://data/images/defaulttheme/action_item'
+    background_down: 'atlas://data/images/defaulttheme/action_item_down'
+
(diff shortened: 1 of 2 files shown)


tito:
@kived +1. You can rebase and merge the changes yourself :)

NumesSanguis:
I know get the following error when using ActionLabel (<ActionLabel@ActionItem+Label>) when I updated to the latest Kivy master branch:

[CRITICAL] [Clock       ] Warning, too much iteration done before the next frame. Check your code, or increase the Clock.max_iteration attribute

Might it be related to this update?

Here is a simple example generating the error:
http://pastebin.com/YAHKVJdr

It has to do with calling:
text: app.foo

PR #3330, opened 2015-05-03
add xclip clipboard/cutbuffer implementation

kived:
Also enables SDL2 clipboard on Linux again, even though it is sometimes buggy it will work sometimes, and hopefully be used rarely due to all these other implementations. :)


akshayaurora:
:+1:

PR #3374, opened 2015-05-18
Screenshots added

picibucor:
Screenshots of examples are added. Naming is following the already existing pattern (for Gallery of Examples).


tito:
Why all the images are not referenced into the documentation?

dessant:
@picibucor, the other thing to consider, which i've mentioned before, is that window decorations get outdated fast, and besides that, kivy being cross-platform, there is no representative window decoration to use.

I would be in favor of removing this commit from the git history of your branch and adding images created with kivy's screenshot method. Window/image sizes reduced to half would work much better, because they are simple examples with very few details, and it would also save bandwidth on both ends.

PR #3377, opened 2015-05-20
fix webdebugger

kived:


Zen-CODE:
Thanks. fixed issue on my side

PR #3093, opened 2015-03-03
Adding an undefined state to progress bar

Vagdish:
by passing "None" as a value.

PR #3378, opened 2015-05-22
Adding an indeterminate mode to ProgressBar

Vagdish:
By passing None as value or as value_normalized.
This commit simplify the pull request 3093 (no new method added) but the
indeterminate mode is still hackable through style.kv

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -63,8 +63,6 @@
             size: (sp(32), sp(32))
             source: 'atlas://data/images/defaulttheme/slider_cursor{}'.format('_disabled' if self.disabled else '')
 
-#:set ProgressBar_relsize 0.2
-#:set ProgressBar_period 2.0
 <ProgressBar>:
     canvas:
         Color:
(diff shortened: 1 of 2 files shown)


Vagdish:
I'm very sorry to open a new pull request for this change. I've done some mistakes with my git repo.

Well, this is similar to #3093 but without adding new methods to API. Devs can still easily change the visuals through style.kv

tito:
Thanks a lot for your contribution! I'm starting to be worried about the complexity of the kv rule. Would you be able to try the separation within python instead? Ie, pos/size/border are depending of a pos/size/border_indeterminate if the current mode is indeterminate, otherwise use the pos/size/border_normal
(ie, you'll need to add 6 variables, and switch from one to another depending the mode).

Unsure exactly about the code in the end, but this is getting unreadable :)

Vagdish:
Yes, it could be done within python. I'll update the pull request according to your advice and put the sin-math-magic into python.

KeyWeeUsr:
@Vagdish Any progress with this?

PR #3380, opened 2015-05-24
cache keyboard height and clear cache on global layout events

denys-duchier:
The get_keyboard_height storm on touch events is due to a change in post_dispatch_input (base.py): it looks up the property keyboard_height.  However this property is not cached, so get_keyboard_height is invoked every time.  This PR adds a cache for the keyboard height, and clears the cache on global layout events.


akshayaurora:
As discussed on irc this is no longer reproducible with master, the caching is now done in p4a and calling android.get_keyboard_height() now simply returns the chached value from PythonActivity.mView.kHeight. http://git.io/vTN4t

PR #3383, opened 2015-05-25
use internal caching for keyboard_height property

denys-duchier:
keyboard_height is an AliasProperty, and it has an optional cache.  this patch activates it.  I verified that qua-non's latest fix to p4a actually causes the cache to be updated.


akshayaurora:
@denys-duchier thanks

PR #3387, opened 2015-05-27
take into account the minimum_widths of all items of a dropdown

denys-duchier:
This fix was suggested by @kived on IRC.

### DIFF
--- a/kivy/uix/actionbar.py
+++ b/kivy/uix/actionbar.py
@@ -307,7 +307,7 @@ class ActionGroup(ActionItem, Spinner):
         children = ddn.container.children
 
         if children:
-            ddn.width = max([self.width, children[0].minimum_width])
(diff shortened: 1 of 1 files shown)


kivy/uix/actionbar.py
```diff
@@ -307,7 +307,7 @@ def _toggle_dropdown(self, *largs):
         children = ddn.container.children
 
         if children:
-            ddn.width = max([self.width, children[0].minimum_width])
+            ddn.width = max(self.width, max(c.minimum_width for c in children))
```
akshayaurora: use a copy? for c in children[:]

denys-duchier: In other circumstances I would, but in this case: nothing can be scheduled during iteration, so children cannot be modified (it is not legal to do it from a different thread), and we may as well not incur the overhead of a copy. or something I missed?

kived:
:+1:

PR #3389, opened 2015-05-27
by default, ActionButton should use texture_size[0] as minimum_width

denys-duchier:
With this change, ActionGroup's dropdown will automatically be sized
as one would naively expect.

I also added a bit of horizontal padding (8dp is what is stipulated
in the material design spec).

what do you think?

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -569,6 +569,8 @@
     size_hint_x: None if not root.inside_group else 1
     width: [dp(48) if (root.icon and not root.inside_group) else max(dp(48), (self.texture_size[0] + dp(32))), self.size_hint_x][0]
     color: self.color[:3] + [0 if (root.icon and not root.inside_group) else 1]
+    padding_x: dp(8)
+    minimum_width: self.texture_size[0]
 
     Image:
         allow_stretch: True


kivy/data/style.kv
```diff
@@ -569,6 +569,8 @@
     size_hint_x: None if not root.inside_group else 1
     width: [dp(48) if (root.icon and not root.inside_group) else max(dp(48), (self.texture_size[0] + dp(32))), self.size_hint_x][0]
     color: self.color[:3] + [0 if (root.icon and not root.inside_group) else 1]
+    padding_x: dp(8)
+    minimum_width: self.texture_size[0]
```
akshayaurora: This is a binding, you need to be careful as if a person wants to set minimum_width after this they wouldn't be able to and whenever the texture changes the minimum_width would be reset to this.

denys-duchier: Ah yes!  Well... they could using `minimum_width: self.minimum_width and dp(50)`, but I take your meaning.  This is a shortcoming of the current implementation of property constraints.  Maybe it can be fixed.  I'll look into it.

akshayaurora:
@denys-duchier please fix the issue mentioned above and re-open the pr.

PR #3392, opened 2015-05-28
allow NumericProperties to accept type double

sirpercival:
There's no reason not to, is there?

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -528,7 +528,7 @@ cdef class NumericProperty(Property):
         if x is None:
             return x
         tp = type(x)
-        if tp is int or tp is float or tp is long:
+        if tp is int or tp is float or tp is long or tp is double:
             return x
         if tp is tuple or tp is list:
             if len(x) != 2:


inclement:
Python's numeric types are int, float, long and complex - floats are probably implemented with doubles internally. Other than that, double isn't a defined variable and (unless I miss something) this code won't compile.

I've never thought about it but maybe we should accept complex numbers...though the lack of ordering etc. might break things, it's not necessarily a natural extension.

sirpercival:
aha - i was using double in cython and i forgot it wasn't a native python type. the actual type ends up being numpy.float64. what about accepting numpy numeric types?

inclement:
It would be nice to accept numpy's numeric types, but we can't rely on the numpy import, and even then it's kind of arbitrary to stop there. We could try to convert anyting unrecognised with float() or similar, but then some things won't end up being what people expect. Maybe this should be considered a doc issue, as I remember the error about it is a bit unclear.

sirpercival:
what about having a compile option to allow numpy types? in my case it seems so far to be ok to convert everything to float, but i'm wondering if there's a situation where it would be problematic.

matham:
I don't think we should make special exceptions for numpy because it's not a python native type. On the other hand, I fail to see why we check if it's a numeric type. Instead we should just check if it's a list/string and do the conversion, otherwise just store the original value. That's how duck typing works anyway. So I'd be in favor of removing the is int etc check.

Issue #2365, opened 2014-07-24
KEP003: Syntax to clear previous kv rules of a property


matham:
Many times one creates e.g. a dynamic class with some properties and their rules which is than used repeatedly in many places. E.g. a simple button with a label and a icon.

Although most of the class instances conform to the rules defined in the dynamic class, sometimes, you want to overwrite the rules for a specific property just in one or two places. This can be hard to impossible  to do currently.

One example:

```
<SwitchIcon@ToggleButtonBehavior+Widget>:
    disabled: app.running
    color: app.that_color
```

Then later in one instance we want to do:

```
SwitchIcon:
    disabled: not app.running
    color: (1, 1, 1, 1)
```

The `disabled` property will now have two rules, with the `disabled: not app.running` executed _after_ the `disabled: app.running` rule, which is why most people don't realize that the second rule definition doesn't overwrite, but just adds a second rule to the disabled property.

For the `color` property, this just won't work. Yes, this `color` will be set to `(1, 1, 1, 1)`, but as soon as `app.that_color` changes, because `color` is still bound to that rule, `color: (1, 1, 1, 1)` will be overwritten.

A solution is to allow clearing the rules for a specific property with a dash, e.g.

```
SwitchIcon:
    -disabled: not app.running
    -color: (1, 1, 1, 1)
```

PR #3407, opened 2015-06-07
ensure that only the last property constraint takes effect

denys-duchier:
Using the kv language, the same property of a widget can be constrained multiple times.  When all constraints are static, then only the last one takes effect:

```
from kivy.app import App
from kivy.lang import Builder

root = Builder.load_string("""
BoxLayout:
    Label:
        text: "Hello World"

<Label>:
    color: (0,1,0,1)

<Label>:
    color: (0,0,1,1)

<Label>:
    color: (1,0,0,1)
""")


class MyApp(App):

    def build(self):
        return root

MyApp().run()
```

When they are all dynamic, then they are all evaluated, but only the last one wins:

```
from kivy.app import App
from kivy.lang import Builder, global_idmap

def show(x):
    print(x)

global_idmap["show"] = show

root = Builder.load_string("""
<MyLabel@Label>:
    tick: True

BoxLayout:
    MyLabel:
        id: label
        text: "Hello World"
    Button:
        text: "Click Me!"
        on_release:
            label.tick = not label.tick

<MyLabel>:
    color: (show('Green'),(0,1,0,1) if self.tick else (0,0.5,0,1))[1]

<MyLabel>:
    color: (show('Blue'),(0,0,1,1) if self.tick else (0,0,0.5,1))[1]

<MyLabel>:
    color: (show('Red'),(1,0,0,1) if self.tick else (0.5,0,0,1))[1]
""")


class MyApp(App):

    def build(self):
        return root

MyApp().run()
```
When there is a mix and the last one is static, then it wins at first, but when dynamic rules need to be reevaluated, then the last dynamic rule wins:

```
from kivy.app import App
from kivy.lang import Builder, global_idmap

def show(x):
    print(x)

global_idmap["show"] = show

root = Builder.load_string("""
<MyLabel@Label>:
    tick: True

BoxLayout:
    MyLabel:
        id: label
        text: "Hello World"
    Button:
        text: "Click Me!"
        on_release:
            label.tick = not label.tick

<MyLabel>:
    color: (show('Green'),(0,1,0,1) if self.tick else (0,0.5,0,1))[1]

<MyLabel>:
    color: (show('Blue'),(0,0,1,1) if self.tick else (0,0,0.5,1))[1]

<MyLabel>:
    color: (show('Red'),(1,0,0,1) if self.tick else (0.5,0,0,1))[1]

<MyLabel>:
    color: (1,1,0,1)
""")


class MyApp(App):

    def build(self):
        return root

MyApp().run()
```
This PR proposes that only the last applied constraint to a widget's property remains in effect.  If there was a previously applied constraint, then it is removed (all its bindings are removed).

This makes kv much more predictable, and makes it possible for kivy to provide default constraints that are generally useful, while still allowing the user to override these constraints.

### DIFF
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -890,14 +890,13 @@ cdef class EventObservers:
         cdef BoundCallback new_callback
 
         while callback is not None:
-            if is_ref and not callback.is_ref:
-                cb_equal = callback.func == observer()
-            elif callback.is_ref and not is_ref:
-                cb_equal = callback.func() == observer
-            else:
-                cb_equal = callback.func == observer
-            if (callback.lock != deleted and callback.largs is None and
-                callback.kwargs is None and cb_equal):
+            cb_equal = ((callback.func() if callback.is_ref else callback.func)
+                        ==
+                        (observer() if is_ref else observer))
+            if (cb_equal
+                and callback.lock != deleted
+                and callback.largs is None
+                and callback.kwargs is None):
                 return
             callback = callback.next
 
--- a/kivy/lang.py
+++ b/kivy/lang.py
@@ -1496,6 +1496,21 @@ def delayed_call_fn(args, instance, v):
         _delayed_start = args
 
 
+def undo_bindings(bound):
+    for f, k, uid in bound:
+        if fun is None:
+            continue
+        try:
+            f.unbind_uid(k, uid)
+        except ReferenceError:
+            pass
+
+
+def undo_bindings_list(bounds):
+    for bound in bounds:
+        undo_bindings(bound)
+
+
 def update_intermediates(base, keys, bound, s, fn, args, instance, value):
     ''' Function that is called when an intermediate property is updated
     and `rebind` of that property is True. In that case, we unbind
@@ -1534,13 +1549,7 @@ def update_intermediates(base, keys, bound, s, fn, args, instance, value):
             The function to be called args, `args` on bound callback.
     '''
     # first remove all the old bound functions from `s` and down.
-    for f, k, fun, uid in bound[s:]:
-        if fun is None:
-            continue
-        try:
-            f.unbind_uid(k, uid)
-        except ReferenceError:
-            pass
+    undo_bindings(bound[s:])
     del bound[s:]
 
     # find the first attr from which we need to start rebinding.
@@ -1598,14 +1607,7 @@ def create_handler(iself, element, key, value, rule, idmap, delayed=False):
     # posting a new one.  there should be at most one constraint for any
     # property iself/key.
     if current_constraint_bounds:
-        for bound in current_constraint_bounds:
-            for f, k, fn, bound_uid in bound:
-                if fn is None:
-                    continue
-                try:
-                    f.unbind_uid(k, bound_uid)
-                except ReferenceError:
-                    pass
+        undo_bindings_list(current_constraint_bounds)
         del current_constraint_bounds[:]
     # at this point current_constraint_bounds is an empty list
 
--- a/kivy/lang.py
+++ b/kivy/lang.py
@@ -1614,7 +1614,8 @@ def create_handler(iself, element, key, value, rule, idmap, delayed=False):
     # if there is already a constraint on iself/key then remove it before
     # posting a new one.  there should be at most one constraint for any
     # property iself/key.
-    current_constraint_bindings = clear_constraint(iself, key)
+    if not delayed:
+        current_constraint_bindings = clear_constraint(iself, key)
 
     # we need a hash for when delayed, so we don't execute duplicate canvas
     # callbacks from the same handler during a sync op
@@ -1672,7 +1673,7 @@ def create_handler(iself, element, key, value, rule, idmap, delayed=False):
                     was_bound = True
             if was_bound:
                 handler_append(bound)
-            if bound:
+            if bound and not delayed:
                 current_constraint_bindings.append(bound)
 
     try:
--- a/kivy/lang.py
+++ b/kivy/lang.py
@@ -831,6 +831,11 @@ else:
 # widget is deleted
 _handlers = defaultdict(list)
 
+# uid -> propname -> [bound]
+# used to correctly unbind callbacks when one rule's property constraint
+# overrides another
+_constraints = defaultdict(lambda:defaultdict(list))
+
 
 class ProxyApp(object):
     # proxy app object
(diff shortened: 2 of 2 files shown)


tito:
I'm all in for that. @matham could you check it please ?

matham:
We talked about this on IRC and I had two main issues with it.

* 1 that this does what https://github.com/kivy/kivy/issues/2365 proposes, however it changes the default and doesn't make it optional. In particular, I think that by default all the rules should be applied, like it's layed out in #2365, and only if a user specifies shall the older rules be discarded.
* 2 is that I thought it should be done in parser instead. I changed my mind since, though about this.
* I also think that the way it's done in the PR is somewhat inefficient, which is important considering that this code is executed many times.

@denys-duchier said he'd look into it. However, I'd volunteer to implement this and KEP003 (#2365), but only if you don't mind me doing it. I can see how to make this more efficient, more general, and also allow it to work with the upcoming compiler more easily, especially since I have been spending some time on lang.py

denys-duchier:
I wanted to take some time thinking things over, and now I am fairly sure that I disagree with matham.  The default should predictable and declarative (declarativity is what kv lang is primarily about).  Consider the following code:

```
from kivy.lang import Builder

root = Builder.load_string("""
<Tasty@Widget>:
    tasty: True
    color: (1,0,0,1) if self.tasty else (1,1,0,1)

<Smelly@Widget>:
    smelly: True
    color: (0,1,0,1) if self.smelly else (0,0,1,1)

<TastyButton@Tasty+Button>:
<SmellyButton@Smelly+Button>:
<TastySmellyLabel@Tasty+Smelly+Label>:

BoxLayout:
    orientation: "vertical"
    tasty: True
    smelly: True
    TastyButton:
        text: "Tasty"
        tasty: root.tasty
        on_release:
            root.tasty = not root.tasty
    SmellyButton:
        text: "Smelly"
        smelly: root.smelly
        on_release:
            root.smelly = not root.smelly
    TastySmellyLabel:
        text: "Tasty+Smelly"
        tasty: root.tasty
        smelly: root.smelly
""")

runTouchApp(root)
```

`TastyButton` and `SmellyButton` are declarative, but the constraints on `TastySmellyLabel` are not: it is not sufficient to know the values of properties `tasty` and `smelly` to deduce the color of the label; you have to know something about past history.  However, I can also see value in making my proposed default optional, and I'll look into it.  In fact, it might be useful to put priorities on individual constraints.

Doing this in the parser is problematic, I think, e.g. in presence of multiple inheritance unknown to the parser.

Regarding efficiency, I followed what was already being done with the global `_handlers` variable.  I don't think I am introducing inefficiency in the building process that wasn't already there; am I?  if so, I'll fix that.

dessant:
@denys-duchier I agree that we should experiment with this as a non-default behavior, as we discussed. Making it the default behavior can't be done, because it would break current apps.

matham:
I'm sorry, but I cannot agree. I get that you cannot predict the value of `color`, but I don't really see the problem with that, not enough to consider changing and disabling how it works currently and breaking app, by default. I think if you do need that predictability and don't want multiple rules, you should have specify that like in KEP003.

I do agree that doing it in the parser will not work well.

The in inefficiency is that this is done for every rule that is applied. But also, you abstracted out the unbinding stuff into the `undo_bindings_list` and `clear_constraint` functions, which should be inlined instead.

matham:
@denys-duchier we had that whole convo about the compiler on IRC a week or two ago.

matham:
What I'd prefer is that we create a method of builder that allows clearing like your `clear_constraint` but called something like clear_rule instead. 

And instead of changing `_handler` like you did, which will make it more difficult for compiler, and I think is unneeded anyway, we should just iterate through the elements of `_handler[widget.uid]` and check the args which are saved there if it matches the rule name and then unbind and remove.

That'll make it more generic and easier for the compiler to replicate. Then all we need to do is add the `-rule` syntax to parser and a flag to the parser property class.

denys-duchier:
I hear a lot about breaking apps, and quite possibly I am being naive, but do we know of apps that actually use kv lang constraints multiply to achieve functionnal effects?  It seems insane to me, but maybe it is just not my turn of mind.  I'd like to look at an example of an actual app doing that, though.  it seems so weird.

kived:
@denys-duchier What you describe is not uncommon in multiple inheritance. Consider the following:

````python
class Base(object):
    def __init__(self):
        self.value = 0

    def fn(self):
        print 'value:', self.value

class A(Base):
    def fn(self):
        self.value += 1
        super(A, self).fn()

class B(Base):
    def fn(self):
        self.value -= 1
        super(B, self).fn()

class C(A, B):
    pass
````

Calling `C().fn()` will call both `A.fn` and `B.fn`. It is up to you, the implementer of `C`, to reconcile these differences in behavior if necessary. Hence the suggestion in KEP003 is an explicit skip previous rules operator `-`.

**Comment by denys-duchier** [comment]
@kived the python behaviour is statically predictable: C.fn will call A.fn which will call B.fn which will call Base.fn.  It doesn't depend on past history.

**Comment by matham** [comment]
I'm pretty sure I have made use of this in the past, although I cannot remember where.

As a specific example, consider

```py
<MyRule>:
    valve: OFF
    valve: self.parent1.valve
    valve: self.parent2.valve
```

Now, when parent valve 1 changes, our valve will change. When parent valve 2 changes, we'll also change. This allows it to listen to two valves and update independently from them.

denys-duchier:
@matham @dessant I started a new branch (unfinished, in fact untested so far) `property-constraints-redux` for tilded and prioritized constraints.  My idea is that:

```
<Foo>:
    10~baz: expr1
```
gives you a constraint on property `baz` with priority 10.  A later constraint:

```
<Foo>:
    5~baz: expr2
```
will not override it.  It would only override if its priority was >= 10.

A subsequent constraint:

```
<Foo>:
    baz: expr3
```
has priority None and is simply added to the existing set of constraints for this property.  Any subsequent constraint with priority not None overrides constraints with priority None.

A tilded property without a priority has implicitly priority 0:

```
<Foo>:
    ~baz: expr4
```
but maybe that's not the right default. perhaps it should have a mid-level priority.

thoughts?

tshirtman:
1/ What problem is it solving, exactly, to add priority levels to properties definitions? (i'm not sure why you call them constraints, they are not restraining the possible values, they are setting the value, i.e defining).
2/ I don't like the idea of numbered level priorities, it feels hardly maintainable and error prone to me, people will end up asking for a way to override even that system and we'll end up with an `!important` like thing, and we'll be as broken as CSS.
3/ I think if we want a definition to be final, it's enough to have a way to remove all the previous definitions, if we want not to override *all* the previous definition, it's enough to place the "clearing" signal at the right place in the stack, for clarity, it may be better to recreate the rules that you in fact didn't want to clear, rather than having to check at multiple levels if definition X or Y has been cleared by definitions of level N.
4/ there is probably no "good" default value for the priorities you are proposing, i fear it'll just be confusing.

denys-duchier:
It addresses the same issues as priority in XSLT or !important in CSS.

yes, it has complexity issues (and that was not my first choice), but so does the choice of tilding or not tilding a property.  as long as you have that (complex) choice, then you might as well have more control over it.  If you don't use it, then you are no worse off.  If you do use it, then presumably it is for good reasons.

If you want a tilded constraint that cannot be overriden, then I can easily add notation (such as ~~baz) to express it.

why I call them **constraints**: I am a *constraint programming* guy.  I am sure this must color my view of things.  A constraint is an expression that restricts the values that a certain variable may take.  When only one such value remains, the variable is said to be **determined**.

There is no confusion in what I proposed: if a non-tilded constraint is applied then it is simply added as it always was. if the constraint is tilded then it is meant to (possibly) override existing constraints. it overrides all constraints that have a priority <= to itself.

The generality I offer can easily be taken out.  I was not able to interpret your point 3.

matham:
I've created a PR implementing KEP003 with how I think it should work: https://github.com/kivy/kivy/pull/3432. It makes it optional by default.

akshayaurora:
closing this in favor of #3432

PR #3415, opened 2015-06-11
Colors in utils.py

andnovar:
Hello this pr adds the SmoothLine to the __init__.py file and adds colors by name in utils.py. Colors by name is an alternative instead of changing colors by rgba.

### DIFF
--- a/kivy/graphics/__init__.py
+++ b/kivy/graphics/__init__.py
@@ -94,12 +94,12 @@ from kivy.graphics.context_instructions import BindTexture, Color, \
     PopMatrix, PushMatrix, Rotate, Scale, Translate, LoadIdentity, \
     UpdateNormalMatrix, gl_init_resources
 from kivy.graphics.vertex_instructions import Bezier, BorderImage, Ellipse, \
-    GraphicException, Line, Mesh, Point, Quad, Rectangle, Triangle
+    GraphicException, Line, Mesh, Point, Quad, Rectangle, Triangle, SmoothLine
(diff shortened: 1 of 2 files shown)


Kovak:
Where did the name and values of these colors come from?

It seems like a rather arbitrary selection of colors to me, and the names rely on a fairly subjective understanding of what they mean. I think it could possibly be better to go with something more like Google's material design palette: http://www.google.com/design/spec/style/color.html#

This provides a full spectrum of colors, covering both general hue and variations in shade of that hue. In addition, it eschews subjective naming of colors for a numerical system qualifying the relative weights of shades. The relative groupings in the weighting system are not mathematically based, but instead based on how vibrant humans appear each to be so that they roughly appear 'equal' to our eyes  (Blue500 and Red500 appear to be of the same vibrancy to the human eye for instance)., an approach that is used heavily in print media discussions of color. 

It seems to me such an approach could provide both a larger spectrum for kivy users to select from, and one less rooted in cultural understands of descriptive words such as 'burlywood', 'goldenrod', or 'cornsilk'.

andnovar:
This was my reference http://avatar.se/molscript/doc/colour_names.html. I took them from here since they are already in rgb (0,1). I see your point specially since this are colors for UI.

tito:
If we need a palette, we need to have the standard web palette.
It is already in kivy, but hidden in the svg.pyx: https://github.com/kivy/kivy/blob/master/kivy/graphics/svg.pyx#L77

So what you really want is to make public this table. Not duplicate color :)

andnovar:
Agree, then I think the best is to make this public.

PR #3422, opened 2015-06-11
[WIP] Modifications to work with SDL2 on Android

inclement:
These are actually quite minimal, and are all that is actually needed to work with SDL2. Some of the details will need to be updated later (e.g. changing the NewPythonActivity references, and maybe changing how Hardware.java works more than I already did), but I believe these can and should be merged now, it will make it easier to test the sdl2 build tools.

Edit: Actually, not to be merged yet.

### DIFF
--- a/kivy/metrics.py
+++ b/kivy/metrics.py
@@ -104,6 +104,7 @@ __all__ = ('Metrics', 'MetricsBase', 'pt', 'inch', 'cm', 'mm', 'dp', 'sp',
 from os import environ
 from kivy.utils import reify, platform
 from kivy.properties import dpi2px
+from kivy.setupconfig import USE_SDL2
 
 
 def pt(value):
@@ -158,8 +159,13 @@ class MetricsBase(object):
             return float(custom_dpi)
 
         if platform == 'android':
-            import android
-            return android.get_dpi()
+            if USE_SDL2:
+                import jnius
+                Hardware = jnius.autoclass('org.renpy.android.Hardware')
+                return Hardware.getDPI()
+            else:
+                import android
+                return android.get_dpi()
         elif platform == 'ios':
             import ios
             return ios.get_dpi()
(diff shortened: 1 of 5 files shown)


inclement:
In one place I left the old android module call for use with pygame, but added the sdl2 one with pyjnius. This is because I had to modify Hardware.java with the line at https://github.com/inclement/p4a-experiment/blob/master/bootstrap_templates/sdl2/src/org/renpy/android/Hardware.java#L166 in order for it to work...I haven't tested if this is necessary with pygame (is it done somewhere else?).

kivy/metrics.py
```diff
@@ -158,8 +159,13 @@ def dpi(self):
             return float(custom_dpi)
 
         if platform == 'android':
-            import android
-            return android.get_dpi()
+            if USE_SDL2:
```
tito: i think we could always use jnius now.

kivy/metrics.py
```diff
@@ -217,8 +223,13 @@ def fontscale(self):
 
         if platform == 'android':
             from jnius import autoclass
-            PythonActivity = autoclass('org.renpy.android.PythonActivity')
-            config = PythonActivity.mActivity.getResources().getConfiguration()
+            if USE_SDL2:
+                # This activity name is temporary, it will need to be changed later
+                NewPythonActivity = autoclass('net.inclem.android.NewPythonActivity')
```
tito: org.kivy ? :)

inclement: Yep, I'll definitely change it later, I started this way to be clear it wasn't final :)

kivy/metrics.py
inclement: I had to modify Hardware.java to make it work (maybe due to an android bug, http://stackoverflow.com/questions/27590213/android-displaymetrics-densitydpi-returns-zero-on-rooted-devices) and I didn't test that with the pygame bootstrap. I'll change it all to pyjnius if that works.

inclement:
Replaced by #3648

PR #3424, opened 2015-06-12
Add get includes

Kovak:
Adds a get_includes function for use when building other modules such as KivEnt. If a module cimports heavily from the graphics section it needs access to headers such as gl_redirect.h or config.h.


matham:
Based on a search of the includes in that dir, I think you're maybe missing this folder/file: https://github.com/kivy/kivy/tree/master/kivy/tools/gles_compat.

PR #3429, opened 2015-06-14
Label "Sizing and text content" documentation and example

arlowhite:
Added a "Sizing and text content" section to the Label API to clarify how text_size and texture_size properties influence text. Link to this section from Button API.
Added an example demonstrating text wrapping: examples/widgets/label_texture_size.py

### DIFF
--- a/examples/widgets/label_texture_size.py
+++ b/examples/widgets/label_texture_size.py
@@ -1,0 +1,182 @@
+
+"""
+Label textsize
+============
+
+This example shows how to size a Label to its content (texture_size) and how
+setting text_size controls text wrapping.
+"""
+from kivy.app import App
+from kivy.clock import Clock
+from kivy.lang import Builder
+
+# Copied from https://en.wikipedia.org/wiki/A_Tale_of_Two_Cities
+# Published in 1859 and public domain.
+_example_text = """It was the best of times, it was the worst of times,
+it was the age of wisdom, it was the age of foolishness, it was the epoch of
+belief, it was the epoch of incredulity, it was the season of Light, it was
+the season of Darkness, it was the spring of hope, it was the winter of despair,
+we had everything before us, we had nothing before us, we were all going
+direct to Heaven, we were all going direct the other way - in short,
+the period was so far like the present period, that some of its noisiest
+authorities insisted on its being received, for good or for evil, in the
+superlative degree of comparison only.
+"""
+
+_kv_code = """
+# Note: StackLayout, ToggleButton, and HeadingLabel are defined at the bottom.
+Widget:
+    GridLayout:
+        id: grid_layout
+        cols: 2
+        top: root.top
+        width: root.width
+        height: cm(6)
+
+        HeadingLabel:
+            text: "GridLayout (1st column): Default, no text_size set"
+
+        HeadingLabel:
+            text: 'GridLayout: (2nd column): text_size bound to size'
+
+        Label:
+            id: left_content
+            canvas:
+                Color:
+                    rgb: 68/255.0, 164/255.0, 201/255.0
+                Line:
+                    rectangle: self.x, self.y, self.width, self.height
+
+        Label:
+            id: right_content
+            text_size: self.size
+            padding: dp(6), dp(6)
+
+            canvas:
+                Color:
+                    rgb: 68/255.0, 164/255.0, 201/255.0
+                Line:
+                    rectangle: self.x, self.y, self.width, self.height
+
+        StackLayout:
+
+            ToggleButton:
+                text: 'Disable'
+                on_state:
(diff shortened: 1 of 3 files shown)


kivy/uix/button.py
```diff
@@ -12,6 +12,9 @@
 
     button = Button(text='Hello world', font_size=14)
 
+.. note:: To fit the Button to its :attr:`~Button.text`, bind size to :attr:`~kivy.uix.label.Label.texture_size`
```
matham: I don't think this should be here, as we cannot document about parent behavior in every subclass. At some point we have to assume that the user will read the parent doc, rather than duplicate it, or mention it again. Especially, since it's mentioned right above that, that it inherits from Label.

matham:
Overall, it's great, however, there are python pep8 style issues. In particular, your lines are too long. Lines are limited to 80 chars. See here how to make git automatically show you the pep8 issues: http://kivy.org/docs/contribute.html#coding-style.

arlowhite:
I also realized I haven't discussed how setting size without text_size works.
There is a Note in the text_size attribute doc:
> This text_size property is the same as the usersize property in the Label class. (It is named size= in the constructor.)

But I think it needs to be explained more, so I'll do a bit more work on that.

kivy/uix/button.py
arlowhite: Ok, what do you think about expanding the last sentence instead?

Currently:
> To configure the button, the same properties are used as for the Label class:

Change to:

> To configure the button, the same properties (padding, font_size, etc) and sizing system are used as for the Label class:

matham: That would be fine.

arlowhite:
Ok, I've fixed PEP8 issues, removed the Note from Button and changed the last sentence instead.
I expanded examples/widgets/label_texture_size.py
* size with no text_size, and text_size bound to size examples.
* Shorten, max_lines, valign buttons
* Demonstrates fairly complex layout itself in how the example places buttons

max_lines doesn't actually behave the way I expect, does it have a bug?
Also, the GridLayout doesn't look the same if you shrink the window width and expand it again. Not sure the best way to fix this.

This example is a bit complicated now. Is this an appropriate location and filename for it?

kivy/uix/label.py
```diff
@@ -16,12 +16,50 @@
...
+
+For example, this label's size will be set to the text content
+(plus :attr:`~Label.padding`):
+
+.. code-block:: python
+
+    Label:
+        size: self.texture_size
+
+This label's text will be clipped to the specified width and height:
```
arlowhite: I think I want to change this to:
> This label's text will wrap at the specified width and be clipped to the height:

Alternatively, is it better style to recommend this?:
```python
Label:
    size: cm(6), cm(4)
    text_size: self.size
```

arlowhite:
I changed that sentence to: "This label's text will wrap at the specified width and be clipped to the height:"
Squashed commits to just two.
I'm done making changes to the branch unless you have any other comments. The only thing I've been debating is whether to rename the example file to something else. Maybe *label_sizing.py*?

examples/widgets/label_texture_size.py
```diff
@@ -0,0 +1,182 @@
...
+                rgb: 68/255.0, 164/255.0, 201/255.0
+            Line:
+                rectangle: self.x, self.y, self.width, self.height
+
+    Button:
+        # Button is a subclass of Label and can be sized to text in the same way
+        text: 'Reset'
+        padding: mm(6), mm(3)
+        size: self.texture_size
+        # Setting right this way didn't work, not sure why.
```
arlowhite: This isn't actually a bug, you can't set right and then change width and expect right to stay aligned. I created #3430 and new documentation explains this. I will remove these comments.

examples/widgets/label_texture_size.py
```diff
@@ -0,0 +1,182 @@
...
+                rgb: 68/255.0, 164/255.0, 201/255.0
+            Line:
+                rectangle: self.x, self.y, self.width, self.height
+
+    Button:
+        # Button is a subclass of Label and can be sized to text in the same way
+        text: 'Reset'
+        padding: mm(6), mm(3)
+        size: self.texture_size
+        # Align to right side and vertically align-middle with title
```
arlowhite: Removed the comment about binding right not working.
Right-align and vertical-align Reset button using property calculations and link to relevant doc. (Violates PEP8, but I think it's excusable for a URL)

- [ ] The question is do we want to encourage developers to do this? Or should I just change it to a FloatLayout using pos_hint as mentioned in the documentation?

arlowhite:
I've decided to split this into two separate new Pull Requests since I have a few questions on the example but the docs are ready to go. Also, this PR is a bit confusing due to the history rewrites.

PR #3432, opened 2015-06-14
Implements KEP003

### LINKED ISSUE #2365: KEP003: Syntax to clear previous kv rules of a property

matham:
Fixes #2365. After this pr, the following code

```py
w = Builder.load_string('''
<MyWidget@Widget>:
    overflow: self.x
    on_overflow: print('Overflow is "{}"!'.format(self.overflow))

Widget:
    MyWidget:
        overflow: self.y * 10
''')
w.children[0].x = 123
w.children[0].y = 321
```
prints
```
Overflow is "123"!
Overflow is "3210"!
```

And changing `overflow: self.y * 10` to `-overflow: self.y * 10` prints instead 
```
Overflow is "3210"!
```

### DIFF
--- a/kivy/lang.py
+++ b/kivy/lang.py
@@ -648,6 +648,43 @@ and in my.kv:
 MyWidget will now have a Color and Rectangle instruction in its canvas
 without any of the instructions inherited from the Label.
 
+Redefining a widget's property style
+------------------------------------
+
+Similar to :ref:`Redefining a widget's style`, sometimes we would like to
+inherit from a widget, keep all its KV defined styles, except for the style
+applied to a specific property. For example, we would
+like to inherit from a :class:`~kivy.uix.button.Button`, but we would also
+like to set our own `state_image`, rather then relying on the
+`background_normal` and `background_down` values. We can achieve this by
+prepending a dash (-) before the `state_image` property name in the .kv style
+definition.
+
+In myapp.py:
+
+.. code-block:: python
+
+    class MyWidget(Button):
+
+        new_background = StringProperty('my_background.png')
+
+and in my.kv:
+
+.. code-block:: kv
+
+    <MyWidget>:
+        -state_image: self.new_background
+
+MyWidget will now have a `state_image` background set only by `new_background`,
+and not by any previous styles that may have set `state_image`.
+
+.. note::
+
+    Although the previous rules are cleared, they are still applied during
+    widget constrction, and are only removed when the new rule with the dash
+    is reached. This means that initially, previous rules could be used to set
+    the property.
+
 Lang Directives
 ---------------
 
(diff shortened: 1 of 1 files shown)


denys-duchier:
I have a proposal (and code that implements it) for having the semantics both your way and my way, and making a possible change of default in kivy 2.0 painless.  Allow the following explicit property annotations:
```
-foo: e1()
```
for explicitly clearing earlier constraints, and
```
+baz: e2()
```
for explicitly adding to the existing constraints.   A non annotated constraint:
```
baz: e3()
```
would then adopt the current default annotation.  The default annotation would be `+` for kivy 1.9, but it could also be explicitly selected by a directive (valid only for the file (or string) in which it appears):
```#:property-annotation-default -```
or
```#:property-annotation-default +```
With such an explicit directive, a kv file written today would retain the same semantics even if the global default was changed in 2.0.

What do you think?

dessant:
@denys-duchier, I think clearing previous rules is where we should draw the line. I'm not a new kivy user, but your proposal feels massively confusing.

akshayaurora:
@matham +1 for the implementation and sane defaults.

PR #3436, opened 2015-06-15
Mpl kivy

andnovar:
PR for matplotlib backend agg with events connected to kivy

### DIFF
--- a/kivy/ext/mpl/backend_kivy.py
+++ b/kivy/ext/mpl/backend_kivy.py
@@ -6,7 +6,7 @@ import six
 import matplotlib
 from matplotlib._pylab_helpers import Gcf
 from matplotlib.backend_bases import RendererBase, GraphicsContextBase,\
-     FigureManagerBase, FigureCanvasBase
+    FigureManagerBase, FigureCanvasBase
 from matplotlib.figure import Figure
 from matplotlib.transforms import Bbox
 from matplotlib.backend_bases import ShowBase
@@ -27,7 +27,8 @@ from kivy.base import EventLoop
 from kivy.graphics import Color, Line
 
 # try:
-#     kivy.require('1.9.0') # I would need to check which release of Kivy would be the best suitable.
+#     kivy.require('1.9.0') # I would need to check which release of
+#     Kivy would be the best suitable.
 # except AttributeError:
 #     raise ImportError(
 #         "kivy version too old -- it must have require_version")
@@ -40,27 +41,28 @@ app = None
 
 
 class MPLKivyApp(App):
- 
+
     def __init__(self, **kwargs):
         super(MPLKivyApp, self).__init__(**kwargs)
         self.figure = kwargs['figure']
- 
+
     def build(self):
         return self.figure
- 
+
     def on_pause(self):
         return App.on_pause(self)
- 
+
     def on_resume(self):
         App.on_resume(self)
- 
- 
+
+
 def _create_App(fig_canvas):
     global app
     if app is None:
         if _debug:
             print("Starting up Kivy Application")
-        app = MPLKivyApp(figure = fig_canvas)
+        app = MPLKivyApp(figure=fig_canvas)
+
 
 class RendererKivy(RendererBase):
     """
@@ -85,12 +87,12 @@ class RendererKivy(RendererBase):
         with self.widget.canvas:
             Color(1.0, 0.5, 0.5)
             Line(points=points_line, width=2.0)
-        
 
     # draw_markers is optional, and we get more correct relative
-    # timings by leaving it out.  backend implementers concerned with
+    # timings by leaving it out. backend implementers concerned with
     # performance will probably want to implement it
-#     def draw_markers(self, gc, marker_path, marker_trans, path, trans, rgbFace=None):
+#     def draw_markers(self, gc, marker_path, marker_trans, path,
+#     trans, rgbFace=None):
 #         pass
 
     # draw_path_collection is optional, and we get more correct
@@ -103,7 +105,7 @@ class RendererKivy(RendererBase):
 #         pass
 
     # draw_quad_mesh is optional, and we get more correct
-    # relative timings by leaving it out.  backend implementers concerned with
+    # relative timings by leaving it out. backend implementers concerned with
     # performance will probably want to implement it
 #     def draw_quad_mesh(self, gc, master_transform, meshWidth, meshHeight,
 #                        coordinates, offsets, offsetTrans, facecolors,
@@ -139,12 +141,12 @@ class RendererKivy(RendererBase):
 
 class GraphicsContextKivy(GraphicsContextBase):
     """
-    The graphics context provides the color, line styles, etc...  See the gtk
+    The graphics context provides the color, line styles, etc... See the gtk
     and postscript backends for examples of mapping the graphics context
     attributes (cap styles, join styles, line widths, colors) to a particular
-    backend.  In GTK this is done by wrapping a gtk.gdk.GC object and
+    backend. In GTK this is done by wrapping a gtk.gdk.GC object and
     forwarding the appropriate calls to it using a dictionary mapping styles
-    to gdk constants.  In Postscript, all the work is done by the renderer,
+    to gdk constants. In Postscript, all the work is done by the renderer,
     mapping line styles to postscript calls.
 
     If it's more appropriate to do the mapping at the renderer level (as in
@@ -159,14 +161,13 @@ class GraphicsContextKivy(GraphicsContextBase):
     """
     pass
 
-
-
 ########################################################################
 #
 # The following functions and classes are for pylab and implement
 # window/figure managers, etc...
 #
 ########################################################################
+
 
 def draw_if_interactive():
     """
@@ -179,26 +180,28 @@ def draw_if_interactive():
         if figManager is not None:
             figManager.canvas.draw_idle()
 
+
 class Show(ShowBase):
     def mainloop(self):
         print("calling show")
         global app
         if app is not None:
             app.run()
-  
+
 show = Show()
 
 # def show():
 #     """
 #     For image backends - is not required
 #     For GUI backends - show() is usually the last line of a pylab script and
-#     tells the backend that it is time to draw.  In interactive mode, this may
-#     be a do nothing func.  See the GTK backend for an example of how to handle
+#     tells the backend that it is time to draw. In interactive mode, this may
+#     be a do nothing func. See the GTK backend for an example of how to handle
 #     interactive versus batch mode
 #     """
 #     for manager in Gcf.get_all_fig_managers():
 #         # do something to display the GUI
 #         manager.show()
+
 
 def new_figure_manager(num, *args, **kwargs):
     """
@@ -207,7 +210,7 @@ def new_figure_manager(num, *args, **kwargs):
     # if a main-level app must be created, this (and
     # new_figure_manager_given_figure) is the usual place to
     # do it -- see backend_wx, backend_wxagg and backend_tkagg for
-    # examples.  Not all GUIs require explicit instantiation of a
+    # examples. Not all GUIs require explicit instantiation of a
     # main-level app (egg backend_gtk, backend_gtkagg) for pylab
     FigureClass = kwargs.pop('FigureClass', Figure)
     thisFig = FigureClass(*args, **kwargs)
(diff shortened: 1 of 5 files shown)


kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
+"""
```
matham: You should replace this documentation with your own documentation specific to kivyagg.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+
+where my_backend.py is your module name. This syntax is also
+recognized in the rc file and in the -d argument in pylab, e.g.,::
+
+  python simple_plot.py -dmodule://my_backend
+
+If your backend implements support for saving figures (i.e. has a print_xyz()
+method) you can register it as the default handler for a given file type
+
+  from matplotlib.backend_bases import register_backend
```
matham: Please implement this, as kivy supports saving to file.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+from matplotlib._pylab_helpers import Gcf
+from matplotlib.backend_bases import RendererBase, GraphicsContextBase,\
+    FigureManagerBase, FigureCanvasBase
+from matplotlib.figure import Figure
+from matplotlib.transforms import Bbox
+from matplotlib.backends.backend_agg import FigureCanvasAgg
+from kivy.input.motionevent import MotionEvent
+
+try:
+    import kivy
```
matham: Only the `import kivy` should be under the try/except, the rest should be after it, and outside it.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+except ImportError:
+    raise ImportError("this backend requires Kivy to be installed.")
+
+try:
+    kivy.require('1.9.0')
+except AttributeError:
+    raise ImportError(
+        "kivy version too old -- it must have require_version")
+
+from PIL import Image
```
matham: You don't need this, see my comment later.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+        #renderer = RendererKivyAgg(self.figure.dpi)
+        #self.figure.draw(renderer)
+
+    # You should provide a print_xxx function for every file format
+    # you can write.
+
+    # If the file type is not in the base set of filetypes,
+    # you should add it to the class-scope filetypes dictionary as follows:
+    filetypes = FigureCanvasBase.filetypes.copy()
+    filetypes['foo'] = 'My magic Foo format'
```
matham: This needs to be changed to support only png.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def blit(self, bbox=None):
+        # If bbox is None, blit the entire canvas to the widget. Otherwise
+        # blit only the area defined by the bbox.
+        self.blitbox = bbox
+
+    def print_foo(self, filename, *args, **kwargs):
```
matham: This needs to be changed to save png, and it should use http://kivy.org/docs/api-kivy.core.image.html#kivy.core.image.Image.save.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+        to their original values after this call, so you don't need to
+        save and restore them.
+        """
+        l, b, w, h = self.figure.bbox.bounds
+        im = Image.frombuffer('RGBA', (w, h), self.get_renderer().buffer_rgba(),
+                              "raw", "RGBA", 0, 1)
+        im.save(filename)
+
+    def get_default_filetype(self):
+        return 'foo'
```
matham: This should also be changed.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+        FigureCanvasAgg.__init__(self, figure)
+        Widget.__init__(self, **kwargs)
+        self.figure = figure
+        self.blit()
+
+    def draw(self):
+        """
+        Draw the figure using the renderer
+        """
+        FigureCanvasAgg.draw(self)
```
matham: You should be using super here, and the same elsewhere.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+            buf_rgba = self.get_renderer().buffer_rgba()
+        else:
+            bbox = self.blitbox
+            l, b, r, t = bbox.extents
+            w = int(r) - int(l)
+            h = int(t) - int(b)
+            t = int(b) + h
+            reg = self.copy_from_bbox(bbox)
+            buf_rgba = reg.to_string_argb()
+        texture = Texture.create(size=(w, h))
```
matham: You need to cache the texture and only create a new texture if the size changed. Like I do at https://github.com/matham/filers/blob/master/filers/misc_widgets.py#L267.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def get_default_filetype(self):
+        return 'foo'
+
+
+class FigureManagerKivyAgg(FigureManagerBase):
+    """
+    Wrap everything up into a window for the pylab interface
+
+    For non interactive backends, the base class does all the work
```
matham: We need to talk and fully understand on how kivy should manage interactive and non-interactive mode, and how it translates to kivy.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+    e.g., backend_gtk.py, backend_wx.py and backend_tkagg.py
+    """
+
+    def __init__(self, figure, **kwargs):
+        if _debug:
+            print('FigureCanvasKivyAgg: ', figure)
+        #super(FigureCanvasKivyAgg, self).__init__(figure, **kwargs)
+        FigureCanvasAgg.__init__(self, figure)
+        Widget.__init__(self, **kwargs)
+        self.figure = figure
```
matham: This is not needed since you call `FigureCanvasAgg.__init__(self, figure)`.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,216 @@
...
+    raise ImportError("this backend requires Kivy to be installed.")
+
+try:
+    kivy.require('1.9.0')
+except AttributeError:
+    raise ImportError(
+        "kivy version too old -- it must have require_version")
+
+from PIL import Image
+EventLoop.ensure_window()
```
matham: I don't think this is needed and should be removed.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,386 @@
...
+EventLoop.ensure_window()
+
+_debug = True
+
+app = None
+
+def _create_App(fig_canvas):
+    global app
+
+    class TestApp(App):
```
matham: This class should be defined outside the function, and instead should take a `fig_canvas` parameter. So you can just do `TestApp(figure=fig_canvas)`.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,386 @@
...
+    class methods button_press_event, button_release_event,
+    motion_notify_event, key_press_event, and key_release_event.  See,
+    e.g., backend_gtk.py, backend_wx.py and backend_tkagg.py
+    """
+    
+    def __init__(self, figure, **kwargs):
+        if _debug:
+            print('FigureCanvasKivy: ', figure)
+        
+        self._keyboard = Window.request_keyboard(self._keyboard_closed, self)
```
matham: Lets talk about using focus behavior here.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,386 @@
...
+        self._keyboard.bind(on_key_down = self._on_keyboard_down)
+        self._keyboard.bind(on_key_up = self._on_keyboard_up)
+        Window.bind(mouse_pos = self._on_mouse_pos)
+        Window.bind(on_close = self._on_close)
```
matham: As mentioned, this should probably bind to the figure being closed, rather than the whole window. Because in kivy, the window will always stay open, even if the figure dies.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,386 @@
...
+        self.bind(size = self._on_size_changed)
+        
+        super(FigureCanvasKivy, self).__init__(figure, **kwargs)
+        _create_App(self)
```
matham: This should probably not happen here, and should depend on whether it's interactive or not.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,386 @@
...
+        #renderer = RendererKivy(self.figure.dpi)
+        #self.figure.draw(renderer)
+
+    # You should provide a print_xxx function for every file format
+    # you can write.
+
+    # If the file type is not in the base set of filetypes,
+    # you should add it to the class-scope filetypes dictionary as follows:
+    filetypes = FigureCanvasBase.filetypes.copy()
+    filetypes['foo'] = 'My magic Foo format'
```
matham: Fix this.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,386 @@
...
+    
+    def on_touch_down(self, touch):
+        if super(FigureCanvasKivy, self).on_touch_down(touch):
+            return True
+        if self.collide_point(*touch.pos):
```
matham: You need to generate `motion_notify_event`, `enter_notify_event`, and `leave_notify_event` events here. You generate this below for mouse, however, what if there's no mouse and just a touch screen, you still need to generate them. In fact, even with a mouse, you should generate it here, the only exception should be if the mouse is not held down, then you should generate them from `_on_mouse_pos`.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,386 @@
...
+
+import six
+
+import matplotlib
+from matplotlib._pylab_helpers import Gcf
+from matplotlib.backend_bases import RendererBase, GraphicsContextBase,\
+     FigureManagerBase, FigureCanvasBase
+from matplotlib.figure import Figure
+from matplotlib.transforms import Bbox
+from backend_kivyagg import FigureCanvasKivyAgg
```
matham: Fix this import to be absolute. Also, we should talk about this import.

kivy/ext/mpl/backend_kivyagg.py
```diff
@@ -0,0 +1,155 @@
...
+                update = True
+        if update or self.img_texture is None:
+            texture = Texture.create(size=(w, h))
+            texture.flip_vertical()
+        else:
+            texture = self.img_texture
+        texture.blit_buffer(buf_rgba, colorfmt='rgba', bufferfmt='ubyte')
+        self.img_texture = texture
+        with self.canvas:
+            Rectangle(texture=texture, pos=self.pos, size=(w, h))
```
matham: This part should only happen if we create a new texture.

andnovar: done

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,377 @@
...
+    motion_notify_event, key_press_event, and key_release_event.  See,
+    e.g., backend_gtk.py, backend_wx.py and backend_tkagg.py
+    """
+    
+    def __init__(self, figure, **kwargs):
+        if _debug:
+            print('FigureCanvasKivy: ', figure)
+        Window.bind(mouse_pos = self._on_mouse_pos)
+        self.bind(size = self._on_size_changed)
+        self.flag = True
```
matham: This really needs to have a better name.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,377 @@
...
+        return EventLoop.window.title
+
+    def set_window_title(self, title):
+        EventLoop.window.title = title
+
+    def resize(self, w, h):
+        Window.size(w, h)
+
+    def destroy(self):
+        EventLoop.close()
```
matham: This should stop the app possibly rather than calling here.

matham:
Lets close it here.

Issue #3430, opened 2015-06-14
Widget.right binding does not work as expected when width changes after being positioned.


arlowhite:
**Edit:**
Originally, I created this issue regarding Labels with size bound to texture_size. But this issue affects any widget where you bind its right property but then its width changes after being positioned. Instead of the widget's right side staying fixed and x changing, x stays the same and the value of right changes (x + width).

**Original Issue:**
Setting Label.right in KV doesn't work correctly when size is bound to texture_size:
```python
    Label:
        text: 'Label'
        size: self.texture_size
        right: root.right
```
Instead, the Label is placed with x=root.right. The Label will jump to the correct position when parent/window is resized.

This can be worked around by setting x this way instead:
```python
    Label:
        text: 'Label'
        size: self.texture_size
        x: root.right - self.width
```

The same issue occurs with top: root.top

arlowhite:
printing out when x, right, or width change:

> initial         x=0 right=100 width=100
> call to GraphicUnitTest.render()
> **window width set to 320, so label.right is set to 320**
> right_change    x=220 right=320 width=100
> x_change        x=220 right=320 width=100
> **texture renders and texture_size changes label.width to 84**
> right_change    x=220 right=304 width=84
> width_change    x=220 right=304 width=84

When the Label's texture_size is determined, the width is set to 84, but instead of x changing relative to right, right changes relative to x. x:220 + width:84 = right:304

arlowhite:
I'm not sure this is a bug exactly. The main issue is that the parent sets Label.right before texture_size updates.

However, even if this were fixed, the issue would still occur if the text were to change after Label was positioned.

For binding in python code there's really nothing that can be done, but this all started with KV language:

Maybe the KV processor can just translate **right: EXPRESSION** to **x: EXPRESSION - self.width** and **top: EXPRESSION** to **y: EXPRESSION - self.height**

andnovar:
Is this Label inside a GridLayout ? if so it could be related to this issue https://github.com/kivy/kivy/pull/3216 ?

arlowhite:
No, it happens with just a Label inside a plain Widget. Though I did notice an issue with GridLayout, I'll comment in #3216.

kived:
Well, it's doing exactly what you told it. Set Label.right to it's
parent.right whenever the parent.right changes. You didn't tell it to check
when the Label.width changes (in the second one, you did, hence it worked).
kv only binds to the properties you mention in your expression.

[...]

PR #3439, opened 2015-06-16
GraphicUnitTest for issue #3430 (Label size: self.texture_size positioning top/right)

arlowhite:
kivy/kivy#3430
Tests that a Label with size bound to texture_size positions on right and top correctly on 1st render.

### DIFF
--- a/kivy/tests/test_issue_3430.py
+++ b/kivy/tests/test_issue_3430.py
@@ -1,0 +1,104 @@
+"""
+Issue #3430 test
+When Label.size is bound to Label.texture_size, it is not initially positioned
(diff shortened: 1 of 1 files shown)


arlowhite:
The two renders in each test_ method should be identical.

Unfortunately, I can't test render comparisons, because when I run **make test** all unittest render images bleed together. I'm not sure if this is an OpenGL issue on my end or something else. However the assertions do catch the bug and I do see the label move between test_right-1 and test_right-2.

arlowhite:
Unittest not necessary since setting Widget.right in this way and expecting it to be width aware is not supported. See conversation in #3430

PR #3443, opened 2015-06-17
Couple of Corrections..

gouzouni:
Couple of corrections that had to do with Image's source reloading and Dropdown height updating.

### DIFF
--- a/kivy/uix/dropdown.py
+++ b/kivy/uix/dropdown.py
@@ -314,9 +314,10 @@ class DropDown(ScrollView):
         h_bottom = wy - self.height
         h_top = win.height - (wtop + self.height)
         if h_bottom > 0:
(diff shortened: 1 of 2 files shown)


akshayaurora:
@gouzouni thanks,  for future reference try to separate the fixes in separate pr's :+1:

gouzouni:
@akshayaurora Ok will seperate all future.
However keep in mind Fix for problem problem: "Height of a DropDown List (once opened) is not properly updated after window resize to smaller heights." could cause too many recursives problem so one need to check on that. My solution is just a hint.

The first correction about "Reloading an image that had incorrect source from the beginning, would cause ('NoneType' object has no attribute 'remove_from_cache')" is fine and works.

PR #3447, opened 2015-06-18
Fix property comparison

matham:
For the code
```py
class MyWidget(Widget):
    obj = ObjectProperty(None)

w = MyWidget()
w.obj = np.arange(10)
w.obj = np.arange(10)
```

Before, it'd do 
```py
   File "G:\Python\libs\Playground\src\playground6.py", line 28, in <module>
     w.obj = np.arange(10)
   File "kivy\properties.pyx", line 404, in kivy.properties.Property.__set__ (kivy\properties.c:4927)
   File "kivy\properties.pyx", line 430, in kivy.properties.Property.set (kivy\properties.c:5449)
 ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

Now it does only 
```
[WARNING           ] [Property    ] Value comparison failed for <ObjectProperty name=obj> with "The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()". Consider setting force_dispatch to True to avoid this.
```

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -209,6 +209,7 @@ from kivy.config import ConfigParser
 from functools import partial
 from kivy.clock import Clock
 from kivy.weakmethod import WeakMethod
+from kivy.logger import Logger
 
 cdef float g_dpi = -1
 cdef float g_density = -1
@@ -282,11 +283,11 @@ cdef class Property:
             If set, it will replace an invalid property value (overrides
             errorhandler).
 
-            If the paramters include `force_dispatch`, it should be a boolean.
-            If True, the property event will be dispatched even if the new
-            value matches the old value (by default identical values are not
-            dispatched to avoid infinite recursion in two-way binds). Be
-            careful, this is for advanced use only.
+            If the parameters include `force_dispatch`, it should be a boolean.
+            If True, no value comparison will be done, so the property event
+            will be dispatched even if the new value matches the old value (by
+            default identical values are not dispatched to avoid infinite
+            recursion in two-way binds). Be careful, this is for advanced use only.
 
     .. versionchanged:: 1.4.2
         Parameters errorhandler and errorvalue added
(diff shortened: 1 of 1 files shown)


inclement:
Just to check I understand, the reason for implementing the compare method on several different properties is that they each shouldn't have the error catching because comparison should be unambiguous?

inclement:
Also, in the interests of science, I'm just abusing this issue to see how long a commit message KGB will print in #kivy-dev. I'll alert the crew. Well, I'll say this for him - he's sure of himself. Captain, why are we out here chasing comets? When has justice ever been as simple as a rule book? Well, that's certainly good to know. I'm afraid I still don't understand, sir. You did exactly what you had to do. You considered all your options, you tried every alternative and then you made the hard choice. Wait a minute - you've been declared dead. You can't give orders around here. Computer, belay that order. I can't. As much as I care about you, my first duty is to the ship. Then maybe you should consider this: if anything happens to them, Starfleet is going to want a full investigation. Talk about going nowhere fast. Mr. Crusher, ready a collision course with the Borg ship. And blowing into maximum warp speed, you appeared for an instant to be in two places at once. Mr. Worf, you sound like a man who's asking his friend if he can start dating his sister. Smooth as an android's bottom, eh, Data? But the probability of making a six is no greater than that of rolling a seven. Worf, It's better than music. It's jazz. Sure. You'd be surprised how far a hug goes with Geordi, or Worf. The game's not big enough unless it scares you a little. Congratulations - you just destroyed the Enterprise.

matham:
@inclement yes.

inclement:
Would it be better to do it the other way around, just add the try/except to ObjectProperty and assume == otherwise? It might be less code reuse.

Since this might be bikeshedding, just say no and I'll merge it if you want.

matham:
I don't think so, because I think by default, whether it be a OptionProperty, a ObjectProperty or alias property we should assume a comparison can throw an error. It's only when we know for sure, e.g. for a string, numeric, or `list` `tuple` typed list that we should not have the except.

inclement:
Okay, you've convinced me. Nice fix :)

PR #3448, opened 2015-06-18
Added "Sizing and text content" section to Label API documentation.

arlowhite:
Added a "Sizing and text content" section to the Label API to clarify how text_size and texture_size properties influence text. Add subtle link to this section from Button API.

@matham This is just the documentation portion from our discussion on #3429

### DIFF
--- a/kivy/uix/label.py
+++ b/kivy/uix/label.py
@@ -16,12 +16,50 @@ strings::
     # size
     l = Label(text='Hello world', font_size='20sp')
 
+.. _kivy-uix-label-sizing-and-text-content:
+
+Sizing and text content
+---------------------------
+
+By default, the size of :class:`Label` is not affected by :attr:`~Label.text`
+content and the text is not affected by the size. In order to control
+sizing, you must specify :attr:`~Label.text_size` to constrain the text
+and/or bind :attr:`~Label.size` to :attr:`~Label.texture_size` to grow with
+the text.
+
+For example, this label's size will be set to the text content
+(plus :attr:`~Label.padding`):
+
+.. code-block:: python
+
+    Label:
+        size: self.texture_size
+
+This label's text will wrap at the specified width and be clipped to the height:
+
+.. code-block:: python
+
+    Label:
+        text_size: cm(6), cm(4)
+
+.. note:: The :attr:`~Label.shorten` and :attr:`~Label.max_lines` attributes
+ control how overflowing text behaves.
+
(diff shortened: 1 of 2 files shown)


kivy/uix/label.py
```diff
@@ -16,12 +16,50 @@
...
+        text_size: root.width, None
+        size: self.texture_size
+
 Text alignment and wrapping
 ---------------------------
 
-The :class:`Label` has :attr:`halign` and :attr:`valign` properties to
-control the alignment of its text, but by default these have no effect
-and the text is always centered within the Label. This is for
+The :class:`Label` has :attr:`~Label.halign` and :attr:`~Label.valign`
```
matham: Upon reading this, I don't think this is True, is it? Because I think any lines shorter than the maximum line width are still properly aligned. Can you double check.

arlowhite: This is correct, though a bit technical.

matham: Try running 
```py
Label:
    halign: 'center'
    text:
        "Lorem ipsum dolor sit amet, consectetur adipiscing elit\\nLorem ipsum dolor sit amet, consecte"
'''))
```
And you'll see that it is not correct. Change the halign and you'll see that it does get aligned within the texture.

arlowhite: Ah, you're right, I should have realized I needed a newline character to test it. Though valign has no effect.

arlowhite:
Ok, I've cleaned up the alignment section and put a few sentences on efficiency in a Note. I'm done with the branch unless you have more changes.

kivy/uix/label.py
```diff
@@ -16,38 +16,73 @@
...
+will also automatically wrap the text of the Label to remain within this area.
+
+.. code-block:: python
+
+    Label:
         text_size: self.size
         halign: 'right'
         valign: 'middle'
 
+.. note:: The reason text_size is not bound to size by default is for
```
matham: This is not really needed, because saying this justification doesn't help users and as devs, we already know this.

matham:
I think it's ready to be merged once you remove that note.

kivy/uix/label.py
arlowhite: Ok, I was just trying to preserve the original author's commentary, but I think you're right. I'll remove it in a minute.

arlowhite:
Ok, removed the efficiency Note.

matham:
Great :+1:

PR #3449, opened 2015-06-18
Label example that demonstrates sizing to texture_size and wrapping with text_size.

arlowhite:
Creates a new example: *examples/widgets/label_texture_size.py*

I spent some time looking through Kivy's examples and never found one that really helped me understand how texture_size and text_size are used. So, I created this example while working on some documentation #3448.

A few issues:
- [x] Is this a good file name and location? Maybe *label_sizing.py* would be better?
- [x] max_lines does not behave as I would expect. Is this a problem with my example or max_lines itself?
- [x] I was thinking of creating a button that outlines the Label.textures, would this be useful or just overcomplicate the example, which is already a bit complicated.

@matham This is just the example from our discussion on #3429

### DIFF
--- a/examples/widgets/label_sizing.py
+++ b/examples/widgets/label_sizing.py
@@ -1,0 +1,188 @@
+
+"""
+Label textsize
+============
+
+This example shows how to size a Label to its content (texture_size) and how
+setting text_size controls text wrapping.
+"""
+from kivy.app import App
+from kivy.clock import Clock
+from kivy.lang import Builder
+from kivy.properties import StringProperty, NumericProperty, BooleanProperty
+
+# Copied from https://en.wikipedia.org/wiki/A_Tale_of_Two_Cities
+# Published in 1859 and public domain.
+# The newline after the title will help demonstrate halign
+_example_title_text = 'A Tale of Two Cities, by Charles Dickens\n'
+_example_text = """It was the best of times, it was the worst of times,
+it was the age of wisdom, it was the age of foolishness, it was the epoch of
+belief, it was the epoch of incredulity, it was the season of Light, it was
+the season of Darkness, it was the spring of hope, it was the winter of despair,
+we had everything before us, we had nothing before us, we were all going
+direct to Heaven, we were all going direct the other way - in short,
+the period was so far like the present period, that some of its noisiest
+authorities insisted on its being received, for good or for evil, in the
+superlative degree of comparison only.
+"""
+
+# Note: Many of the Widgets (StackLayout, ToggleButton, Spinner) have
+# defaults set at the bottom of the KV, where  DemoLabel and HeadingLabel
+# are also defined.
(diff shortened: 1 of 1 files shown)


examples/widgets/label_texture_size.py
```diff
@@ -0,0 +1,182 @@
...
+                rgb: 68/255.0, 164/255.0, 201/255.0
+            Line:
+                rectangle: self.x, self.y, self.width, self.height
+
+    Button:
+        # Button is a subclass of Label and can be sized to text in the same way
+        text: 'Reset'
+        padding: mm(6), mm(3)
+        size: self.texture_size
+        # Align to right side and vertically align-middle with title
```
arlowhite: Here, I right-align and vertical-align the Reset button using property calculations and link to relevant doc. (Line length violates PEP8, but I think it's excusable for a URL)

- [ ] Should developers be encouraged to do this? Or would it be better to demonstrate using a FloatLayout with pos_hint as mentioned in the documentation?

examples/widgets/label_texture_size.py
```diff
@@ -0,0 +1,182 @@
...
+        # Note: These calculations make the property dependencies explicit, see:
+        # http://kivy.org/docs/api-kivy.uix.widget.html#usage-of-widget-center-widget-right-and-widget-top
+        x: root.right - self.width - dp(4)
+        y: title.center_y - self.height / 2.0
```
arlowhite: Here, I right-align and vertical-align the Reset button using property calculations and link to relevant doc. (Line length violates PEP8, but I think it's excusable for a URL)

- [ ] Should developers be encouraged to do this? Or would it be better to demonstrate using a FloatLayout with pos_hint as mentioned in the documentation?

matham:
- Yeah, label_sizing is probably better. 
- Open a issue about the max_lines, I'd have to look into that.
- I don't know what you mean by "I was thinking of creating a button that outlines the Label.textures", can you give an example?

examples/widgets/label_texture_size.py
matham: Either way is fine.

arlowhite:
Ok, renamed to label_sizing
I made the buttons modify all Labels, which is more instructive. Also, the text now has a newline so that you can see that halign does have an effect. (As you mentioned in the other PR)

Outlining the texture doesn't really demonstrate much; only the top left text has a texture not equal to the size, and that's already obvious.

- [ ] I can't figure out how to make the Spinners wider, setting width doesn't work. Maybe I should just use a Dropdown?

arlowhite:
Created a max_lines issue: #3450

arlowhite:
Figured out the Spinner problem, it inherits Button and I modified <Button> in KV.
Squashed everything to one commit. I'm done unless you have more comments.

matham:
:+1:

PR #3453, opened 2015-06-20
warn if only one of the minimum window size properties is bigger than 0

dessant:
If not both of them is bigger than 0, the restriction does not take effect. I don't know if this is by design or an sdl2 bug.

We could either do this, or just document it. Thoughts?


akshayaurora:
@dessant  :+1:

PR #3456, opened 2015-06-22
[WIP]: KV lang compiler

matham:
Implements a KV compiler. It should be complete and fully functional as far as I can tell.

To see the changes, I suggest you do `make html` (or look at the source) and look at the docs of `Builder.compile_kv`, `Builder.compile_load_string`, and the `batch_bind` docs at `kivy.tools.kvcompiler`. The last one in particular describes much of the arch changes to the KV rule application, and in particular is something we need to decide whether it's acceptable.

The main things/changes to note, other than those docs is that:
- I did not test using rebind much yet, but it should all work.
- Some parts of the code still need to be polished and documented more (mainly `KVCompiler` and `unbind_widget`).
- I changed the usage of proxy_ref so that while direct objects are never stored in bindings, whenever we refer to a object in KV, e.g. `obj: root`, the actual `root` object is saved in `obj`. I don't really see how this will lead to memory leaks and why we should use `proxy_ref` instead, as it just reduces performance. Just like in python when you do `self.obj = other_obj` we don't save the `proxy_ref` of `other_obj`, we should not do it for KV either. I'd like to understand the reasoning for why `proxy_ref` was used in the past.
- About **backward compat**: unless you made some very specific assumptions about how kv evaluates the rules that are not typical, it should all be backward compat as far as I can tell.

Other changes in the PR:
- `proxy_ref` was moved from `Widget` to `EventDispatcher`. It actually resulted in a perf boost.
- In the past, when a widget died, all the KV rules that set that widget's properties are unbound using `unbind_widget`. However, this did not happen for Canvas instructions who have properties set by KV rules. Now, they also are unbound when the instruction dies.
- Fixes kv language string recognition to recognize more string variations than before.
- Instead of applying KV rules after the widget is created depending on `__no_builder`, we always apply the rules in `__init__`. To ensure that `self.parent` is not `None`, the following syntax is supported: `Widget(parent=obj)`. And in this case, `Widget` will be added as a child of `obj`. This allows setting `parent` before any rules are applied. I think this is a better approach than before.
- What else?


tshirtman:
:coffee:

tito:
:kissing:

dessant:
I'm here to test and merge in the next few days if you have time for a rebase.

auvipy:
@matham could u plz rebase to fix merge conflicts?

KeyWeeUsr:
@dessant any progress with this? 😋

auvipy:
I think someone should take it over. and just rebase

doc/sources/images/kvrules1.gv
```diff
@@ -0,0 +1,46 @@
...
+    FloatLayout -> GridLayout;
+
+
+    MyWidgetCW [xlabel="0"];
+    Label [xlabel="1"];
+    FloatLayout [xlabel="2"];
+    GridLayout [xlabel="3"];
+
+    MyWidgetCW [shape=diamond, label=MyWidgetC, color="salmon3", style=filled];
+}
```
auvipy: no new line

kivy/uix/widget.py
```diff
@@ -288,11 +281,14 @@ class Widget(WidgetBase):
     .. versionchanged:: 1.5.0
         The constructor now accepts on_* arguments to automatically bind
         callbacks to properties or events, as in the Kv language.
+
+    ..versionchanged:: 1.9.1
```
auvipy: this should be 2.0.0 or intended version

dessant:
@KeyWeeUsr, this is one of those prs which require considerable work before it could be merged, but it's worth leaving open.

KeyWeeUsr:
I tried to rebase it, but without success. There are two files:

    kivy/tools/kvcompiler.py
    kivy/lang.py

which are old ones. The `kvcompiler.py` is a new one from this branch, so that wouldn't be a problem at all, however that **_`lang.py`_** file is an old thing, that was separated to:

    kivy/lang/__init__.py
    kivy/lang/parser.py
    kivy/lang/builder.py

Until that conflict is resolved, the PR is unusable even for testing with the latest changes in master. 😿

manthansharma:
@matham I have tired to rebase this branch with kivy/master branch. It is almost completed other than `lang.py` file as mentioned earlier by @KeyWeeUsr it has been separated to:
```
kivy/lang/__init__.py
kivy/lang/parser.py
kivy/lang/builder.py
```
So, I am thinking to rewriting changes already done in lang.py to these 3 diff. files or is there any other efficient way.

Second problem i faced is every commit is now changed to `+ manthansharma` is it a problem or will be ok.

I haven't opened a PR but you check the changes from https://github.com/kivy/kivy/compare/master...manthansharma:lang-compiler. It's completely working with the sample.kv file.

matham:
In that branch, all the `lang.py` changes has been removed, so I'm not sure how it could work with a compiled file? I don't see where those changes went to, they just disappeared.

manthansharma:
@matham yeah, i have understand the whole `lang.py` and `kvcompile.py` every line of code of it. I have some doubt in some of the methods implemented that i tried to get clarified from you but you were unavailable at that moment.

So, I am just revising my cython concept as it been a long time, then I can start rewriting the changes again to `lang.py` and other.

PR #3458, opened 2015-06-23
Recognise left and right control for select

stuartmarsden:
On Linux the scancode is either lctrl or rctrl depending on which is pressed. Only ctrl is checked for so it means ctrl select was not working on Linux (Kubuntu 15.04).


matham:
:+1:

PR #3485, opened 2015-07-06
Allow multiprocessing to work correctly with pyinstaller

brentpicasso:
This mod passes multiprocessing-fork option to sys.argv so it can be handled correctly when running under the contexty of windows pyinstaller.


brentpicasso:
cancelling PR; applying to 1.9.0 instead.

tshirtman:
I'm not sure what you mean, PRs should be against master, not latest stable.

brentpicasso:
my bad. I will re-submit PR against master. :)

dessant:
There is no need to resubmit.

brentpicasso:
Thanks; looks like the merge seems correct with 1.9 as it did with 1.8.x.

So more background on this PR:

A kivy app that uses multiprocessing on windows will fail to run if packaged with pyinstaller. This allows the environment variable to be forwarded so the app can successfully run.

PR #3494, opened 2015-07-09
Adding rotation to mtdev

tpulatha:
We needed rotation functionality for mtdev similar to the functionality of hidinput.


bj0:
I have the 2.8 capacitive pitft, and it does not work with hidinput (throws a KeyError: 'x').  mtdev works but the axes are flipped, so I needed to be able to rotate them.  With this patch + custom input line in .kivy/config.ini the touch input is working.

akshayaurora:
@tpulatha :+1: thanks.

PR #3498, opened 2015-07-11
remove super call from TreeViewNode.__init__

LogicalDash:
Calling the superclass's initializer is normally a good idea, but ``TreeViewNode`` only has ``object`` for a superclass. ``object.__init__`` doesn't do anything, and trying to call it with ``**kwargs`` was giving me ``TypeError``s.

### DIFF
--- a/kivy/uix/treeview.py
+++ b/kivy/uix/treeview.py
@@ -128,7 +128,6 @@ class TreeViewNode(object):
     def __init__(self, **kwargs):
         if self.__class__ is TreeViewNode:
             raise TreeViewException('You cannot use directly TreeViewNode.')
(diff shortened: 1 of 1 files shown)


matham:
Super is required in order to do inheritance. So this will break if someone does `class MyNode(TreeViewNode, MyOtherClass)`.

What is the contents of `kwargs` that is forwarded? And do you have example code that reproduces the problem?

rnixx:
https://github.com/kivy/kivy/pull/3498/files#diff-43568edec886077d46a1013c2787f0afR130
already raises an exception if some tries to instanciate TreeViewNode directly.

object(foo=1) raises a type error because it takes no arguments. But this never happens here.

Issue can be closed.

PR #3499, opened 2015-07-11
Added ca_file argument for custom CA validation

Cheaterman:
This is for people like me who use self-signed certificates and would get a SSL validation error without a custom CA to validate against.


tito:
Very good, thanks you!

PR #3506, opened 2015-07-15
Fix misalignment of multi-touch emu feedback

cpbotha:
Visual feedback of multi-touch emulation first touch was being placed
at coordinates uncorrected for density != 1. Fixed by factoring out
and using same scaling as for WindowBase.on_motion().

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -528,6 +528,21 @@ class WindowBase(EventDispatcher):
     '''Real size of the window ignoring rotation.
     '''
 
+    def _get_effective_size(self):
+        w, h = self.system_size
+        if platform == 'ios' or self._density != 1:
+            w, h = self.size
+
+        return w,h
+
+    effective_size = AliasProperty(_get_effective_size, None)
+    '''On density=1 and non-ios displays, return system_size, else
+    return scaled / rotated size.
+
+    Used by MouseMotionEvent.update_graphics() and WindowBase.on_motion().
+    '''
+
+
     borderless = BooleanProperty(False)
     '''When set to True, this property removes the window border/decoration.
 
@@ -942,9 +957,7 @@ class WindowBase(EventDispatcher):
                 The Motion Event currently dispatched.
         '''
         if me.is_touch:
-            w, h = self.system_size
-            if platform == 'ios' or self._density != 1:
-                w, h = self.size
+            w,h = self.effective_size
             me.scale_for_screen(w, h, rotation=self._rotation,
                                 smode=self.softinput_mode,
                                 kheight=self.keyboard_height)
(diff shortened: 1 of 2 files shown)


kivy/core/window/__init__.py
```diff
@@ -528,6 +528,21 @@ def _get_system_size(self):
...
+        w, h = self.system_size
+        if platform == 'ios' or self._density != 1:
+            w, h = self.size
+
+        return w,h
+
+    effective_size = AliasProperty(_get_effective_size, None)
+    '''On density=1 and non-ios displays, return system_size, else
+    return scaled / rotated size.
+
```
akshayaurora: .. versionadded: 1.9.1

kivy/core/window/__init__.py
```diff
@@ -942,9 +957,7 @@ def on_motion(self, etype, me):
                 The Motion Event currently dispatched.
         '''
         if me.is_touch:
-            w, h = self.system_size
-            if platform == 'ios' or self._density != 1:
-                w, h = self.size
+            w,h = self.effective_size
```
akshayaurora:
```
w,  h 
```

You need the pep8hook.
run 

```
make hook
```

from the kivy root dir.

akshayaurora: Let's use the function and get rid of the property.

kivy/core/window/__init__.py
cpbotha: My reasons for doing it this way was indeed to keep it DRY (having two different places checking the same logic is a bit fragile), and to follow the lead of the other properties.

Would you prefer that we rather just remove the property, and then rename the function to "get_effective_size()" (without underscore) and use that?

akshayaurora: This function would be used internally only and thus it makes sense to keep
it private, i.e have it start using `_`.

cpbotha:
*bump* Could you take another look at this?

(I've just noticed that there's similar buggy behaviour when handling a drop event on the window. would like to have this PR merged before I try and fix the next related issue.)

kivy/core/window/__init__.py
```diff
@@ -528,6 +528,18 @@ def _get_system_size(self):
     '''Real size of the window ignoring rotation.
     '''
 
+    def get_effective_size(self):
```
akshayaurora: There is no need for this to be exposed to the user, this is used internally only and thus should start with a `_` `_get_effective_size`.

cpbotha:
Hi there @akshayaurora !

Have renamed method with underscore. The travis checks that have failed are in kivy/config.py (lines too long) -- but my commit has not touched those files.

Please let me know if you would prefer me squashing the three commits together, ok?

akshayaurora:
@cpbotha nice job!

PR #3509, opened 2015-07-16
Smoothline problem

andnovar:
SmoothLine is not in the list of graphics that is a problem when trying to invoke it.


matham:
You need to squash all the commits into a single commit and then force push to this branch.

andnovar:
ok I have to use rebase

PR #3512, opened 2015-07-16
alpha version of the kivy backend for matplotlib

andnovar:
Kivy backend for matplotlib.


kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,850 @@
...
+
+
+class MPLKivyApp(App):
+
+    figure = ObjectProperty(None)
+    toolbar = ObjectProperty(None)
+
+    def __init__(self, **kwargs):
+        super(MPLKivyApp, self).__init__(**kwargs)
+        self.figure = kwargs['figure']
```
matham: This should not be `kwargs['figure']`, but `kwargs.get('figure')` b/c what if it was not specified? Similarly for `toolbar`.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,850 @@
...
+        self.toolbar = kwargs['toolbar']
+
+    def build(self):
+        EventLoop.ensure_window()
+        layout = FloatLayout()
+        layout.clear_widgets()
```
matham: Why are you clearing? Remove it.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,850 @@
...
+        self.figure.size_hint_y = 0.9
+        layout.add_widget(self.figure)
```
matham: What if `self.figure` or `toolbar` is `None`? You need to check and add them conditionally.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,850 @@
...
+        self.toolbar.size_hint_y = 0.1
+        layout.add_widget(self.toolbar)
+        return layout
+
+    def on_pause(self):
+        return App.on_pause(self)
```
matham: Why are these even here? Remove it. The same for `on_resume`. And you should be calling super anyway if you used it.

kivy/ext/mpl/backend_kivy.py
andnovar: I missunderstood the function I though these removes.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,850 @@
...
+from math import cos, sin, pi
+
+# try:
+#     kivy.require('1.9.0') # I would need to check which release of
+#     Kivy would be the best suitable.
+# except AttributeError:
+#     raise ImportError(
+#         "kivy version too old -- it must have require_version")
+
+_debug = True
```
matham: This should be removed. Instead, wherever you're printing debug messages use kivy logger instead with debug or info level etc.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,850 @@
...
+
+app = None
```
matham: Instead of using a global app, use `App.get_running_app()`. If it returns `None`, you know an app is not running yet.

andnovar: Ow thanks for the tip !

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,850 @@
...
+        layout.add_widget(self.toolbar)
+        return layout
+
+    def on_pause(self):
+        return App.on_pause(self)
+
+    def on_resume(self):
+        App.on_resume(self)
+
+    def on_stop(self):
```
matham: Remove this altogether. See my comment below.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,850 @@
...
+        global app
+        app = None
+
+
+def _create_App(fig_canvas, toolbar):
+    global app
+    if app is None:
```
matham: Instead of using `app` global use `App.get_running_app()`. Also, if `App.get_running_app()` is not `None` and a app is already running, you should raise an exception here for now.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,850 @@
...
+from kivy.uix.boxlayout import BoxLayout
+from kivy.uix.floatlayout import FloatLayout
+from kivy.uix.relativelayout import RelativeLayout
+from kivy.uix.popup import Popup
+from kivy.properties import ObjectProperty
+from kivy.clock import Clock
+
+import numpy as np
+import io
+from _functools import partial
```
matham: Why this underscore `_functools`?

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,850 @@
...
+        if _debug:
+            print("Starting up Kivy Application")
+        app = MPLKivyApp(figure=fig_canvas, toolbar=toolbar)
+
+
+class RendererKivy(RendererBase):
+    """
+    The renderer handles drawing/rendering operations.
+
+    This is a minimal do-nothing class that can be used to get started when
```
matham: This part of the docs will need to be replaced and a abstract description on how the renderer is handled for the kivy backend.

kivy/ext/mpl/backend_kivy.py
matham: It does remove all. You just instantiated the instance so there's nothing to remove so there's no reason to call it.

kivy/ext/mpl/backend_kivy.py
andnovar: matham sorry what is your advice here ? should I create an exception or return an empty object ?

kivy/ext/mpl/backend_kivy.py
andnovar: right, I don't know what I was thinking when I did this.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,832 @@
...
+    writing a new backend. Refer to backend_bases.RendererBase for
+    documentation of the classes methods.
+    """
+    def __init__(self, dpi, widget):
+        self.dpi = dpi
+        self.widget = widget
+        #  Can be enhanced by using TextToPath matplotlib, textpath.py
+        self.mathtext_parser = MathTextParser("Bitmap")
+        self.list_goraud_triangles = []
```
matham: I would prefer that any instance variables you create are declared first at class level so please init these in the class with e.g. `_clipd = None`.

kivy/ext/mpl/backend_kivy.py
andnovar: should I enable the kivy.require ?? I'll remove the _debug

matham: Sure, do require('1.9.1') for now.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,832 @@
...
+        self._clipd = {}
+        self.clip_rectangles = []
+
+    def contains(self, widget, x, y):
```
matham: In what coordinate system is the x, y?

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,832 @@
...
+            l, b, w, h = bbox.bounds
+        else:
+            l = 0
+            b = 0,
+            w = self.widget.width
+            h = self.widget.height
+        h, w = im.get_size_out()
+        rows, cols, image_str = im.as_rgba_str()
+        texture = Texture.create(size=(w, h))
+        texture.blit_buffer(image_str, colorfmt='rgba', bufferfmt='ubyte')
```
matham: One of the big issues is that instructions need to be cached and re-used rather than created every time. But we should leave that for much later.

andnovar: what about when resizing ? or just a specific instruction is changed ? But I understand the point.

matham: We should talk about this later. For now just focus on the bugs.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,834 @@
...
+            plot_text.text = "{}".format(s)
+            if prop.get_style() == 'italic':
+                plot_text.italic = True
+            if weight_as_number(prop.get_weight()) > 500:
+                plot_text.bold = True
+            plot_text.refresh()
+            with self.widget.canvas:
+                if isinstance(angle, float):
+                    PushMatrix()
+                    Rotate(angle=angle, origin=(x, y))
```
matham: This rotation does not work properly. Try in test_plt to do ```
ax.set_yticklabels(('Ahh', '-----------G1-------', 'G2', 'G3', 'G4', 'G5', 'G5', 'G5', 'G5'), rotation=90)
ax.set_xticklabels(('-----------G1-----------', 'G2', 'G3', 'G4', 'G5'), rotation=90)``` and compare it to how it looks for e.g. agg vs kivy. It needs to rotate around the center. Also, if labels sit on top of each other they seem to be obscuring each other. We should not be creating labels for every text object, but rather draw directly on the canvas.

andnovar: I'll check it out.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,831 @@
...
+            h = self.widget.height
+        h, w = im.get_size_out()
+        rows, cols, image_str = im.as_rgba_str()
+        texture = Texture.create(size=(w, h))
+        texture.blit_buffer(image_str, colorfmt='rgba', bufferfmt='ubyte')
+        with self.widget.canvas:
+            Color(1.0, 1.0, 1.0, 1.0)
+            Rectangle(texture=texture, pos=(x, y), size=(w, h))
+
+    def draw_gouraud_triangle(self, gc, points, colors, transform):
```
matham: Non of this is implemented. What is going on here?

andnovar: Is not implemented in any of the gui backends, but it is on the image ones, I did not found any example that uses it neither.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,831 @@
...
+        w = ftimage.get_width()
+        h = ftimage.get_height()
+        texture = Texture.create(size=(w, h))
+        texture.blit_buffer(ftimage.as_rgba_str(), colorfmt='rgba',
+                            bufferfmt='ubyte')
+        texture.flip_vertical()
+        with self.widget.canvas:
+            Rectangle(texture=texture, pos=(x, y), size=(w, h))
+
+    def draw_markers(self, gc, marker_path, marker_trans, path,
```
matham: Mpl says that for optim reasons `draw_markers`, `draw_path_collection`, and `draw_quad_mesh` need to be implemented, yet none of them are. No wonder it's so slow.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,831 @@
...
+            plot_text = CoreLabel(font_size=prop.get_size_in_points(),
+                            font_name=prop.get_name())
+        plot_text.text = str(s.encode("utf-8"))
+        plot_text.refresh()
+        return plot_text.texture.size[0], plot_text.texture.size[1], 1
+
+    def new_gc(self):
+        return GraphicsContextKivy(self.widget)
+
+    def points_to_pixels(self, points):
```
matham: This needs to do the actual conversion.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,831 @@
...
+
+    _capd = {
+        'butt': 'square',
+        'projecting': 'square',
+        'round': 'round',
+    }
+
+    def __init__(self, renderer):
+        #super(GraphicsContextBase, self).__init__()
+        GraphicsContextBase.__init__(self)
```
matham: You **need** to use super everywhere rather than just calling the parent class directly.

andnovar: Some super breaks, that's why I kept leaving them for later.

kivy/ext/mpl/backend_kivy.py
andnovar: Not a problem. I'll add them this weekend.

kivy/ext/mpl/backend_kivy.py
matham: You should probably ask on the mpl mailing list then for an example that uses it, or why it there.

kivy/ext/mpl/backend_kivy.py
matham: When you hit a bug, you need to fix it right there or open a issue for it. You cannot leave stuff like this in the code.

kivy/ext/mpl/backend_kivy.py
andnovar: screenshot from 2015-08-08 23 17 48
screenshot from 2015-08-08 23 18 03
screenshot from 2015-08-08 23 18 28

andnovar: matham I get the same result in different backends. The rotation is not from the center for the look of it is right, and the superposition would happen anyway if we don't use CoreLabel.

kivy/ext/mpl/backend_kivy.py
andnovar: This is solved but mpl 1.4.3 class definition is old. It is fixed on 1.5.1.

kivy/ext/mpl/backend_kivy.py
matham: KIvy:

kivy

TkAgg:

agg

matham: I included both an x and y axis rotated label in the example code I gave you because yes, on the y-axis it looks the same, but on the x-axis it looks different. The x-axis needs to be rotated around the center, or the end, actually. Or maybe we need to figure out what mpl expects to rotate it around. You need to ask on their mailing list if you cannot find out for sure.

andnovar: matham sorry but this is the same result it is just that when a label is too big in mplkivy is not shown on screen, which is the case for G1

dessant:
There seems to be a white border around the plot, maybe a pos offset issue. https://i.imgur.com/UuqRtSX.png

@andnovar please fix the travis error.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,1013 @@
...
+            layout.add_widget(self.toolbar)
+        return layout
+
+
+def _create_App(fig_canvas, toolbar):
+    '''Method to instantiate a MPLKivyApp.
+    '''
+    app = App.get_running_app()
+    if app is None:
+        app = MPLKivyApp(figure=fig_canvas, toolbar=toolbar)
```
matham: This doesn't make sense here. You're creating a app here, but not actually doing anything with it. Then you run the current app, in show's mainloop. At that point, the app that is run could be any app, even one created by the user later. Instead, you need to create and run the app in mainloop.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,1013 @@
...
+            figManager.canvas.draw_idle()
+
+
+class Show(ShowBase):
+    '''mainloop needs to be overwritten to define the show() behavior for kivy
+       framework.
+    '''
+    def mainloop(self):
+        app = App.get_running_app()
+        if app:
```
matham: Here's where you should create the app and run it if one does not exist.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,1013 @@
...
+        self.figure = figure
+        super(FigureCanvasKivy, self).__init__(figure=self.figure, **kwargs)
+        self.focus = True
+
+    def draw(self):
+        '''Draw the figure using the KivyRenderer
+        '''
+        self.clear_widgets()
+        self.canvas.clear()
+        renderer = RendererKivy(self)
```
matham: Why are you creating a new renderer for every draw? You should create it once and resuse it for the same canvas.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,1013 @@
...
+    .. versionadded:: 1.9.1
+    '''
+
+    def __init__(self, figure, **kwargs):
+        Window.bind(mouse_pos=self._on_mouse_pos)
+        self.bind(size=self._on_size_changed)
+        self.inside_figure = True
+        self.figure = figure
+        super(FigureCanvasKivy, self).__init__(figure=self.figure, **kwargs)
+        self.focus = True
```
matham: Why are you focusing the widget by default? A widget should not have focus automatically just be being created.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,1013 @@
...
+        self.figure.draw(renderer)
+
+    def on_touch_down(self, touch):
+        '''Kivy Event to trigger the following matplotlib events:
+           `motion_notify_event`, `scroll_event`, `button_press_event`,
+           `enter_notify_event` and `leave_notify_event`
+        '''
+        if super(FigureCanvasKivy, self).on_touch_down(touch):
+            return True
+        FigureCanvasBase.motion_notify_event(self, touch.x,
```
matham: Do you have actual tests showing these coordinates are correct. I'm suspicious that the coordinates you're passing will not be what mpl events are expecting. They are in kivy coordinates. Also, the inside figure logic seems wrong and doesn't seem to work in test_backend for example.

kivy/ext/mpl/backend_kivy.py
andnovar: I followed the convention of the other backends my bad, I'll change it

kivy/ext/mpl/backend_kivy.py
andnovar: I tried with the examples the inside figure and it works, however having more than 1 widget probably won't work I need to change coordinates here as well.

kivy/ext/mpl/backend_kivy.py
andnovar: if not focused it does not receive the keyboard events by default.

matham: Yes, that's the point. Imagine you start off with a textinput as well as a mpl widget. Say you want the textinput to have focus, but then the mpl widget will steal the focus. The better way is to not set focus and instead by the user clicking on the widget it gets focus - that's how focus works in kivy.

kivy/ext/mpl/backend_kivy.py
andnovar: Ok sorry but I was looking at the code and I need this App to be instantiated with the figurecanvas and the toolbar. I can not see a way I can pass these objects directly to the Show()

kivy/ext/mpl/backend_kivy.py
andnovar: The reason is in the other comment.

kivy/ext/mpl/backend_kivy.py
andnovar: http://pastebin.com/upwvZ4B0 You can try this. When I was trying this I found out that events are not triggered when using kivy and not plt.

kivy/ext/mpl/backend_kivy.py
andnovar: Ok I see the point now. Changed.

kivy/ext/mpl/backend_kivy.py
matham: That seems to work for the most part (except for not getting a leave event when leaving the window altogether). Why isn't e.g. test_backend working?

matham: test_backend doesn't work but test_plt works. Also you seem to be setting `inside_figure` to False/True right after you have a notify event, but that isn't consistent with the variable name.

kivy/ext/mpl/backend_kivy.py
```diff
@@ -0,0 +1,1013 @@
...
+        self.blitbox = bbox
+
+    filetypes = FigureCanvasBase.filetypes.copy()
+    filetypes['png'] = 'Portable Network Graphics'
+
+    def print_png(self, filename, *args, **kwargs):
+        '''Call the widget function to make a png of the widget. It does a
+           one second delay to allow the widget to be completed.
+        '''
+        Clock.schedule_once(partial(self._print_image, filename), 1)
```
matham: Why is it delayed? What if the widget changed in the meantime? I don't think this should be delayed.

kivy/ext/mpl/backend_kivy.py
andnovar: yes the events work. I tried creating two widgets with the same code and it got confused on the axes_enter and axes_leave event. I am checking that. I just found out that the events need to be connected after the widget is instantiated if not it does not work due to the fact that there is nothing to map to. http://pastebin.com/bCfCwBNW <-- I am changing the coordinates in the events.

andnovar: The inside_figure is just to say I am already inside so don't keep calling enter_notify, do you think enter_figure is better name ?

matham: `enter_figure` is not better because you're also using it for leaving. Something like `entered_figure` may be better.

matham: If you look at test_backend vs test_plt both connect before the figure is created. So that's not the real issue. I think it's more that one uses show and create the app while the other uses it differently.

kivy/ext/mpl/backend_kivy.py
andnovar: If not delayed it does not have enough time to export a complete render.

andnovar: specifically when you do a draw() and immediately after a savefig()

matham: That doesn't explain it. These are all blocking operation, how can there not be enough time?

matham:
Moved to https://github.com/kivy-garden/garden.matplotlib.

PR #3515, opened 2015-07-17
Fix for "disable_multitouch" in config

angryrancor:
This is a fix for interaction on desktop systems w/no multitouch capability.  Current code does not allow disabling of multitouch and creating the "Red Circle" on right click.

Requires the following config change to take effect:
* Edit /Applications/Kivy.app/Contents/Resources/.kivy/config.ini to add "disable_multitouch" to input/mouse.

So the config file will then read (if changed from default):
[input]
mouse = mouse,disable_multitouch

Credit goes to Bill Janssen and this thread from the Google Group: https://groups.google.com/forum/#!topic/kivy-users/6KuO-DaBNIE

### DIFF
--- a/kivy/input/providers/mouse.py
+++ b/kivy/input/providers/mouse.py
@@ -199,7 +199,7 @@ class MouseMotionEventProvider(MotionEventProvider):
         cur.is_double_tap = is_double_tap
         self.touches[id] = cur
         if do_graphics:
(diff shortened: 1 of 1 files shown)


angryrancor:
This actually works in master, apologies for faulty testing.  Retracting request.

PR #3517, opened 2015-07-18
Small fixes to CompoundSelectionBehavior.

matham:
Required for https://github.com/kivy-garden/garden.recycleview/pull/11.


kivy/uix/behaviors.py
```diff
@@ -1090,6 +1093,12 @@ def deselect_node(self, node):
     defaults to 0.
     '''
 
+    nodes_order_reversed = BooleanProperty(True)
```
akshayaurora: Internal props should be preceeded with `_` ?

kivy/uix/behaviors.py
```diff
@@ -1373,6 +1382,12 @@ def get_selectable_nodes(self):
         '''
         return self.children
 
+    def get_index_of_node(self, node, selectable_nodes):
```
akshayaurora: if these are internel, let's precede them with `_`.

matham: It's not truly internal, they are meant for the person that overwrites these classes, but not for the end user. We have a lot of these types of methods which are just labeled internal, but don't start with underscore. They are meant to be overwritten and and do belong in the proper external API, but, they are not meant to be used directly by the user.

The same for `nodes_order_reversed`.

PR #3518, opened 2015-07-19
Re-parse / re-render svg tree if default color is changed

brentpicasso:
* allow setting of current_color via a new color property
* this will re-trigger a re-parse and re-render of the XML DOM representing the SVG tree
* some clean up of logging

If an svg file specifies 'currentColor' as a color for stroke or fill, the Svg renderer will attempt to use current_color as a default, if set.  This patch extends this concept by allowing the color property to be updated during runtime, allowing the default color to be changed on the fly.

### DIFF
--- a/kivy/graphics/svg.pyx
+++ b/kivy/graphics/svg.pyx
@@ -574,6 +574,9 @@ cdef class Svg(RenderContext):
     '''Set the default color.
 
     Used for SvgElements that specify "currentColor"
+
+    .. versionadded:: 1.9.1
+
     '''
     property color:
         def __set__(self, color):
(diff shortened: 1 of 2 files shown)


brentpicasso:
Example of effect: https://youtu.be/eaYbPqEIpVg

kivy/graphics/svg.pyx
```diff
@@ -568,14 +571,24 @@ cdef class Svg(RenderContext):
             return self._anchor_y
 
 
+    '''Set the default color.
+
+    Used for SvgElements that specify "currentColor"
+    '''
```
tito: Can you add the tag:

```
.. versionadded:: 1.9.1
```

?

Thanks you :)

brentpicasso: done!

tito:
Thanks!

brentpicasso:
\o/  My pleasure. Thank you for making an amazing framework.

PR #3519, opened 2015-07-19
Docs: New Linux install instructions for Virtualenv and SDL2

ernstp:
Virtualenv now defaults to using --no-site-packages and that's how it
should be used, contain all software in the virtualenv.
Also there's no need to do a lot of wierd things to the users system.
Recommend users to install the exact version of Cython they need
with virtualenv.
No need to keep instructions for ancient Ubuntu versions (11.10).

### DIFF
--- a/doc/sources/installation/installation-linux.rst
+++ b/doc/sources/installation/installation-linux.rst
@@ -142,60 +142,122 @@ Gentoo
    `spell: Standard flag, provide enchant to use spelling in kivy apps.`
 
 
-Using software bundles ( also known as tarballs )
-=================================================
-
-*Providing dependencies*
-~~~~~~~~~~~~~~~~~~~~~~~~
+*Installation in a Virtual Environment*
+=================================================================
 
 
-General
+Common dependencies
+~~~~~~~~~~~~~~~~~~~
+
+
+Cython
 -------
 
-The following software is needed, even if your distribution is not listed above:
 
-- `Python >= 2.7 and Python < 3 <http://www.python.org/>`_
-- `PyGame <http://www.pygame.org/>`_
-- `PyEnchant <http://packages.python.org/pyenchant/>`_
-- `gst-python <http://gstreamer.freedesktop.org/modules/gst-python.html>`_
-- `Cython == 0.20 <http://cython.org/>`_
+Different versions of Kivy have only been tested up to a certain Cython version.
+It may or may not work with a later version.
 
-We prefer to use a package-manager to provide these dependencies.
+========   =============
+Kivy       Cython
+========   =============
+1.8        0.20.2
+1.9        0.21.2
+1.9.1      0.23
+========   =============
 
 
-Ubuntu
-------
+*System dependencies with modern SDL2*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+Ubuntu example
+------------------------------
+
 
 ::
 
-    $ sudo apt-get install \
-        pkg-config \
-        python-setuptools \
-        python-pygame \
-        python-opengl \
-        python-gst0.10 \
-        python-enchant \
-        gstreamer0.10-plugins-good \
-        python-dev \
+    # Install necessary system packages
+    sudo apt-get install -y \
+        python-pip \
         build-essential \
-        libgl1-mesa-dev \
-        libgles2-mesa-dev \
-        cython
+        mercurial \
+        git \
+        python \
+        python-dev \
+        ffmpeg \
+        libsdl2-dev \
+        libsdl2-gfx-dev \
+        libsdl2-image-dev \
+        libsdl2-mixer-dev \
+        libsdl2-net-dev \
+        libsdl2-ttf-dev \
+        libportmidi-dev \
+        libswscale-dev \
+        libavformat-dev \
+        libavcodec-dev \
+        zlib1g-dev
 
-*Upgrade Cython ( <= Oneiric [11.10] )*
 
-:Using Cython's daily PPA: ::
+Installation
+------------
 
-    $ sudo add-apt-repository ppa:cython-dev/master-ppa
-    $ sudo apt-get update
-    $ sudo apt-get install cython
 
-.. ``
+::
 
-:Using PIP: ::
+    # Make sure Pip, Virtualenv and Setuptools are updated
+    sudo pip install --upgrade pip virtualenv setuptools
 
-    $ sudo apt-get install python-pip
-    $ sudo pip install --upgrade cython
+    # Create a virtualenv
+    virtualenv --no-site-packages kivyinstall
+
+    # Enter the virtualenv
+    . kivyinstall/bin/activate
+
+    # Use correct Cython version here
+    pip install Cython==0.21.2
+
+    # Install stable version of Kivy into the virtualenv
+    pip install kivy
+    # For the development version of Kivy, use the following command instead
+    # pip install git+https://github.com/kivy/kivy.git@master
+
+Python 3
+--------
+
+
+If you want to use Python 3 you install "python3" and "python3-dev" and then pass
+"-p python3" to virtualenv.
+
+
+*Dependencies with legacy PyGame*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+Ubuntu example
+------------------------------
+
+
+::
+
+    # Install necessary system packages
+    sudo apt-get install -y \
+        python-pip \
+        build-essential \
+        mercurial \
+        git \
+        python \
+        python-dev \
+        ffmpeg \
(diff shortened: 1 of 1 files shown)


rnixx:
it makes no sense to have a section "py2.7 with pygame" and "py3.3 with sdl2". actually i use "py2.7 with sdl2" which is not mentioned.

Better would be to distinguish between dependencies, legacy (pygame) and recent (SDL2) install, and recent install might be done with py2.7 or py3.3

further i compiled kivy with cython 0.22.1 and it seem to work.

rnixx:
also i think its not very clever to document dependencies for specific ubuntu versions. this is basically dependency related and here it would be more appropriate to have dependency documentation with <14.04 and >=14.04

akshayaurora:
site-packages were necessary with pygame backend, pygame refuses to compile completely if you don't enable that. We are moving away from pygame to sdl2 so this might make sense but it would be better to document that you need site-packages for compiling pygame.

rnixx:
@akshayaurora which system site packages are needed to compile pygame? imho it should be possible to install them isolated in a virtualenv as well...

ernstp:
>Better would be to distinguish between dependencies, legacy (pygame) and recent (SDL2) install, and recent install might be done with py2.7 or py3.3

Sounds good. Actually I see that python 3.4 is default on 14.04.

>further i compiled kivy with cython 0.22.1 and it seem to work.

Yes it does but since you get a big warning I felt that I couldn't put that in the documentation.
(And IMHO I think Kivy could just have proper PyPI dependencies...)

My thought is to cover a number of common scenarios which includes most bits and then people can use that to make new combinations. Perhaps we could add Virtualenv instructions for Fedora/Arch/etc also.

ernstp:
I think the "Using software bundles ( also known as tarballs )" section could be removed also, it would be better use virtualenv there.

ernstp:
Regarding Python 2.7 vs 3.3, I need to use something in the examples, and it doesn't make sense to make a new example for each python version. So I just picked to different, that should give people enough info to create their own combinations.

dessant:
I propose to not mention python versions at all in the section titles, nor the install commands, just use `python`, `python-dev`, `pip` etc.

dessant:
@rnixx i think pygame itself cannot be installed in a venv.

rnixx:
@dessant these are the same system deps you need to install to system python. i installed the system deps, created a venv and installed cython and numpy and pygame. nothing special

rnixx:
and +1 not to use specific python version. one sentence clarifying this should be enough

ernstp:
Sorry for being a bit slow here, let's try to get this in!

akshayaurora:
@ernstp nice work :+1:

mpizenberg:
Hi, I tried to follow the instructions to install kivy 1.9.0 with pip in a virtualenv (python3) and somehow it uses pygame instead SDL2.

```
[INFO   ] [Logger      ] Record log in /home/matthieu/.kivy/logs/kivy_15-08-27_11.txt
[INFO   ] [Kivy        ] v1.9.0
[INFO   ] [Python      ] v3.4.0 (default, Jun 19 2015, 14:20:21) 
[GCC 4.8.2]
[INFO   ] [Factory     ] 173 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_gif, img_pygame, img_pil (img_ffpyplayer ignored)
[INFO   ] [Text        ] Provider: pygame
[INFO   ] [Window      ] Provider: pygame(['window_egl_rpi'] ignored)
```

Is there a way to force using SDL2 instead of pygame ?

ernstp:
Did you include --no-site-packages? Needed if you have old virtualenv.

mpizenberg:
my version of virtualenv is 13.0.1 and I had to reinstall cython (otherwise kivy won't install) pillow and pygame (otherwise the kivy app won't start) with pip, so I think it is without site-packages (but did not explicitly used --no-site-packages). Is there a way I could check ?

rnixx:
@mpizenberg check virtualenv --help. --no-site-packages is already the default in 13.x. But i wonder what you installed at all. If you create a venv without pygame installed, it's not possible that kivy chooses pygame as provider.

Can you import pygame when running your venv python?

ernstp:
You should not install pygame in this virtualenv.

akshayaurora:
@mpizenberg kivy's setup will try to find SDL2 libs and only if it can't find them will it try using pygame, make sure you have SDL2 libs installed and accessible as mentioned in the documentation. I have updated the doc to explicitly mention this.

mpizenberg:
I uninstalled pygame and kivy, then reinstalled kivy and when I launched the application I have this message :
```
[INFO   ] [Window      ] Provider: pygame(['window_egl_rpi'] ignored)
[CRITICAL] [Window      ] Unable to find any valuable Window provider at all!
```

Maybe the problem is that if pygame is installed systemwide, it is usable in a virtual env (I thought it should not if using --no-site-packages) because when I do this : 
```
(cv3)matthieu@nus-idmi-pc:[~]$ pip uninstall pygame
Cannot uninstall requirement pygame, not installed
(cv3)matthieu@nus-idmi-pc:[~]$ python
Python 3.4.0 (default, Jun 19 2015, 14:20:21) 
[GCC 4.8.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pygame
>>>
```

It seems that pygame is still importable in virtualenv

mpizenberg:
@akshayaurora , yes every lib is installed system-wide : 
```
(cv3)matthieu@nus-idmi-pc:[~]$ sudo apt-get install -y \
>     python-pip \
>     build-essential \
>     mercurial \
>     git \
>     python \
>     python-dev \
>     ffmpeg \
>     libsdl2-dev \
>     libsdl2-gfx-dev \
>     libsdl2-image-dev \
>     libsdl2-mixer-dev \
>     libsdl2-net-dev \
>     libsdl2-ttf-dev \
>     libportmidi-dev \
>     libswscale-dev \
>     libavformat-dev \
>     libavcodec-dev \
>     zlib1g-dev
[sudo] password for matthieu: 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
build-essential is already the newest version.
python is already the newest version.
python-dev is already the newest version.
zlib1g-dev is already the newest version.
libportmidi-dev is already the newest version.
libsdl2-gfx-dev is already the newest version.
libsdl2-image-dev is already the newest version.
libsdl2-mixer-dev is already the newest version.
libsdl2-net-dev is already the newest version.
libsdl2-ttf-dev is already the newest version.
git is already the newest version.
libsdl2-dev is already the newest version.
mercurial is already the newest version.
python-pip is already the newest version.
ffmpeg is already the newest version.
libavcodec-dev is already the newest version.
libavformat-dev is already the newest version.
libswscale-dev is already the newest version.
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

rnixx:
@mpizenberg did you install pygame with your system's package manager?

try to apt-get remove --purge python-pygame

rnixx:
@mpizenberg and also python-numpy and cython if you want to go the virtualenv only approach

mpizenberg:
@rnixx cython and numpy are installed with pip in virtualenv only.
But for pygame I used the following : 
```
(cv3)matthieu@nus-idmi-pc:[~]$ sudo apt-get build-dep python-pygame
(cv3)matthieu@nus-idmi-pc:[~]$ pip install 'Pygame == 1.9.1release' --allow-external Pygame --allow-unverified Pygame
```

rnixx:
@mpizenberg then kick your system pygame and try again ;)

mpizenberg:
@rnixx I find no way to find where this pygame is installed

It seems it is not with apt-get : 
```
(cv3)matthieu@nus-idmi-pc:[~]$ sudo apt-get remove --purge python-pygame
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package 'python-pygame' is not installed, so not removed
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

It is not with pip in the virtualenv either apparently : 
```
(cv3)matthieu@nus-idmi-pc:[~]$ pip uninstall pygame
Cannot uninstall requirement pygame, not installed
```

Though it is somewhere : 
```
(cv3)matthieu@nus-idmi-pc:[~]$ python
Python 3.4.0 (default, Jun 19 2015, 14:20:21) 
[GCC 4.8.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> import pygame
>>> sys.modules
{ ... , 'pygame': <module 'pygame' (namespace)>, ... }
```

?

rnixx:
@mpizenberg and what about dpkg -r pygame (or python-pygame)?

akshayaurora:
@mpizenberg you have not given your compile log, come over to #kivy on irc.freenode.net and we can help you solve your problem, it's not related to this issue so I am closing this for any further discussion.

Issue #3474, opened 2015-06-29
Line with dynamic dash on/off length


andnovar:
A new feature could be added to Line, so a Line can be created from a list of numbers that define how the line will be rendered. The list would be on an on/off configuration. For instance we can have [t1, z1, t2, z2, t3, z3, ... ] where the set of tn values can be the segments in which the line will be rendered and the zn values the one that the segments won't be rendered.
##

matham:
Can you give an example of how that line would then be instantiated? I'm not sure I fully understand what you mean.

andnovar:
a = Line(points=points, dash_list=[2,4,1,6,8,2]) This would create a line with the first dash length 2 then an offset of 4 then a dash lenght of 1 then an offset of 6 and so on.

matham:
Implemented [here](https://github.com/kivy/kivy/pull/6269).

PR #3524, opened 2015-07-22
Line with multiple values for length and offset so dashes can be set …

andnovar:
…with a list of numbers [on, off, on, off, ...] where on will be length and off offset.

This was discussed in this issue and extends kivy to support an mpl feature.

https://github.com/kivy/kivy/issues/3474

screenshot from 2015-07-22 02 03 16

This line was created by the following list --> [20, 10, 100, 20, 10, 20]


tito:
Continued and merged in #6269

PR #3526, opened 2015-07-22
new MasterSlave class added

debeshm:
Hope this new class will be beneficial for kivy application developers in a wide variety of ideas.
The usage of this class is in its documantation

### DIFF
--- a/kivy/masterslave.py
+++ b/kivy/masterslave.py
@@ -1,0 +1,200 @@
+'''
+MasterSlave
+===========
+
+:class:`MasterSlave` is used to make a widget follow or follow opposite
+to another widget in terms of size or path it covers on screen. The
+widget which is followed can be reffered to as 'master' and the
+following widget can be referred to as 'slave'.
+
+To use MasterSlave, follow these steps:
+
+    * Setup a MasterSlave object on two widgets(master and slave)
+    * Call the desired follow methods of MasterSlave class with positive
+      arguments to follow in direction of master and with negative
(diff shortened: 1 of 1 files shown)


matham:
This probably belongs in kivy-garden rather than in kivy directly.

debeshm:
matham : thanks for the info

how am I able to know??

matham:
See http://kivy-garden.github.io/

tshirtman:
Also, it seems remotely similar to garden.magnet (unless i'm missing the point)

debeshm:
tshirtman: magnet is like a container but MasterSlave is a normal class which can make a widget follow another widget or follow opposite to another widget at a slower or faster rate

debeshm:
it can be used in many applications or games....and it does not only follows animation but any type of movement...may be due to dragging by user

tshirtman:
Ok, I understand better, no worries.

debeshm:
so is this contribution suitable here or is it suitable for kivy-garden?

tshirtman:
I think it's better in a garden, it's not really a widget, and not a common use case for kivy applications.

debeshm:
ok..all ri8
Thanks....
now I need to know about garden as I am not experienced with it

tshirtman:
The link @matham gave should give you all info, but please ask if you have any question :).

debeshm:
sure

akshayaurora:
closed for garden.

PR #3534, opened 2015-07-25
small fixes to codeinput themes

aron-bordin:
## small fixes related with themes

* updates the text background color from theme
* pygments uses a bbcode format, so some themes tries to use [u] tag. removes this extra-tags


akshayaurora:
@aron-bordin :+1:

PR #3539, opened 2015-07-27
Revert "Couple of Corrections.."

gouzouni:
Reverts kivy/kivy#3443

I'll revert it and pull just the first one which is correct. The other could crash the App.


akshayaurora:
fixed in 12a54b70dcda7cffa2d4305d6d93efbf8562a707

PR #3547, opened 2015-08-02
Change default font to Roboto

dessant:
cc @akshayaurora


dessant:
The hinting issue was fixed in the latest version of roboto, so light hinting is not required anymore.

hint

@akshayaurora we could keep the default hinting `normal` in kivy, but the discussed config option would still be great.

PR #3552, opened 2015-08-04
add display_border property to BorderImage

### LINKED ISSUE #3551: BorderImage should support border size, like the css border
Currently, pixels are always represented 1:1, the apparent size of the border is determined by the number of pixels taken from the texture.

as an example
http://border-image.com

open the Border size section, and change the values, it's currenly impossible to do that with `BorderImage` in kivy, and it's sometime missing.

the values could either be added to the "border" attribute (and made optional, if only 4 values are here, they would be used for that too), or added to a new optional property, whose name can't be `size` of course (already taken), and `border_size` sounds bad, since there is already `border`, `display_size` maybe?

kived:
Fixes #3551

### DIFF
--- a/kivy/graphics/vertex_instructions.pyx
+++ b/kivy/graphics/vertex_instructions.pyx
@@ -851,6 +851,7 @@ cdef class BorderImage(Rectangle):
 
     '''
     cdef list _border
+    cdef list _display_border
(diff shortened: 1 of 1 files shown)


tshirtman:
:+1:

PR #3555, opened 2015-08-04
Move colormap to utils.

matham:
Moves colormap from svg to utils. It doesn't make sense to leave it in svg and import svg from utils which leads to circular imports anyway.


kivy/utils.py
```diff
@@ -136,6 +136,157 @@ def is_color_transparent(c):
...
+    'turquoise': '#40e0d0',
+    'violet': '#ee82ee',
+    'wheat': '#f5deb3',
+    'white': '#ffffff',
+    'whitesmoke': '#f5f5f5',
+    'yellow': '#ffff00',
+    'yellowgreen': '#9acd32',
+}
+
+colormap = {k: get_color_from_hex(v) for k, v in hex_colormap.items()}
```
dessant: This could be a reified function disguised as a property, to build the dict only if it is used.

dessant:
:+1:

PR #3566, opened 2015-08-06
Fix KIVY_VIDEO provider list

Cheaterman:
FFmpeg doesn't seem to work, while ffpyplayer does, so I guess the docs need updating?


dessant:
:+1:

Issue #3411, opened 2015-06-09
multitouch_on_demand doesn't work with 1.9.0 on OS X


wjanssen:
This is 1.9.0 on OS X 10.7.

I have this in my ~/.kivy/config.ini:

[input]
mouse = mouse,multitouch_on_demand

But when I right-click I get the dreaded red dot, and the info about profiles and button is this:

['pos', 'button', 'multitouch_sim'] right

I thought I got to control whether "multitouch_sim" is in the profile?
##

janssen:
This analysis is not right.  multitouch_on_demand will always put multitouch_sim in the profile.  However, the red dot should not be drawn.

KeyWeeUsr:
Probably fixed by #5396

tito:
Just tested with 8f48a117e6fe7936e935234807930cf9bcaa8f22 / master, i got multitouch sim on touchtracer, but not the red dot. So yes probably fixed by #5396.

PR #3573, opened 2015-08-11
remove red dot appropriately, issue #3411

janssen:
The red ellipse should only be drawn when either multitouch_disabled is not True, or when multitouch_on_demand is not True (because in that case we don't know if multitouch is desired or not until we get the release event).


akshayaurora:
@janssen :+1:

PR #3578, opened 2015-08-14
Context has lr_shader, not l_shader

Cheaterman:
Probably historical reasons or missed during a refactor.


akshayaurora:
@Cheaterman :+1:

PR #3580, opened 2015-08-15
Make sure negative widget insertions are possible

Davideddu:
With the current implementation it's not possible to add a widget at the beginning of the carousel. If the user passes -1, it's added at position -1 because -1 != 0, but when the user passes 0, it's appended, which is the same thing!


akshayaurora:
@Davideddu thanks

Davideddu:
@akshayaurora Sure, no problem! ;Thanks for developing Kivy ;)

PR #3599, opened 2015-08-23
Update OpenCV's Camera provider to use cv2 module

hctsp:
**Abstract :** The `cv` module is deprecated in OpenCV 3.0.0, and thus completely unavailable for Python 3.

**To reproduce :** Run Kivy's Camera example (`examples/camera/main.py`) using Python 3 with a working camera
**Output :** No camera provider
**Expected output  :** camera output

**Cause :** `cv` module not available in Python 3
**Solution :** modify the OpenCV CameraBase implementation in `camera_opencv.py` to use `cv2` instead of `cv` and `cv.highgui`.

**Question :** Maybe it would be better to implement this as a new provider, rather than replacing the current module ? I don't know if Kivy uses the same codebase for Python 2 and Python 3, in which case it would make sense. Otherwise, it is my understanding that the current OpenCV camera provider is totally unusable using Python 3. I'll be happy to fix the PR to make this a new provider (basically just renaming the file and adding it in `__init__.py`.


hctsp:
Tests are passing in all Python versions, but the OpenCV3 camera provider is actually not compatible with Python 2. Closing the PR for now, as more work is needed for that.

I think I'm actually going to refactor the CameraBase class and its current implementations, as I feel it's needed to simplify provider implementation and to hide details that don't need to be exposed to camera providers, such as scheduling frame updates, or handling camera state (acquire camera if started with no camera open, unschedule updates when releasing camera (if asked by users, upon exception, or when no frame error threshold is reached), ask for camera reconfiguration when resolution is changed, ask for rescheduling when FPS is changed, and many other things).

I'm not sure I'll be able to reimplement the PyGST+Android+VideoCapture+AVFoundation providers ,for Python 2 and 3, on my own in reasonable time, especially to have it working in Python 2 and 3, so there is a big chance that this code unfortunately will never get merged in Kivy's official branch.

I'll continue to provide an updated repoistory (and patches, if needed), for OpenCV3 camera support, and thus, Camera support on Linux for Python 3 environments.

PR #3601, opened 2015-08-27
accept KIVY_SDL2_PATH for darwin builds

poke1024:
I'm building Kivy on OS X using homebrew (and python3) and am including SDL2 as homebrew libs (e.g. myhomebrew/lib/libSDL2.a), not as OS X frameworks. Currently setup.py ignores KIVY_SDL2_PATH for darwin and you have to have SDL2 frameworks installed under /Library. With this change, setting KIVY_SDL2_PATH to your homebrew SDL2 header directory, Kivy builds without SDL2 frameworks. The compiled binary works (i.e. the homebrew SDL2 is correctly linked). Tested under OS X 10.10.5.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -378,7 +378,11 @@ if platform not in ('ios', 'android') and (c_options['use_gstreamer']
 sdl2_flags = {}
 if platform not in ('android',) and c_options['use_sdl2'] in (None, True):
 
-    if c_options['use_osx_frameworks'] and platform == 'darwin':
+    if platform == 'darwin' and environ.get('KIVY_SDL2_PATH', None):
+    	# paths and link flags will be configured in determine_sdl2()
+        c_options['use_sdl2'] = True
+
+    elif c_options['use_osx_frameworks'] and platform == 'darwin':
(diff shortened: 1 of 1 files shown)


akshayaurora:
@poke1024 try compiling kivy using USE_OSX_FRAMEWORKS=0 this will make kivy search for SDL2 in other paths than OS X frameworks.

```
    export USE_OSX_FRAMEWORKS=0
    make
```

To ignore USE_OSX_FRAMEWORKS is not the right option imo other paths should only be looked at when USE_OSX_FRAMEWORKS is set to 0 and that works right now.

poke1024:
using USE_OSX_FRAMEWORKS = 0 works and solves my problem. thanks!

since setup.py is now using pkg-config to find the sdl libraries, one nice thing to have would be to have a configurable pkg-config path (setup.py, line 61), not the global one or "default" one. I'm using several different installation systems with different pkg-configs (for example to build kivy for python2 and python3). Of course I can work around setting PATH accordingly, but still: a configurable pkg-config path in setup.py would be nice to have.

PR #3603, opened 2015-08-28
Adds clock to compat

matham:


kived:
:+1: as long as this all still works on py3 :)

dessant:
@matham could you rebase?

PR #3608, opened 2015-09-02
textinput-filter: comma

picibucor:
textinput-filter accepts commas as separator as well

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -656,6 +656,10 @@ class TextInput(FocusBehavior, Widget):
             else:
                 int_pat = self._insert_int_patu
 
+            if mode == 'float_comma':
+                substring = re.sub(',', '.', substring)
+                mode = 'float'
+
             if mode == 'int':
                 substring = re.sub(int_pat, chr(''), substring)
             elif mode == 'float':


matham:
Python doesn't seem to accept a comma:

```py
>>> float('3.5')
3.5
>>> float('3,5')
  File "<input>", line 1, in <module>
ValueError: invalid literal for float(): 3,5

```

matham:
If it's a local specific thing, then we would need to make it local dependent in kivy too.

picibucor:
Just one directive, out of many:
http://www.iec.ch/standardsdev/resources/draftingpublications/writing_editing/directives/numbers.htm

picibucor:
Check the standards of IEC, for example:
http://www.iec.ch/standardsdev/resources/draftingpublications/writing_editing/directives/numbers.htm

picibucor:
sorry, missclicked

dessant:
http://babel.pocoo.org/ could be used to get the localization data.

picibucor:
@dessant : Thank you, Babel seems to be very useful.
According to my experiences the input_filter: float  filters out everything what not number '[^0-9]' or  the first dot is.
Do I understand it right:
If I want to use Babel to convert the "local" input-text (comma instead of dot) into a python-accepted format (dot) I cannot use the "input_filter: float" option in the kv-file.

dessant:
@picibucor you could pass a function to input_filter that filters out everything that is not a dot, comma or number, the sanitized string would be passed to babel and you would just return the localized string from your function.

picibucor:
Do you mean so:
TextInput:
    input_filter:  [^0-9.,-]
?

akshayaurora:
Replaced by https://github.com/kivy/kivy/pull/3745, where we aim reuse internal python checking int/float to check if the value is valid.

PR #3620, opened 2015-09-09
deleted property to notify the widget has been removed

andnovar:
When a widget is removed I could not find a way to let the widget know it has been removed. I added an attribute that sets true when deleted.

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -464,6 +464,11 @@ class Widget(WidgetBase):
         for child in self.children:
             child.disabled = value
 
+    def on_deleted(self, instance, value):
+        if value and self.parent is None:
+            for child in self.children:
+                child.deleted = value
+
     #
     # Tree management
     #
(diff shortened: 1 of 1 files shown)


kivy/uix/widget.py
```diff
@@ -1216,3 +1222,16 @@ def on_opacity(self, instance, value):
     :attr:`disabled` is a :class:`~kivy.properties.BooleanProperty` and
     defaults to False.
     '''
+
+    deleted = BooleanProperty(False)
+    '''Indicates whether this widget has been removed from the parent or not.
```
dessant: There's already a property for this, `parent`.

matham:
Well, this wouldn't work because what if something is `"deleted"` and then re-added? From this, once it's deleted it'd never be triggered again, even if re-added. Anyway, deletion is not a state that is `True/False`, it's an event that happens.

Also, this concept of deleted doesn't quite work. What if you remove a widget from its parent, but it's parant never had a parent. The child would still trigger deleted, but it wasn't shown in the first place. And finally, as dessant says, you can get this info from just checking when the parent changes, or just overwriting add/remove_widget. So there's no reason to add this. You're basically adding a boolean to monitor another object property (parent)?

PR #3624, opened 2015-09-10
add show_hidden and dirselect parameters to SettingPath

janssen:
I found a need to be able to control the "show_hidden" and "dirselect" properties of SettingPath.  I created a branch with that capability.  Note that I also add an initial_path (in case "value" is None or an empty string), and delete a redundant assignment to self.textinput.


janssen:
I think I've screwed this up by trying to merge from master.  Closing it and starting again.

janssen:
No, I think it's OK.  Don't understand why all the merge changes are showing up, though.

janssen:
Whoops!  Failed the PEP 8 check -- too many spaces after a period.  Seems like nit-picking, but I'll fix it.

PR #3625, opened 2015-09-10
Add show_hidden and dirselect to SettingPath

janssen:
I found a need to be able to control the "show_hidden" and "dirselect" properties of SettingPath. I created a branch with that capability. Note that I also add an initial_path (in case "value" is None or an empty string), and delete a redundant assignment to self.textinput.


kivy/uix/settings.py
```diff
@@ -443,6 +443,21 @@ class SettingPath(SettingItem):
     defaults to None.
     '''
 
+    show_hidden = BooleanProperty(False)
+    '''Whether to show 'hidden' filenames. What that means is
+    operating-system-dependent.
+
```
akshayaurora: You are missing version added.

kivy/uix/settings.py
```diff
@@ -443,6 +443,21 @@ class SettingPath(SettingItem):
...
+    :attr:`show_hidden` is an :class:`~kivy.properties.BooleanProperty` and
+    defaults to False.
+    '''
+
+    dirselect = BooleanProperty(True)
+    '''Whether to allow selection of directories.
+
+    :attr:`dirselect` is an :class:`~kivy.properties.BooleanProperty` and
```
akshayaurora: Same as above.

janssen:
OK, I've updated with version-added.  Not sure what this Travis-CI failure is.

ngoonee:
Hope this can be merged, once the PEP8 is settled. However I think dirselect may need to default to False as that's the behaviour of FileChooser to begin with. Though in the specific case of a SettingPath we would more normally want dirselect=True, not sure if that's worth changing default behaviour.

janssen:
Well, I fixed the PEP 8 warning -- looks like it was some kind of Python 3.4 bug.  With Python 3.5, it's OK, but now it looks like the nose-test system is broken.  Don't think there's anything I can do to fix that.

KeyWeeUsr:
@janssen Would you mind merging from master, updating `versionadded` to the current master and maybe even squashing the commits into one? I'd like to see a result of travis after an update, though I'm not sure how to trigger the new checking, because manual restart of the job still uses the old checking, so I kind of hope force pushing a squashed commit could be the thing.

Anyway, seems good to me. Maybe switching to False as ngoonee suggested wouldn't be bad. ^^

dessant:
Adding a countdown. 😋 This could be handled by us if OP is gone.

janssen:
Still here, but I'm not a git ninja.  The steps suggested by @KeyWeeUsr would take some figuring out.

janssen:
I have merged from master and updated the pull request to make the versionadded be 1.9.2.  Not sure how to squash the commits, though.

KeyWeeUsr:
Don't worry. If the tests pass, I'll squash and merge. 🍡

kivy/uix/settings.py
```diff
@@ -476,10 +495,11 @@ def _create_popup(self, instance):
             width=popup_width)
 
         # create the filechooser
+        initial_path = self.value or os.getcwd()
         self.textinput = textinput = FileChooserListView(
-            path=self.value, size_hint=(1, 1), dirselect=True)
+            path=initial_path, size_hint=(1, 1),
+            dirselect=self.dirselect, show_hidden=self.show_hidden)
         textinput.bind(on_path=self._validate)
-        self.textinput = textinput
```
KeyWeeUsr: Does this break anything?

janssen:
I have no clue as to what causes that Python 3.4 error.  Over to you folks.

kivy/uix/settings.py
janssen: I think it's just redundant -- self.textinput is bound already above.

KeyWeeUsr:
Thanks. 👍

PR #3626, opened 2015-09-11
audio_sdl2: support audio formats beyond wav

bionoid:
audio_sdl2 would always only WAV support, because mix_flags was not global and no flags were supplied to Mix_Init(). I successfully played an OGG file with it, **but this needs review/testing**, I'm not familiar with SDL2.

### DIFF
--- a/kivy/core/audio/audio_sdl2.pyx
+++ b/kivy/core/audio/audio_sdl2.pyx
@@ -150,7 +150,8 @@ class SoundSDL2(Sound):
         self.unload()
         if self.filename is None:
             return
-        mc.chunk = Mix_LoadWAV(<char *><bytes>self.filename)
+        fn = self.filename.encode('UTF-8')
+        mc.chunk = Mix_LoadWAV(<char *><bytes>fn)
         if mc.chunk == NULL:
             Logger.warning('AudioSDL2: Unable to load %r' % self.filename)
         else:
--- a/kivy/core/audio/audio_sdl2.pyx
+++ b/kivy/core/audio/audio_sdl2.pyx
@@ -32,6 +32,7 @@ cdef mix_init():
     cdef int audio_channels = 2
     cdef int audio_buffers = 4096
     global mix_is_init
+    global mix_flags
 
     # avoid next call
     if mix_is_init != 0:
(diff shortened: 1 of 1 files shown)


kivy/core/audio/audio_sdl2.pyx
```diff
@@ -149,7 +148,8 @@ class SoundSDL2(Sound):
         self.unload()
         if self.filename is None:
             return
-        mc.chunk = Mix_LoadWAV(<char *><bytes>self.filename)
+        fn = self.filename.encode('UTF-8')
```
kived: I'm guessing you tested this in py3? If `self.filename` is already `bytes()` then this will throw an exception. You should probably do something like `fn = self.filename.encode('utf8') if not isinstance(self.filename, bytes) else self.filename` instead.

kivy/core/audio/audio_sdl2.pyx
```diff
@@ -42,7 +43,7 @@ cdef mix_init():
         mix_is_init = -1
         return 0
 
-    mix_flags = Mix_Init(0)
+    mix_flags = Mix_Init(MIX_INIT_FLAC|MIX_INIT_MOD|MIX_INIT_MP3|MIX_INIT_OGG)
```
kived: If you're missing support for one of these formats, does `Mix_Init` error out or does it just ignore the format?

bionoid: AFAIK it just ignores it. My test VM doesn't have MP3 support due to a bug in the sdl packaged with ubuntu 14.04, and the result is that mix_flags does not have MIX_INIT_MP3 set

bionoid: There is an issue with FLAC files - mix_flags gets MIX_INIT_FLAC (and the library is installed), but Mix_LoadWAV() returns NULL. Maybe it has to do with my SDL version

MIX_INIT_FLAC could just be removed from Mix_Init(). At least tests will pass on a system with only sdl2 audio provider, as long as libvorbis is available.

akshayaurora:
@bionoid :+1:

kivy/core/audio/audio_sdl2.pyx
```diff
@@ -111,13 +112,11 @@ class SoundSDL2(Sound):
...
-        Mix_QuerySpec(&freq, &fmt, &channels)
-        if fmt == AUDIO_S8 or fmt == AUDIO_U8:
-            mixerbytes = 1
-        else:
-            mixerbytes = 2
-        numsamples = mc.chunk.alen / mixerbytes / channels
-        return <double>numsamples / <double>channels
+        if not Mix_QuerySpec(&freq, &fmt, &channels):
+            return 0
+        points = mc.chunk.alen / ((fmt & 0xFF) / 8)
```
tito: ((fmt & 0xFF) / 8)
^ Really ? Where is the reference about it?

bionoid: https://wiki.libsdl.org/SDL_AudioFormat

PR #3632, opened 2015-09-15
fix recursion depth error

Zen-CODE:
From this ticket. Setting focus = True on a widget inheriting FocusBehavior gives stack overflow or maximum recursion depth exceeded. This happens only when the 'is_focusable' property is True.

https://groups.google.com/forum/#!topic/kivy-users/eGsaZYOrH5Q

A minimal, runnable example follows;

    #!/usr/bin/env python
    from kivy.app import App
    from kivy.lang import Builder

    kv = '''
    <MyTextInput@TextInput>:
        is_focusable: False

    BoxLayout:
        Button:
            text: 'Set focus'
            on_release: ti.focus = True
        MyTextInput:
            id: ti
    '''

    class TestApp(App):
        def build(self):

    TestApp().run()

### DIFF
--- a/kivy/uix/behaviors.py
+++ b/kivy/uix/behaviors.py
@@ -873,12 +873,19 @@ class FocusBehavior(object):
                 keyboards[keyboard] = None
 
     def _bind_keyboard(self):
-        self._ensure_keyboard()
-        keyboard = self._keyboard
+        if self.disabled or not self.is_focusable:
+            keyboard = False
+        else:
+            self._ensure_keyboard()
+            keyboard = self._keyboard
 
-        if not keyboard or self.disabled or not self.is_focusable:
-            self.focus = False
+        if not keyboard:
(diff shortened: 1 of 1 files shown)


akshayaurora:
This was discussed before, I will repeat the main points.  Calling on a schedule for focus or unfocus on core code will break apps that rely on focus being set/unset immediately for example..

```
    self.focus = False
    do code that depends on focus being False
```

This will cause far more breakage than simply allowing the user to check for `is_focusable`.

I do think that UI programming has to be based on events and proposing users to use `on_focus` event instead might be a good idea, however this is a breaking change. This might be a good candidate for 2.0 as it changes the existing behavior.

Zen-CODE:
Is does feel hackish i must admit. Is there another way to stop the focus
proprety from changing? A better way i could investigate ?

PR #3634, opened 2015-09-16
Corrected on_frame in doc

bliz937:
Couldn't get ```on_frame``` event to work, found that the docs show the event as ```on_frame``` but it is actually ```on_texture```.


akshayaurora:
@bliz937 thanks! Don't know how this went unnoticed for so long Oo.

PR #3636, opened 2015-09-16
audio_sdl2: support sequenced formats

bionoid:
Adds support for mod/s3m/midi/etc playback, which can only be done on mixer music channel.

Only tested with libmodplug, assume it will work if built for mikmod or both libraries. (EDIT: works with mikmod and both at once)

* Fixed incorrect MIX_INIT_* bitmasks in lib/sdl2.pxi
* Renamed MixContainer to ChunkContainer. MusicContainer was introduced, and both relate to the mixer, so MixContainer became ambigious
* Added MusicSDL2 to support playing sequenced formats on SDL2-mixer music channel. No support for wave formats - we'd need a mixer abstraction to separate them
* WARNING: Can't _get_length for sequenced formats; using 0 if nothing loaded and -1 if loaded (better suggestions?)

### DIFF
--- a/kivy/core/audio/audio_sdl2.pyx
+++ b/kivy/core/audio/audio_sdl2.pyx
@@ -21,11 +21,11 @@ Depending the compilation of SDL2 mixer and/or installed libraries:
 .. Warning::
 
     Sequenced formats use the SDL2 Mixer music channel, you can only play
-    one at a time, and .length will be 0 if no music is loaded, and -1
-    if loaded (can't get duration of these formats)
+    one at a time, and .length will be -1 if music fails to load, and 0
+    if loaded successfully (we can't get duration of these formats)
 '''
 
-__all__ = ('SoundSDL2', )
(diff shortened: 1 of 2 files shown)


tito:
I would have actually done the inverse (except if there is a case i don't know about): 0 when everything is "ok", -1 as an error. Otherwise, good PR, all the rest is ok :)

akshayaurora:
@bionoid nice work :+1: merge if you are comfortable with the current state.

bionoid:
Only thing is that we don't call ``Mix_Quit()`` on app exit to free dynamically loaded libraries, I'm not sure about the consequence of this or what is the best approach. It's probably harmless to just call it before ``SDL_Quit()`` here?

https://github.com/kivy/kivy/blob/332bc89e1888b322cec4c2255a08f9f0398f2a40/kivy/core/window/_window_sdl2.pyx#L223

tito:
Merge ok then :)

Issue #3630, opened 2015-09-14
Installation Doc Ubuntu packages


dec0dedab0de:
In /kivy/blob/master/doc/sources/installation/installation.rst  The Ubuntu section says to install libgl1-mesa-dev-lts-quantal, and  libgles2-mesa-dev-lts-quantal.   Both of which are not available in 15.04. I think it should have  libgl1-mesa-dev, and  libgles2-mesa-dev instead.

mondayrain:
Just tested on my Vivid, can confirm.

mondayrain:
@dec0dedab0de @dessant 
this can be closed!

dessant:
@mondayrain thanks :+1:

PR #3639, opened 2015-09-18
Updated Ubuntu installation packages docs for 15.04

mondayrain:
As @dec0dedab0de pointed in issue #3630 that the correct apt-get packages for Ubuntu 15.04 were not specified; added it into documentation. Turns out they're the same as < 12.04!


mondayrain:
@Zen-CODE, as you seem to take care of majority of the documentation stuff here :)

Zen-CODE:
Great stuff. Thanks for that :-)

Issue #3640, opened 2015-09-18
Touchscreen not working in Kivy on RaspberryPi official touchscreen


dazzag24:
Update kernel and rebooted as per details here:
https://www.raspberrypi.org/blog/#the-eagerly-awaited-raspberry-pi-display

Then followed instructions here:
http://kivy.org/docs/installation/installation-rpi.html

[
As an aside these instructions result in the demos not running.  You need to install some additional Python modules:
sudo pip install pygments docutils
]

The demos now run, BUT none of them respond to the touchscreen.    The touchscreen DOES work in X windows, so it isn't a hardware issue I hope.

Output from running the pictures demo:

[INFO              ] [Kivy        ] v1.9.1-dev
[INFO              ] [Python      ] v2.7.3 (default, Mar 18 2014, 05:13:23) 
[GCC 4.6.3]
[INFO              ] [Factory     ] 177 symbols loaded
[INFO              ] [Image       ] Providers: img_tex, img_dds, img_gif, img_pygame, img_pil (img_ffpyplayer ignored)
[INFO              ] [Window      ] Provider: egl_rpi
[INFO              ] [GL          ] OpenGL version <OpenGL ES 2.0>
[INFO              ] [GL          ] OpenGL vendor <Broadcom>
[INFO              ] [GL          ] OpenGL renderer <VideoCore IV HW>
[INFO              ] [GL          ] OpenGL parsed version: 2, 0
[INFO              ] [GL          ] Shading version <OpenGL ES GLSL ES 1.00>
[INFO              ] [GL          ] Texture max size <2048>
[INFO              ] [GL          ] Texture max units <8>
[INFO              ] [Shader      ] fragment shader: <Compiled>
[INFO              ] [Shader      ] vertex shader: <Compiled>
[INFO              ] [Window      ] virtual keyboard not allowed, single mode, not docked
[INFO              ] [GL          ] NPOT texture support is available
[INFO              ] [Text        ] Provider: pygame
[INFO              ] [OSC         ] using <multiprocessing> for socket
[INFO              ] [ProbeSysfs  ] device match: /dev/input/event0
[INFO              ] [HIDInput    ] Read event from </dev/input/event0>
[INFO              ] [Base        ] Start application main loop
[INFO              ] [HIDMotionEvent] using <FT5406 memory based driver>
[INFO              ] [HIDMotionEvent] <FT5406 memory based driver> range ABS X position is 0 - 800
[...]

PR #3642, opened 2015-09-19
fix RPI touchscreen not working with KIVY_BCM_DISPMANX_ID=4

jehutting:
This fix (for issue #3640) allows it to use the Raspberry Pi official touchscreen besides the HDMI screen.

### DIFF
--- a/kivy/lib/vidcore_lite/egl.pyx
+++ b/kivy/lib/vidcore_lite/egl.pyx
@@ -603,7 +603,7 @@ def bcm_display_open(bcm.uint32_t device):
     cdef:
         bcm.DISPMANX_DISPLAY_HANDLE_T disp
         bcm.DisplayHandle D
-    disp = bcm.vc_dispmanx_display_open( 0 )
+    disp = bcm.vc_dispmanx_display_open( device )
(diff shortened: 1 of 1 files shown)


inclement:
Thanks a lot for this fix!

RussNelson:
Not working for me. Touch screen works in X, not in Kivy examples.

jehutting:
OK...when running the GUI on the HDMI screen and the demo on the touch screen I see that the touch screen touches and movements are also driving the mouse on the HDMI screen. So there is no differentiation between those two. Is this the behaviour you had and fixed?

RussNelson:
The Kivy demo screen drew just fine, but no mouse was visible and no touch was recognized. I wasn't using the HDMI screen -- serial console instead. FWIW, I used x2x to connect and saw the same thing. No mouse, no clicks were recognized, however clicks and drags took effect on the other windows.

makuk66:
Thanks for this fix -- works for me on rpi2,with `KIVY_BCM_DISPMANX_ID=4 python ./examples/demo/pictures/main.py -d`, including touch.

Issue #3385, opened 2015-05-26
input_filter : negative numbers


picibucor:
The input_filter: float / int  property does not allow negative numbers. 
##

picibucor:
regular-expression for negative and non-negative numbers:     ^[-0-9][0-9]+

Is there a way to invert this expression?

Paradoxeuh:
I'm sorry but I can't find a simple way to have integer positive and negative, how is it working now? (I'm using the daily-repository of kivy on linux). Might be the wrong place but I didn't find any clear explanation in the commits

matham:
Negative integers do not work currently. https://github.com/kivy/kivy/pull/3745 should fix it.

Paradoxeuh:
Thank you for your answer. Is it possible to know when is this change is going to be merged ? Roughly ?

dessant:
@Paradoxeuh, there is no eta, testing the pr and giving feedback helps speeding up the merge.

matham:
I want to finally add this but I don't to remove the ability to just have positive numbers. So I think we need options like `int_pos`, `int`, `float`, and `float_pos`. However, this breaks previous behavior for e.g. `int` not allowing negatives. Also, positive ints do not include zero. So I'm not sure what the best options are.

PR #3646, opened 2015-09-21
input_filter: negative numbers

picibucor:
allowing negative numbers by the filter options: int, float
trying to solve issue #3385

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -659,22 +659,16 @@ class TextInput(FocusBehavior, Widget):
                 int_pat = self._insert_int_patu
 
             if mode == 'int':
-                try:
+                if self.text != '':
                     if self.text[0] == '-':
-                        substring = re.sub('-', '', substring)
-                    else:
-                        substring = re.sub('-', '', substring)
-                except:
-                    pass
+                        pass
+                    substring = re.sub('-', '', substring)
                 substring = re.sub(int_pat, chr(''), substring)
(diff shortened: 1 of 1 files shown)


kivy/uix/textinput.py
```diff
@@ -659,8 +659,22 @@ def insert_text(self, substring, from_undo=False):
                 int_pat = self._insert_int_patu
 
             if mode == 'int':
+                try:
+                    if self.text[0] == '-':
+                        substring = re.sub('-', '', substring)
+                    else:
+                        substring = re.sub('-', '', substring)
+                except:
+                    pass
```
matham: Why do you need the try/except? Also, both of the if/else cases have identical code...

kivy/uix/textinput.py
```diff
@@ -2406,8 +2420,8 @@ def _refresh_hint_text(self):
     _lines = ListProperty([])
     _hint_text_lines = ListProperty([])
     _editable = BooleanProperty(True)
-    _insert_int_patu = re.compile(u'[^0-9]')
-    _insert_int_patb = re.compile(b'[^0-9]')
+    _insert_int_patu = re.compile(u'[^0-9-]')
+    _insert_int_patb = re.compile(b'[^0-9-]')
```
matham: I suspect you need to do `[^0-9\\-]`.

akshayaurora:
Replaced by https://github.com/kivy/kivy/pull/3745

PR #3652, opened 2015-09-24
Update pyinstaller hooks to use proper api. Fixes #3622.

### LINKED ISSUE #3622: Python 3.4 packager KeyError: 'rthooks'
As described here:
http://stackoverflow.com/questions/31083061/kivy-1-9-0-windows-package-keyerror-rthooks/32366409#32366409

I've partially troubleshooted this:

---------------

The [kivy docs](http://kivy.org/docs/guide/packaging-windows.html) you reference mention adding the following three lines to the top of the `.spec` file:

    from kivy.tools.packaging.pyinstaller_hooks import install_hooks
    import os
    install_hooks(globals())

The error is happening in `install_hooks(globals())`, which is defined at `\Lib\site-packages\kivy\tools\packaging\pyinstaller_hooks\__init__.py`:

    from os.path import dirname, join
    from functools import partial
    
    curdir = dirname(__file__)

    def install_hooks(sym, hookspath=None):
    
        _hookspath = [curdir]
        if hookspath is not None:
            _hookspath += hookspath
    
        sym['rthooks']['kivy'] = [join(curdir, 'rt-hook-kivy.py')]
        sym['Analysis'] = partial(sym['Analysis'], hookspath=_hookspath)

But the second last line is causing the message: `WARNING: stderr: KeyError: 'rthooks'`.

So it looks like it's expecting a variable `rthooks` to be in the global namespace, but it's not.

I'm not sure what to do next.

matham:
Fixes #3622.

### DIFF
--- a/doc/sources/guide/packaging-windows.rst
+++ b/doc/sources/guide/packaging-windows.rst
@@ -8,10 +8,11 @@ package of Kivy.
 The package will be either 32 or 64 bits depending on which version of Python
 you ran it with.
 
-+-----------------------------------------------------------------------------+
-| NOTE: Currently, packages for Windows can be generated with Python 2.7 and  |
-| Python 3.3+. However, Python 3.3+ support is still experimental             |
-+-----------------------------------------------------------------------------+
+
+.. note::
+    Currently, packages for Windows can be generated with Python 2.7 and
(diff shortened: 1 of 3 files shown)


akshayaurora:
@matham :dancer:

PR #3655, opened 2015-09-28
Fix PyEnchant URL

ohaz:
The URL was broken, it linked to a 404. The new page is the one that the developers page links to.


akshayaurora:
@ohaz thanks.

PR #3656, opened 2015-09-28
Added verify option to UrlRequest

Ramalus:
If verify is False, SSL CA certificate validadion is disabled.


dessant:
Works great. :+1: 

```python
from kivy.app import App
from kivy.network.urlrequest import UrlRequest


class TestApp(App):

    def callback(self, req, result):
        print(result)

    def on_start(self, *args):
        UrlRequest('https://www.flaskapi.org/', on_success=self.callback,
                    on_failure=self.callback, on_error=self.callback,
                    verify=False)


if __name__ == '__main__':
    TestApp().run()
```

Issue #3253, opened 2015-04-08
kivy-examples package broken


inclement:
It's reported at https://groups.google.com/forum/#!topic/kivy-users/hSyITNXQyb0 that the kivy-examples package doesn't currently work on Ubuntu.

thopiekar:
Need a more detailed output, eg. a commandline output, and more important, which PPA was used.

TimoKramer:
I tried to run the kivycatalog and got this error:

Traceback (most recent call last):
   File "main.py", line 191, in <module>
     KivyCatalogApp().run()
   File "/usr/lib/python2.7/dist-packages/kivy/app.py", line 824, in run
     runTouchApp()
   File "/usr/lib/python2.7/dist-packages/kivy/base.py", line 487, in runTouchApp
     EventLoop.window.mainloop()
   File "/usr/lib/python2.7/dist-packages/kivy/core/window/window_sdl2.py", line 539, in mainloop
     self._mainloop()
   File "/usr/lib/python2.7/dist-packages/kivy/core/window/window_sdl2.py", line 300, in _mainloop
     EventLoop.idle()
   File "/usr/lib/python2.7/dist-packages/kivy/base.py", line 327, in idle
     Clock.tick()
   File "/usr/lib/python2.7/dist-packages/kivy/clock.py", line 483, in tick
     self._process_events()
   File "/usr/lib/python2.7/dist-packages/kivy/clock.py", line 615, in _process_events
     event.tick(self._last_tick, remove)
   File "/usr/lib/python2.7/dist-packages/kivy/clock.py", line 374, in tick
     ret = callback(self._dt)
   File "/usr/lib/python2.7/dist-packages/kivy/uix/video.py", line 177, in _do_video_load
     self._video.play()
   File "/usr/lib/python2.7/dist-packages/kivy/core/video/video_gstplayer.py", line 88, in play
     self.player.set_volume(self.volume)
 AttributeError: 'NoneType' object has no attribute 'set_volume'

TimoKramer:
same when running python3 kvrun.py app_logo.kv when I push the button

 Traceback (most recent call last):
   File "kvrun.py", line 35, in <module>
     KvApp(filename=sys.argv[1]).run()
   File "/usr/lib/python3/dist-packages/kivy/app.py", line 824, in run
     runTouchApp()
   File "/usr/lib/python3/dist-packages/kivy/base.py", line 487, in runTouchApp
     EventLoop.window.mainloop()
[...]

PR #3664, opened 2015-10-05
Fix for sample path in examples

josephbreihan:
Updates path mentioned in issue #3253 to /usr/share/kivy-examples/widgets.

### DIFF
--- a/examples/kv/app_logo.kv
+++ b/examples/kv/app_logo.kv
@@ -1,7 +1,7 @@
 Widget:
     Video:
         id: myvideo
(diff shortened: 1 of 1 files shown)


dessant:
Thanks for contributing. You might want to use a relative path like [here](https://github.com/kivy/kivy/blob/2cbf08ced7ec206280b5e61cf5e24791af691bde/examples/demo/kivycatalog/container_kvs/MediaContainer.kv#L8), it would work on all platforms.

dessant:
Closed in favor of https://github.com/kivy/kivy/pull/4332.

PR #3666, opened 2015-10-05
ignore redis ConnectionError in RedisStore test

dessant:
When redis-py is installed, the test is failing unless redis is started manually, ignoring ConnectionError should be safe.


dessant:
capture

PR #3669, opened 2015-10-05
Fix breakage in Python < 2.7.9

Ramalus:
Checked if ssl has attribute create_default_context before using it.
This was introduced in Python 2.7.9 so in older versions this code would break without the check.


dessant:
:+1:

PR #3672, opened 2015-10-07
Maintain last selection if not allowing empty selection

ngoonee:
Self-explanatory. There doesn't seem any good reason to always default to the first item if the user deselects the last item, in the case where allow_empty_selection is already False.

In particular, this means accidental double/multiple-selections (especially common when using a mouse rather than touch) become less of an issue, and the likely intention of the user is carried through.

### DIFF
--- a/kivy/adapters/listadapter.py
+++ b/kivy/adapters/listadapter.py
@@ -299,7 +299,14 @@ class ListAdapter(Adapter, EventDispatcher):
                 else:
                     self.select_item_view(view)
         else:
-            self.deselect_item_view(view)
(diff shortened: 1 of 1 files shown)


dessant:
@ngoonee, nice change :+1:

dessant:
The merge makes `test_list_adapter_selection_handle_selection` fail on osx and windows, while it passes on linux.

dessant:
Test adapted in https://github.com/kivy/kivy/commit/24f4b605b23ba4f0bd4ecf2993a3bf2a0f363336.

PR #3675, opened 2015-10-08
Fix python-for-android documentation URL

ohaz:
Was http://python-for-android.readthedocs.org/en/latest/android/, which lead to a 404, changed to http://python-for-android.readthedocs.org/en/latest/


inclement:
Thanks ohaz (and maltouzes).

I didn't think of this, but probably the kivy p4a doc should all be reworded to note about the new toolchain, so thanks for that too!

PR #3679, opened 2015-10-09
Documentation: update OS X instructions

cbenhagen:
- remove warning from 4e1704f as we do not use pygame on OS X anymore since 1.9.0
- rename MacOSX to OS X
- add a chapter on using pip and homebrew
- use $ signs
- remove instructions to install dev version in kivy.app

The pip instructions depend on #3682


dessant:
Great work! :+1:

PR #3685, opened 2015-10-11
search for frameworks AND via pkg-config

cbenhagen:
this allows installing via pip on brew and non-brew systems
without the need for USE_OSX_FRAMEWORKS=0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -366,7 +366,8 @@ if platform not in ('ios', 'android') and (c_options['use_gstreamer']
                     '-framework', 'GStreamer'],
                 'include_dirs': [join(f_path, 'Headers')]}
 
-    else:
+    # use pkgconfig if gstreamer not already configured using frameworks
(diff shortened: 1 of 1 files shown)


cbenhagen:
@akshayaurora any news on this?

akshayaurora:
@cbenhagen as it is this actually breaks compilation for frameworks,  we are about to release anyways please wait for 1.9.2 for major changes like this.

cbenhagen:
Why and how would this break compilation for frameworks? If so I would love to fix that. It will use frameworks as before and also gives preference to them if both a homebrew and a framework version is present.

As for the release. I don't mind waiting for 1.9.2. But also I do not see how this is a major change.

cbenhagen:
@akshayaurora Can we now merge this?

cbenhagen:
Sorry i had to delete my kivy fork some time ago and creating a new PR (#4027) was the only option to update it.

PR #3686, opened 2015-10-11
Add option to mimic Spinner height in dropdown

ngoonee:
A fairly arbitrary size 48 height is currently the default (where's this set, by the way? Certainly not in this file. I couldn't find it).

This patch adds an option `mimic_size` which would have each dropdown element's height set according to the current size of the Spinner. More visually pleasing IMO. User customization is unaffected, perhaps a note should be made that this overwrites any changes in SpinnerOption.

For consideration before accepting:-
mimic_size or mimic_height?
I only set item.height (based on self.height) as size_hint_x of the default SpinnerOption is 1, hence setting item.size would not make a difference and may be a bit unclear.

### DIFF
--- a/kivy/uix/spinner.py
+++ b/kivy/uix/spinner.py
@@ -105,6 +105,15 @@ class Spinner(Button):
 
     .. versionadded:: 1.4.0
     '''
+    
+    mimic_size = BooleanProperty(False)
+    '''Each element in the dropdown list uses a default/user-supplied height.
+    Set to True to propagate the Spinner's height value to each dropdown list
+    element.
+    :attr:`mimic_size` is a :class:`~kivy.properties.BooleanProperty` and
+    defaults to False.
+    .. versionadded:: No idea
+    '''
 
     def __init__(self, **kwargs):
         self._dropdown = None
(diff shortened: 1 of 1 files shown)


ngoonee:
Ah ignore the first question, a 'grep 48' found me style.kv. Man, shouldn't this be going into individual kv files for each .py? This file is humongous. I guess its for efficiency purposes rather than having a Builder line for each python source.

ngoonee:
Also, I see this patch failed travis-CI pep8. Not sure why, it passes pep8 here on my machine (using the vim plugin for that), and the changes are so minimal it's unlikely I made a copy paste mistake.

dessant:
@ngoonee, we're ready to merge the pr, nice work.
Would you rebase and rename the property to `sync_height`? The version is `1.9.2`.

ngoonee:
No issue doing that @dessant but I don't have this checked out and am currently submitting PRs via the github web interface. Can I update this PR or should I just submit a new one?

dessant:
Submitting a new one is ok, or we can also commit the change for you.

ngoonee:
Okay resubmitted a new pull request @dessant Closing this one

PR #3692, opened 2015-10-12
Allow using unicode configparser

tshirtman:
First, try using configparser, so users of python2 configparser backport
can use it.
Second, add an as_unicode parameter to write, so it's possible to use
codecs.open instead of open to save the file.


tshirtman:
- Could be a breaking change, for people that have configparser but didn't use it in kivy (i don't think it's a big issue though).
- `as_unicode` may not be the best parameter, but i didn't want to use `unicode`, because shadowing builtins is bad.

PR #3695, opened 2015-10-13
Extended OSC library

LoialOtter:
Added tags: T,F,I,N (true, false, impulse, null)
Fixed error with timetag - now a 64bit integer
Added 'default' to dispatch for handling unknown addresses.

### DIFF
--- a/kivy/lib/osc/OSC.py
+++ b/kivy/lib/osc/OSC.py
@@ -34,6 +34,25 @@ import string
 import pprint
 from kivy.compat import string_types
 
+class binaryString(bytearray):
+    pass
+
+class impulse(object):
+    def __nonzero__(self):
+        return True
+    def __str__(self):
+        return "Impulse"
+    def __repr__(self):
+        return "Impulse"
+
+class null(object):
+    def __nonzero__(self):
+        return False
+    def __str__(self):
+        return "NULL"
+    def __repr__(self):
+        return "NULL"
+
 
 def hexDump(data):
     """Useful utility; prints the string in hexadecimal"""
@@ -98,6 +117,18 @@ class OSCMessage:
     def __repr__(self):
         return self.getBinary()
 
+def readTrue(data):
+    return (True, data)
+
+def readFalse(data):
+    return (False, data)
+
+def readImpulse(data):
+    return (impulse(), data)
+
+def readNull(data):
+    return (null(), data)
+
 def readString(data):
     if isinstance(data, str):
         length = string.find(data, '\0')
@@ -108,9 +139,21 @@ def readString(data):
 
 
 def readBlob(data):
-    length   = struct.unpack(">i", data[0:4])[0]
-    nextData = int(math.ceil((length) / 4.0) * 4) + 4
-    return (data[4:length+4], data[nextData:])
+    if(len(data)<4):
+        print("Error: too few bytes for blob", data, len(data))
+        rest = data
+        blob = ""
+    else:
+        length   = struct.unpack(">i", data[0:4])[0]
+        if(len(data)-4 < length):
+            print("Error: too few bytes for blob", data, len(data))
+            rest = data
+            blob = ""
+        else:
+            nextData = int(math.ceil((length) / 4.0) * 4) + 4
+            rest = data[nextData:]
+            blob = data[4:length+4]
+    return (blob, rest)
 
 
 def readInt(data):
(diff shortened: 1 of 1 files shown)


kivy/lib/osc/OSC.py
```diff
@@ -108,9 +139,21 @@ def readString(data):
...
 def readBlob(data):
-    length   = struct.unpack(">i", data[0:4])[0]
-    nextData = int(math.ceil((length) / 4.0) * 4) + 4
-    return (data[4:length+4], data[nextData:])
+    if(len(data)<4):
+        print("Error: too few bytes for blob", data, len(data))
+        rest = data
+        blob = ""
+    else:
+        length   = struct.unpack(">i", data[0:4])[0]
```
tshirtman: multiple spaces, will break style guide.

kivy/lib/osc/OSC.py
```diff
@@ -108,9 +139,21 @@ def readString(data):
 
 
 def readBlob(data):
-    length   = struct.unpack(">i", data[0:4])[0]
-    nextData = int(math.ceil((length) / 4.0) * 4) + 4
-    return (data[4:length+4], data[nextData:])
+    if(len(data)<4):
```
tshirtman: I think instead of testing length, using try/except and returning ('', data) when data is wrong incomplete could be better performance-wise, and probably shorter, but nothing wrong with this way either, afaik.

kivy/lib/osc/OSC.py
```diff
@@ -177,24 +231,45 @@ def OSCArgument(data):
     """Convert some Python types to their
     OSC binary representations, returning a
     (typetag, data) tuple."""
+        
+    try:
```
tshirtman: Any particular reason to eat exceptions here? Did you have a specific issue to fix?

kivy/lib/osc/OSC.py
```diff
@@ -220,33 +295,36 @@ def parseArgs(args):
...
-            for tag in typetags[1:]:
-                value, rest = table[tag](rest)
-                decoded.append(value)
-        else:
-            print("Oops, typetag lacks the magic ,")
-
-    return decoded
+    try:
+        """Converts a typetagged OSC message to a Python list."""
+        table = { "i" : readInt, "f" : readFloat, "s" : readString, "b" : readBlob, "d" : readDouble, "t" : readLong, "T" : readTrue, "F" : readFalse, "I" : readImpulse, "N" : readNull }
```
tshirtman: would be nice to reflow the dict here, for readability. :)

kivy/lib/osc/OSC.py
```diff
@@ -220,33 +295,36 @@ def parseArgs(args):
...
-        decoded.append(typetags)
-        if typetags[0] == ",":
-            for tag in typetags[1:]:
-                value, rest = table[tag](rest)
-                decoded.append(value)
-        else:
-            print("Oops, typetag lacks the magic ,")
-
-    return decoded
+    try:
```
tshirtman: Same, any reason for the try/except?

kivy/lib/osc/OSC.py
```diff
@@ -268,22 +346,33 @@ def handle(self, data, source = None):
...
         if type(message[0]) == list :
             # smells like nested messages
             for msg in message :
                 self.dispatch(msg, source)
         elif type(message[0]) == str :
+            valid = False
             # got a single message
             try:
                 address = message[0]
                 callbackfunction = self.callbacks[address]
```
tshirtman: you can simplify the fallback to defaults with a self.callbacks.get('address', self.callbacks.get('default')) and checking result is not None, you should be able to get ride of both try/except, and of the "valid" variable.

kivy/lib/osc/OSC.py
LoialOtter: There's an issue if data being given isn't valid; probably poor coding practices on my own end as I don't wrap the sending method to catch error data.
This closer matches incorrect data coming from on the TCP connection though.

kivy/lib/osc/OSC.py
```diff
@@ -34,6 +34,25 @@
 import pprint
 from kivy.compat import string_types
 
+class binaryString(bytearray):
```
tshirtman: I think you didn't use it in the end, if not needed, can be removed.

kivy/lib/osc/OSC.py
LoialOtter: Was just applying the style found there. Changing it to try/except makes more sense.

kivy/lib/osc/OSC.py
LoialOtter: This one is much more touchy; if incorrect data is coming on TCP connection, the packet should be reported and ignored. If this didn't eat the exception, the program would close with little indication of why.
This should be made more robust. Also; I would like to get 'logging' in here but haven't looked at how much work that'll take yet.

kivy/lib/osc/OSC.py
LoialOtter: good catch. I'm pretty sure that's my original workaround method and then I realized I could just use binarystrings directly and check in OSCArgument for bytearray/string.
Fine to remove.

dessant:
@tshirtman :pager:

tshirtman:
Sorry for the long delay, finally found some time to test in to be sure there was no issue with my use of TUIO with it, and it's all good, thanks!

Issue #3650, opened 2015-09-23
TypeError: object.__init__() takes no parameters


dessant:
Reported by TwoCubed on irc.

https://github.com/kivy/kivy/blob/master/examples/widgets/settings.py
```python
 Traceback (most recent call last):
   File "E:\dev\envs\py34_x86\kivy\examples\widgets\settings.py", line 93, in <module>
     SettingsApp().run()
   File "e:\dev\envs\py34_x86\kivy\kivy\app.py", line 825, in run
     runTouchApp()
   File "e:\dev\envs\py34_x86\kivy\kivy\base.py", line 487, in runTouchApp
     EventLoop.window.mainloop()
   File "e:\dev\envs\py34_x86\kivy\kivy\core\window\window_sdl2.py", line 606, in mainloop
     self._mainloop()
   File "e:\dev\envs\py34_x86\kivy\kivy\core\window\window_sdl2.py", line 361, in _mainloop
     EventLoop.idle()
   File "e:\dev\envs\py34_x86\kivy\kivy\base.py", line 330, in idle
     self.dispatch_input()
   File "e:\dev\envs\py34_x86\kivy\kivy\base.py", line 315, in dispatch_input
     post_dispatch_input(*pop(0))
   File "e:\dev\envs\py34_x86\kivy\kivy\base.py", line 221, in post_dispatch_input
     listener.dispatch('on_motion', etype, me)
   File "kivy\_event.pyx", line 717, in kivy._event.EventDispatcher.dispatch (kivy\_event.c:7635)
   File "e:\dev\envs\py34_x86\kivy\kivy\core\window\__init__.py", line 986, in on_motion
     self.dispatch('on_touch_down', me)
   File "kivy\_event.pyx", line 717, in kivy._event.EventDispatcher.dispatch (kivy\_event.c:7635)
   File "e:\dev\envs\py34_x86\kivy\kivy\core\window\__init__.py", line 1002, in on_touch_down
     if w.dispatch('on_touch_down', touch):
   File "kivy\_event.pyx", line 717, in kivy._event.EventDispatcher.dispatch (kivy\_event.c:7635)
   File "e:\dev\envs\py34_x86\kivy\kivy\uix\widget.py", line 438, in on_touch_down
     if child.dispatch('on_touch_down', touch):
   File "kivy\_event.pyx", line 717, in kivy._event.EventDispatcher.dispatch (kivy\_event.c:7635)
[...]

PR #3696, opened 2015-10-14
super in base object should not pass on kwargs

ngoonee:
Based on comments in https://github.com/kivy/kivy/issues/3650 - this fixes crashes whenever opening Settings (including in the examples)

GBlame shows that this line was added in commit d7bf6dbd by Matthew Einhorn on 2015-06-29 

Not sure if that commit needs reverting as I don't really understand what was being done. But yes, this fixes a 100% crash.

### DIFF
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -250,7 +250,7 @@ cdef class EventDispatcher(ObjectWithUid):
         properties = self.properties()
         prop_args = [
             (k, kwargs.pop(k)) for k in list(kwargs.keys()) if k in properties]
-        super(EventDispatcher, self).__init__(**kwargs)
+        super(EventDispatcher, self).__init__()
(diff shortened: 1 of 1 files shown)


ngoonee:
Interestingly enough... this crash only occurs with python3. When I finally got around to testing on python2, the crash does not occur (both in my app as well as in the Settings example)

matham:
The correct solution is not to revert that change and not forward kwargs, because for cooperative inheritance to work, we need to pass on any kwargs we didn't use. So you need to find where in the code you're passing on kwargs that never gets eaten and not forward them with super.

matham:
We appreciate your PR and work on this, though.

ngoonee:
The default settings example hits this crash as well. I believe it has to do with 'title' not being a property.

Issue #3690, opened 2015-10-12
Sound example doesn't show all buttons when built for mobile


stuaxo:
Heres a screenshot -

http://imgur.com/qzSvH0m

You can't press any sound buttons as they are offscreen

udiboy1209:
I would like to fix this issue. Just to confirm, this is a bug in the `example/audio` app, right?

stuaxo:
Yup !

udiboy1209:
From what I understand in the example, it searches for `*.wav` files in the directory where `main.py` lives an makes a button for each file found.

So I don't think the problem is that the buttons are off-screen, but that it cant find any sound files in the current directory when installed on Android. I tested this by moving all the sound files to another dir, and it had the same effect on Linux too.

I am not sure exactly what directory it would check when on Android ( what will `dirname(__file__)` retunr? ). If anybody has an idea, this could be tested further by moving a few sound files to that folder.

stuaxo:
I'll grab that info for you...

pythonic64:
@udiboy1209  Did you add ",wav" at the end of this line in your buildozer.spec:
https://github.com/kivy/buildozer/blob/master/buildozer/default.spec#L16 ?

stuaxo:
Ah - gets the buttons displaying and they work; though they could probably do some tweaking so the text inside them fits and they don't overlap the volume slider.
http://imgur.com/pwVTPzS

The orientation is wrong too.

(note - if the buttons are bigger there doesn't seem to be able scroll when they are offscreen either).


Buildozer:

I wonder if there should be something like

buildozer.defaults

That could overwrite settings when   buildozer init   was run - that way we the examples could all specify stuff like ,wav where needed and have correct names (they are all just My Application now)

pythonic64:
Sizes for AudioButton instances in [main.py](https://github.com/kivy/kivy/blob/master/examples/audio/main.py#L66) and for widgets in audio.kv needs to be in dp. 
You can add audio buttons container [sl](https://github.com/kivy/kivy/blob/master/examples/audio/audio.kv#L45) inside of ScrollView to enable scrolling.

Also, you can add buildozer.spec file, which will be default for this example, with appropriate values set for title, package.name, orientation etc. 
When creating apk developer can further customize .spec file, (with paths for android sdk, ndk etc.) if needed and then execute "buildozer android debug deploy run" to run this example on android device.

udiboy1209:
[...]

PR #3698, opened 2015-10-14
Fix audio example not including sound files on Android

udiboy1209:
This is a pull request to fix the issue #3690


akshayaurora:
@udiboy1209 nice contribution :+1:

PR #3700, opened 2015-10-15
Update firstwidget.rst

emon10005:
hyperlinks updated


akshayaurora:
Thanks for the contribution.

However hen you get a link like :class:`~kivy.input.motionevent.MotionEvent` we want the actual link to be to that class not to a specific implementation of it.

This would just further confuse people trying to figure out what a MotionEvent is.

PR #3703, opened 2015-10-17
Add a DynamicOptions setting item

yoelk:
It subclasses SettingOptions, with a modification that enables the
options to be dynamic. Instead of defining the options statically as a
list, a function path is given (attribute function_path) so that
function is called each time the user clicks on the drop-down list to
see the options.

It was developed for the case where an options setting was needed to
let a user choose between the currently available serial ports.


KeyWeeUsr:
Maybe should be closed, ref https://github.com/kivy/kivy/issues/4452#issuecomment-267651778 .

PR #3709, opened 2015-10-19
More intuitive handling of dirselect

ngoonee:
If dirselect=True, selecting '..' and clicking 'OK' would append .. to cwd instead of the currently viewed directory, this patch fixes that. Also makes directory selection more in line with standards with OS-specific file selectors, where doing a select when viewing a directory (no selections made) would select that directory.


dessant:
:+1:

Issue #3708, opened 2015-10-19
Impossible (or very hard) to traverse into/out-of folder when dirselect=True


ngoonee:
When using dirselect=True (a requirement if using the various file choosers to select a directory, especially in settings), clicking a directory once will select it rather than enter it, which is expected behaviour.

However, this means that if I want to traverse in to the directory (without using the arrow on the left) I would have to double click it, which would both go into the directory and select (and enter) the next directory in that same physical location on the screen.

Same issue (even more annoying) with the .. 'special' directory, if I'm in /a/b/c there's no way to actually go to /a/b, a double click brings me straight to /a

Looking at filechooser.py, I believe the reason for this is that in both lines 591 (in entry_touched) and line 623 (in entry_released) open_entry is called. This is not triggered when dirselect=False because line 604 is malformed (just open_entry with no function call).

Removing line 591 bugs out multiselect+dirselect, so I think line 623 should be removed. Also line 604 is obviously not doing anything and should be removed.

ngoonee:
PR been pulled, so this is fixed.

PR #3710, opened 2015-10-19
Fix directory selection double-selecting

ngoonee:
open_entry gets called twice when double-tapping (which is necessary to open/go into a directory with dirselect=True). Remove one of them.

Also remove a line which doesn't do anything (in the dirselect=False case) which prevented this bug from being more obvious, since it would only be hit in the dirselect=True codepath.

Linked to this issue (reported by me) - https://github.com/kivy/kivy/issues/3708


ngoonee:
Am surprised this wasn't merged for 1.9.1, is there something else I need to do?

dessant:
@ngoonee, thanks a lot.

PR #3711, opened 2015-10-21
Add scrollleft / scrollright to x11 backend.

stuaxo:
Add scroll left, scroll right to x11 backend.

Mappings taken from here

https://bbs.archlinux.org/viewtopic.php?id=194776

And tested with xev


dessant:
:+1:

stuaxo:
Thanks :D

PR #3712, opened 2015-10-21
Removed oltab

kiok46:
oltab was unnecessarily created and then removed without being used.


kiok46:
Where did I get wrong?

dessant:
It does not seem to be unnecessary, the default tab is removed if a new tab is present. Does your change solve an issue?

tito:
This break any consecutive call. Obviously, it's a no-op when the _default_tab is None (the first time), but required when it's not. The documentation also state that any previous default tab will be removed.

By doing this, you change the default behavior, and remove the convenience. I don't see what you tried to solve here :)

Issue #3707, opened 2015-10-19
Scrolling left and right with trackpad generates 'left' mousebutton events not 'scrollleft' 'scrollright'


stuaxo:
In Ubuntu, I'm testing the list widget examples, when I scroll up and down using the trackpad (I haven't tested on a mouse) it seems to actually press the buttons in all examples I've tried.

ngoonee:
That's an artifact of how trackpads work, you're probably using two-finger scroll or something. One finger is actually a bit faster than the other, that finger would register as an on_press before the other finger touches the trackpad.

Alternatives include using a different scroll method (not sure which would work, I use two-finger myself and do see this behaviour) or just putting a nice fat scroll bar on the Scroll Widget (requires you to modify ListView a bit) and grab that.

stuaxo:
I am indeed using the two finger scrolling... I think it's not checking which mouse button is being pressed.

If I run xev and do the same scrolling I see inputs for  button 4, 5 and 7  if I click I get button 1.


If think if the examples, or the api didn't respond to those higher buttons it would fix it.

Two goes on xev, 2nd one I actually click:
```
$ xev | grep "button "
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
    state 0x10, button 5, same_screen YES
    state 0x1010, button 5, same_screen YES
[...]

PR #3713, opened 2015-10-21
Add scrollleft / scrollright to sdl2 backend.

stuaxo:
Not sure why the action of the event for scrolling left/right is being interpreted as "mousebuttondown"/"mousebuttonup" only for scroll left/right, but it is.

This makes all the list examples useable when tested on Ubuntu.

### DIFF
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -416,9 +416,9 @@ class WindowSDL(WindowBase):
                     btn = 'right'
                 elif button == 2:
                     btn = 'middle'
-                if button == 6:  # fix for # 3707 - trackpad scrollleft gets to here
+                if button == 6:
                     btn = 'scrollleft'
-                elif button == 7:  # fix for # 3707 - trackpad scrollright gets to here
+                elif button == 7:
                     btn = 'scrollright'
                 eventname = 'on_mouse_down'
                 self._mouse_buttons_down.add(button)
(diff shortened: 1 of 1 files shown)


dessant:
Scollleft/right seems to be handled further down https://github.com/kivy/kivy/pull/3713/files#diff-5f05a9fe836a8283afe4a523b2038a32R437. Could you check how does the patch relate to that?

stuaxo:
I was a bit surprised that the code gets here, but then I'm not fully sure how this bit of code works.

Yep - the code doesn't seem to end up there for some reason, the action is always mousebuttondown, mousebuttonup, not mousewheeldown, mousewheelup for left/right scrolling so stays in the button section, never getting to the section where it is handled.

I don't know what decides between mousebutton or mousewheel - it might be that the fix needs to be there instead.

stuaxo:
I investigated further, putting debug in _window_sdl2.pyx and the wrong event type is indeed coming from SDL2.

I guess this is an upstream bug.

stuaxo:
Have asked a question on the SDL2 list  http://lists.libsdl.org/pipermail/sdl-libsdl.org/2015-October/099026.html

akshayaurora:
@stuaxo if this is a sadly bug,  I guess we should just make sure that we include SDL 2.0.4 in the next packaging and this should be fixed.

closing this and opening a issue for packaging.

@stuaxo thanks for all the work and research. :+1:

dessant:
It is not evident if 2.0.4 fixed the issue, nobody tested it yet. It would probably make sense to leave the issue open until 2.0.4 is confirmed to be fixed.

dessant:
Although there's already an issue for it. :P https://github.com/kivy/kivy/issues/3707

stuaxo:
Hi,
   I somehow missed the email for SDL 2.0.4, I've tested it just now and it works.

S

PR #3720, opened 2015-10-23
doc: update raspberry pi installation instructions

cbenhagen:
(wip) do not merge yet

I am still unhappy with the developer installation part. I think we need to explain more clearly when and how to install the developer version. It depends if one would want to develop kivy itself (get via git and install inplace) or just use the dev version right?


cbenhagen:
apparently @bionoid has some WIP docs cooking which "separates prerequisites for each platform from the actual compilation, and elaborates more on pythonpath/virtualenv etc"

akshayaurora:
@cbenhagen this is good enough for now, I am merging this,  if you have further additions do it as a separate pr.

akshayaurora:
:+1:

PR #3722, opened 2015-10-24
OSC to allow binary data

lipi:
Added typehint parameter to oscSend, to allow sending binary data. 

Without this typehint binary data may get changed. Found the issue when trying to send JPEG data.


tito:
thanks!

PR #3733, opened 2015-10-28
VLC mediaplayer integration example for android

ibobalo:
p4a vlc recipe required

### DIFF
--- a/examples/android/vlc/Makefile
+++ b/examples/android/vlc/Makefile
@@ -1,0 +1,70 @@
+NDK_PATH=$(HOME)/android-ndk-r10c
+ADK_PATH=$(HOME)/android-sdk-linux_r23.0.2
+API=19
+PKG=org.test.NativeVlcPlayer
+
+kivyvlc.apk: dist_vlc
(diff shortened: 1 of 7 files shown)


dessant:
I see you've made some pep8 commits, the log here contains a complete list of style issues to fix: https://travis-ci.org/kivy/kivy/jobs/88336479.

tito:
I don't feel that this should lend into kivy itself. It would be much better for you to commit in kivy-garden in order to maintain it later.

We're not going to handle any maintenance on that part, it's too specific i guess.

akshayaurora:
@ibobalo nice work!

However I share @tito 's sentiment. This is clearly a lot more than just a example.

The right place for this is inside a garden module, garden is essentially a extensions kind of a aggregate which aggregates user created widgets/modules/extensions.

dessant:
@ibobalo I've created https://github.com/kivy-garden/garden.androidvlc and added you as a maintainer with write access.

Feel free to push your work there, thanks a lot for your work so far. :cherries:

Issue #3739, opened 2015-10-31
Mention widget naming requirements


dessant:
Widget names must follow PEP8 in order to work in kv. We could specify this in `lang.rst` and/or `widget.py` and link to https://www.python.org/dev/peps/pep-0008/#class-names.

dessant:
Allowing any valid class name would be even better. @matham, would it require many changes?

emon10005:
hey i would like to work on it

tshirtman:
@emon10005 
Feel free to make a pull request to the documentation :).

PR #3740, opened 2015-10-31
Update lang.rst

emon10005:
Mention widget naming requirements #3739

### DIFF
--- a/doc/sources/guide/lang.rst
+++ b/doc/sources/guide/lang.rst
@@ -156,7 +156,8 @@ To have your display updated when your data change, you can now have just:
 
     GridLayout:
         cols: len(root.data)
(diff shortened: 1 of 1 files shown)


inclement:
Thanks for the contribution!

The specific requirement is that widgets start with a capital letter (and that properties do not) if they are to work when manipulated via kv language. Would you be able to modify your change to make this clear, since it isn't actually pep8 itself that has to be followed?

emon10005:
should i not mention pep8

inclement:
You could recommend it in the context of being a worthwhile style guide
to adhere to in any case, but I don't think it's essential.

dessant:
@emon10005, it is a good start, thanks a lot! :+1:

PR #3741, opened 2015-10-31
Fix scrollview issue when one do_scroll=False

SapienTech:
We still want to compute scroll distance in directions where
do_scroll_direction = False, since if the user continues to a drag,
they may start moving in a direction that is allowed to scroll.
This is particularly needed for mobile devices, as finger scrolls
often arc horizontally to vertically.


akshayaurora:
@SapienTech nice work! :+1:

PR #3745, opened 2015-11-03
 simplify and improve input filtering

akshayaurora:
Tests the whole string for conversion to int or float rather than just the incoming substring.

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -88,7 +88,7 @@ this function. You can either set it to `int` or `float` or set it to a method.
 For example, to write only in capitalized characters::
 
     capitalInput = TextInput()
-    
+
     def capitalize_input(substring, new_text):
         return substring.upper()
 
@@ -98,7 +98,7 @@ For example, to write only in capitalized characters::
 Or to only allow hex::
 
     hexInput = TextInput()
-    
+
     def capitalize_input(substring, new_text):
         if hex(new_text):
             return substring
@@ -537,7 +537,7 @@ class TextInput(FocusBehavior, Widget):
         if platform == 'linux':
             self._ensure_clipboard()
 
-    def _filter_input(self, substring, new_text):
+    def _filter_input(self, substring, new_text, from_del=False):
         # do tests here
         mode = self.input_filter
         if mode is not None:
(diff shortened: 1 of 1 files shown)


matham:
You're changing the behavior. Before, if the mode was int and the substring was `33d56`, what would end in the textinput was the old value plus `3356`. With your change, nothing will be added, including the valid numbers `3356`. Right? 

I think the old behavior is the desired one as we want the invalid strings to be filtered out, rather than removing everything including the potentially valid characters.

akshayaurora:
@matham,  yes,  invalid input will be discarded as a whole. This not while typing though as we return.
So this only happens when you paste or insert programatically `33d56` as that's a invalid input; we discard it as a whole.

We want invalid input to be filtered out not invalid parts of the input, we were doing the latter. I would not call the old behavior the desired one or accurate.

matham:
To me, it's about input conditioning to the correct format, in which case even when pasting we should still condition rather than reject. That's how I've seen it used I think.

akshayaurora:
I would think if one pastes hello9 interpreting that to be `9` is wrong.

dessant:
You've had a chat with @matham about passing more granular data to the callback (insert pos etc), will that be added in this pr?

When a different substring is returned than the one passed to the callback, `on_input_filtered` is not fired, but it does fire with `int` and `float` filtering. The init value of `text` should also be filtered.

I've used this:
```python
from kivy.app import App
from kivy.lang import Builder

kv = """
<Test>:
    TextInput:
        text: 'gfg9'
        on_input_filtered: print('filtered')
        input_filter: root.callback
"""

Builder.load_string(kv)


class Test(BoxLayout):

    def callback(self, *args):
        print(args)
        return 'bla'


class TestApp(App):
    def build(self):
        return Test()

if __name__ == '__main__':
    TestApp().run()
```

akshayaurora:
@dessant just set the filter before the text it will work.
If you set the filter after then it should not work.

Fixing the firing of on_input_filtered on change of substring...

dessant:
@akshayaurora, the text was moved after the filter declaration, but it is not filtered on init.

I've tested the latest state a bit, select `fg9` and press the backspace, the selected part will be deleted and it leaves us with `g` in the field, while `on_input_filtered` is fired correctly.

```python
from kivy.app import App
from kivy.lang import Builder

kv = """
<Test>:
    TextInput:
        on_input_filtered: print('filtered')
        # input_filter: root.callback
        input_filter: 'int'
        text: 'gfg9'
"""

Builder.load_string(kv)


class Test(BoxLayout):

    def callback(self, *args):
        print(args)
        return 'bla'


class TestApp(App):
    def build(self):
        return Test()

```

KeyWeeUsr:
Anyone can test the negative numbers? Would be nice to have it in the new stable as people requested this input_filter change a long ago.

dessant:
The pr has a number of issues and probably requires a new discussion with the team, I don't remember if the desired behavior was agreed upon in the end. Please reopen if/when you resume working on the feature.

Issue #71, opened 2011-02-26
Image: invalid ratio when displayed


tito:
Sometime, the ratio is not respected. That's depend of the ratio of the container and the ratio of the source image. Check all the case, and ensure all are working.

PR #3750, opened 2015-11-06
fix merge conflicts for loadidentity-fix

Kovak:
This fix looked good to me, just needed to be updated. #2756 was so old I figured it best to just do so myself.


dessant:
@Kovak `#71 Failure: NameError (name 'unittest' is not defined) ... ERROR`

akshayaurora:
@Kovak :+1:

PR #3767, opened 2015-11-14
Emacs textinput

el-ethan:
Add basic Emacs keyboard shortcuts for common actions and movements to TextInput. These bindings can be enabled by specifying `emacs` as the value for the `key_bindings` setting in the default `~/.kivy/config.ini` configuration file. More Emacs shortcuts, or other shortcuts can easily be defined later.


inclement:
This looks very nice, thanks. I'll wait for any implementation comments from others, but I didn't spot any problems on a quick look through.

akshayaurora:
@el-ethan it seems we forgot to move the doc along with the [code][1], most of the code for navigaion/custom bindings, key/mouse movements is already in behaviors, keeping TextInput Light and allowing people to mix `CodeNavigationBehavior` with `VimBehavior` or `EmacsBehavior`.

Could you please move the code from TextInput into a separate EmacsBehavior, this will also remove the need for multiple.

[1]:https://github.com/kivy/kivy/blob/master/kivy/uix/behaviors/codenavigation.py

el-ethan:
@akshayaurora that makes perfect sense! I will get everything moved over to behaviors and fix a few other problematic things I just found with my code. Thanks for the feedback!

el-ethan:
I am about ready to push the new EmacsBehavior code, but it seems like it might be cleanest to just do it from a new branch rather than undoing all the changes I made to textinput and config - none of which are necessary any more. What do you think? Either way is fine with me, and I will go with whatever workflow is preferred by those more experienced than myself.

inclement:
Do whatever seems sensible to you, it's fine to make a new PR if that
seems right.

If you do want to stay in this one, you can also just force push a new
branch with different changes, so that you have a new branch locally but
everything ends up in this PR (and the old stuff is overridden and
forgotten).

el-ethan:
OK, I have got everything moved into `EmacsBehavior`!

akshayaurora:
@el-ethan  nice work! :+1:

akshayaurora:
@el-ethan can we have a example in the examples folder? It could be added to examples/widgets/codeinput.py

el-ethan:
Thank you! I am proud to be part of the project, even in a small way! I will work on an example too.

PR #3772, opened 2015-11-16
Update OS X packaging docs

cbenhagen:
More to come ;) I will add a section on packaging using homebrew installed dependencies and pip installed kivy when i have fully tested it.


dessant:
:+1:

Issue #2297, opened 2014-06-27
Scrollview blurs content


NumesSanguis:
Some uses of the ScrollView blurs the content in it. Also when clicking buttons it only works once every 5 clicks. Probably because the content is set to non-integer points like 1,0000001 or something and some rendering kicks in.

I posted this on the google groups: https://groups.google.com/forum/#!topic/kivy-dev/dthCWk8lzCs
Code:
http://dpaste.com/3FVZ9N6/
Screenshot:
https://i.imgur.com/bmStIbT.png
Try (un)commeting size_hint_y: None to see the blurry effect yourself.

Here some small code to try it:
https://gist.github.com/NumesSanguis/3e0282bc8252cd86b76b

Same code without ScrollView:
https://gist.github.com/NumesSanguis/ded56793795ed79697ce

pyfrid:
[INFO              ] [Kivy        ] v1.9.1-dev
[INFO              ] [Python      ] v2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2]
[INFO              ] [Factory     ] 177 symbols loaded
[INFO              ] [Image       ] Providers: img_tex, img_dds, img_gif, img_sdl2, img_pil (img_ffpyplayer ignored)
[INFO              ] [Text        ] Provider: sdl2
[INFO              ] [OSC         ] using <multiprocessing> for socket
[INFO              ] [Window      ] Provider: sdl2(['window_egl_rpi'] ignored)
[INFO              ] [GL          ] OpenGL version <3.0 Mesa 10.1.3>
[INFO              ] [GL          ] OpenGL vendor <Intel Open Source Technology Center>
[INFO              ] [GL          ] OpenGL renderer <Mesa DRI Intel(R) Sandybridge Desktop >
[INFO              ] [GL          ] OpenGL parsed version: 3, 0
[INFO              ] [GL          ] Shading version <1.30>
[INFO              ] [GL          ] Texture max size <8192>
[INFO              ] [GL          ] Texture max units <16>
[INFO              ] [Window      ] auto add sdl2 input provider
[INFO              ] [Window      ] virtual keyboard not allowed, single mode, not docked
![screenshot from 2015-10-12 20 26 37](https://cloud.githubusercontent.com/assets/965931/10435477/855c62b6-7120-11e5-9d8e-cbceeb61ebc8.png)

dessant:
I've seen this many times with canvas lines, not just in a scrollview.

[...]

PR #3778, opened 2015-11-19
fixing issue #2297 with blurred labels in scrollview

pyfrid:
1) changing min and mag filter of label texture to nearest in kivy.core.text
2) rounding scrollview translate coordinates to integer in kivy.uix.scrollview

### DIFF
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -646,8 +646,8 @@ class LabelBase(object):
                                      mipmap=self.options['mipmap'],
                                      callback=self._texture_fill)
             #switching off linear interpolation
-            texture.mag_filter="nearest"
-            texture.min_filter="nearest"
+            texture.mag_filter = "nearest"
+            texture.min_filter = "nearest"
             texture.flip_vertical()
             texture.add_reload_observer(self._texture_refresh)
             self.texture = texture
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -645,6 +645,9 @@ class LabelBase(object):
             texture = Texture.create(size=(width, height),
                                      mipmap=self.options['mipmap'],
                                      callback=self._texture_fill)
+            #switching off linear interpolation
+            texture.mag_filter="nearest"
+            texture.min_filter="nearest"
             texture.flip_vertical()
             texture.add_reload_observer(self._texture_refresh)
             self.texture = texture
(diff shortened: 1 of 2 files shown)


inclement:
I don't think the rounding to integer is desirable, as it will lead to jerky motion on lower-resolution displays. Maybe it could be an option.

pyfrid:
Without rounding, text will be pixelized if position of scrollview has been changed.

inclement:
That's why I suggest making it an option - both rounding and not rounding are apparently desirable modes.

dessant:
If we go the config option way, one that also affects all pos/size props would be great, because blurring caused by subpixel rendering is a problem with all textures, not just text.

dessant:
It is also not just a problem inside scrollview, but it can appear anywhere, commonly when `size_hint` is used.

pyfrid:
I propose a general config option for the default filter for all textures and a boolean property for scrollview for translation rounding. Concerning common problem everywhere where size_hint is used I need examples. In my large application with a lot of layouts and fixed size widgets I have issue only with scrollview. Even in my example to the corresponding issue tab labels become blurred because they are placed to scrollview by default with horizontal scrolling. 

I will work on that during weekends.

tito:
Changing the mag filter or using int will mislead the resolution of this issue. I mean, as an user, if i activate this option, i expect that it will be not blurry and works. But it wont, if the widget tree is not pixel aligned.

I think we can make the distinction between 2 phases:
- when something animate: we want smooth animation, int is not desirable, as well as mag filter.
- when it's static: we want pixel-perfect rendering, no blur.

This could be resolved by a total another approach, more general:
1. ensure all the layout are computing pixel perfect layout, no half pixel: ie, use int() when it should be, as long as the width or height is an integer as well (we can't have pixel perfect calculation if the width/height are not integer). size_hint could compute integer width/height as well.
2. ensure any scrollview will end the scrolling to a pixel, not a portion of it. So when moving / animating, stay as today, but when we reach the end of the movement (delta too low to say there is a movement), then move to the nearest pixel in the direction of the movement.

That should cover all the standard case. The one that wont work will be the one when user use its own matrix and do it's own translation.

tito:
Changing the label rendering and using integer will break all the current animation or usage of subpixel application, so it's not acceptable :)

But thanks for your first contribution!

pyfrid:
Thank you Mathieu, for clear explanation. I will try your proposal.

Issue #3717, opened 2015-10-22
release packaging fix to pypi as 1.9.0.post0


cbenhagen:
see https://github.com/kivy/kivy/pull/3682

PR #3782, opened 2015-11-20
OS X: pip install kivy from git

cbenhagen:
This is needed until https://github.com/kivy/kivy/issues/3717 can be closed.


akshayaurora:
@cbenhagen thanks.

PR #3784, opened 2015-11-20
[WIP] Fixing blurred text

pyfrid:
The following changes has been done:

1) rounding pos and size of layouts children
2) rounding last scrollview translation values after Animation has been finished

Still not sure about Accordion and TreeView.
Blurring of textures as expected appears only during animations.

### DIFF
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -646,8 +646,8 @@ class LabelBase(object):
                                      mipmap=self.options['mipmap'],
                                      callback=self._texture_fill)
             #switching off linear interpolation
-            texture.mag_filter="nearest"
-            texture.min_filter="nearest"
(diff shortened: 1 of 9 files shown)


dessant:
Related issue: https://github.com/kivy/kivy/issues/2297#issuecomment-158348811.

dessant:
@tito, could you look into this?

KeyWeeUsr:
Seems like a really bad approach to me if it should be the default one mainly because using `Animation` could cause issues. On the other hand, we could perhaps have a Config/Widget option to have integers for `pos` only with a big warning?

tito:
I'm closing this PR.

It won't work before it break the Animation by forcing everything to be an int. The right resolution of the issue is much harder, like every widget should be pixel-aligned but can animate smoothly.

Issue #3773, opened 2015-11-16
Stop marketing the .app packaging method on osx as official


dessant:
We have pyinstaller, which does not have any outstanding issue documented, people seem to be using it with success.

We are well aware that the .app method is not a real solution for packaging in its current state, it lacks many features pyinstaller has, and the output also depends on the target system's python.

If pyinstaller is indeed something that should be avoided on osx, I would like to see issues being opened for our repo, documenting how and why packaging with it does not work. Maybe we can find fixes for those problems and resume recommending a real, established packager.

akshayaurora:
https://github.com/kivy/kivy/issues/1908

akshayaurora:
@dessant I guess this assigned to you since you think this is solvable with pyinstaller there ws a bounty of 50$ plus people ready to offer 500$, best of luck.

dessant:
I don't think it is solvable, but I do believe in listing our problems and documenting them.

cbenhagen:
I do think this is solvable. Here is a proof of concept:
https://www.dropbox.com/s/8q770mccsv57dkp/ogg_test.zip?dl=1

Packaged with pyinstaller using homebrew dependencies.

dessant:
@cbenhagen does ogg work (can't test)? What was missing on #1908 when packaging?

cbenhagen:
Yes it does. It plays an ogg file from wikipedia. Hard to tell what went wrong there. But it will be straightforward in the future.

tito:
You can have binary clash at any moment if you mix our Kivy.app release and your own dependencies with homebrew. That's what happened on few computer. There is people that actually use Kivy.app and drag&drop, others use brew.

There is no point to remove Kivy.app, not all developers use brew as the only way to package. And it's not because the documentation is not finished that it must be killed, i completely disagree here.

And no, i'm also having trouble to have a correct package with pyinstaller, and it's using me too much time than just copy&paste for fast release to a customer. And my last point: it's more newbie friendly (in the concept). If we are not here yet, it's mostly just a documentation issue, which can be resolved easily.

Let's don't kill the concept.

akshayaurora:
@cbenhagen did you test the app on another sys than the one you packaged on? The test.app does not work for me. I hear no sound.

bionoid:
ogg playback works for me on 10.10.5, I don't have any relevant libraries installed with brew etc

akshayaurora:
My system is 10.10.5 too I don't have brew installed. No Sound :/

[...]

Issue #2817, opened 2015-01-02
SDL2 OSX packaging should use latest master to enable on_dropfile event to work properly on macs


akshayaurora:
SDL2 OSX packaging should use latest master to enable on_dropfile event to work properly on macs.

tito:
I vote in favor to delay this one, and wait the official frameworks. As we don't use source code anymore.

akshayaurora:
+1 retagging

akshayaurora:
2.0.4 rc is out we should be using that in absence of a release for 1.9.1

akshayaurora:
New Packaging uses 2.0.4

PR #3785, opened 2015-11-20
OS X packaging

### LINKED ISSUE #3667: Update pyinstaller packaging guide for OS X
http://kivy.org/docs/guide/packaging-macosx.html#unofficial-method-using-pyinstaller

PyInstaller 3.0 was just released, the section with the tree inclusion should also be revisited, the linked issue seems to be solved on our part.
https://github.com/pyinstaller/pyinstaller/issues/1338

cbenhagen:
Release early, release often they say. So here is my somewhat unpolished proposition to fix  #3667, #3773, and #2817. I know this is a controversial subject. But i feel very positive about documenting and hopefully fixing any problems that may arise with the usage of homebrewed dependencies. I really like this way for many reason like the Python 3 or OpenCV support.

### DIFF
--- a/doc/sources/guide/packaging-osx.rst
+++ b/doc/sources/guide/packaging-osx.rst
@@ -1,82 +1,89 @@
 Creating packages for OS X
 ==========================
 
-Packaging your application for the OS X 10.6 platform can only be done inside OS X.
-
-The package will only work for the 64 bit OS X. We no longer support 32 bit OS X platforms.
-
 .. note::
-    Currently, packages for OS X can only be generated with Python 2.7. Python 3.3+ support is on the way...
+    - Packaging your application for the OS X platform can only be done inside OS X.
 
-.. _osx_requirements:
 
-Official Packaging method
--------------------------
+Using kivy-sdk-packager
+-----------------------
+.. note::
+    - This is currently the preferred way to package an app if you are new to Kivy
+      and/or you know that you do not need newer or more libraries than provided by the Kivy.app
+    - The package will only work for the 64 bit OS X. We no longer support 32 bit OS X platforms.
+    - Currently, packages built using this method can only be generated with Python 2.7.
+      Python 3.3+ support is on the way...
 
-Due to a lot of problems with including libraries and files on OS X with other methods
-we now have a simpler and easier way to package Kivy apps on OS X.
+Since Kivy 1.9 Kivy package on OS X is a self contained portable distribution.
+It is now possible to package Kivy apps using the method described below to make
+it easier to include frameworks like SDL 2 and GStreamer.
 
-Since kivy 1.9 kivy package on OS X is a self contained portable distribution.
-It is now possible to package kivy apps using the method described below to make
-it easier to include frameworks like sdl2 and gstreamer::
+1. Make sure you have the unmodified Kivy.app from the download page.
 
-1: Make sure you have the Kivy.app(unmodified) from the download page.
-2: run the following commands::
+2. Run the following commands::
 
-    > mkdir  packaging
+    > mkdir packaging
     > cd packaging
     packaging> git clone https://github.com/kivy/kivy-sdk-packager
     packaging> cd kivy-sdk-packager/osx
     osx> cp -a /Applications/Kivy.app ./Kivy.App
 
-This step above is important, you have to make sure to preserve the paths and permissions. A command like cp -rf will copy but make the app unusable and lead to error later on.
+  .. note::
+    - This step above is important, you have to make sure to preserve the paths and permissions.
+      A command like ``cp -rf`` will copy but make the app unusable and lead to error later on.
 
-Now all you need to do is to include your compiled app into the Kivy.app, simply run the following command::
+3. Now all you need to do is to include your compiled app into the Kivy.app. Simply run the following command::
 
-    osx> ./package-app.sh path/to/your/app
+    osx> ./package-app.sh /Applications/Kivy.app/Contents/Resources/kivy/examples/demo/touchtracer/
 
-This should copy Kivy.app to `yourapp.app` and include a compiled copy of your app into this package.
+  This copies Kivy.app to `touchtracer.app` and includes a compiled copy of the demo app into this package.
 
-That's it, your self contained package is ready to be deployed!
+4. That's it, your first self contained package is ready to be deployed!
+   You can now further customize your app as described bellow.
 
-when you double click this app you can see your app run.
 
+Shrink the app size
+^^^^^^^^^^^^^^^^^^^
 This is a pretty big sized app right now however you can simply remove the unneeded parts from this package.
 
-For example if you don't use Gstreamer, simply remove it from YourApp.app/Contents/Frameworks.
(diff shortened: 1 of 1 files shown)


doc/sources/guide/packaging-osx.rst
```diff
@@ -1,82 +1,192 @@
...
 
-Packaging your application for the OS X 10.6 platform can only be done inside OS X.
-
-The package will only work for the 64 bit OS X. We no longer support 32 bit OS X platforms.
-
 .. note::
-    Currently, packages for OS X can only be generated with Python 2.7. Python 3.3+ support is on the way...
+    - Packaging your application for the OS X platform can only be done inside OS X.
 
-.. _osx_requirements:
```
akshayaurora: Why remove the link?

doc/sources/guide/packaging-osx.rst
```diff
@@ -1,82 +1,192 @@
...
 
-Official Packaging method
--------------------------
+Using kivy-sdk-packager
```
akshayaurora: Nope, there was a reason for the exact words. Official packaging method == the one we support and expect every one to use.

doc/sources/guide/packaging-osx.rst
```diff
@@ -1,82 +1,192 @@
...
+-----------------------
+.. note::
+    - This is currently the preferred way to package an app if you are new to Kivy
+      and/or you know that you do not need newer or more libraries than provided by the Kivy.app
```
akshayaurora: Nope,  this is not for new users and one can easily include other libraries that was the whole point of this new method,  to make packaging external libs like streamer easier.

doc/sources/guide/packaging-osx.rst
```diff
@@ -1,82 +1,192 @@
...
+    - The package will only work for the 64 bit OS X. We no longer support 32 bit OS X platforms.
+    - Currently, packages built using this method can only be generated with Python 2.7.
+      Python 3.3+ support is on the way...
```
akshayaurora: We simply don not support older OsX 32bit, not just for packaging but for kivy itself too. has nothing to do with current packaging method.

The limitation to python2.7 was for pyinstaller that didn't include py3 support instill recently not for this packaging method.

doc/sources/guide/packaging-osx.rst
dessant: @akshayaurora we support both packaging methods.
@cbenhagen please leve that section intact, i will rephrase that part.

doc/sources/guide/packaging-osx.rst
```diff
@@ -1,82 +1,192 @@
...
-This step above is important, you have to make sure to preserve the paths and permissions. A command like cp -rf will copy but make the app unusable and lead to error later on.
+  .. note::
+    - This step above is important, you have to make sure to preserve the paths and permissions.
+      A command like ``cp -rf`` will copy but make the app unusable and lead to error later on.
 
-Now all you need to do is to include your compiled app into the Kivy.app, simply run the following command::
+3. Now all you need to do is to include your compiled app into the Kivy.app. Simply run the following command::
 
-    osx> ./package-app.sh path/to/your/app
+    osx> ./package-app.sh /Applications/Kivy.app/Contents/Resources/kivy/examples/demo/touchtracer/
```
akshayaurora: path/to/your.app is pretty self explanatory, using your mentioned method will confuse users where one might think that you can only use apps inside kivy/app

doc/sources/guide/packaging-osx.rst
```diff
@@ -1,82 +1,192 @@
...
 
-This should copy Kivy.app to `yourapp.app` and include a compiled copy of your app into this package.
+  This copies Kivy.app to `touchtracer.app` and includes a compiled copy of the demo app into this package.
```
akshayaurora: Same as above there is no reason for this change and is more confusing than before.

doc/sources/guide/packaging-osx.rst
```diff
@@ -1,82 +1,192 @@
...
+    osx> ./create-osx-dmg.sh YourApp.app
 
 This should give you a compressed dmg that will even further minimize the size of your distributed app.
 
 
+Using PyInstaller and Homebrew
+------------------------------
+.. note::
+    - Only use this method if you are an advanced user and know that you need more
+      or newer libraries than Provided by the Kivy.app or Python 3 support.
```
akshayaurora: This is miss leading again, as mentioned above the whole point of Kivy.app method was to make inclusion of custom libs easier.

akshayaurora: if you compile your Kivy.app using kivy-sdk-packager using python 3 you will have a python3 package,  just like with pyinstaller. So Kivy.app method does support python3

bionoid: @akshayaurora but that won't work unless the target system has a suitable py3 interpreter installed, right? If pyinstaller bundles it (not sure if this is the case), that may be preferable

akshayaurora: One can use python3 from python.org, point is the method does not prevent you from deploying a python3 app

doc/sources/guide/packaging-osx.rst
akshayaurora: @dessant we don't officialy support it cause we can't , how will you support users if pyinstaller stops working tomorrow ? We can only support methods that we have control over.

doc/sources/guide/packaging-osx.rst
akshayaurora: @bionoid,  you are right in that we do not package python binary itself,  I think this should be added as a option to the new packaging method.  Infact we should document that the Kivy.app method does not package the python binary itself but uses system binary with the `kivy` venv.

doc/sources/guide/packaging-osx.rst
cbenhagen: At first I thought it was not being used anymore. But then I found the duplicate https://github.com/kivy/kivy/blob/master/doc/sources/gettingstarted/packaging.rst which I think should either be removed or have the same content as https://github.com/kivy/kivy/blob/master/doc/sources/guide/packaging.rst where the TOC is autogenerated and does not use the label.

akshayaurora:
@cbenhagen,  nice work, though somethings are a bit misleading,  I guess I could merge then fix these myself.

doc/sources/guide/packaging-osx.rst
cbenhagen: Then I really missed that point. We should absolutely document how this can be done. Maybe with an example of a newer library. Let's take SDL 2 here too.

doc/sources/guide/packaging-osx.rst
cbenhagen: Ok so let's document how to use and package Python 3 projects with the app method.

doc/sources/guide/packaging-osx.rst
cbenhagen: Packaging touchtracer is what we document for every other OS too. And I think it is quite nice to have an example everyone uses.

PR #3786, opened 2015-11-21
Autogenerate TOC

cbenhagen:


akshayaurora:
We are using the custom refs in many places for direct referencing , not just for toc,  this will stop that from working.

cbenhagen:
Even if we do use those labels anywhere else it does not matter in this PR. We would need to add it in the documentation not here.

akshayaurora:
this is removing those refs from doc,  any reason why?

cbenhagen:
Because they point to labels which are not there anymore. We could either add back those labels or just do as we do here 
https://raw.githubusercontent.com/kivy/kivy/master/doc/sources/guide/packaging.rst and let this TOC get autogenerated.

akshayaurora:
@cbenhagen https://raw.githubusercontent.com/kivy/kivy/master/doc/sources/guide/packaging-windows.rst these labels exist.

cbenhagen:
Thats not the point. We can also add the missing ones. But this will just break very easily again. Why do want to create and maintain this list here by hand when you could do it automatice??

akshayaurora:
Cause automatically we don't get a list of all the custom refs we want.
The only missing ref is for `mac_Create-the-spec-file` which should actually added and linked to.

 We also do not want all refs in the doc in the TOC, we only want some of them that are important, some of the refs are used only for direct linking from other docs.

I'll add the `mac_Create-the-spec-file` reference back manually.

cbenhagen:
All that those ``:ref:`` do is create a link to the section. Its not that you can use them from anywhere else. For that a label is what you want to use. But if your point is that you want more control on what gets into this list here but do want to get it autogenerated on the same list here http://kivy.org/docs/guide/packaging.html then please tell me why and what you want on those lists so i can either search for a nicer way to create them or at least document for me somewhere that when i update the documentation that i have to update this list too and what exactly should go in there and what should not..

akshayaurora:
You have to be aware of the existing labels,  that's all, don't just remove them. Yes,  we are using :ref: to point to those custom labels that's the whole point. 

Ideally we want every important section listed there. what's important is decided by whether a user would want to skip to that section directly.

PR #3789, opened 2015-11-21
OS X Packaging: Fix PyInstaller method

cbenhagen:


doc/sources/guide/packaging-osx.rst
```diff
@@ -108,54 +108,65 @@ Complete guide
...
+    $ export HOOKSPATH=`python -c "
+    import imp, os
+    print(os.path.join(imp.find_module('kivy')[1], 'tools/packaging/pyinstaller_hooks'))"`
+
 #. Package your app using the path to your main.py::
 
-    $ pyinstaller -y --clean --windowed --name touchtracer /usr/local/share/kivy-examples/demo/touchtracer/main.py
+    $ pyinstaller -y --clean --windowed --name touchtracer \
+      --additional-hooks-dir $HOOKSPATH \
+      --runtime-hook $HOOKSPATH/rt-hook-kivy.py \
```
akshayaurora: what was wrong with there earlier method of using runtime hooks from kivy/tools/packaging/pyinstaller_hooks ?

akshayaurora: In other words what are the advantages of doing it this way?

cbenhagen: - It's easier to document in as the user can just copy paste the command instead of having to add the lines.
- Less prone to errors on the users side.
- Easier scriptable

Also if he needs to add more hooks he could do it on the same command line. Otherwise he would have to write code in the spec file which extends the list which is returned by ``get_hooks()``

cbenhagen:
Please do not merge yet:

I will squash the commits at the end so not to mess up the discussion.

doc/sources/guide/packaging-osx.rst
cbenhagen: @akshayaurora what do you think?

dessant: +1 for passing the hook in the command, this method will make the most sense once the kivy hook gets merged in pyinstaller.

cbenhagen:
@dessant (let's better discuss this https://github.com/kivy/kivy/pull/3809#issuecomment-161953609 here)
No the discussion here is about how the hooks are loaded at all. If they are passed as commandline arguments or hacked into the spec. I explicitly call it a hack here because it is not the way anyone using PyInstaller would expect it and also it was mentioned by a PyInstaller developer that this would break the users ability to add additional hooks.

dessant:
@cbenhagen i agree and i'm in favor of this change.

dessant:
To avoid yet another update after the pyinstaller hook merge, please document the usage and overriding as if it was already merged. :)

doc/sources/guide/packaging-osx.rst
```diff
@@ -103,76 +98,62 @@ Complete guide
...
-     $ brew reinstall --build-bottle https://github.com/cbenhagen/homebrew/raw/patch-3/Library/Formula/gst-python.rb
+    $ pip install -I Cython==0.21.2
+    $ USE_OSX_FRAMEWORKS=0 pip install git+https://github.com/kivy/kivy.git@1.9.0
 
 #. Install the development version of PyInstaller which includes fixes for the
    GStreamer hooks::
 
     $ pip install git+https://github.com/pyinstaller/pyinstaller.git@develop
 
+#. Export the ``HOOKSPATH`` environment variable::
```
dessant: Is the env var also used by pyinstaller, or just by this example? In the latter case, I would find it better to avoid this indirection and explain that the hook already comes with pyinstaller, but optionally, a custom one can be used, like the one shipped with kivy.

cbenhagen: No it is only used by us. And yes I agree that as soon as the hook is in PyInstaller we can move this part to an advanced section where we explain how the provided hooks can be overridden and why one would want to do this.

PR #3802, opened 2015-11-28
Replace print on Logger.info

ivlevdenis:
Small fix.
```
[INFO   ] [Shader      ] fragment shader: <Compiled>
[INFO   ] [Shader      ] vertex shader: <Compiled>
adding binding for mouse move   <<<<<<
```


dessant:
:+1:

PR #3806, opened 2015-11-30
Add EmacsBehavior to CodeInput and add example

el-ethan:
Have CodeInput inherit EmacsBehavior and add "Key bindings" option to CodeInputTest so that users can try out the bindings.

### DIFF
--- a/examples/widgets/codeinput.py
+++ b/examples/widgets/codeinput.py
@@ -3,12 +3,13 @@ from kivy.extras.highlight import KivyLexer
 from kivy.uix.spinner import Spinner, SpinnerOption
 from kivy.uix.boxlayout import BoxLayout
 from kivy.uix.codeinput import CodeInput
+from kivy.uix.behaviors import EmacsBehavior
 from kivy.uix.popup import Popup
 from kivy.properties import ListProperty
 from kivy.core.window import Window
 from kivy.core.text import LabelBase
 from pygments import lexers
-
+from pygame import font as fonts
 import codecs
 import glob
 import os
@@ -109,6 +110,13 @@ class SaveDialog(Popup):
         self.dismiss()
 
 
+class CodeInputWithBindings(EmacsBehavior, CodeInput):
+    '''CodeInput with keybindings.
+    To add more bindings, add behavior before CodeInput in class definition
+    '''
+    pass
+
+
 class CodeInputTest(App):
 
     files = ListProperty([None, ])
@@ -154,11 +162,12 @@ class CodeInputTest(App):
         menu.add_widget(key_bindings)
         b.add_widget(menu)
 
-        self.codeinput = CodeInput(
-
+        self.codeinput = CodeInputWithBindings(
             lexer=KivyLexer(),
             font_size=12,
-            text=example_text)
+            text=example_text,
+            active_key_bindings='default',
+        )
 
         b.add_widget(self.codeinput)
 
--- a/examples/widgets/codeinput.py
+++ b/examples/widgets/codeinput.py
@@ -57,6 +57,26 @@ public static byte toUnsignedByte(int intVal) {
     s.parentNode.insertBefore(po, s);
   })();
 </script>
+----------------------Emacs key bindings---------------------
+This CodeInput inherits from EmacsBehavior, so you can use Emacs key bindings
+if you want! To try out Emacs key bindings, set the "Key bindings" option to
+"Emacs". Experiment with the shortcuts below on some of the text in this window
+(just be careful not to delete the cheat sheet before you have made note of the
+commands!)
+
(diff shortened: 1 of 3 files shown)


kivy/uix/codeinput.py
```diff
@@ -85,6 +85,13 @@ class CodeInput(CodeNavigationBehavior, TextInput):
 
     '''
 
+    active_key_bindings = StringProperty('default')
+    '''String name which determines the type of key bindings to use.
+
+    :attr:`active_key_bindings` is a :class:`~kivy.properties.StringProperty`
+    and defaults to ``'default'``
+    '''
```
akshayaurora: ..versionadded:: 1.9.2

kivy/uix/behaviors/emacs.py
```diff
@@ -44,6 +47,17 @@ class EmacsBehavior(object):
     keyboard shortcuts for the :class:`~kivy.uix.textinput.TextInput` widget.
     '''
 
+    active_key_bindings = StringProperty('emacs')
+    '''String name which determines the type of key bindings to use with the
+    :class:`~kivy.uix.textinput.TextInput`. This allows Emacs key bindings to
+    be enabled/disabled programmatically for widgets that inherit from
+    :class:`EmacsBehavior`. If the value is not ``'emacs'`` Emacs bindings will
+    be disabled.
+
```
akshayaurora: ..versionadded...

akshayaurora:
@el-ethan I think the vim side if this debate is absent... :)

Which behavior to have by default for code input... I feel this would be better served as a garden widget, where we can have EmacsCodeInput VIMCodeinput... instead of limiting to one type of behavior let's keep the core widget open for re-usability.

akshayaurora:
Oh, btw nice job!

dessant:
This repo is a great place for the behavior itself, but i agree with @akshayaurora that we should not enable it for codeinbut by default. Documenting the usage in the behavior docstring is enough.

el-ethan:
@akshayaurora, do you think the vim community would mind if VIMBehavior inherited from EmacsBehavior? Seriously though, I think it would be great if there was a VIMBehavior mixin as well - but better that it be written by a vim user :)

Also, even though CodeInput inherits from EmacsBehavior in this branch, by default the behavior doesn't change from that of kivy master. It will only change if you pass a different value for `active_key_bindings`. 

The reason why I initially added it to the CodeInput is that I was asked to add an example of EmacsBehavior to `examples/widgets/codeinput`, and this was what I came up with at the time. The `active_key_bindings` property allows the behavior of the CodeInput to be changed programmatically so that you don't have to decide what kind of behavior you want to use ahead of time. VIMBehavior could be added to CodeInput later, and used the same way. 

However, I am not so attached to this example, so if you want me to make some changes, or simply close this pull request, it is fine with me. Probably an example is not necessary anyway, since anyone who will use EmacsBehavior is likely already familiar with Emacs.

kivy/uix/codeinput.py
```diff
@@ -52,7 +52,7 @@
 # TODO: color chooser for keywords/strings/...
 
 
-class CodeInput(CodeNavigationBehavior, TextInput):
+class CodeInput(CodeNavigationBehavior, EmacsBehavior, TextInput):
```
dessant: @el-ethan, we would like to merge your work. Could you remove `EmacsBehavior, ` from this line?

el-ethan:
I removed `EmacsBehavior` from `CodeInput` but I added made a few minor changes so that the example in `examples/widgets/codeinput` still works and can be further extended in the future with VIMBehavior, PicoBehavior ;),  etc. I can still remove the example altogether, if necessary.

dessant:
@el-ethan, thanks a lot! :+1:

dessant:
I've cleaned up a bit in https://github.com/kivy/kivy/pull/3894.

The pr limits EmacsBehavior (feature released in 1.9.1) usage to CodeInput only, before it could also be used with TextInput.

@team, should we move CodeInput.key_bindings to TextInput and preserve the original behavior?

examples/widgets/codeinput.py
```diff
@@ -3,12 +3,13 @@
...
 from kivy.uix.boxlayout import BoxLayout
 from kivy.uix.codeinput import CodeInput
+from kivy.uix.behaviors import EmacsBehavior
 from kivy.uix.popup import Popup
 from kivy.properties import ListProperty
 from kivy.core.window import Window
 from kivy.core.text import LabelBase
 from pygments import lexers
-
+from pygame import font as fonts
```
tito: This should be removed

dessant:
@tito, it was removed.

dessant:
ref https://github.com/kivy/kivy/pull/3898

PR #3809, opened 2015-11-30
Merge the additional libraries section into the main pyinstaller guide

dessant:


dessant:
@cbenhagen regarding this patch https://github.com/cbenhagen/homebrew/raw/patch-3/Library/Formula/gst-python.rb, is it something that can be fixed upstream?

dessant:
A different thing that not clear, is it feasible to use the system python and frameworks (no brew) for pyinstaller packaging.

dessant:
@cbenhagen could you also integrate the second part of the pyinstaller section into the numbered steps? I understand that it is a leftover from a previous version, but it becomes rather confusing.

cbenhagen:
@dessant i agree that the guide is rather confusing right now. There are several leftovers and stuff that is in wrong order and not easy to follow.

Some of this is long fixed but is waiting on decisions:
https://github.com/kivy/kivy/pull/3789

I am not sure how to proceed here. Maybe separate the fixes in two PRs so we can have a better guide online soon and discuss the remaining issue later?

cbenhagen:
Yes it would should be no issue to use system python and or frameworks installed by hand. But this gets really messy to document. We could maybe leave a note that this is possible.

dessant:
@cbenhagen i've skimmed the discussion there, if the main argument is about pytinstaller provided hook vs local one, we should just document both and explain the use cases, how can the library hook be overridden etc. The pyinstaller hook needs testing with wheels, but i feel we will be in favor of merging it.

dessant:
@cbenhagen i agree that maybe we should not open that bottle for now :D. Let's have pyinstaller + brew cleaned up then.

PR #3816, opened 2015-12-02
Support for SDL2's outline rendering

vitorio:
Hi!  Please consider this pull request a starting point for a conversation; while the code works, I'm not sure if I've developed it (particularly the `kivy.uix.label` portion) to project norms or standards.  I'm happy to make changes as recommended and re-create the pull request.

A month and a half ago, someone on _kivy-users_ asked about rendering text with outlines: https://groups.google.com/forum/#!msg/kivy-users/IxH1yT6plAU/VAikQ1FaBAAJ

A few suggestions were made, and the user ended up going with repeated renderings using canvas.before.  I had the same need and experimented with all of the options suggested: http://imgur.com/a/eqpz5

Of the three text renderers in Kivy, only SDL2 supports outline rendering, and you still have to render the font twice if you need the outline color to be different from the font color, which I assume is the most common use case: http://www.archivum.info/sdl@lists.libsdl.org/2010-01/00605/Re-(SDL)-Getting-outlined-text-in-SDL_TTF.html

This code adds support for SDL2's outline rendering, through three properties added to `kivy.core.text`: outline, outline_color, and outline_size.  It handles the outlining in `_text_sdl2.pyx`, and adds support to `kivy.uix.label`.

I'm currently using this code in an application, having also added support for this in `kivy.garden.scrolllabel`.

Looking forward to your comments!

Thanks,
Vitorio

### DIFF
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -46,6 +46,7 @@ from kivy.core.text.text_layout import layout_text, LayoutWord
 from kivy.resources import resource_find, resource_add_path
 from kivy.compat import PY2
 from kivy.setupconfig import USE_SDL2
+from kivy.metrics import dp
 
 DEFAULT_FONT = 'Roboto'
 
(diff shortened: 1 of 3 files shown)


vitorio:
Just tested this on a Retina display, and the outline_size seems to be much smaller than it is on a non-Retina display.  Probably need some sort of DPI compensation factor.  Where might that come from?

pythonic64:
Try changing: outline_size = NumericProperty('1dp')

dessant:
@vitorio thanks for your contribution. :+1: Would you rename `outline_size` to `outline_width`?

vitorio:
@dessant Sure. By version tag, do you mean the e.g. `.. versionchanged:: 1.5.0` comments?  If so, what version number should I use?

dessant:
It should be `.. versionadded:: 1.9.1`.

akshayaurora:
should be 1.9.2 we are in feature freeze for 1.9.1

dessant:
@vitorio, sorry for the late comeback, we are ready to test the pr. Could you rebase it?

vitorio:
Hi, okay!  This rebase assumes that @ali65's kerning/blending changes should be used in outline modes.  I'm not actually sure this is the case!  Perhaps they can confirm/deny.

dessant:
@vitorio, I've made some minor changes to the api and merged your code by preserving the commits here: https://github.com/kivy/kivy/pull/4265. Thanks again for taking the time to contribute.

PR #3817, opened 2015-12-03
Fix collasped accordionItem bug

jpoyau:
Fix bug where collapsed accordionItem children widgets were getting
on_touch_up and on_touch_move events.

[Gist](https://gist.github.com/jpoyau/62c67dd8f50c134bc76c) test file that shows the issue.

### DIFF
--- a/kivy/uix/accordion.py
+++ b/kivy/uix/accordion.py
@@ -276,7 +276,7 @@ class AccordionItem(FloatLayout):
 
     def on_collapse(self, instance, value):
         accordion = self.accordion
-        self.collapsing = True;
+        self.collapsing = True
         if accordion is None:
             return
         if not value:
@@ -323,8 +323,6 @@ class AccordionItem(FloatLayout):
             return True
         else:
             return super(AccordionItem, self).on_touch_up(touch)
-
-   
 
     def _update_title(self, dt):
         if not self.container_title:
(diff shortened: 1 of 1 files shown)


kivy/uix/accordion.py
```diff
@@ -359,14 +379,13 @@ class Accordion(Widget):
...
-        fbind('size', update)
-        fbind('pos', update)
-        fbind('min_space', update)
+        self._trigger_layout = Clock.create_trigger(self._do_layout, -1)
+        self.bind(
+            orientation=self._trigger_layout,
+            children=self._trigger_layout,
+            size=self._trigger_layout,
+            pos=self._trigger_layout,
+            min_space=self._trigger_layout)
```
tshirtman: what's the reason to revert to old style bindings, is this intended?

kivy/uix/accordion.py
```diff
@@ -1,4 +1,4 @@
-'''
+﻿'''
```
tshirtman: also, check your editor's config, there is an invisible character (U+FEFF) that was introduced at the beginning of the file here, had to use hexedit to replace it by something visible and edit it out later.

KeyWeeUsr:
Doesn't fix really. It works only first time after an item is collapsed, otherwise `on_touch_up` is triggered in all collapsed tabs when:

* clicked second time in the tab
* uncollapsing from the right side
* clicked at the button/tab header that does the uncollapsing.
* etc

It's still a bug, but I'm not sure if minor or something bigger. Feel free to open an issue or PR with better fix.

PR #3844, opened 2015-12-14
Preventing from swiping outside the slides range in Carousel widget

Nykakin:
If `Carousel` widget has its `loop` property set as `False`, then this code blocks attempts to swipe further after reaching last item and swipe back from the first item - the widgets are locked in place.

Tested with:

``` python
from kivy.app import App
from kivy.lang import Builder
from kivy.uix.carousel import Carousel

Builder.load_string("""
<MyWidget>:
    direction: 'left'
    loop: False
    Label:
        text: "Screen 1"
    Label:
        text: "Screen 2"
    Label:
        text: "Screen 3"
""")

class MyWidget(Carousel):
    pass

class MyApp(App):
    def build(self):
        return MyWidget()

if __name__ == '__main__':
    MyApp().run()
```

### DIFF
--- a/kivy/uix/carousel.py
+++ b/kivy/uix/carousel.py
@@ -524,10 +524,21 @@ class Carousel(StencilView):
         direction = self.direction
         if ud['mode'] == 'unknown':
             if direction[0] in ('r', 'l'):
-                distance = abs(touch.ox - touch.x)
+                distance = touch.ox - touch.x
             else:
-                distance = abs(touch.oy - touch.y)
-            if distance > self.scroll_distance:
+                distance = touch.oy - touch.y
+            if not self.loop:
+                if direction[0] in ('r', 't'):
+                    if self.index == 0 and distance < 0:
+                        return False
+                    if self.index == len(self.slides)-1 and distance > 0:
+                        return False
+                else:
+                    if self.index == 0 and distance > 0:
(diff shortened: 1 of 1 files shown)


tshirtman:
i think this behavior is interesting, but should be optional, just like ScrollView has multiple ways to deal with overscroll, application designers may have different needs for that, and changing the current behavior won't be acceptable for everybody.

KeyWeeUsr:
I like the old behavior really much and with this edit it looks a little bit like it's not even registering the touch (because it stays in place without feedback). Maybe add something like those fancy blue strip-like things if you do the same scrolling in stackexchange's [carousel-like widgets](http://phandroid.s3.amazonaws.com/wp-content/uploads/2014/01/stack-exchange-app-640x511.png) or something? If it's going to be merged, an optional behavior would be nice. ^^

Also, don't know if it's intended or not, but the PR will work only if you drag it hm... like you tried when attempting to scroll to the next of carousel pages(Screen1 -> Screen3), however, it won't work if you try to scroll to the previous page and back to the next (Screen1 -> Screen2 -> Screen3) with a single movement. (do [this](https://thumbs.dreamstime.com/z/touch-screen-gesture-2-hands-19553383.jpg) to compare the different behaviors, can't describe it well).

PR #3849, opened 2015-12-16
Update pyinstaller docs and hook for use with wheels.

matham:
To test, checkout https://github.com/pyinstaller/pyinstaller/pull/1678. Then copy and overwrite the rt hook and the kivy-hook in that pr located under pyinstaller/... with the ones in this pr. Then in the copied kivy-hook uncomment the line with hiddenimports.

The docs explain from the pov that this has been done because that's how it'll look like once that pr is merged and we fix the hooks.


schitzN:
@matham really looking forward to this pr. when do you think this will be done? would love to see a new wheel which i can use to package for windows!

PR #3858, opened 2015-12-21
Added Spinner text property update when values are updated.

rafalo1333:
If Spinner's values property is updated, the Spinner text is not changing. This is the wrong behavior. Let's make some examples:

1. User creates the Spinner with the values ('Home', 'Work', 'Holidays') and without the text property set. Spinner will show no text when some options are available; there will be no default option shown - and it is bad for the user experience.
2. User creates the Spinner with the empty values property. Then the Spinner instance is updated with the list of ('Home', 'Work', 'Holidays') but Spinner wont show any default value - empty text string suggests that no options are available to choose from.
3. User creates the Spinner with the values set to ('Home', 'Work', 'Holidays') and text set to 'Home'. Then Spinner values are updated to ('New Home', 'Another Work', 'Next Holidays'). The text will still show 'Home' as selected value, so without the user interaction the selection will be wrong.

I have added the text_autoupdate property that indicates if the user want the Spinner to automatically set the text property to values[0].

If set to True and values is not empty, the Spinner will automatically set its text to values[0]. Else, it will leave the text value untouched.

I believe it will make Spinner widget funtionality and complexity better,

### DIFF
--- a/kivy/uix/spinner.py
+++ b/kivy/uix/spinner.py
@@ -63,6 +63,16 @@ class Spinner(Button):
     :attr:`values` is a :class:`~kivy.properties.ListProperty` and defaults to
     [].
     '''
+    
+    text_autoupdate = BooleanProperty(False)
+    '''Indicates if the spinner's :attr:`text` should be automatically updated with the
+    first value of the :attr:`values` property. Setting it to True will cause
+    the spinner to update its :attr:`text` property every time
+    attr:`values` are changed.
+    
+    :attr:`text_autoupdate` is a :class:`~kivy.properties.BooleanProperty` and
+    defaults to False.
+    '''
 
     option_cls = ObjectProperty(SpinnerOption)
     '''Class used to display the options within the dropdown list displayed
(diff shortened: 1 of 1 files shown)


dessant:
I feel there is no need for the new property, the spinner button text should propably be set to an empty string whenever `.values` changes.

On the other hand, if you also want to set `Spinner.text` to one of the options, this makes it difficult, because _update_dropdown would overwrite it during dispatch.

It seems the current way of setting the `Spinner.text` manually for the main button if you update `Spinner.values` brings the least of complications.

dessant:
@rafalo1333, if you can address the issue about being able to set `Spinner.text` manually when autopdate is False, we could work with it.

dessant:
:+1:

matham:
If the original text is in the values shouldn't it reuse that rather than
picking the first value? It should only pick the first value if the current
text is not in the new values.

dessant:
@matham, yeah, that would be nice.

dessant:
I've added a check https://github.com/kivy/kivy/commit/90ccf0a2a1c1955abe8614097d6e3013fdc703cc.

PR #3859, opened 2015-12-21
deprecate MIN_STATE_TIME in favor of min_state_time

### LINKED ISSUE #3843: Button behavior has a property MIN_STATE_TIME
This property can't be used in kv cause of the uppercase.

tshirtman:
fixes #3843

### DIFF
--- a/kivy/uix/behaviors/button.py
+++ b/kivy/uix/behaviors/button.py
@@ -4,8 +4,10 @@
 __all__ = ('ButtonBehavior', )
 
 from kivy.clock import Clock
-from kivy.properties import OptionProperty, ObjectProperty, BooleanProperty
+from kivy.properties import OptionProperty, ObjectProperty, \
+    BooleanProperty, NumericProperty, AliasProperty
 from time import time
+from kivy.logger import Logger
 
 
 class ButtonBehavior(object):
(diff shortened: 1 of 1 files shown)


dessant:
:+1:

akshayaurora:
@kived explained the var was static and allowed setting it globally for every instance. That negates the need to set it from kv. reverting the merge till we have a proper discussion on this as I think the static var was a nice choice just needed more documentation.

dessant:
@akshayaurora a global var can be a nice choice, but that is just as true for most kivy properties. This gives us a way to change it for an individual widget, and also to change it globally in kv for all buttons.

akshayaurora:
ok, I guess there is no need to revert then, +1

Matteljay:
Hopefully on topic: is it me or is min_state_time broken for the Button widget? I tried simply spamming it with my mouse on a Linux desktop: updated Arch Linux: python-kivy 1.10.1-2, python 3.7.1
1. not only is it useless: firing the on_press event every time AND:
2. getting a value 'down' for the .state property for the first click instance, making it impossible to ignore future clicks within the min_state_time period.
3. it also visually acts weird with different intervals (i tried setting it to 4 seconds, try clicking it like a madman :))
In my [LabelPush](https://github.com/Matteljay) project is a simple & hacky fix, if deemed useful please add something similar to Kivy:
```
'''
    btn_PRINT: id_btn_PRINT
    Button:
        id: id_btn_PRINT
        on_press: root.press_btn_PRINT()
        text: 'PRINT'
        font_size: 20
        background_down: self.background_normal
'''


    def press_btn_PRINT(self):
        btn = self.btn_PRINT
        # spam control: alternative to Kivy's min_state_time button
        if btn.background_color[1] == 2: return
        btn.background_color[1] = 2
        def greenback(dt): btn.background_color[1] = 1
        Clock.schedule_once(greenback, 4)
```

Issue #3481, opened 2015-07-02
KEP: allow tuple/list coordinates for Line.points


kived:
Could be handled at assignment time.

````python
if points and isinstance(points[0], (list, tuple)):
    points = list(itertools.chain(*points))
````

trugate:
Thanks for suggesting this officially.  I would think most use-cases in kivy for Line points would be tuples or something similar, like widget.pos, widget.center.  Is there a use case where a list of x, y, x, y makes more sense?

dessant:
test

udiboy1209:
I'd like to work on this issue! Could you tell me where to start ?

tshirtman:
I'd say here
https://github.com/kivy/kivy/blob/master/kivy/graphics/vertex_instructions_line.pxi#L675-L677

udiboy1209:
I made the change suggested by @kived and tried testing it using the `lines` example: https://github.com/kivy/kivy/blob/master/examples/canvas/lines.py

When I run the example as it is, it works. But when I change the points declaration to this

    points = ListProperty([(500, 500), 300, 300, 500, 300, 500, 400, 600, 400])
    
it shows error:

    Exception TypeError: 'a float is required' in 'kivy.graphics.vertex_instructions.SmoothLine.build_smooth' ignored

I dont know whats wrong and what float is required? Does it mean that the elements in the list (coordinate values) should be float, but it is getting a tuple? (which would mean the change isn't working). Please help!

dessant:
Fixed in https://github.com/kivy/kivy/pull/3866.

PR #3866, opened 2015-12-23
Allow Line.points definition to be a mix of lists/tuples

udiboy1209:
Fix for issue #3481

`Line.points` can be defined as list of lists, tuples and coordinates (float values).
Setter method modified to check each element of the argument list and append it to `self._points` accordingly.
Line example file /examples/canvas/line.py updated with the new possible way of defining points.

**Note:** appending a tuple to `points` causes problems


kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -672,8 +670,15 @@ cdef class Line(VertexInstruction):
         '''
         def __get__(self):
             return self._points
+
         def __set__(self, points):
-            self._points = list(points)
+            tmp = []
+            for p in points:
```
dessant: I haven't tested, but it looks like a slow implementation. I think mixing lists/tuples with plain values is not a priority. `points` could be either a plain list, a list of lists, or a list of tuples. This way, a faster algo could be used, something like what @kived proposed in the issue.

udiboy1209:
@dessant I made the changes you requested. I agree the previous algorithm was slower. Given the conditions you mentioned we can use kived's algorithm

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -672,8 +670,12 @@ cdef class Line(VertexInstruction):
         '''
         def __get__(self):
             return self._points
+
         def __set__(self, points):
-            self._points = list(points)
+            import itertools
```
dessant: Is there a reason for doing the import in the method?

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -672,8 +670,12 @@ cdef class Line(VertexInstruction):
...
+            if points and isinstance(points[0], (list, tuple)):
```
dessant: What happens if points in not a list or tuple? :stuck_out_tongue_winking_eye:

kivy/graphics/vertex_instructions_line.pxi
udiboy1209: I just realised that will slow things down :P

kivy/graphics/vertex_instructions_line.pxi
udiboy1209: Oops! missed that!

dessant:
:+1:

Issue #3842, opened 2015-12-14
ToggleButton documentation does not mention BooleanProperty: allow_no_selection


DavidA2014:
ToggleButton documentation does not mention BooleanProperty: allow_no_selection 

http://kivy.org/docs/api-kivy.uix.togglebutton.html#

https://github.com/kivy/kivy/blob/master/kivy/uix/behaviors/togglebutton.py#L30

DavidA2014:
Sorry, I now see that allow_no_selection is defined in the base class docs. But wouldn't it be better if inherited properties showed up in inheriting classes?

tshirtman:
Yes, it's a matter of configuring sphinx for it, i believe, and it would indeed make reading easier, it's not like we are printing the pdf anyway.

Zen-CODE:
Yup. The shinx autodoc extension has the option ':inherited-members:' for autoclass and automodule directives.

Although I'm not a fan of doing this as it makes the docs much more verbose, I can't deny that showing inherited methods is clearer, more complete and avoids many possible issues. I've seen that with our trainees. Will submit a PR for that :-)

Zen-CODE:
PR merged

https://github.com/kivy/kivy/commit/61b9555352e27d13abe45088334f3ec8de8ac015

dessant:
related discussion https://github.com/kivy/kivy/pull/3870

PR #3870, opened 2015-12-27
doc: show inherited members. references #3842

Zen-CODE:


akshayaurora:
@Zen-CODE :dancers:

dessant:
In it's current form it makes the doc hard to parse. Would it be possible tho list inherited stuff at the end of the class docs and mark clearly from which base class each item comes from?

Zen-CODE:
I don't think shpinx allows this kind of customization.

http://sphinx-doc.org/ext/autodoc.html

Search for 'inherit' and you'll see the options are pretty limited. From my
comment on the original ticket:

"Although I'm not a fan of doing this as it makes the docs much more
verbose, I can't deny that showing inherited methods is clearer, more
complete and avoids many possible issues. I've seen that with our trainees."

On Sun, Dec 27, 2015 at 6:43 PM, dessant <notifications@github.com> wrote:

> In it's current form it makes the doc hard to parse. Would it be possible
> tho list inherited stuff at the end of the class docs and mark clearly from
> which base class each item comes from?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/kivy/kivy/pull/3870#issuecomment-167425487>.
>



-- 
===============
ZenKEY - Become one.
http://www.camiweb.com/zenkey

dessant:
In that case I would vote against the change, @tshirtman was not a fan of it either. We'll see how the rest of the team feels after using the online docs.

akshayaurora:
ok, seems I was too eager in merging this...It would be good to have inherited members but not in this form :/

Zen-CODE:
Okay, I'll commit a reversal of this change with a link to this discussion for easy reversal following a proper discussion.

Zen-CODE:
Reversed following proper discussion. 

https://github.com/kivy/kivy/commit/16e710f89c78755c1b754b3f52c95cb8b0d608a0

Issue #3312, opened 2015-04-29
ButtonBehavior.always_release should default to False


kived:
Defaulting to False provides a much better UX, as it is the expected behavior. If I click/touch a button, then drag my mouse/finger off the button, the `on_release` event should not fire. This flag should be turned on for the special use cases where you do want the `on_release` to fire this way.
##

dessant:
I agree, but that would break certain apps. Are you ok with tagging it for 2.0?

akshayaurora:
+1 for 2.0

GP89:
I didn't know this had been added- it's a good suggestion. Would you say as well if this is False, the state should return to normal as well when the drag goes outside the button (and back to down when it returns)? then the visual of the button would no longer appear in the down state, making it clearer that an `on_release` wont get fired when releasing. Maybe this would break some apps though

rafalo1333:
+1 for defaulting it to False.

tito:
I didn't noticed the changes of the default until today.
The current default change is incomplete and introduce more subtile bugs to resolve.

Before: as stated @kived, if you wanted not to fire the `on_release`, it wasn't possible. Except if you used `always_release` to True, even if UX was bugged.
Now: Stay press, `on_press` fired, move out, done. But the UX state is still `down`, therefore and not `normal` as the UX would suggest in the first place. And then to make the button back to `normal`, you need to click on it, so `on_press` and `on_release` is fired, which defeat the initial purpose.

Either the fast `bad` fix would be to put it back to True, until the UX is fixed
Either fix the default and the behavior.

Whatever the solution is, this default change didn't helped, and IMO, make the bug about always_release to True and UX state to down more visible than before.

tito:
Actually, this issue was opened and explained in #4478.

PR #3891, opened 2016-01-04
ButtonBehavior.always_release default to False

dessant:
Fixes https://github.com/kivy/kivy/issues/3312.

There were some discussions on irc about introducing incremental breaking changes after 1.9.1 instead of doing everything in 2.0, the team seemed to be in favor of this route.

We could also post on #kivy-dev for those who might want to now revisit other issues that were previously tagged 2.0: https://github.com/kivy/kivy/milestones/2.0.0.


akshayaurora:
:+1:

PR #3900, opened 2016-01-08
Add kivy.uix.behaviors.CoverBehavior

rnixx:
Introduces a behavior which can be used for covering image and video to widget size keeping aspect ratio


akshayaurora:
Simpler approach, https://gist.github.com/tito/b438d8573db1f7926d69

KeyWeeUsr:
@akshayaurora so... what now? Close/merge/rework?

matham:
I don't think I understand what exactly the point of this is? What is it solving?

rnixx:
@matham 

> I don't think I understand what exactly the point of this is? What is it solving?

It's desired for aspect ratio independent full widget size background images. Similar to CSS 3 background-size: cover;

May you re-open this please?

dessant:
@matham, @rnixx: could this be added to garden instead?

I do see the use case, but it's also arbitrary enough to set a precedent for including many other similar widgets and behaviors.

@rnixx, there has been a sentiment lingering within the team, some would favor trimming the current widget set, though it mostly manifests itself by opposing certain new additions. It may eventually happen, but garden will need to be more discoverable and will need better integration with kivy.

akshayaurora:
@dessant I think this behavior is frequent enough that it warrants to be included in the Image widget.
Most UI frameworks have this by default.

Plus this is a behavior so it does not make the widget heavy unless you add this behavior.

I had linked to the simpler POC tito did to get a idea from rniix of what he thought about it .

On further glance on the code though I see that the properties provided are definitely a improvement allowing further behavior of this behavior.

@rnixx could you please add version added tags and a example in the examples folder with this so people can actually see this in action.

rnixx:
I added examples for cover image and cover video, my local dev install has a problem with video provider, need to check what's wrong.

Docs still need to be polished.

akshayaurora:
Eagerly awaiting for the docs :)

rnixx:
@akshayaurora ready to review.

akshayaurora:
@rnixx thanks this works like a charm :)

rnixx:
Thanks!

PR #3901, opened 2016-01-08
Add kivy.uix.behaviors.touchripple module

rnixx:
Introduces general TouchRippleBehavior and TouchRippleButtonBehavior classes


dessant:
@rnixx, thanks for your work, this will be a popular feature.
There were some concerns about the ways variables are set, @Kovak could you give some pointers what needs to be changed in the pr?

rnixx:
@Kovak, if you provide some details I can adopt the pull request where appropriate. Thanks

Kovak:
All of the controlling properties were made into non-properties, things like:

ripple_fade_from_alpha = .5
ripple_fade_to_alpha = .8

This makes it needlessly difficult to style the ripple. Also is a step back from the original implementation this code is based on.

In addition we touch ungrab but never grab, which means we are randomly ungrabbing some grab caused by another logic. This of course is the reason my original implementation actually had a GrabBehavior that no one else seems to have liked. It is the only really solid way to handle widgets that may have several mixins that depend on touch.grab without ending up with tons of touch.grab for the same widget.

rnixx:
> This makes it needlessly difficult to style the ripple
Is there really a scenario where some want to bind these settings? I mean, normally you do not change these settings during an animation and if you change the settings on a subclass or an instance, changes get respected for the next animation.

> In addition we touch ungrab but never grab
Ok, what's your suggestion then? Remove the ungrab in ``on_touch_up`` or add a grab to ``on_touch_down``?

> This of course is the reason my original implementation actually had a GrabBehavior that no one else seems to have liked
I cannot remember a GrabBehavior, but I'm fine if one exists.

rnixx:
@Kovak @dessant may you review again please?

kivy/uix/behaviors/touchripple.py
```diff
@@ -0,0 +1,262 @@
...
+    '''
+
+    ripple_func_out = StringProperty('out_quad')
+    '''Animation callback for hiding the overlay.
+
+    :attr:`ripple_func_out` is a :class:`~kivy.properties.StringProperty`
+    and defaults to `out_quad`.
+    '''
+
+    ripple_rad = NumericProperty(ripple_rad_default)
```
Kovak: Does this actually run for you?

ripple_rad_default here is a NumericProperty so its value cannot be accessed like this

rnixx: uh, missed that one. I have not fired up the recent version yet. will adopt

Kovak:
@rnixx  if you can fix the last error, we'd like to merge this this weekend if possible.

rnixx:
@Kovak Sorry for the late reply, missed that one in my mails. Fix is committed

rnixx:
I fear the TouchRippleButtonBehavior cannot stay as it is right now.

We have different behviour how on_release gets triggered. thus is think we cannot work with super calls because on_touch_up of ButtonBehavior would be called then, which causes the behaviour trouble.

But we want to delegate on_touch_down and on_touch_up anyways to other superclasses, right?

An option would be to create a ButtonBehaviorBase class, not implementing on_touch_down and on_touch_up. This base gets inherited from then in ButtonBehavior and TouchRippleButtonBehavior which implement the touch event handlers.

Next, if i understand ButtonBehavior and how it is used in uix.button and kv, state is only used for visual effect, which is also not desired in a TouchRippleButton because the ripple animation is the visual effect there. So the base class also might not provide state and min_state_time.

rnixx:
I changed TouchRippleButtonBehavior not to derive from ButtonBehavior now.

Event registration, on_mouse_move, default on_press and on_release handler, last_touch and always_release share the same code, so this might go to a base class.

After thinking a little more about state, it might be worth supporting it as well in TouchRippleButtonBehavior, if someone wants a toggle button with ripple effect on each state change.

What do you think?

dessant:
Thanks for the update @rnixx. @Kovak, are you ok with the current state, api names etc?

KeyWeeUsr:
@Kovak ping 😄

rnixx:
Last time we talked about this topic in IRC, @Kovak complained about the reimplementation of a touch ripple related button behaviour.

I was pointing out problems with super chains when starting to inherit from multiple behaviours, especially if we have in detail slightly different needs when exactly events are triggered, etc. The discussion ended with no result.

We have some options now:

* Remove ``TouchRippleButtonBehavior``, merge with ``TouchRippleBehavior`` only and work/discuss on the multiple inheritance problem in a new PR/issue.
* Add ``experimental`` note to  ``TouchRippleButtonBehavior`` and merge as is.
* Keep discussing ;)

rnixx:
@Kovak @dessant @KeyWeeUsr 

Hi,

how should we proceed with this one?

kivy/uix/behaviors/touchripple.py
```diff
@@ -0,0 +1,312 @@
...
+        self.ripple_ellipse = None
+        self.ripple_col_instruction = None
+
+    def ripple_show(self, touch):
+        '''Begin ripple animation on current widget.
+
+        Expects touch event as argument.
+        '''
+        Animation.cancel_all(self, 'ripple_rad', 'ripple_color')
+        self._ripple_anim_complete(self, self)
```
KeyWeeUsr: is this _really_ intended? `_ripple_anim_complete` does nothing with any of them, so why do you pass them into the method?

kivy/uix/behaviors/touchripple.py
```diff
@@ -0,0 +1,312 @@
+'''
+Touch Ripple
+============
+
+.. versionadded:: 1.9.2
```
KeyWeeUsr: Please change it to `1.10.0` :)

kivy/uix/behaviors/touchripple.py
```diff
@@ -0,0 +1,312 @@
...
+:class:`~kivy.uix.behaviors.button.ButtonBehavior` but rendering the ripple
+animation instead of default press/release visualization.
+'''
+from kivy.animation import Animation
+from kivy.clock import Clock
+from kivy.graphics import CanvasBase
+from kivy.graphics import Color
+from kivy.graphics import Ellipse
+from kivy.graphics import ScissorPush
+from kivy.graphics import ScissorPop
```
KeyWeeUsr: Maybe a more compact import would be better for eyes i.e. `from this import x, y, z`

rnixx: That's a matter of taste I guess. Personally I prefer one import per line. If there is a consensus among kivy developers preferring multiple imports per line I can change it for sure.

kivy/uix/behaviors/touchripple.py
rnixx: The same function is registered as animation ``on_complete`` callback in Line 205 which gets passed ``anim`` and ``instance`` as parameters.

kivy/uix/behaviors/touchripple.py
rnixx: done

kivy/uix/behaviors/touchripple.py
KeyWeeUsr: I have this from the docs: `on_complete: animation, widget`, so I still think it's little bit weird to pass `self, self` to that method as the `self` is an instance of `TouchRippleBehavior`, which is neither widget nor animation I think?

Didn't you intend to do this rather?

    self.anim = Animation( ...)

and in that method:

    self._ripple_anim_complete(self.anim, <something>)?

rnixx: Actually the intention is to reset the ripple radius and clear the ripple_pane both on ripple_fade animation complete and before performing ripple_show animation in case ripple_fade gets interrupted.

I move the code to a seperate function and call it explicit in _ripple_anim_complete and ripple_show for clarification

rnixx: done

KeyWeeUsr: I meant it like this, but nevermind :)

    def foo(self, *args):
        pass
    ----
    self.bind(on_complete=self.foo)
    self.foo(it, collects, the, arguments)
    # and doesn't require any
    self.foo()

rnixx: if you take a look at ripple_show function. There happens some stuff including adding canvas instructions before instanciate and starting the animation. thus clearing the pane in an animation event callback will not work.

rnixx:
@Kovak Hi, do you have some time so we can finally finish this task?

rnixx:
@KeyWeeUsr Hi,

I updated the import formatting, added experimental note to this module and added a fix which ensures ripple animation completes if ``disabled`` property gets set to True in ``TouchRippleButtonBehavior``.

Hope this can get merged in this state now.

Thanks

KeyWeeUsr:
@rnixx Thanks for waiting and for the work 🍡

rnixx:
@KeyWeeUsr Thank you likewise.

PR #3905, opened 2016-01-10
OS X Packaging: Update for PyInstaller 3.1

cbenhagen:


dessant:
:+1:

dessant:
It would be good to document hook overriding for osx, it's already there in the win packaging docs, maybe there could be a shared pyinstaller section. Thoughts?
https://kivy.org/docs/guide/packaging-windows.html#overwriting-the-default-hook

PR #3906, opened 2016-01-10
Adds gl mocking so tests can run

matham:
The adds gl mocking via the `USE_OPENGL_MOCK` env variable. With this the appveyor, rtd and travis mac builds work.


.travis.yml
```diff
@@ -118,18 +117,19 @@ script:
...
          make;
-         echo -e "make test";
+         make test;
       fi;
     fi;
 
 
 after_success:
-  - if [ "${TRAVIS_OS_NAME}" == "osx" ] && [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
+  - if [ "${TRAVIS_OS_NAME}" == "os" ] && [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
```
tito: is this intented?

.travis.yml
```diff
@@ -52,18 +52,18 @@ install:
       curl -O -L https://www.libsdl.org/projects/SDL_image/release/SDL2_image-2.0.1.dmg;
       curl -O -L https://www.libsdl.org/projects/SDL_mixer/release/SDL2_mixer-2.0.1.dmg;
       curl -O -L https://www.libsdl.org/projects/SDL_ttf/release/SDL2_ttf-2.0.13.dmg;
-      curl -O -L http://gstreamer.freedesktop.org/data/pkg/osx/1.6.2/gstreamer-1.0-1.6.2-x86_64.pkg;
-      curl -O -L http://gstreamer.freedesktop.org/data/pkg/osx/1.6.2/gstreamer-1.0-devel-1.6.2-x86_64.pkg;
+      curl -O -L https://www.googledrive.com/host/0B1_HB9J8mZepS0hXamNnd2pRWkE/gstreamer-1.0-1.6.2-x86_64.pkg;
```
tito: please revert :)

kivy/graphics/c_opengl.pxd
```diff
@@ -384,147 +384,148 @@ cdef extern from "gl_redirect.h":
...
-    cdef void  glVertexAttrib1fv(GLuint indx,  GLfloat* values) nogil
-    cdef void  glVertexAttrib2f(GLuint indx, GLfloat x, GLfloat y) nogil
-    cdef void  glVertexAttrib2fv(GLuint indx,  GLfloat* values) nogil
-    cdef void  glVertexAttrib3f(GLuint indx, GLfloat x, GLfloat y, GLfloat z) nogil
-    cdef void  glVertexAttrib3fv(GLuint indx,  GLfloat* values) nogil
-    cdef void  glVertexAttrib4f(GLuint indx, GLfloat x, GLfloat y, GLfloat z, GLfloat w) nogil
-    cdef void  glVertexAttrib4fv(GLuint indx,  GLfloat* values) nogil
-    cdef void  glVertexAttribPointer(GLuint indx, GLint size, GLenum type, GLboolean normalized, GLsizei stride,  GLvoid* ptr) nogil
-    cdef void  glViewport(GLint x, GLint y, GLsizei width, GLsizei height) nogil
+    IF not USE_OPENGL_MOCK:
```
tito: If the inclusion is already controlled by USE_OPENGL_MOCK, maybe don't include it here too?

.travis.yml
dessant: Yes, the block was "disabled" :-P because it's unfinished, it should upload to g drive instead of the server.

.travis.yml
matham: The reason for this, with the old links directly from libsdl their connection was so slow, that it'd take at least half an hour to download and fairly often the tests would time out. However, from google drive it takes a few seconds to download making the test result a lot quicker.

Should I still revert? What is the problem with hosting on drive?

kivy/graphics/c_opengl.pxd
matham: Not quite. Even when mocking we still use this file for the constants, which would cause linking issues if not guarded like this. So this file is always included. Otherwise, we'd had to reimport all the constants in the mocking pxd file as well. This seemed cleaner. What do you think?

matham:
As you can see, even with a download manager with multiple connection it took 40 min to complete: https://travis-ci.org/kivy/kivy/builds/102838915. However, quite often it takes longer and it just times out. So I think going back to hosting gst on drive is better. What do you think @tito?

dessant:
We could do this until https://github.com/travis-ci/travis-ci/issues/4011.

dessant:
Ok, let's merge this. :+1:

PR #3909, opened 2016-01-12
Add apply_rules to BuilderBase

Gagaro:
The idea behind this feature is to be able to load a set of rules and apply them without applying all the already existing rules or adding the rules applied to the existing ones.

It's my first contribution to Kivy, so don't hesitate to tell me if I forgot something.

Thanks.

### DIFF
--- a/kivy/lang.py
+++ b/kivy/lang.py
@@ -1795,6 +1795,9 @@ class ParserSelectorName(ParserSelector):
             parents[cls] = classes
         return self.key in parents[cls]
 
+    def match_rule_name(self, rule_name):
+        return self.key == rule_name.lower()
+
 
 class BuilderBase(object):
     '''The Builder is responsible for creating a :class:`Parser` for parsing a
@@ -1805,6 +1808,7 @@ class BuilderBase(object):
     '''
 
     _match_cache = {}
+    _match_name_cache = {}
 
     def __init__(self):
         super(BuilderBase, self).__init__()
@@ -1953,6 +1957,25 @@ class BuilderBase(object):
         self._apply_rule(widget, rule, rule, template_ctx=proxy_ctx)
         return widget
 
+    def apply_rules(self, widget, rule_name, ignored_consts=set()):
+        '''Search all the rules that match `rule_name` widget
+        and apply them to `widget`.
+
+        .. versionadded:: 1.9.2
+
+        `ignored_consts` is a set or list type whose elements are property
+        names for which constant KV rules (i.e. those that don't create
+        bindings) of that widget will not be applied. This allows e.g. skipping
+        constant rules that overwrite a value initialized in python.
+        '''
+        rules = self.match_rule_name(rule_name)
+        if __debug__:
+            trace('Builder: Found %d rules for %s' % (len(rules), rule_name))
+        if not rules:
+            return
+        for rule in rules:
+            self._apply_rule(widget, rule, rule, ignored_consts=ignored_consts)
+
     def apply(self, widget, ignored_consts=set()):
         '''Search all the rules that match the widget and apply them.
 
@@ -1971,6 +1994,7 @@ class BuilderBase(object):
 
     def _clear_matchcache(self):
         BuilderBase._match_cache = {}
+        BuilderBase._match_name_cache = {}
 
     def _apply_rule(self, widget, rule, rootrule, template_ctx=None,
                     ignored_consts=set()):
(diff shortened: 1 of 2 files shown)


Gagaro:
Is this an issue with travis?

```
The job exceeded the maxmimum time limit for jobs, and has been terminated.
```

matham:
Yeah, it's a travis issue.

Regarding the PR, it is really nice, however, I'm not sure why this would be needed? I.e. in what real situations is this helpful? If it's for situations where you want to apply a rule to widget but not keep that rule around so that it's applied to other widgets of the same class. Then I'd do it differently,

I'd not add new functions for loading kv strings/files just once and not store them, but rather force having to load the kv string/file and store it like in the past, but instead also add a method which force applies a kv rule (even if the class doesn't match) to the widget. This way, you can still load the rule like normal, but give the kv rukle a special name so it's not applied to any normal class and instead is only applied when you force it on the widget.

This would reduce the amount of new code added, and it'll be a lot more efficient and compatible with the upcoming compiled kv code. As you could precompile the rule, and then apply to whatever widget you like.

Gagaro:
My need is to be able to load level in a game (an arkanoid) using kv files. Currently, it would keep the rules and loading a level would reload anything already linked to the widget (you can see it here https://github.com/Gagaro/arkanivy/tree/395187ab92def750b92c5d6aba0b23eb662e2453/arkanivy).

So it is mainly to be able to have one time rules.

Are you suggesting to had something new to the kv language directly? It seems like more work to do (and it would make it more difficult to use). My way re-uses what already exists (to load the file, match the widgets to the rules, ...) and only really add one method (`apply_file` using `apply_string`).

How is the compiled kv going to work (at load or before loading, ..) ? Is there something I can read on it somewhere?

matham:
So for your example, the problem is that you want to apply the `ArkanivyGame` rule from `level1.kv` after the `ArkanivyGame` object has been created and its `ArkanivyGame` rule from `arkanivy.kv` has been applied. But you don't want to read and store the new `ArkanivyGame` rule from ` level1.kv` because then it'll affect every `ArkanivyGame` created, but you don't want that. Right?

If so, what my alternate solution is as follows. Rename `ArkanivyGame` in `level1.kv` to `ArkanivyGame576698434` and then load `level1.kv` as normal and rules will be read and saved. Now, when you want to apply `ArkanivyGame576698434` to the existing `ArkanivyGame`, all you'd have to do is call `Builder.apply_rule(my_game, 'ArkanivyGame576698434')`. This will apply that rule to the widget even though the class names don't match.

This alternate solution requires much less to implement, because all you have to do is locate the rule and apply it.

With regards to compiling kv, this loading and not saving would be quite expensive because you'd have to recompile every time. However, if you load it once and apply when asked it should be quite fast. It's the same with current kv though. The kv compiler is a work in progress, you can see the PR here: https://github.com/kivy/kivy/pull/3456.

Gagaro:
Ok I see what you mean now. It sounds better indeed. Thanks for your input, I'll try to rewrite my PR asap.

Gagaro:
Done. I had to add a `match_widget_name` because the `match` method needs an actual widget to get the matching rules.

kivy/lang.py
```diff
@@ -1795,6 +1795,10 @@ def match(self, widget):
             parents[cls] = classes
         return self.key in parents[cls]
 
+    def match_widget_name(self, widget_name):
+        return self.key == widget_name.lower()
+
+
```
matham: Did you run the pep8 hook? Because it should be exactly 2 blank lines: https://kivy.org/docs/contribute.html#coding-style.

kivy/lang.py
```diff
@@ -1795,6 +1795,10 @@ def match(self, widget):
             parents[cls] = classes
         return self.key in parents[cls]
 
+    def match_widget_name(self, widget_name):
```
matham: `rule_name` is better than `widget_name`.

kivy/lang.py
```diff
@@ -1953,6 +1958,23 @@ def template(self, *args, **ctx):
         self._apply_rule(widget, rule, rule, template_ctx=proxy_ctx)
         return widget
 
+    def apply_rules(self, widget, widget_name, ignored_consts=set()):
```
matham: `rule_name`. rather than `widget_name` is a better choice. Also, you need to add the versionadded think, like here: https://github.com/udiboy1209/kivy/blob/styling/kivy/uix/label.py#L645. The version is 1.9.2.

kivy/lang.py
```diff
@@ -2176,6 +2199,23 @@ def match(self, widget):
         cache[k] = rules
         return rules
 
+    def match_widget_name(self, widget_name):
```
matham: `rule_name`.

kivy/lang.py
Gagaro: Sorry, I had an issue with the hook so I ran it manually but forgot for the second commit..

Gagaro:
Fixed pep8 and renamed widget_name.

matham:
Awesome!

PR #3932, opened 2016-01-22
Changed ffpyplayer.version to ffpyplayer.__version__

Laspimon:
I was getting the following error when running make test:

      File "~/kivydev/kivy/core/__init__.py", line 129, in core_register_libs
        level=0)
      File "~/kivydev/kivy/core/audio/audio_ffpyplayer.py", line 63, in <module>
        Logger.info('SoundFFPy: Using ffpyplayer {}'.format(ffpyplayer.version))
    AttributeError: 'module' object has no attribute 'version'

### DIFF
--- a/kivy/core/audio/audio_ffpyplayer.py
+++ b/kivy/core/audio/audio_ffpyplayer.py
@@ -60,7 +60,7 @@ from kivy.core.audio import Sound, SoundLoader
 from kivy.weakmethod import WeakMethod
 import time
 
-Logger.info('SoundFFPy: Using ffpyplayer {}'.format(ffpyplayer.version))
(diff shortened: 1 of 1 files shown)


matham:
The problem with this PR is that it's going to break it for people using an older version of ffpyplayer. Right now this change is only in master, and is quite recent. A better solution is to try `version` and then `__version__` if the first doesn't exist.

matham:
On second thought, I re-added `version` to ffpyplayer. So it should be fixed with ffpyplayer master.

Laspimon:
It makes sense to re-add the version attribute, but isn't it cleaner to make the check for __version__?

matham:
Sure, if you make a pr that checks for `__version__` and then falls back on `version` that would be fine.

PR #3935, opened 2016-01-22
Altered get_cursor_from_xy to intuitively place cursor

Clarvel:
Altered get_cursor_from_xy to provide more natural text positioning with mouse by placing the cursor to the left or right of the clicked on letter depending where on the letter the mouse is clicked, no more pixel perfect placements required.
If the cursor is to the left half of the letter, the cursor is placed on the left.
If the cursor is placed on the right half of the letter, the cursor is placed on the right.

Improved efficiency of the for loop by removing a redundant variable (dcx) and correcting the loop's range.

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -1147,17 +1147,15 @@ class TextInput(FocusBehavior, Widget):
         scrl_y = scrl_y / dy if scrl_y > 0 else 0
         cy = (self.top - padding_top + scrl_y * dy) - y
         cy = int(boundary(round(cy / dy - 0.5), 0, len(l) - 1))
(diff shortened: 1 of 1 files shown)


dessant:
@Clarvel, thanks for your contribution. :+1: 

Would you alter the threshold a bit? I've tested a couple of apps and they seem to place the cursor on the right only if the click is about 0-40% from the right edge of the letter. This is probably done to ensure that the cursor is placed on the left if the click is in the middle of the letter.

Clarvel:
That's a fairly simple fix, the file should be updated to place the threshold 60% from the left.

dessant:
@great, thanks again. :100:

PR #3936, opened 2016-01-23
Ffpyplayer __version__ check

Laspimon:
As with #3932
This checks for ffpyplayer.\__version__ but falls back on ffpyplayer.version, to accommodate older versions of ffpyplayer.


matham:
:+1:

PR #3955, opened 2016-01-30
Deprecate the interactive launcher

dessant:


matham:
Why?

dessant:
@matham, there were discussions on irc about deprecating it for some time.
It is generally buggy and broken, and people seem to think that it should be rewritten rather than patched.

akshayaurora:
This is a good idea, we need to provide a alternative too, though for now just deprecating is necessary. It's unmaintained code. We can simply use twisted manhole that already works wonders for us on kivy-remote-shell.

kawing-chiu:
What is the alternative to interactive launcher? Does kivy support interactive usage in ipython currently?

Issue #3959, opened 2016-02-01
Label color is not correct when markup is true


dessant:
After discovering https://github.com/kivy/kivy/issues/3920, I did this https://github.com/kivy/kivy/commit/1d783c9c93e5083ca4aa9bc2e3b3e5f78639e0cd, but i doesn't really fix the issue, just shows it from a different angle.

![1](https://cloud.githubusercontent.com/assets/7513068/12719222/7b8b48cc-c8fc-11e5-9da7-ebcb0dfae57c.PNG)

```python
from kivy.app import App
from kivy.lang import Builder

kv = """
<Lab@Label>:
    font_size: dp(100)
    text: 'O O'
    color: 1, 0, 0, 0.3
    disabled_color: 0.9, 0.8, 0.7, 0.6

AnchorLayout:
    canvas:
        Color:
            rgba: 0, 0, 0, 1
        Rectangle:
            size: self.size
            pos: self.pos
    GridLayout:
        cols: 3
        size_hint_y: None
        height: '300dp'
        Widget:
        Label:
            text: 'disabled: False'
        Label:
            text: 'disabled: True'
        Label:
            text: 'markup: False'
        Lab:
            disabled: False
        Lab:
            disabled: True
        Label:
            text: 'markup: True'
        Lab:
            markup: True
            disabled: False
        Lab:
            markup: True
            disabled: True
"""


class TestApp(App):
    def build(self):
        return Builder.load_string(kv)

if __name__ == '__main__':
    TestApp().run()
```

udiboy1209:
I'll look into this then @dessant.

udiboy1209:
@dessant why do you set the color = `(1,1,1,1)` when `self.markup` is true in [style.kv](https://github.com/kivy/kivy/blob/ab3c3fe52ce353d77e803f68b116c982f59aeef9/kivy/data/style.kv#L6). Does `MarkupLabel` use some other way of coloring text?

dessant:
@udiboy1209 i'm not sure how is the coloring done for markup text, i've set (1,1,1,1) for disabled text to be consistent with what was done when the label is not disabled, but i have not looked into it past that.

dessant:
We had a discussion with quanon, I'll mention the desired behavior here.

After https://github.com/kivy/kivy/issues/4010 is implemented, when the markup hex color does not contain alpha, we use the alpha component from `color` or `disabled_color`.

dessant:
[...]

PR #3963, opened 2016-02-03
Show disabled_color when disabled=True for markup label

udiboy1209:
Fix for #3959 

Regarding the alpha thing, I dont think markup currently supports alpha because I tried this:

    Label:
        markup: True
        color: 0,0,1,0.5
        text: 'A [color=#77FF0000]B[\color]'

and it outputs [this image](http://imgur.com/xGdAl4z):

output

@dessant please review this.


dessant:
I agreee, alpha is a different issue, the pr looks good. :100:

Issue #2916, opened 2015-01-26
Glerror 1282 after view matrix is set.


olymk2:
running glerror on 3Drendering.py kivy example right after the projection or model matrix shows error 1282, spent ages trying to track this down in my code :p 

the changes i made are in this gist https://gist.github.com/olymk2/49db889c94488bed5019

 def gl_error(self, text = '', kill=True):
        err = glGetError()
        if not err:
            return 
        while err:
            print '## GL ## = ' + text + 'OPENGL Error Code = ', err
            err = glGetError()

olymk2:
using opengl debugging i can see use_shader is being cleared which is likely causing the issue,

('GL glUseProgram( program = ', 0, ', )')
('GL glUniformMatrix4fv( location = ', 2, ', count = ', 1, ', transpose = ', 0, ', value*=', "'0x7fffc12106e0'", ', )')
OpenGL Error 1282 / 502

olymk2:
I have been looking into this it looks like when you use schedule an update interval it calls your update method but does not make the shader active before trying to set any uniforms on the current program.

still not sure on a fix

PR #3970, opened 2016-02-04
fix upload uniform without calling useprogram

Kovak:
Shader's set_uniform can be called by RenderContext without first calling the appropriate UseProgram. This appears to happen mainly on window resize (at least for me). 

It's a pretty simple fix, we just check to see if the current program is the expected program and if not we call glUseProgram.

I believe this should address issue #2916 as well.

### DIFF
--- a/kivy/graphics/shader.pyx
+++ b/kivy/graphics/shader.pyx
@@ -235,6 +235,12 @@ cdef class Shader:
     cdef int set_uniform(self, str name, value) except -1:
         if name in self.uniform_values and self.uniform_values[name] == value:
             return 0
+        cdef GLint data
+        glGetIntegerv(GL_CURRENT_PROGRAM, &data)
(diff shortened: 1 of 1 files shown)


Kovak:
Thinking on this a little more, I have 2 concerns:

1. the program name is an unsigned int but in es2, we get a signed int back from the glGetIntegerv. In practice this shouldn't really be an issue for the first 32,767 programs, of which I don't think anyone will ever have so much anyway.

2. I'm not entirely certain on how we end up with RenderContext.set_state (https://github.com/kivy/kivy/blob/master/kivy/graphics/instructions.pyx#L778) being called without having actually called RenderContext.use, perhaps the actual fix would be making sure we call use before this occurs.

@tito - do you have any advice on the proper way to handle this issue?

dessant:
@tito, :wave:

dessant:
Thanks @Kovak. :horse_racing:

PR #3971, opened 2016-02-04
Bug fix: value of -1 anim_delay setting correctly stops an image anim…

toomanybrians:
The documentation  in kivy.uix.Image.anim_delay states that you can set
an value of -1 to cause an animation to be stopped.
kivy.core.image.Image._set_anim_delay() checks for this, however
kivy.core.image.Image.anim_reset() does not, incorrectly scheduling a
Clock event for -1 (same frame) leading to a “too much iteration”
error. This fix adds a check to anim_reset() to only schedule the Clock
event if anim_delay is >= 0.

### DIFF
--- a/kivy/core/image/__init__.py
+++ b/kivy/core/image/__init__.py
@@ -574,7 +574,7 @@ class Image(EventDispatcher):
         '''
         # stop animation
         Clock.unschedule(self._anim)
-        if allow_anim and self._anim_available:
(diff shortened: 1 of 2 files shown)


toomanybrians:
I added another commit which fixes a deprecated method in Python 3 in img_gif.py. I wrapped it in an if block checking for PY2 so it works for both versions.

PR #3972, opened 2016-02-04
Fixed loading unicode file names with gstplayer

Cheaterman:


kivy/core/video/video_gstplayer.py
```diff
@@ -59,6 +59,7 @@ def _on_gst_eos_sync(self):
         Clock.schedule_once(self._do_eos, 0)
 
     def load(self):
+        self._filename = self._filename.encode('utf8')
         Logger.debug('VideoGstplayer: Load <{}>'.format(self._filename))
```
matham: You will run into issues here if the default encoding cannot decode the filename. Because the format op will try to convert it to unicode using the default encoding which may fail. And the same anywhere else that tries to use `_filename` as a string in py3. So it's best to use the bytes version only at `GstPlayer(uri...)`. 

Also, you shouldn't unconditionally encode. In py3 bytes does not have a encode method. Although I'm not sure whether filename is allowed to be a bytes object in py3. Still if someone does pass a bytes filename this will fail.

Cheaterman:
I don't know mate, I just shared whatever works. If you wanna help, you should instead apply a more proper fix to this, and to ffpyplayer as well ;-)

PR #3974, opened 2016-02-04
Remove duplicate code for SelectableView

zeeMonkeez:
SelectableView is both defined in kivy.uix.selectableview and kivy.uix.listview, using identical code.
This removes duplicate code, but adds an import to maintain compatibility with code importing kivy.uix.selectableview.


dessant:
Nice find, but shouldn't this be done the other way around?

zeeMonkeez:
OK, I did not look at the history of the code. However, the fact kivy.uix.selectableview is only ever referred to in two examples suggests that that was the original location and it was later integrated into kivy.uix.listview?

zeeMonkeez:
@dessant Checking history it should indeed be the other way around. Will close this and prepare another one (if nobody else is faster).

PR #3975, opened 2016-02-04
Remove duplicate definition of SelectableView from kivy.uix.listview

zeeMonkeez:
replace references to `kivy.uix.listview.SelectableView` with references to `kivy.uix.selectableview.SelectableView`


Zen-CODE:
Nice and thorough, thanks :-). All the examples and both import paths work for me (i.e. from uix.listview still work, so should not break anything). But a bit nervous to merge: will leave that to the more heavyweight devs to look at, but will follow up if not merged soon.

dessant:
@zeeMonkeez, thanks.

PR #3981, opened 2016-02-07
Use the correct value of index in _sizes

ngoonee:
Currently using an offset value. Index does need incrementing in order to prevent an infinite loop, but the sizes dict should be filled using the previous value.


dessant:

PR #3985, opened 2016-02-08
Add functions to ListAdapter and DictAdapter to select ListViewItems by key

zeeMonkeez:
Currently, if a `ListAdapter`'s or `DictAdapter`'s selection is to be changed programmatically, the views have to be passed to `select_list`. This may be inconvenient, particularly when the adapter's data are updated and previously selected items should be selected again. This PR adds convenience functions that do the retrieval of the views in the background.


dessant:

PR #3986, opened 2016-02-08
DictAdapter: Fix index retrieval in trim* and cut_to_sel methods

zeeMonkeez:
Change trim_*_sel and cut_to_sel tests to use a view text different from the dict key.

Fix bug in DictAdapter.trim_*_sel and cut_to_sel that found index of
selection by text property of the view rather than the index property.


dessant:

PR #3995, opened 2016-02-11
Fix ListAdapter.select_list

zeeMonkeez:
ListAdapter.select_list has two issues: when `extend=False`, `self.selection` is cleared, but the corresponding views are not marked `is_selected=False`. Thus they keep their selected appearance. Furthermore, if an empty list is passed to `select_list`, the selection is cleared and no item is selected, which may not be allowed (if `selection_mode!='none'` and `allow_empty_selection=False`).
This PR both adds unit tests to discover the issues and introduces a fix for both.


dessant:

Issue #3364, opened 2015-05-16
Delimeter problem in kivy language.


gouzouni:
text: "Select an Index to read it's Description.." if root.selected_indices["feat_index"] is None else root.selected_indices["feat_index"].note

Delimeter in (it's) is ruining the statement.

kived:
String removal during the binding stage of the parser doesn't ensure that matched quotes are removed.

dessant:
Fixed in https://github.com/kivy/kivy/pull/4015.

PR #4006, opened 2016-02-17
Fix delimiter issue #3364

chenesan:
See issue #3364. When we have single quote in the double quoted
string in .kv property definition, It will failed to parse it
correctly.

For example:

```
<TestClass>
    Label:
        text: "I'm " + str(root.year) + " years old"
```

The regex, [lang_str]
(https://github.com/kivy/kivy/blob/master/kivy/lang/parser.py#L36)
in parser.py, will remove the double/single quote incorrectly.
(https://github.com/kivy/kivy/blob/master/kivy/lang/parser.py#L185)

This PR fix that problem.

### DIFF
--- a/kivy/lang/parser.py
+++ b/kivy/lang/parser.py
@@ -33,7 +33,7 @@ Cache.register('kv.lang')
 __KV_INCLUDES__ = []
 
 # precompile regexp expression
-lang_str = re.compile('([\'"][^\'"]*[\'"])')
+lang_str = re.compile(r'((((?<!\\)").*?((?<!\\)"))|(((?<!\\)\').*?((?<!\\)\')))')
 lang_key = re.compile('([a-zA-Z_]+)')
 lang_keyvalue = re.compile('([a-zA-Z_][a-zA-Z0-9_.]*\.[a-zA-Z0-9_.]+)')
 lang_tr = re.compile('(_\()')
@@ -160,7 +160,6 @@ class ParserRuleProperty(object):
 
         # first, remove all the string from the value
         tmp = sub(lang_str, '', self.value)
-
         # detecting how to handle the value according to the key name
         mode = self.mode
         if self.mode is None:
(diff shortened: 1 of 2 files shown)


chenesan:
Maybe I didn't describe the bug clearly, so here's the bug reproducing code:

`test.py`

```python
from kivy.app import App
from kivy.clock import Clock


class TestWidget(Widget):
    year = NumericProperty(0)

    def update(self, dt):
        self.year += 1

class TestApp(App):
    def build(self):
        wid = TestWidget()
        Clock.schedule_interval(wid.update, 1.0 / 60.0)
        return wid

if __name__ == '__main__':
    TestApp().run()
```

`test.kv`:

```python

<TestWidget>
    Label:
        font_size: 70
        center_x: root.width / 2
        center_y: root.height / 2
        text: "I am " + str(root.year) + " years old"
```

When you run the `test.py`, the label text will update label text regularly. But once we change the label text as `text: "I'm " + str(root.year) + " years old"`, It will fail to update the `root.year` property since the kv lang parser will fail to precompile the `root.year` property correctly. The PR fixes this issue.

matham:
This only fixes this particular issue, however there are lots of other quoting issues with the lang_str pattern that is not fixed here. The proper fix is https://github.com/matham/kivy/blob/lang-compiler/kivy/lang.py#L905 from the compiler branch. I.e. 

```py
lang_str = re.compile(
    "((?:'''.*?''')|"
    "(?:(?:(?<!')|''')'(?:[^']|\\\\')+?'(?:(?!')|'''))|"
    '(?:""".*?""")|'
    '(?:(?:(?<!")|""")"(?:[^"]|\\\\")+?"(?:(?!")|""")))', re.DOTALL)
```

matham:
I believe it also fixes your issue.

chenesan:
Thank you! I'll close this PR. And, when will the compiler fix be merged in kivy:master?

matham:
Probably a long way off. But I guess I can merge that fix now. I'll merge https://github.com/kivy/kivy/pull/4015 as soon as the tests pass.
Thanks for the PR though.

PR #4009, opened 2016-02-17
hint_text in TextInput shows when focused and no text entered

### LINKED ISSUE #3977: TextInput hint_text is hidden on focus
Hiding the hint text on focus limits the usability of the feature, `hint_text` should always be visible unless there is text in the input.

HTML input fields also behave this way: http://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_placeholder.

udiboy1209:
Fixes #3977 

screenshot from 2016-02-18 02 00 01

Works like the HTML input tag

Sample Code:

```python
from kivy.app import App
from kivy.lang import Builder

kv = """
AnchorLayout:
    TextInput:
        size_hint: 0.4, 0.05
        hint_text: 'hint text'
"""

class TestApp(App):
    def build(self):

if __name__ == '__main__':
    TestApp().run()
```

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -176,7 +176,7 @@
             pos: [int(x) for x in self.cursor_pos]
             size: 1, -self.line_height
         Color:
-            rgba: self.disabled_foreground_color if self.disabled else (self.hint_text_color if not self.text and not self.focus else self.foreground_color)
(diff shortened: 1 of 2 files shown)


dessant:
Beautiful. :+1:

PR #4022, opened 2016-02-22
Add option to sync Spinner dropdown children heights

ngoonee:
Currently the default height is 48 (from style.kv). This patch adds an option `sync_height` which would have each dropdown element's height set according to the current size of the Spinner, which aesthetically makes sense for most cases. If this defaults to True, user will not be able to customize sizes, so a False default makes more sense.

A resubmission of https://github.com/kivy/kivy/pull/3686#issuecomment-186065508 updated for rebasing.


dessant:
:+1:

PR #4024, opened 2016-02-22
Always show the textinput cursor at the moment of touch

### LINKED ISSUE #4017: Always show the textinput cursor at the moment of touch
This is just a small annoyance, but it would be nice to clean it up. The cursor blinking timer is not started/reset on a new touch, so unless the focus gain time coincides with the time when the cursor should appear, you'll see a delay until the cursor finally appears.

udiboy1209:
Fixes #4017

@dessant review please :)

A (pretty lame) test for this is repeated clicking, in which case the cursor should stay visible.

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -447,7 +447,6 @@ class TextInput(FocusBehavior, Widget):
 
     def __init__(self, **kwargs):
         self.is_focusable = kwargs.get('is_focusable', True)
-        self._cursor_blink_time = Clock.get_time()
         self._cursor = [0, 0]
         self._selection = False
         self._selection_finished = True
@@ -1721,13 +1720,11 @@ class TextInput(FocusBehavior, Widget):
     def _do_blink_cursor(self, dt):
         # Callback called by the timer to blink the cursor, according to the
         # last activity in the widget
-        b = (Clock.get_time() - self._cursor_blink_time)
-        self.cursor_blink = int(b * 2) % 2
+        self.cursor_blink = not self.cursor_blink
 
     def on_cursor(self, instance, value):
         # When the cursor is moved, reset the activity timer, and update all
         # the graphics.
-        self._cursor_blink_time = Clock.get_time()
         self._trigger_update_graphics()
 
     def _delete_line(self, idx):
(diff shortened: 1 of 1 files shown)


dessant:
I think `_cursor_blink_time` could be removed, I see no reason for having it.
When the cursor moves, on a click/touch, or on focus, we can just reset (unschedule etc), then schedule a method which toggles `cursor_blink`.

This means whenever one of the above three events occur, the cursor is immediately shown.

udiboy1209:
@dessant from what I understand about the implementation of `_do_blink_cursor`, not setting the `_cursor_blink_time` could lead to varying **on** time for the first blink, depending on when the touch happened after the last blink **on**->**off**.

This is because `_do_blink_cursor` doesn't toggle `cursor_blink`, it calculates it based on current time and `_cursor_blink_time`.

dessant:
Yes, that is the old and wrong implementation.

dessant:
We should not care about when any of the events happened, we just reset and show, then schedule.

udiboy1209:
Okay I'll remove `_cursor_blink_time` for now. We can fix the `cursor_blink` implementation next and this wont get affected.

dessant:
Well, i see it differently, the proper fix for the linked issue must entail rethinking how cursor_blink is handled, your pr would fix only one corner case.

udiboy1209:
Let me try a quick new implementation of `_do_cursor_blink`. If that breaks other stuff (that function is being used at many places), probably ditching this pr and fixing the `cursor_blink` issue first would be best.

udiboy1209:
Just changed to flipping `cursor_blink` instead of using time. Works pretty well on my machine. that was simple :grin:

udiboy1209:
@dessant I created a common method, and called it on mouse click, focus and cursor move as you requested.

dessant:
I've tested it, the cursor is shown now when needed.

The pr also fixes the issue when a blink is skipped if the clock drifts 1-2 frames, previously the blinking was sometimes irregular, even while having no interaction with the focused input (the cursor was visible for two cycles or disappeared for two).

A possible optimization would be to not call the reset 2-3 times when the focus is gained during a touch, or the cursor position changes because of a touch, but this could also introduce other corner cases and bugs. If someone has a clever solution, please make a new pr.

@udiboy1209 great job, your work is getting better and better. :trophy:

udiboy1209:
Thanks :)

PR #4032, opened 2016-02-25
Fix ScrollView not properly ignoring touch_up

Cheaterman:
Hi!

The condition in ScrollView's on_touch_up (added by Ryan in c4fe24c0) looks wrong, self.uid isn't used anywhere else in the code.

Sorry, I don't have a minimal example proving the behaviour to be wrong, it just simply seems self.uid won't ever be added to touch.ud.

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -771,7 +771,7 @@ class ScrollView(StencilView):
         return rv
 
     def on_touch_up(self, touch):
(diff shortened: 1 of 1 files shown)


Cheaterman:
Kived: Tell me if I was wrong, but I assume you intended to write what I commited?

dessant:
@kived, could you take a look?

tshirtman:
we have used this internally for some time without issues, merging.

Issue #3196, opened 2015-03-28
GridLayout ignores pos_hint/size_hint at layout time.


tshirtman:
GridLayout ignore pos_hint/size_hint (at least) for widgets that are smaller than the computed col_width/row_height

https://github.com/kivy/kivy/blob/prevent_scatter_scrollview_crash/kivy/uix/gridlayout.py#L431-443

andnovar:
Hello, Can you develop more on this ? I am trying to reproduce this issue but no success.

tshirtman:
as discussed on irc, here is a simple example:
https://gist.github.com/df54d13618c71158ffe6

- Label should always be the size of the texture, since (size_hint: None, None/size: self.texture_size), and this way it could be aligned with pos_hint inside the cell, and it's not
- for additional fun, halign doesn't work in this situation
- i know i could use text_size: self.size instead, but that's wasteful, and doesn't give as much flexibility as pos_hint
- another way would be to test all the labels in AnchorLayout, giving me full flexibility again, but quite wasteful too (creating a lot of overwise unnecessary widgets, and a lot of do_layout events)

My workaround in this case ended up being overriding label's canvas rule and place the texture where i wanted it, but that's not really beginner-friendly.

arlowhite:
I came across this bug when working on a Label example:
https://github.com/arlowhite/kivy/blob/label-text-size-doc/examples/widgets/label_texture_size.py
a8ad46b

It has a bunch of Labels inside a GridLayout. I noticed that the GridLayout does not size its children correctly when you shrink the window and expand its width again. The GridLayout height is fixed, but the children in rows 1 & 3 change their heights as the Labels wrap and StackLayout rearranges buttons. The 2nd row of GridLayout children (2 Labels) is the only row where size_hint_y is not None. This row does not re-expand to fill all GridLayout space for some reason. Though it does grow a little bit.

I overwrote my gridlayout.py with the one from andnovar/kivy/gridbug and it does fix this issue. #3216

dessant:
`size_hint` fixed in https://github.com/kivy/kivy/pull/4035.

dessant:
The pos_hint code should be lifted from here and tested: https://github.com/kivy/kivy/pull/3216/files#diff-ea7c2b5766ae42b4c545371831a6ef79R443.

matham:
@dessant although that diff is not quite correct.

dessant:
@matham, yeah, just a food for thought.

[...]

PR #4035, opened 2016-02-26
Respect size_hint in gridlayout.

matham:
I'm not sure how this has not been a issue before? 

Imagine a 3-row 1-column gridlayout. Imagine that the first child has `size_hint_x` `None` and the other two being `1`. Now, the first the first child will always be set to the column's width because the `size_hint_x` is ignored. 

This is a problem if e.g. you use the first child's "fixed" width to set other things. But if the first child is not fixed because the `size_hint_x` is ignored, everything else gets messed up.

### DIFF
--- a/kivy/uix/gridlayout.py
+++ b/kivy/uix/gridlayout.py
@@ -428,8 +428,11 @@ class GridLayout(Layout):
                 c = children[i]
                 c.x = x
                 c.y = y - row_height
-                c.width = col_width
-                c.height = row_height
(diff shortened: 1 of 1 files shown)


dessant:
I've avoided this before by not using gridlayout in some cases :P, thinking that's how the widget was designed to work, it has been definitely a problem for a long time. :+1:

dessant:
@matham `pos_hint` also appears to be an issue https://github.com/kivy/kivy/issues/3196.

PR #4042, opened 2016-02-28
fixed bug in pagelayout

DerThorsten:
Fixed bug if zero or one widgets are in pagelayout


kivy/uix/pagelayout.py
```diff
@@ -198,8 +198,12 @@ def on_touch_up(self, touch):
                 self._trigger_layout()
 
             touch.ungrab(self)
-        return self.children[-self.page + 1].on_touch_up(touch)
-
+        try:
+            r =  self.children[-self.page + 1].on_touch_up(touch)
+        except:
```
akshayaurora: do not use a blanket exception

kivy/uix/pagelayout.py
```diff
@@ -198,8 +198,12 @@ def on_touch_up(self, touch):
...
+            r = None
+        if r is not None:
+            return r
```
tshirtman: this is equivalent to "return r", since a function without return, still returns None.

It could be nicer to check for `if self.children` to check if the list is not empty, instead of the try/except, but a try/except IndexError, would be appropriate too.

PR #4047, opened 2016-02-29
Fix alpha rendering of text color for pygame

### LINKED ISSUE #4043: Pygame text provider does not render text opacity properly
Text renders as if opacity is set to 100% regardless of what's set in a Label's color attribute.

SDL2 works just fine.

Tested with master pulled today on Linux.

udiboy1209:
Fixes #4043

set color using set_colorkey, which is the way to include alpha when anti-aliasing=True.
ref: http://www.pygame.org/docs/ref/font.html#pygame.font.Font.render

I haven't gotten to testing this, had problems setting up pygame. I'd appreciate if someone does, but will try by myself too.

### DIFF
--- a/kivy/core/text/text_pygame.py
+++ b/kivy/core/text/text_pygame.py
@@ -97,6 +97,7 @@ class LabelPygame(LabelBase):
         color[0], color[2] = color[2], color[0]
         try:
             text = font.render(text, True, color)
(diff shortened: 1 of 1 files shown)


akshayaurora:
@udiboy1209 thanks, keep up the god work!

Issue #3853, opened 2015-12-19
TypeError: key_up() missing 2 required positional arguments: 'text' and 'modifiers'


dessant:
https://github.com/kivy/kivy/blob/master/examples/keyboard/main.py#L204
The callback method assumes a different event signature for on_key_up, note that both Keyboard and Window have this event, and they have different signatures.

```python
 Traceback (most recent call last):
   File "E:\dev\envs\py34_x86\kivy\examples\keyboard\main.py", line 220, in <module>
     KeyboardDemo().run()
   File "e:\dev\envs\py34_x86\kivy\kivy\app.py", line 828, in run
     runTouchApp()
   File "e:\dev\envs\py34_x86\kivy\kivy\base.py", line 487, in runTouchApp
     EventLoop.window.mainloop()
   File "e:\dev\envs\py34_x86\kivy\kivy\core\window\window_sdl2.py", line 613, in mainloop
     self._mainloop()
   File "e:\dev\envs\py34_x86\kivy\kivy\core\window\window_sdl2.py", line 546, in _mainloop
     self.dispatch('on_key_up', key, scancode)
   File "kivy\_event.pyx", line 714, in kivy._event.EventDispatcher.dispatch (kivy\_event.c:7654)
   File "kivy\_event.pyx", line 1224, in kivy._event.EventObservers.dispatch (kivy\_event.c:13497)
   File "kivy\_event.pyx", line 1148, in kivy._event.EventObservers._dispatch (kivy\_event.c:13023)
   File "e:\dev\envs\py34_x86\kivy\kivy\core\window\__init__.py", line 163, in _on_window_key_up
     return self.dispatch('on_key_up', keycode)
   File "kivy\_event.pyx", line 714, in kivy._event.EventDispatcher.dispatch (kivy\_event.c:7654)
   File "kivy\_event.pyx", line 1224, in kivy._event.EventObservers.dispatch (kivy\_event.c:13497)
   File "kivy\_event.pyx", line 1148, in kivy._event.EventObservers._dispatch (kivy\_event.c:13023)
 TypeError: key_up() missing 2 required positional arguments: 'text' and 'modifiers'
```

akshayaurora:
closed by #3883

dessant:
The events from both classes are used in the example, fitting the callback to the other one will still produce a crash.
Select `dock` mode, restart the app, press `Continue` and select one of the keyboard layouts, then use the docked keyboard.

```python
 Traceback (most recent call last):
[...]

PR #4060, opened 2016-03-06
Corrected arguments of key_up() function in keyboard example

CuriousLearner:
This is a fix for Issue #3853

### DIFF
--- a/examples/keyboard/main.py
+++ b/examples/keyboard/main.py
@@ -201,9 +201,9 @@ class KeyboardScreen(Screen):
         """ The callback function that catches keyboard events. """
         self.displayLabel.text = u"Key pressed - {0}".format(text)
 
-    def key_up(self, keyboard, keycode):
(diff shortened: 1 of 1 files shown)


dessant:
And now we've come full circle :D.
I don't expect this to fix the original issue: https://github.com/kivy/kivy/issues/3853#issue-123087902.

CuriousLearner:
Hi @dessant ,

I tried this patch in docked keyboard mode. I cannot re-produce the same error. Any pointers from your side would be really helpful. Thanks

dessant:
In the issue thread you linked, it can be seen that this pr will just revert the function signature to the original state.
https://github.com/kivy/kivy/issues/3853
https://github.com/kivy/kivy/pull/3883/files

The right fix must result in all keyboard modes working with the example.

dessant:
The pr reverts to the original state, please open a new one if you have the right fix (replacing some of the largs with keyword arguments should probably suffice).

Issue #2087, opened 2014-04-18
Carousel loop=True crash on load_slide()


bionoid:
When loop=True and there are > 2 slides, I get a crash when trying to load the last slide. It works if loop=False or the target slide is not the last in the list.

```python
from kivy.app import App
from kivy.uix.carousel import Carousel
from kivy.uix.label import Label

class TestApp(App):
    def build(self):
        c = Carousel(loop=True)
        c.add_widget(Label(text='index 0'))
        c.add_widget(Label(text='index 1'))
        c.add_widget(Label(text='index 2'))
        c.load_slide(c.slides[2])
        return c

if __name__ == '__main__':
    TestApp().run()
```

```
kivy.uix.widget.WidgetException: Cannot add <kivy.uix.relativelayout.RelativeLayout object at 0xa2b13ec>, it already has a parent <kivy.uix.carousel.Carousel object at 0xa02dbfc>
```

PR #4081, opened 2016-03-11
Fixed repeated addition of widget. references #2087

Zen-CODE:
Just had the same issue as #2087 when using the carousel as a page-able menu, The runnable example in the linked ticket demonstrates the issue.


dessant:
Thanks. :+1:

PR #4088, opened 2016-03-15
fix #4069 about readonly in TextInput

### LINKED ISSUE #4069: Disabling readonly on Double Tap in TextInput; backspace remains disabled
Found a strange issue in TextInput when trying to implement a double tap text entry by utilizing readonly. When double tapped entering text works fine, while backspace is still disabled. A crude way of fixing this is to unfocus and focus textinput.  For reproducability, please find the code below.

Kivy v1.9.1
Python v2.7.10
Windows 10

```python
class DoubleTapInput(TextInput):
    def __init__(self,**kwargs):
        super(DoubleTapInput,self).__init__(**kwargs)
        self.readonly=True
        self.txt_Dbl_Tap = False
    
    def on_double_tap(self):
        if self.txt_Dbl_Tap == False:
            self.readonly=False
            #self.focus = False  
            #self.focus = True
            self.txt_Dbl_tap = True
```

xg-wang:
This commit fixes #4069 

##Illustration
This issue is because in the src we have such bindings:
```Python
def handle_readonly(instance, value):
    if value and (not _is_desktop or not self.allow_copy):
        self.is_focusable = False

fbind('readonly', handle_readonly)
fbind('focus', self._on_textinput_focused)
handle_readonly(self, self.readonly)
```

Notice that here Kivy actually does nothing when you alters `self.readonly`.

However, when `on_double_tap` is triggered, it checks `self.readonly` before you actually sets it tobe `False`, then it sets the `self._editable` to be `False`.
```Python
def _on_textinput_focused(self, instance, value, *largs):
        self.focus = value

        win = EventLoop.window
        self.cancel_selection()
        self._hide_cut_copy_paste(win)

        if value:
            if (not (self.readonly or self.disabled) or _is_desktop and
                self._keyboard_mode == 'system'):
                self._trigger_cursor_reset()
                self._editable = True
            else:
                self._editable = False
        else:
            Clock.unschedule(self._do_blink_cursor)
            self._hide_handles(win)
```

This explains the reason that redundant focus call fixes.

I added the `_editable` property check&change binding to the `readonly` and it makes sense.

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -508,6 +508,11 @@ class TextInput(FocusBehavior, Widget):
         def handle_readonly(instance, value):
             if value and (not _is_desktop or not self.allow_copy):
                 self.is_focusable = False
+            if (not (value or self.disabled) or _is_desktop and
+                self._keyboard_mode == 'system'):
+                self._editable = True
+            else:
+                self._editable = False
 
         fbind('padding', update_text_options)
         fbind('tab_width', update_text_options)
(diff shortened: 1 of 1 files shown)


xg-wang:
@akshayaurora hi, would you check if this pr is correct?

akshayaurora:
@xg-wang nice work!

Issue #4090, opened 2016-03-15
Colorpicker example crashes when the clear button is pressed


dessant:
`examples/widgets/colorpicker.py`

```python
 Traceback (most recent call last):
   File "E:\dev\prs\kivy\examples\widgets\colorpicker.py", line 230, in <module>
     MainApp().run()
   File "E:\dev\prs\kivy\kivy\app.py", line 828, in run
     runTouchApp()
   File "E:\dev\prs\kivy\kivy\base.py", line 487, in runTouchApp
     EventLoop.window.mainloop()
   File "E:\dev\prs\kivy\kivy\core\window\window_sdl2.py", line 622, in mainloop
     self._mainloop()
   File "E:\dev\prs\kivy\kivy\core\window\window_sdl2.py", line 365, in _mainloop
     EventLoop.idle()
   File "E:\dev\prs\kivy\kivy\base.py", line 330, in idle
     self.dispatch_input()
   File "E:\dev\prs\kivy\kivy\base.py", line 315, in dispatch_input
     post_dispatch_input(*pop(0))
   File "E:\dev\prs\kivy\kivy\base.py", line 281, in post_dispatch_input
     wid.dispatch('on_touch_up', me)
   File "kivy\_event.pyx", line 718, in kivy._event.EventDispatcher.dispatch (kivy\_event.c:7726)
   File "E:\dev\prs\kivy\kivy\uix\behaviors\button.py", line 187, in on_touch_up
     self.dispatch('on_release')
   File "kivy\_event.pyx", line 714, in kivy._event.EventDispatcher.dispatch (kivy\_event.c:7681)
   File "kivy\_event.pyx", line 1224, in kivy._event.EventObservers.dispatch (kivy\_event.c:13524)
   File "kivy\_event.pyx", line 1108, in kivy._event.EventObservers._dispatch (kivy\_event.c:12356)
   File "E:\dev\prs\kivy\kivy\lang\builder.py", line 64, in custom_callback
     exec(__kvlang__.co_value, idmap)
   File "<string>", line 89, in <module>
 AttributeError: 'NoneType' object has no attribute 'canvas'
```

xg-wang:
fixed in #4094 

The demo crashes when clicking clear button without selecting any image.
Fixed by adding a check on the current_image.

Also fixed the display image name that may happen on different computers.

PR #4094, opened 2016-03-18
Fix colorpicker

xg-wang:
#4090 
The demo crashes when clicking clear button without selecting any image.
Fixed by adding a check on the current_image.

Also fixed the display image name that may happen on different computers.


akshayaurora:
@xg-wang thanks

PR #4106, opened 2016-03-20
Fix for shader editor example crash when editing on python 2

andong777:
The example shader editor crashed when editing on Python. The error is shown below.

> TypeError: Expected str, got unicode

It is due to a mix of unicode and str. We can simply use str() to wrap the variable, like this:

`fs = fs_header + str(self.fs)`

It works except for non-ascii chars. So I decide to use encode('utf-8') to do this.

`fs = fs_header + self.fs.encode('utf-8')`

I have tested it on Python 2.7 and Python 3.5.


akshayaurora:
@andong777 thanks, :+1:

PR #4107, opened 2016-03-20
ScreenManager: fix Screen removal leaving screen.parent property != None

rafalo1333:
Fixes situation where after removal the Screen obj from SM, the obj.parent was the reference of the SM that we have removed the screen from. The parent should be None, so we need to real_remove_widget() from SM widget tree.

### DIFF
--- a/kivy/uix/screenmanager.py
+++ b/kivy/uix/screenmanager.py
@@ -913,7 +913,7 @@ class ScreenManager(FloatLayout):
     def real_add_widget(self, *l):
         # ensure screen is removed from it's previous parent before adding'
         if l[0].parent:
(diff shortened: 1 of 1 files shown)


dessant:
Good strart, some other changes will be needed to clean up the sm state, i'll follow up with a pr. :+1:

rafalo1333:
Could the fix be merged to the stable 1.9.1 branch also or we need to use SM subclassing and overriding proper methods to avoid the bug?

dessant:
@rafalo1333, usually only critical fixes get backported into kivy stable.

rafalo1333:
OK. Hope 1.9.2 will debut soon ;)

PR #4111, opened 2016-03-21
Remove the last screen and leave ScreenManager in a valid state

### LINKED ISSUE #2655: ScreenManager.clear_widgets()
What should ScreenManager.clear_widgets() do?
At the moment, it can change its "current" attribute as it is iterating through the screens:
https://gist.github.com/cropleyb/b3f65d29fddc579ce741
(unless the current screen is the last screen to be iterated over)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5866483-screenmanager-clear_widgets?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dessant:
#4107 interferes with transitions, removing the current widget too soon, this pr fixes that.
`clear_widgets` is also overridden to ensure proper cleaning, fixes #2655.

### DIFF
--- a/kivy/uix/screenmanager.py
+++ b/kivy/uix/screenmanager.py
@@ -789,7 +789,7 @@ class ScreenManager(FloatLayout):
     By default, the manager will show only one screen at a time.
     '''
 
-    current = StringProperty(None)
+    current = StringProperty(None, allownone=True)
     '''
     Name of the screen currently shown, or the screen to show.
 
@@ -845,7 +845,7 @@ class ScreenManager(FloatLayout):
     [], read-only.
     '''
 
-    current_screen = ObjectProperty(None)
+    current_screen = ObjectProperty(None, allownone=True)
     '''Contains the currently displayed screen. You must not change this
     property manually, use :attr:`current` instead.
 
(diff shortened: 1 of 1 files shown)


kivy/uix/screenmanager.py
```diff
@@ -897,34 +897,46 @@ def remove_widget(self, *l):
...
+        remove_widget = self.remove_widget
+        for screen in screens:
+            remove_widget(screen)
 
-    def real_remove_widget(self, *l):
-        super(ScreenManager, self).remove_widget(*l)
+    def real_add_widget(self, screen, *args):
+        # ensure screen is removed from its previous parent
+        if screen.parent:
+            screen.parent.real_remove_widget(screen)
```
tshirtman: Since Screen can be added to any `Widget` not just `ScreenManager`, it's probably better to check if `screen.parent` is a `ScreenManager` instance and call `remove_widget` if it's not the case.

kivy/uix/screenmanager.py
```diff
@@ -897,34 +897,46 @@ def remove_widget(self, *l):
...
 
         if not screen in self.screens:
             return
+
         if self.current_screen == screen:
             other = next(self)
-            if other:
+            if screen.name == other:
+                self.current = None
+                screen.parent.real_remove_widget(screen)
```
tshirtman: same as line 927

kivy/uix/screenmanager.py
dessant: Thanks, nice find! :+1:

kivy/uix/screenmanager.py
dessant: `parent` must be a sm here because it's the current screen.

tshirtman:
LGTM

Issue #4092, opened 2016-03-17
DropDown/Spinner does not toggle if added to RelativeLayout (if layout pos != 0, 0)


werton:
on_touch_down() method of DropDown widget not properly handling touch event if self.attach_to attached to relativelayout. I.e. self.attach_to has relative coordinates, but touch.pos has absolute positioning. touch.pos should be translated to_widget position, otherwise there will be unnecessary call self.dismiss().
the problem this piece of code:

```
    def on_touch_down(self, touch):
        if super(DropDown, self).on_touch_down(touch):
            return True
        if self.collide_point(*touch.pos):
            return True
        if self.attach_to and self.attach_to.collide_point(*touch.pos):  # touch.pos should be translated to_widget position
            return True
        if self.auto_dismiss:
            self.dismiss()
```
Here runnable example, the Button 1 is attached to RelativeLayout and can't be toggled as opposed to Button 2 which not attached to RelativeLayout

```
from kivy.lang import Builder
from kivy.base import runTouchApp

w = Builder.load_string('''
Widget:
    RelativeLayout:
        pos: 0, 200

        Button:
            id: btn1
            pos: 0,0
            text: 'Button 1'
            on_release: dropdown.open(self)

    Button:
        id: btn2
        pos: 200,0
        text: 'Button 2'
        on_release: dropdown.open(self)

    DropDown:
        id: dropdown
        Button:
            text: 'Value A'
            size_hint_y: None
            on_release: dropdown.select('A')

        Button:
            text: 'Value B'
            size_hint_y: None
            on_release: dropdown.select('B')

''')

runTouchApp(w)
```

werton:
Updated example. In it two buttons, one button is placed on RelaytiveLayout. If you click the buttons you will see that dropdown's button1 is not correctly handling the event on_touch_down(i.e. dropdown cannot be closed by clicking on
button after opening), it occurs in consequence of the fact that on_touch_down method does not translate the coordinates of button 1 to the absolute coordinates.

```
from kivy.base import runTouchApp
from kivy.lang import Builder
from kivy.uix.widget import Widget

Builder.load_string('''
<MyWidget>:
    # This is a debug rendering dropdown's button position which received dropdown.on_touch_down method
    touch_coordinates: [0, 0]
    dropdown_button_coord: [0, 0, 0, 0]

    canvas.after:
        Color:
            rgba: 0, 1, 0, 0.5
        Rectangle:
            pos: self.dropdown_button_coord[:2]
[...]

PR #4112, opened 2016-03-21
Fixed issue #4092 in dropdown.py

werton:
Fixed issue #4092.The absolute coordinates of the touch.pos converted to relative coordinates of self.attach_to(dropdown's button) to prevent incorrect processing when the dropdown's button attached to RelaytiveLayout.


akshayaurora:
@werton thanks!

PR #4120, opened 2016-03-28
Sdl2 vsync

tshirtman:
needs testing, only support sdl2.

### DIFF
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -304,7 +304,7 @@ from weakref import ref
 _is_rpi = exists('/opt/vc/include/bcm_host.h')
 
 # Version number of current configuration format
-KIVY_CONFIG_VERSION = 15
+KIVY_CONFIG_VERSION = 16
 
 Config = None
 '''The default Kivy configuration object. This is a :class:`ConfigParser`
@@ -795,7 +795,7 @@ if not environ.get('KIVY_DOC_INCLUDE'):
             Config.setdefault('graphics', 'min_state_time', '.035')
 
         elif version == 15:
-            Config.setdefault('graphics', 'vsync', 0)
+            Config.setdefault('graphics', 'vsync', 'None')
 
         # elif version == 1:
         #    # add here the command for upgrading from configuration 0 to 1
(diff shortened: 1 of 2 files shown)


dessant:
Looks great! It should be added to the config docs, and a versionadded at the bottom of the config list.

kived:
The config should probably accept three values - 0, 1 and None, defaulting to None. None wouldn't call `SDL_GL_SetSwapInterval` at all, leaving it at the user's default setting. Also if set to 1, maybe it should call `SDL_GL_SetSwapInterval(-1)` first, to attempt to enable late swap tearing, then call it with 1 if that call fails.

> Some systems allow specifying -1 for the interval, to enable late swap tearing. Late swap tearing works the same as vsync, but if you've already missed the vertical retrace for a given frame, it swaps buffers immediately, which might be less jarring for the user during occasional framerate drops. If application requests late swap tearing and the system does not support it, this function will fail and return -1. In such a case, you should probably retry the call with 1 for the interval.
> 
> Late swap tearing is implemented for some glX drivers with GLX_EXT_swap_control_tear and for some Windows drivers with WGL_EXT_swap_control_tear.

dessant:
@kived mentioned that it is not clear how sdl vsync behaves with the kivy clock or even if it has any effect on a gl surface, these should be investigated.

ignus2:
Adding my experience:
VSync is not a binary thing, that's why it's called Swap**Interval**. It can very well have a value of 2,3,etc., all of which have a valid meaning and use-case in certain scenarios.
If someone wants to fiddle with vsync, there is a reason for that. As such, I believe if vsync is set to a value, swapinterval should be set exactly to that value.
I have specifically a need for vsync, and currently I worked around it in kivy using pysdl2 to call directly SDL_SetSwapInterval() to the value I need (which can also change while the application is running in my use-case).
Moreover, I think the setting "vsync" should be renamed to "vsynccount" or "vsyncinterval" or something similar.
Also, this setting should be possible to change on-the-fly while the app is running.

Some more observations:
Having vsync on usually makes SDL_SwapWindow() block. As a side note (I'll open another issue for this), SDL_SwapWindow() should be done in "with nogil:", as currently it blocks the whole python program, which can make things behave very badly if other threads are running (a  video playback thread for example).

@kived What issues are there regarding kivy clock and vsync?

Issue #4124, opened 2016-03-29
Crash when Slider is imported before forking process on Mac OS X


zeeMonkeez:
Consider

```
#!/usr/bin/env python2
import multiprocessing
from kivy.app import App

from kivy.uix.slider import Slider

class Test(App):
    def build(self):
        return Slider()

def run_app():
    app = Test()
    app.run()

running_app = multiprocessing.Process(target=run_app)
running_app.daemon = True
running_app.start()
running_app.join()
```

This currently crashes on Mac OS X:

> *** multi-threaded process forked ***
> crashed on child side of fork pre-exec
> USING_FORK_WITHOUT_EXEC_IS_NOT_SUPPORTED_BY_FILE_MANAGER
> 

This is because the property `padding` is initialized with `NumericProperty(sp(16))`. This call to `sp` will attempt to initialize SDL. Cf. [this question on StackOverflow](http://stackoverflow.com/questions/8106002/using-the-python-multiprocessing-module-for-io-with-pygame-on-mac-os-10-7)

akshayaurora:
Can you explain what you are trying to do here?

akshayaurora:
Kivy does not support multiple apps. So simple question what is the intention here? To try and make kivy support multiple apps or ...?

akshayaurora:
ok so simply use subprocess.popen. 
Do you have to use multiprocessing?

akshayaurora:
I guess using multiprocessing makes it easier for you to send messages between the processes there. 

It's just one of the use cases that I haven't seen be used till now, people usually use separate processes of service on mobiles and osc or other options for communication.

If are submitting a pr fixing this then could you please make sure to add a test too?

zeeMonkeez:
@akshayaurora This solution was actually suggested by @inclement (or so I interpreted his comment to [my question on SO](http://stackoverflow.com/questions/34915939/nested-alternating-apps-with-kivy)).

I'd be more than happy to submit a test case. Not sure however what to test for. I suppose there'd be one test for each user interface module to make sure importing before forking works?

zeeMonkeez:
For now, I'll leave both #4127 and #4126 open. I like the latter better for it separates styling from logic. But either one solves my issue.

PR #4125, opened 2016-03-29
Workaround for issue on MaxOS X: pygame/SDL and multiprocessing

zeeMonkeez:
This is a workaround for #4124, allowing to import `kivy.uix.slider` before a process running an app is forked.

### DIFF
--- a/kivy/uix/slider.py
+++ b/kivy/uix/slider.py
@@ -50,7 +50,7 @@ class Slider(Widget):
     :attr:`max` is a :class:`~kivy.properties.NumericProperty` and defaults to
     100.'''
 
-    padding = NumericProperty(sp(16))
+    padding = NumericProperty(16)
     '''Padding of the slider. The padding is used for graphical representation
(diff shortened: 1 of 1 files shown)


dessant:
@zeeMonkeez, it is unlikely for this change to get merged, there are thousands of places where kivy will ensure a window being present.

I would suggest working around the issue and using Popen if you intend to start multiple apps.

zeeMonkeez:
There is a difference though between ensuring the window is open at time of module import and at time of object instantiation.

zeeMonkeez:
@dessant That parsing must happen at a later stage. Just tested #4124 with `Label`, which indeed uses `font_size = NumericProperty('15p')`, but that does not crash.

zeeMonkeez:
@dessant I can fix a PR that uses the `'16sp'` syntax if it is more likely to be accepted than #4126. Slider is the only instance calling `sp` directly to initialise a property, as far as I can tell.

dessant:
If `NumericProperty('16sp')` fixes your issue, we are also ok with it.

PR #4127, opened 2016-03-30
Change use of dimension conversion in Slider

### LINKED ISSUE #4124: Crash when Slider is imported before forking process on Mac OS X

zeeMonkeez:
Change to convert unit at instantiation, as is done in other classes, such as kivy.uix.slider.Slider.

Fixes #4124.

Discussed with @dessant in #4125.

### DIFF
--- a/kivy/uix/slider.py
+++ b/kivy/uix/slider.py
@@ -23,7 +23,6 @@ __all__ = ('Slider', )
 from kivy.uix.widget import Widget
 from kivy.properties import (NumericProperty, AliasProperty, OptionProperty,
                              ReferenceListProperty, BoundedNumericProperty)
-from kivy.metrics import sp
 
 
 class Slider(Widget):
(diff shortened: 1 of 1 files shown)


dessant:
@zeeMonkeez, thanks. :wink:

PR #4131, opened 2016-03-30
Fix ImportError kivy:  (missing -l python3.5m flags)

haricot:
Fix [658] (https://github.com/kivy/python-for-android/issues/658)


dessant:
@inclement plop.

kived:
This will break compilation on Python 3.4. This is an Android-specific problem, and should probably be fixed in python-for-android.

dessant:
Please fix the issue on the p4a side, if possible.

PR #4132, opened 2016-03-31
DropDown: include a min delay before dismissing

akshayaurora:


matham:
I think it should use the button's min delay (the button that is the main dropdown button). And if that has been exceeded as measured from the time when the touch was pressed, then it should schedule with a delay of zero. Also, I think a option should be added to dismiss so that by default it uses schedule but it could also be made without the clock schedule.

tshirtman:
Ok, the issue is that there is a button only if Spinner (or some other user-defined class we don't know about) is used, if one uses a DropDown on, say, a TextInput (not uncommon for filter search), there is no button to bind to, so DropDown needs to have an option too, and Spinner needs to pass its value to it, but now it means the option_cls passed by user has to accept it, which is an api break (as it wasn't a requirement before, and it will break apps).

matham:
But the problem may not be present for a textinput, only a button?

tshirtman:
Ok, forget about my comment, got confused about the Button, though it was the one used to open the `DropDown`, not the one used by options, since we already assume `option_cls` to be a subclass of `ButtonBehavior` (as we bind on its `on_release` event), we can also check its `min_state_delay` value.

tshirtman:
Ok, after more discussion, since the `Clock` event is in `DropDown`, and that its `select` data is arbitrary, (might not be a widget at all), we can't use it to determine a delay, `Spinner` has the right information, but its not managing the dismiss.

The easier solution, as pointed by @akshayaurora is to just add a `dismiss_delay` to `DropDown`, and let user put whatever value, or bind it to whatever they want, we can keep .1 as a default value (so button's state is visible), but 0 could be another acceptable default (the bug this PR is about is still fixed with it).

PR #4136, opened 2016-04-01
Fix model center not syncing with window center.

matham:
What happened in the past, a model resize, say in `height`, would cause a recalc of `center_y` which would bring the model out of sync with the window. Next in the chain would be a dispatch for `size` that calls `_align_center`. but `center` would still be stale and would equal the window center from before because the window center is always the same (a reference property does not always check the underlying props so it can become stale temporarily) so when `center` is set in `_align_center` to the window center nothing would happen because it was already that and `y` will not be updated to make it center for real. Finally, `center` would be updated due to the `center_y` only **after** `_align_center` is done and at that point center reflects that it's not the same as the window's `center`, but nothing is bound to `center` so it won't fix it.

By binding to `center` it fixes everything.

### DIFF
--- a/kivy/uix/modalview.py
+++ b/kivy/uix/modalview.py
@@ -182,19 +182,11 @@ class ModalView(AnchorLayout):
             on_resize=self._align_center,
             on_keyboard=self._handle_keyboard)
         self.center = self._window.center
-        self.fbind('size', self._update_center)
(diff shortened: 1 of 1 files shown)


dessant:
Flow diagram needed. :stuck_out_tongue_winking_eye:

PR #4138, opened 2016-04-02
Accept unicode strings as window title in SDL2

### LINKED ISSUE #4104: How to do that self.title use utf-8 string
```
# coding:utf-8
import kivy
kivy.require('1.9.1')
# python v 2.9.10
from kivy.app import App


class TestApp(App):

    def build(self):
        #self.title = u'啦唔呦'
        #self.title = u'\u5566\u5514\u5466'
        #self.title = (u'\u5566\u5514\u5466').encode('utf-8')

```
I know Button(text='你好', font_name="1.ttf") and TextInput(text="你好",font_name="1.ttf") 
But self.title ，how to do with the single string set

udiboy1209:
Fix #4104

I'm not sure whether this will break other window providers. I looked at the code for pygame and x11, they both seem to accept unicode strings. But I have only tested with SDL2.

### DIFF
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -814,7 +814,7 @@ class App(EventDispatcher):
         window = EventLoop.window
         if window:
             self._app_window = window
-            window.set_title(self.get_application_name())
+            window.set_title(self.get_application_name().encode('utf-8'))
             icon = self.get_application_icon()
             if icon:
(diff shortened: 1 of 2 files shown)


akshayaurora:
closed by 191158a6ae4bebd0e5b859b9349696230d50eaef

PR #4146, opened 2016-04-05
Improve touch experience in ScrollView widget

### LINKED ISSUE #4145: ScrollView: Imprecise touch events registered as scroll, should propogate touch
Imprecise touch events - especially on mobile/tablet - require exact accuracy even if the motion is within scroll_distance - they register as scroll events even though they really should be touch.

This leads to a cumbersome and difficult to use experience when dealing with touch aware widgets such as text boxes / buttons, etc. inside a scroll view.

If the scrollView receives a touch event within scroll_distance it should register as a touch.

brentpicasso:
register imprecise touches as touch events instead of scroll events, making mobile / touchscreen UIs work dramatically better when dealing with button or touchable type widgets in a ScrollView. Resolves #4145

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -658,6 +658,8 @@ class ScrollView(StencilView):
             'mode': 'unknown',
             'dx': 0,
             'dy': 0,
+            'y_start': touch.y,
+            'x_start': touch.x,
             'user_stopped': False,
             'frames': Clock.frames,
             'time': touch.time_start}
@@ -809,7 +811,6 @@ class ScrollView(StencilView):
         if self._get_uid() not in touch.ud:
             return False
 
-        self._touch = None
         uid = self._get_uid()
         ud = touch.ud[uid]
         if self.do_scroll_x and self.effect_x:
@@ -827,6 +828,17 @@ class ScrollView(StencilView):
             if not ud['user_stopped']:
                 self.simulate_touch_down(touch)
             Clock.schedule_once(partial(self._do_touch_up, touch), .2)
+
+        #check if we're touch selecting imprecisely
+        #but within range of scroll_distance
+        y_start = ud.get('y_start')
+        x_start = ud.get('x_start')
+        y_select = self.do_scroll_y and y_start and abs(y_start - touch.y) < self.scroll_distance
+        x_select = self.do_scroll_x and x_start and abs(x_start - touch.x) < self.scroll_distance
+        if ud['mode'] == 'scroll' and (x_select or y_select):
+            self.simulate_touch_down(touch)
+            Clock.schedule_once(partial(self._do_touch_up, touch), .1)
+
         Clock.unschedule(self._update_effect_bounds)
         Clock.schedule_once(self._update_effect_bounds)
 


kivy/uix/scrollview.py
```diff
@@ -809,7 +811,6 @@ def on_scroll_stop(self, touch, check_children=True):
         if self._get_uid() not in touch.ud:
             return False
 
-        self._touch = None
```
brentpicasso: This looks like a duplicate line, so removing. Not really related to this PR.

akshayaurora:
Please provide explanation about the issue first and a small snippet that reproduces the issue that this is supposed to fix.

brentpicasso:
The issue is summarized in the PR text, but the more elaborate version:

When using Kivy's ScrollView with internal button type widgets (check boxes, toggle buttons, etc) there is a problem easily activating these buttons. 

When pressing the button with a small amount of im-preciseness, such as a slight amount of scroll, it immediately registers as a scroll event, and button is not pressed.

What should happen is that the button press should occur without scrolling, providing the slight scrolling is within a threshold.  In this PR, I chose the scroll_distance value as that threshold.

Without this fix, selecting items in a scroll view is very inconsistent and provides a very bad experience. 

I don't have time immediately to create a canned demo, but it would look like this:

A scrollView with checkboxes or toggle buttons listed inside.  An android .buildozer configuration to show on a mobile platform (or launch on a laptop with a touch screen).

tito:
I can relate to that, android devices have poor experience with scrollview + touchable widgets. You're far more precise on desktop than a mobile.

The PR looks good enough, but unsure how to test. I think it can be merged.

brentpicasso:
With this PR it's dramatically easier to select widgets in a scroll view. Originally I thought it was kivy missing touch events, since so many of our panes were in ScrollViews; but I realized that ScrollView was indeed getting the touch events, but just dropping them on the ground.

I was watching other apps and it seems that there's always a little bit of friction where the view won't scroll until X amount of motion is detected, then it starts scrolling. If there was a touch release within that threshold, it registers as a touch.  I think ScrollView could be re-written to emulate this, but this PR is as close as I could get for now.

brentpicasso:
Here's a code sample that can repro it. To test on android you'll need to supply your own buildozer.spec file.









import kivy
kivy.require('1.9.1')

from kivy.app import App
from kivy.uix.togglebutton import ToggleButton

class ScrollTest(BoxLayout):
    def __init__(self, **kwargs):
        super(ScrollTest, self).__init__(**kwargs)

        sv = ScrollView(size_hint=(1.0, 1.0), do_scroll_x=False)
        self.add_widget(sv)
        grid = GridLayout(height=5000, cols=1, size_hint_y=None)
        sv.add_widget(grid)
        for x in range(0,40):
            grid.add_widget(ToggleButton(text='Select me {}'.format(x)))

class MyApp(App):

    def build(self):
        return ScrollTest()

if __name__ == '__main__':
    MyApp().run()

dessant:
@akshayaurora, @brentpicasso: the example shared on this thread reproduces #4150, please create a new pr and make sure it is tested.

brentpicasso:
Hi @akshayaurora  - can you paste the buildozer.spec you used to repro this on android, so I can verify ?  Thanks.

dessant:
@brentpicasso, the bug should be there on desktop too, start_y and touch.y are equal, regardless of how long is the drag, check out the issue linked above.

Issue #1078, opened 2013-04-03
Updating ScrollView from outside and scrolling with mouse after.


Naomak:
When I set ScrollView's scroll_y(x) property from outside (by some button), position of content is updated as supposed. But then when you try to scroll (with mouse wheel) after increasing scroll_y value (still in range 0-1), content and its move animation starts from place before 'outside' updating. When decreasing scroll_y and then scrolls with mouse, it works nicely though.

Test code in gist: https://gist.github.com/Goffrey/5299929
Steps to see problem:
1. Scrolls completely down with wheel.
2. Scroll UP to middle with Up button.
3. Now scroll with mouse wheel:
    \* when scrolling up - it jumps down to little scrolled height from bottom
    \* when scrolling down - it jump completely to the bottom
##

fgmcoelho:
This issue still happens on version 1.8 and 1.9.

Here is some info:
https://groups.google.com/forum/?hl=en-US&fromgroups=#!topic/kivy-users/PlSwk_MXe3s

dessant:
I can confirm that scrollview is jumping to the bottom in certain cases.

fgmcoelho:
Hi, i checked the code to try to track the issue. It seems that the problem is on the _update_effect_x and y. Unfortunately it is a holiday in my country and I won't be able to look further to this issue until Monday, but this should give you a pointer if you want to take a look.

Thanks for revisiting this issue so fast.

fgmcoelho:
Adding further info:
calling _update_effect_x_bounds() and _update_effect_y_bounds() after setting the values of scroll_x and scroll_y will remove problem of it jumping back to the old position.

But if you try to replace the scroll view when it is still moving from a long drag, the position will still bug.

dessant:
refs
https://github.com/kivy/kivy/issues/3382
https://github.com/kivy/kivy/issues/2038#issuecomment-46158938

tshirtman:
I worked around this in my app by calling `_update_effect_y_bounds()` when scroll_y was changed from code, i'm not familiar enough with SV code to know how wrong/inefficient it is, but maybe adding it to _update_from_scroll would solve the issue in a more general way.

Bakterija:
I just wanted to do a pull request that solves this problem and accidentaly commited the changes.
[...]

PR #4180, opened 2016-04-16
Scroll to touch pos if the touch is within the scrollbar but does not collide with the handle

dessant:
For testing:

```python
from kivy.app import App
from kivy.lang import Builder

kv = """
<NewB@Button>:
    size_hint: None, None
    size: dp(4000), dp(40)


<Test>:
    ScrollView:
        scroll_type: ['bars', 'content']
        scroll_wheel_distance: dp(80)
        bar_width: dp(24)
        GridLayout:
            id: grid
            cols: 1
            size_hint: None, None
            size: self.minimum_size
"""

Builder.load_string(kv)


class Test(BoxLayout):
        super(Test, self).__init__(*args, **kwargs)
        for b in range(40):
            self.ids.grid.add_widget(Factory.NewB(text=str(b)))


class TestApp(App):
    def build(self):
        return Test()

if __name__ == '__main__':
    TestApp().run()
```

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -667,16 +667,14 @@ class ScrollView(StencilView):
             return self.simulate_touch_down(touch)
 
         if in_bar:
-            if (touch.ud['in_bar_y'] and not
+            if (ud['in_bar_y'] and not
                     self._touch_in_handle(
                         self._handle_y_pos, self._handle_y_size, touch)):
                 self.scroll_y = (touch.y - self.y) / self.height
-                return True
-            elif (touch.ud['in_bar_x'] and not
+            elif (ud['in_bar_x'] and not
                     self._touch_in_handle(
                         self._handle_x_pos, self._handle_x_size, touch)):
                 self.scroll_x = (touch.x - self.x) / self.width
-                return True
 
         # no mouse scrolling, so the user is going to drag the scrollview with
         # this touch.
@@ -688,7 +686,7 @@ class ScrollView(StencilView):
             'mode': 'unknown',
             'dx': 0,
             'dy': 0,
-            'user_stopped': False,
+            'user_stopped': in_bar,
             'frames': Clock.frames,
             'time': touch.time_start}
 
@@ -703,7 +701,7 @@ class ScrollView(StencilView):
 
         if not in_bar:
             Clock.schedule_once(self._change_touch_mode,
-                                    self.scroll_timeout / 1000.)
+                                self.scroll_timeout / 1000.)
         return True
 
     def on_touch_move(self, touch):
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -584,6 +584,11 @@ class ScrollView(StencilView):
             touch.grab(self)
             return True
 
+    def _touch_in_handle(self, pos, size, touch):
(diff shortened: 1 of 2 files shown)


kivy/uix/scrollview.py
```diff
@@ -584,6 +584,11 @@ def on_touch_down(self, touch):
             touch.grab(self)
             return True
 
+    def _touch_in_handle(self, pos, size, touch):
+        x, y = pos
+        width, height = size
+        return x <= touch.x <= x + width and y <= touch.y <= y + height
```
dessant: Is there a kivy api for this? The line was mostly copied from collide_point.

kivy/uix/scrollview.py
```diff
@@ -661,6 +666,18 @@ def on_scroll_start(self, touch, check_children=True):
         if scroll_type == ['bars'] and not in_bar:
             return self.simulate_touch_down(touch)
 
+        if in_bar:
+            if (touch.ud['in_bar_y'] and not
+                    self._touch_in_handle(
+                        self._handle_y_pos, self._handle_y_size, touch)):
+                self.scroll_y = (touch.y - self.y) / self.height
+                return True
```
kived: I think it works better if you remove both of these `return True` statements. That way it jumps to the new position, and if you keep holding it will let you continue scrolling in that area.

dessant: @kived, that's a nice tweak, I really like the behavior. :dancer:

kived:
Also, needs to cancel the scroll effect if active - if I drag to scroll, then click in the bar outside the handle it jumps back and continues scrolling as it was.

dessant:
@kived, that may be an issue with scroll_x/y: https://github.com/kivy/kivy/issues/1078#issuecomment-148952835.

kived:
No, I'm saying that if you start scrolling via touch, and the scroll effect is still active - the content is still moving - you have to stop the scroll effect. Otherwise, when you touch in the bar, it does jump to the new scroll position properly, but then the scroll effect keeps running causing it to jump back.

dessant:
@kived, wouldn't that mean it is unsafe to use the scroll_y api while the content is scrolling?

kived:
I wouldn't call it unsafe, but it won't work. Basically, you're setting scroll_y once, but the effect is setting scroll_y every frame, and you didn't tell it to stop.

dessant:
@kived, it sounds like the scroll_y api should be fixed, and running effects should be stopped in the scroll_y callback.

dessant:
@kived, I've opened an issue for the scroll_x/y bug, is the rest of the change ok?

akshayaurora:
Should this be made optional? I use this only for scroll_type = ['bars'] personally however I see there could be other use cases. Maybe we should make this optional for those who want to enable it or disable it?

dessant:
I don't see a need for making it optional if the scroll bar is enabled, what would be the use case?

kived:
The only use case I can think of is if you want your scroll bars to page down/up when clicking outside the handle, like some desktop apps do. However, that's not how Kivy works right now anyway, you just end up grabbing the handle from wherever you clicked.

So if someone wants to add that functionality, then I can agree with making an option to select which functionality you want. But this is far better than the current functionality.

PR #4187, opened 2016-04-18
BUGFIX: Parser.execute_directive() doesn't use resource_find() for in…

ericatkin:
…clude directive.


rafalo1333:
I've tested on 1.9.1 and it fixes the issue of not recognizing the kv files loaded multiple times with #:include directive.

akshayaurora:
I can't reproduce this, please a provide a snippet for the issue this is supposed to solve.

rafalo1333:
@akshayaurora https://bitbucket.org/rafalo1333/kvlangincludekverror/src

akshayaurora:
@ericatkin @rafalo1333 thanks

PR #4192, opened 2016-04-19
cancel suggestion_text when lost focus

### LINKED ISSUE #4044: TextInput suggestion_text bugs
When the input focus is lost, the suggestion text stays there.

Pressing tab moves the cursor above the suggestion text.

I've ran kivy/uix/textinput.py while testing, the example makes use of the api.

xg-wang:
fix #4044 test passed on kivy/uix/textinput.py

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -1636,6 +1636,7 @@ class TextInput(FocusBehavior, Widget):
 
         win = EventLoop.window
         self.cancel_selection()
+        self.cancel_suggestion_text()
         self._hide_cut_copy_paste(win)
 
         if value:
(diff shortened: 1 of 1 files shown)


kivy/uix/textinput.py
```diff
@@ -2827,9 +2828,25 @@ def on_suggestion_text(self, instance, value):
...
         self._lines_labels[cr] = lbl.texture
+        self.suggestion_row = cr
         rct.size = lbl.size
         self._update_graphics()
 
+    def cancel_suggestion_text(self):
+        '''Cancel current suggestion text(if any)
+        '''
+        if self.suggestion_text != "" and self.suggestion_row >= 0:
+            print(self.suggestion_row)
```
akshayaurora: left over from testing?

xg-wang: Yes it is, should I create another pr or another commit would be fine?

PR #4200, opened 2016-04-20
The FloatLayout Ignores children's size_hints with values of zero.

FluxIX:
The FloatLayout has been modified to handle children's size_hint values of 0. Prior to this change, the size_hint value would be ignored and the size would not be modified.

### DIFF
--- a/kivy/uix/floatlayout.py
+++ b/kivy/uix/floatlayout.py
@@ -78,12 +78,17 @@ class FloatLayout(Layout):
         for c in self.children:
             # size
             shw, shh = c.size_hint
-            if shw and shh:
-                c.size = w * shw, h * shh
-            elif shw:
-                c.width = w * shw
-            elif shh:
-                c.height = h * shh
+            if shw is not None:
+                local_width = w * shw
+            else:
+                local_width = c.width
+
+            if shh is not None:
+                local_height = h * shh
+            else:
+                local_height = c.height
+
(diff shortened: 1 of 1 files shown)


kivy/uix/floatlayout.py
```diff
@@ -78,12 +78,17 @@ def do_layout(self, *largs, **kwargs):
...
+                local_width = w * shw
+            else:
+                local_width = c.width
+
+            if shh is not None:
+                local_height = h * shh
+            else:
+                local_height = c.height
+
+            c.size = local_width, local_height
```
matham: You're reassigning the child's width/height to itself when the hint is None. That's very inefficient because the property gets re-evaluated. You should probably do it like before but just replace e.g. `if shw and shh:` with `if shw is not None and shh is not None:`.

FluxIX:
How expensive if property evaluation? The check to determine if the value changed shouldn't be expensive.

kived:
Not horribly expensive, but certainly more expensive than avoiding it.

matham:
Yeah, in addition to the fact that you're assigning to `size` which then checks/sets `width` and `height`. So it's multiple properties that get checked.

FluxIX:
OK. I'll change it.

FluxIX:
I changed the checking form to the previous form and added the explicit None checks.

PR #4202, opened 2016-04-20
StringProperty: convert None to empty string

kived:
Makes using `StringProperty` easier. No more `text: root.my_text or ''`. Option `none_is_empty` defaults to `True` but can be disabled, and does nothing if `allownone` is `True`.

### DIFF
--- a/kivy/properties.pxd
+++ b/kivy/properties.pxd
@@ -45,7 +45,7 @@ cdef class NumericProperty(Property):
     cdef float parse_list(self, EventDispatcher obj, value, ext)
 
 cdef class StringProperty(Property):
-    pass
+    cdef bint _none_is_empty
 
 cdef class ListProperty(Property):
     pass
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -589,11 +589,19 @@ cdef class StringProperty(Property):
     :Parameters:
         `defaultvalue`: string, defaults to ''
             Specifies the default value of the property.
+        `none_is_empty`: bool, defaults to True
+            Convert None to an empty string '' if None is not an allowed value.
 
     '''
 
     def __init__(self, defaultvalue='', **kw):
+        self._none_is_empty = kw.get('none_is_empty', True)
         super(StringProperty, self).__init__(defaultvalue, **kw)
+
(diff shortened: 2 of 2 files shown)


matham:
I don't fully understand where this would be useful. When `allownone` is `False` this would only work for the initial case where the default value was set to `None` explicitly (later you cannot even set it to `None`), because the default is still the empty string. If you don't want to do `text: root.my_text or ''` why set `my_text` to default to `None` in the first place? Just leave it empty or set it to the empty string. In Kivy itself, I don't think we ever set it to None explicitly.

Also, I often set the default value to `None` and I'd rather not have to also set this new parameter `none_is_empty` to `False`.

kived:
We use this `x or ''` pattern in Kivy itself for `SettingString` and `SettingPath` and `SettingOptions`. Beyond that, I just feel that setting the value to `None` shouldn't break your app. With my code, you can indeed set `None` when `allownone` is `False`, it just converts it to the empty string.

And if `allownone` is `True`, this does nothing anyway. You don't need to change the `none_is_empty` parameter to allow setting an actual `None`.

Kovak:
I think this is a really nice change, and I will say x or '' comes up fairly often for me.

matham:
`SettingsString` and I assume the other have `text: root.value or ''` because `value` is a `ObjectProperty`, not a `StringProperty`. Can you give one actual example where `x or ''` is used where `x` is a `StringProperty`?

matham:
I may have missunderstood where the None is coming from. I thought it was meant to fix a `None` from when the prop was initialized. But I guess @kived wanted to use this in general situations when it is assigned a None value.

I think in those situations we should not be converting None to a empty string. Because a) it's unexpected - you'd expect an error if allownone is respected, and b) you can use the error handler to just convert it to the empty string. It also feels a bit hacky to me adding such a param just to solve a very specific problem. Also, why only StringProperty? Why not also convert None of a bool prop to False. Simialrly for lists etc.

dessant:
So, closing? Rage merging?

I also think it is a step we should not take (in this form), we could easily find justification for many other cases from there.

dessant:
It seems this is a tie, please reopen if there are new views on the issue.

PR #4203, opened 2016-04-20
Add buttons for nightly wheels

KeyWeeUsr:
Based on [kivy-website PR#67](https://github.com/kivy/kivy-website/pull/67) this will add four buttons, each for different nightly wheel for easier access than searching in google drive folder. Click on the button will open a "popup" as accessing actual file through google url and changing the name right on the website doesn't seem efficient to me, because there is already mentioned renaming wheel and how to use it.


dessant:
@KeyWeeUsr, if it's not too much hassle, would you turn it into an unordered list of links and squash the commits (so the images are not preserved in the git history)? In retrospect, I should have said links on irc instead of buttons, sorry for that. :yum:

KeyWeeUsr:
I wasn't sure if I should join two commits when you said images shouldn't be preserved in history, so I rather reverted last commit and force pushed. Shouldn't do any damage if it was done on my branch I think and the result is clear commit. ^^

dessant:
@KeyWeeUsr, looks delightful, thanks a lot for your work. :tropical_drink:

PR #4204, opened 2016-04-21
TextInput: hint_text decode text by default

### LINKED ISSUE #4201: unicode text_hint in TextInput
using a unicode value for `hint_text` in kv (loaded with load_file) fails, i get encoding issues.

```
TextInput:
    hint_text: u'é'
```

a successful workaround is to use `decode` instead of `u''`

```
TextInput:
   hint_text: 'é'.decode('utf-8')
```

edit, this is not specific to `hint_text`, `text` has the same issue.

akshayaurora:
closes #4201


dessant:
`unicode` does not exist in python 3.

PR #4217, opened 2016-04-24
[Fix] x11 window provider: Window resizing. The new size never got pa…

jegger:
…ssed to the base class.

I guess this feature was never working?


tshirtman:
Confirmed fix, thanks!

PR #4219, opened 2016-04-24
Added nogil to _WindowSDL2Storage.flip()

ignus2:
Add nogil when calling SDL_GL_SwapWindow().
The reason being SDL_GL_SwapWindow() can block when vsync is turned on, effectively blocking the whole python program and all threads because of the GIL.

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -397,7 +397,9 @@ cdef class _WindowSDL2Storage:
             pass
 
     def flip(self):
-        SDL_GL_SwapWindow(self.win)
+        win = self.win
+        with nogil:
+            SDL_GL_SwapWindow(win)
 
     def save_bytes_in_png(self, filename, data, int width, int height):
 
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -7,6 +7,7 @@ from kivy.config import Config
 from kivy.logger import Logger
 from kivy import platform
 from kivy.graphics.cgl import cgl_get_backend_name
+from kivy.graphics.cgl cimport *
 
 from cpython.mem cimport PyMem_Malloc, PyMem_Realloc, PyMem_Free
 
@@ -400,6 +401,7 @@ cdef class _WindowSDL2Storage:
         win = self.win
         with nogil:
             SDL_GL_SwapWindow(win)
+            cgl.glFinish()
 
     def save_bytes_in_png(self, filename, data, int width, int height):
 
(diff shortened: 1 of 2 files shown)


matham:
Seems fine (without testing). The only thing is I'd type the `win`variable i.e. `cdef SDL_Window win = self.win`.

ignus2:
Along the same lines I also added glFinish() after the SwapWindow() call in my version.
The reason is that there is no guarantee that the swap will actually happen when SwapWindow() is called, it can be deferred by the driver. In my app I observed my app also got completely blocked (ie. other threads in the app) despite the SwapWindow() being in nogil, and the reason was that the actual swap and waiting for vsync happened sometime later on in some glXXX call outside of nogil, hence blocking the main thread and the whole app as a result (because of the GIL).
Putting glFinish() after a swap is also no guarantee, but it fixed it for me (the swap happened there, in nogil). I can make a pull request for that as well.

matham:
Seems alright. @Kovak ?

ignus2:
Hmmm..., so the glFinish() thing fails on OSX. Can someone take a look?

matham:
I tested this and on my laptop and it didn't seem to make much of a difference. I.e. it didn't seem to lock it to the refresh rate of the monitor. With nvidia graphics card I got 60fps whether with master and this code, and with intel integrated graphics I got no framerate limiting either way. So it seems this makes no difference.

ignus2:
Do you refer to glFinish being added or the original nogil patch?
nogil is needed so other threads don't get blocked because of the GIL while the swap call is waiting on vsync.
glFinish was needed because swapwindow might not wait for vsync, but at a later time in another glXXX call, and glFinish, though not required by drivers, actually do flush swap calls.
This patch makes no difference if your program is single threaded.

matham:
I'm referring to glFinish. The nogil stuff is needed. glFinish didn't seem
to make any difference for me as that didn't force the kivy clock to run at
the refresh rate, which I assume is the point of it.

ignus2:
glFinish is an attempt to make the swap actually happen right there. Drivers are not required to flush a swap at glFinish, as swap is not part of gl but rather wgl,glx,cgl etc. ie. the windowing interface. In my experience though, they do flush the swap.
You say nogil is required. However, the swap function call is not required to actually wait for the swap to happen, it might return immediately, making nogil useless, as the very next glXXX call will wait/block for the swap to happen, which is not in nogil, so the python app is blocked.
I'm writing a fairly large app based on kivy with multiple threads and video playback, and have traced back video decoding jitters to the GIL, and through profiling I saw a big portion of time was spent inside some random function performing opengl calls. As I put glFinish right after the swap, the block and the jitter was gone, and profiling confirmed that glFinish did force the swap to happen, in nogil.

tito:
Tested on OSX, works.

Whatnoww:
Afaik reverting this creates other issues in Android. At least on first look, it causes scrollviews to always run at 30fps when being dragged through.

PR #4223, opened 2016-04-25
should fix Issue 4222

tshirtman:
just wrote a test to reproduce, and a code to fix the tests, not a lot of manual testing, but seems OK :)

### DIFF
--- a/kivy/animation.py
+++ b/kivy/animation.py
@@ -426,6 +426,13 @@ class Sequence(Animation):
         self.anim2.cancel(widget)
         super(Sequence, self).cancel(widget)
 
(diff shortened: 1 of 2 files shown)


kivy/animation.py
```diff
@@ -426,6 +426,13 @@ def cancel(self, widget):
         self.anim2.cancel(widget)
         super(Sequence, self).cancel(widget)
 
+    def cancel_property(self, widget, prop):
```
akshayaurora: We should add documentation for this.

akshayaurora:
👍 merge after adding docs.

PR #4234, opened 2016-04-27
Add Crash Course to docs

KeyWeeUsr:
This tutorial helped me a lot in the beginning and is basically a set of videos, therefore easier to go through than text and code for beginners. When on website, it should clarify some stuff more, such as properties or bindings.

### DIFF
--- a/doc/sources/tutorials/crashcourse.rst
+++ b/doc/sources/tutorials/crashcourse.rst
@@ -1,0 +1,47 @@
+.. _crashcourse:
+
+Crash Course
+============
+
+Crash Course (or Kivy Crash Course) is a series of YouTube video tutorials by
+Kivy core developer `inclement <https://github.com/inclement>`_ which provides
+a simple walkthrough in Kivy for users who know how to code in Python, but is
+also friendly even to beginners with Python as with Kivy. After Pong and Paint
+tutorials, this set of videos covers more features and tweaks that can be used
+to create your app quicker, your code elegant and eye-friendly.
+
+
+Basic Info
+----------
+
+Crash Course primarily consists of a series of YouTube videos each ~10 minutes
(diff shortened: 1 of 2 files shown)


doc/sources/tutorials/crashcourse.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Crash Course(or Kivy Crash Course) is a set of 14 video tutorials by Kivy core
+developer `inclement <https://github.com/inclement>`_ which provides a simple
+walkthrough in Kivy for users who know how to code in Python, but is also
+friendly even to beginners with Python as with Kivy. After Pong and Paint
+tutorials, this set of videos covers more features and tweaks that can be used
+to create your app quicker, your code elegant and eye-friendly.
+
+
+Basic Info
```
dessant: Could `Basic Info`, `What the Crash Course covers` and `Let's start` be removed? The course is an external resource and we do not track if the content and number of videos and articles change, so these sections will just become outdated and incorrect.

The playlist link could be simply added to the `Links` section.

kived: I would agree about `Let's start`, but I'd say the other sections should stay. They should just be less specific - i.e. `Crash Course primarily consists of a series of YouTube videos.` (And YouTube should be written properly.) `Topics covered by the Crash Course include: ...`

dessant: That's actually a better approach than ripping them out :yum:, +1 for having a more general description.

KeyWeeUsr:
I edited it a little bit, removed the numbers, ... Wasn't sure about the series length(minutes), so I let it be. Should I remove that, or something else?

Zen-CODE:
Thanks :-)

Issue #4232, opened 2016-04-27
ScrollView scroll_distance and scroll_wheel_distance values are further converted with sp


dessant:
This will lead to incorrect user supplied values, there should be no conversion happening.

Issue #2480, opened 2014-09-20
Cannot test kivy.utils.platform without importing the rest of kivy


winstonwolff:
There are occasions when one needs to check your platform before loading kivy, such as setting some config environment variables only on iOS. Trying this in v1.8.0, importing kivy.utils uses the @deprecated decorator which imports kivy.logger and the rest of kivy.

Would it be possible to break that connection? e.g. not use @deprecated so it just crashes, but at least allows you check your platform before loading kivy?
##

FeralBytes:
Also Affected by this issue. Unable to setup my own logging facilities because Kivy beets me too it.

FeralBytes:
Work around is to do this:
```python
logging.basicConfig(filemode='w', level=log_level)
```
Before you import anything Kivy. You have to beat Kivy or else Kivy will crash.
Then when you are ready to set your logging to the file that you want, do this:
```python
logging.root.handlers = []  # This removes the root handler and allows you to run basicConfig again.
logging.basicConfig(filename=LOG_FILE_PATH, filemode='w', level=log_level)
```

KeyWeeUsr:
`Logger` is already imported in the main `__init__.py` file:
https://github.com/kivy/kivy/blob/fed61946603baf381404ae74009f742ab277d848/kivy/__init__.py#L38

therefore to make this happen we would need to make the logger completely separated from the framework. Perhaps using `KIVY_NO_FILELOG` and/or `KIVY_NO_CONSOLELOG` would do it for this particular case, but I think we might introduce a new, mocked `Logger` object if e.g. `KIVY_NO_LOGGER` is specified, but that would mean no logging from Kivy side.

If that's not the case i.e. it's about removing `Logger` from `__init__.py` and still keep the Kivy logging in the rest, then I'd mark it as `wontfix` and close.

winstonwolff:
The original issue was not about logging, it's about decoupling kivy.utils.platform from the rest of Kivy because when writing platform-specific code, one might need to configure Kivy before importing it. E.g. on Linux I might want to choose one audio driver, and on iOS a different audio driver. Kivy allows me to do this by setting and environment var before I load Kivy. So here would be my steps:

(1) detect platform
(2) set environment var to choose audio driver, depending on platform
(3) import Kivy

The problem is that step (1) would ideally be done using kivy.utils.platform, but as soon as I import kivy.utils.platform, I'm importing the rest of Kivy too so I can no longer do (2)

Julian-O:
[...]

PR #4235, opened 2016-04-27
ScrollView: make sure import does not load a window.

akshayaurora:


dessant:
@akshayaurora, looks great. :dancer: 
It also fixes https://github.com/kivy/kivy/issues/4232.
I'll open a separate issue for the config issue we discussed.

kived:
Seems like it would be better to fix the root problem - that importing Window creates a window.

dessant:
There must be an issue opened about agressive kivy imports.

dessant:
This was also one of them: https://github.com/kivy/kivy/issues/2480
But i also remember a general one.

PR #4236, opened 2016-04-28
fix stacklayout not sizing if children is empty

Kovak:
Right now if we go from 1 element to 0 element, the bail out on calculation prevents us from properly calculating the size of the stacklayout. If there are no children we should set minimum_size to 0, 0, otherwise it will always be the size of the last element.


akshayaurora:
@Kovak 👍

Issue #3554, opened 2015-08-04
Doc needs MORE Screenshots


akshayaurora:
For this project to gain traction, there need to be screenshots for every page that has an example application. Especially the ones that us the .kv files.

References #710  #3374
##

picibucor:
check #3374 as well.

Zen-CODE:
Perhaps it's that I'm not a visually driven person, but this does not seem so important to me.

There are lots of screenshots in the guides, and in other places I feel enough for a person to be pretty fluid in Kivy by the time they need them? How do others feel?

If it's felt important, I'm certainly willing to try a sweep through adding these, I think initially just for the kv examples as I agree these are more important and kivy specific.

dessant:
We are closing some of the stale issues, please create a new pr if you'd like to work on this feature, or get in touch with us on #kivy-dev.

PR #4237, opened 2016-04-28
Add screenshots for widgets

KeyWeeUsr:
This PR adds screenshots for dynamic widgets i.e. the widgets provide a specific movement and pure `.png`/`.jpg` isn't enough. PR also completes screenshots for static widgets that may be neccessary to see(design):

static:
- label(markup)
- treeview(how it looks)

dynamic:
- carousel
- dropdown
- scatter
- pagelayout
- stencilview
- screenmanager(all transitions ordered as in docs)

ref #3554 ?


dessant:
Nice job @KeyWeeUsr, I like that the images have a compact size and do not contain unnecessary elements.

dessant:
@KeyWeeUsr, the images do not appear on the built doc pages.

KeyWeeUsr:
image
 I see them there O.o

dessant:
Pff, this is odd, I was browsing the same page and this is what i've seen, but now it works.
capture

dessant:
We should also ensure that the image is contained in the parent div and does not cover other elements.

KeyWeeUsr:
I was thinking exactly the same when I saw [ColorPicker](https://kivy.org/docs/api-kivy.uix.colorpicker.html) mainly because the "code" div keeps whole width of a parent and images are placed right above that div. If I make the images smaller or something it won't fix the issue. The image is already in the parent div, the problem is that the other elements are too. I'm not really sure with this, but I can't place images into paragraphs, can I?
image
The necessary change should maybe target the default template for docs i.e. there should be a place for an image and hopefully image of consistent shape if it's not something like ActionBar.

PR #4238, opened 2016-04-28
Xinput2 input provider

jegger:
This is an xinput2 input provider for kivy. This allows the use of multiple "native" x11 inputs (mouses) as they are available on some mutltiouch screens on Linux. You are not limited to fullscreen when you have more than one (mouse) input.

This currently only works in combination with the x11-window as I need the window-id.
See: https://groups.google.com/forum/#!topic/kivy-dev/wlrRfIBi-2c
Therefore I needed to slightly alter the x11-window.

I am open for any feedback, or just happy if you could merge as it is.

### DIFF
--- a/kivy/core/window/window_x11.pyx
+++ b/kivy/core/window/window_x11.pyx
@@ -192,11 +192,11 @@ class WindowX11(WindowBase):
         if 'KIVY_WINDOW_X11_CWOR' in environ:
             CWOR = True
 
-        if x11_create_window(size[0], size[1], pos[0], pos[1],
-                resizable, fullscreen, border, above, CWOR,
-                <char *><bytes>self.title) < 0:
+        window_id = x11_create_window(size[0], size[1], pos[0], pos[1], resizable, fullscreen, border, above, CWOR, <char *><bytes>self.title)
+        if window_id < 0:
             Logger.critical('WinX11: Unable to create the window')
             return
+        self.window_id = window_id
 
         size[0] = x11_get_width()
         size[1] = x11_get_height()
--- a/kivy/input/providers/__init__.py
+++ b/kivy/input/providers/__init__.py
@@ -51,6 +51,11 @@ if platform == 'linux' or 'KIVY_DOC' in os.environ:
     except:
         err = 'Input: LinuxWacom is not supported by your version of linux'
         Logger.exception(err)
+    try:
+        import kivy.input.providers.xinput2
+    except:
+        err = 'Input: Xinput2 is not supported by your version of linux'
+        Logger.exception(err)
 
 if platform == 'android' or 'KIVY_DOC' in os.environ:
     try:
--- a/kivy/input/providers/__init__.py
+++ b/kivy/input/providers/__init__.py
@@ -53,9 +53,11 @@ if platform == 'linux' or 'KIVY_DOC' in os.environ:
         Logger.exception(err)
     try:
         import kivy.input.providers.xinput2
+    except ImportError:
+        Logger.debug('Input: Did not import xinput2 '
+                     'If you want it you have to compile kivy with USE_X11=1')
     except:
-        err = 'Input: Xinput2 is not supported by your version of linux'
-        Logger.exception(err)
+        Logger.exception('Input: ')
 
 if platform == 'android' or 'KIVY_DOC' in os.environ:
     try:
--- a/kivy/input/providers/xinput2.pyx
+++ b/kivy/input/providers/xinput2.pyx
@@ -19,7 +19,6 @@ class InputX11():
 
     def start(self, window_id, on_touch):
         self.on_touch = on_touch
-        print "forward window id", window_id
         init(window_id)
 
     def x11_idle(self):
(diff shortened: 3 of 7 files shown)


kivy/input/providers/__init__.py
```diff
@@ -51,6 +51,13 @@
     except:
         err = 'Input: LinuxWacom is not supported by your version of linux'
         Logger.exception(err)
+    try:
+        import kivy.input.providers.xinput2
+    except ImportError:
+        Logger.debug('Input: Did not import xinput2 '
+                     'If you want it you have to compile kivy with USE_X11=1')
+    except:
+        Logger.exception('Input: ')
```
tshirtman: More descriptive logging maybe? :)

tshirtman:
Need to test, but that does look correct :)

kivy/input/providers/__init__.py
jegger: Yes, indeed - this does not make much sense.

KeyWeeUsr:
@tshirtman could you take a look again? :)

Zen-CODE:
@jegger Would be a pity to throw this away. Please could you fix the merge conflicts so we can test and merge? Thanks

matham (approved):
Seems ok. If no one else wants to test, we should just merge soon.

kivy/input/providers/__init__.py
```diff
@@ -52,6 +52,13 @@
     except:
         err = 'Input: LinuxWacom is not supported by your version of linux'
         Logger.exception(err)
+    try:
+        import kivy.input.providers.xinput2
+    except ImportError:
+        Logger.debug('Input: Did not import xinput2 '
```
pythonic64, line 58: Missing `.` after "... after xinput2". Should be "Input: Did not import xinput2. "

kivy/input/providers/xinput2.pyx
```diff
@@ -0,0 +1,124 @@
...
+
+
+class Xinput2Event(MotionEvent):
+
+    def depack(self, args):
+        if args[0] is None:
+            return
+        self.profile = ('pos',)
+        width, height = EventLoop.window.system_size
+        rx = args[0] / float(width)
```
pythonic64, line 60: Probably should use `width - 1` and `height - 1` when converting to relative coords. Same as in https://github.com/kivy/kivy/blob/master/kivy/input/providers/mouse.py#L275 . See https://github.com/kivy/kivy/pull/7425.

kivy/core/window/window_x11.pyx
```diff
@@ -210,6 +210,7 @@ class WindowX11(WindowBase):
                 <char *><bytes>title) < 0:
             Logger.critical('WinX11: Unable to create the window')
             return
+        self.window_id = window_id
```
pythonic64, line 213: Should we declare `window_id` to `WindowBase` as property or attribute? Or declare it in `__init__` of `WindowX11`?
We should have a doc for this public attribute.

kivy/input/providers/xinput2.pyx
matham: I wonder if we should add a `normalize_pos` method to window that does this that can just be re-used everywhere, because otherwise it'd inevitably be forgotten in new providers.

kivy/core/window/window_x11.pyx
matham: Well, it could be a variable just for x11. But I think correspondingly, we should test more robustly whether the provider is x11 (currently it's checking if window_id  is defined, which is not robust).

kivy/input/providers/xinput2.pyx
pythonic64: I agree for `normalize_pos` method.

kivy/core/window/window_x11.pyx
pythonic64: Yes, that's what I meant. It could be a property or attribute just for `WindowX11`, but with property we can add the doc.

kivy/input/providers/xinput2_core.c
```diff
@@ -0,0 +1,168 @@
...
+	{
+		int rc;
+		int major = 2;
+		int minor = 3;
+
+		rc = XIQueryVersion(dpy, &major, &minor);
+		if (rc != Success)
+		{
+			printf("No XI2 support. (%d.%d only)\n", major, minor);
+			exit(1);
```
pythonic64, line 66: Doesn't seem safe to call `exit(1)`. Wouldn't that exit the running process? 
Can we propagate errors so that they can be logged using Kivy `Logger`?

jegger:
@Zen-CODE I am sorry for not responding at all. Somehow I did not got any notifications when you all did the code-review. 

I would really like to invest the time into getting the comments resolved - as I profited a lot from Kivy in the past :-)! Thank you all again, and I love to see that this project is still maintained 👍  

But my main problem is that I do not have any environment ready to test the provider (touch hardware) and have not even worked with Kivy within the last 3 years. I am lacking the knowledge completely. In example, It would cost me quite some time to figure out how to properly handle this one: https://github.com/kivy/kivy/pull/4238#discussion_r613562703  And I can honestly tell you that I will not find the time to read me again into it right now. I am sorry for that. Not sure how you want to proceed?

PR #4239, opened 2016-04-28
x11-window: honor borderless configuration

jegger:
There is the kivy graphics config-option "borderless". The x11-window did not honor this configuration. Now it does.


akshayaurora:
@jegger thanks 👍

PR #4245, opened 2016-04-29
Fix docs content's height

### LINKED ISSUE #4167: Docs content height when the descriptions are hidden
I think website documentation behaves kind of weird when it's changed from "Show Descriptions" to "Hide Descriptions" option. All items are folded, but the `content` div element is still the same height. Why it's weird? Although I know it's only cosmetic issue, I expect `td.leftlink` and `td.rightlink` to behave as navigation at the bottom as it behaves normally, but when `min-height` is fixed at the beginning, a user will end up with scrolling manually to find the bottom navigation until the page refreshes manually.

What causes it: `div#content.min-height` is set to a height of fully unfolded website, but when the function that folds the items is called, it doesn't recalculate the `min-height`. When the website is visited after a refresh and it remembers that a user checked "Hide Descriptions", but it uses directly the `min-height` of unfolded items. Try [Window](https://kivy.org/docs/api-kivy.core.window.html) for example, hide descriptions and press `End` on keyboard/scroll to the bottom.

KeyWeeUsr:
This PR ensures that the height of an div#content is recalculated properly, so that the bottom navigation is really at the bottom of a page.

PR affects only `min-height`, therefore even if `min-height==0`, the page will have a proper size and the left content navigation won't break.

Fix #4167.


KeyWeeUsr:
This seems to be a pretty old thing, because even [here](https://web.archive.org/web/20121015184637/http://kivy.org/docs/api-kivy.uix.slider.html) it doesn't behave as it really should.

dessant:
I don't understand which bug is being solved, could you describe what are you seeing?

dessant:
@KeyWeeUsr, ok, i've missed the issue link. Isn't pressing `End` supposed to always take you to the bottom of the page?

dessant:
@KeyWeeUsr, haha, and *now* I also understand it. :grin: 
+1

KeyWeeUsr:
Hehe, I was trying to write it again, but you found it :+1:

KeyWeeUsr:
Not that much obvious on small files, but when the class has a lot of properties, it'll leave a big hole under the folded items.

dessant:
@KeyWeeUsr, were you able to test the change locally? I'm not able to duild docs here.

KeyWeeUsr:
Only with chrome's dev-tools. get to `div#content` and change `min-height` to something smaller, or even delete it and you'll see the change. JS is tested also with the dev-tools, but works. Don't know if `min-height` is really required by something, so I rather leave it there and just edit it the way that `min-height` gets updated.
Building docs on windows is still bugged, right?

KeyWeeUsr:
Also, more important than this is that the bottom navigation seems broken, which I mentioned [here](https://github.com/kivy/kivy/issues/4246). :-/

dessant:
Ok, dev-tools should be enough. Yeah, it still doesn't build on win.

Nice job, let's merge this. Please check out the other screenshot pr, the pics are not visible on the site.

PR #4254, opened 2016-05-03
kivy.utils.platform is a string and it is not callable anymore

dessant:


dessant:
Should be listed here after merging: https://github.com/kivy/kivy/wiki/Breaking-changes-in-Kivy-2.0.

matham:
Why?

dessant:
@matham, why not? :dancer: 
It was deprecated 3 years ago.

matham:
Ok.

PR #4257, opened 2016-05-04
Fix bottom navigation

### LINKED ISSUE #4246: Docs - Bottom navigation doesn't work properly
Bottom navigation sends you on the previous and on the next article, but not always.

- https://kivy.org/docs/api-kivy.uix.html (next)
- https://kivy.org/docs/api-kivy.uix.abstractview.html (previous)
- https://kivy.org/docs/api-kivy.uix.anchorlayout.html (this rather skips whole `behaviors` folder though it's in the left navigation)
- https://kivy.org/docs/api-kivy.uix.widget.html (and this pretty much doesn't give a damn at all)

How is the bottom navigation list(?) created and how can I fix it? :hammer:

KeyWeeUsr:
Fix #4246

### DIFF
--- a/doc/autobuild.py
+++ b/doc/autobuild.py
@@ -216,7 +216,7 @@ for package in packages:
 
     # search modules
     m = list(modules.keys())
-    m.sort(key=lambda x: extract_summary_line(sys.modules[x].__doc__))
+    m.sort(key=lambda x: extract_summary_line(sys.modules[x].__doc__).upper())
     for module in m:
         packagemodule = module.rsplit('.', 1)[0]
         if packagemodule != package:


KeyWeeUsr:
I have built the docs on my pc and there are two options:
1) vkeyboard.py after .py->.html, widget.py contains on the bottom nav only "vkeyboard" on the previous link side
image
image
All will be lowercase at the bottom navigation(which is this PR by default) and it will not happen that something will skip the alphabet such as "VKeyboard" "reStruc..." i.e. not in format `<one capital><the rest lowercase>.

2) There will be some rules for contributing about how to write first line of `__doc__` in files so that it won't break the navigation i.e. one capital letter only and only at the beginning of a word.

KeyWeeUsr:
Found a proper way for sorting now without any other unnecessary editing needed. This is bottom nav from [Widget](https://kivy.org/docs/api-kivy.uix.widget.html) that works properly according to the left sidebar navigation. Rst and pretty much everything now works as it should.
image

PR #4283, opened 2016-05-15
Fix syntax error in `_local_func()` example.

cpburnz:
Fixed a syntax error in a `_local_func()` function example: `pos=.pos` should be `pos=pos`.


tshirtman:
Thanks!

PR #4286, opened 2016-05-16
adding get_current_display_mode method

kiok46:
disp

You may use this code to check.

    from kivy.app import App
    from kivy.lang import Builder
    from kivy.properties import ListProperty

    Builder.load_string('''
    <Interface>:

        Label:
            text: str(root.disp)

    ''')

    class Interface(BoxLayout):
        disp = ListProperty()

        def __init__(self, **kwargs):
            super(Interface, self).__init__(**kwargs)
            self.display()

        def display(self):
            self.disp = Window.get_current_display_mode()
            print self.disp

    class DispApp(App):

        def build(self):
            return Interface()

    if __name__ == "__main__":
        app = DispApp()
        app.run()

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -158,6 +158,11 @@ cdef class _WindowSDL2Storage:
         SDL_GL_GetDrawableSize(self.win, &w, &h)
         return w, h
 
+    def get_currentdisplaymode(self):
+        cdef SDL_DisplayMode mode
+        SDL_GetCurrentDisplayMode(0, &mode)
+        return mode.w, mode.h
+
     def resize_display_mode(self, w, h):
         cdef SDL_DisplayMode mode
         cdef int draw_w, draw_h
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -700,3 +700,6 @@ class WindowSDL(WindowBase):
     def unmap_key(self, key):
         if key in self.key_map:
             del self.key_map[key]
+
+    def get_current_display_mode(self):
+        return self._win.get_currentdisplaymode()


KeyWeeUsr:
I'm not really sure what does it really do except returning Window.size, which is already implemented [here](https://github.com/kiok46/kivy/blob/f143512ce3de02f698d72b78d193f56b0d1abe68/kivy/core/window/_window_sdl2.pyx#L156). Maybe you wanted to _set_ display mode? Or do you want to return total size of a screen e.g. 1366x768?

kiok46:
@KeyWeeUsr  The reason it returns **1366x663** instead of **1366x768** is because the window was a VirtualBox (running linux) also it is different from Window.size which returns the size of "kivy" window.

dessant:
It returns the current screen resolution, right? In that case i'm not entirely sure it should be added there. Plyer might be a better fit for it, as part of a device info facade, if such a thing gets implemented.

kiok46:
@dessant,   @kived  and I had a little discussion about it.
https://botbot.me/freenode/kivy-dev/2016-05-06/?msg=65603175&page=1
https://botbot.me/freenode/kivy-dev/2016-05-06/?msg=65614415&page=2

KeyWeeUsr:
@kiok46 Sorry, my mistake. Hm, maybe the idea was to use fullscreen under the full resolution directly with Kivy, so he didn't do it for Plyer? Or maybe to set starting Window/Form position on the screen? Besides these I don't see how you can use resolution in kivy only. However it'd be good to have this size there too if it'd work on each platform. ^^

dessant:
@kiok46, it's not clear wether or not @kived wants this api to be added directly to kivy, his main message was that we should not depend on kivy inside plyer.

kiok46:
@dessant  It's up to you guys to include it ot not :)
There is a facade for this in one of the PRs for plyer.
https://github.com/kivy/plyer/pull/201/files#diff-9335e4ea990b09538de388ea61f9da32R45

dessant:
I would suggest a bit different api if we add it, something like `Window.screen_size` or `Window.screen_resolution`.

When the screen resolution changes, this property should be updated.
In this form I could see the use case of being able to react not just to a window size change (through Window.size) but also to a screen resolution change.

KeyWeeUsr:
I'm not sure if kivy works with multiple monitors, but if the `Window.screen_resolution` is going to happen, this would be a nice place to check even for each monitor's resolution maybe?

dessant:
@KeyWeeUsr, yeah, the property could update to the current screen res as you drag the kivy window between monitors.

kiok46:
@dessant  sounds good, which value should it return when kivy window moves from one screen to other? I mean of the first or second screen or it could be something like {'sc1': [100: 768], 's2': [200, 768]}

dessant:
@kiok46, always just the resolution of the current screen, I guess sdl2 decides between one of the displays even if the window is between the two, but it's just an assumption, it should be looked into.

kiok46:
@dessant  This example made me think that dictionary could be used.
https://wiki.libsdl.org/SDL_GetCurrentDisplayMode#CA-pygments_614_15

dessant:
@kiok46, the window is associated with a single display at a time based on the window center coordinates, I'm not sure we should go beyond that granularity at this time.
https://wiki.libsdl.org/SDL_GetWindowDisplayIndex

dessant:
@kiok46, the main problem is how do we get notified about a display resolution change, does sdl2 have such event?

dessant:
A pr implementing the features described in https://github.com/kivy/kivy/pull/4286#issuecomment-219574595 could be merged in kivy, but in it's current form the api will be available in plyer.

PR #4289, opened 2016-05-17
Add Matrix.transform() to easily get transformed 2D coordinate

udiboy1209:
Looks like a useful addition. Wanted to use it [here](https://github.com/udiboy1209/kivent/blob/2121e58b905cbf3bd948a88094c1bae5de0484bf/modules/core/kivent_core/systems/gameview.pyx#L128).

Test code:

```
import kivy
from kivy.graphics.transformation import Matrix

m = Matrix()
m.rotate(3.14159/4, 0, 0, 1)

print m.transform([1,1])
```

### DIFF
--- a/kivy/graphics/transformation.pyx
+++ b/kivy/graphics/transformation.pyx
@@ -611,6 +611,20 @@ cdef class Matrix:
 
         return (winx, winy, winz)
 
+    def transform(self, double x, double y):
+        '''Return a transformed point by applying this Matrix's tranformation.
+
+        :Parameters:
+            `point`: 2D point to tranform
+
+        .. versionadded: 2.0.0
+        '''
+
+        cdef double *m = <double *>self.mat
+        transformed_point = (x * m[0] + y * m[4] + m[12],
+                            x * m[1] + y * m[5] + m[13])
+        return transformed_point
+
     def __str__(self):
         cdef double *m = <double *>self.mat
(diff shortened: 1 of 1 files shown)


kivy/graphics/transformation.pyx
```diff
@@ -611,6 +611,20 @@ cdef class Matrix:
 
         return (winx, winy, winz)
 
+    def transform(self, list point):
```
matham: I'd suggest to not type it as list, because what if point is a tuple? Or event better, type it as `double x, double y`.

udiboy1209: `double` seems to be the best way to go @matham. Some of Kivent's points were stored as lists so that was the reason for this, but there could be use cases for tuples.

udiboy1209:
On second thought, I might have overlooked this [existing cpdefed method](https://github.com/udiboy1209/kivy/blob/93d13d465d7081910d6b8e83001d0e4353310366/kivy/graphics/transformation.pyx#L440). My method is just a python-accessible version of this.

There should be a docstring describing what `transform_point` actually does. Git blame shows me @tito added this years ago :stuck_out_tongue:! What do `double t` and `double tt` do?

dessant:
@udiboy1209, what will be the fate of this?

udiboy1209:
It's no use :stuck_out_tongue: . `transform_point` does the same job better. I was kinda confused with the matrix math before but now its clear.

PR #4294, opened 2016-05-20
Fix typos and links

KeyWeeUsr:
I hope it's not too small commit. If yes, just include this "y" to "v" replacement in something bigger. ^^

### DIFF
--- a/doc/gallery.py
+++ b/doc/gallery.py
@@ -35,7 +35,7 @@ def iter_filename_info(dir_name):
     """
     Yield info (dict) of each matching screenshot found walking the
     directory dir_name.  A matching screenshot uses double underscores to
-    separate fields, i.e., path__to__filename__py.png as the screenshot for
+    separate fields, i.e. path__to__filename__py.png as the screenshot for
     examples/path/to/filename.py.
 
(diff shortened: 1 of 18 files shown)


KeyWeeUsr:
I'll add fixes for other typos I found soon.

KeyWeeUsr:
Commas after `e.g.` or `i.e.` aren't common([i.e.,](http://google.com/#q=site:kivy.org+"i.e.,"), [e.g.,](http://google.com/#q=site:kivy.org+"e.g.,")) in these docs except those edited cases and according to wiki and [SO](http://english.stackexchange.com/a/16215) it's either used by preference or English standard.

About `::`, these colons disappear if there is a space before them, leaving only pretty text, so there will be no `.:` or `,:` endings in sentences.

Also there were some links broken at the end of pong tutorial to `Circle` and `source` because the property/class was moved. I'm not sure if [values](https://kivy.org/docs/api-kivy.html#kivy.kivy_options) should be really displayed this way. I left them un changed.

Zen-CODE:
Nice improvements. Good catch on v/y :-)

PR #4296, opened 2016-05-21
Add tint color to accordion

KeyWeeUsr:
Adds tinting color for accordion if changing images is either not necessary or undesirable.


dessant:
We discussed this with @matham and we felt this is the wrong path. We can keep adding these apis, but it never ends. A more flexible solution could be to have `style_cls` or something similar for all components of complex widgets like these, or maybe something else. Perhaps these are just the symptoms of kivy lacking proper theming support, but we should not add these new endpoints, but push for a new discussion about theming instead.

PR #4304, opened 2016-05-24
AliasProperty should update when underlying prop changes even if cache is True

### LINKED ISSUE #4303: AliasProperty cache doesn't dispatch updates
When setting `cache=True` on an `AliasProperty`, it does not dispatch updates. In the same situation but without `cache=True` the updates are dispatched appropriately.

```python
import kivy
kivy.require('1.9.1')

from kivy.app import App
from kivy.lang import Builder
from kivy.properties import StringProperty, AliasProperty
from kivy.clock import Clock

class TestWidget(BoxLayout):
    real_text = StringProperty('')
    
    def get_alias_text(self):
        print 'getting alias text'
        return self.real_text
    alias_text = AliasProperty(get_alias_text, bind=('real_text',))
    
    def get_cached_alias_text(self):
        print 'getting cached alias text'
        return self.real_text
    cached_alias_text = AliasProperty(get_cached_alias_text, bind=('real_text',), cache=True)

TestWidget:
    real_text: str(self.size)
    Label:
        text: root.real_text
    Label:
        text: root.alias_text
    Label:
        text: root.cached_alias_text
''')

class TestApp(App):
    def build(self):
        root.bind(cached_alias_text=self.changed)
        root.bind(alias_text=self.changed2)
        root.bind(real_text=self.changed3)
        Clock.schedule_interval(self.printvalues, 2)
        return root
    
    def printvalues(self, dt):
        print root.real_text, root.alias_text, root.cached_alias_text
    
    def changed(self, *args):
        print 'changed cached alias!'
    
    def changed2(self, *args):
        print 'changed alias!'
    
    def changed3(self, *args):
        print 'changed text!'

if __name__ == '__main__':
    TestApp().run()

```

matham:
Fixes #4303 .

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -1369,8 +1369,8 @@ cdef class AliasProperty(Property):
 
     cpdef trigger_change(self, EventDispatcher obj, value):
         cdef PropertyStorage ps = obj.__storage[self._name]
-        ps.alias_initial = 1
-        dvalue = self.get(obj)
+        ps.alias_initial = 0
+        dvalue = ps.getter(obj)
         if ps.value != dvalue:
             ps.value = dvalue
             self.dispatch(obj)


akshayaurora:
@matham :+1:

PR #4310, opened 2016-05-25
[Change] x11-window: implement on_title

jegger:
This enables to set the window title at runtime. Until now, the window just had the name "Kivy" (as Kivy is the init value of the Window.title property. With this implemented you see the title "Touchtracer" when you launch the demo/touchtracer example.


inclement:
Looks good, thanks!

PR #4314, opened 2016-05-25
Don't cache until first dispatch, otherwise it's never dispatched if read before the dispatch.

matham:


akshayaurora:
@matham thanks 👍

dessant:
One of the test cases fails on osx.
```
FAIL: test_aliasproperty_with_cache (kivy.tests.test_properties.PropertiesTestCase)

----------------------------------------------------------------------

  File "/Users/travis/build/kivy/kivy/kivy/tests/test_properties.py", line 319, in test_aliasproperty_with_cache

    self.assertEqual(observe_called, 1)

AssertionError: 2 != 1
```

matham:
It's osx...

PR #4316, opened 2016-05-26
Animate the window content based on `softinput_mode`, introducing

akshayaurora:
`keyboard_padding` and `keyboard_anim_args`


akshayaurora:
This should work as is but when used with https://github.com/kivy/kivy/pull/4316 this should animate the content instead of just moving it.

Issue #3569, opened 2015-08-10
Close camera


hymenoby:
Please; i am a new user of kivy and i find that the camera widget is a good one but i don't know how to close the camera.

when i make camera.play = False; it do stop the widget but don't close the camera. could someone help me please

tshirtman:
Seems the CameraBase doesn't provide any method to actually close the camera, since stop is actually pause (at least in pygst provider), one can stop the camera (at least the light goes down) by changing the state to STATE_NULL in pygst provider, but then the camera can't start again, i guess one needs to reinit the pipeline in this situation. Our API is a bit simplistic there.

hymenoby:
I understand; i used pygst to close the camera and it works. thank you

keithlaaks:
https://www.voxtelecom.co.za/


Disclaimer:

Vox does not accept responsibility or liability for the unauthorized use of its e-mail facility and/or the use of its e-mail facility other than for its own authorized business purposes. Save for statements and/or opinions relating to bona fide company matters, Vox denies responsibility or liability for the contents of this communication.

This email has been scanned for viruses and malware, and may have been automatically archived by Mimecast Ltd, an innovator in Software as a Service (SaaS) for business. Providing a safer and more useful place for your human generated data. Specializing in; Security, archiving and compliance.

PR #4328, opened 2016-05-28
Deleting the _device to release the camera device in opencv

schoolsplay:
I noticed that when stopping the camera by setting play=False the device isn't released. (The webcam light stays on for example)
In python-opencv I have to explicit delete the reference to the opencv device so I just added a 'del self._device' to get the same result.
Tested it and the webcam is switched off.

Regards,
Stas (stas@btp.nl)

### DIFF
--- a/kivy/core/camera/camera_opencv.py
+++ b/kivy/core/camera/camera_opencv.py
@@ -101,3 +101,5 @@ class CameraOpenCV(CameraBase):
     def stop(self):
         super(CameraOpenCV, self).stop()
         Clock.unschedule(self._update)
+        # Release the device
+        del self._device


dessant:
@schoolsplay, would this mean they need to call init_camera if they want to use the same camera instance again?

schoolsplay:
Yes as the reference to the opencv instance is removed.
So perhaps one needs a call to init in the "start" method ?

dessant:
Either that or have a separate `deinit_camera` method. @tito, what do you think?

dessant:
Related https://github.com/kivy/kivy/issues/3569.

matham:
Yes, the camera should be stopped, but then it needs to be init during start. Please re-open once you implement this.

PR #4332, opened 2016-05-29
#3253 examples bug fix

SepehrHml:
bugfix for issue #3253 ... wanna start to contribute on kivy 💃

### DIFF
--- a/examples/kv/app_logo.kv
+++ b/examples/kv/app_logo.kv
@@ -1,7 +1,7 @@
 Widget:
     Video:
         id: myvideo
-        source: '/home/tito/code/pymt/examples/apps/videoplayer/softboy.mpg'
+        source: '../widgets/softboy.mpg'
         play: mybutton.state == 'down'
         canvas:
             Color:


dessant:
The doc pdf creation fails after the merge.
https://travis-ci.org/kivy/kivy/jobs/133714453

SepehrHml:
😲 i try to find the problem as fast as i can ...

SepehrHml:
ummmm ... i saw logs ... it says : ImportError: No module named 'pygame' ... i never worked with travis but i guess something in the travis config changed that it cant find pygame ... becuase the thing i changed isnt in docs or something :/ dont you think ?

dessant:
The pygame error is normal (see the previous builds), it must be something else.

KeyWeeUsr:
The path is almost the same in [app_video.kv](https://github.com/kivy/kivy/blob/2c67a7b93ff1ce99e6382ea1ec0c849f39fa6b19/examples/kv/app_video.kv), so.. maybe it failed only because of early merge or something, so the test was skipped? Still, makes no sense why it fails [here](https://travis-ci.org/kivy/kivy/jobs/133714453). Maybe an issue with `.mpg` file?

dessant:
@SepehrHml, it was just a coincidence, I've reverted in a branch and the build still fails https://travis-ci.org/kivy/kivy/builds/133858247.

PR #4334, opened 2016-05-29
[Feature] add German keyboard-layout

jegger:
The already existing "qwertz" is not really a german keyboard layout, just some german-like one.


akshayaurora:
@jegger thanks

PR #4343, opened 2016-05-30
[WIP] ANGLE as an alternative for GL rendering on Windows

tito:
[ANGLE](https://github.com/MSOpenTech/angle) is a OpenGL ES to DirectX library, initiated by Google, and used in many project such as Chrome, Firefox, and more, mostly for their WebGL support on Desktop. I'm currently using the Microsoft ANGLE version maintained at https://github.com/MSOpenTech/angle, where they also contribute to Cocos2dx for pushing native OpenGL project in their Windows Store.

This can be achieve by asking a ES profile implementation to SDL2, and compile then copy ANGLE `libEGL.dll` and `libGLESv2.dll` to `c:\Python27\share\sdl2\bin`.
Support in kivy for ANGLE must be activated at compilation: `set USE_ANGLE=1` + `make`

The changes intruduce a "dynamic" opengl backend, that uses SDL2 `SDL_GL_GetProcAddress` to gather all the GL functions that we need for OpenGL, and expose them in "cgl" context.
All the graphics implementation now use cgl (either a module or the new context for dynamic gl) as a base for accessing GL functions. The GL definitions are extracted and shared in c_opengl_def.

ANGLE give us:
- ability to execute Kivy application with default driver (at least for Virtualbox and intel, no need to install specific vendor graphics)
- works starting Direct3d 9 support (use dxdiag to find out)
- works on older computer (intel drop opengl support older graphics card (< 2011, according to a customer) on Windows 10 while there where no issue on Windows 7)

Known issues:
- stencil doesn't work on Windows 7 on Virtual BOX, but even their samples doesn't work, while firefox and chrome that uses angle (according to http://www.browserleaks.com/webgl)
- line doesn't show on Windows 10

TODO:
- continue to separate fully the GL context in order to dynamically switch to another implementation at runtime
  - be able to use MOCK without recompilation
  - be able to use DEBUG opengl (check glGetError after every command) without recompilation
  - be able to either use DESKTOP opengl or ANGLE opengl without recompilation
- stencil fixes or alternative
- line fixes or alternative

Suggestions:
- if stencil is really an issue, there is a possibility to use scissor. It might be faster, but doesn't support stack, and have addional limitation, such as: scissor must be aligned to the windows while stencil can be rotated + position of the scissor must be in windows coordinate. They may have a possibility to make "hybrid" instructions that either use scissor or stencil depending the current matrix. On standard ui (without any scatter/rotation), only scissor would be used.


kivy/graphics/context.pyx
```diff
@@ -25,11 +25,17 @@ from kivy.graphics.texture cimport Texture, TextureRegion
...
+from kivy.graphics.c_opengl_def cimport *
+IF USE_OPENGL_DEBUG:
+    cimport kivy.graphics.c_opengl_debug as cgl
+ELIF USE_OPENGL_DYNAMIC:
+    from kivy.graphics.c_opengl_dynamic cimport cgl
+ELIF USE_OPENGL_MOCK:
+    cimport kivy.graphics.c_opengl_mock as cgl
+ELSE:
+    cimport kivy.graphics.c_opengl as cgl
+
```
matham: I remember I was having issues here when doing mock although I don't remember precisely what. So you should probably double check compiling and using with mock and normal.

tito: This is going to be away, as my goal is to be able to switch to another GL provider at runtime, not at compilation.

tito: Somethine like KIVY_GL_BACKEND, set to either ANGLE, DYNAMIC, MOCK, DYNAMIC_DEBUG, STATIC, STATIC_DEBUG, etc

dessant:
Closing in favor of https://github.com/kivy/kivy/pull/4385.

PR #4361, opened 2016-06-04
Fix multiprocessing freeze support

wsmith-nwfsc:
Kivy and `multiprocessing.freeze_support()` wasn't working correctly on Windows when launching multiple processes from an app packaged with PyInstaller or cx_Freeze.
The  `--multiprocessing-fork` `sys.argv` flag was not being restored in the correct order. It is, surprisingly, hard coded as the first argument in `multiprocessing`'s implementation of this feature.

The patch fixes the check for the `--multiprocessing-fork` argument, as well as restoring it in the correct position in `sys.argv`. Tested and verified on windows 7 with an app frozen with pyinstaller.


kivy/__init__.py
```diff
@@ -331,14 +331,19 @@ def kivy_usage():
 
         mp_fork = None
         try:
-            mp_fork = opts['multiprocessing-fork']
-        except:
+            for opt, arg in opts:
+                if opt in '--multiprocessing-fork':
```
inclement: This line doesn't look right - 'for arg in string' will check each character of the string. Should `in` just be `==`?

wsmith-nwfsc: You are right--I'd mindlessly used the pattern from below. I'll update the PR.

wsmith-nwfsc: I've committed this update (I did it with --rebase so there are now 22 commits in this PR, not sure if this is your preferred method. If not, I can just create another PR)

inclement:
It looks like this branch now has merge conflicts, and in any case pulls in far more commits than it should (perhaps from a bad rebase). Are you able to fix it up ready for merging again?

wsmith-nwfsc:
Indeed, I've made a mess of it. I'll try to wrangle it back down to a single commit.

wsmith-nwfsc:
I went ahead and made a new, much cleaner pull request ( #4427 ) Thanks for being patient with that!

PR #4362, opened 2016-06-04
Fix notes example

### LINKED ISSUE #3697: Notes tutorial crashes when adding a new note
https://github.com/kivy/kivy/blob/master/examples/tutorials/notes/final/main.py
Related to https://github.com/kivy/kivy/pull/3469.
```python
   File "E:\dev\envs\py34_x86\kivy\examples\tutorials\notes\final\main.py", line 158, in <module>
     NoteApp().run()
   File "e:\dev\envs\py34_x86\kivy\kivy\app.py", line 828, in run
     runTouchApp()
   File "e:\dev\envs\py34_x86\kivy\kivy\base.py", line 487, in runTouchApp
     EventLoop.window.mainloop()
   File "e:\dev\envs\py34_x86\kivy\kivy\core\window\window_sdl2.py", line 606, in mainloop
     self._mainloop()
   File "e:\dev\envs\py34_x86\kivy\kivy\core\window\window_sdl2.py", line 361, in _mainloop
     EventLoop.idle()
   File "e:\dev\envs\py34_x86\kivy\kivy\base.py", line 336, in idle
     Clock.tick_draw()
   File "e:\dev\envs\py34_x86\kivy\kivy\clock.py", line 490, in tick_draw
     self._process_events_before_frame()
   File "e:\dev\envs\py34_x86\kivy\kivy\clock.py", line 640, in _process_events_before_frame
     event.tick(self._last_tick, remove)
   File "e:\dev\envs\py34_x86\kivy\kivy\clock.py", line 374, in tick
     ret = callback(self._dt)
   File "e:\dev\envs\py34_x86\kivy\kivy\uix\listview.py", line 950, in _spopulate
     self.populate()
   File "e:\dev\envs\py34_x86\kivy\kivy\uix\listview.py", line 998, in populate
     item_view = self.adapter.get_view(index)
   File "e:\dev\envs\py34_x86\kivy\kivy\adapters\listadapter.py", line 211, in get_view
     item_view = self.create_view(index)
   File "e:\dev\envs\py34_x86\kivy\kivy\adapters\listadapter.py", line 234, in create_view
     view_instance = cls(**item_args)
   File "e:\dev\envs\py34_x86\kivy\kivy\uix\boxlayout.py", line 102, in __init__
     super(BoxLayout, self).__init__(**kwargs)
   File "e:\dev\envs\py34_x86\kivy\kivy\uix\layout.py", line 72, in __init__
     super(Layout, self).__init__(**kwargs)
   File "e:\dev\envs\py34_x86\kivy\kivy\uix\widget.py", line 312, in __init__
     super(Widget, self).__init__(**kwargs)
   File "kivy\_event.pyx", line 253, in kivy._event.EventDispatcher.__init__ (kivy\_event.c:4863)
 TypeError: object.__init__() takes no parameters
```

KeyWeeUsr:
Partially fixes #3697 - the part about `json.load()` in Python 3 where you can use only `str` and not bytes. Then there is an issue with `return True` on double-click, which if is present will prevent you from staying in focused MutableTextInput.

However I can't find anything I could catch to fix the whole issue as at [line 254](https://github.com/kivy/kivy/blob/master/kivy/_event.pyx#L254) there is `**kwargs` which should take all parameters into dictionary, am I right? Please help me?

Python 2 works as expected.

### DIFF
--- a/examples/tutorials/notes/final/main.py
+++ b/examples/tutorials/notes/final/main.py
@@ -62,6 +62,11 @@ class NoteView(Screen):
 
 class NoteListItem(BoxLayout):
 
+    def __init__(self, **kwargs):
+        print(kwargs)
+        del kwargs['index']
+        super(NoteListItem, self).__init__(**kwargs)
+    note_content = StringProperty()
     note_title = StringProperty()
     note_index = NumericProperty()
 
@@ -91,15 +96,12 @@ class NoteApp(App):
     def load_notes(self):
         if not exists(self.notes_fn):
             return
-        with open(self.notes_fn, 'rb') as fd:
-            # ensure it's str not bytes for json.load()
-            from codecs import getreader
-            reader = getreader('utf-8')
-            data = json.load(reader(fd))
+        with open(self.notes_fn) as fd:
+            data = json.load(fd)
         self.notes.data = data
 
     def save_notes(self):
-        with open(self.notes_fn, 'wb') as fd:
+        with open(self.notes_fn, 'w') as fd:
             json.dump(self.notes.data, fd)
 
     def del_note(self, note_index):
--- a/examples/tutorials/notes/final/main.py
+++ b/examples/tutorials/notes/final/main.py
@@ -36,7 +36,6 @@ class MutableTextInput(FloatLayout):
     def on_touch_down(self, touch):
         if self.collide_point(*touch.pos) and touch.is_double_tap:
             self.edit()
-            return True
         return super(MutableTextInput, self).on_touch_down(touch)
 
     def edit(self):
@@ -93,7 +92,10 @@ class NoteApp(App):
         if not exists(self.notes_fn):
             return
         with open(self.notes_fn, 'rb') as fd:
-            data = json.load(fd)
+            # ensure it's str not bytes for json.load()
+            from codecs import getreader
+            reader = getreader('utf-8')
+            data = json.load(reader(fd))
         self.notes.data = data
 
     def save_notes(self):


examples/tutorials/notes/final/main.py
```diff
@@ -93,7 +92,10 @@ def load_notes(self):
         if not exists(self.notes_fn):
             return
         with open(self.notes_fn, 'rb') as fd:
```
dessant: Could we just open the file in text mode? (i have not tested :P)

dessant:
`NoteListItem` is instantiated by passing the `note_content` kwarg among others, but it has no
property with that name and that's why it crashes, a quick fix could be to not pass the kwarg or declare `note_content = StringProperty()`.

The note is opened after doing either of them, but i'm not able to type anything in that thing :yum:.

This is a bit old and unmaintained code, an awesome route would be to port it to RecycleView, this would also mean updating the tutorial which comes with this app.

KeyWeeUsr:
I did a quick fix for now. There is an argument `index` and `note_index` which are basically the same, but `index` isn't required by anything(+ missing note_content property), so it crashes. So I see probably two solutions:

- rewrite `note_index` as `index`
- remove `index`

However, I'm not sure if `index` isn't something a ListView creates probably by an accident(?) with `args_converter`.

dessant:
@KeyWeeUsr :grin:, it seems there is no such thing.

`index` comes from ListAdapter, but even if it's a bug there, it's ok to fix it here for now since that code will be ejected together with LIstView one day. :P

Issue #1735, opened 2013-12-29
No method to gracefully handle image loading exceptions


blhiggins:
See this thread for context: https://groups.google.com/forum/#!topic/kivy-dev/XtHr3kddqGk

To prevent crashing when attempting to load a bad image file into an Image widget (see https://github.com/kivy/kivy/issues/1561 also), Kivy master eats any exception that may be raised. This is suboptimal, as it presents no mechanism for the app itself to appropriately handle exceptions.

In my case, I want to be able to ignore images that cannot be loaded, which means not adding an image widget if there is a problem with it. As it stands, the widget is added (no exception is thrown)vbut the image is white, and the only way to check if the image was loaded successfully is to query if image._coreimage == None, which is not Pythonic, requires accessing a private variable, and gives the app no way to choose the most appropriate behaviour depending on the type of exception thrown.

tshirtman proposed 3 potential solutions:
1. revert the change referenced above and deal with it at application level 
2. avoid crashing only when loaded from kv (if possible) 
3. make it an option (BooleanProperty on Image, 'silentfail' for 
   example)

Any of those would solve the present issue, certainly. My suspicion is that #3, with silentfail = True by default, would be the most sensible way to address this in the short term.

AsyncImage also has no method to determine the status/success/failure of loading an image. There, perhaps the approach should be that it should catch exceptions and record them. The app could then query the AsyncImage widget to determine if the load has finished and what, if any, exceptions were caught. Perhaps, and maybe even better, this could be accomplished via a callback, like passing on_error = apps_exception_handler_function to the constructor, and having that be called when something goes wrong.

I suspect this issue betrays a more fundamental problem with the way kivy manages exception handling, particularly for things that it automates (like construction from a .kv) or does in the background (like AsyncImage loading). From what of the API and docs I have read, it doesn't sound like this probem has been thought out to any extent. Perhaps the on_error callback idea could be applied to Widgets generally, then, and solve this issue for the general case. It wouldn't be the ideal solution in the case of Image construction (not in a .kv file), but it might be a solution to consider for other problems regarding exception handling.

Peace,
Brendon
##

dessant:
This is fixed for remote loading with the `on_error` event, and there is currently a log message if a local image cannot be loaded. As an improvement, Image could also have `on_error`, please see https://github.com/kivy/kivy/pull/4545/files for how it could be handled.

Julian-O:
Appears to be fixed. Closing as complete.

If this assessment is incorrect, please let us know.

PR #4366, opened 2016-06-04
Fix #4062. Adds on_error event to ProxyImage

### LINKED ISSUE #4062: ProxyImage/Loader should have on_error event
Following [this](https://kivy.org/docs/api-kivy.loader.html) guide.

Here's the basic code (for an intercom system):

```
def update(self, dt):
        self.proxyImg = Loader.image(myUrl, nocache=True)
        self.proxyImg.bind(on_load=self.image_loaded)

def image_loaded(self, proxyImage):
    if proxyImage.image.texture:
        self.image.texture = proxyImage.image.texture
        pass
    if not self.paused:
        Clock.schedule_once(self.update, 1.0 / 30.0)
```

I noticed that the Loader/ProxyImage class doesn't have an `on_error` event (only `on_load`) which is troublesome.

The issue: if `myUrl` fails to load, then there is no way for me to signal that to the user/retry.

Looking into the code, it should be fairly simple to implement: In [`loader.py`](https://github.com/kivy/kivy/blob/master/kivy/loader.py#L348) line `348` the `Loader` class logs an exception. It would be nice if it also notified the ProxyImage that the image has failed to load.

ismailof:
This PR adds the 'on_error' event to ProxyImage, which can be dispatched when a remote image cannot be loaded, as proposed in #4062. The cause of the error (the exception info) is provided as suggested in #1735, so the user code can notify the user, retry the loading, debug or whatever it fits.
If the event is not binded, nothing extra happens, just as now.

### DIFF
--- a/kivy/loader.py
+++ b/kivy/loader.py
@@ -348,7 +348,8 @@ class LoaderBase(object):
             Logger.exception('Loader: Failed to load image <%s>' % filename)
             # close file when remote file not found or download error
             try:
-                close(_out_osfd)
+                if _out_osfd:
+                    close(_out_osfd)
             except OSError:
                 pass
             return self.error_image
--- a/kivy/loader.py
+++ b/kivy/loader.py
@@ -63,9 +63,12 @@ class ProxyImage(Image):
     :Events:
         `on_load`
             Fired when the image is loaded or changed.
+        `on_error`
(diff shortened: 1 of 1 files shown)


dessant:
Merged in https://github.com/kivy/kivy/pull/4545.

PR #4372, opened 2016-06-05
Added pitch shifting to audio_sdl2

Cheaterman:
This adds simple pitch shifting to audio_sdl2 as well as an example file.

### DIFF
--- a/examples/audio/pitch.py
+++ b/examples/audio/pitch.py
@@ -1,0 +1,44 @@
+# encoding: utf8
+
+from kivy.app import App
+from kivy.lang import Builder
+from kivy.clock import Clock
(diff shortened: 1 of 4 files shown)


examples/audio/pitch.py
```diff
@@ -0,0 +1,42 @@
...
+from kivy.core.audio import SoundLoader
+from kivy.uix.boxlayout import BoxLayout
+from kivy.uix.button import Button
+
+from sys import version_info
+
+
+NOTES = (
+    ('Do',  1),
+    ('Ré',  9/8.),
```
inclement: This should probably just be `Re`

tshirtman: or maybe just A, B, C…

PR #4383, opened 2016-06-10
Add Bubble.background_color explanation

KeyWeeUsr:
Adds explaining how to set the property as it doesn't work alone, but only with bg/arrow_image changed first.


akshayaurora:
@KeyWeeUsr thanks!

PR #4392, opened 2016-06-14
Added build method for homebrew-installed SDL2 libraries

JamesMakela:
Homebrew on Mac OS X has become the de-facto installer for 3rd party libraries,
and it doesn't install inside the OSX Framework packaging area. So kivy needs
a way of discovering whether the SDL2 libraries were installed via Hombrew.
I have modified setup to look for a Homebrew installation after failing to find
the SDL2 packages inside the Framework area.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ import sys
 
 from copy import deepcopy
 import os
-from os.path import join, dirname, sep, exists, basename, isdir
+from os.path import join, dirname, sep, exists, basename, isdir, isfile
 from os import walk, environ
 from distutils.version import LooseVersion
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


inclement:
This seems reasonable and a great improvement to me, thanks for the contribution! I'll leave it to an OS X using dev to check though, I'm not familiar with homebrew.

tito:
I don't understand, homebrew install libraries and link them in /usr/local/, which is already used by us. No others projects check for "homebrew" installed libraries, as it's not up to the project to check that imo.

For example, i do use brew on 2 different osx, i never had to hack anything, because it's linked in /usr/local, it just works. Do we miss anything or is your installation have an issue?

akshayaurora:
`USE_OSX_FRAMEWORKS=0 python setup.py`

Still gives you a issue while compiling?

JamesMakela:
@tito: If you are on a Mac in which you have no administrator privileges, which happens to be my case, then Homebrew may be configured to use a repository somewhere inside your home folder.  It is always a good policy to check.

dessant:
@akshayaurora, could you please guide this to a merge for 1.10?

akshayaurora:
@JamesMakela could you please rebase this ?

tito:
The approach is to fallback on pkg-config for both SDL2 and gstreamer, if frameworks have not been found. It shoudl work even if your homebrew is on your local directory, you just have to "brew install pkg-config" if you don't have it.

I saw https://github.com/Homebrew/legacy-homebrew/issues/39384 - but that indicated from the beginning that it works as PKG_CONFIG_PATH is defined to extend the path to your home directory.

So if you have pkg-config in your PATH (check the homebrew issue, it show where is it), then he should find the sdl2 configuration, and everything will work. There is no need to do something homebrew specific as pkg-config is a standard way to detect libraries available on your system, and brew already manage that. :)

PR #4402, opened 2016-06-16
Add allow_sides to DropDown

KeyWeeUsr:
Adds a property that allows placing items on the right side of a widget DropDown is attached to. If there is no space, it'll switch to the left side.

This allows behavior such as "nested" menu if there is hovering (or disabled dismiss) set for DropDown, so that an item would trigger `on_release: <another dropdown>.open(self)`.

(random image for this kind of "nested" menu)
image

**Example:**

<pre>from kivy.base import runTouchApp
from kivy.uix.dropdown import DropDown

def show_another(button, *largs):
    dp = DropDown(allow_sides=True)
    dp.bind(on_select=lambda instance, x: setattr(button, 'text', x))
    for i in range(10):
        item = Button(text='hello %d' % i, size_hint_y=None, height=20)
        item.bind(on_release=lambda btn: dp.select(btn.text))
        dp.add_widget(item)
    dp.open(button)

def show_dropdown(button, *largs):
    dp = DropDown(allow_sides=True, auto_dismiss=False)
    dp.bind(on_select=lambda instance, x: setattr(button, 'text', x))
    for i in range(10):
        item = Button(text='hello %d' % i, size_hint_y=None, height=20)
        item.bind(on_release=show_another)
        dp.add_widget(item)
    dp.open(button)

def touch_move(instance, touch):
    instance.center = touch.pos

btn = Button(text='SHOW', size_hint=(None, None), pos=(300, 200))
btn.bind(on_release=show_dropdown, on_touch_move=touch_move)

runTouchApp(btn)
</pre>

Also, there is a "bug", which if there is no space >= like DropDown's GridLayout(inner widget), it'll automatically switch to the bottom part of a widget in `attach_to` as a starting point as it would do with a default DropDown. This is clearly visible with `Dropdown.pos` if `pos <= GridLayout.height`

Should I add this example to the file too?

### DIFF
--- a/kivy/uix/dropdown.py
+++ b/kivy/uix/dropdown.py
@@ -30,8 +30,9 @@ dropdown. ::
     # create a dropdown with 10 buttons
     dropdown = DropDown()
     for index in range(10):
-        # when adding widgets, we need to specify the height manually (disabling
-        # the size_hint_y) so the dropdown can calculate the area it needs.
+        # when adding widgets, we need to specify the height manually
+        # (disabling the size_hint_y) so the dropdown can calculate the area
+        # it needs.
         btn = Button(text='Value %d' % index, size_hint_y=None, height=44)
 
         # for each button, attach a callback that will call the select() method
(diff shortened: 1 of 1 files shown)


dessant:
Same sentiment as here https://github.com/kivy/kivy/pull/4296#issuecomment-231480546.

PR #4404, opened 2016-06-16
logger: Use a shorter field width for non-colored output

gebart:
avoids unnecessarily long prefixes when not using color:

    [INFO              ] [Foo] Info message

instead becomes:

    [INFO   ] [Foo] Info message

to test, use `TERM=` set to something unknown:

    % TERM=asdf python examples/demo/showcase/main.py


gebart:
Added another commit which does the same change for the FileHandler. It makes the log files more closely match what is seen in the console.

inclement:
Seems reasonable, thanks.

PR #4410, opened 2016-06-18
Fix docs build errors

KeyWeeUsr:
When building docs without `silenced=yes` it throws _a lot_ of errors (~ 30+), so I decided to fix them + some pep8 edge(>79) cutting. This PR reduces errors to 3 basic ones:

<pre>WARNING: while setting up extension autodoc: directive 'autoclass' is already registered, it will be
 overridden

WARNING: while setting up extension preprocess: directive 'automethod' is already registered, it wil
l be overridden

C:\Users\...\doctest\kivy\doc\sources\examples\gen__canvas__repeat_texture__py.rst:34: WA
RNING: duplicate label canvas_mtexture1.png, other instance in C:\Users\...\doctest\kivy\
doc\sources\examples\gen__canvas__multitexture__py.rst</pre>

The last one means that two files use the same file as "anchor" set with ```.. _`<name>` ``` because they use the same file. Excluding the file wouldn't make sense and renaming the anchor would take too much time. Another solution might be duplicating the file and renaming it to `mtexture2` or something like that - or removing any reference from the second file that could trigger this warning.

*The most visible are:*

  - [excluded OSC library](https://kivy.org/docs/api-kivy.lib.html)
  - [note in packaging-android](https://kivy.org/docs/guide/packaging-android.html)
  - [missing headers mostly in osx](https://kivy.org/docs/guide/packaging-osx.html)
  - broken links/`:ref:`/anchors/`:attr:`

*Edits in `gallery.py`:*

To get rid of warnings I made a longer anchors using folder(s) prefix, therefore creating a unique anchor for each file. I think it's better because it works if someone wants to link to that file with `:ref:`

    gen__demo__kivycatalog__main__py.html#demo-kivycatalog-kivycatalog-kv

And a little bonus, I excluded old `gsoc` files so that it wouldn't ping with warnings about not being included in any toctree, because it's basically the latest `gsoc` file with some edits and to see different versions of files you can use github anyway. Hopefully I didn't break anything with `gallery.py` (building works, links too).

### DIFF
--- a/doc/gallery.py
+++ b/doc/gallery.py
@@ -10,6 +10,7 @@ in the examples/ directory to create rst files in the generation_dir
 
 import os
 import re
+from os.path import sep
 from os.path import join as slash  # just like that name better
 from kivy.logger import Logger
 import textwrap
@@ -52,7 +53,7 @@ def iter_filename_info(dir_name):
                     yield {'error': 'png filename not following screenshot'
                                     ' pattern: {}'.format(filename)}
                 else:
-                    d = m.group(2).replace('__', os.path.sep)
+                    d = m.group(2).replace('__', sep)
                     yield {'dunder': m.group(1),
                            'dir': d,
                            'file': m.group(3),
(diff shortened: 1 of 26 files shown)


inclement:
Looks good to me, and your changes sound sensible. I'll let the CI tests run before merging, just in case and so that others can comment if they like.

inclement:
Uh-oh, something has introduced conflicts, would you mind rebasing? It can't be anything too complicated.

KeyWeeUsr:
This https://github.com/kivy/kivy/commit/224c57f001f90fa24bcced30b76baa6912b8fa99 :yum:

tito:
Cool, thanks!

PR #4412, opened 2016-06-19
Make kivy clock thread safe and switch to linked list

matham:
This makes the Kivy clock fully thread safe and uses a linked list so that the callbacks are deterministic.

Benchmarks using https://github.com/kivy/kivy/wiki/Kivy-clock-experiments
Master:
```
schedule_once: 3.04238663305
trigger:  5.99657834953
unschedule_by_callback: 0.28418420994
unschedule_by_event: 7.20933188983
tick: 7.33034039193
```

PR:
```
schedule_once: 2.14062707372
trigger:  4.78726804852
unschedule_by_callback: 48.8890041079
unschedule_by_event: 6.2745237454
tick: 7.08543473329
```

### DIFF
--- a/kivy/clock.py
+++ b/kivy/clock.py
@@ -366,6 +366,7 @@ class ClockEvent(object):
                 else:  # we are in middle
                     prev.next = next
                     next.prev = prev
+            self.prev = self.next = None
 
         clock._lock_release()
 
--- a/kivy/clock.py
+++ b/kivy/clock.py
@@ -165,7 +165,7 @@ Threading
 The kivy clock is now fully thread safe.
 '''
 
-__all__ = ('Clock', 'ClockBase', 'ClockEvent', 'mainthread')
+__all__ = ('Clock', 'CyClockBase', 'ClockBase', 'ClockEvent', 'mainthread')
 
 from sys import platform
 from os import environ
@@ -177,7 +177,11 @@ from kivy.logger import Logger
 from kivy.compat import clock as _default_time
 import time
 from threading import Lock
+from kivy._clock import CyClockBase, ClockEvent
 
+
+def _get_sleep_obj():
+    pass
 try:
     import ctypes
     if platform in ('win32', 'cygwin'):
@@ -188,16 +192,15 @@ try:
 
         _kernel32 = ctypes.windll.kernel32
 
-        class _ClockBase(object):
-            def __init__(self):
-                self._timer = _kernel32.CreateWaitableTimerA(None, True, None)
+        def _get_sleep_obj():
+            return _kernel32.CreateWaitableTimerA(None, True, None)
 
-            def usleep(self, microseconds):
-                delay = ctypes.c_longlong(int(-microseconds * 10))
-                _kernel32.SetWaitableTimer(
-                    self._timer, ctypes.byref(delay), 0,
-                    ctypes.c_void_p(), ctypes.c_void_p(), False)
-                _kernel32.WaitForSingleObject(self._timer, 0xffffffff)
+        def _usleep(microseconds, obj=None):
+            delay = ctypes.c_longlong(int(-microseconds * 10))
+            _kernel32.SetWaitableTimer(
+                obj, ctypes.byref(delay), 0,
+                ctypes.c_void_p(), ctypes.c_void_p(), False)
+            _kernel32.WaitForSingleObject(obj, 0xffffffff)
     else:
         if platform == 'darwin':
             _libc = ctypes.CDLL('libc.dylib')
@@ -246,9 +249,8 @@ try:
         _libc.usleep.argtypes = [ctypes.c_ulong]
         _libc_usleep = _libc.usleep
 
-        class _ClockBase(object):
-            def usleep(self, microseconds):
-                _libc_usleep(int(microseconds))
+        def _usleep(microseconds, obj=None):
+            _libc_usleep(int(microseconds))
 
 except (OSError, ImportError, AttributeError):
     # ImportError: ctypes is not available on python-for-android.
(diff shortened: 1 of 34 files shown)


kivy/clock.py
```diff
@@ -626,36 +618,71 @@ def unschedule(self, callback, all=True):
...
+            except:
+                raise
+            else:
+                acquire()
+            event = self._next_event
+        release()
 
     def _process_events_before_frame(self):
-        found = True
+        acquire, release = self._lock_acquire, self._lock_release
```
tito: just a "perf" notation, it's less bytecode to write 2 lines, instead of generating a tuple and unpack it here ^^

kivy/clock.py
```diff
@@ -626,36 +618,71 @@ def unschedule(self, callback, all=True):
...
-            remove = events.remove
-            for event in events[:]:
-                # event may be already removed from original list
-                if event in events:
-                    event.tick(self._last_tick, remove)
+        acquire, release = self._lock_acquire, self._lock_release
+
+        acquire()
+        event = self._root_event
+        while event is not None:
```
tito: If we are going to allow changes in the linked list during the execution (as we were doing before), i think we could acquire/release on the whole loop (with a try/finally), and use Reentrant Lock instead. That would remove the use of acquire/release/try/catch within the loop :)

tito:
That's interesting. I should have try using lock for my original idea with indices, but i wrongly assumed this would slow things down. Look like python is already slow compared to lock congestion!

dessant:
Is there a way to make unschedule_by_callback a bit faster? Sometimes saving event ids is not practical and it also leads to more boilerplate.

kivy/clock.py
matham: Interesting, I would have though python would optimize such constant tuples.

kivy/clock.py
matham: There are two issues with that. 1) with a RLock, other threads will essentially be blocked during this whole stage until we are done processing all the events, and 2) in py3, RLock is twice as slow as Lock.

matham:
> Is there a way to make unschedule_by_callback a bit faster? Sometimes saving event ids is not practical and it also leads to more boilerplate.

It is doable if we also save a dict copy of all the callbacks, but that will be slower as it will require that we store them in lists in addition to the linked list.

Alternatively (probably a bad idea), we can store the event as a attr of the callback, e.g. `__kivy_clock_event` and use that to unlock.

Also, the reason that it looks so slow is that this is the worse performance as we always have to go through the entire linked list since `all` in unschedule defaults to True.

tito:
Saving in attr of a callback could be optional (if that's doable), because this will create a weakref and add into a list, for maybe an usage you don't want, as it's not optional.

Usually when you create interval / schedule in others toolkit, you have something in return to refer to it later (js, qt, gtk, glib, libevent, etc...). Instead of:

```
Clock.schedule_once(self.callback)
Clock.unschedule_once(self.callback)
```

It would be:
```
self._callback_id = Clock.schedule_once(self.callback)
self._callback_id.cancel()
```

We loose also the possibility to not check if the callback has already been released with Clock.unschedule().

matham:
I agree, but unschedule is used widely it seems. Personally, I don't use it that much. But looking in kivy itself it's used quite widely. Maybe we can change `all` to False which won't always cause worst time.

Or we can add a `cache` param to schedule/create_trigger which defaults to (False/True), and only saves it to a dict, like before, if true to preserve the old functionality. But I don't like this too much. But it may be used as a transitional step.

tito:
Adding a transitional step for deprecating it later, i don't see the point. But i think the cache may deserve a purpose: if you know you'll unschedule a callback later, but you don't want to keep the reference (or can't for any reason), use the cache?

We need to do the move anyway, because i'm sure the way you rewritten the Clock may have better performance later.

matham:
So should we default `cache=False` for now?

matham:
I suppose in kv often it's harder to save the ref. But, current users will experience a slowdown. Because if they do cache=False, the default and then call unschedule on the callback, we'll need to check the dict first (which won't have it), and then fallback to walking the linked list. So by default current users will experience a bigger slowdown. If we default cache to True, well, there's no point to it then.

tito:
I think the best would go and identify all the point where we actually use it, and see if it has a real impact on kv.

tito:
Not to mention the difference between perceived time, and actual time. It may not been an issue, we need to really test, even in blind test :)

matham:
By kv, I didn't mean kv lang by kivy itself, but I meant a user doing e.g. :
```
MyWidget:
    on_cheese: Clock.schedule_once(self.trap)
    on_caught: Clock.unschedule(self.trap)
```
In such cases users typically would not save the event. But I doubt anyway this is often used with a unschedule by users.

In kivy itself, I don't think it's an issue because a) it's not used in kv-lang at all. And for all the other places it should be easy to replace it with saving an event.

tito:
I never though that Clock.schedule_once in Kv itself was an actual thing (as you need to at least import it into the scope..) This look very messy!

matham:
Yeah, it's probably not that common. Personally I'm not too worried about this slowdown as canceling using the event itself is how it should be done.

And, implementing it in cython will probably make it a loooot faster.

matham:
After moving the core to cython I get:

```
schedule_once: 0.802818013001
trigger:  2.9213438012
unschedule_by_callback: 29.7284551574
unschedule_by_event: 3.06405578833
tick: 3.31444748379
```

matham:
I think it's done now and ready to merge.

tito:
Woot!

dessant:
@thopiekar, clock has a new compiled module, the daily ppa config needs updating.

flyyn:
How do we update this ?

PR #4424, opened 2016-06-23
update emacs mode docs to modern way of enabling newline and indent

duelafn:
Binding C-m to newline-and-indent no longer works in emacs. Not sure when the switch to electric-indent-local-mode became available or correct, but it works fine in Debian stable (emacs 24.4) so the change has been around long enough to justify updating the documentation.


inclement:
Sounds good, thanks!

PR #4440, opened 2016-06-30
Moved Slider styling to properties. Introducing value_track* properties.

rafalo1333:
Removed hardcoded slider kv and moved it's contents into Slider properties. It will allow to style the widget from both Python and kv.

Introducing value_track, value_track_color and value_track_width properties to let the slider to visualize the value as the line starting form 0 up to value_pos.


tshirtman:
merged in 79aeb700ebba0a884ba83b47abe59678d6c1a351

PR #4441, opened 2016-06-30
Fix dismiss in ActionGroup

### LINKED ISSUE #1906: ActionGroup in spinner mode doesn't close on selection.
If I run run the actionbar.py module directly, executing the "if main" block. The actionbar example loads without error. On my system both of the ActionGroups are displayed in spinner mode due to the available space. If I click an ActionGroup spinner it opens, but selecting an item in the group does not close the dropdown. The only way I have found to close the dropdown it to click elsewhere on the screen.

https://gist.github.com/philwrenn/9121379

There were no errors. I just added the log per the posting instructions. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1421602-actiongroup-in-spinner-mode-doesn-t-close-on-selection?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

KeyWeeUsr:
If there's no desire to actually _change_ group name after selecting an item there, then it's good to go. Maybe a dirty fix, but does the work. Still don't know what exactly prevents DropDown from closing. Maybe because it's inherited and put in another widget, therefore if I click on the button some function from DropDown isn't triggered?

I think using `return True` may be a problem here, because it doesn't even get to the autodismiss part(which is in dropdown as the last written), so... I moved it to the top basically.

Try `on_release`, for `Btn3`, works good.

Fix #1906

### DIFF
--- a/kivy/uix/actionbar.py
+++ b/kivy/uix/actionbar.py
@@ -15,9 +15,9 @@ An :class:`ActionBar` contains an :class:`ActionView` with various
 :class:`ContextualActionViews <kivy.uix.actionbar.ContextualActionView>`.
 An :class:`ActionView` will contain an :class:`ActionPrevious` having title,
 app_icon and previous_icon properties. An :class:`ActionView` will contain
-subclasses of :class:`ActionItems <ActionItem>`. Some predefined ones inlcude an
-:class:`ActionButton`, an :class:`ActionToggleButton`, an :class:`ActionCheck`,
-an :class:`ActionSeparator` and an :class:`ActionGroup`.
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
Now it should fix even those badly displayed properties and stuff in docs.

akshayaurora:
@KeyWeeUsr thanks

PR #4367, opened 2016-06-04
In textinput, don't try to split lines shorter than 1px

LogicalDash:
A line shorter than 1px, but longer than 0px, can legitimately
occur in layouts with size_hints that make a textinput shorter
than its text or hint_text.

Fixes #4244.

PR #4445, opened 2016-07-01
Prevent an infinite loop when trying to fit an overlong word into TextInput

### LINKED ISSUE #4244: TextInput.hint_text too long for TextInput.size_hint_x locks the app
If I have a TextInput whose `size_hint_x` makes it too small to hold all its `hint_text`, my app freezes (apparently in an infinite loop) instead of truncating the `hint_text` as I'd expect.

Steps to reproduce:

1. Make a TextInput, give it long hint_text and small size_hint_x (or put it in a BoxLayout so that a "large" size_hint_x still makes it too small)
2. Launch app

Expected behavior: get an app window that I can interact with
Actual behavior: app eats 100% CPU and doesn't display anything

Example code:
```
from kivy.lang import Builder


class TestBoxLayout(BoxLayout):
    pass


Builder.load_string("""
<TestBoxLayout>:
    orientation: 'vertical'
    Widget:
        id: foo
    Widget:
        id: bar
    BoxLayout:
        TextInput:
            hint_text: 'Some hint text'
            size_hint_x: 0.5
        Widget:
            id: spam
        Widget:
            id: eggs
""")


if __name__ == '__main__':
    runTouchApp(TestBoxLayout())
```

LogicalDash:
This is similar to my previous PR but handles a subtly different case.

As in #4367, this was making my app freeze on start as it tried to layout a TextInput with too much text in it. This time it's happening specifically because it can't split a word.

Re-fixes #4244. The test code in #4244 may also be used to test this PR.

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -2212,7 +2212,7 @@ class TextInput(FocusBehavior, Widget):
                     # split the word
                     for c in word:
                         cw = self._get_text_width(
-                            c, self._tab_width, self._label_cached
+                            c, self.tab_width, self._label_cached
                         )
                         if split_width + cw > width:
                             break
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -2167,17 +2167,6 @@ class TextInput(FocusBehavior, Widget):
             oldindex = index + 1
         yield text[oldindex:]
 
-    def _split_word(self, text, width):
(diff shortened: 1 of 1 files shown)


LogicalDash:
Refactored away the old _split_word method, by dessant's suggestion

dessant:
Thanks @LogicalDash :+1:

Issue #4137, opened 2016-04-01
"on_key_down" event in Windows is documented badly


janssen:
There's an extra undocumented argument, which seems to be a list of the modifier keys.

abject:
The argument `something_undocumented` seems to be an integer.

Furthermore `keycode` is not any more a list of the keycode value and text, it's only the keycode integer value.

A simple example showing this is:
``` python
from kivy.app import App
from kivy.uix.widget import Widget
from kivy.core.window import Window


class KeyDown(App):
    def build(self):
        Window.bind(on_key_down=self.key_action)
        return Widget()

    def key_action(self, *args):
        print("got a key event: %s" % list(args))


if __name__ == '__main__':
    KeyDown().run()
```

When pressing on a key you will some message like: 
```
got a key event: [<kivy.core.window.window_sdl2.WindowSDL object at 0x7f4831c012b8>, 113, 4, 'q', []]
```
for eg when I pressed on 'q' key.
or:
```
got a key event: [<kivy.core.window.window_sdl2.WindowSDL object at 0x7fb6a7e1d2b8>, 305, 224, 'ı', []]
got a key event: [<kivy.core.window.window_sdl2.WindowSDL object at 0x7fb6a7e1d2b8>, 113, 4, 'q', ['ctrl']]
```
 for Ctrl+q combo.

I cant' find any example in the documentation about the third argument passed by the event. All examples have only the following arguments: `keyboard, keycode, text, modifiers`.

So if I changed the previous example by:
```python
from kivy.app import App
from kivy.uix.widget import Widget
from kivy.core.window import Window


class KeyDown(App):
    def build(self):
        Window.bind(on_key_down=self.key_action)
        return Widget()

    def key_action(self, keyboard, keycode, text, modifiers):
        print("got a key event: %s" % list(args))


if __name__ == '__main__':
    KeyDown().run()
```
When I run it I get the following error:
```python
 Traceback (most recent call last):
   File "test.py", line 16, in <module>
     KeyDown().run()
   File "/usr/lib/python3/dist-packages/kivy/app.py", line 828, in run
     runTouchApp()
   File "/usr/lib/python3/dist-packages/kivy/base.py", line 504, in runTouchApp
     EventLoop.window.mainloop()
   File "/usr/lib/python3/dist-packages/kivy/core/window/window_sdl2.py", line 663, in mainloop
     self._mainloop()
[...]

PR #4446, opened 2016-07-01
Fix typos

KeyWeeUsr:
Fixes:
- #4137 (there's missing word)
- code block in event (hiding `::`)
- parser aliases - in docs they are jumping out of casual formatting, because of `parse_int = int`. This wraps it into functions.

### DIFF
--- a/kivy/parser.py
+++ b/kivy/parser.py
@@ -103,7 +103,7 @@ def parse_bool(text):
 
 
 def parse_string(text):
-    '''Parse a string to a string (removing single and double quotes)'''
+    '''Parse a string to a string (removing single and double quotes).'''
     if len(text) >= 2 and text[0] in ('"', "'") and text[-1] in ('"', "'"):
         text = text[1:-1]
     return text.strip()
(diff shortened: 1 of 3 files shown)


kivy/parser.py
```diff
@@ -150,5 +150,13 @@ def parse_float4(text):
     return value
 
 
-parse_int = int
-parse_float = float
+def parse_int(text):
```
akshayaurora: please separate doc changes to not include any code changes

KeyWeeUsr: @akshayaurora Do you mean that wrapping `int()` and `float()` to functions? That's actually the point, because then It'll highlight them as the remaining functions. Otherwise it'll be like [this.](https://kivy.org/docs/api-kivy.parser.html#kivy.parser.parse_int)

akshayaurora: In that case the existing doc display is ok. You are adding a extra method call where it is really not needed. Just for the sake of keeping doc uniform.

KeyWeeUsr: Ah, I understand now.

akshayaurora:
Thanks for all your nice work @KeyWeeUsr reopen when highlighted issues are fixed.

PR #4447, opened 2016-07-01
Fix size text align

### LINKED ISSUE #4416: FileChooserListView - bad position for Sizes
What title says. It's positioned badly even inside folders. Maybe some separator was removed or Label size is its text's size?

image

KeyWeeUsr:
Fixes #4416


dessant:
Thanks :dancers:

PR #4448, opened 2016-07-01
Fix url in UrlRequest

KeyWeeUsr:
UrlRequest somehow isn't able to switch to https automatically(issue or not?) and here if the api isn't through https, the request fails. However, there it gets interesting, because the error is empty. Rewriting it to https fixes it and the result is available.


akshayaurora:
@KeyWeeUsr thanks for the short fix, this is definitely a issue.

PR #4449, opened 2016-07-02
Store kivy_home_dir as a unicode string in python 2.

### LINKED ISSUE #3912: Logger does not create log files when username contains non-ascii characters
This happens on Windows.

You need to create a user with a non-ascii name:

Control panel->Change keyboards->Add keyboard
Polish (programmers)

Change your locale (without this, nothing will work)
"Administrative" tab -> Change system locale -> Polish

Add user Łukasz (right Alt+L to get the Ł with polish keyboard selected)

Log in as Łukasz, install python, kivy (if you don't have polish locale, this will fail). Alternatively, you could package a python file with pyinstaller --onefile and then pass it to the non-ascii user. This will work the same but you won't be able to debug so easily.
Run python and do "import kivy".

You will get:
```
>>> import kivy
[ERROR             ] Error while activating FileHandler logger
  File "C:\Python27_Lukasz\lib\site-packages\kivy\logger.py", line 205, in emit
    self._configure()
  File "C:\Python27_Lukasz\lib\site-packages\kivy\logger.py", line 153, in _configure
    _dir = os.path.join(_dir, log_dir)
  File "C:\Python27_Lukasz\lib\ntpath.py", line 85, in join
    result_path = result_path + p_path
UnicodeDecodeError: 'ascii' codec can't decode byte 0xa3 in position 7: ordinal not in range(128)
[INFO              ] [Kivy        ] v1.9.1
[INFO              ] [Python      ] v2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:32:19) [MSC v.1500 32 bit (Intel)]
>>>
```
Kivy works fine but the logger won't create any file and no logs will be saved to disk.
Decoding user_home_dir to an unicode string fixes the problem:

C:\Python27_Lukasz\Lib\site-packages\kivy>git diff
```
diff --git a/__init__.py b/__init__.py
index 0f9e128..af2b970 100644
--- a/__init__.py
+++ b/__init__.py
@@ -278,7 +278,7 @@ if not environ.get('KIVY_DOC_INCLUDE'):
     if 'KIVY_HOME' in environ:
     else:
-        user_home_dir = expanduser('~')
+        user_home_dir = expanduser('~').decode(sys.getfilesystemencoding())
         if platform == 'android':
         elif platform == 'ios':
```		 
My guess is the same thing should be done with the kivy_home_dir variable as it is also join()ed in the same way.

overfl0:
For Python 3 there is no change. For python 2 and users without
special characters, there is no change (because the string is easily
decoded with ascii codec anyway). The only change is for users that
have special characters in their username.

See https://github.com/kivy/kivy/issues/3912#issuecomment-229679674 for rationale and issue #3912 for reproduction.
Fixes #3912.

### DIFF
--- a/kivy/__init__.py
+++ b/kivy/__init__.py
@@ -36,6 +36,7 @@ from getopt import getopt, GetoptError
 from os import environ, mkdir
 from os.path import dirname, join, basename, exists, expanduser
 import pkgutil
+from kivy.compat import PY2
 from kivy.logger import Logger, LOG_LEVELS
 from kivy.utils import platform
 
@@ -284,6 +285,10 @@ if not environ.get('KIVY_DOC_INCLUDE'):
         elif platform == 'ios':
             user_home_dir = join(expanduser('~'), 'Documents')
         kivy_home_dir = join(user_home_dir, '.kivy')
+
+    if PY2:
+        kivy_home_dir = kivy_home_dir.decode(sys.getfilesystemencoding())
+
     kivy_config_fn = join(kivy_home_dir, 'config.ini')
     kivy_usermodules_dir = join(kivy_home_dir, 'mods')
     kivy_userexts_dir = join(kivy_home_dir, 'extensions')


dessant:
@matham, please review this. :cat:

matham:
I wonder, wouldn't `user_home_dir = expanduser(u'~')` solve the issue?

overfl0:
Unfortunately, it won't.

```
>>> os.path.expanduser(u'~')
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\ntpath.py", line 311, in expanduser
    return userhome + path[i:]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xa3 in position 9: ordinal not in range(128)
```

```
>>> os.path.expanduser('~').decode(sys.getfilesystemencoding())
u'C:\\Users\\\u0141ukasz'
```

overfl0:
And in case you're curious:
```
>>> os.path.expanduser('~')
'C:\\Users\\\xa3ukasz'
>>> sys.getfilesystemencoding()
'mbcs'
```

In theory, you could hardcode those values to verify how the patch is working without creating a special user just for this PR but I'm afraid that strange things will happen if you don't have the right locale set up on your machine.

Why `u'~'` does not work on python 2 is beyond me, but well... there is a reason python 3 was made :).

matham:
Alright :+1:

PR #4455, opened 2016-07-08
Focused widget inside ScrollView should unfocus on tap. Fixes #4399

### LINKED ISSUE #4399: TextInput does not lose focus inside ScrollView
Hi. I found the issue with behavior of TextInput focus inside ScrollView. If i touch other widget - TextInput does not lose focus despite the property unfocus_on_touch: True.

Sample program:
```
<ScrollView>:
    BoxLayout:
        orientation: 'vertical'
        Label:
            text: 'lol'
        TextInput:
            text: 'Hello World'
```
When I click on the Label, Textinput does not lose focus. But If I remove ScrollView from code - it works as intended.

Fak3:


Fak3:
FocusBehavior._handle_post_on_touch_up() handles unfocus logic, taking in the consideration the list of ignored events, which should not trigger unfocus. Currently ScrollView makes any touch ignored by all focusable widgets. This happens in the ScrollView.on_scroll_start() - this method is called for every touch down event inside the ScrollView. Even the simple tap ouside of the currently focused widget will get ignored.
This PR changes behavior so that ScrollView will make focusable widgets to ignore only the touches, that resulted in the actual scroll - touch-and-drag events. So the tap outside the focused widget will be handled, and it will unfocus itself properly.

matham:
`sv.handled` being in touch is not guaranteed to mean anything. So we cannot rely on it. Checking `...[sv.handled]['x'] or ...[sv.handled]['y']` is also not reliable because even if it didn't fully scroll it probably shouldn't be used to defocus the touch. So maybe adding `...['sv.ignore_touch'] = True` should be used instead.

In fact, there may be other cases where we should ignore the touch, but with your change it won't anymore. I believe https://github.com/kivy/kivy/blob/master/kivy/uix/scrollview.py#L646 is one such example.

Fak3:
> `sv.handled` being in touch is not guaranteed to mean anything. So we cannot rely on it. 

Oh, thats right. Thanks.

> Checking `...[sv.handled]['x'] or ...[sv.handled]['y']` is also not reliable because even if it didn't fully scroll it probably shouldn't be used to defocus the touch. 

Why? Checking `...[sv.handled]['x'] or ...[sv.handled]['y']` should do exactly the thing you said: _"even if it didn't fully scroll it is not used to defocus"_ That means any touch+move is ignored by FocusBehavior._handle_post_on_touch_up().

> In fact, there may be other cases where we should ignore the touch, but with your change it won't anymore. I believe https://github.com/kivy/kivy/blob/master/kivy/uix/scrollview.py#L646 is one such example.

Scrolling with mousewheel is ignored in the FocusBehavior._handle_post_on_touch_up(), so the line you pointed looks irrelevant to me. Mousewheel events will not defocus anything.

matham:
> Why? Checking ...[sv.handled]['x'] or ...[sv.handled]['y'] should do exactly the thing you said: "even if it didn't fully scroll it is not used to defocus" That means any touch+move is ignored by FocusBehavior._handle_post_on_touch_up().

...

> Scrolling with mousewheel is ignored in the FocusBehavior._handle_post_on_touch_up(), so the line you pointed looks irrelevant to me. Mousewheel events will not defocus anything.

This was just an example. You can see identical behavior even when not using the wheel. E.g. at https://github.com/kivy/kivy/blob/master/kivy/uix/scrollview.py#L759. That will not set `touch.ud['sv.handled']['x']` to True, yet that touch should not be used for defocusing. Even though it didn't scroll. There's a difference between a touch that doesn't scroll and a touch that de-focuses.

Fak3:
> Even though it didn't scroll.

This made me wonder what is the semantics of the `['sv.handled']['x']`? Why it is False when overscroll happens? Shouldn't it actually be True? I would say that overscroll is still the scroll.

matham:
I didn't really write that code, so I'm not sure. But, whatever changes are made will need to retrace all the code paths to make sure it doesn't break corner cases.

Fak3:
I updated PR. Introduced `sv.can_defocus` flag, independent of `sv.handled`.

PR #4456, opened 2016-07-08
Update installation-linux.rst

JackAnderson5:
Unless (e.g.) `-p python2.7` is provided, it will use the default interpreter when creating the virtualenv. The default might be unwanted (e.g. use the python3.5 while the user needs specifically 2.7) and go unnoticed for quite a while wasting lots of time for the user.

The second change (irrelevant to the first) is uncommenting a command.


Zen-CODE:
Thanks :-)

PR #4485, opened 2016-07-23
Add default font function

mixedCase:
This replicates the functionality from PR kivy/kivy#2412 but makes it
mergeable with current master.

It probably needs documentation, not sure where to add it. Thoughts?

### DIFF
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -57,6 +57,15 @@ FONT_BOLDITALIC = 3
 whitespace_pat = re.compile('( +)')
 
 
+def set_default_font(fontname):
+    global DEFAULT_FONT
+    DEFAULT_FONT = fontname
+    from kivy.uix.label import Label
+    Label.font_name.defaultvalue = fontname
+    from kivy.uix.textinput import TextInput
+    TextInput.font_name.defaultvalue = fontname
+
+
 class LabelBase(object):
     '''Core text label.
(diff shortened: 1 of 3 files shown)


rafalo1333:
@mixedCase good one, but in set_default_font I would use kivy.resources.resource_find() method, instead of simply assigning the value.

dessant:
@rafalo1333, `resource_find` is used internally for resolving `font_name`.

dessant:
I think the previous pr was held back because it didn't properly fix the issue. The correct solution would also let us specify a default font *before* `LabelBase.register` is called [here](https://github.com/kivy/kivy/blob/c779254148c3e45d0fd2525e9f0b165e12f7bec0/kivy/core/text/__init__.py#L778), and unregister the current default font if set_default_font is called during runtime.

In any case, DEFAULT_FONT should not be a string, but a dict so we can properly pass values to `register`.
For the current default, it would look like this (`italic`, `bold` and `bolditalic` are optional):

```python
DEFAULT_FONT = {'name': 'Roboto',
                'regular': 'Roboto-Regular',
                'italic': 'Roboto-Italic',
                'bold': 'Roboto-Bold',
                'bolditalic': 'Roboto-BoldItalic'}
```

mixedCase:
Truth be told I wasn't the one originally interested in adding this functionality and I thought the problem was just getting the previous PR ready to merge. Therefore I'll be closing this PR, sorry.

PR #4490, opened 2016-07-25
return `abspath` in `resource_find`. Seems to be needed when android …

akshayaurora:
…uses apkexpansion

needed for https://github.com/kivy/python-for-android/issues/852

### DIFF
--- a/kivy/resources.py
+++ b/kivy/resources.py
@@ -51,12 +51,12 @@ def resource_find(filename):
         return None
     if filename[:8] == 'atlas://':
         return filename
-    if exists(filename):
+    if exists(abspath(filename)):
         return abspath(filename)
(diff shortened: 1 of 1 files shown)


kivy/resources.py
```diff
@@ -52,11 +52,11 @@ def resource_find(filename):
     if filename[:8] == 'atlas://':
         return filename
     if exists(filename):
-        return filename
+        return abspath(filename)
```
matham: There may be some subtle differences between `exists` and `abspath`, so that even if `exists(filename)`, `exists(abspath(filename))` may be False (although it probably is the same, but it never hurts to make sure). So it should probably be changed to `if exists(abspath(filename)): return abspath(filename)`. Similarly, for `output` below.

PR #4494, opened 2016-07-27
AnimationTransition: update ClutterAlpha url.

rafalo1333:
Changed the url to the newer one in gnome.org domain. Old one is broken (404).


inclement:
Thanks!

Issue #4477, opened 2016-07-18
touch.pos for on_touch_down / on_touch_up differ when not moving the mouse


johanneshk:
I don't understand if this is a bug or a feature:
https://gist.github.com/johanneshk/5bfa9a3983e81a20ed16ed1ff0d80257
Click on the red square and observe the console prints: For me touch.pos for on_touch_down and on_touch_up differ. Why? If you replace ScrollView with RelativeLayout in L.48 it both positions coincide (imo the correct behavior). 
Example:

down <__main__.Element object at 0x7f9b507ebf50> touch.pos (21.0, **32.0**) to window (221.0, 562.0) True
up <__main__.Element object at 0x7f9b507ebf50> touch.pos (21.0, **-448.0**) to window (221.0, 82.0) False

The y-coordinates is shifted in some way I don't understand.

bionoid:
If you click-hold for a second, releasing the mouse button produces an extra "down" event, which probably should not be happening. The output looks like this:

```
down <__main__.Element object at 0x0000000003E118D0> touch.pos (27.0, -503.0) to window (227.0, 27.0) False
down <__main__.Element object at 0x0000000003E118D0> touch.pos (27.0, 27.0) to window (227.0, 557.0) True
up <__main__.Element object at 0x0000000003E118D0> touch.pos (27.0, -503.0) to window (227.0, 27.0) False
```

johanneshk:
Huh... so the position in the second 'down' (which is actually triggered upon button release) is correct. Unfortunately I don't really understand how touch events are digested by ScrollView (but in my opinion the culprit is somewhere in there). So we seem to have two issues:

- why is another 'down' touch event dispatched when releasing the button
- why is the position different

uclatommy:
I am working on a retina screen and this behavior causes buttons to never trigger the `on_release` event. The `touch.pos` when `on_touch_down` occurs is converted `to_local` so `collide_point` is `True`. However, when the mouse is released in the exact same position, the `touch.pos` is in absolute coordinates which are outside of the button's `right` and `top` coordinates. For example, here are the button coordinates:
```
>>> print('x:{}, right:{}, y:{}, top:{}'.format(self.x, self.right, self.y, self.top))
x:0, right:216.0, y:0, top:38
```
I put a breakpoint at `button.py:137` (the `on_touch_down` function) and print the touch position:
```
>>> print(touch.pos)
(72.0, 18.0)
```
[...]

PR #4498, opened 2016-07-28
Correctly transform touch_up position

johanneshk:
See issue #4477

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -1058,7 +1058,7 @@ class ScrollView(StencilView):
     def _do_touch_up(self, touch, *largs):
         # touch is in window coords
         touch.push()
-        touch.apply_transform_2d(self.to_widget)
+        touch.apply_transform_2d(self.to_local)
         super(ScrollView, self).on_touch_up(touch)
         touch.pop()
         # don't forget about grab event!


inclement:
I notice you mention in the linked issue that it might have some issues - did you work out what was going on there?

johanneshk:
@inclement Yes, the FileChooserListView somehow changes (can't really describe: Some clicks seem to go missing, opens wrong folders, slower). See this Gist: https://gist.github.com/johanneshk/5bfa9a3983e81a20ed16ed1ff0d80257#file-main2-py

This PR seems to fix the initial behavior, but for some reason the FilechooserListView is broken (at least for me?). I don't know why, but that's what I'm experiencing...

johanneshk:
Could someone confirm my Edit above? Plus maybe the behavior in general? If this is correct, I could go ahead and look into FileChooserListView and why it behaves differently.

Thanks!

matham:
At a first glance you're correct. Because `_do_touch_up` is `"called"` from `on_scroll_stop` which receives touches in parent coordinates, not window coordinates. So `to_local` rather than `to_widget` should be used. That is your reasoning, right?

The problem with this though is that it's `"scheduled"`, not called from `on_scroll_stop`. So in between the scheduling, when it's in parent coordinates, and the execution; the touch is likely to have been converted back to window coordinates. That's because he touch starts and ends in window coordinates.

At least this is how I understand this.

johanneshk:
@matham  your explanation makes perfect sense. Plus the first comment in '_do_touch_up' is 'touch is in window coords'. With the pop statements you added this now seems to be true in all cases.

So please don't merge this PR but rather #4508 .

PR #4501, opened 2016-07-31
Add on_stop to recorder

KeyWeeUsr:
Adds `on_stop` event to the recorder. `stop` is not a property, so this shouldn't be a problem. I think it should be included, because even if there is a way how to do this now, it's not pretty, because it'd mean rewriting `on_play` i.e. making a class that inherits from `Recorder`, then calling `Recorder.on_play` probably via `super()` in that custom class and then checking for the `play` value.

However, this is dispatched automatically when `self.play = False` in `update()` function, so it's pretty easy to work with. Also, it's placed _after_ `self.play = False`, therefore the event would run when `on_play` is finished.

### DIFF
--- a/kivy/input/recorder.py
+++ b/kivy/input/recorder.py
@@ -112,6 +112,13 @@ class RecorderMotionEvent(MotionEvent):
 
 class Recorder(EventDispatcher):
     '''Recorder class. Please check module documentation for more information.
+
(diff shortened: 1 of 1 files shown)


akshayaurora:
@KeyWeeUsr thanks 👍

PR #4502, opened 2016-08-01
Fix regex for new nightly name

KeyWeeUsr:
Change in wheel's name breaks fetching it via javascript on [installation-windows](https://kivy.org/docs/installation/installation-windows.html):

    Kivy-1.9.2.dev0_30062016_git_8d5451f-cp27_30062016_git_8d5451f-none-win32
    ↓
    Kivy-1.9.2.dev0_31072016_git_cfec8f3-cp27-cp27m-win32.whl

I added it only as another "or" to the regex in case the name would be reverted or changed in a way the old part would match the wheel.

### DIFF
--- a/doc/sources/.static/kivy.js
+++ b/doc/sources/.static/kivy.js
@@ -322,7 +322,7 @@ cs\\/doclist\\/images\\/icon_9_archive_xl128\\.png"\\/><\\/div>\
 https:\\/\\/ssl\\.gstatic\\.com\\/docs\\/doclist\\/images\\/icon_9_\
 archive_list\\.png"\\/><\\/div><div class="flip-entry-title">)';
     var wheel = '(Kivy-\\d\\.\\d\\.\\d)(\\.\\w{4}_$date$_git\\_?\\w\
-{7}-$pyVer$)(-none|_\\d{8}_git_\\w{7}-$pyVer$m)(-$arch$.whl)';
+{7}-$pyVer$)(-none|_\\d{8}_git_\\w{7}-$pyVer$m|-$pyVer$m)(-$arch$.whl)';
     var date = new Date();
     var yesterday = addZeros(date.getDate() - 1, 2);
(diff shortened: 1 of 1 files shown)


matham:
I did change (fix) the names. However, I also added files with constant names that get updated with every nightly so that we can have direct links and remove the javascript altogether. @kived was saying that the java script seemed to be very round-about. But also, I'd prefer that we do not have js if we don't need it.

KeyWeeUsr:
I that case I'm all for the removing that js. For some people, including me, it may seem as unsafe solution using js to request files from other website even if it is for framework within its own website and therefore discourage them from even trying latest Kivy without compiling.

I'm of the same opinion that it's a big "round-about", because it has to find the link for the latest wheel file through ajax (which isn't even pure ajax, but some yahoo api that makes it easier) -- I wasn't aware that you can have a single link that would work this way with Drive except manually updating the file e.g. when you right click on the file -> versions -> upload new version for each nightly (4x manually, each night).

But I think it was still better than scrolling and checking for the latest name manually, especially when there was `grid` display by default and user couldn't even sort the files with display changed to `list` :yum:

matham:
Those 4 nightlies are, or at least should be updated in place every night. That means the files are not changed, just their content is updated. So we should be able to get that the link directly to the 4 files and not have to update the link every night. Wouldn't that work instead of the js? Or is this what you were saying? I'm not that familiar with js/html so I'm not really sure how to update the website to reflect that.

KeyWeeUsr:
Yup, that's exactly what I'm saying. If there's an option to avoid insecure-looking workaround, then let's do it.

Btw, is there some google drive api you are doing it with or how exactly do you change the content of the files?

matham:
I use PyDrive. The re-use happens here: https://github.com/kivy/kivy-sdk-packager/blob/master/win/gdrive.py#L99

KeyWeeUsr:
ref #4516

Issue #4504, opened 2016-08-02
StackLayout children rearrange themselves unexpectedly when their parent's size changes 


JackAnderson5:
Here is [the code ](https://gist.github.com/JackAnderson5/2c24c2bcf6da4bcee6219f1fabb16014).

I'm trying to populate a StackLayout with 5 buttons, so i set their `size_hint_x` to "one fifth" of their parent, or `1/5.`. This is equal to `0.2`. 

So, when parent width is `800` everything is fine; 5 buttons per row.

When the parent width is `801` each button has width `801*0.2`, that is `160.20000000000002` (floating point precision issues... the actual value should be `160.2`). The problem is that if we add 5 of those we get `160.20000000000002*5` which is `801.0000000000001`, which is greater than `801`. Therefor **i assume** StackLayout says "that exceeds my dimensions! Send it to the next raw."; results in 4 buttons per row.

In other words, floating point precision increases `width` or `height` by a tiny bit making children not fit sometimes. 

<i>Note that `801*(1/5.)` **!=** `801*1/5`, kivy gets the first one.</i>

matham:
That seems like a problem that would be hard to solve generally.

JackAnderson5:
Wouldn't simply rounding down the size of children solve it? 

e.g. `160.20000000000002` -> `160.2000000000000`

matham:
Yeah, but is a significant amount? What works for your case will not work when the number of children is a lot higher. And rounding down arbitrarily the size of children by one will display incorrectly in other situations. 

I'm not sure if rounding will not effect how the user sees graphics when we display anything that is smaller than one pixel on the same line. Because sometimes it really does need to be displayed on the next row. Although I'm not sure how sensitive users are to these sub-pixel issues.

JackAnderson5:
I'm not very familiar with kivy but.. 

> [...]will not work when the number of children is a lot higher

.. in order to be off by a single pixel in a whole StackLayout there would have to be trillions* of widgets on the same row of that StackLayout 

*(assuming we use something like `round(x, 12)` instead of `round(x, 3)`)

matham:
Yeah, I guess you are right. Something like when `sum(children) - available_width < 1e-9` they are considered to be on the same line. A PR is welcome.

JackAnderson5:
PR nearly ready, will do some benchmarking on performance of the 2 about solutions (rounding vs subtracting), and will PR.

JackAnderson5:
[...]

PR #4507, opened 2016-08-03
Update stacklayout.py; Fixes issue 4504

JackAnderson5:
Fixes [issue 4504](https://github.com/kivy/kivy/issues/4504). 

`lc =[]` is irrelevant; simply moved closer to where it's used.

### DIFF
--- a/kivy/uix/stacklayout.py
+++ b/kivy/uix/stacklayout.py
@@ -152,8 +152,6 @@ class StackLayout(Layout):
         padding_y = padding_top + padding_bottom
         spacing_x, spacing_y = self.spacing
 
-        lc = []
-
         # Determine which direction and in what order to place the widgets
         posattr = [0] * 2
         posdelta = [0] * 2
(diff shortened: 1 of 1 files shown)


tshirtman:
i'd really like a better solution, but unless someone comes up with a better idea, it'll have to get merged…

matham:
The only suggestion I have is to not actually change the size of the child as this pr does using `sizes`, but instead only use that `1e-12` trick when checking if ti fits. I'm worried that e.g. on a 640 width with two children instead of each child's size being 320, it'll now be `320 - 1e-12` which may have other effects, but will also make it harder to display the number cleanly.

user5061:
Sorry for the mess. Didn't know how to post it here. There [you go](https://github.com/kivy/kivy/pull/4579).

<sub>Offtopic: Original issue with StackLayout and original fix are raised by my old account</sub>

user5061:
TL;DR: `1e-12` should be fine. (Much smaller values could be a problem though.)

Initially i used `1e-9` just to be on the safe side, since i didnt know up to how many digits a float is "precise". For example `101*.2=20.200000000000003`, error occurs on 15th decimal so subtracting `1e-15` would be a bad idea. However, with `2201*.8=1760.8000000000002`, the error is on 13th decimal.

dessant:
Closing in favor of https://github.com/kivy/kivy/pull/4579.

PR #4510, opened 2016-08-04
fixed py2 kivy/uix/image ascii error

Bakterija:
Fixed unicode source ascii errror for python2 and
Added logging for texture_update exceptions


Bakterija:
added simple app example https://github.com/Bakterija/push_examples/tree/master/kivy_4510%20fixed%20py2%20kivy_uix_image%20ascii%20error

Zen-CODE:
Thanks. That's a nice refinement :-)

PR #4511, opened 2016-08-04
Fix dropdown and spinner issue

matham:
Fixes:

* Drowdown being dismissed even when clicking within the dropdown
* Spinner creating everything anew even when only size changed. It led to infinite size computations when bound to size of the children.

### DIFF
--- a/kivy/uix/dropdown.py
+++ b/kivy/uix/dropdown.py
@@ -314,6 +314,8 @@ class DropDown(ScrollView):
             return True
         if 'button' in touch.profile and touch.button.startswith('scroll'):
             return
+        if self.collide_point(*touch.pos):
+            return True
         if self.auto_dismiss:
             self.dismiss()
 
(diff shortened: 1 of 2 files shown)


AndiEcker:
I think the code in line 169 of spinner.py should be:

for item in container.children:

instead of:

for value in container.children:



2016-08-05 0:43 GMT+01:00 matham <notifications@github.com>:

AndiEcker:
Actually the new method could be written more "clean·" as:

+  def _update_dropdown_size(self, *largs):
 +        if self.sync_height and self._dropdown and self._dropdown.container:
 +            for item in self._dropdown.container.children:
 +                item.height = self.height

matham:
Thanks, fixed with https://github.com/kivy/kivy/commit/f8f156fd41fba54fbf80071fc2992aef09ac86e4.
About the more compact version, it's less efficient than the current version because there are more attribute access.

PR #4513, opened 2016-08-05
Add an example for multiple Window.on_dropfile

KeyWeeUsr:
Useful and not mentioned yet thing in the examples. :yum:

### DIFF
--- a/examples/miscellaneous/multiple_dropfile.py
+++ b/examples/miscellaneous/multiple_dropfile.py
@@ -1,0 +1,50 @@
+from kivy.app import App
+from kivy.uix.button import Button
+from kivy.core.window import Window
+from kivy.uix.boxlayout import BoxLayout
+
+
+class DropFile(Button):
+    def __init__(self, **kwargs):
+        super(DropFile, self).__init__(**kwargs)
+
+        # get app instance to add function from widget
+        app = App.get_running_app()
+
+        # add function to the list
+        app.drops.append(self.on_dropfile)
+
+    def on_dropfile(self, widget, filename):
+        # a function catching a dropped file
+        # if it's dropped in the widget's area
+        if self.collide_point(*Window.mouse_pos):
+            # on_dropfile's filename is bytes (py3)
+            self.text = filename.decode('utf-8')
+
+
+class DropApp(App):
+    def build(self):
+        # set an empty list that will be later populated
+        # with functions from widgets themselves
+        self.drops = []
+
+        # bind handling function to 'on_dropfile'
+        Window.bind(on_dropfile=self.handledrops)
+
+        box = BoxLayout()
+        dropleft = DropFile(text='left')
+        box.add_widget(dropleft)
+        dropright = DropFile(text='right')
+        box.add_widget(dropright)
+        return box
+
+    def handledrops(self, *args):
+        # this will execute each function from list with arguments from
+        # Window.on_dropfile
+        #
+        # make sure `Window.on_dropfile` works on your system first,
+        # otherwise the example won't work at all
+        for func in self.drops:
+            func(*args)
+
+DropApp().run()


Zen-CODE:
Sorry, this does not work for me, on 1.9.0 or master? Is this for a specific OS?

KeyWeeUsr:
On master, I tested it on Windows, but should work on others too as it basically executes functions from list when `on_dropfile` is dispatched.

Zen-CODE:
Pity. Seems that does not work on linux (Mint 17.3 KDE).

KeyWeeUsr:
Does `on_dropfile` itself work? I don't see a reason why it shouldn't work if the event itself works.

Zen-CODE:
Well, the 'on_dropfile' never fires. The button captions stay the same whatever one drops on them. Looking over the code, that's what should happen?

KeyWeeUsr:
Exactly what shouldn't happen. On each drop depending on the area where the drop is the appropriate Label should change its text.

As I read the code, the problem might be here depending on your window provider https://github.com/kivy/kivy/blob/master/kivy/core/window/__init__.py#L1407

Check console with this:

```
from kivy.app import App

class DropApp(App):
    def printer(self, *args):
        print(args)

    def build(self):
        Window.bind(on_dropfile=self.printer)
        box = BoxLayout()
        return box
DropApp().run()
```

Zen-CODE:
Nope, nothing happens.

[INFO   ] [Kivy        ] v1.9.2-dev0
[INFO   ] [Python      ] v2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2]
[INFO   ] [Factory     ] 193 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_gif, img_sdl2, img_pil (img_ffpyplayer ignored)
[INFO   ] [Text        ] Provider: sdl2

examples/miscellaneous/multiple_dropfile.py
```diff
@@ -0,0 +1,48 @@
...
+        # get app instance to add function from widget
+        app = App.get_running_app()
+
+        # add function to the list
+        app.drops.append(self.on_dropfile)
+
+    def on_dropfile(self, widget, filename):
+        # a function catching a dropped file
+        # if it's dropped in the widget's area
+        if (self.pos[0] < Window.mouse_pos[0] < self.pos[0] + self.width and
```
dessant: This could be:
```python
if self.collide_point(*Window.mouse_pos):
    self.text = filename.decode('utf-8')
```

KeyWeeUsr:
Then it's probably window provider's fault, because it should print all args if the event is dispatched with the check I pasted here. If it's not dispatched, therefore there's an issue probably with getting user's input(or making it available for provider) from the provider.

Are you using different provider than the default one(sdl2)?

examples/miscellaneous/multiple_dropfile.py
KeyWeeUsr: but only if he'd at least see something in the console when any file is dropped, otherwise it shouldn't matter (tested with both py2 and py3).

Zen-CODE:
No, just SDL2.

examples/miscellaneous/multiple_dropfile.py
KeyWeeUsr: Ah, sorry I misunderstood. Yes, I'll rewrite it.

tshirtman:
works for me, on linux (ubuntu 16.04) with sdl2, with i3wm.

matham:
Then I guess it should be fine to merge, maybe with an additional warning saying it's only supported where drop support is working?

KeyWeeUsr:
Hm, fine then, I'll edit it in a while.

PR #4517, opened 2016-08-06
Osc fixes

inclement:
These are the hacky fixes I had to make to get osc working with python3 - it seems that mostly it didn't understand bytes at all. I also have to pass `typehint='b'` to osc.sendMsg, and make everything bytes explicitly.

This probably shouldn't be merged, but maybe it can help anyone else having python3 problems, and serve as the base for a more robust fix.

### DIFF
--- a/kivy/lib/osc/OSC.py
+++ b/kivy/lib/osc/OSC.py
@@ -196,7 +196,7 @@ def OSCBlob(next):
     """Convert a string into an OSC Blob,
     returning a (typetag, data) tuple."""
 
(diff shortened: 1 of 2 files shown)


dessant:
@tshirtman, could you guide this to a merge/close?

tshirtman:
i did a more robust fix some time ago, in pr #5716

tshirtman:
closing since kivy is now using oscpy for tuio, and the osc package will be removed in the future.

PR #4518, opened 2016-08-07
Fix input_filter for direct changing

KeyWeeUsr:
I have encountered a problem with `input_filter` which isn't that much obvious, but may be crucial for unittesting the input - I mean explicitly changing `textinput.text` through TextInput's instance. This fixes it, although I'm not 100% sure if it's the right approach.

Example:
```
# -*- coding: utf-8 -*-
import re
from kivy.lang import Builder
Builder.load_string('''
<Test>:
    TextInput:
        id: input
        input_filter: 'float' #root.simple_chars
    Button:
        on_release: input.text = '1Ťéşt 1Ťéşt1.2'
''')
class Test(BoxLayout):
    @staticmethod
    def simple_chars(substring, from_undo):
        chars = re.findall(r'([a-zA-Z0-9.])', substring)
        return u''.join(chars)
runTouchApp(Test())
```

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -2893,6 +2893,26 @@ class TextInput(FocusBehavior, Widget):
         return text
 
     def _set_text(self, text):
+        mode = self.input_filter
+        if mode is not None:
+            if text is bytes:
+                int_pat = self._insert_int_patb
+            else:
+                int_pat = self._insert_int_patu
+
+            if mode == 'int':
+                text = re.sub(int_pat, '', text)
+            elif mode == 'float':
+                if '.' not in text:
+                    text = re.sub(int_pat, '', text)
+                else:
+                    text = '.'.join([re.sub(int_pat, '', k) for k
+                                          in text.split('.', 1)])
+            else:
+                text = mode(text, from_undo=False)
+            if not text:
+                return
+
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
Bad, bad, bad! Cripples RecycleView's refreshing from data.

matham:
Also, input_filter is meant to control user input, not programmatical input. So I don't think it's rules should be applied when assigning to text anyway.

KeyWeeUsr:
@matham what about inserting the text into TextInput e.g. from a file - kind of like template? It'll break user's app if someone builds the logic in this way.

matham:
Not sure what you mean by template. And if a user can insert text (by selecting a file which is then loaded?) then it's the programmer's responsibility to filter the text. input_filter is just a convenience for front facing interactions (i.e. by end user), it is not meant to be a generic way to filter text. The programmer is still responsible for making sure invalid text doesn't get assigned there in other ways.

KeyWeeUsr:
I meant e.g. putting .spec-like file (template) into TextInput, which would be later edited by the end user or something like that. Anyway, you're right that programmer is responsible, but I thought this might be an easy way to make it simplier for a dev (built-in filter for text), however I see that it could do more damage than help.

PR #4522, opened 2016-08-09
carousel scrollview children touch_move  fix

Bakterija:
Current carousel has to break scrollview scrolling because it has to try scrolling itself before dispatching scrolling to children.
My solution is to add new "do_scroll" variables (just like scrollview already has) and use those to switch off carousel's own scrolling when x or y "scroll_distance" threshold crossed and "do_scroll_x/y" is False.

### DIFF
--- a/kivy/uix/carousel.py
+++ b/kivy/uix/carousel.py
@@ -523,7 +523,7 @@ class Carousel(StencilView):
             'time': touch.time_start}
         self._change_touch_mode_ev = Clock.schedule_once(
             self._change_touch_mode, self.scroll_timeout / 1000.)
-        self.touc_mode_change = False
+        self.touch_mode_change = False
         return True
 
     def on_touch_move(self, touch):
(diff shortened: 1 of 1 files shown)


Bakterija:
With this I am also including an example app that demonstrates the difference    
https://github.com/Bakterija/push_examples/tree/master/kivy_4522%20carousel%20scrollview%20children%20touch_move%20fix

tshirtman:
Ok, the idea is right, i think, but i was seeing the API a bit differently, since carousel already has a `direction` property and it doesn't make sense to ignore swipes in the direction your carousel is used, right? (or one would simply disable it), i think the only new thing needed is a `BooleanProperty` telling to ignore/avoid touches in the other direction, but i'm not sure how to name it.

`ignore_opposite_swipes`
`ignore_opposite_scroll`
`avoid_other_direction_touches`
`avoid_perpendicular_swipes`
`avoid_other_swipes`

?

Then changing on_touch_move's condition to use `(self.direction in 'top', 'bottom' and self.avoid_other_swipes)` for the first condition, and same with 'left', 'right'.

Bakterija:
adjustments were made

tshirtman:
looks good to me :+1:

tshirtman:
merged in 17319bf9ee00935a25e174e41505a7c9970d8958

Issue #1570, opened 2013-10-19
arabic text input


nim4n:
how can I convert text in typing time? http://postimg.org/image/u40ftpwm9/  http://postimg.org/image/ibefw8b89/ as you can see when I type in kivy the word is out of shape so I used "arabic reshaper" to fix this problem. now I can use convert button to fix It but how should I fix this in typing time without pressing convert button?
##

sajadbanooie:
for textinput it wont work

sajadbanooie:
def _create_line_label(self, text, hint=False):
        reshaped_text = arabic_reshaper.reshape(text)
        text = get_display(reshaped_text)
adding it to TextInput will get:
 Traceback (most recent call last):
   File "/home/sajad/PycharmProjects/kimiastone/Persian.py", line 112, in <module>
     myApp().run()
   File "/usr/local/lib/python2.7/dist-packages/kivy/app.py", line 824, in run
     runTouchApp()
   File "/usr/local/lib/python2.7/dist-packages/kivy/base.py", line 487, in runTouchApp
     EventLoop.window.mainloop()
   File "/usr/local/lib/python2.7/dist-packages/kivy/core/window/window_pygame.py", line 401, in mainloop
     self._mainloop()
   File "/usr/local/lib/python2.7/dist-packages/kivy/core/window/window_pygame.py", line 288, in _mainloop
     EventLoop.idle()
   File "/usr/local/lib/python2.7/dist-packages/kivy/base.py", line 327, in idle
     Clock.tick()
   File "/usr/local/lib/python2.7/dist-packages/kivy/clock.py", line 483, in tick
     self._process_events()
   File "/usr/local/lib/python2.7/dist-packages/kivy/clock.py", line 615, in _process_events
     event.tick(self._last_tick, remove)
   File "/usr/local/lib/python2.7/dist-packages/kivy/clock.py", line 374, in tick
     ret = callback(self._dt)
   File "/usr/local/lib/python2.7/dist-packages/kivy/uix/textinput.py", line 1512, in _refresh_line_options
     self._refresh_text_from_property()
   File "/usr/local/lib/python2.7/dist-packages/kivy/uix/textinput.py", line 1531, in _refresh_text_from_property
     self._refresh_text(self._get_text(encode=False), *largs)
   File "/usr/local/lib/python2.7/dist-packages/kivy/uix/textinput.py", line 1551, in _refresh_text
     _line_rects.append(Rectangle(size=lbl.size))
[...]

PR #4524, opened 2016-08-09
Supporting multiline, Right-To-Left Persian/Arabic text rendering, via setting the rtl option for text

pylover:
This is just a snapshot of my works,

So, please review and say something about it.

The next step is to fix textinput cursor.

All comments and suggestions are appreciated.


pylover:
Refer to: #1570

akshayaurora:
- The way you have it implemented right now, the text can be either rtl or ltr.

The library that you use does support bi-directional features, that is we can have both kind of text in the same line. 

There is no need for the `rtl` property,  the text input should dynamically support any language based on input not based on how it is programmed.

This is the reason, this goes along with another topic, that is of font Glyph Fallback, 

- You are specifying arabic,  while we need to be able to support reshaping for all of the languages, not just for arabic.

pylover:
Currently, i am searching for a library for reshaping to support all languages and cultures,

About rtl property, if so, we have a little process cost to detect language and direction. is it ok?

Or we can use a global config value?

pylover:
In addition. most text rendering tools has an attribute about direction.

pylover:
it seems the fribidi is enough

tshirtman:
Hey, is this in a mergeable state? :)

pylover:
No. Its just for review and comment about it. Its not mergable.
The fribidi complex text layout is supported. But its unstable
On Aug 22, 2016 5:55 AM, Gabriel Pettier <notifications@github.com> wrote:Hey, is this in a mergeable state? :)

pylover:
Current problems:
1. After each space the cursor goes forward for one character.
2. kivy requires rtl now. should it be in setyp.py requirments?

motad333:
hellow
i am begin to learn python and kivy
when i try to use your class some erorr happen and i cant fix it
No module named globalization
and when try to find it in core.text nothing not found 
what is it?

pylover:
Checkout the `monkeypatch` branch from my fork.

motad333:
hi again
i am very begin to python and when check your work was not found any file with monkeypath
what is easiest way to persian in textinput in kivy?
thank for answer

KeyWeeUsr:
@pylover what is the license of `.ttf` files? I don't feel right about including random files to the repository.

pylover:
@KeyWeeUsr, yes, this files are just for test, before, i mentioned to do not merge this pull request. this is only a try for rtl.

engmms:
Hi pylover ,, Are you reach to solve arabic fonts problem in Python+Kivy App

pylover:
@engmms, Yes, it's working partially, Needs some work about multiline-text selection and cursor location,

Unfortunately, I'm very busy at this time, but i'll continue a few months later.

engmms:
How I can use it now,,as I build project and just I want to change
labels and buttons to appear with Arabic fonts

pylover:
Just clone and build my fork

eteamin:
@pylover Would you consider moving your pull request towards mergable state? thank you

pylover:
Rebasing this pull request would not be a good idea. I think a from scratch try is so better, because the mainstream is changed a lot.

PR #4525, opened 2016-08-09
[WIP] Asyncio support

tshirtman:
This changeset allows to make use of asyncio Tasks and coroutines in kivy programs, and contains a little example doing so.
(see examples/kivy_asyncio.py)

Right now it's slightly hacky and only tested with WindowSDL.

The issue is that currently, and since the first kivy commit, the mainloop is (or at least can be) actually driven by the Window instead of the EventLoop, so calling EventLoop.idle() is actually not enough to have a fully functional app, so `KivyEventLoop` (asyncio event loop implementation for kivy) directly calls EventLoop.window._mainloop(), instead, it would be better if EventLoop.idle() called its Window step function (be it called _mainloop or whatever).

I didn't check, but since Window runs its `mainloop()` function, and KivyEventLoop also calls `_mainloop()`, the workload is probably doubled to create a frame.

for this reason, it could be interesting to revamp the system to make Window the worker for EventLoop, not the opposite.

Testable, but probably not mergeable in this state.

tested with python3.5.2 on linux, but should work with any python version with asyncio.

### DIFF
--- a/examples/asyncio/http_client.py
+++ b/examples/asyncio/http_client.py
@@ -1,0 +1,134 @@
+from kivy.base import KivyEventLoop
+from kivy.app import App
+from kivy.lang import Builder
+from kivy.properties import NumericProperty, BooleanProperty
+from kivy.clock import Clock
+from random import sample
+from string import printable
+from time import sleep
+import asyncio
+import aiohttp
+import logging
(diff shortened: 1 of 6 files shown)


tshirtman:
closing in favor of #5241

PR #4530, opened 2016-08-11
Update installation-linux.rst

user5061:
The python3 sentence should be higher, closer to the dependencies it refers to. 

The details about default virtualenvs ( which exist in the other section about virtualenvs) are now included in this virtualenv-section as well.


doc/sources/installation/installation-linux.rst
```diff
@@ -175,6 +175,7 @@ Kivy       Cython
 Ubuntu example
 --------------
 
+In the following command use "python" and "python-dev" for Python 2, or "python3" and "python3-dev" for Python 3.
```
thopiekar: Shouldn't it be ```In the following commands [...]```?

PR #4532, opened 2016-08-12
uix:TextInput convert hint_text to unicode only if string isn't already unicode

BabyWolf:
Testing with the following code:

```py
#-*- coding: utf-8 -*-
from kivy.app import App
from kivy.lang import Builder

TextInput:
    hint_text: 'привіт світ'
    text: ''
''')

class TestApp(App):
    def build(self):
        return root


if __name__ == '__main__':
    TestApp().run()
```


matham:
Your example code works fine for me in py2 and 3. Also, I'm I don't think `u' ,\'".;:\n\r\t'` will ever cause an exception.

PR #4534, opened 2016-08-13
Removed long , annoying tracebacks

Bakterija:
i get long tracebacks when loading video files with kivy
removed traceback, replaced with Image: Error loading texture {filename}'..

currently http://i.prntscr.com/19310636b6f348f292d2563144a29478.png

after http://i.prntscr.com/abfcbf20f638458190ddf6ad58a31b34.png

### DIFF
--- a/kivy/uix/image.py
+++ b/kivy/uix/image.py
@@ -260,7 +260,8 @@ class Image(Widget):
                                                  keep_data=self.keep_data,
                                                  nocache=self.nocache)
             except Exception as e:
-                Logger.exception(e)
+                Logger.error('Image: Error reading file {filename}'.
+                                    format(filename=self.source))
                 self._coreimage = ci = None
 
             if ci:
--- a/kivy/uix/image.py
+++ b/kivy/uix/image.py
@@ -260,7 +260,7 @@ class Image(Widget):
                                                  keep_data=self.keep_data,
                                                  nocache=self.nocache)
             except Exception as e:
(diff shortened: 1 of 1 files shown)


Cheaterman:
No! This removes the traceback when trying to load a thumbnail for the video! This traceback is normal and to be expected!

matham:
I think I may agree with @Cheaterman on this.

KeyWeeUsr:
@matham @dessant @Cheaterman So... what's going to happen with this PR?

Cheaterman:
@KeyWeeUsr from what I understand, the PR stays, and Kivy's architecture needs to evolve?

dessant:
Yeah, it's worth noting the history of these changes.

We went from silently eating the exception to logging it (https://github.com/kivy/kivy/pull/4510), which surfaced a bug in Video (https://github.com/kivy/kivy/issues/4535).
People did not like the traceback so the logging method was changed in this pr, and in the end we arrived at the conclusion that there should be an on_error event, moved from AsyncImage to Image (discussed in https://github.com/kivy/kivy/pull/4549), and the exception could be passed with the event, giving the user liberty to log and handle the error.

KeyWeeUsr:
Thanks, just checking if there isn't something bad hanging around while putting together the release notes. 🐱

matham:
I think an on_error approach is bad design. But anyway, I guess it's not urgent for this release.

PR #4549, opened 2016-08-22
remove exception variable that is not used anymore

Bakterija:


dessant:
From here we could evolve to have an on_error event for Image, though starting to log the error in some form is already much better. ref https://github.com/kivy/kivy/pull/4510/files#diff-5febd68249e83301b81850ca856b0969R263

dessant:
And it seems we already have one, it just needs to be used https://github.com/kivy/kivy/pull/4545/files#diff-5febd68249e83301b81850ca856b0969R335.

Issue #4353, opened 2016-06-02
DropDown not affected by max_height


Leva7:
`DropDown`'s property `max_height` doesn't seem to do anything to the size of the drop-down. Code and more explanation can be found in [this Stack Overflow question](http://stackoverflow.com/questions/37573325/kivy-dropdown-not-affected-by-max-height)
Kivy version: 1.9.2-dev0

rafalo1333:
It's still present on master and very annoying. Does someone knows how we can fix it?

tshirtman:
From what i understand, max_height is applied to the wrong thing, the container is the GridLayout, that contains the items, not the ScrollView that contains it, it makes no sense to try to change the size of the list of item, what one might wants to limit is the ScrollView that displays it, so imho the `_container_minimum_size` (https://github.com/kivy/kivy/blob/69f6bd89a18602e0b91df8b7c97b3b642d87413f/kivy/uix/dropdown.py#L278-L288) is not needed at all, what needs to be done is use `max_height` in `_reposition`, probably around here. https://github.com/kivy/kivy/blob/69f6bd89a18602e0b91df8b7c97b3b642d87413f/kivy/uix/dropdown.py#L349-L368).

tshirtman:
please test the pr #4550, i do think it solves the issue.

rafalo1333:
@KeeWeeUsr Can be closed as @tshirtman fixed the issue.

PR #4550, opened 2016-08-23
proposed fix for #4353 DropDown.max_height not working

tshirtman:


rafalo1333:
Good work, @tshirtman. After the fix, `max_height` works, but the new problem appeared. When the window height is very small (with max_height set), or when it's height is smaller than container's `minimum_height` (when `max_height` is None) the dropdown refuses to open or just disappears (f.e. when we resize the window to be small in height). This behavior is easy to see with Spinner, but also with module example. I tried to search for a fix, but with no success. The bug is somewhere in _reposition, for sure.

tshirtman:
hm, i though i had tested that, but will look again later today, hopefully.

tshirtman:
Should be ok now :)

rafalo1333:
Great work @tshirtman , works like a charm! Ready to merge.

tshirtman:
great, merging then :)

PR #4558, opened 2016-08-25
Separate image and button in Switch

KeyWeeUsr:
Puts the Button in the right place. If they are on "the same" canvas, it's hard to customize the widget, however when they are separated, widget becomes highly customizable even if it's in a state that requires image(s) as the background.

Example:

```
from kivy.lang import Builder
Builder.load_string('''
<Custom@Switch>:
    canvas:
        Color:
            rgb: 0.2, 0.709, 0.898, 1
        Rectangle:
            size: [sp(41.5), sp(20)]
            pos: [self.center_x - sp(41.5), self.center_y - sp(10)]
        Color:
            rgb: 0.4, 0.4, 0.4, 1
        Rectangle:
            size: [sp(41.5), sp(20)]
            pos: [self.center_x, self.center_y - sp(10)]
    Label:
        text: '[b]OFF[/b]'
        markup: True
        font_size: 13
        pos: [root.center_x - sp(70), root.center_y - sp(50)]
    Label:
        color: 0.75, 0.75, 0.75, 1
        text: '[b]ON[/b]'
        markup: True
        font_size: 13
        pos: [root.center_x - sp(30), root.center_y - sp(50)]

<Test>:
    Widget:
        canvas:
            Color:
                rgb: 0.2, 0.709, 0.898, 1
            Rectangle:
                size: 800, 800
                pos: [0,0]
    Custom:
    Switch:
    Widget:
        # change pos to [0,0] to see
        canvas:
            Color:
                rgb: 0.2, 0.709, 0.898, 1
            Rectangle:
                size: 800, 800
                pos: [500,0]
            
''')
class Test(BoxLayout): pass
runTouchApp(Test())
```

First widget demonstrates that separating doesn't cripple `canvas`, nor `canvas.after`, because it's added as a child sooner. The second one demonstrates the same for `canvas.before`.

As some of my PRs were closed because of theming-like content(ref https://github.com/kivy/kivy/pull/4296#issuecomment-231480546), I'm kind of worried, however, as I don't see any sane reason why this should be bad(correct me if I'm wrong), I think it may be a good change, because it introduces new possibilities e.g. to cover the text on the widget.

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -472,6 +472,9 @@
             source: 'atlas://data/images/defaulttheme/switch-background{}'.format('_disabled' if self.disabled else '')
             size: sp(83), sp(32)
             pos: int(self.center_x - sp(41)), int(self.center_y - sp(16))
+    canvas.after:
+        Color:
+            rgb: 1, 1, 1
         Rectangle:
             source: 'atlas://data/images/defaulttheme/switch-button{}'.format('_disabled' if self.disabled else '')
             size: sp(43), sp(32)


udiboy1209:
This isn't bad but it doesn't look as a good API if I want to modify the switch button. I think most use cases that would require to change the background would also require to change the button image. Sure I could do it by adding another rectangle to `canvas.after` but then it just becomes recreating the Switch widget myself.

Also, is there some problem in providing a way to modify the image source directly? and making the `pos` equation dynamic on size?

P.S. wont hurt to merge this though.

KeyWeeUsr:
Don't get me wrong, it's not supposed to _be_ API for modifying the Switch widget. It's only supposed to make the problem easier(for now at least) to handle mainly because the widget is ~90% image with a fixed text(chars, font, size, place), which is _bad_. Especially when your app's language isn't english, because in mobile apps this widget(don't confuse with layouts) is probably the most common after Buttons and Labels.

And also it kind of jumped on my mind when I saw solutions of customizing the widget with making own image, which I think is stupid because the rest of Kivy theme works good and is highly customizable except this black spot. :smile: 

I'm going to rewrite Switch soon to be a widget that's not dependent on an image(or not fully at least), but until it's ready this could be the way (as I intend to separate the layers anyway) :smile:

udiboy1209:
I agree to all of that! What I meant by the poster equation breaking down is that you may need to rewrite it for different size of the background. And yes Switch having text as a background image is bad in itself and can lead to bad practices being followed.  It's good that you will be rewriting this widget.

For now this change is good and it works, so merging it! Thanks!

PR #4559, opened 2016-08-26
Fix scrolling in TabbedPanel

KeyWeeUsr:
`switch_to(tab, do_scroll=False)` is the behavior until this patch. This introduces `do_scroll` keyword, which if set to True(by default), scrolls the tab list to the header of the tab TabbedPanel switches to.

### DIFF
--- a/kivy/uix/tabbedpanel.py
+++ b/kivy/uix/tabbedpanel.py
@@ -487,8 +487,8 @@ class TabbedPanel(GridLayout):
         self._setup_default_tab()
         self.switch_to(self.default_tab)
 
-    def switch_to(self, header, do_scroll=True):
-        '''Switch to a specific panel header.
+    def switch_to(self, header, do_scroll=False):
+        '''Switch to a specific panel header. If used with `do_scroll=True`, it scrolls to the header's tab too.
         '''
         header_content = header.content
         self._current_tab.state = 'normal'
--- a/kivy/uix/tabbedpanel.py
+++ b/kivy/uix/tabbedpanel.py
@@ -487,7 +487,7 @@ class TabbedPanel(GridLayout):
         self._setup_default_tab()
         self.switch_to(self.default_tab)
 
-    def switch_to(self, header):
+    def switch_to(self, header, do_scroll=True):
         '''Switch to a specific panel header.
         '''
         header_content = header.content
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
Better this way, as it throws errors if it scrolls by default (and will break api for versions older than this patch)

kivy/uix/tabbedpanel.py
```diff
@@ -487,8 +487,8 @@ def __init__(self, **kwargs):
         self._setup_default_tab()
         self.switch_to(self.default_tab)
 
-    def switch_to(self, header):
-        '''Switch to a specific panel header.
+    def switch_to(self, header, do_scroll=False):
+        '''Switch to a specific panel header. If used with `do_scroll=True`, it scrolls to the header's tab too.
```
akshayaurora: pep8!

kivy/uix/tabbedpanel.py
```diff
@@ -503,6 +503,10 @@ def switch_to(self, header):
             parent.remove_widget(header_content)
         self.add_widget(header_content)
 
+        if do_scroll:
+            tabs = self._tab_strip
+            tabs.parent.scroll_to(header)
```
akshayaurora: let's add animation too!

KeyWeeUsr: @akshayaurora what's with the animation? It scrolls as casual ScrollView, doesn't it?

akshayaurora: I guess it's true by default so no need to specify it.

kivy/uix/tabbedpanel.py
```diff
@@ -487,8 +487,9 @@ def __init__(self, **kwargs):
         self._setup_default_tab()
         self.switch_to(self.default_tab)
 
-    def switch_to(self, header):
-        '''Switch to a specific panel header.
+    def switch_to(self, header, do_scroll=False):
+        '''Switch to a specific panel header. If used with
+        `do_scroll=True`, it scrolls to the header's tab too.
```
dessant: A versionchanged tag could be added for the new arg.

KeyWeeUsr:
Ah, I can't edit after merge. O.o

Issue #3572, opened 2015-08-11
TreeView: `select_node(None)` should unselect the selected node


lmNt:
As the title suggests, I think calling `select_node(None)` should reset the selection in a TreeView and not crash.

Currently, at least if I am correct, it is not possible to unselect a selected node in a TreeView in a valid/clean way, i.e. without setting the nodes' `is_selected` property to false.

So instead of doing this:

    tree_view.get_selected_node().is_selected = False

I think

    tree_view.select_node(None)

should do the job.
Correct me, if I am missing something here.

KeyWeeUsr:
@lmNt Implemented clean method for deselecting, closing. :yup:

PR #4561, opened 2016-08-26
Add TreeView.deselect_node()

KeyWeeUsr:
This solves the issue #3572 but I'm not quite sure if it's a good idea. On the other hand, it could be tweeked a little bit more to `node=None` by default and that way it could prevent crashing an app.

### DIFF
--- a/kivy/uix/treeview.py
+++ b/kivy/uix/treeview.py
@@ -345,7 +345,16 @@ class TreeView(Widget):
 
     def select_node(self, node):
         '''Select a node in the tree.
+
+        .. versionchanged:: 1.9.2
+
+        If ``None`` is used as a value for `node` and there's
+        already a selected node, it will unselect it.
         '''
+        if not node:
+            if self._selected_node:
+                self._selected_node.is_selected = False
+            return
         if node.no_selection:
             return
         if self._selected_node:
--- a/kivy/uix/treeview.py
+++ b/kivy/uix/treeview.py
@@ -345,22 +345,21 @@ class TreeView(Widget):
 
     def select_node(self, node):
         '''Select a node in the tree.
-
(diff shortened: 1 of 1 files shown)


matham:
I'm not sure I like the idea of simply leaving node to None in a select function implying something about deslecting. Why would it not work to deselect directly? Also, that kind of thing makes sense when touching a node, i.e. one touch selects, the next deselects. But not in underlying api code.

KeyWeeUsr:
It's not about None directly, but it deselects only if there's something other selected if I'm right. I didn't find a deselecting function and I used already working function to deselect, which in the beginning(everything deselected), the variable holding node is equal to None.

Use case would be deselecting everything with clock maybe.

matham:
Right, the problem is with the the implicit behavior that when it's None and there's something selected, it deselects it. This method is called select not toggle so that  behavior would be very unexpected. The method currently on deselects when a new one is selected instead which is expected.

While there isn't a deselect method, there really should be, with the same behavior you're suggesting to deselect the current one if node is None (in that method).

KeyWeeUsr:
Ah, got it! Will look into that soon.

matham:
Just a small comment, I think that method should have a node parameter to be used to delect that node. Only if the node param is None should we deselect the currently selected node.

KeyWeeUsr:
Well, I tried to select multiple nodes with `shift` or `ctrl`, but no succes there, so I guess it's not really possible from user's side. Also `select_node()` always deselects the old node, so that wouldn't work too.

There's is however a possibility to select multiple nodes with passing `is_selected=True`, but this way a user can't deselect by clicking on some other node. A click on the selected node is necessary to "select" it once more and then it deselects. (probably bug anyway)

PR #4563, opened 2016-08-27
Draw animation every frame by default, use step=0 instead of 1 / 60.0

rafalo1333:
Animation should be as smooth as possible by default. Drawing animation on every frame will be the best default to provide rich, smooth user experience. As discussed with @tshirtman, Config.graphics.maxfps should limit that if set, and the user would have the option to degrade performance by using other kwarg parameter value.


dessant:
Thanks @rafalo1333, I don't see a drawback to changing this either.

Issue #4338, opened 2016-05-30
PDF doc builds fail on travis


dessant:
The build has been disabled until a fix is found. The issue does not stem from a recent kivy commit, so it's likely a change in one of the dependencies needed for the build.
https://github.com/kivy/kivy/pull/4337
https://travis-ci.org/kivy/kivy/jobs/133713221

KeyWeeUsr:
If you have some list of packages in a file, I may try to look for what has been updated in 28th - 29th May, because that's where it crashed and if I find it, a simple `==<version` could fix the pdf. Depends on how long it'll take to port the docs to RTD i.e. if it makes sense to even try to fix it, or just let it go, hehe.

dessant:
Kivy does compile on their backend now and the docs are generated, but for some reason the api reference is left out both from the html and pdf version, html styling worked for the first builds, but now that is also broken.

https://kivy.readthedocs.io/en/latest/

dessant:
Fixed in https://github.com/kivy/kivy/pull/4566, pdf builds are now managed by RTD.

Issue #1197, opened 2013-05-13
Make documentation available by version (pdf/html)


tshirtman:
Currently, the doc, html and pdf versions, refer to the latest master, while this is most of the time convenient, for various reasons, we should make at least the last stable version available, and possibly older versions too.

Now, what version should be the default one, and if it's the stable, should we push the doc fixes to the older versions (and regenerate it) when the doc appear to be wrong?

KeyWeeUsr:
What about making https://stable.kivy.org/docs just for copy of stable parts from the latest commit of stable version? You could basically go for the same approach that is used now for docs building, but changing git source to the release package and make another subdomain(if it's not money-eater).

dessant:
Fixed in https://github.com/kivy/kivy/pull/4566, after the next release both stable and master will be available on RTD.

PR #4566, opened 2016-08-28
Fix docs for RTD

KeyWeeUsr:
* Removed old `AUTOBUILD_STAMP`, autobuild.py now works as import in conf.py(RTD ignores such commands in Makefile). Check for the touch file still present, but filename changed to `autobuild.py-done` as it sounds simpler and I haven't found other checks for the specific filename + getting exact python from env shouldn't be necessary.

* `autobuild.py` has as default behavior set `BE_QUIET=True` as this it when the docs are built by default with make locally.

* touch `autobuild.py-done` is done within python, which removes my ugly fix for Win ^^

* Some pep8 fixes of `conf.py`

* `templates_path = ['.templates']` is required for local build of docs from kivy.org, however the underscore version is requirer on RTD

* Implemented check for RTD and if, choose their theme - leaves local build's behavior alone.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -39,8 +39,8 @@ else
 	_TESTS = _TEST
 endif
 
-AUTOBUILD_STAMP_BASE = autobuild.py-done-with-
-AUTOBUILD_STAMP      = "$(AUTOBUILD_STAMP_BASE)$(PYTHON)"
+# Created in autobuild.py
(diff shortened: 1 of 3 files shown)


KeyWeeUsr:
ref #4338
ref #1197 (in RTD it leaves the old versions there, so it's just the matter of right settings)

dessant:
LGTM, thanks @KeyWeeUsr!

PR #4569, opened 2016-08-28
WM_Pointer to get pressure on Windows

steveway:
This isn't yet tested since I haven't gotten kivy to compile yet, haven't set that up.
But this is in principle how that works, as you can see there is more info in the structs that we can use.
This should work with both Wintab based devices like those from Wacom for example and with other ones .
It would be great if someone could test and possibly extend this.
Also the windows section of the __init__.py file in the input providers directory should be changed to something like this:
```
if platform == 'win' or 'KIVY_DOC' in os.environ:
    use_wm_pointer = False
    try:
        import kivy.input.providers.wm_pointer
        use_wm_pointer
    except:
        use_wm_pointer = False
        err = 'Input: WM_Pointer not supported by your version of Windows'
        Logger.warning(err)
    if use_wm_pointer == false:
        try:
            import kivy.input.providers.wm_touch
            import kivy.input.providers.wm_pen
        except:
            err = 'Input: WM_Touch/WM_Pen not supported by your version of Windows'
            Logger.warning(err)
```
With this it should only try to use if WM_Pointer doesn't work.

### DIFF
--- a/kivy/input/providers/wm_pointer.py
+++ b/kivy/input/providers/wm_pointer.py
@@ -1,0 +1,172 @@
+'''
+Support for WM_POINTER messages (Windows platform)
+==============================================
+'''
+
+__all__ = ('WM_PointerProvider', 'WM_Pointer')
+
+import os
+from kivy.input.providers.wm_common import *
+from kivy.input.motionevent import MotionEvent
+from ctypes.wintypes import *
+
+GET_POINTERID_WPARAM = LOWORD
+GetPointerPenInfo = windll.user32.GetPointerPenInfo
+
+(PT_POINTER,
+PT_TOUCH,
+PT_PEN,
+PT_MOUSE,
+PT_TOUCHPAD)=map(c_int,range(1,6))
+
+WM_POINTERUPDATE = 0x0245
+WM_POINTERDOWN = 0x0246
+WM_POINTERENTER = 0x0249
+WM_POINTERLEAVE = 0x024A
+WM_POINTERUP = 0x0247
+
+
+class POINTER_INFO(Structure):
+    _fields_ = [
+        ('pointerType', c_int),
+        ('pointerId', c_uint32),
+        ('frameId', c_uint32),
+        ('pointerFlags', c_uint32),
+        ("sourceDevice", HANDLE),
+        ("hwndTarget", HWND),
+        ("ptPixelLocation", POINT),
+        ("ptHimetricLocation", POINT),
+        ("ptPixelLocationRaw", POINT),
+        ("ptHimetricLocationRaw", POINT),
+        ("dwTime", DWORD),
+        ('historyCount', c_uint32),
+        ('inputData', c_int32),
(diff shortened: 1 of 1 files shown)


steveway:
Ok, I've changed a few things and made some small fixes.
I've also compiled and tested it, and nothing happens.
It seems that Kivy does not even call the start method of WM_PointerProvider.
Not sure what is wrong, if someone has any idea, I would be glad if you chimed in.

kivy/config.py
```diff
@@ -709,6 +709,7 @@ def name(self, value):
             # from 1.0.9, don't activate mactouch by default, or app are
             # unusable.
             if platform == 'win':
+                Config.setdefault('input', 'wm_pointer', 'wm_pointer')
```
matham: This needs to be added and documented with the other config variables in kivy/__init__.py (I believe).

matham:
I'd suggest you compare how/when kivy calls start in the other wm_xxx modules to see why it isn't calling yours.

steveway:
Hello matham, sorry I forgot to clarify that the current version that I have here does now indeed start for me.
I also already get the pressure information.
I currently need to adjust the x and y coordinates to the application window coordinates.
I have a multi-monitor setup, where the right one is my main monitor and the left one is a Wacom Cintiq.
With the current code, the x coordinates go from -1980 (left edge of the Cintiq) to +1980 (right edge of monitor).
And the y coordinate goes from 0 (top edge) to 1080 (bottom edge).
Problem 1 with this is, that I'm not using the x and y coordinates inside the structs since these now contain bogus values.
Problem 2 is that we need to find a way to map these coordinates to the current kivy window coordinates.
It's getting there, but it's not quite done.

Also, in the TouchTracer example, what is this line supposed to do exactly:
`#if not .95 < (touch.pressure / ud['pressure']) < 1.05:`
There is no check done if ud['pressure'] is 0, so it seems to crash.
If you uncomment the lines 89 to 93 in my wm_pointer.py file then all input devices, like mice and touchscreens/pads and similar, should only produce wm_pointer events.
For mice the pressure stays of course at 0 and I can pretty reliably crash the TouchTracer demo with that.
Actually, looking at this now, it would make sense if ud['pressure'] would contain the maximum pressure value. But I can't find where for example linuxwacom or androidjoystick set these values.

msprengholz:
Is there any update regarding this feature? I'd really like to use pressure from pen events on Windows, but it seems to still not support this profile.

steveway:
I hadn't had time to work on this myself.
I had hoped that someone with more knowledge about kivys eventsystem would take a look at this.
In the meantime I had opened a similar bug/feature-request in the SDL Bugtracker.
https://bugzilla.libsdl.org/show_bug.cgi?id=3386
Sam Lantinga seemed to be interested in that and Icculus also is interested but mentioned that he doesn't have much time currently (August 2017) for that:
https://twitter.com/steveway/status/892648799524487168
I believe having this directly in SDL makes a bit more sense, but someone has to program it first either way.

PR #4571, opened 2016-08-30
Update properties docs

user5061:
[Here](https://gist.github.com/user5061/7433e8049c9f44c755180b85f6c05139) is the code to reproduce the behavior in the commit. The same can be replicated with DictProperty, ObjectProperty, and probably all properties. I only tested those 3 Kivy properties though.

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -51,6 +51,29 @@ With Kivy, you can do::
         a = NumericProperty(1.0)
 
 
+Depth being tracked
+~~~~~~~~~~~~~~
+
+Only the "top level" of a nested object is being tracked. For example:: 
+
+    my_list_prop = ListProperty([1, {'hi': 0}])
+    # Changing the any top level element will trigger all `on_my_list_prop` callbacks
+    my_list_prop[0] = 4  
+    # Changing any deeper element will be ignored by all `on_my_list_prop` callbacks
+    my_list_prop[1]['hi'] = 4 
+
+Another example::
+
+    class A(object):
+        pass
+    
+    instance_a = A()
+    my_obj_prop = ObjectProperty(instance_a)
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
May seem strange, but it's kind of obvious, because you change a value of a different object(the dictionary) and the check is only set for the object that is instance of `ListProperty`, which inherits from EventDispatcher.

I wonder if it behaves the same e.g. for `ListProperty([0, ListProperty([1, 2])])`

user5061:
I initially incorrectly assumed that properties create a deepcopy of the object. Perhaps #4571 should be used instead of this PR.

rafalo1333:
Properties must be declared at class level, I do not think KeyWeeUsr example is possible to produce with Kivy.

KeyWeeUsr:
@user5061 Oh, I thought I was alone with that error. Anyway, I was only curious if that'd work. If, then maybe this way it could store an additional info, which could result in making ping from nested Properties possible e.g.:

```
p = ListProperty([0, 1])
p[1] = ListProperty([1, 2], parent=p)
```

Otherwise, I think the addition to docs is good, because it probably isn't mentioned there and will spare troubles for someone thinking about why this or that don't work, so... I'd merge it, though I don't know if I can :smile:

matham:
I'd say that the top part is good, but I'd remove everything after `Another example::`. From there on it just adds an example that is not recommended (will it even work?) usage of properties and is already clear by the first example.

I'd also change `The same holds true for all kivy properties.` to `The same holds true for all container type kivy properties.` For `ObjectProperty` e.g. even the top level will not be tracked. So it only applies for list/dict properties.

user5061:
@matham Ah yes, there is already a "Warning"-section in ObjectProperty docs, which makes that part of the PR redundant. What's the procedure when a PR should be changed before pulled? Should I change it?

PR #4573, opened 2016-08-30
Add clipboard example

KeyWeeUsr:
(...and fix ugly import in docs.)

Example that makes use of clipboard for creating widgets e.g. for pasting values from spreadsheet.

### DIFF
--- a/examples/miscellaneous/clipboard.py
+++ b/examples/miscellaneous/clipboard.py
@@ -1,0 +1,64 @@
+from kivy.lang import Builder
+from kivy.uix.label import Label
+from kivy.base import runTouchApp
+from kivy.uix.boxlayout import BoxLayout
+
+import os
+
+Builder.load_string('''
+#:import Clipboard kivy.core.clipboard.Clipboard
+<Clip>:
+    orientation: 'vertical'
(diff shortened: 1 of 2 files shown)


rafalo1333:
@KeyWeeUsr actually removing this ugly import would KV from docs crashing ;) Clipboard is not uix class and its not available in KV namespace, so without this import the code would not run.

KeyWeeUsr:
@rafalo1333 sure:

```
from kivy.lang import Builder
Builder.load_string('''
#:import Clipboard kivy.core.clipboard.Clipboard
<Test>:
    Button:
        on_release:
            self.text = Clipboard.paste()
            Clipboard.copy('Data')
''')
class Test(BoxLayout): pass
runTouchApp(Test())
```

rafalo1333:
Its working when Clipboard is imported. When you remove it, after button release app crashes with the obvious error: 

```
  File "<string>", line 5, in <module>
 NameError: name 'Clipboard' is not defined
```

KeyWeeUsr:
That's why I put the python import from `on_release` to the top and made it kv import. Of course if you remove it completely it'll crash. Did you see the diff btw?

PR #4578, opened 2016-09-02
Some misspellings

kwadrat:


udiboy1209:
Nice contribution! Did you use some tool for doing all that? Other project repos could use this too :grin:. 
I saw some changes affecting non-comment areas (mostly strings). I don't think the change would produce any errors but it would be better to check.

Also, not that it matters much but it would look cleaner to squash all those commits into one.

KeyWeeUsr:
@kwadrat Thanks :smile:

KeyWeeUsr:
@udiboy1209 I hope GitHub squashed them right with "Squash and merge"

kwadrat:
I have a simple Python program to generate an index of words
and then I do a lot of human reading and proofchecking using Google Search
and internet dictionaries:
------------------------------------------------------------------------------
1. Read all text files to RAM.
2. Join them.
3. Split joined text into a list of words.
4. Use set() for removing duplicates.
5. Use list() to convert set to list.
6. Sort the list.
7. Save all the words to text file.
8. Use vim text editor to read the text file ("manually" - using my own eyes) from top to bottom.
9. Look for duplicated words (these are almost identical words with a misspelling).
10. Look for strange (possibly misspelled) words.
------------------------------------------------------------------------------

I prepared another Pull-Request with a bunch of fixes.

Which projects would you like to check in the next turn?

PR #4579, opened 2016-09-03
Fix #4504

### LINKED ISSUE #4504: StackLayout children rearrange themselves unexpectedly when their parent's size changes

user5061:
Fixes issue https://github.com/kivy/kivy/issues/4504.

### DIFF
--- a/kivy/uix/stacklayout.py
+++ b/kivy/uix/stacklayout.py
@@ -248,7 +248,7 @@ class StackLayout(Layout):
             # causing unexpected children reordering when parent resizes.
             # e.g. if size is 101 and children size_hint_x is 1./5
             # 5 children would not fit in one line because 101*(1./5) > 101/5
-            if (availsize + 1e-12) >= 0 or not lc:
+            if (availsize + 1e-10) >= 0 or not lc:
                 # even if there's no space, we always add one widget to a row
                 lc.append(c)
                 sizes = testsizes
--- a/kivy/uix/stacklayout.py
+++ b/kivy/uix/stacklayout.py
@@ -245,7 +245,11 @@ class StackLayout(Layout):
                     testsizes[-1] = max(0, c.size[innerattr])
                 availsize -= testsizes[-1]
 
-            if availsize >= 0 or not lc:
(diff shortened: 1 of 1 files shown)


user5061:
According to answers and comments at [Determine how many decimal digits of a float are precise](http://stackoverflow.com/questions/39316849/determine-how-many-decimal-digits-of-a-float-are-precise), there are at least 15 significant digits, (that is, the error always occurs on the 16th significant digit or more). 

That means that significant decimals can will be less that 15. For example, `443*.2` results in a float `88.60000000000001` with 15 significant digits, 2 digits being ints and 13 being decimals. Subtracting `1e-12` fixes the issue the PR targets, in this case. 

However, if the int significant digits are 4, then subtracting `1e-12` is ineffective since the error could occur on the 11 decimal. 

Another way for `1e-12` to be ineffective is if we have multiple widgets, e.g. 10 widgets each having an error on 12 decimal, would accumulate with a total error on 11th decimal. 

**Conclusion**: 
We need a value higher than `1e-12`. In case of `10000.xxxx` then the subtraction has to occur on the 15-5=10th decimal or earlier, therefor i ll change the subtracted value to `1e-10`.

PR #4582, opened 2016-09-04
Simplify debian installation instructions and make them safer

dkg:
https://bugs.debian.org/836555 documents a cryptographic vulnerability
in the debian installation instructions.

The instructions are also out of date, and more complex than they need
to be.  Attached is a simplification for the instructions for debian,
which happens to also fix the cryptographic weakness.

### DIFF
--- a/doc/sources/installation/installation-linux.rst
+++ b/doc/sources/installation/installation-linux.rst
@@ -32,39 +32,12 @@ Ubuntu / Kubuntu / Xubuntu / Lubuntu (Saucy and above)
         $ sudo apt-get install python-kivy-examples
 
 
-Debian  (Jessie or newer)
--------------------------
+Debian
+------
 
-#. Add one of the PPAs to your sources.list in apt manually or via Synaptic
-
-    * Jessie/Testing:
-
-        :stable builds:
-            deb http://ppa.launchpad.net/kivy-team/kivy/ubuntu trusty main
-        :daily builds:
-            deb http://ppa.launchpad.net/kivy-team/kivy-daily/ubuntu trusty main
-
-    * Sid/Unstable:
-
-        :stable builds:
-            deb http://ppa.launchpad.net/kivy-team/kivy/ubuntu utopic main
-        :daily builds:
-            deb http://ppa.launchpad.net/kivy-team/kivy-daily/ubuntu utopic main
-
-        **Notice**: Wheezy is not supported - You'll need to upgrade to Jessie at least!
-
-#. Add the GPG key to your apt keyring by executing
-
-    as user:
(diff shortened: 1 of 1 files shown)


dessant:
I agree about updating the key.

The ppa docs may need to stay there though, both stable and daily packages are served through launchpad. In many cases even the stable ppa is desired over what's published in the main Debian repos, because the latter tends to be outdated.

dkg:
On Sun 2016-09-04 08:10:10 -0400, dessant wrote:
> I agree about updating the key.
>
> The ppa docs may need to stay there though, both stable and daily
> packages are served through launchpad. In many cases even the stable
> ppa is desired over what's published in the main Debian repos, because
> the latter tends to be outdated.

If you want to keep the ppa docs, it sounds like there are a few
specific things that you should do:

 0) update your ppa signing key to be something stronger than a 1024-bit
    RSA key.

 1) include a full copy of your PGP public key in the same git
    repository that holds these instructions

 2) for debian with apt version 1.3 (debian stretch and later):

     * place the key somewhere outside of `/etc/apt`, and use a
       `signed-by` directive in your proposed sources.list entry, like:


             deb [ signed-by=/usr/local/share/keyrings/kiva-archive.gpg ] http://ppa.launchpad.net/kivy-team/kivy/ubuntu trusty main

     for debian jessie:

     * place the key directly in
       `/etc/apt/trusted.gpg.d/kiva-archive.gpg`

This avoids the use of the potentially-surprising "apt-key adv
--keyserver ... --recv ..." command, and avoids any question of short
key IDs vs long fingerprints.

Regards,

    --dkg

matham:
@thopiekar can you please take a look?

thopiekar:
0: Signing keys are managed by Launchpad.net. If you are targetting a higher Ubuntu release, the signing key should be stronger. Before upgrading to Xenial or Yakkety (can't remember exactly) I also had these warnings about weak signatures.. Switching to the next release made the message disappear.
1: The key can be found on the launchpad page of the kivy team. Don't see a reason to copy it all around.
2: (Is that Debian specific?) I remember that I've added a command to add the key to apt's main keyring. Basically it isn't our intention to support Debian directly - we only have luck that it works with Ubuntu and Debian at the same time. So you should follow the instructions given by Launchpad.net how to add the PPA manually. (We actually copy/pasted their instructions and made the general instructions work with our PPAs).

Regards

thopiekar:
Oh, and using add-apt-repository won't work on Debian, afaik, because it will use 'jessie', 'sid', etc. as distribution and not the Ubuntu releases. Currently, there are no packages built for Debian.

PR #4583, opened 2016-09-04
Use consistent Lang headers for logs in kivy.lang module

rafalo1333:
…" title coming from nowhere

Logger.warning itself is a warning, no need to repeat it twice. User wants to know which part of Kivy raises such warning.

My first idea was to use "Lang" instead of "Parser" header, but for that we should use it everywhere in lang package. Ill be glad to hear what you think about it.

### DIFF
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -67,12 +67,12 @@ def custom_callback(__kvlang__, idmap, *largs, **kwargs):
 def call_fn(args, instance, v):
     element, key, value, rule, idmap = args
     if __debug__:
-        trace('Builder: call_fn %s, key=%s, value=%r, %r' % (
+        trace('Lang: call_fn %s, key=%s, value=%r, %r' % (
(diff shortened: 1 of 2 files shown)


dessant:
We use both Lang and Builder in builder.py, Parser would be ok if we want to make a distinction from Builder, though Lang seems enough for me.

rafalo1333:
So you think it will be okay to change the headers in builder.py/parser.py
to use "Lang" descriptor? If so Ill do another commit for that.

04.09.2016 5:57 PM "dessant" <notifications@github.com> napisał(a):

dessant:
Yes, I'm personally ok with it, but i'd wait until @matham drops by.

matham:
Yeah, I think Lang is probably better to be consistent.

rafalo1333:
@matham @dessant OK I reworked it to use Lang everywhere, I would be glad if you can test it. If you'll find places where we can also change header ping me, otherwhise its ready to merge.

Issue #4581, opened 2016-09-04
cannot compile current kivy master


ecdsa:
this commit breaks it: 0e46e3d6b302c94e82843fb08000b0ba12e67bda
I get the same error as described in the comment

udiboy1209:
@ecdsa can you give some more details about what you are trying to do? Compile on what system and for what architecture?

ecdsa:
I was initially trying to build an APK with buildozer.
However, I also get the same error when trying to build kivy on ubuntu (without buildozer).
I do "sudo python setup.py install"
and I get 
```
cythoning kivy/graphics/instructions.pyx to kivy/graphics/instructions.c
 
 Error compiling Cython file:
 ------------------------------------------------------------
 ...
             if do_parent == 1 and self.parent is not None:
                 instrs.append(self)
                 self.parent.flag_update(do_parent=1, _instrs=instrs)
             self.flags |= GI_NEEDS_UPDATE
     ELSE:
         cpdef void flag_update(self, int do_parent=1):
              ^
 ------------------------------------------------------------
 
 kivy/graphics/instructions.pyx:75:14: Return with value in void function
 
 Error compiling Cython file:
 ------------------------------------------------------------
 ...
             if do_parent == 1 and self.parent is not None:
                 instrs.append(self)
                 self.parent.flag_update(do_parent=1, _instrs=instrs)
             self.flags |= GI_NEEDS_UPDATE
     ELSE:
         cpdef void flag_update(self, int do_parent=1):
              ^
 ------------------------------------------------------------
 
 kivy/graphics/instructions.pyx:75:14: Cannot convert 'void' to Python object

```

udiboy1209:
I get this warning at that same point, but no error. Make sure you have the correct cython(0.23) version, can't think of anything else that would cause such an issue.

```
cythoning kivy/graphics/instructions.pyx to kivy/graphics/instructions.c
building 'kivy.graphics.instructions' extension
x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -I/usr/include/python3.5m -I/home/udiboy/.apps/venvs/kivy3/include/python3.5m -c kivy/graphics/instructions.c -o build/temp.linux-x86_64-3.5/kivy/graphics/instructions.o
[...]

PR #4585, opened 2016-09-04
Update cython min version to 0.23 in setup.py

udiboy1209:
Latest master doesn't compile with '0.22'

This will help prevent bugs like #4581

I hope it isn't a problem that both min and max versions are '0.23'

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ if PY3:  # fix error with py3's LooseVersion comparisons
     LooseVersion.__eq__ = ver_equal
 
 
-MIN_CYTHON_STRING = '0.20'
+MIN_CYTHON_STRING = '0.23'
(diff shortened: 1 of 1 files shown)


dessant:
We could also rise the max version. @inclement, is p4a working with the latest cython release?

udiboy1209:
Is a successful compilation enough to say that kivy will work correctly with that cython version?

matham:
Yes, compilation and make test.

akshayaurora:
I have it working with 0.24

udiboy1209:
Travis jobs are using the latest stable cython (0.24.1). I guess that would work with Kivy then. Don't know about p4a though.

akshayaurora:
I was talking about p4a,  it works with 0.24

PR #4586, opened 2016-09-05
Switch rework

KeyWeeUsr:
New:
* color
* background_* and button_* properties (allow customizing widget later)
* `markup == True`

Changes:
* remove text from switch's images in theme
* make disabled button non-transparent and darked, so that it resembles the original semi-transparent image. Why: → the text from Label is visible through the button when disabled - changes in properties while checking still in canvas are kind of "ignored" even if Switch was directly created with `active: True` (`Color:\n\rrgb: this if self.active else that`... always shows the default value, not the actual one)
* `values` as a property for text used in switch

Why it should be merged:
* text burned into an image is really bad and introduces even worse ways of fixing it
* background_* and button_* properties are available in most of the widgets anyway, so it should be here too
* markup is just for fancy styling, because text is handled with Labels
* makes Switch an easily customizable widget and works(_should_, e.g. kanji symbols or encoding depends on Label+font) out-of-box even for non-english apps

Not sure about:
* color → the theme is blue-ish, this makes colorful text either badly visible through the background or hm... hurting eyes if a bad color is used
* text sizing → should it be within the area available for the default `ON/OFF` and font_size handled according to that or leave it like that (and let user report/fix it if needed)?
* should font be added as property too?


KeyWeeUsr:
After a talk with @dessant it seems that to preserve Switch as it is and allow localization the best option might be to remove the text completely. [This](http://ankri.de/wp-content/uploads/2012/07/Screenshot_2012-07-30-13-54-10.png) seems like a bad idea to me too as it changes width according to text size and may break layouts. I still struggle about this, because the widget seems too empty now and I think reverting to previous design (`I/O`) might be a good idea then.

To compare it with android versions, ICS had on/off switch (maybe even on devices using non-ascii language), but on my note 2 I see revert to `I/O` design on KitKat. Any ideas? @tito ?

KeyWeeUsr:
Ok, this wasn't probably a happy idea. Anyway, I'm glad at least the canvas in Switch is "separated", so that, for anyone interested, a trick like this is possible without creating additional image(s). 😄

```
from kivy.lang import Builder
Builder.load_string('''
<Custom@Switch>:
    values: ['OFF', 'ON']
    canvas:
        Color:
            rgb: 0.2, 0.709, 0.898, 1
        Rectangle:
            size: [sp(41.5), sp(20)]
            pos: [self.center_x - sp(41.5), self.center_y - sp(10)]
        Color:
            rgb: 0.4, 0.4, 0.4, 1
        Rectangle:
            size: [sp(41.5), sp(20)]
            pos: [self.center_x, self.center_y - sp(10)]
    Label:
        text: '[b]{}[/b]'.format(root.values[0])
        markup: True
        font_size: 13
        pos: [root.center_x - sp(70), root.center_y - sp(50)]
    Label:
        color: 0.75, 0.75, 0.75, 1
        text: '[b]{}[/b]'.format(root.values[1])
        markup: True
        font_size: 13
        pos: [root.center_x - sp(30), root.center_y - sp(50)]

<Test>:
    Custom:
    Switch:
    Custom:
        values: ['Yes', 'No']
''')
class Test(BoxLayout): pass
runTouchApp(Test())
```

Reopen/ping if there is a different approach to Switch without burning the text into image.

PR #4587, opened 2016-09-07
Fix remaining misspellings

kwadrat:


KeyWeeUsr:
I think editing variables' names should have a separate PR in the future. Changes in a casual text aren't as important as changes in core (especially cython) parts and it may happen that those files will hide in the sea of docs changes(92 files, so quite possible). There are some tests and the PR passed, but what if there's no test for some particular change and a file slips reviewer's sight?

Anyway, seems good to me, thanks for the contribution :+1:

kwadrat:
Thank you. Next time I will create separate PR for code changes and separate PR for string/comment changes.

PR #4588, opened 2016-09-07
Update boxlayout.py

user5061:
[Note contents](https://kivy.org/docs/api-kivy.uix.boxlayout.html#module-kivy.uix.boxlayout) were not displayed on different lines.


matham:
Yeah, it needs to be indented.

user5061:
Have you indented it? I can't view the result (can't build the docs). If not, should just each code line be indented 4 spaces to the right?

matham:
I have not, and yes they should be indented.

Issue #3753, opened 2015-11-08
Add 'is_shortened' BooleanProperty to Label


dessant:
There should be a boolean property that can be checked to find out if the label was shortened or not.
One use case is to display the full text in a tooltip, on hover.

matham:
Maybe an event is better? So one can have access to the shortened text? Although a prop is probably just as good.

dessant:
A prop might be better, with an event only, one would end up needing to store the state in a prop.
But yeah, we could have `shortened_text` instead of a boolean.

matham:
Yeah, in which case, a value of `None` means no shortened text.

dessant:
We could also opt for an empty string and use a StringProperty.

PR #4589, opened 2016-09-07
Add Label.is_shortened

dessant:


dessant:
Closes https://github.com/kivy/kivy/issues/3753.

matham:
What about...[markup](https://github.com/kivy/kivy/blob/master/kivy/core/text/markup.py#L252)?! I'm rethinking whether a bool might be better because for markup we'd have to compute the shortened text which would not be needed. Would a user ever need the text in shortened form?

Also, I believe your code has issues where `shortened_text` will be wrong since it's only updated when render is called and it has reached the shorten path, but the text may change well before that without actually rendering, in which case shortened_text should be reset to None.

dessant:
@matham, I'm not currently seeing a legitimate use case for having access to the shortened text either, `is_shortened` then? I'll add markup and look into the rest.

PR #4594, opened 2016-09-08
Add on_release_outside event in ButtonBehavior, fix #4593

### LINKED ISSUE #4593: ToggleButton can get released with allow_no_selection=False
Ohai buddiez, I hope you're all doing goodie ^__^

I found a new bug probably due to "always_release" on ButtonBehavior having been changed recently:
`Changed in version 1.9.2: The default value is now False.`

Take the following example:
https://gist.github.com/42e02d13c31a6504b57d5cd3ac23a460

If you try to press a button, then release outside of it, its state will be "normal", even though it should remain "down".

I have made a small change to ButtonBehavior which adds an event "on_release_outside" and makes it set the state to normal by default, and overrode it in ToggleButtonBehavior to do nothing instead.

Incoming PR, so please give feedback! (EDIT: See #4594 )

Thanks for reading, see you around buddiez bliblibli ^__^

Cheaterman:
Ohai buddiez, hope you doing goodie ^__^

Everything is in the title and in #4593 .

Thanks in advance!

Have a good one buddiez ^__^

### DIFF
--- a/kivy/uix/behaviors/button.py
+++ b/kivy/uix/behaviors/button.py
@@ -116,6 +116,7 @@ class ButtonBehavior(object):
     def __init__(self, **kwargs):
         self.register_event_type('on_press')
         self.register_event_type('on_release')
+        self.register_event_type('on_release_outside')
         if 'min_state_time' not in kwargs:
             self.min_state_time = float(Config.get('graphics', 'min_state_time'))
         super(ButtonBehavior, self).__init__(**kwargs)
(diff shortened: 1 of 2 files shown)


matham:
I don't think a new event is required, it's just a issue with the button logic. The fix should be `             self._do_release()` where `self.state = 'normal'` currently is.

Cheaterman:
Totally correct! Please feel free to close/amend my PR, and fix the issue upstream :-)

Issue #4465, opened 2016-07-13
App: support background persistence on Android/iOS by default


rafalo1333:
I've been developing Kivy apps for Android for over 2.5 years. There was no single app that I wanted to get killed after user pressed home button. Every app I have been writing required me to write:

``` python
class SomeApp(App):

    def on_pause(self):
        return True
```

Across the web, on #kivy IRC, in tutorials, on user groups, reddit forums: everywhere users are asking how to prevent the app from being killed when minmised on Android and iOS. Some users using Kivy for the first time think it's not possible to develop the Kivy app that stays in background when minimised. So, its the thing that requires the developers to write more code and at the same time makes the bad first impression that Kivy is limited and cannot do the same as Java/ObjC for the new Kivy users.

My proposal is that default behaviour should be that `on_pause()` method returns `True` on mobile platforms and if someone wants to disable the background persistence of the app, he can override that behaviour. As we see everywhere on the web, we can bet that more than 90% of the users do want the background persistence of the app. It should point to the conclusion, that turning it off should require the override, not turning it on.

I understand that we can label it as `api-break`, but I believe it can give a lot of love to the Kivy from beginning users and more experienced people too, as they wont have to write this simple 2-liner for 1074334764 time in their life.

It will make kivy inline with other framework, which support the default iOS/Android behaviour out-of-the-box.

What do you think?

dessant:
I fully agree with everything you said, defaults should reflect the most common use cases.

tito:
You're right, it's time to make the change and fix the possible issue with it.

Issue #2572, opened 2014-10-11
SDL2 - the ui is not visible after minimizing the window


dessant:
Minimize the window, then restore it; the app canvas is white/black until the window is resized or a button/textinput is clicked, or anything else that triggers a canvas update. Seen on Windows.

**Edit**: skip forward to the example and video: https://github.com/kivy/kivy/issues/2572#issuecomment-106606800
##

akshayaurora:
@dessant do you experience the issue if you set pause_on_minimise to True and return True on your on_pause? It does not seem to be a issue for my tests.

dessant:
Yes, i have reproduced it like that on windows.

dessant:
@akshayaurora https://vid.me/e/p1cw

PR #4603, opened 2016-09-11
Made on_pause default to True

inclement:
Not sure if there's a reason not to do this. I think it's a better default on Android almost all the time.

It might affect desktop behavior with the fancy sdl2 pausing we have there now (at least on OS X, as I remember), but I haven't seen anyone have problems with that so I'd guess it isn't a big issue.

### DIFF
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -868,7 +868,7 @@ class App(EventDispatcher):
 
         .. versionadded:: 1.1.0
         '''
-        return False
+        return True
 
     def on_resume(self):
         '''Event handler called when your application is resuming from


akshayaurora:
:+1:

rafalo1333:
I am for it, that's why I started #4465. It's natural default for Android/iOS.

matham:
The docs saying `False` is the default should be updated as well.

akshayaurora:
Please make sure to return false on desktops. we had it default to True and it lead to issues on desktops earlier if I remember correctly.

dessant:
@akshayaurora, there is no need to return False from `on_pause` only on desktops. You were remembering the `pause_on_minimize` config option (which defaults to False), and there is a separate bug report for that issue: https://github.com/kivy/kivy/issues/2572#issuecomment-106606800.

PR #4609, opened 2016-09-14
Update setup.py for SDL2 on Linux

pptime:
issue:
Platform:  Centos 7 x86
When I activate environment variable USE_SDL2=1 on centos 7, the compilation flags for SDL2 is thrown away somewhere before compiling the source code. Before doing this, I ensure that the SDL2 suites are correctly installed and can be resolved correctly.
Solution:
merge sdl2_flags with the flags such that compiling check will not fail when sdl2_flags is correct

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -603,7 +603,7 @@ def determine_sdl2():
 
     sdl2_path = environ.get('KIVY_SDL2_PATH', None)
 
-    if sdl2_flags and not sdl2_path and platform == 'darwin':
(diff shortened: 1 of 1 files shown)


matham:
Yeah, that seems like a bug currently because `and` is evaluated before `or`. Also, I think the line length is over 80 chars. But I may be wrong there.

pptime:
Yes, indeed. I've made a quick local change without paying too much
attention on that. I just made a fix. BTW, is there anyway to make a pull
request without passing through github?

2016-09-28 1:24 GMT+02:00 matham <notifications@github.com>:

PR #4614, opened 2016-09-18
Update texture.pyx

matumaros:
Added note that texture size needs to be a multiple of 8


KeyWeeUsr:
Seems kind of strange to me, I've always thought the texture needs to be of power 2 and that's it.

matumaros:
Oh yes, sorry. I had discussed that in chat the other day and it was a mistake on my part. I also thought that somebody was closing this.

PR #4615, opened 2016-09-19
Kv Lang: fix profiling tool HTML output generation

rafalo1333:
Fixes `TypeError: coercing to Unicode: need string or buffer, NoneType found` error during HTML output generation. Skips one loop iteration if the file cant be opened.

Works like a charm for me and I believe it's better way of handling file than None check. Tool can encounter various problems while handling files and try/except is the most reliable solution in my opinion.

### DIFF
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -865,7 +865,10 @@ if 'KIVY_PROFILE_LANG' in environ:
             '</style>']
         files = set([x[1].ctx.filename for x in Builder.rules])
         for fn in files:
-            lines = open(fn).readlines()
+            try:
+                lines = open(fn).readlines()
+            except:
+                continue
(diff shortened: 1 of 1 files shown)


kivy/lang/builder.py
```diff
@@ -865,7 +865,10 @@ def dump_builder_stats():
             '</style>']
         files = set([x[1].ctx.filename for x in Builder.rules])
         for fn in files:
-            lines = open(fn).readlines()
+            try:
+                lines = open(fn).readlines()
```
matham: This should really be `with open(fn) as fh: fh.readlines()`. Also, rather than excepting everything, you should only except IOError.

rafalo1333: Good point, i would update this :+1:

rafalo1333: @matham updated, I think its good now.

PR #4617, opened 2016-09-19
Adds size_hint_min/max to widgets.

matham:
The addition is explained in Widget, partially quoted below:

```py
size_hint_min_x = NumericProperty(None, allownone=True)
    '''When not None, the X-direction minimum size (in pixels,
    like :attr:`width`) when :attr:`size_hint_x` is also not None.

    When :attr:`size_hint_x` is not None, it is the minimum width that the
    widget will be set due to the :attr:`size_hint_x`. I.e. when a smaller size
    would be set, :attr:`size_hint_min_x` is the value used instead for the
    widget width. When None, or when :attr:`size_hint_x` is None,
    :attr:`size_hint_min_x` doesn't do anything.
    '''

    size_hint_min_y = NumericProperty(None, allownone=True)
    '''When not None, the Y-direction minimum size (in pixels,
    like :attr:`height`) when :attr:`size_hint_y` is also not None.

    When :attr:`size_hint_y` is not None, it is the minimum height that the
    widget will be set due to the :attr:`size_hint_y`. I.e. when a smaller size
    would be set, :attr:`size_hint_min_y` is the value used instead for the
    widget height. When None, or when :attr:`size_hint_y` is None,
    :attr:`size_hint_min_y` doesn't do anything.
    '''

    size_hint_min = ReferenceListProperty(size_hint_min_x, size_hint_min_y)

    size_hint_max_x = NumericProperty(None, allownone=True)
    '''When not None, the X-direction maximum size (in pixels,
    like :attr:`width`) when :attr:`size_hint_x` is also not None.

    Similar to :attr:`size_hint_min_x`, except that it sets the maximum width.
    '''

    size_hint_max_y = NumericProperty(None, allownone=True)
    '''When not None, the Y-direction maximum size (in pixels,
    like :attr:`height`) when :attr:`size_hint_y` is also not None.

    Similar to :attr:`size_hint_min_y`, except that it sets the maximum height.
    '''

    size_hint_max = ReferenceListProperty(size_hint_max_x, size_hint_max_y)
```

All layouts including RecycleView layouts support it. 
In terms of reduced perf, if size_hint_min/max is not used the layout incurs minimum cost (except maybe for StackLayout since the code there generally isn't as optimized as it could probably be), except of course due to the addition of these new properties to the widgets.

Example code:
```py
from kivy.lang import Builder

BoxLayout:
    orientation: 'vertical'
    GridLayout:
        rows: 1
        Button:
            text: '1'
            width: 100
            size_hint_x: None
        Button:
            text: '2'
            size_hint_min_x: 300
        Button:
            text: '3'
        Button:
            text: '4'
            size_hint_max_x: 25
    BoxLayout:
        Button:
            text: '1'
            width: 100
            size_hint_x: None
        Button:
            text: '2'
            size_hint_min_x: 300
        Button:
            text: '3'
        Button:
            text: '4'
            size_hint_max_x: 25
'''))
```

### DIFF
--- a/examples/widgets/recycleview/basic_data.py
+++ b/examples/widgets/recycleview/basic_data.py
@@ -15,8 +15,6 @@ kv = """
             size: self.size
             pos: self.pos
     value: ''
-    size_hint_y: None
-    height: dp(100)
     Label:
         text: root.value
 
--- a/examples/widgets/recycleview/basic_data.py
+++ b/examples/widgets/recycleview/basic_data.py
@@ -15,6 +15,8 @@ kv = """
             size: self.size
             pos: self.pos
     value: ''
+    size_hint_y: None
+    height: dp(100)
     Label:
         text: root.value
 
--- a/kivy/compat.py
+++ b/kivy/compat.py
@@ -7,14 +7,19 @@ to aid in writing Python 2/3 compatibile code.
 '''
 
 __all__ = ('PY2', 'clock', 'string_types', 'queue', 'iterkeys',
-           'itervalues', 'iteritems')
+           'itervalues', 'iteritems', 'isclose')
 
 import sys
 import time
+from math import isinf, fabs
 try:
     import queue
 except ImportError:
     import Queue as queue
+try:
+    from math import isclose
+except ImportError:
+    isclose = None
 
 PY2 = sys.version_info[0] == 2
 '''True if this version of python is 2.x.'''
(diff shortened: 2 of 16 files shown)


KeyWeeUsr:
Perfect! A few minutes ago I was struggling with proper sizing. This should solve it. :yum:

matham:
Now added support to all layouts including recycleview layout. **Testing is welcome!** Especially for recycleview since I didn't test that part yet (don't need to test that hint_max/min work in rv, just that I didn't break the rv widgets with these changes).

Also, I changed the signature of `refresh_view_layout` from listing all the sizing info to a single layout dict. It should only affect you if you overwrote that method, which is unlikely to be the case.

PR #4618, opened 2016-09-19
Update pagelayout.py

llfkj:
Remove what looks like a pasto introduced in commit c951666


KeyWeeUsr:
The line doesn't seem to do anything. In the worst case it could wait for some error, however it doesn't seem it's the thing. Thanks ^^

PR #4625, opened 2016-09-21
Ellipsis Style for markup label

### LINKED ISSUE #4580: Label ellipsis cannot be styled
The desired solution is described by @matham here https://github.com/kivy/kivy/pull/2642#issuecomment-61826012.

udiboy1209:
Fixes #4580 

Keeping a different font size breaks baseline alignment of ellipsis. I don't know if that is supposed to happen (as an outcome of bad UI config). If it is a required feature I'll look into it.

Example app

```
from kivy.app import App
from kivy.lang import Builder

kv = """
Label:
    text_size: 200, None
    text: 'Some very long line which will be cut'
    markup: True
    shorten: True
    ellipsis_options: {'color':(1,0.5,0.5,1),'underline':True}
"""


class TestApp(App):
    def build(self):

if __name__ == '__main__':
    TestApp().run()
```

### DIFF
--- a/kivy/core/text/markup.py
+++ b/kivy/core/text/markup.py
@@ -752,7 +752,14 @@ class MarkupLabel(MarkupLabelBase):
         s = self.get_extents(last_text)
         if len(last_text):
             line1.append(LayoutWord(last_word.options, s[0], s[1], last_text))
-        elps.options = last_word.options
+
+        elps_opts = old_opts
+        if 'ellipsis_options' in old_opts:
+            user_opts = old_opts['ellipsis_options']
+            for k in user_opts:
+                elps_opts[k] = user_opts[k]
+
+        elps.options = elps_opts
         line1.append(elps)
 
         # now add back the right half
--- a/kivy/core/text/markup.py
+++ b/kivy/core/text/markup.py
@@ -627,6 +627,12 @@ class MarkupLabel(MarkupLabelBase):
             return lw + 2 * xpad, lh + 2 * ypad, [LayoutLine(0, 0,
             lw, lh, 1, 0, line)]
 
+        elps_opts = copy(old_opts)
+        if 'ellipsis_options' in old_opts:
+            elps_opts.update(old_opts['ellipsis_options'])
+
+        # Set new opts for ellipsis
+        self.options = elps_opts
         # find the size of ellipsis that'll fit
         elps_s = textwidth('...')
         if elps_s[0] > uw:  # even ellipsis didn't fit...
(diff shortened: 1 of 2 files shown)


kivy/core/text/markup.py
```diff
@@ -752,7 +752,14 @@ def p_restricted(line, uw, c):
         s = self.get_extents(last_text)
         if len(last_text):
             line1.append(LayoutWord(last_word.options, s[0], s[1], last_text))
-        elps.options = last_word.options
+
+        elps_opts = old_opts
```
matham: This stuff needs to happen before `elps_s = textwidth('...')` above because that's when we compute the ellipse size. If the ellipse changes size according to these new opts, we'll have to know it there to make sure it fits.

kivy/core/text/markup.py
```diff
@@ -752,7 +752,14 @@ def p_restricted(line, uw, c):
         s = self.get_extents(last_text)
         if len(last_text):
             line1.append(LayoutWord(last_word.options, s[0], s[1], last_text))
-        elps.options = last_word.options
```
matham: Hmm, I cannot remember why I overwrite the ellipse options with the last word's options. As far as I remembered the ellipse should be the same opts as the main text, but this doesn't seem to comply with that. Unless the last word's opts is the same as the main text, but then I'm not sure why I overwrote it here.

matham: I think this should have been `elps.options = old_opts`, which is basically redundant since the ellipse options had been set properly earlier. It would be good to test to see if the ellipse ever takes the opts of the word to the left, because that would be wrong then and this line should just be removed.

kivy/core/text/markup.py
udiboy1209: I can't see how `elps_opts` will affect the `textwidht()` function. Nonetheless, I made the change but it didn't fix the variable font case.

matham: What was the change, and how did it not fix it? As for `textwidth`, check the function from the providers and you'll see it relies on the current options. That's why self.option need to be updated and restored before doing that.

udiboy1209: I moved the `elps_opts` assignment before `textwidth('...')` call. I didn't update self.options, I'll try once with that.

udiboy1209: @matham check the latest update, now different font sizes work fine.

kivy/core/text/markup.py
```diff
@@ -627,6 +627,14 @@ def p_restricted(line, uw, c):
             return lw + 2 * xpad, lh + 2 * ypad, [LayoutLine(0, 0,
             lw, lh, 1, 0, line)]
 
+        elps_opts = copy(old_opts)
+        if 'ellipsis_options' in old_opts:
```
matham: This whole block can be written as `elps_opts.update(old_opts.get('ellipsis_options', {}))`. It's up to you if you want to change it, but either way feel free to merge.

udiboy1209: Thanks :+1:

PR #4626, opened 2016-09-21
Doc fix for split_str in Label

udiboy1209:
Looked like a mistake, one of @kivy/contributors contributors please verify!


KeyWeeUsr:
Yup, it's a mistake. Thanks. :)

PR #4627, opened 2016-09-21
Fix property numeric

llfkj:
NumericProperty won't convert something that looks like a number unless it has units attached to it.

I found this trying to populate a `padding` value from a `Settings` pane.
Since such an item can be a bare string, or any of a 1, 2, or 4 item list, I was reading it in as a string, and then converting things to NumericProperties.

Here's how you can see the issue.

```python
class TestProperty(EventDispatcher):
    pass

wid = TestProperty()

a = NumericProperty()
a.link(wid, 'a')
a.link_deps(wid, 'a')
print(a.get(wid))
a.set(wid, '1px')
print(a.get(wid))
a.set(wid, '2')
print(a.get(wid))
```

After the fix, something like '2' will be handled just like '2px'

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -579,7 +579,10 @@ cdef class NumericProperty(Property):
                 self.name, x))
 
     cdef float parse_str(self, EventDispatcher obj, value):
-        return self.parse_list(obj, value[:-2], value[-2:])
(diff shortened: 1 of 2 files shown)


matham:
I think that's probably fine.

matham:
We might similarly want to fall back on `float(value)` in convert, but I guess it's unrelated.

PR #4628, opened 2016-09-22
AnchorLayout is badly broken

llfkj:
An asymmetric padding list will show the problem. If you use a constant pad value, you likely wouldn't have noticed.

In short:
* Top/bottom padding indices were mixed up
* `size_hint` calculation is wrong
* `center` calculations ignore padding entirely (placing all children in the middle of the screen regardless of padding)

These are all easy to see by running the code at: https://gist.github.com/llfkj/08c2ccede6cf271d4238405d9ba5e2d6

The linked example should create a button that is half the size of the padded area, and attempts to place it within the rectangle. Short version: it doesn't. The white rectangle is the area inside the padding.

### DIFF
--- a/kivy/uix/anchorlayout.py
+++ b/kivy/uix/anchorlayout.py
@@ -82,33 +82,30 @@ class AnchorLayout(Layout):
         height = self.height
         anchor_x = self.anchor_x
         anchor_y = self.anchor_y
-        padding = self.padding
-
-        left, top, right, bottom = range(4)
-        horizontal, vertical = range(2)
+        pad_left, pad_top, pad_right, pad_bottom = self.padding
 
         for c in self.children:
             x, y = _x, _y
             cw, ch = c.size
-            if c.size_hint[horizontal] is not None:
-                cw = c.size_hint[horizontal] * (width - padding[left] - padding[right])
-            if c.size_hint[vertical] is not None:
-                ch = c.size_hint[vertical] * (height - padding[top] - padding[bottom])
+            size_hint_h, size_hint_v = c.size_hint
+
+            if size_hint_h is not None:
+                cw = size_hint_h * (width - pad_left - pad_right)
+            if size_hint_v is not None:
+                ch = size_hint_v * (height - pad_top - pad_bottom)
 
             if anchor_x == 'left':
-                x = x + padding[left]
-            if anchor_x == 'right':
-                x = x + width - (cw + padding[right])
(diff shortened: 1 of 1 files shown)


kivy/uix/anchorlayout.py
```diff
@@ -84,28 +84,31 @@ def do_layout(self, *largs):
         anchor_y = self.anchor_y
         padding = self.padding
 
+        left, top, right, bottom = range(4)
```
matham: This should be `padd_left, padd_top, padd_right, padd_bottom = self.padding`. Similarly, it should be `shw, shh = self.size_hint` rather than `horizontal, vertical = range(2)` and then indexed.

kivy/uix/anchorlayout.py
```diff
@@ -84,28 +84,31 @@ def do_layout(self, *largs):
...
+            cw, ch = c.size
+            if c.size_hint[horizontal] is not None:
+                cw = c.size_hint[horizontal] * (width - padding[left] - padding[right])
+            if c.size_hint[vertical] is not None:
+                ch = c.size_hint[vertical] * (height - padding[top] - padding[bottom])
 
             if anchor_x == 'left':
-                x = x + padding[0]
+                x = x + padding[left]
             if anchor_x == 'right':
```
matham: These may as well be changed to a bunch of `elifs`.

kivy/uix/anchorlayout.py
```diff
@@ -84,28 +84,31 @@ def do_layout(self, *largs):
...
-                x = x + width - (w + padding[2])
+                x = x + width - (cw + padding[right])
             if self.anchor_x == 'center':
```
matham: And the `self.` part can be removed.

kivy/uix/anchorlayout.py
```diff
@@ -84,28 +84,31 @@ def do_layout(self, *largs):
...
             if anchor_y == 'center':
-                y = y + (height / 2) - (h / 2)
+                y = y + (height - padding[top] + padding[bottom] - ch) / 2
 
             c.x = x
             c.y = y
-            c.width = w
-            c.height = h
+            c.width = cw
+            c.height = ch
```
matham: Also, `c.size = cw, ch` and similarly for `pos` is more performant.

llfkj:
Any other comments on this? I can't imagine anyone actually using the AnchorLayout in a way that they depend on the broken behavior.

PR #4629, opened 2016-09-22
Add Widget.remove_edge

KeyWeeUsr:
Not sure about the naming, but it could be included as it makes easier for user to just put `<widget>.remove_edge()` in kv without making additional function to care about error handling with `remove_widget()`. Kind of like `list.pop()`.

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -587,6 +587,28 @@ class Widget(WidgetBase):
             self.canvas.before.remove(widget.canvas)
         widget.parent = None
 
+    def remove_edge(self, newest=True):
+        '''Remove the newest or the oldest added widget from the children
+        of this widget.
+
+        :Parameters:
+            `newest`: bool
+                Select the newest or the oldest widget.
+
+    .. code-block:: python
+
+        >>> from kivy.uix.button import Button
+        >>> root = Widget()
+        >>> root.add_widget(Button())
+        >>> root.add_widget(Button(text='Bye!'))
+        >>> root.remove_edge()  # removes ^
+        '''
+        try:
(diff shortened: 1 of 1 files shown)


matham:
I don't think I've ever needed to use something like that, especially not
from kv. Why would you to need remove the last or first widget? Even more,
if you do need to remove it you should know if it exists so you wouldn't
need the exception catching at which point you don't need the function
since its a one liner.

On Sep 22, 2016 1:26 AM, "Peter Badida" <notifications@github.com> wrote:

Not sure about the naming, but it could be included as it makes easier for
user to just put <widget>.remove_edge() in kv without making additional
function to care about error handling with remove_widget(). Kind of like
list.pop().
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/kivy/kivy/pull/4629
Commit Summary

   - Add Widget.remove_edge

File Changes

   - *M* kivy/uix/widget.py
   <https://github.com/kivy/kivy/pull/4629/files#diff-0> (22)

Patch Links:

   - https://github.com/kivy/kivy/pull/4629.patch
   - https://github.com/kivy/kivy/pull/4629.diff

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/kivy/kivy/pull/4629>, or mute the thread

KeyWeeUsr:
I was thinking about some button with this in on_release to remove a recently added child. I forgot about the check :disappointed: this makes the pr irrelevant, I guess, as the check can be performed in kv. Only pro of the pr now seems to be shorter line of code if kv: `if parent.children: parent.remove_widget(parent.children[0])`

PR #4632, opened 2016-09-24
Fix #4331

### LINKED ISSUE #4331: TextInput selection_text not cleared on backspace/delete or widget defocus
Select part of the text and press backspace or delete, then press the print button, the previously selected text will still be there. The same is the case when the widget is defocused after selection.

```python
from kivy.app import App
from kivy.lang import Builder

kv = """
BoxLayout:
    orientation: 'vertical'
    TextInput:
        id: input
        text: 'select me'
    Button:
        text: 'Print selected text'
        on_press: print(input.selection_text)
"""


class TestApp(App):
    def build(self):

if __name__ == '__main__':
    TestApp().run()
```

KeyWeeUsr:
`selection_text` is a property that stores the value of selected text. Although everything else was cleared, this one remained the same, therefore issues like 4331 happened.

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -1187,6 +1187,7 @@ class TextInput(FocusBehavior, Widget):
         self._selection = False
         self._selection_finished = True
         self._selection_touch = None
+        self.selection_text = u''
         self._trigger_update_graphics()
 
     def delete_selection(self, from_undo=False):


udiboy1209:
Looks good :+1: Thanks!

dessant:
Just a minor thing, it's nice to use more descriptive commit names, like `clear selection_text when TextInput selection is canceled` or something similar, it helps a lot while browsing the commit log, and also when compiling the changelog before releases.

KeyWeeUsr:
Yeah, I should have named it better. But at least the pr title can be renamed.

PR #4633, opened 2016-09-26
postpone scroll_to

nadaz:
postpone the execution of the method if the viewport has pending layout operation


KeyWeeUsr:
Thanks, cProfiler likes it ^^

Issue #4634, opened 2016-09-26
self.stop() fails with AttributeError: 'NoneType' object has no attribute 'children' in app's build()


ilovetogetspamed:
I was trying to trap for not being able to connect to the Serial port so I could log a message, send alert, etc. and then cleanly shut down Kivy.  

When I force the error in opening the port, by adding an 's' to the device name, i.e. the real name of the serial device on my Mac is '/dev/cu.usbmodem14671', I get an error message instead of the app.stop() shutting down the app, as I had expected it to.

rafalo133 on IRC #Kivy asked me to open this.

```
self.serial = SerialPort(McuProtocol(), '/dev/cu.usbmodem14671s', reactor, ...
```

Offending code follows:

```
def build(self):

    # Build the screen manager
    self.root.ids.screen_manager.current = 'splash_screen'
    # ...

    # Open the Serial Port
    SERIAL_PORT = '/dev/ttyACM0'

    try:
        # load twisted protocol
        self.serial = SerialPort(McuProtocol(), SERIAL_PORT, reactor, baudrate=115200)
        Logger.exception('EPIC.AR: Serial: Unable to open the serial port.  Check connections!')
        self.stop()
    except SerialException as e:
        try:
            self.serial = SerialPort(McuProtocol(), '/dev/cu.usbmodem14671s', reactor, baudrate=115200)
        except SerialException as e:
            Logger.exception('EPIC.AR: Serial: Unable to open the serial port.  Check connections!')
            self.stop()

    return self.root
```

```

..
[ERROR             ] [MY.APP     ] Serial: Unable to open the serial port.  Check connections!
Traceback (most recent call last):
  File "/Users/cfo64nc/Documents/GitHub/MyApp-PROJECT/MyApp-Kivy-Python/MY.APP/main.py", line 289, in build
    self.serial = SerialPort(McuProtocol(), '/dev/cu.usbmodem14671s', reactor, baudrate=115200)
  File "/Users/cfo64nc/.env/myapp_ar_django/lib/python2.7/site-packages/twisted/internet/_posixserialport.py", line 37, in __init__
    xonxoff=xonxoff, rtscts=rtscts)
  File "/Users/cfo64nc/.env/myapp_ar_django/lib/python2.7/site-packages/serial/serialutil.py", line 182, in __init__
    self.open()
  File "/Users/cfo64nc/.env/myapp_ar_django/lib/python2.7/site-packages/serial/serialposix.py", line 247, in open
[...]

PR #4636, opened 2016-09-26
App.stop(): clear window children only if window exists

rafalo1333:


rafalo1333:
Partially fixes #4634

PR #4637, opened 2016-09-26
stopTouchApp(): don't simply return, try stop in the next frame

### LINKED ISSUE #4634: self.stop() fails with AttributeError: 'NoneType' object has no attribute 'children' in app's build()

rafalo1333:
If someone wants to close the app, its the app which is responsible of successful finishing the eventloop.

There are use cases, as described in #4634 when user wants to close the app on startup, for example in build() or on_start(). Current solution breaks such possibility: when eventloop is not available, the request to close will be forgotten. That together with #4636 fixes #4634 completely.

### DIFF
--- a/kivy/base.py
+++ b/kivy/base.py
@@ -491,9 +491,8 @@ def runTouchApp(widget=None, slave=False):
 
 def stopTouchApp():
     '''Stop the current application by leaving the main loop'''
-    if EventLoop is None:
-        return
-    if EventLoop.status != 'started':
+    if EventLoop is None or EventLoop.status != 'started':
+        Clock.schedule_once(lambda dt: stopTouchApp(), 0)
         return
     Logger.info('Base: Leaving application in progress...')
     EventLoop.close()


kivy/base.py
```diff
@@ -491,9 +491,8 @@ def runTouchApp(widget=None, slave=False):
 
 def stopTouchApp():
     '''Stop the current application by leaving the main loop'''
-    if EventLoop is None:
-        return
-    if EventLoop.status != 'started':
+    if EventLoop is None or EventLoop.status != 'started':
+        Clock.schedule_once(lambda dt: stopTouchApp(), 0)
```
matham: This seems dangerous. Imagine you want to stop and start again the event loop (which is not really support now, but still...), calling stop too late after it's already stopping will keep scheduling to stop until the next time you run it, which is not what was wanted. I think we should only schedule if we know that the event loop is starting, but not before it started at all and not if it's already stopping.

rafalo1333: @matham I dont understand the problem. In my PR calling app.stop() before eventloop is ready will schedule it, then callback will again check if eventloop is started. If it has, app will close, otherwhise the stop will be again scheduled. If the eventloop is stopped, there is no schedule, so next time user calls runTouchApp there should be no crash because callback is scheduled somewhere.

Am I not correct? Maybe I misunderstood the problem :)

rafalo1333: Ok, got it. I'll try smarter approach.

dessant:
@rafalo1333, please reopen if you discover a better fix.

PR #4641, opened 2016-09-29
fix on_scroll_move to obey scroll_distance

darkopevec:
the function did not correctly check for minimum distance to travel before performing a scroll

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -758,24 +758,10 @@ class ScrollView(StencilView):
             self._touch = False
             return self.on_scroll_start(touch, False)
         ud = touch.ud[uid]
+        mode = ud['mode']
 
         # check if the minimum distance has been travelled
-        if ud['mode'] == 'unknown':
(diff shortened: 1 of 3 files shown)


kivy/uix/scrollview.py
```diff
@@ -758,10 +758,24 @@ def on_scroll_move(self, touch):
             self._touch = False
             return self.on_scroll_start(touch, False)
         ud = touch.ud[uid]
-        mode = ud['mode']
```
KeyWeeUsr: Why? You just put it on two more lines with `ud['mode']`.

darkopevec: Sorry, I have a custom renamed widget and I edited this part multiple times. Obviously I missed this.

matham:
There are unrelated rpi changes that you pushed (maybe unintentionally?).

darkopevec:
I am still learning to use git.. please see https://github.com/kivy/kivy/pull/4644

PR #4653, opened 2016-10-03
added checks to start scroll if do_scroll enabled for axis

darkopevec:
ie. scroll in y would start if touch dx > scroll_distance and do_scroll_x == False

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -761,18 +761,18 @@ class ScrollView(StencilView):
 
         # check if the minimum distance has been travelled
         if ud['mode'] == 'unknown':
+            if not self.do_scroll_x and not self.do_scroll_y:
+                # touch is in parent, but _change expects window coords
+                touch.push()
+                touch.apply_transform_2d(self.to_local)
+                touch.apply_transform_2d(self.to_window)
+                self._change_touch_mode()
+                touch.pop()
+                return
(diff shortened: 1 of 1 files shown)


kivy/uix/scrollview.py
```diff
@@ -763,8 +763,8 @@ def on_scroll_move(self, touch):
         if ud['mode'] == 'unknown':
             ud['dx'] += abs(touch.dx)
             ud['dy'] += abs(touch.dy)
-            if ((ud['dx'] > self.scroll_distance) or
-                    (ud['dy'] > self.scroll_distance)):
+            if ((ud['dx'] > self.scroll_distance and self.do_scroll_x) or
+                    (ud['dy'] > self.scroll_distance and self.do_scroll_y)):
```
KeyWeeUsr: do_scroll_x = True
do_scroll_y = False
scroll_distance > ud['dx']
scroll_distance > ud['dy']

    if ((True and True) or (True and False)):  # True, k, let's go further
        if not True and not False:                    # False, oops.

The bottom part won't be executed unless both scrollings are enabled. Was it intender or am I the only one confused here?

darkopevec: You are correct, though that check is superfluous - one is using scrollview with deliberately disabled both axis - no scrolling. For now I moved it.

PR #4657, opened 2016-10-04
Explain delay for Touchtracer

### LINKED ISSUE #4425: Touchtracer - delay when trying to draw
I don't even see it if I don't look close, but if I'm right, the thing that's mentioned in [this question](http://stackoverflow.com/q/37933920/5994041) is in fact a delay I see here. If not, there's a delay anyway. It's like there is an invisible(because bg isn't only single color) brush covering the touch - such thing you can see even in Paint as it shows you where the brush actually is.

Touchtracer seems to catch every touch, yet something slips and is displayed hm... after there's enough items in a list or something? E.g. it won't draw until there's at least 3-4+ touches in let's say `touches = []` so that a line could be drawn? Even if, you can construct a line with two points, but I see that more than two touches are not present when I drag the mouse until another part is drawn(see the red circle). This thing is present even with master and from each side(it's not an issue with misplaced touch coords).

image

KeyWeeUsr:
Closes #4425 

Yup, I'm blind and didn't see the first function's arguments back then. 🌵

### DIFF
--- a/examples/demo/touchtracer/main.py
+++ b/examples/demo/touchtracer/main.py
@@ -8,6 +8,12 @@ should see cross-hairs with the coordinates written next to them. As
 you drag, it leaves a trail. Additional information, like pressure,
 will be shown if they are in your device's touch.profile.
 
+.. note::
+
+   A function `calculate_points` handling the points which will be drawn
+   has by default implemented a delay of 5 steps. To get more precise visual
+   results lower the value of the optional keyword argument `steps`.
+
 This program specifies an icon, the file icon.png, in its App subclass.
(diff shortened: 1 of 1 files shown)


examples/demo/touchtracer/main.py
```diff
@@ -8,6 +8,12 @@
 you drag, it leaves a trail. Additional information, like pressure,
 will be shown if they are in your device's touch.profile.
 
+.. note::
+
+   A function `calculate_points` handling the points which will be drawn
+   has by default implemented a delay of 5 steps. To get more presious visual
```
matham: *precise.

PR #4661, opened 2016-10-06
setup: search /usr/lib recursively for libMali

jakogut:
When detecting platform 'mali', the current setup code uses a
predefined path to detect the presence of the userspace mali GLES
library, which works on Debian based distros.

Other distros may install the library in a different location. Arch,
for instance, installs it to /usr/lib/mali-egl/libMali.so.

This patch searches /usr/lib/ recursively for the library.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,6 +86,13 @@ if sys.platform == 'darwin':
     else:
         osx_arch = 'i386'
 
+# search for libMali.so
+libmali_path = None
+for root, subFolders, files in walk('/usr/lib'):
+    for fn in files:
+        if fn == 'libMali.so':
+            libmali_path = os.path.join(root, fn)
+
 # Detect Python for android project (http://github.com/kivy/python-for-android)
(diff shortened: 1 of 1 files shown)


tshirtman:
`libmali_path` defaults to `None`, and `exists(None)` raises an exception, shouldn't it default to `''` ?

jakogut:
@tshirtman Thanks for the feedback, you are correct.

Looking at this now, I don't think this is the proper fix for this issue. I originally created this PR because Arch Linux ARM installs libMali.so to a different location than Debian, but a recursive search could cause problems for some users, especially if it's on by default. I could see this search taking a long time on a system with a lot of installed libraries, and a slow disk.

I think a better solution here may be to search the common install locations for libMali, and allow the search path to be overridden with an environment variable.

Thoughts?

matham:
I'm closing for now until a proper fix can be implemented. I don't think searching the file system is generally a good idea as that would be quite slow.

PR #4665, opened 2016-10-09
Added softinput_mode handling for SDL2

inclement:
This works by using the SDL2 built in keyboard height handling. below_target and pan mode work easily, but resize doesn't do anything.

To get the keyboard handling you also need an up to date p4a version (pending me merging the small change necessary).

To add `resize` it would be necessary to modify PythonActivity's windowSoftInputMode to adjustResize, and also to change the java widget structure to something other than an AbsoluteLayout. Without doing that, I think the resize of the root widget works but the opengl canvas isn't itself resized, just shifted.

@akshayaurora I know you weren't enamoured with using SDL2 because of the limitations (e.g. still need more work to actually get the keyboard height within a kivy app), but I think it has the major advantages that it's very simple and easy, and faster than the old method (the shift is instant), without really any downsides.

I limited this code to set the text region on Android only because that's where it's tested. I think the same thing would work on iOS with SDL2, and I think SDL2 can also do something in relation to the software keyboard on Windows.

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -516,6 +516,8 @@ class WindowBase(EventDispatcher):
     :attr:`softinput_mode` is an :class:`~kivy.properties.OptionProperty` and
     defaults to `None`.
 
+    .. note:: The `resize` option does not currently work with SDL2 on Android.
+
     .. versionadded:: 1.9.0
 
     .. versionchanged:: 1.9.1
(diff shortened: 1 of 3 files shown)


inclement:
Actually, turns out python-for-android didn't need any changes for this to work, SDL2 really did have it all handled. So, it's good to go as far as I'm concerned.

akshayaurora:
@inclement 💯

rafalo1333:
Fantastic work @inclement, works like a charm. Thanks!

mrhdias:
Thanks and have a nice day!

PR #4667, opened 2016-10-11
Fix gstreamer links in Windows installation

KeyWeeUsr:
I know at least about gstreamer, but if even other links are broken, it can be (for now) replaced just with the drive folders until a suitable solution comes up.

This, or I can link directly to those files, but I don't think pip will be available to fetch `whl` from that kind of link without e.g. requests or other workaround.

### DIFF
--- a/doc/sources/installation/installation-windows.rst
+++ b/doc/sources/installation/installation-windows.rst
@@ -28,6 +28,11 @@ be installed for each Python version that you want to use Kivy.
 Installation
 ------------
 
+.. |gst| replace:: Google Drive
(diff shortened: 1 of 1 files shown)


matham:
I actually fixed it yesterday. The wheels are now directly on the server so
it should work like before.

KeyWeeUsr:
*_dev packages too? I can't see them in `downloads/simple`. Or is it linked to the appveyor folder?

matham:
It's on pypi. Only gstreamer wasn't because of its size.

PR #4670, opened 2016-10-14
Added proper handling for back button on Android

inclement:
With this PR, the back button will exit to the homescreen without stopping Kivy, just like the home button.

I think this is correct behaviour because, although SDL2 maps the back button to the esc keycode, it doesn't have the same implications as on the desktop. That said, I just threw this together from the code in my own app and have made the PR to raise discussion, the best solution may not be the same as this.

Something definitely needs changing, as right now kivy halts without informing Android, so it appears to have crashed and will raise an error popup when the user navigates back to the app.

Edit: I also haven't tested this exact code, even if you really love it then that needs doing before merging.

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -1399,8 +1399,24 @@ class WindowBase(EventDispatcher):
         if WindowBase.on_keyboard.exit_on_escape:
             if key == 27 or all([is_osx, key in [113, 119], modifier == 1024]):
                 if not self.dispatch('on_request_close', source='keyboard'):
-                    stopTouchApp()
-                    self.close()
+                    if platform == 'android':
+                        # back button on Android should not be treated
+                        # the same as desktop's ESC
+                        from jnius import autoclass, JavaException
+                        try:
+                            activity = autoclass('org.kivy.android.PythonActivity')
+                        except JavaException:
+                            activity = None
+                        if not activity:
+                            try:
+                                activity = autoclass('org.renpy.android.PythonActivity')
+                            except JavaErception:
+                                activity = None
+                        if activity is not None:
+                            activity.moveTaskToBack(True)
+                    else:
+                        stopTouchApp()
+                        self.close()
                     return True
 
     if Config:


kivy/core/window/__init__.py
```diff
@@ -1399,8 +1399,24 @@ def on_keyboard(self, key, scancode=None, codepoint=None,
...
+                        # the same as desktop's ESC
+                        from jnius import autoclass, JavaException
+                        try:
+                            activity = autoclass('org.kivy.android.PythonActivity')
+                        except JavaException:
+                            activity = None
+                        if not activity:
+                            try:
+                                activity = autoclass('org.renpy.android.PythonActivity')
+                            except JavaErception:
```
KeyWeeUsr: typo ^^

KeyWeeUsr:
I'm not sure if this is a good idea, because it reminds me some of the games made with Unity which didn't have back button implemented at all and the only thing that could be done was using home button, which only paused the game and left a lot of stuff in memory (not sure for how long or if it is removed later, I just purged it with force). With Kivy there's always packaged the whole interpreter, so what would happen if there were more Kivy apps on a phone with this behavior launched one after another and exited only with home button (the new back button behavior)?

Will be everything collected from memory, or will it hang there and if, for how long? Leaving app paused is a good thing (with home button as an _intentional_ pause), yet I expect a back button to _exit_ the app, not just put it in background.

inclement:
> With Kivy there's always packaged the whole interpreter, so what would happen if there were more Kivy apps on a phone with this behavior launched one after another and exited only with home button (the new back button behavior)?

Android would quietly halt the app and reclaim the memory if it ever needs it. This is the recommended Android activity model.

KeyWeeUsr:
What about data loss while paused and the memory would get reclaimed? I haven't really poked into how Java apps handle the back button, but if there's no real danger making the back button behave like that, I'm for that. Except the potential memory & data issues I mentioned it can only bring the good stuff such as quicker loading (if in memory) and converging to the Java Android app model. 👍

inclement:
When the app is paused, it gets a few seconds to store data (via
onPause), but after that Android provides no guarantee that it will ever
run again. It is expected as part of this execution model that if you
don't save data, it may be lost.

dessant:
Will it be possible to cleanly exit without pausing when the back button is pressed?

rafalo1333:
@inclement in the standard Android Activity model, when you press back button, Activity's onDestroy() is called and activity is killed. Only when you'll override onBackPressed(), you can achieve behavior from the PR. While I like it, it is not the default that Android activities are working like.

@dessant to achieve that you'll need to bind to Window's on_keyboard, and from there call app.stop() when key 27 is pressed.

KeyWeeUsr:
I again feel this is not a good idea mainly because of [Activity lifecycle](https://developer.android.com/training/basics/activity-lifecycle/recreating.html):

> There are a few scenarios in which your activity is destroyed due to normal app behavior, such as when the user presses the _Back_ button or your activity signals its own destruction by calling finish().

> When your activity is destroyed because the user presses _Back_ or the activity finishes itself, the system's concept of that Activity instance is gone forever because the behavior indicates the activity is no longer needed.

These quotes show that the current behavior of back button in Kivy app is actually the most accurate if compared to the Java Android app.

dessant:
@inclement, @KeyWeeUsr: so, was it settled? 😋

KeyWeeUsr:
Well, all apps seem to have back button to exit/close an app, even android seems to do so in its docs, so I don't feel it's a good change. Let's leave this behavior to _Home_ button.

inclement:
I realise I was wrong about this being the most correct thing to do according to the docs, but I still think we should consider it - it seems that one reason for the java model is that the java ui can be quickly regenerated, but restarting a Kivy app means completely restarting the python process which is very slow. I think apps like games often override the back button in the same way, presumably for the same reason

KeyWeeUsr:
But this way everything remains as if the app was paused, which just kills everyone else's app that might depend on cleaning the environment properly each time. Maybe a better approach would be a property set in `App.build()` or in `config.ini`?

inclement:
It should be toggleable, but I'm not sure there's really a concern about
killing other peoples apps - it depends on how Android decides what to
halt, but it seems more likely that the Kivy app would be the first one
killed if it's consuming resources disproportionately.

KeyWeeUsr:
Ah, bad wording. Let's say you make a casual application, not a game. You have there some variables that need to reset each time an App starts e.g. you fetch some token, get a key from url or something. This change doesn't allow the app to exit properly, only relies on Android taking care of it and

> ... that if you don't save data, it may be lost.

in the worst case Android might even restore the data e.g. if you have `on_pause` enabled, press home button and quickly run the app again. I see the same thing I left it when paused. When a user works with files it might be even less desirable just as out-of-box feature. That's why if it should be included, let user decide how to handle this stuff with Config.

germn:
I think this idea makes sense since Kivy app starts much longer then Java-based Android apps.

On other hand if app uses to process files, this PR right now may [lead to crush](https://github.com/kivy/python-for-android/issues/967).

KeyWeeUsr:
What about `Config.set('kivy', 'pause_on_minimize', 1)`? Wouldn't be better to reuse this functionality to pause the app instead of making back button behave as the second home button?

dessant:
@KeyWeeUsr that's more about pausing the event loop or keeping it runnig, but you don't have a choice for that on mobile afaik. Respecting `exit_on_escape` might be a better fit.

KeyWeeUsr:
I meant something like this:

    if WindowBase.on_keyboard.exit_on_escape:
        if key == 27 or all([is_osx, key in [113, 119], modifier == 1024]):
            if not self.dispatch('on_request_close', source='keyboard'):
                if platform == 'android' and Config.getbool('kivy', 'pause_on_minimize'):
                    # back button on Android should not be treated
                    # the same as desktop's ESC
                    from jnius import autoclass, JavaException
                    .
                    .
                    .
                else:
                    stopTouchApp()
                    self.close()
                return True

Which basically means a user would have an option to toggle this functionality.

dessant:
There isn't a way to signal in java that the app wants to exit *now* instead of waiting for the process to be stopped by android?

dessant:
@KeyWeeUsr, yeah, but if we do that it just means we hide the bug by default instead of fixing it, just like the pr does in its current form. I agree about this being an undesired change.

KeyWeeUsr:
It's not a bug, it's just Kivy being loaded a little bit more than desired, which I think we can't really do anything about it except attempt to hide it like in this PR or `on_pause->return True` or another alternative.

I do agree that it's not nice to have hanging app just because it loads long enough for an end-user to notice and if I had some big app with Kivy I'd probably make it pause by default too, but there shouldn't be just a big switch from a common android behavior to pausing everything just because Kivy can become a loading beast similar to Unity if you really want to make it so.

Hence the bool option. Maybe `pause_on_minimize` isn't the best choice and another variable directly related to this functionality would be better.

Also, about the signal in java, I don't know anything except [`finishActivity()`](https://developer.android.com/reference/android/app/Activity.html#finishActivity(int)) and system exit(like `exit()` in py), which probably won't help here.

kollivier:
Just came across this during my deep dive into the p4a bootstrap behaviors. As someone who's been using cross-platform toolkits for many years now, I've found that I always want to be given the choice of what to do in my app code. e.g. I'd just like to have some on_back_key event or on_android_back_key to stress that it's an android-specific callback. If the callback isn't handled, then close the app.

The 'close the app' behavior is generally correct in the case when you're at the app's home screen, but, for example, if you're building an ebook reader and the user has started reading a book, users would probably expect the back key to take them back to the book list. It also serves as a "go to previous screen" button. That's how Kindle, Chrome and the Settings app works. 

Once that callback is available, devs who don't want their app to fully close on back would have the choice to change that behavior as well. And then if they get complaints about users that back should close the app and free resources, then they can change it back too. ;)

FeralBytes:
I just wanted to say; I do not like the idea of Config setting as compared to a method or callback. Callbacks give you the power to be dynamic like in the e-reader example above. But the Config settings are hard coded and can not be changed after loading Kivy.

akshayaurora:
Current setup allows you listen to escape key like [so](https://github.com/pythonindia/PyCon-Mobile-App/blob/master/eventsapp/utils/keyboard.py) and specify your own behaviour of what to do? Go Back in Screen Higherarchy or to pause app  like [so](https://github.com/pythonindia/PyCon-Mobile-App/blob/master/eventsapp/utils/__init__.py#L15)

or to follow the default behaviour of closing the app as the it does for a Java App.

This issue has been open for a long time; Closing the app is the expected behaviour now for a long time. 

I feel the correct solution here is to give a example in doc on how to customise your app to change default behaviour If the developer so chooses.

RobinPicard:
Do you think it would make sense to create an event method that would be fired when the back button is pressed on Android and that the user could overwrite to implement their own handling (just like what is done with on_pause for instance)? I was thinking it could be easier to use, especially since I think handling this back button is a quite common use case @akshayaurora

RobertFlatt:
@RobinPicard For custom back button/gesture handling, one way is to bind the keyboard and test for key == 27 , like this <https://github.com/Android-for-Python/Android-for-Python-Users#back-button-and-gesture>

Btw, I just submitted a PR that addresses a different issue with the back button/gesture. https://github.com/kivy/kivy/pull/7989#issuecomment-1231037983

RobinPicard:
> #7989 (comment)
I use the binding of the keyboard and the test on test for key == 27 for my own app and it works fine, but I just thought there could be a way of making it more accessible by already doing a part of the work with something like that (before you own modifications):

    ```def on_keyboard(self, key, scancode=None, codepoint=None,
                    modifier=None, **kwargs):
        [...]
        if WindowBase.on_keyboard.exit_on_escape:
            if platform == 'android' and key == 27:
                if not self.dispatch('on_request_close', source='keyboard') and not self.dispatch('on_back_key_android'):
                    stopTouchApp()
                    self.close()
                    return True
        [...]```

    ```def on_back_key_android(self):
        pass```

Then the user would just have to add `Window.on_back_key_android = custom_function` to handle what happens when the back button is pressed (just like what you'd do with on_enter, on_pause...)

Issue #4535, opened 2016-08-13
kivy.uix.videoplayer attempts to load source path as Image when loading file


Bakterija:

dessant:
This seems to happen because Video is subclassed from Image, so when setting Video.source, the loading logic of Image is also executed.

rayrase:
I'm new to open source and feel this is something I can fix. Is this issue still up for grabs?

dessant:
@rayrase, yes, of course.

rayrase:
Ok, thank you!

BittaBoesaBiba:
Same issue here. Can be reproduced with `Video` and `VideoPlayer` respectively

```
Screen:
    Video:  # or VideoPlayer
        size_hint: None, None
        size: dp(640), dp(380)

        # https://sample-videos.com/video321/mp4/360/big_buck_bunny_360p_1mb.mp4
        source: 'big_buck_bunny_360p_1mb.mp4'
        on_touch_up: self.state = 'play'
```

Fix would be to adjust `__init__` like
```
    def __init__(self, **kwargs):
        self._video = None
        self.source = kwargs.pop('source')  # <----- this
        super(Video, self).__init__(**kwargs)
        self.fbind('source', self._trigger_video_load)

        if "eos" in kwargs:
            self.options["eos"] = kwargs["eos"]
        if self.source:
            self._trigger_video_load()
```

For now one could override the init in a subclass.
```
from kivy.uix.video import Video
class GentleVideo(Video):
    """Prevents warning
    Image: Error loading <path/to/file.mp4>
    """
    def __init__(self, **kwargs):
        self._video = None
        self.source = kwargs.pop('source')
        super(Video, self).__init__(**kwargs)
        self.fbind('source', self._trigger_video_load)

        if "eos" in kwargs:
            self.options["eos"] = kwargs["eos"]
        if self.source:
            self._trigger_video_load()
```

I'll see if I can quickly create a pull request.

PR #4679, opened 2016-10-18
Issue #4535 kivy.uix.videoplayer attempts to load source path as kivy.uix.image when loading file

rayrase:
I added a check for the file extension of source in Image **init** method.

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -516,6 +516,8 @@ class WindowBase(EventDispatcher):
     :attr:`softinput_mode` is an :class:`~kivy.properties.OptionProperty` and
     defaults to `None`.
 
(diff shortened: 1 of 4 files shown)


tshirtman:
1/ I don't think that's the right fix, every core image provider can define a list of file extensions it supports, so you can't put a list here, especially such a short one.

2/ somehow other changes went into the commit, i don't think that was intended.

rayrase:
Ok, thank you very much for your feedback. You're much more knowledgable than I am, what would you suggest I try? I'm thinking I should make Video a subclass of Widget and make it so it's not moveable like an Image. 

Again thank you very much!

dessant:
As @tshirtman said, this looks like the wrong fix, let's continue tracking the issue at https://github.com/kivy/kivy/issues/4535.

PR #4686, opened 2016-10-21
changed VKeyboard's 'font_size' to an external parameter

johnyburd:
Allows manual setting of VKeyboard's font_size parameter for cases when size based on width doesn't work.

### DIFF
--- a/kivy/uix/vkeyboard.py
+++ b/kivy/uix/vkeyboard.py
@@ -252,6 +252,14 @@ class VKeyboard(Scatter):
 
     '''
 
+    font_size = NumericProperty('20dp')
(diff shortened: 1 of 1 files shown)


tito:
This is not going to work. Before, the font size was resolved during layout, where self.width is known. Now by setting the font_size in init, it has no purpose because, self.width is not known (100 by default), so the default value is replaced with a wrong value anyway.

IMO, create a new variable like "fixed_font_size = BooleanProperty(False)", and use either the font_size set if fixed_font_size is True or the dynamic font size if fixed_font_size is False.

Please reopen with the new changes :)

PR #4689, opened 2016-10-21
Docs: popup has on_open event, not on_enter

rafalo1333:


KeyWeeUsr:
Nice catch! Niether `popup.py`, nor `focus.py` mention `on_enter`.

PR #4694, opened 2016-10-29
Compile Kivy with MSVC

KeyWeeUsr:
As a lot of you know, there's a problem with MinGW/Python for 3.5 and higher versions which if you use multiple frameworks/packages and want to stay up-to-date ends up with having multiple Python installations. That's not nice.

There are experimental wheels for stable version, but that has a lot of differences, therefore the newest version is sometimes more desired than relying on stable, so I kind of hope we can hammer some instructions together for compiling Kivy in MSVC which could result in making wheels with MSVC soon-ish after that.

I tried it for Win8.1 x64 for Python 3.5, 32bit, so it might be a little bit different with the SDK versions, but the rest should be the same:
## Environment
1. Get the [compiler](http://landinghub.visualstudio.com/visual-cpp-build-tools) (~5GB total size)
2. Open `cmd.exe` in `<python.exe dir>` (`shift` + right click)
3. Upgrade pip and setuptools*, [get deps](https://drive.google.com/drive/folders/0B1_HB9J8mZepdUl0bEZubXZiSUE) and install them
4. Run `"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat"`
5. Set environment:
   
   ```
   set MSSdk=1
   set USE_SDL2=1
   REM ^^^^^^ and USE_GSTREAMER if used
   set DISTUTILS_USE_SDK=1 
   set LIB=%cd%\libs;%LIB%
   set INCLUDE=%cd%\include;%INCLUDE%
   ```
6. `git clone https://github.com/kivy/kivy`

*`The setuptools Python package version must be at least 24.0.` according to Python's [wiki](https://wiki.python.org/moin/WindowsCompilers)
## Compile Kivy
1. `python -m pip install kivy\` (the local path, not "kivy" as stable from pypi)
2. Still some magic missing

Returns a [link error](http://pastebin.com/7rqXjEG5) for now, but I don't have a clue what/how to link the missing thing.

### DIFF
--- a/doc/sources/installation/installation-windows.rst
+++ b/doc/sources/installation/installation-windows.rst
@@ -212,7 +212,49 @@ with git rather than a wheel there are some additional steps:
 MSVC
 ~~~~
 
-...
+.. warning::
+   The branch is still in progress and some things might not work when compiled
+   with msvc.
+
+.. |msvc| replace:: compiler
+.. _msvc: http://landinghub.visualstudio.com/visual-cpp-build-tools
+.. |py3deps| replace:: py3.5 deps
+.. _py3deps: https://drive.google.com/drive/folders/0B1_HB9J8mZepdUl0bEZubXZiSUE
+.. |gl-back| replace:: feature-gl-backend
+.. _gl-back: https://github.com/kivy/kivy/pull/4385
+
+Environment
+^^^^^^^^^^^
+
+1. Get the |msvc|_ (~5GB total size)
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
Just so that it's more obvious, the complete instructions are visible [here](https://github.com/KeyWeeUsr/kivy/blob/606445f7a66e15ce4e28cc75c43cb976efc3f955/doc/sources/installation/installation-windows.rst#msvc). 😋

Issue #4695, opened 2016-10-29
Sequential Animations aren't removed from Animation._instances when complete


scastette:
Sequential animations are not removed from the Animation._instances collection when they complete.

PR #4696, opened 2016-10-29
Added test and fix for Issue #4695

scastette:
Sequential Animations were not being removed from Animation._instances when complete.
Also, test for WeakRef animations did not run GC.


tito:
Thanks you!

PR #4698, opened 2016-10-30
Fix pointer mismatch

KeyWeeUsr:
Something is wrong with passing `wglGetProcAddress` like a parameter via `void *(*f)(const char *)`. I tried to add/remove `*` or even refactor it a little bit, but still it either let the whole compilation explore or returned this:

```
    cgl_glew.c
    C:\Users\X\AppData\Local\Temp\pip-qse2svnt-build\kivy\graphics\cgl_glew
.c(2513): error C2440: 'function': cannot convert from 'PROC (__stdcall *)(LPCST
R)' to 'void *(__cdecl *)(const char *)'
    C:\Users\X\AppData\Local\Temp\pip-qse2svnt-build\kivy\graphics\cgl_glew
.c(2513): warning C4022: '__pyx_f_4kivy_8graphics_8cgl_glew_gl_dynamic_binding':
 pointer mismatch for actual parameter 1
     error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\
BIN\\cl.exe' failed with exit status 2
```

It's probably simple to fix it, I just don't see it there maybe. Anyway I saw no visible restriction like pep8 for the cython files, therefore maybe it's a usable fix until someone finds a way how to do it the fancy way 😋


dessant:
@matham, ping.

matham:
The reason for the indirection is so that glew would not be required at all during compiling/linking. I.e. you could build it on windows without the glew dlls and then load dynamically. Although right now, it doesn't do that anyway because `wglGetProcAddress` is accessed directly.

The solution is probably to change `cdef void gl_dynamic_binding(void *(*f)(const char *)) except *:` to something like `cdef void gl_dynamic_binding(void *(stdcall *f)(const char *)) except *:`

KeyWeeUsr:
@matham Nah, doesn't seem to help. With that or a similar guessing the best result I got was this:

void *(__stdcall *f)(const char *) without `&`:

             gl_dynamic_binding(wglGetProcAddress)
                                                ^
     ------------------------------------------------------------

     kivy\graphics\cgl_glew.pyx:38:44: Cannot assign type 'void *(const char *)'
     to 'void *(__stdcall *)(const char *)'

with:

             gl_dynamic_binding(&wglGetProcAddress)
                               ^
     ------------------------------------------------------------

    kivy\graphics\cgl_glew.pyx:38:27: Cannot assign type 'void *(*)(const char *)'
    to 'void *(__stdcall *)(const char *)'
    building 'kivy.graphics.cgl_glew' extension

In other cases it returns the same error with `PROC`.

matham:
Hmm, I guess I'll have to test this as it worked for me before. What version of python are you running? I hope not 3.5.

The reason for dynamic linking is that we would like to be able to compile kivy without glew being present at all and then at runtime be able to provide glew, angle etc and it'd just work. I.e. imagine wanting to use it with angle, yet you still needed glew present. Of course you may disable glew during the build so it would not be needed and you couldn't use it. But it would nice to be able to switch backend providers at runtime and not have to decide at compile time.

KeyWeeUsr:
Exactly 3.5, but it's kind of weird that bug or how to call it if it works on lower versions.

Anyway, I see now, this is a bad fix. Yet, it makes py3.5 work even if it's experimental. Angle isn't ready yet, am I right?

Issue #4442, opened 2016-06-30
Feature Request: Multiple Windows For Desktop Applications.


phil-andrews:
This would be absolutely huge. I desperately want to use Kivy to build a number of applications but this hold back is the one and only reason I can't do it. There would be no reason to use another GUI framework.

inclement:
This has come up before - I believe it's likely to be non-trivial, as Kivy's internals were written to not worry about multiple window states, I think as a deliberate choice when it emerged that full multi-window support (with multiple opengl contexts etc) would be hard. A PR to add this feature or work towards it would probably be welcome, but for these reasons I don't think any core dev is likely to look at it soon.

KeyWeeUsr:
I'd like to add that you can do pseudo-multiwindow with kivy in three ways(app = file with `<main class>().run()`):

1. Single app, one space - widgets as forms("windows")

2. Single app divided into admin & client parts launched multiple times where admin would probably handle communication

3. One+ app (launching forms from running app):
  - Running another interpreter instance via subprocess (`python another.py` or similar)
  - Running an app via multiprocessing (if crashes/freezes, probably all will go down)

However it's not that hard to have multiple windows with kivy as it is with those steps, rather the communication is (or at least for me) kind of problematic and non-elegant (read _ugly_).

Also, I don't want it to sound rude, but what is that special snowflake use case that _needs_ multiple windows?

phil-andrews:
You're right, opening a second window isn't terribly difficult. You can do it in about 12 lines of code with `subprocess`. You can even handle killing it with `subprocess`. The issue is passing data back and forth between windows. 

The scenario is multiple displays. I need to display a window on screen two that takes in data from the client portal on screen one. Then, the data needs to be refreshed as the client sees fit. It's not massively complicated but passing data back and forth is, you're right, ugly.

KeyWeeUsr:
That's why I mentioned that bad bad communication :yum: What you may try is using a local server between opened windows. I've had this idea in mind since I've seen (now it's called) [remi](https://github.com/dddomodossola/remi), which is nicer than using e.g. files or accessing the app with some other ugly way. This way you may write app that autoconnects & checks something `on_start()`

[...]

PR #4699, opened 2016-10-31
Add MultiForm demo/tutorial

KeyWeeUsr:
Similar to #4442 feature request, but this one is about one app with multiple window-like widgets. Around ~200lines, but could be probably shorter.

Except for the obvious, I mentioned there:

* traversing parent-children tree from bottom (from root it's just annoying, hungry and sometimes didn't work for me, to I excluded it)
* post_initialization (because `self.parent` is `None` in `__init__`)
* working with WeakProxies
* hiding widgets
* how properties work

... and so on. I just hope it's not confusing much ^^

image

### DIFF
--- a/examples/demo/multiform/multiform.kv
+++ b/examples/demo/multiform/multiform.kv
@@ -1,0 +1,115 @@
+#:import dp kivy.metrics.dp
+#:import Clock kivy.clock.Clock
+
+<Area>:
+    BoxLayout:
+        size_hint_y: root.formbar_height
+
+        ScrollView:
+            size_hint_x: 0.9
+            GridLayout:
+                id: formbar
+                size_hint_x: None
+                width: self.minimum_width
+                rows: 1
+
+        Button:
+            text: 'Create!'
+            size_hint_x: 0.1
+            pos_hint: {'center_x': 0.5, 'y': 0}
+            on_release: root.create_form()
+
+<FormItem>:
+    size_hint_x: None
+    width: dp(100)
+
+<Form>:
+    size_hint: [None, None]
+    size: [400, 350]
+    do_rotation: False
+    auto_dismiss: False
+
+    BoxLayout:
+        id: body
+        size: root.size
+        orientation: 'vertical'
+        canvas:
+            Color:
+                rgb: (0.1, 0.1, 0.1)
+
+            Rectangle:
+                size: self.size
+                pos: self.pos
+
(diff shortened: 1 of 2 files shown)


dessant:
Wouldn't this be happier as a garden flower?

KeyWeeUsr:
I've thought about it too, but I wasn't sure about it as in garden there is code that you can use (install from it) and just import & go on with your stuff. This is rather demo. I don't mind having it in garden though. :)

dessant:
A part of me would love to see the `examples` folder moved to a `kivy-examples` repo and let it grow more freely, while not clogging the main kivy repository. What do you guys think?

KeyWeeUsr:
Well, I still love how Kivy automatically installs the examples too, so if this doesn't change, it might be a good thing. But then again, if it's a separate repo, it might be less "controlled" and less useful examples might come in. If that means this demo won't be merged because it's on the edge with its usefulness, then rather close it instead of making a new repo that could become congested later.

I may even move it to wiki if that's the case ^^

dessant:
I'm also ok with merging, the code looks ok. My main issue is that we could come up with a large number of examples of this category, and I wouldn't know how to choose and which one to keep. Would you?

KeyWeeUsr:
Most likely I wouldn't be able to choose which one should be included or not (except obvious stuff). However, I'm not sure if this is the way examples should go. It's just a feeling though and I'd like to see the examples grow too. Maybe even in a separate repo. I'm just judging from my own experience as sometimes I'd like to include a lot of stuff and sometimes not so much because making it is just too obvious even without demo.

Don't get me wrong, I'm kind of for both (separate or not) solutions, I'm just unsure what would be the right one.

dessant:
Yeah, these are the hardest decisions. 😋 I regard the examples folder as a place for showing only how to use the kivy api, basic layouts etc, without complex details.

There are already some odd choices, like pyjnius apps: https://github.com/kivy/kivy/tree/master/examples/android. These things could be moved to a kivy-examples repo, where we could import external libs, or show specific ui setups etc. The wiki also has several examples which could belong there.

I'm not sure how the rest of the team feels about this, but I believe some kind of distinction has to be made.

dessant:
@KeyWeeUsr, if not the garden, then this could live in the wiki until some new structure is figured out for examples. I'm hesitating, but I'll close this. 👼

PR #4707, opened 2016-11-04
allow to grab mouse in sdl2 window

mahomahomaho:
It's useful in windowed game, to not able mouse cursor to leave window. This methods allows it.


dessant:
Nice work, this could also be a boolean property.

tshirtman:
looks good to me.

PR #4708, opened 2016-11-05
Get actual image format instead of extension

ismailof:
Allow to load any image file by guessing its type from file metadata instead of extension, using the builtin utility imghdr.what('filename').  Using this, the specific loader is always selected, not depending on the file extension, which can be a problem when loading from servers where extension type is not controlled.


dessant:
This would only work reliably if `imghdr` would be extended to recognize all file types supported by the different image providers. The list of formats recognized by default also differs depending on the python version.

https://github.com/ismailof/kivy/blob/f0f45e29c7e6ac1ceb54fe2391169d5aad509e54/kivy/core/image/img_sdl2.py#L22
https://github.com/ismailof/kivy/blob/f0f45e29c7e6ac1ceb54fe2391169d5aad509e54/kivy/core/image/img_tex.py#L17
https://github.com/ismailof/kivy/blob/f0f45e29c7e6ac1ceb54fe2391169d5aad509e54/kivy/core/image/img_pygame.py#L29
https://github.com/ismailof/kivy/blob/f0f45e29c7e6ac1ceb54fe2391169d5aad509e54/kivy/core/image/img_pil.py#L45
https://github.com/ismailof/kivy/blob/f0f45e29c7e6ac1ceb54fe2391169d5aad509e54/kivy/core/image/img_ffpyplayer.py#L47
https://github.com/ismailof/kivy/blob/f0f45e29c7e6ac1ceb54fe2391169d5aad509e54/kivy/core/image/img_dds.py#L16

ismailof:
I made a simple change to default back to extension if the image format cannot be guessed by imghdr. This way all of the formats can still be supported, but the most common ones (those on imghdr) does not rely on the file extension. 
Do you think that this would be enough by the moment? imghdr can still be extended for extra cases if necessary.
My personal motivation for this PDR is that I have to retrieve images (usually jpg or png) from a server where the extension does not match the format, so I was getting `Unknown <obj> type, no loader found.` exceptions.

dessant:
I have not tested, but I believe this should work now. 👍

dessant:
@ismailof, we could also do the same here for the bytestream: https://github.com/ismailof/kivy/blob/77c6cdac069a4ecd772a7040ea72e40a18525e91/kivy/core/image/__init__.py#L332.

ismailof:
Ok, I'll take a look on it. Adding the imghrd part doesn't seem complicated.
In addition, the ImageLoader code looks a little bit messy to me, as it plays with atlases, zips and loaders in a different fashion but kind of alltogether, so if you don't mind, I would like to reorder it on a different PR.

Issue #2325, opened 2014-07-11
Image Texture not freed on delete?


Xeno22:
I load many images from files on my raspberry, at some point, it does not load anything anymore. Also I can't start my app afterwards because of an open gl init error.
I tracked it down to the massive textures in the GPU memory.

I load the image this way:

```
         self.image = Image(source = cachedFile, allow_stretch = fullscreen,
                          keep_ratio = True, mipmap = False,
                         pos = self.pos, size = self.size, keep_data = False,
                         nocache = True
                         )
```

and remove this way:

```
     def remove_oldimage(self, *l):
         print "remove oldimage"
         self.remove_widget(self.oldimage)
         del self.oldimage
         self.oldimage = False

     def removeImage(self, effect, *l):
         if self.image != False:
             self.oldimage = self.image
             self.image = False
             if effect != None:
                 anim = Animation(opacity = 0, t=effect)
                 anim.bind(on_complete = self.remove_oldimage)
                 anim.start(self.oldimage)
             else:
                 self.oldimage.opacity = 0
                 self.remove_oldimage()
```

So my question: did I do something wrong while using the Kivy API, or do I really have to deallocate the texture by myself (and how)?
##

derekstavis:
I have been experiencing the same problem. Seems like that when there's no space left for textures, the program goes pretty crazy, with missing textures blinking left and right, screen managers switching back and forth, etc. Killing and reopening the app deallocates all textures and everything work again.

bionoid:
Maybe this issue was fixed by #5664

Julian-O:
No-one has complained for 5 years. Assuming it was fixed.

If you think this is in error, please let us know.

PR #4709, opened 2016-11-06
Cache improvements

tshirtman:
This didn't get a lot of testing (only ran touchtracer/showcase/kivycatalog), so more are welcome, in short.

- Added a few tests cases for Cache
- Found out the `limit` argument wasn't used at all, so only the timeout would prevent the cache from growing indefinitly (if applied)
- Fixed that and the purge logic that was disabled because of being broken
- removed a few try/except/pass that didn't seem necessary (so extra tests are very welcome)
- added another limitation system, allowing one to define a function to compute the size of each element, and call the same purge logic as for the limit argument, when the total size of a cache outgrow its defined limit, this does add some overhead both in computation and memory footprint, put is entirely optional and shouldn't add much overhead when not used (which is the default).
- converted print statements to Logger.debug/trace and added a few more
edit:
- The PR now tries to use the feature with texture cache, adding a setting in kivy to configure a max size, sadly, while the feature seems to work at removing textures from the cache, it doesn't seem to make the gpu release the textures, so it's not really useful at the moment.

### DIFF
--- a/kivy/cache.py
+++ b/kivy/cache.py
@@ -103,7 +103,6 @@ class Cache(object):
             return
         timeout = timeout or cat['timeout']
 
-        # FIXME: activate purge when limit is hit
         limit = cat['limit']
         if limit is not None and len(Cache._objects[category]) >= limit:
             Cache._purge_oldest(category)
--- a/kivy/cache.py
+++ b/kivy/cache.py
@@ -39,7 +39,8 @@ class Cache(object):
     _objects = {}
 
     @staticmethod
-    def register(category, limit=None, timeout=None):
+    def register(category, limit=None, timeout=None, max_size=None,
+                 compute_size=None):
         '''Register a new category in the cache with the specified limit.
 
         :Parameters:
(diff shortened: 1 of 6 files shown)


dessant:
May be related: https://github.com/kivy/kivy/issues/2325.

dessant:
ref https://github.com/kivy/kivy/issues/4739.

dessant:
@tshirtman, would you like to have this for 1.10 or should we retag?

tshirtman:
I'm not sure about merging the last commit, which don't seem to achieve its objective, but the rest could be merged.

dessant:
@tshirtman, you've mentioned that part before, what i'm asking is if you'll clean up the pr for this release or should we retag. :)

dessant:
Ok, retagging.

dessant:
The relevant parts were merged in https://github.com/kivy/kivy/pull/5107.

dessant:
@tshirtman, I've missed this also contains some new features, please separate them when you have time.

PR #4716, opened 2016-11-08
Allow window creation with sub 24-bit color

### LINKED ISSUE #4715: Can't create window on displays that don't support 24-bit color
I'm working on a Kivy app that runs on an FBTFT display. The display only supports 16-bit color, and Kivy fails to create a Window on it. I found a workaround, as detailed here:

http://forum.odroid.com/viewtopic.php?f=145&t=24315&p=166820#p166820

Essentially, allowing Kivy to acquire a window with lower color depth works. Is there a reason Kivy requires 24-bit color by default? Can we drop this requirement?

jakogut:
Closes #4715

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -94,16 +94,6 @@ cdef class _WindowSDL2Storage:
                 'LandscapeLeft LandscapeRight')
         SDL_SetHint(SDL_HINT_ORIENTATIONS, <bytes>orientations)
 
-        SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1)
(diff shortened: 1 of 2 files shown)


jakogut:
I'm not sure if this is the proper solution. Can anyone think of a reason to explicitly specify a minimum color depth during compilation? This seems to me to be more a runtime/display configuration issue.

janssen:
There's some image code, as in Widget.export_to_png(), which eventually calls kivy.core.image.Image.save(), which assumes 8 bits per color.

tito:
Removing completely the requirements is not going to work. Because we enforced the color and depth, then you can be sure some widget implementation works. Removing everything can also lead to a surface that doesn't meet our requirements (having depth buffer, stencil buffer 8 bits etc).

The original issue can't be solved with this PR. A minimal approach would be to have a configuration option to activate this behavior.

PR #4718, opened 2016-11-08
Fix auto_dismiss

KeyWeeUsr:
return causes ignoring auto_dismiss if passed in e.g. kv via <ActionDropdown>:, however the on_touch_down is still necessary to dismiss it properly, which has a side effect of dismissing when an item's on_release or similar are triggered.

### DIFF
--- a/kivy/uix/actionbar.py
+++ b/kivy/uix/actionbar.py
@@ -264,7 +264,6 @@ class ActionDropDown(DropDown):
         if super(ActionDropDown, self).on_touch_down(touch):
             if self.auto_dismiss:
                 self.dismiss()
-            return True
 
 
 class ActionGroup(ActionItem, Spinner):


KeyWeeUsr:
Someone please run the `actionbar.py` with `<ActionDropDown>`. I wouldn't like this bug to get into the new version. Try with and without `return True` in `ActionDropDown.on_touch_down()`, this PR removes the `return`.

```
if __name__ == "__main__":

    # XXX clean the first registration done from '__main__' here.
    # otherwise kivy.uix.actionbar.ActionPrevious != __main__.ActionPrevious
    Factory.unregister('ActionPrevious')

    Builder.load_string('''
<MainWindow>:
    ActionBar:
        pos_hint: {'top':1}
        ActionView:
            use_separator: True
            ActionPrevious:
                title: 'Action Bar'
                with_previous: False
            ActionOverflow:
            ActionButton:
                text: 'Btn0'
                icon: 'atlas://data/images/defaulttheme/audio-volume-high'
            ActionButton:
                text: 'Btn1'
            ActionButton:
                text: 'Btn2'
            ActionGroup:
                text: 'Group 2'
                ActionButton:
                    text: 'Btn3'
                ActionButton:
                    text: 'Btn4'
            ActionGroup:
                text: 'Group1'
                ActionButton:
                    text: 'Btn5'
                ActionButton:
                    text: 'Btn6'
                ActionButton:
                    text: 'Btn7'
<ActionDropDown>:
    auto_dismiss: False
''')

    class MainWindow(FloatLayout):
        pass

    float_layout = MainWindow()
    runTouchApp(float_layout)
```

Issue #4682, opened 2016-10-19
Linux install doc - input group


mixmastamyk:
Under a recent Ubuntu, I found I needed to add my user to the "input" group to get access to things like the track pad.  A short note to the Linux install doc would be helpful:

```
sudo adduser $USER input  # log out then back in to apply
```

Otherwise you'll see lots of errors such as: `permission denied accessing  /dev/input/eventX…`

Zen-CODE:
@KeyWeeUsr. I once submitted a PR to present a message instead of throw an error, but someone felt that might make it incompatible. I don't think that's legitimate: it's bad practice to rely on errors. 

I will prepare anotherPR that logs these messages instead of throwing an error. I think that makes much more sense as errors are things you cannot predict. This is predictable, and not an error so much as a permissions issue.

KeyWeeUsr:
@Zen-CODE I put the documentation+easy tag here, because the adding user to input seemed like a solution for the permission errors too. Or just a note about that in docs. A beginner won't most likely know what's going on except "permission denied" and I think it could lead to opening Kivy apps with `sudo` because of an error just to make things smooth in the log. ^^

Zen-CODE:
Yes, I am also going to add that to the docs. But even with that in the docs, I think it's wrong for an error to be thrown. It's really not an error, it's a permission issue so should appear as such. It is unnecessary to spit such an alarming to the console and has been the cause of many posts and questions. 

Agreed? Or do you think it should throw and error as it currently does?

KeyWeeUsr:
I'm not really sure if it should be thrown. As I read the questions it doesn't seem to cause problems except for that denied access, which is probably only for fetching devices, therefore catching and printing is the good thing to do, right?

Zen-CODE:
If you look at the PR, you'll see two posts by people alarmed by the message. So was I when I first started using kivy (also being new to linux). And those are just two I could google quickly. There have also been many questions on the forums about it, because it does throw an error and people think something is wrong.

I was actually going to ask you to test and merge it :-) I think it causes too much confusion to see errors spat our directly, as if they are not expected and need attention. They are expected, and I think we should handle them as such....

But yes, will get to the docs (hopefully) tonight.

mixmastamyk:
I didn't like the error either, but I gathered that since it had been ignored for years someone supported it.  Therefore I though docs could help.  But, if you'd like to log it instead, by all means.  The log message might include reference to the "input" group if it needed under other popular distros as well.

[...]

Issue #4413, opened 2016-06-19
Is this normal? 


h2Devel:
So I was running manjaro and kept getting this error. I tried everything I could to track it down, but it didn't seem too common. So I wiped my entire system, and loaded a fresh/clean install of Ubuntu 16.04. 

Copied the hello world text from the homepage and got the same thing, then copy and pasted the pong source code. It runs, but gives me this error every time. Am I missing something here?

[kivy_output.txt](https://github.com/kivy/kivy/files/322160/kivy_output.txt)

dessant:
@matham, is this still valid? https://github.com/kivy/kivy/issues/1503#issuecomment-48820806

dessant:
@h2Devel, do you have mtdev installed?

h2Devel:
Just installed 'mtdev-tools', and 'libmtdev-dev', and still no dice.

h2Devel:
Oh.... But I'm not using a touch screen.... And from what I've gathered, the permission error has something to do with the track pad of some laptops. What I'm mostly concerned with is the threading issue.

matham:
@dessant no clue.

h2Devel:
updated a bunch of stuff, and still no change.

from kivy.app import App


class MyWindowApp(App):
    pass

window = MyWindowApp()
window.run()

=============================

/usr/bin/python3.5 /home/harley/PycharmProjects/Kivy/basic.py
[INFO              ] [Logger      ] Record log in /home/harley/.kivy/logs/kivy_16-06-20_10.txt
[INFO              ] [Kivy        ] v1.9.2-dev0
[INFO              ] [Python      ] v3.5.1+ (default, Mar 30 2016, 22:46:26) 
[GCC 5.3.1 20160330]
[INFO              ] [Factory     ] 193 symbols loaded
[INFO              ] [Image       ] Providers: img_tex, img_dds, img_gif, img_sdl2 (img_pil, img_ffpyplayer ignored)
[INFO              ] [OSC         ] using <multiprocessing> for socket
[INFO              ] [Window      ] Provider: sdl2(['window_egl_rpi'] ignored)
[INFO              ] [GL          ] OpenGL version <b'4.5.0 NVIDIA 361.42'>
[INFO              ] [GL          ] OpenGL vendor <b'NVIDIA Corporation'>
[INFO              ] [GL          ] OpenGL renderer <b'GeForce GTX 560M/PCIe/SSE2'>
[...]

PR #4725, opened 2016-11-12
Catch permission errors

Zen-CODE:
These error are over dramatic and cause a lot of alarms amongst new
users. This pull request logs a warning instead of throwing an error.

references #4413, #4682

See also:

http://stackoverflow.com/questions/34477313/my-permission-is-denied-
using-kivy

http://stackoverflow.com/questions/26164723/kivy-and-buildozer-
permission-denied

### DIFF
--- a/kivy/input/providers/mtdev.py
+++ b/kivy/input/providers/mtdev.py
@@ -222,7 +222,16 @@ else:
             # open mtdev device
             _fn = input_fn
             _slot = 0
-            _device = Device(_fn)
+            try:
+                _device = Device(_fn)
+            except OSError as e:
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
@Zen-CODE Prettier than the raw raise and works for me in VM. I'd merge it. Will you include the docs changes here, or it'll be a separate PR?

Zen-CODE:
Thanks. Please merge. I'll commit the docs directly to master (we have a more relaxed policy on 'doc' only changes as they can't really break stuff) but reference this commit so it's all track-able.

And so you can suggest changes if you're not 100% happy :-) 

Peace out

PR #4726, opened 2016-11-13
Add multiline comment support

### LINKED ISSUE #4650: [Feature request] add multiline comment in KV lang
I think it would be nice to have the possibility to use multiline comments (with """ as the delimiter) in KV lang too: it's useful for debugging and for adding long comments.
Putting a # on every line it's very frustrating..

Thanx.

KeyWeeUsr:
Implement `'''` or `"""` alone to symbolize a multiline comment won't do as there will surely appear a person who will use multiline strings in kv like in python (`''' text on multiple lines'''`), so I added a `#` symbol to be even more obvious. Using `''# #''` might be a bad idea too.

Now it basically has a higher priority than a casual comment and to fetch this multiline comment the whole line (`stripped` variable) is checked for an occurrence of the comment start and then it starts removing _everything_ unless the comment ending (`#'''`or `#"""`) is present. Python breaks too with such a mistake, so I don't see a problem with it in kv.

Note that multiline comments **_won't work_** in `Builder.load_string(''' <kv> ''')` unless you choose the other variant of the quotes e.g. `Builder.load_string(''' '''# comment #''' ''')` will obviously break.

Test string:
```
'''#
Multiline
comment
#'''
    '''#
    Multiline
    comment
    #'''
'''#Multiline
comment#'''
#:import dp kivy.metrics.dp
'''#
Multiline
comment
#'''
    '''#
    Multiline
    comment
    #'''
'''#Multiline
comment#'''
<Test>:
    '''#
    Multiline
    comment
    #'''
        '''#
        Multiline
        comment
        #'''
    '''#Multiline
    comment#'''
    canvas:
        '''#
        Multiline
        comment
        #'''
            '''#
            Multiline
            comment
            #'''
        '''#Multiline
        comment#'''
        Color:
            '''#
            Multiline
            comment
            #'''
                '''#
                Multiline
                comment
                #'''
            '''#Multiline
            comment#'''
            rgb: 1, 0, 0
            '''#
            Multiline
            comment
            #'''
                '''#
                Multiline
                comment
                #'''
            '''#Multiline
            comment#'''
        Rectangle:
            '''#
            Multiline
            comment
            #'''
                '''#
                Multiline
                comment
                #'''
            '''#Multiline
            comment#'''
            size: self.size
            '''#
            Multiline
            comment
            #'''
                '''#
                Multiline
                comment
                #'''
            '''#Multiline
            comment#'''
            pos: self.pos
            '''#
            Multiline
            comment
            #'''
                '''#
                Multiline
                comment
                #'''
            '''#Multiline
            comment#'''
    Button:
        '''#
        Multiline
        comment
        #'''
            '''#
            Multiline
            comment
            #'''
        '''#Multiline
        comment#'''
        '''# Yes, this will work too #'''
```

Closes #4650

### DIFF
--- a/kivy/lang/parser.py
+++ b/kivy/lang/parser.py
@@ -508,12 +508,18 @@ class Parser(object):
            Comments need to be on a single line and not at the end of a line.
            i.e. a comment line's first non-whitespace character must be a #.
         '''
+        mult_com = False
         # extract directives
         for ln, line in lines[:]:
             stripped = line.strip()
             if stripped[:2] == '#:':
                 self.directives.append((ln, stripped[2:]))
-            if stripped[:1] == '#':
+            if mult_com or "'''#" in stripped or '"""#' in stripped:
+                mult_com = True
+                lines.remove((ln, line))
+                if "#'''" in stripped or '#"""' in stripped:
+                    mult_com = False
+            elif stripped[:1] == '#':
                 lines.remove((ln, line))
             if not stripped:
                 lines.remove((ln, line))
--- a/kivy/lang/parser.py
+++ b/kivy/lang/parser.py
@@ -514,11 +514,22 @@ class Parser(object):
             stripped = line.strip()
             if stripped[:2] == '#:':
                 self.directives.append((ln, stripped[2:]))
-            if mult_com or "'''#" in stripped or '"""#' in stripped:
-                mult_com = True
+            if mult_com or '###' in stripped:
                 lines.remove((ln, line))
-                if "#'''" in stripped or '#"""' in stripped:
+                begin = stripped.find('###') + 1
+                end = stripped.find('###', begin + 2) + 1
+
+                if begin and mult_com:
+                    # end of multiline com found
                     mult_com = False
+                    continue
+
+                if end:
+                    # end of multiline com in single line
+                    mult_com = False
+                    continue
+                mult_com = True
+
             elif stripped[:1] == '#':
                 lines.remove((ln, line))
             if not stripped:


inclement:
I don't like this syntax - it seems awkward, being ugly due to joining both existing 'comment' syntaxes, and also because it's more artificial than in Python, since ''' normally works precisely because it declares a string and therefore does the 'right' thing in different places, but that doesn't map to a kv structure in the same way. It's also unpleasant that it doesn't work well with Builder.load_string.

That said, I don't really like using `'''` for multiline comments anyway, but it's clearly popular. If others like this change, I wouldn't strongly oppose it.

KeyWeeUsr:
Same here, but I haven't come up with anything that would be python-related enough. If there's a prettier symbol, I'm all for the different thing. If you want C comments, let there be `/* */` multilines of all I care 😋 the basic structure is there, but it can't be `'''` or `"""`, because if someone uses such string in property, kv goes boom.

On the other hand, if there was a lot more complicated way of identifying the comment symbol (watching for an exact place somewhere in a line for each different indentation, etc), the python multiline comment could probably work. But the implementation would be much uglier than having `#` in multiline. 🙍

udiboy1209:
Apparently `'''` or `"""` are discouraged by PEP8 to be used as multiline comments. Refs: [Block Comments in PEP8](https://googleweblight.com/?lite_url=http://www.python.org/dev/peps/pep-0008/%23block-comments&lc=en-IN&s=1&m=96&host=www.google.co.in&ts=1479110176&re=1&sig=AF9NednxbIdrJhmWDb923Z3hail5FKXuFg#a-block-comments) and [Codeacademy forum discussion](https://www.codecademy.com/en/forum_questions/505ba3cfc6addb000200e33c).

We could do one of two things with KV then.

 * Interpret strings normally and check if they are assigned to a variable. If not throw them away. (Indirectly works like a multiline comment)

 * Discourage multiline comments entirely. Most text editors have commenting features which let you comment entire blocks of text using the `#`. 

Anyway, three hashes `###` might be intuitive and prettier than the current alternative.

The only place multiline comments should be used is for docstrings.

KeyWeeUsr:
The `###` seems like a better idea than to check for each string. It's easier to debug if a user breaks kv code with this (finding the opening and ending) in a code with multiple multiline strings, because now it checks in the beginning and throws away lines until told not to do so.

Python makes them real strings which are gc-ed, but there's no need to slow down kv because they implemented it like that. Unless you want to implement docstrings in kv too, then I'd understand. You can break more things with such checking, though.

Discouraging multiline comments on the other hand seems reasonable too as probably every editor except the basic notepad-like, yet they could be really useful if coding on mobile devices (where no such editor either exists or the editor isn't easy to operate e.g. emacs).

FluxIX:
''' and """ define strings and the syntax makes that apparent. If you want a multiline comment, it should have a concept of opening and closing (C does this with /* and */, but in no way am I advocating that specific syntax).

I do feel we need to support simple text editors as they will be more numerous in the mobile domain in the near-term.

dessant:
Please don't hunt me down. ❤️  https://github.com/kivy/kivy/issues/4650#issuecomment-266204122

PR #4737, opened 2016-11-18
Remove sdl2 presplash after initialised

KeyWeeUsr:
Remove android presplash if either `App` or `runTouchApp` is used after the Kivy application is [correctly initialised](https://github.com/kivy/kivy/blob/master/kivy/app.py#L804) with `build` method.

ref kivy/python-for-android#928

### DIFF
--- a/kivy/base.py
+++ b/kivy/base.py
@@ -203,6 +203,15 @@ class EventLoopBase(EventDispatcher):
         if mod in self.postproc_modules:
             self.postproc_modules.remove(mod)
 
+    def remove_android_splash(self, *args):
+        '''Remove android presplash in SDL2 bootstrap.'''
+        try:
+            from android import remove_presplash
+            remove_presplash()
+        except:
+            Logger.error('Base: Could not remove android presplash')
+            return
+
     def post_dispatch_input(self, etype, me):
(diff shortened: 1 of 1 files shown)


akshayaurora:
This is android specific.  Better space for this would be the android module in p4a.

kivy/base.py
```diff
@@ -203,6 +203,17 @@ def remove_postproc_module(self, mod):
...
 
+    def remove_android_splash(self, *args):
+        '''Remove android presplash in SDL2 bootstrap.'''
+        try:
+            from jnius import autoclass
+        except ImportError:
+            Logger.error('Base: Could not remove android presplash')
+            return
+
+        activity = autoclass('org.kivy.android.PythonActivity').mActivity
```
inclement: This part should also be wrapped in try/except I think - basically, this call should never be allowed to crash.

KeyWeeUsr: just default `except` or some specifical error?

KeyWeeUsr:
ref https://github.com/kivy/python-for-android/pull/930

KeyWeeUsr:
Apparently the kivy part wasn't merged yet, though p4a was.

PR #4745, opened 2016-11-20
Add cursor_width to TextInput

KeyWeeUsr:
A new fancy property! It should be a nice feature if someone would like to use the ancient-like cursor from DOS/emacs/... However I'm not sure if the cursor should be centered at the pos, or left-aligned (as it is now) - meaning:

    word<cursor>   (default)
    wo<cursor>rd   (better?)

or even cursor _above_ the letter e.g. like if inserting/overwriting in console. That would however need a little bit of tweaking.

Test with:

    from kivy.lang import Builder
    Builder.load_string('''
    <Test>:
        TextInput:
            cursor_width: 20
    ''')
    class Test(BoxLayout): pass
    runTouchApp(Test())

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -2615,6 +2615,15 @@ class TextInput(FocusBehavior, Widget):
     defaults to [1, 0, 0, 1].
     '''
 
+    cursor_width = NumericProperty('1sp')
+    '''Current width of the cursor.
+
+    .. versionadded:: 1.9.2
+
+    :attr:`cursor_width` is a :class:`~kivy.properties.NumericProperty` and
+    defaults to '1sp'.
+    '''
+
     line_height = NumericProperty(1)
     '''Height of a line. This property is automatically computed from the
     :attr:`font_name`, :attr:`font_size`. Changing the line_height will have
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -187,7 +187,7 @@
             rgba: (self.cursor_color if self.focus and not self.cursor_blink else (0, 0, 0, 0))
         Rectangle:
             pos: [int(x) for x in self.cursor_pos]
-            size: 1, -self.line_height
+            size: root.cursor_width, -self.line_height
         Color:
             rgba: self.disabled_foreground_color if self.disabled else (self.hint_text_color if not self.text else self.foreground_color)
 


kivy/uix/textinput.py
```diff
@@ -2615,6 +2615,15 @@ def _get_cursor_row(self):
     defaults to [1, 0, 0, 1].
     '''
 
+    cursor_width = NumericProperty(1)
+    '''Current width of the cursor.
+
+    .. versionadded:: 1.9.2
+
+    :attr:`cursor_color` is a :class:`~kivy.properties.NumericProperty` and
```
Leva7: You've got `cursor_color` here, correct it

dessant: @Leva7, yes sir.

KeyWeeUsr: Thanks! 👻

akshayaurora:
There is no reason why the default TextInput should be including options like these, there could be cursor shape, cursor color... these would be welcome as extensions though just like code input is. or in garden.

dessant:
`cursor_color` is already implemented. I would consider `cursor_width` a needed feature for any input with bigger font sizes, in those cases you'd want to make the cursor a bit thicker (2-3px) to better match the text.

Though I may also draw the line before cursor shaping, it sounds a bit exotic for it to be included in a general purpose text input.

KeyWeeUsr:
@akshayaurora There actually is [`cursor_color`](https://kivy.org/docs/api-kivy.uix.textinput.html#kivy.uix.textinput.TextInput.cursor_color) right above my change in `style.kv`, that's why I thought width would be good addition too as it manipulates the same thing and isn't really something _new_ or advanced, only gives an access to that `1`, which if editing, might result in user crippling the `style.kv` file.

kivy/uix/textinput.py
```diff
@@ -2615,6 +2615,15 @@ def _get_cursor_row(self):
     defaults to [1, 0, 0, 1].
     '''
 
+    cursor_width = NumericProperty(1)
```
dessant: We could also have '1sp' for the default value.

KeyWeeUsr: Nice catch, thanks.

tshirtman:
looks good to me, if you want to think about alignment, either you can give an offset (offset_x?) property, or halign ('left', 'center', 'right'). But that can be added later if needed anyway.

dessant:
untitled

```python
from kivy.app import App
from kivy.lang import Builder

kv = """
TextInput:
    text: 'TEST'
    font_size: dp(150)
    cursor_width: dp(3)
"""


class TestApp(App):
    def build(self):

if __name__ == '__main__':
    TestApp().run()
```

PR #4749, opened 2016-11-22
Update texture.pyx

doyousketch2:
texture_create_from_data  defaults to  mipmap=False
so if we try, and fail to find  im.have_mipmap
then we pass, and continue with default value.

solves this issue:

flip = Texture .create_from_data(self .image, mipmap = False) .flip_horizontal()
   File "kivy/graphics/texture.pyx", line 585, in kivy.graphics.texture.texture_create_from_data (kivy/graphics/texture.c:7399)
 AttributeError: 'Sprite' object has no attribute 'have_mipmap'

### DIFF
--- a/kivy/graphics/texture.pyx
+++ b/kivy/graphics/texture.pyx
@@ -584,8 +584,11 @@ def texture_create_from_data(im, mipmap=False):
         allocate = 0
 
     # if imagedata have more than one image, activate mipmap
-    if im.have_mipmap:
-        mipmap = True
+    try:
+        if im.have_mipmap:
+            mipmap = True
+    except Exception:
+        pass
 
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
I don't really work with core (at least not with this part), but `try-except-pass` seems bad enough to me. At least ping the user with [Logger](https://kivy.org/docs/api-kivy.logger.html) that something happened.

tshirtman:
is `self.image` here an `ImageData` instance, what is the `Sprite` class here ? you don't give a lot of info about what you are trying to do.

Even if the usage is correct, you'd want to use `hasattr('has_mipmap', im)` instead of try/except/pass.

Issue #4746, opened 2016-11-20
listening keys and fullscreen on second monitor


picibucor:
I have two monitors. If I run my project it appears on the main monitor. I can resize the window and make it full-screen by:
--> clicking on the full-screen button in the upper-right corner
--> pulling it to the top of the screen

As usual for Microsoft-Windows.

But this changes if I pull my App in the second window. By clicking on the full-screen icon, or pulling it to the top i will get the following error-message:

    [INFO   ] [Logger      ] Record log in C:\Mirk\CETACal\logs\kivy_16-11-18_53.txt
    [INFO   ] [Kivy        ] v1.9.2-dev0
    [INFO   ] [Python      ] v3.4.4 (v3.4.4:737efcadf5a6, Dec 20 2015, 20:20:57) [MSC v.1600 64 bit (AMD64)]
    [INFO   ] [Factory     ] 193 symbols loaded
    [INFO   ] [Image       ] Providers: img_tex, img_dds, img_gif, img_sdl2, img_pil (img_ffpyplayer ignored)
    [INFO   ] [Text        ] Provider: sdl2
    [INFO   ] [OSC         ] using <thread> for socket
    [INFO   ] [Window      ] Provider: sdl2
    [INFO   ] [GL          ] GLEW initialization succeeded
    [INFO   ] [GL          ] OpenGL version <b'4.3.0 - Build 10.18.14.4432'>
    [INFO   ] [GL          ] OpenGL vendor <b'Intel'>
    [INFO   ] [GL          ] OpenGL renderer <b'Intel(R) HD Graphics 4600'>
    [INFO   ] [GL          ] OpenGL parsed version: 4, 3
    [INFO   ] [GL          ] Shading version <b'4.30 - Build 10.18.14.4432'>
    [INFO   ] [GL          ] Texture max size <16384>
    [INFO   ] [GL          ] Texture max units <32>
    [INFO   ] [Window      ] auto add sdl2 input provider
    [INFO   ] [Window      ] virtual keyboard not allowed, single mode, not docked
    [INFO   ] [Base        ] Start application main loop
    [INFO   ] [GL          ] NPOT texture support is available
    [INFO   ] [Base        ] Leaving application in progress...
     Traceback (most recent call last):
       File "C:\Mirk\CETACal\cls\test4.py", line 67, in <module>
         ChildApp().run()
       File "C:\Python34\lib\site-packages\kivy\app.py", line 828, in run
         runTouchApp()
[...]

PR #4751, opened 2016-11-24
fix for #4746

picibucor:
if the target has no "to_window", just set targettop to 0


KeyWeeUsr:
Single monitor works normally, right? What about e.g. android or rpi? I'm asking, because you check for `to_window` in object, yet it might be better to check for a specific value of it if such keyword appeared out of nowhere. (e.g. `to_window=None`)

Issue #4727, opened 2016-11-13
Saving an Image using 'save()' throws error


udiboy1209:
Trying to save using this simple example from docs

On Ubuntu 16.04, python v2.7.11, kivy master
```
import kivy
from kivy.core.image import Image

img = Image('img.png')
img.save('img2.png')
```

Gives this error:

```
[INFO   ] [Logger      ] Record log in /home/udiboy/.kivy/logs/kivy_16-11-13_17.txt
[INFO   ] [Kivy        ] v1.9.2-dev0
[INFO   ] [Python      ] v2.7.11+ (default, Apr 17 2016, 14:00:29) 
[GCC 5.3.1 20160413]
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_gif, img_sdl2, img_pil (img_ffpyplayer ignored)
 Traceback (most recent call last):
   File "test_img.py", line 5, in <module>
     img.save('img2.png')
   File "/home/udiboy/.apps/venvs/kivy/local/lib/python2.7/site-packages/kivy/core/image/__init__.py", line 848, in save
     self.populate()
 AttributeError: 'Image' object has no attribute 'populate'
```

Error is at this line: https://github.com/kivy/kivy/blob/master/kivy/core/image/__init__.py#L848

The `populate()` method is in `ImageLoaderBase` class and not in `Image` class.

saqib1707:
@udiboy1209  In  https://github.com/kivy/kivy/blob/master/kivy/core/image/__init__.py#L843 
Removing the 'not' in this line solves the error since if data.fmt is not in ('rgba' , 'rgb') then it has to invoke self.populate() to convert it to rgba format.

udiboy1209:
@saqib1707 nice catch! Why don't you make a pull request with that change and add a test here https://github.com/kivy/kivy/blob/master/kivy/tests/test_image.py for testing the save function!

PR #4753, opened 2016-11-27
Issue#4727:Image saving using 'save()' throws error-- resolved

saqib1707:
If data.fmt is not in ('rgba' ,'rgb') then change it to rgba using self.populate()


udiboy1209:
Ahh, I see you made a PR! you should mention the issue number in the description so github automatically links the two: #4727 

See why the appveyor check failed and add the test I mentioned(on the issue) in the tests module, then this will be ready to merge!

KeyWeeUsr:
@udiboy1209 appveyor is dead for now, just ignore it. The problem is basically because of missing yml file (and config), which is in my PR. There was probably a redirection from matham's account to the KivyOrg account and the yml is only in a separate branch waiting to be tested. The branch also kind of depends on py35 wheels (well, not necessarily), so that makes the delay a little bit longer.

It's going to resolve soon enough, but for testing Travis is the key. If Travis is ok with PR, it shouldn't break the tested stuff.

KeyWeeUsr:
@saqib1707 Please remove `solution.py` file. It's not the test udiboy mentioned:

> add the test I mentioned(on the issue) in the tests module, then this will be ready to merge!

just a copied `kivy/core/image/__init__.py` with your fix, that's already present. Sorry for the confusion. 🙍

akshayaurora:
@saqib1707 thanks. Nice fix.

PR #4758, opened 2016-12-01
Add more usable error for new-line chars

KeyWeeUsr:
I think I've seen it in docs already, but this seems to be a better approach to just explicitly tell a user what to do rather than expecting that the user will dig throught the whole documentation or google for that. Although the error is quite readable for people who work with the python often, it's completly unusable for beginners who encounter kv-lang for the first time and at the same time are python beginners.

I made pep8 fixes too. Please help me test L448, because I'm not sure how to trigger it. ^^

### DIFF
--- a/kivy/lang/parser.py
+++ b/kivy/lang/parser.py
@@ -49,6 +49,9 @@ lang_cls_split_pat = re.compile(', *')
 # widget is deleted
 _handlers = defaultdict(partial(defaultdict, list))
 
+# global msg for premature EOL
+eol_msg = 'Consider escaping new-line characters (\\n)'
+
 
 class ProxyApp(object):
     # proxy app object
@@ -175,13 +178,27 @@ class ParserRuleProperty(object):
         if mode == 'eval':
             # if we don't detect any string/key in it, we can eval and give the
             # result
+            if re.search(lang_key, tmp) is None and value in ["'", '"']:
+                try:
+                    self.co_value = eval(value)
+                except SyntaxError as e:
+                    raise ParserException(
+                        self.ctx, self.line,
+                        '\n'.join([str(e), eol_msg]))
+                return
             if re.search(lang_key, tmp) is None:
                 self.co_value = eval(value)
                 return
 
         # ok, we can compile.
         value = '\n' * self.line + value
-        self.co_value = compile(value, self.ctx.filename or '<string>', mode)
+        try:
+            self.co_value = compile(
+                value, self.ctx.filename or '<string>', mode)
+        except SyntaxError as e:
+            raise ParserException(
+                self.ctx, self.line,
+                '\n'.join([str(e), eol_msg]))
 
         # for exec mode, we don't need to watch any keys.
         if mode == 'exec':
@@ -295,9 +312,16 @@ class ParserRule(object):
             self._build_template()
         else:
             if self.ctx.root is not None:
-                raise ParserException(
-                    self.ctx, self.line,
-                    'Only one root object is allowed by .kv')
+                err_msg = 'Only one root object is allowed by .kv'
+                sc_previous = self.ctx.sourcecode[self.line - 1]
+                sc_current = self.ctx.sourcecode[self.line]
+
+                for quote in ["'", '"', "''", '""']:
+                    if sc_previous.count(quote) == sc_current.count(quote):
+                        raise ParserException(
+                            self.ctx, self.line,
+                            '\n'.join([err_msg, eol_msg]))
+                raise ParserException(self.ctx, self.line, err_msg)
             self.ctx.root = self
 
     def _build_rule(self):
@@ -353,7 +377,7 @@ class ParserRule(object):
             raise ParserException(self.ctx, self.line,
                                   'Invalid template (must be inside [])')
         item_content = name[1:-1]
-        if not '@' in item_content:
+        if '@' not in item_content:
             raise ParserException(self.ctx, self.line,
                                   'Invalid template name (missing @)')
         template_name, template_root_cls = item_content.split('@')
@@ -428,17 +452,17 @@ class Parser(object):
                     break
                 if ref in __KV_INCLUDES__:
                     if not os.path.isfile(resource_find(ref) or ref):
-                        raise ParserException(self, ln,
-                                              'Invalid or unknown file: {0}'
-                                              .format(ref))
+                        raise ParserException(
+                            self, ln,
+                            'Invalid or unknown file: {0}'.format(ref))
                     if not force_load:
                         Logger.warn('Lang: {0} has already been included!'
                                     .format(ref))
                         continue
                     else:
-                        Logger.debug('Lang: Reloading {0} ' +
-                                     'because include was forced.'
-                                     .format(ref))
+                        Logger.debug(
+                            'Lang: Reloading {0} because include was forced.'
+                            .format(ref))
                         kivy.lang.builder.Builder.unload_file(ref)
                         kivy.lang.builder.Builder.load_file(ref)
                         continue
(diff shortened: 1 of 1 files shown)


udiboy1209:
Do you have a test case for this PR? I still get a different error when I use `\n` in a kv string I load using Builder and a kv file works correctly.

KeyWeeUsr:
@udiboy1209 What error do you get?

**single file:**
```
from kivy.lang import Builder
Builder.load_string('''
<Test>:
    canvas:
        Color:
            rgb: 1, 0, 0
        Rectangle:
            size: self.size
            pos: self.pos
    Button:
        text: 'test\ntest'  # triggers the error
''')
class Test(BoxLayout): pass
runTouchApp(Test())
```

**multiple files (main.py, my.kv):**
```
from kivy.app import App
from kivy.lang import Builder
Builder.load_file('answertesting.kv')
class Test(BoxLayout): pass
class My(App):
    def build(self):
        return Test()
My().run()
```
```
<Test>:
    canvas:
        Color:
            rgb: 1, 0, 0
        Rectangle:
            size: self.size
            pos: self.pos
    Button:
        text: 'test\ntest'  # wont trigger, probably even shouldnt
```

udiboy1209:
@KeyWeeUsr I am running an example similar to the first one. This is the error I get:

```
[INFO   ] [Logger      ] Record log in /home/udiboy/.kivy/logs/kivy_16-12-07_28.txt
[INFO   ] [Kivy        ] v1.9.2-dev0
[INFO   ] [Python      ] v3.5.1+ (default, Mar 30 2016, 22:46:26) 
[GCC 5.3.1 20160330]
[INFO   ] [Factory     ] 193 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_gif, img_sdl2 (img_pil, img_ffpyplayer ignored)
   File "kvtest.py", line 14, in <module>
     TestApp().run()
   File "/home/udiboy/.apps/venvs/kivy3/lib/python3.5/site-packages/kivy/app.py", line 802, in run
     root = self.build()
   File "kvtest.py", line 11, in build
   File "/home/udiboy/.apps/venvs/kivy3/lib/python3.5/site-packages/kivy/lang/builder.py", line 348, in load_string
     parser = Parser(content=string, filename=fn)
   File "/home/udiboy/.apps/venvs/kivy3/lib/python3.5/site-packages/kivy/lang/parser.py", line 401, in __init__
     self.parse(content)
   File "/home/udiboy/.apps/venvs/kivy3/lib/python3.5/site-packages/kivy/lang/parser.py", line 505, in parse
     objects, remaining_lines = self.parse_level(0, lines)
   File "/home/udiboy/.apps/venvs/kivy3/lib/python3.5/site-packages/kivy/lang/parser.py", line 563, in parse_level
     '%s spaces' % spaces)
 kivy.lang.parser.ParserException: Parser: File "<inline>", line 4:
 ...
       2:Label:
       3:    text: 'Some 
 >>    4: very long line which will be cut'
 ...
 Invalid indentation, must be a multiple of 4 spaces
```

and this is the code:

```python
from kivy.app import App
from kivy.lang import Builder

kv = '''
Label:
    text: 'Some \n very long line which will be cut'
'''

class TestApp(App):
    def build(self):

if __name__ == '__main__':
    TestApp().run()
```

KeyWeeUsr:
Right, that will be here no matter what. The indentation is because of `"\n "`, which just puts a single space at the beginning, which messes up the whole thing. 

I may try to catch those things too if there is `\n`

udiboy1209:
From what I think, there are only two possible errors which can arise from not escaping newlines

 * Unexpected EOL in string
 * Indentation error

We can output a warning with the EOL error to check for unescaped newlines and missed ending  quotes `'`.

And the Indentation error outputs the actual KV text being processed which makes it obvious that the newline needs to be escaped, so we can leave that as it is. 

I think this is better than predicting whether there was an unescaped newline. What do you think?

KeyWeeUsr:
Seems reasonable and shouldn't really slow things down. The edit should be a little bit better catching even the EOL that it didn't before. Hopefully now remains only the indentation error, which I also think is rather more obvious that user broke something and if the user tries to fix the indentation error, sooner or later the EOL will be triggered, therefore even the warning would be visible.

udiboy1209:
The indentation error shouldn't be given higher priority over unexpected EOL. Because in some cases I also get the error `Only one root object is allowed by .kv` . Does the parser run through the entire KV once before evaluating it? I think the final step would be to ensure EOL is thrown when it is clearly the cause of problems.

KeyWeeUsr:
It should now fetch both indentation and root error. I guess there's nothing to slow down at all if an exception is already going to be thrown, so let's at least make something good with it 😋

matham:
This reminds me of https://mail.python.org/pipermail/python-ideas/2016-November/043848.html. While error suggestions are good when we know for sure what causes the errors, I think guessing is probably not a good idea overall since it may also mislead. But it's been a while since I looked at lang so I'm not sure if other issues could cause the errors at location they were changed.

KeyWeeUsr:
Yeah, I saw something similar before writing this, but then again a beginner user of kv lang might not guess that a newline causes the error and that might just backfire with opening issues. Seems like none was opened recently, but I think I saw such question at least on IRC.

I think it's only `\n` character that causes such issues (that's why the condition for that), but if not, then it raises just a casual error without an additional message.

KeyWeeUsr:
This change might be useful for beginners, because the output from KV parser might be not very readable in the beginning especially if Python's traceback throws stuff into console too. However, if you feel the change is not really needed, then feel free to close it.

matham:
I'll close for now because I do think this could raise exceptions where the error mentioned isn't the true cause. If there were extensive tests I would probably feel more comfortable with this though.

Soichi-Tamashiro:
> 
> 
> @KeyWeeUsr I am running an example similar to the first one. This is the error I get:
> 
> ```
> [INFO   ] [Logger      ] Record log in /home/udiboy/.kivy/logs/kivy_16-12-07_28.txt
> [INFO   ] [Kivy        ] v1.9.2-dev0
> [INFO   ] [Python      ] v3.5.1+ (default, Mar 30 2016, 22:46:26) 
> [GCC 5.3.1 20160330]
> [INFO   ] [Factory     ] 193 symbols loaded
> [INFO   ] [Image       ] Providers: img_tex, img_dds, img_gif, img_sdl2 (img_pil, img_ffpyplayer ignored)
>  Traceback (most recent call last):
>    File "kvtest.py", line 14, in <module>
>      TestApp().run()
>    File "/home/udiboy/.apps/venvs/kivy3/lib/python3.5/site-packages/kivy/app.py", line 802, in run
>      root = self.build()
>    File "kvtest.py", line 11, in build
>      return Builder.load_string(kv)
>    File "/home/udiboy/.apps/venvs/kivy3/lib/python3.5/site-packages/kivy/lang/builder.py", line 348, in load_string
>      parser = Parser(content=string, filename=fn)
>    File "/home/udiboy/.apps/venvs/kivy3/lib/python3.5/site-packages/kivy/lang/parser.py", line 401, in __init__
>      self.parse(content)
>    File "/home/udiboy/.apps/venvs/kivy3/lib/python3.5/site-packages/kivy/lang/parser.py", line 505, in parse
>      objects, remaining_lines = self.parse_level(0, lines)
>    File "/home/udiboy/.apps/venvs/kivy3/lib/python3.5/site-packages/kivy/lang/parser.py", line 563, in parse_level
>      '%s spaces' % spaces)
>  kivy.lang.parser.ParserException: Parser: File "<inline>", line 4:
>  ...
>        2:Label:
>        3:    text: 'Some 
>  >>    4: very long line which will be cut'
>  ...
>  Invalid indentation, must be a multiple of 4 spaces
> ```
> 
> and this is the code:
> 
> ```python
> from kivy.app import App
> 
> kv = '''
> Label:
>     text: 'Some \n very long line which will be cut'
> '''
> 
> class TestApp(App):
>     def build(self):
>         return Builder.load_string(kv)
> 
> if __name__ == '__main__':
>     TestApp().run()
> ```
A fix to this error is by replacing every ```\n``` with ```\\n```  and it will compile without errors.
A fix to your code is below:
```
from kivy.app import App
from kivy.lang import Builder

kv = '''
Label:
    text: 'Some \n\n very long line which will be cut'
'''

class TestApp(App):
    def build(self):

if __name__ == '__main__':
    TestApp().run()
```

Issue #4672, opened 2016-10-15
Select all, Copy and Paste is not reachable on Android if TextInput is first / upper element


oxidworks:
I have this problem with Android 5.0.2

Program code:

```
from kivy.app import App
from kivy.uix.button import Button
from kivy.uix.textinput import TextInput
from kivy.uix.label import Label
from kivy.uix.boxlayout import BoxLayout

class DEESCA(App):

    def on_pause(self):
        return True

    def on_translate_button_clicked(self, instance):
        pass

    def build(self):
        self.icon = "data/icon.png"        
        self.inputtext = TextInput(hint_text="insert here")
        self.inputtext.bind(on_text_validate=self.on_translate_button_clicked)
        self.l1 = TextInput(text="DE", readonly=True)
        self.l2 = TextInput(text="ES", readonly=True)
        self.l3 = TextInput(text="CA", readonly=True)
        b = Button(text='Translate')
        b.bind(on_press=self.on_translate_button_clicked)
        layout = BoxLayout(orientation='vertical', padding=10)
        layout.add_widget(self.inputtext)
        layout.add_widget(b)
        layout.add_widget(self.l1)
        layout.add_widget(self.l2)
        layout.add_widget(self.l3)
        return layout 

if __name__ == '__main__':
    DEESCA().run()
```

KeyWeeUsr:
Not sure if an issue or not. If I create `TextInput`s with `use_bubble=True` (as it'd work on mobile), the bubble is clearly visible when clicked in the right area i.e.:

* directly under the top = invisible (bubble is too far above widget)
* a little bit lower = visible

Should the `Bubble` in `TextInput` switch its orientation too e.g. when triggered by a touch in upper part of a widget? The default state is body up, arrow down - should the arrow be on the top and body at the bottom + pushed down to the `TextInput`'s area?

If not, then the issue is really only about [`inch(.25)`](https://github.com/kivy/kivy/blob/master/kivy/uix/textinput.py#L1593) being too big and should be changed to let's say `.15`.

saqib1707:
@KeyWeeUsr one idea would be to place the bubble right or left of the touch when the touch is made directly under the top(when it becomes invisible).If this has to be done then I want to work on this issue.

KeyWeeUsr:
Well, this is the current behavior. The right thing would most likely be to check for the collisions with outer bonds of TextInput and threw the Bubble the other way. I saw some code related to it, maybe it's just still not done or something.

[...]

PR #4761, opened 2016-12-03
Fix#4672 : Select_all,Copy, Paste options reachable on android when textinput is upper/top widget

saqib1707:
The pos argument of _show_cut_copy_paste function in textinput.py file has to be relative , so changed that to relative(to the widget) by subtracting self.y (position of textinput widget w.r.t the window). Also in function on_touch_up in textinput.py the call to _show_cut_copy_paste has been made using absolute position, so introduced a variable touch_rel_pos to pass the relative touch position to _show_cut_copy_paste.
To make the textinput bubble respond on touching , added a always_release=True in the TextInputCutCopyPaste Class.

### DIFF
--- a/kivy/uix/bubble.py
+++ b/kivy/uix/bubble.py
@@ -80,7 +80,7 @@ class BubbleButton(Button):
     Rather use this BubbleButton widget that is already defined and provides a
     suitable background for you.
     '''
-    pass
+    always_release = BooleanProperty(True)
 
 
 class BubbleContent(GridLayout):
@@ -372,3 +372,4 @@ class Bubble(GridLayout):
     def _update_arrow(self, *dt):
         if self.arrow_pos in ('left_mid', 'right_mid'):
             self._sctr.center_y = self._arrow_layout.center_y
+
--- a/kivy/uix/bubble.py
+++ b/kivy/uix/bubble.py
@@ -372,4 +372,3 @@ class Bubble(GridLayout):
     def _update_arrow(self, *dt):
         if self.arrow_pos in ('left_mid', 'right_mid'):
             self._sctr.center_y = self._arrow_layout.center_y
-
--- a/kivy/uix/bubble.py
+++ b/kivy/uix/bubble.py
@@ -80,7 +80,7 @@ class BubbleButton(Button):
     Rather use this BubbleButton widget that is already defined and provides a
     suitable background for you.
     '''
-    always_release = BooleanProperty(True)
+    pass
 
 
 class BubbleContent(GridLayout):
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -485,7 +485,7 @@ class TextInput(FocusBehavior, Widget):
         self._refresh_text_from_property_ev = None
         self._long_touch_ev = None
         self._do_blink_cursor_ev = Clock.create_trigger(
-            self._do_blink_cursor, .5, interval=True)
+            self._do_blink_cursor, .5)
         self._refresh_line_options_ev = None
         self.interesting_keys = {
             8: 'backspace',
@@ -1596,7 +1596,7 @@ class TextInput(FocusBehavior, Widget):
         win_size = win.size
         bubble_pos = (t_pos[0], t_pos[1] + inch(.25))
 
-        if (bubble_pos[0] - bubble_hw) < 0:
+        if (bubble_pos[0] - bubble_hw + self.x) < 0:
             # bubble beyond left of window
             if (bubble_pos[1] + self.y) > (win_size[1] - bubble_size[1]):
                 # bubble above window height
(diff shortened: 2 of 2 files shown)


kivy/uix/bubble.py
```diff
@@ -80,6 +80,9 @@ class BubbleButton(Button):
     Rather use this BubbleButton widget that is already defined and provides a
     suitable background for you.
     '''
+    def __init__(self, **kwargs):
+        super(BubbleButton, self).__init__(**kwargs)
+        self.always_release = BooleanProperty(False)
```
akshayaurora: properties are to be used at class level.

saqib1707: @akshayaurora sorry for the delay . I realized that the always_release should be a class property and hence changed bubble.py

saqib1707:
#4672

kivy/uix/textinput.py
```diff
@@ -1402,7 +1402,8 @@ def on_touch_up(self, touch):
             # show Bubble
             win = EventLoop.window
             if self._selection_to != self._selection_from:
-                self._show_cut_copy_paste(touch.pos, win)
+                touch_rel_pos = (touch.pos[0] - self.x, touch.pos[1] - self.y)
```
akshayaurora: There should be no adjustment made outside the cut copy paste function as positioning is done in it.

Doing it in multiple places will lead to confusion.

saqib1707: But the show_cut_copy_paste function accepts relative positions because in one place local (wrt the widget) position was provided to the cut_copy_paste function while here it was absolute(wrt the window) .

kivy/uix/bubble.py
```diff
@@ -80,7 +80,7 @@ class BubbleButton(Button):
     Rather use this BubbleButton widget that is already defined and provides a
     suitable background for you.
     '''
-    pass
+    always_release = BooleanProperty(True)
```
dessant: Please see https://github.com/kivy/kivy/issues/4672#issuecomment-267908661.

kivy/uix/textinput.py
akshayaurora: The show_cut_copy_paste function is supposed to manage the display of the bubble,
 thus it should be taking care of the conversion if any, this has the added advantage that we won't have the conversion in many places.

kivy/uix/textinput.py
```diff
@@ -1593,27 +1594,27 @@ def _show_cut_copy_paste(self, pos, win, parent_changed=False,
         win_size = win.size
         bubble_pos = (t_pos[0], t_pos[1] + inch(.25))
 
-        if (bubble_pos[0] - bubble_hw) < 0:
+        if (bubble_pos[0] - bubble_hw + self.x) < 0:
```
saqib1707: @akshayaurora got your point about half-visiblity but when it is off-screen (not on the screen)  then  no touch is made on the textinput so anyway no bubble is going to appear.

kivy/uix/bubble.py
saqib1707: @dessant  if the always_release is not explicitly made True for BubbleButton class (as highlighted above) then the on_release event won't be triggered by the button here https://github.com/kivy/kivy/blob/master/kivy/data/style.kv#L202-L217 making the bubbles non-functioning. Other way would be to use on_press instead of on_release in style.kv...

dessant: `on_press` would just sidestep the bug, the same way `always_release = True` would. We're looking for a solution which lets users set both `True` and `False` for `always_release`.

kivy/uix/textinput.py
```diff
@@ -454,7 +454,7 @@ class TextInput(FocusBehavior, Widget):
                   'on_quad_touch')
 
     def __init__(self, **kwargs):
-        self.always_release_bubble = kwargs.get('always_release_bubble', False)
+        self.always_release_bubble = kwargs.get('always_release_bubble', True)
```
dessant: I think this is just hiding the behavior change, but we are still able to see it. 😁  Do not modify `always_release` in any form.

saqib1707: But u said that there should be a way so that users can set always_release for the BubbleButton object. From here https://github.com/kivy/kivy/blob/master/kivy/uix/behaviors/button.py#L168-L169 if self.dispatch('on_release') event has to be triggered then either always_release should be True or self.collide_point(*touch.pos) should  be True. 
I created a new variable which can be set by the users .Other way would be to change the touch.pos in local or windows co-ordinates.(not sure about the second option)

dessant: I think the keyword argument is not strictly needed to fix the bug you were trying to fix. That's a different feature and customization option. What i was saying is, you should not care about the value of always_release, and most importantly don't change the default value yourself. I'm going to close this pr because it's a collective waste of time for everyone involved, please open a new pr if you have a fix that doesn't merely hide the bug.

Issue #4736, opened 2016-11-18
setting suggestion_text crashes if textinput.text is empty string


janssen:
Example:

```
import sys

from kivy.app import App
from kivy.uix.textinput import TextInput

class MyApp(App):

    def on_start(self):
        self.root.text = ""
        self.root.bind(text=lambda w, v: sys.stdout.write(v + "\n"))
        self.root.bind(suggestion_text=lambda w, v: sys.stdout.write("suggestion_text changed to: " + v + '\n'))
        self.root.suggestion_text = "foo"

    def build(self):
        return TextInput(readonly=False, multiline=False, focus=True, on_parent=lambda w, x: w.setter("focus")(True))

if __name__ == "__main__":
    MyApp().run()
```

and the result:

```
 Traceback (most recent call last):
   File "/tmp/test5.py", line 18, in <module>
     MyApp().run()
   File "/usr/lib/python2.7/site-packages/kivy/app.py", line 827, in run
     self.dispatch('on_start')
   File "kivy/_event.pyx", line 718, in kivy._event.EventDispatcher.dispatch (kivy/_event.c:7699)
   File "/tmp/test5.py", line 12, in on_start
     self.root.suggestion_text = "foo"
   File "kivy/properties.pyx", line 408, in kivy.properties.Property.__set__ (kivy/properties.c:5114)
   File "kivy/properties.pyx", line 446, in kivy.properties.Property.set (kivy/properties.c:5876)
   File "kivy/properties.pyx", line 501, in kivy.properties.Property.dispatch (kivy/properties.c:6557)
   File "kivy/_event.pyx", line 1224, in kivy._event.EventObservers.dispatch (kivy/_event.c:13497)
   File "kivy/_event.pyx", line 1130, in kivy._event.EventObservers._dispatch (kivy/_event.c:12696)
   File "/usr/lib/python2.7/site-packages/kivy/uix/textinput.py", line 2772, in on_suggestion_text
     txt = self._lines[self.cursor_row]
 IndexError: list index out of range
```

janssen:
There's clearly an initialization bug and race condition here.  This crashes:

```
import sys

from kivy.app import App
from kivy.uix.textinput import TextInput
from kivy.clock import Clock
from kivy.uix.boxlayout import BoxLayout

class MyApp(App):

    def setit(self, dt):
        if self.ti._lines:
            self.ti.suggestion_text = "foo"

    def on_start(self):
        Clock.schedule_once(self.setit)

    def build(self):
[...]

PR #4762, opened 2016-12-03
Prevent error if canvas is not setup yet

Zen-CODE:
references issue #4736

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -707,7 +707,8 @@ class TextInput(FocusBehavior, Widget):
         wrap = (self._get_text_width(
             new_text,
             self.tab_width,
-            self._label_cached) > (self.width - self.padding[0] - self.padding[2]))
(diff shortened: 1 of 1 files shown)


udiboy1209:
@Zen-CODE your doesn't fix the error mentioned in the issue. I still get the same error for both python 2.7 and 3.5

Zen-CODE:
@udiboy1209. Please can you tell me how you make it error, or post an example? The fix mainly addresses setting on initialization, but it works for me whatever I try...

udiboy1209:
I am running the same example the OP posted in #4736 and getting the same error he gets. I checked the installed textinput.py and it does have your change. Also check my latest comment on that issue's thread.

Zen-CODE:
@usiboy1209. Thanks. Latest push should fix, but it's hackish. I think it's probably better to re-visit the implementation....

akshayaurora:
@Zen-CODE thanks

PR #4763, opened 2016-12-03
Added guide on how to package for android on windows10's Ubuntu Bash …

el3:
Created a new file /kivy/doc/sources/guide/packaging_android_windows.rst
And changed /kivy/doc/sources/guide/packaging_android.rst toinclude the above file

### DIFF
--- a/doc/sources/guide/packaging-android-windows.rst
+++ b/doc/sources/guide/packaging-android-windows.rst
@@ -121,7 +121,7 @@ Download the android-sdk and android-ndk commandline tools for linux::
     # wget https://dl.google.com/android/repository/android-ndk-r13b-linux-x86_64.zip
 
 
-Now unzpack the NDK and SDK::
+Now unpack the NDK and SDK::
 
     # unzip android-ndk-r13b-linux-x86_64.zip # Check version
     # tar -xvzf android-sdk_r24.4.1-linux.tgz # Check version
--- a/doc/sources/guide/packaging-android-windows.rst
+++ b/doc/sources/guide/packaging-android-windows.rst
@@ -127,7 +127,7 @@ Download the android-sdk and android-ndk commandline tools for linux::
     # wget https://dl.google.com/android/repository/android-ndk-r13b-linux-x86_64.zip
 
 
-Now unzpack the NDK and SDK:
+Now unzpack the NDK and SDK::
 
     # unzip android-ndk-r13b-linux-x86_64.zip # Check version
     # tar -xvzf android-sdk_r24.4.1-linux.tgz # Check version
--- a/doc/sources/guide/packaging-android-windows.rst
+++ b/doc/sources/guide/packaging-android-windows.rst
@@ -91,6 +91,7 @@ When in the root shell, run these commands::
                                   pkg-config libtool python-pip openjdk-8-jdk
     # update-alternatives --config java
     # update-alternatives --config javac
+    # pip install pyhton-for-android
 
 
 Now make a directory to install SDK and NDK in.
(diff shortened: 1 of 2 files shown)


doc/sources/guide/packaging-android-windows.rst
```diff
@@ -0,0 +1,212 @@
...
+                                  python2.7-dev libncurses5:i386 libstdc++6:i386 cython \
+                                  zlib1g:i386 unzip ant ccache python-virtualenv \
+                                  autoconf automake pkg-config libtool
+    # add-apt-repository ppa:webupd8team/java -y
+    # apt-get update
+
+
+Install java, and accept license::
+
+    # apt-get install -y oracle-java8-installer oracle-java8-set-default
```
KeyWeeUsr: openjdk doesn't work or something?

doc/sources/guide/packaging-android-windows.rst
```diff
@@ -0,0 +1,212 @@
...
+    Check the link. The versions might be different today
+
+
+Download the android-sdk and android-ndk commandline tools for linux::
+
+    # wget https://dl.google.com/android/android-sdk_r24.4.1-linux.tgz
+    # wget https://dl.google.com/android/repository/android-ndk-r13b-linux-x86_64.zip
+
+
+Now unzpack the NDK and SDK:
```
KeyWeeUsr: missing colon -> `::`

doc/sources/guide/packaging-android-windows.rst
el3: I had some troubles with java errors in the process. Thats why I switched.
But that did not fix the errors. It was because of windows virtual memory space was too small.
That said, openjdk might work. I did not test it tho.

KeyWeeUsr: Please test with openjdk, it's even in p4a docs, so the consistency would be nice. That's only my opinion though, it might be necessary. Great job! 🍡

el3: Yea I will do that.
Then I will update it, if it works.

phunsukwangdu:
I worked out this guide on windows10 bash shell and it works perfectly fine! :D  

--when i followed till  "apt-get upgrade -y" .i copied the dependency command from [here](https://python-for-android.readthedocs.io/en/latest/quickstart/ ) 
--Rest was smooth.

Suggestions:-
  --on windows 10 it is "This pc" than "My computer".So replacing "My Computer" with "This Pc" is             preferable.
--on setting out virtual memory size the Custom size shows two options.They are intial and maximal size.It will be doubtful for user where to put the value.
--activation of bash shell can be showed in the guide itself.

KeyWeeUsr:
That `My Computer` thing seems to differ on other versions of english or on lower windows version. I've seen `My Computer`, `Computer`, `This PC` or even `PC`. So much for consistency. It might be good to put there `*` and somewhere below put the different names if necessary.

The dependency commands are basically the same. There's however **_Cython_** missing, so that would be nice to correct too.

++ for the virtual memory size edit

and I'm not really sure about the bash shell, but I think it can't hurt if it's short.

phunsukwangdu:
Activation of bash shell required only three steps.
-enable developer settings
-tick windows subsystem for linux in turn windows feature on or off.
-restart and start bash.
But yes using of bash shell can be difficult  for windows user.
plz do guide me and spot my mistake as i tend to make a lot.thx

dessant:
@el3, great work! Would you mind reopening this for https://github.com/kivy/python-for-android?

dessant:
As discussed on irc, this will be adapted to the p4a docs.

Issue #4752, opened 2016-11-25
Line ends are not drawn correctly on canvas


hackalog:
My line ends seem to be less than 100% of a circle if I draw a line anywhere towards the left.

e.g.
<img width="237" alt="line_ends_bug" src="https://cloud.githubusercontent.com/assets/11911411/20634764/464bd8b4-b322-11e6-8a38-8a1d7a0bac19.png">

This is on macos (retina), 

Simple reproducer below:
```python
from kivy.app import App
from kivy.uix.boxlayout import BoxLayout
from kivy.lang.builder import Builder
from kivy.graphics import Color, Line

kv = '''

BugTest:

<BugTest>:
    canvas.before:
        Color:
            rgba: 1,1,1,1
        Rectangle:
            size: self.size
            pos: self.pos
    Label:
        text: 'test'
'''


class BugTest(BoxLayout):
    def __init__(self, **kwargs):
        super(BugTest, self).__init__(**kwargs)
        with self.canvas:
            Color(0, 0, 0, 1)
        self.line_width=10

    def on_touch_down(self, touch):
        with self.canvas:
            touch.ud['current_line'] = Line(
                points=(touch.x, touch.y), width=self.line_width, cap='round', joint='round')
            touch.ud['current_line'].points += (touch.x, touch.y)

    def on_touch_move(self, touch):
        if 'current_line' in touch.ud:
            touch.ud['current_line'].points += (touch.x, touch.y)


class LineEndBugApp(App):

    def build(self):
        self.app = Builder.load_string(kv)
        return self.app

LineEndBugApp().run()
```

saqib1707:
@hackalog  On trying to reproduce I got this traceback..Basically there is no Builder class defined inside ..kivy/lang/builder.py

``Traceback` (most recent call last):
   File "test.py", line 3, in <module>
     from kivy.lang.builder import Builder
 ImportError: No module named builder`

KeyWeeUsr:
@saqib1707 No problem with that code for me. Check if you have Kivy installed properly or try to make it `from kivy.lang import Builder` only.

saqib1707:
Thanks @KeyWeeUsr .Its working .

700brains:
Hi friend why does my app crashes whenever I initialize it(start it)
meanwhile am running it in root directory

700brains:
Everytime I tried starting my kivy app it crashes automatically. Am running
it in a root  directory

saqib1707:
[...]

PR #4772, opened 2016-12-07
Fix#4752:Line end drawn correctly on canvas when cap='round'

saqib1707:
For making the cap at both ends round , instead of taking the vertices  from ~(0-173) degrees at the two ends , I have sampled the vertices from complete (0-360) degrees and that's why the vertices_count and indices_count has been increased . Moreover range of for loop has been increased to cover the complete rounded circle at both ends.This works only for round cap.For square cap , the sangle(starting angle) =0 always hence the starting cap vertices point  [https://github.com/kivy/kivy/blob/master/kivy/graphics/vertex_instructions_line.pxi#L564-L569](url)  are coming the same whatever the line direction be.

### DIFF
--- a/kivy/graphics/vertex_instructions_line.pxi
+++ b/kivy/graphics/vertex_instructions_line.pxi
@@ -285,11 +285,15 @@ cdef class Line(VertexInstruction):
         self._bxmax = -999999999
         self._bymax = -999999999
 
+        cap = self._cap
+        if cap == LINE_CAP_SQUARE:
+            p = p[2:]
+            count -= 1
+
         if count < 2:
             self.batch.clear_data()
             return
 
-        cap = self._cap
         if self._close and count > 2:
             p = p + p[0:4]
             count += 2
--- a/kivy/graphics/vertex_instructions_line.pxi
+++ b/kivy/graphics/vertex_instructions_line.pxi
@@ -314,8 +314,8 @@ cdef class Line(VertexInstruction):
             indices_count += 12
             vertices_count += 4
         elif cap == LINE_CAP_ROUND:
-            indices_count += (self._cap_precision * 3) * 2
-            vertices_count += (self._cap_precision) * 2
+            indices_count += 2 * (self._cap_precision + 1) * 3 * 2
+            vertices_count += 2 * (self._cap_precision + 1) * 2
 
         vertices = <vertex_t *>malloc(vertices_count * sizeof(vertex_t))
         if vertices == NULL:
@@ -592,7 +592,7 @@ cdef class Line(VertexInstruction):
             vertices[iv].s0 = 0
             vertices[iv].t0 = 0
             iv += 1
-            for i in xrange(0, self._cap_precision - 1):
+            for i in xrange(0, 2 * self._cap_precision + 1):
                 vertices[iv].x = cx + cos(a1 + step * i) * w
                 vertices[iv].y = cy + sin(a1 + step * i) * w
                 vertices[iv].s0 = 1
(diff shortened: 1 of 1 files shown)


saqib1707:
#4752

saqib1707:
@udiboy1209 Well the requested changes have been done.

udiboy1209:
The example on #4752 is still giving an error with `cap='square'` as you mentioned. I think you should fix both cases before this PR is merged.

dessant:
@saqib1707, when that's solved please get @tito, @kovak or someone else who has experience with the Line instruction to take a look at the pr. We can't afford a breaking change in this part of the code.

saqib1707:
@tito Can u please review this pr related to Line instruction as suggested by dessant above.

sastels:
Hi all! Was pleasantly surprised that an issue that's been bugging me has a fix in! This LGTM - @saqib1707 @udiboy1209 @dessant @tito is there a reason it hasn't been merged / if so is there something I can do to help? Thanks!

tshirtman:
This indeed works fine, i wanted to check if it was not too wasteful, because multiplying indices by two seemed a bit much, but i don't see an obvious way to avoid that now, so merging, maybe it'll get optimized later. Thanks.

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -314,8 +318,8 @@ cdef class Line(VertexInstruction):
             indices_count += 12
             vertices_count += 4
         elif cap == LINE_CAP_ROUND:
-            indices_count += (self._cap_precision * 3) * 2
-            vertices_count += (self._cap_precision) * 2
+            indices_count += 2 * (self._cap_precision + 1) * 3 * 2
+            vertices_count += 2 * (self._cap_precision + 1) * 2
```
KeyWeeUsr: @tshirtman This is *not* the right approach. If you look in the docs, you'll see what [`cap_precision`](https://kivy.org/docs/api-kivy.graphics.html#kivy.graphics.Line.cap_precision) is documented as and this line basically invalidates everything in there, no? The same would apply for the indicies as it's bound to the vertices.

Basically what I'm trying to say:

image

Let's say the "cap" is a semicircle and let's take the `cap_precision == 4` for simplicity. The red lines in the previous implementation would represent this part:

> Number of iteration for drawing the "round" cap [...]

Therefore the total count of points/vertices should be `self._cap_precision - 1` for the semicircle i.e. the line 599 or one of these: `2 * cap_precision` or `2 * cap_precision - 1` depending on whether the center point should be duplicated on both ends or if it takes the point from the already existing line segment. I believe _**`4`**_` * (cap_precision + 1)` is bad approach. That would also make the indices `vertices_count * 3`.

The PR author pretty much stated he draws the full circles, but it seems strangely wasteful and instead I believe the wrong thing (perhaps except +/- 1 in the numbers, but if it worked before, don't fix it) is actually the starting angle ([picture](https://math.stackexchange.com/a/1201367/320970)). 

```python
>>> from math import atan2, degrees
>>> degrees(atan2(0, 1))
0.0
>>> degrees(atan2(1, 0))
90.0
>>> degrees(atan2(0.5, 0.5))
45.0
>>> degrees(atan2(0.3, 0.7))
23.19859051364819
```

I haven't really poked in the code, I only read it, however when I draw the line from the example in the issue up ↔ down, the start angle is either `0`(most likely?) or `90`, when I draw left → right the line is correct and when I draw right → left, the caps are hidden in the line itself (good angle, bad direction because we default to counterclockwise drawing for that).

Basically the PR "fixes" only the "looks" not the "brains" and works only because you can't see the wrong direction if you overdraw the only part that actually shows that the direction is wrong (the diameter segment).

tshirtman: Yes, i agree the fix is crude, and thanks for the writeup, i'm looking again at it, and hope to fix it proper.

Issue #4484, opened 2016-07-23
CoumpoundSelection / RecycleView selection: Allow to unselect an item when "multiselect: False"


youpsla:
To allow to unselect an item when mutliselect is False, I suggest deleting line 294 of Coumpoundselection.py : « self.select_node(node) »

It works for me. Allow to unselect the only selected item.

saqib1707:
@dessant I think this feature (to be able to  unselect an item when multiselect=False ) has to be there if one doesn't wants to select anything.I am opening a PR for the same.

dessant:
Fixed in https://github.com/kivy/kivy/issues/4782.

PR #4776, opened 2016-12-08
#4484: Allow to unselect an item when multiselect=False

saqib1707:
When multiselect = False , then just added a condition to deselect a already selected node.
Test code-https://github.com/kivy/kivy/blob/master/examples/widgets/compound_selection.py
#4484

### DIFF
--- a/kivy/uix/behaviors/compoundselection.py
+++ b/kivy/uix/behaviors/compoundselection.py
@@ -312,7 +312,6 @@ class CompoundSelectionBehavior(object):
                 self.deselect_node(node)
             else:
                 self.clear_selection()
-                self.select_node(node)
         elif range_select:
             # keep anchor only if not multiselect (ctrl-type selection)
             self._select_range(multiselect, not multiselect, node, 0)
--- a/kivy/uix/behaviors/compoundselection.py
+++ b/kivy/uix/behaviors/compoundselection.py
@@ -308,7 +308,7 @@ class CompoundSelectionBehavior(object):
             return False
 
         if (node in self.selected_nodes and (not range_select)):  # selected
-            if ((not multi) or multiselect):
+            if multiselect:
                 self.deselect_node(node)
             else:
                 self.clear_selection()
--- a/doc/sources/installation/installation-windows.rst
+++ b/doc/sources/installation/installation-windows.rst
@@ -77,13 +77,13 @@ To use them, instead of doing ``python -m pip install kivy`` we'll install one o
 these wheels as follows.
 
 .. |cp27_win32| replace:: Python 2.7, 32bit
-.. _cp27_win32: https://kivy.org/downloads/appveyor/kivy/Kivy-1.9.2-cp27-cp27m-win32.whl
+.. _cp27_win32: https://kivy.org/downloads/appveyor/kivy/Kivy-1.9.2.dev0-cp27-cp27m-win32.whl
 .. |cp34_win32| replace:: Python 3.4, 32bit
-.. _cp34_win32: https://kivy.org/downloads/appveyor/kivy/Kivy-1.9.2-cp34-cp34m-win32.whl
+.. _cp34_win32: https://kivy.org/downloads/appveyor/kivy/Kivy-1.9.2.dev0-cp34-cp34m-win32.whl
 .. |cp27_amd64| replace:: Python 2.7, 64bit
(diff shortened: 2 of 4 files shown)


kivy/uix/behaviors/compoundselection.py
```diff
@@ -308,7 +308,7 @@ def select_with_touch(self, node, touch=None):
             return False
 
         if (node in self.selected_nodes and (not range_select)):  # selected
-            if multiselect:
+            if ((not multi) or multiselect):
```
udiboy1209: Don't think the brackets are necessary for python, you may remove them.

saqib1707: deselecting when multiselect=True works with ctrl pressed.

saqib1707: Also in android or other touch devices , one has to set multiselect=True as well as touch_multiselect=True for getting the feature of multiselection of nodes.

udiboy1209: Okay! didn't know that. It works when multiselect=True too :smile:

kivy/uix/behaviors/compoundselection.py
```diff
@@ -308,7 +308,7 @@ def select_with_touch(self, node, touch=None):
...
                 self.deselect_node(node)
             else:
                 self.clear_selection()
```
dessant: I may be wrong, but line 315 (`self.select_node(node)`) may need to be removed instead of the the changes in this pr, for getting the desired effect of being able to deselect the last selected item, when multiselect is False.

dessant: Aaaand the sulution was spelled out in the original issue, only the line numbers changed. 😋 
https://github.com/kivy/kivy/issues/4484#issue-167191367

saqib1707: @dessant No actually what happens is deleting line 315 works for multiselect=False but it causes another problem that when self.multiselect=True and touch_multiselect=False(suppose) then if one doesn't presses the shift key then range_select=False => and when  one clicked on one of the selected nodes(out of many) then the first if condition evaluates to True . The multiselect in line 291 is False since he/she hasn't pressed either ctrl btn => it will clear all selections in line 314 and even won't select the node which was clicked on but the behaviour should be that the node which is clicked should remain selected and others should get deselected. Thats why that line has to be there.

dessant: @saqib1707, removing line 315 also changes the current behavior for `touch_multiselect=False`, not just `multiselect=False`, and allows deselecting the last item by touch when touch_multiselect=False.

Though you are right about the edge case you have described, it looks like line 315 should not be deleted, but put behind an `if` block and only called if the conditions are met for your use case.

saqib1707: @dessant  I think its already been taken care of since line 315 is executed only when self.multiselect=True and touch_multiselect as well as ctrl_down are False and that is my use case. So no need to put an if I guess.

dessant: With your changes, the last item is *not* deselected on touch, when `touch_multiselect=False` and `multiselect=True`.

dessant: The goal of a fix would be to be able to deselect the last selected item regardless of the value of `touch_multiselect` and `multiselect`, while still handling edge cases, one of them being the one you have found.

saqib1707: @dessant when touch_multiselect=False and self.multiselect=True , then to deselect any single selected node one has to (press + ctrl_down) to deselect it (that was the feature earlier) . But if u want that on clicking on an already selected node should clear all selections including the one which is clicked upon then simply removing the line 315 works fine.... I think i was wrong earlier then.

dessant: Yes, it should deselect the last selected item on a simple touch, without `Ctrl`.

Though you were right about the other edge case, if we remove the line as I said, and then select a range of widgets with `Shift`, and then click on one of the selected widgets, everything is deselected instead of keeping the clicked widget selected and deselecting the rest. That is wrong.

Try discarding your changes and work from there, ensuring that the reselection happens only if needed. I expect it to be a change (an `if` statement) right before that line, but of course you have to test and see what really works.

PR #4779, opened 2016-12-09
Obvious typo in the collide_point check

llfkj:
My first attempt to use multigesture in my own code failed, but the bug looks to be a typo in the collision check for `on_touch_move`


llfkj:
This bug doesn't show in the examples/demo/multigesture example, but that's because the entire y=y diagonal is within the GestureSurface, so the check will essentially always pass.

udiboy1209:
Nice catch @llfkj ! Merging this. Thanks :tada:

llfkj:
My wife gets the credit for spotting this one.

PR #4781, opened 2016-12-10
Fix#4484 : Allow unselect an item when multiselect is False

saqib1707:
#4484


dessant:
Please read our discussion on https://github.com/kivy/kivy/pull/4776#discussion_r91833587 again, this is not the right fix.

PR #4782, opened 2016-12-10
Fix#4484 :Allow unselect an item when multiselect is False

saqib1707:
#4484


dessant:
@saqib1707, thanks a lot, this solution seems good.

PR #4790, opened 2016-12-13
Different ways to express same colour in  a  Kivy App (rgb,rgba,hex,hsv).

phunsukwangdu:
The following PR shows the different ways to express same colours in the Kivy App.


KeyWeeUsr:
@Zen-CODE now I understand what you meant at irc. I don't like the layout examples as they are basically [here](https://kivy.org/docs/api-kivy.uix.layout.html) and from there you can navigate to each layout and there are screenshots or gifs.

Also there are tutorial explaining some of the layouts already mentioned and some of the layouts (if not all of them) already have a runnable example in its `.py` file (the stuff under `if __name__ == '__main__'`).

However, the color example is nice.

dessant:
I second the same sentiment as @KeyWeeUsr and @Zen-CODE, the essence of setting a color in different formats could have been presented in less than 50 lines.

@phunsukwangdu, paths like `examples/RGB usage example/main.py ` are problematic from multiple reasons. It needs to be quoted when referenced on the command line because of the spaces, and it contains parts which are not necessary, like `usage example`. My first attempt would be `examples/color/formats/main.py `.

And as always, pep8 needs to be followed. 😋 

+1 for closing.

Zen-CODE:
Perhaps the example/widgets/colorpicker.py is a good place to put this? It already shows how colors can be represented in multiple ways, although the way it presents is obscure. Perhaps you could refine that, and clarify the representation in different formats?

phunsukwangdu:
Yes i will remove the layout example as if it is not required.But the example was to illustrate how we can nest various layouts.There is no internal resources on it.
I will close this PR and send new PRs which will the following changes:-
    -remove layout example
    -appropriate naming of folders
    -appropriate placing of the example(most probably in  )
    -separate PR for different examples
    -recheck pep8 errors(although i did it.got it from (sudo apt-get install pep8) )

Issue #4793, opened 2016-12-13
Recycleview: apply_selection gets called way too often


ngoonee:
So I'm trying to get back the behavior I used to have with the 'old' recycleview on latest master. Here's my slightly modified Selectable example to illustrate the issue.

    from kivy.app import App
    from kivy.lang import Builder
    from kivy.uix.recycleview import RecycleView
    from kivy.uix.recycleview.views import RecycleDataViewBehavior
    from kivy.uix.label import Label
    from kivy.properties import BooleanProperty
    from kivy.uix.recycleboxlayout import RecycleBoxLayout
    from kivy.uix.behaviors import FocusBehavior
    from kivy.uix.recycleview.layout import LayoutSelectionBehavior
    from kivy.animation import Animation

    Builder.load_string('''
    <SelectableLabel>:
        canvas:
            Color:
                rgba: (.0, 0.9, .1, .3) if self.selected else (1, 1, 1, 0)
            Rectangle:
                pos: self.pos
                size: self.size

    <RV>:
        viewclass: 'SelectableLabel'
        SelectableRecycleBoxLayout:
            default_size: None, dp(56)
            default_size_hint: 1, None
            size_hint_y: None
            height: self.minimum_height
            orientation: 'vertical'
    ''')

    class SelectableRecycleBoxLayout(FocusBehavior, LayoutSelectionBehavior,
                                     RecycleBoxLayout):
        ''' Adds selection and focus behaviour to the view. '''

    class SelectableLabel(RecycleDataViewBehavior, Label):
        ''' Add selection support to the Label '''
        index = None
        selected = BooleanProperty(False)
        selectable = BooleanProperty(True)

        def refresh_view_attrs(self, rv, index, data):
            ''' Catch and handle the view changes '''
            self.index = index
            return super(SelectableLabel, self).refresh_view_attrs(
                rv, index, data)

        def on_touch_down(self, touch):
            ''' Add selection on touch down '''
            if super(SelectableLabel, self).on_touch_down(touch):
                return True
            if self.collide_point(*touch.pos) and self.selectable:
                return self.parent.select_with_touch(self.index, touch)

        def apply_selection(self, rv, index, is_selected):
            ''' Respond to the selection of items in the view. '''
            self.selected = is_selected
            if is_selected:
                print("selection changed to {0}".format(rv.data[index]))
                anim = Animation(height=88, d=1)
                #self.height = 88
            else:
                print("selection removed for {0}".format(rv.data[index]))
[...]

PR #4794, opened 2016-12-13
Check for change in selected-ness before apply_selection

ngoonee:
`apply_selection` should only need to be called if the value is different from the current value. Linked to issue https://github.com/kivy/kivy/issues/4793


ngoonee:
Agreed. I'll continue discussion in the issue.

PR #4795, opened 2016-12-13
[WIP] Pep8 fixes of the current master

udiboy1209:
I have updated the Makefile to check pep8 with the newer script (`kivy/tools/pep8checker/pep8.py`) but these changes should also pass with the old script.

Help appreciated in fixing the whole repo :smile:


dessant:
> I have updated the Makefile to check pep8 with the newer script (kivy/tools/pep8checker/pep8.py)

I don't understand the Makefile change, `pep8kivy` imported `pep8` and added other features, like being able to exclude paths.

udiboy1209:
@dessant  Ohh my bad! I thought `pep8.py` was a newer script. Now I get what you meant by

> udiboy1209: the included pep8 module is old

I thought that referred to `pep8kivy.py`.

So do we want to update the pep8 module? It's missing a lot of newer errors. Anyway why are we using a custom pep8 and not the pip-installed version ?

dessant:
Yes, updating `pep8.py`would be good, the original code was taken from here: https://github.com/PyCQA/pycodestyle/blob/master/pycodestyle.py

There are two commits which should not be lost though:
https://github.com/kivy/kivy/commit/a2ea270f750d90673f63dbc7b6ed63b8c5c1d7ce#diff-43b700e60750dd13b002d04cc62dd894
https://github.com/kivy/kivy/commit/0c34d760570ec19d4ce0eca1317e441603113443

dessant:
@udiboy1209, we could also switch to installing from pip, if we set up a way to add custom checks, like the commits linked above.

This update/switch can be done in a separate pr from the pep8 fixes of course.

udiboy1209:
> If you want to provide extensibility / plugins, please see flake8 - pep8 doesn’t want or need a plugin architecture.

This is in the Developer's notes :stuck_out_tongue: .

dessant:
A very old leftover indeed, i'll make a pr. 😁 https://github.com/PyCQA/pycodestyle/commit/4b48c39ca79f9bd6b6ef2490d867209ef5f0f336#diff-f0b41b6133ffe3efa4aed9b6033b32d2R12

PR #4798, opened 2016-12-14
Update pep8.py to version 1.7.0

udiboy1209:
Update pep8.py to 1.7.0 while copying over [a2ea270](https://github.com/kivy/kivy/commit/a2ea270f750d90673f63dbc7b6ed63b8c5c1d7ce#diff-43b700e60750dd13b002d04cc62dd894) and [0c34d76](https://github.com/kivy/kivy/commit/0c34d760570ec19d4ce0eca1317e441603113443). 

@dessant  there are 1109 instances of E402(module level import not at top of file). Should I add it to the ignore list?

### DIFF
--- a/kivy/tools/pep8checker/pep8.py
+++ b/kivy/tools/pep8checker/pep8.py
@@ -1,5 +1,9 @@
-# pep8.py - Check Python source code formatting, according to PEP 8
-# Copyright (C) 2006 Johann C. Rocholl <johann@rocholl.net>
+#!/usr/bin/env python
+# pycodestyle.py - Check Python source code formatting, according to PEP 8
+#
+# Copyright (C) 2006-2009 Johann C. Rocholl <johann@rocholl.net>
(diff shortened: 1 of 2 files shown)


dessant:
Yes, that could be ignored.

dessant:
Just seen that the current version is 2.2.0, did you use an alternate source?
https://github.com/PyCQA/pycodestyle/releases
https://github.com/PyCQA/pycodestyle/blob/2.2.0/pycodestyle.py

udiboy1209:
@dessant  I downloaded the source from PyPi and its 1.7.0 on PyPi so that is what I used. Want to update to 2.2.0 from github?

dessant:
Yes, it would be preferred. They renamed the project to pycodestyle: https://pypi.python.org/pypi/pycodestyle.

udiboy1209:
Ahh lol alright :+1:

PR #4747, opened 2016-11-21
Update known issues for iOS packaging

gfyoung:
Removes known issue that iOS project is not exportable because that was patched in <a href="https://github.com/kivy/kivy-ios/pull/228">kivy/kivy-iOS #228</a>.  In addition, rewrites known issue section to point to the GitHub issues tracker page for `kivy/kivy-ios`.

cc @Zen-CODE

PR #4800, opened 2016-12-14
Add missing word to iOS packaging issues section

gfyoung:
Follow-up from #4747.  Looks like this was overlooked.

### DIFF
--- a/doc/sources/guide/packaging-ios.rst
+++ b/doc/sources/guide/packaging-ios.rst
@@ -105,9 +105,15 @@ for more information.
 Known issues
 ------------
 
-All known issues with packaging for iOS are currently tracked on our `issues <https://github.com/kivy/kivy-ios/issues>`_ page. If you encounter an issue specific to packaging for iOS that isn't listed there, please feel free to file a new issue, and we will get back to you on it.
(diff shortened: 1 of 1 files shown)


gfyoung:
Added [ci skip] to commit because this is a simple doc change, and test failures seem to be unrelated to this PR judging from what I'm seeing on Travis and Appveyor.

KeyWeeUsr:
kivy-ios seems to like pep8, please trim the lines appropriately :)

gfyoung:
But `pep8` only applies to `.py` files?  Also, I don't see any offending lines.

KeyWeeUsr:
Every file in here (except travis, appveyor and some cython files) is under pep8 at least what I see. Mostly it applies only to python & cython, but documentation is at least under `<=79` rule (because it's more readable). The only lines that seem to be violating this rule are 108 and 110.

Also, it's kind of obvious if you look at the whole file 😋

gfyoung:
If you are PEP8-ing the documentation files, fair enough.  Fixed.

KeyWeeUsr:
++ Thanks. 🐱

PR #4803, opened 2016-12-15
Fix a silly multistroke crash

llfkj:
The multistroke engine will crash if you draw a perfectly horizontal (or presumably vertical) line.

This just looks like a fencepost error.

This fix ensures the computed bounding box is at least 1 coordinate high (or wide). 
While here, improve the error message so it is clear why it is crashing.

### DIFF
--- a/kivy/multistroke.py
+++ b/kivy/multistroke.py
@@ -1351,7 +1351,7 @@ def scale_dim(points, size, oneDratio):
     bbox_x, bbox_y, bbox_w, bbox_h = bounding_box(points)
 
     if bbox_h == 0 or bbox_w == 0:
-        raise MultistrokeError('scale_dim() called with invalid points')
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
Is there any way how to reproduce it on user's side or is it an error when inputting the line manually (code)?

KeyWeeUsr:
@llfkj ping

llfkj:
You should be able to try this example (load it as a database in `examples/demo/multistroke`, then try drawing anything.

https://gist.github.com/llfkj/6b907dd973a5c9ec3b6ec98616c84b5b

KeyWeeUsr:
When I loaded the file, it crashed even without any perfect line. It was enough that it actually _recognized_ the gesture. This fix removes that issue + the perfect lines issue too. Most likely related.

Thanks! 🐼

PR #4805, opened 2016-12-15
Fix #4730 :ActionGroup behaviour becomes normal.First click works perfectly.

### LINKED ISSUE #4730: ActionGroup needs clicking twice
Is it only me, or does the `ActionGroup` requires clicking twice? It worked normally before, at least in the beginning of 2016, but recently when I worked with it for a while I noticed this thing. The first touch seems to be ignored, but _isn't_ (try `on_release`), yet the `DropDown` won't show. When clicked the second time, it works normally. Seems like there's some issue with initialising the widget or something.

Run just the basic example in `actionbar.py` and try to click on either "Group 1" or "Group 2". I ran the  `showcase` demo too, but there the `Spinner` that looks like `ActionGroup` behaves normally because it inherits directly from `Spinner` and `ActionItem` in kv file and isn't built as a casual `ActionGroup`. Any ideas what's going on?

saqib1707:
#4730 The problem occurs when one doesn't specifies max_height of the dropdown then in these lines https://github.com/kivy/kivy/blob/master/kivy/uix/dropdown.py#L340-L343 , the max_height = None and hence for the first click self.container.minimum_height = 0 => self.height = 0  due to which first click on the ActionDropDown doesn't shows dropdown list.

### DIFF
--- a/kivy/uix/dropdown.py
+++ b/kivy/uix/dropdown.py
@@ -33,6 +33,7 @@ dropdown. ::
         # When adding widgets, we need to specify the height manually
         # (disabling the size_hint_y) so the dropdown can calculate
         # the area it needs.
+
         btn = Button(text='Value %d' % index, size_hint_y=None, height=44)
 
         # for each button, attach a callback that will call the select() method
@@ -223,7 +224,8 @@ class DropDown(ScrollView):
             return True
 
     def on_container(self, instance, value):
-        pass
+        if value is not None:
+            self.container.bind(minimum_size=self._reposition)
 
     def open(self, widget):
         '''Open the dropdown list and attach it to a specific widget.


KeyWeeUsr:
Nice fix. Please run it through `pep8`, so that you can remove the errors :)

saqib1707:
@KeyWeeUsr But the pep8  error is in the entire kivy directory not only in dropdown.py .

udiboy1209:
@saqib1707 you should use kivy's pep8 script like this for specific files: 

    python kivy/tools/pep8checker/pep8kivy.py <filename>

I don't think there are pep8 errors in your change, but it would be great if you can fix pep8 errors in the entire file, and test it. Thanks!

dessant:
`#when` to `# when`

KeyWeeUsr:
Please remove the commit that merges from `master`, correct the pep8 errors and ensure you use correct branches in future. ^^ PR with 221 changed files isn't trivial to check and bad stuff might happen after such merge (errors might be overlooked).

    git checkout actiongroup-needs-clicking-twice
    git reset HEAD~1
    git push -f

KeyWeeUsr:
Thanks! 🍰

PR #4813, opened 2016-12-16
Change default font to core.text.DEFAULT_FONT

KeyWeeUsr:
Partial rebase of #2412 without the setting function. Closes #2412

### DIFF
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -192,6 +192,11 @@ Available configuration tokens
         time.
     `kivy_clock`: one of `default`, `interrupt`, `free_all`, `free_only`
         The clock type to use with kivy. See :mod:`kivy.clock`.
+    `default_font`: list, defaults to ['Roboto',
+    'data/fonts/Roboto-Regular.ttf', 'data/fonts/Roboto-Italic.ttf',
+    'data/fonts/Roboto-Bold.ttf', 'data/fonts/Roboto-BoldItalic.ttf']
+
+        Default font used for widgets displaying any text.
 
 :input:
 
@@ -266,6 +271,7 @@ Available configuration tokens
 .. versionchanged:: 1.9.2
     `min_state_time` has been added to the `graphics` section.
     `kivy_clock` has been added to the kivy section
+    `default_font` has beed added to the kivy section
 
 .. versionchanged:: 1.9.0
     `borderless` and `window_state` have been added to the graphics section.
@@ -312,7 +318,7 @@ from weakref import ref
 _is_rpi = exists('/opt/vc/include/bcm_host.h')
 
 # Version number of current configuration format
-KIVY_CONFIG_VERSION = 16
+KIVY_CONFIG_VERSION = 17
 
 Config = None
 '''The default Kivy configuration object. This is a :class:`ConfigParser`
@@ -804,6 +810,14 @@ if not environ.get('KIVY_DOC_INCLUDE'):
 
         elif version == 15:
             Config.setdefault('kivy', 'kivy_clock', 'default')
+
+        elif version == 16:
+            Config.setdefault('kivy', 'default_font', [
+                'Roboto',
+                'data/fonts/Roboto-Regular.ttf',
+                'data/fonts/Roboto-Italic.ttf',
+                'data/fonts/Roboto-Bold.ttf',
+                'data/fonts/Roboto-BoldItalic.ttf'])
 
         # elif version == 1:
         #    # add here the command for upgrading from configuration 0 to 1
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -36,9 +36,11 @@ __all__ = ('LabelBase', 'Label')
 
 import re
 import os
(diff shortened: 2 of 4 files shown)


dessant:
It's a bit tricky, I also played with a fix but gave up at the time :P. See this thread for some of the pitfalls, and possible solutions: https://github.com/kivy/kivy/pull/2319.

The main benefit of setting a default font would be to not have to ship/load the font files provided by kivy, but if we must customize it by importing `kivy.core.text`, it's already too late because the module [loads](https://github.com/kivy/kivy/blob/e7299f4e7f6eb3ddf773c26a303af1dd52b90e27/kivy/core/text/__init__.py#L778) the default ones on import.

Setting/reading the default font from Config may be the only practical route at this time. That would involve declaring the font name and 4 paths, and finding a syntax which doesn't offend anyone. 😁

Here is one possible version:
```python
Config.set('kivy', 'default_font', '<Name>,<path_regular>,<path_italic>,<path_bold>,<path_bolditalic>')
```
A similar syntax is used already for some of the config options, but the comma could cause issues with certain paths.

KeyWeeUsr:
Config seems to me like the was as even fullscreen is set this way. It's a _default_ value, it should be fetched from config, right?

Also, what's the problem with commas as separator? In the worst case we can do this:

    Config.set('kivy', 'default_font', '[<Name>,<path_regular>,<path_italic>,<path_bold>,<path_bolditalic>]')
 
and call Python's `ast` this way:

    ast.literal_eval('[<Name>,<path_regular>,<path_italic>,<path_bold>,<path_bolditalic>]')

which will return a list and is [quite safe](https://docs.python.org/3/library/ast.html#ast.literal_eval).

KeyWeeUsr:
@dessant re the pitfalls, override it like this:

    Config.set(...)
    Config.write()

and the `config.ini` is overwritten, therefore on the **_next_** run of the app the font is properly changed.

dessant:
I think this is now a decent way of solving it, if someone doesn't object until tomorrow, we can merge it.

kivy/config.py
```diff
@@ -805,6 +811,14 @@ def name(self, value):
         elif version == 15:
             Config.setdefault('kivy', 'kivy_clock', 'default')
 
+        elif version == 16:
+            Config.setdefault('kivy', 'default_font', [
```
dessant: Should the value be a string here?

dessant: Sorry, we do it internally; https://github.com/kivy/kivy/blob/master/kivy/config.py#L433.

KeyWeeUsr: it's a list O.o

dessant: Yeah, i thought it would fail, but kivy handles it nicely.

KeyWeeUsr:
However, there might be a problem with a need of **local** config:

    Config.read(<path to local .ini>)
    Config.set(...)
    Config.write()

Config somehow skips the local configuration and goes for `.kivy/config.ini`, but I think it's a bug in Config, not a problem of this PR.

kivy/config.py
KeyWeeUsr: Yup and `literal_eval` comes to the rescue for converting it back 🍡 . It might be included in Config though.

KeyWeeUsr:
Yep, was a config thing. Unfortunatelly, `kivy/__init__.py` sets it by default. It's nice though, because if prohibits dirty tricks ^^

```
import os
from os.path import dirname, abspath, join, exists
local_config = join(dirname(abspath(__file__)), 'local_kivy')
os.environ['KIVY_HOME'] = local_config

if not exists(os.environ['KIVY_HOME']):
    os.mkdir(local_config)

Config.set('kivy', 'default_font', [
    'Arial',
    local_config + '\\..\\arial.ttf',
    local_config + '\\..\\arial.ttf',
    local_config + '\\..\\arial.ttf',
    local_config + '\\..\\arial.ttf',
])
Config.write()

from kivy.lang import Builder
Button:
    text: 'blaaaaaaaaaaaaa'
'''))
```

First run is still "cached", because Config isn't reloaded, but even after `import kivy` and some other stuff I tried, it was still the default font until the app was killed and ran for the second time, when the local config was finally read.

Any ideas how to restart Config after `Config.write()`?

ref #4826

dessant:
@KeyWeeUsr, super, it works well with my simple test.
```python
Config.set('kivy', 'default_font', [
    'AveriaSansLibre',
    'AveriaSansLibre-Regular.ttf',
    'AveriaSansLibre-Italic.ttf',
    'AveriaSansLibre-Bold.ttf',
    'AveriaSansLibre-BoldItalic.ttf',
])

from kivy.app import App
from kivy.lang import Builder


kv = """
Label:
    font_size: dp(30)
    text: "I feel pretty,\\nOh, so pretty,"
"""


class TestApp(App):
    def build(self):

if __name__ == '__main__':
    TestApp().run()
```

Il do a small change, mainly for allowing to skip italic, bold and bolditalic.

If someone has an idea for a different config value syntax, please share.

PR #4819, opened 2016-12-17
Remove extension support, closes #3594

### LINKED ISSUE #3594: Remove the current extension support
There was a discussion about extensions, @tshirtman and @tito mentioned that they do not seem to be used at all and should be removed.

dessant:


dessant:
@thopiekar, in case it affects daily ppa builds, these folders are gone:
```
kivy/ext
kivy/tools/extensions

Issue #4697, opened 2016-10-29
OpenCV Camera initialization error (kivy 1.9.1+)


keigezellig:
When using the OpenCV camera provider it gives the following error when starting up:
[INFO              ] [Window      ] virtual keyboard not allowed, single mode, not docked
[INFO              ] [Text        ] Provider: sdl2
[INFO              ] [GL          ] NPOT texture support is available
[INFO              ] [Clipboard   ] Provider: gtk3(['clipboard_dbusklipper'] ignored)
xclip version 0.12
Copyright (C) 2001-2008 Kim Saunders et al.
Distributed under the terms of the GNU GPL
[INFO              ] [Cutbuffer   ] Provider: xclip
[INFO              ] [CutBuffer   ] cut buffer support enabled
[CRITICAL          ] [Camera      ] Unable to find any valuable Camera provider at all!
pygst - ImportError: No module named 'pygst'
  File "/usr/lib/python3/dist-packages/kivy/core/**init**.py", line 59, in core_select_lib
    fromlist=[modulename], level=0)
  File "/usr/lib/python3/dist-packages/kivy/core/camera/camera_pygst.py", line 15, in <module>
    import pygst

**opencv - ImportError: No module named 'cv'
  File "/usr/lib/python3/dist-packages/kivy/core/**init**.py", line 59, in core_select_lib
    fromlist=[modulename], level=0)
  File "/usr/lib/python3/dist-packages/kivy/core/camera/camera_opencv.py", line 20, in <module>
    import cv**

I am using Python 3.4, OpenCV 3.1.0 and verified that OpenCV from the Python interpreter works. However i notice that the import needed is `import cv2` instead of `import cv`

udiboy1209:
There is this breaking change in OpenCV 3 where they have removed `cv.py` entirely ([ref](http://stackoverflow.com/questions/32048313/cannot-import-cv-on-opencv2)). So functions in `camera_opencv.py` need to be updated to work with OpenCV3. I suggest you use some other camera provider for the meanwhile.

keigezellig:
Well that's really unfortunate, since the other provider (pygst) doesn't
work either (at least with python 3 it is)
I get this error (i installed the python3-gst, python-gst and
python-gst-dev packages)

pygst - ImportError: No module named 'pygst'
  File "/usr/lib/python3/dist-packages/kivy/core/__init__.py", line 59, in
core_select_lib
    fromlist=[modulename], level=0)
[...]

PR #4828, opened 2016-12-18
added support for opencv 2 and 3

amateusz:
I updated this file to deal with current opencv versions 2.4 and 3.x.
There are elif branches to initalize everything accordingly. Methods differ. My fav one is @ line 106 'frame.width' -> 'frame.shape[0]' indicating different datatype..
Opencv's naming convention is sick IMO.
(tested both on win10 and ver.3 on RPi)

*** that's my first PR, so please correct me if I'm wrong or something ***

### DIFF
--- a/kivy/core/camera/camera_opencv.py
+++ b/kivy/core/camera/camera_opencv.py
@@ -43,7 +43,8 @@ except ImportError:
     # opencv 2 case (and also opencv 3, because it still uses cv2 module name)
     try:
         import cv2
-        # here missing this OSX specific highgui thing. I'm not on OSX so don't know if it is still valid in opencv >= 2
+        # here missing this OSX specific highgui thing.
+        # I'm not on OSX so don't know if it is still valid in opencv >= 2
     except ImportError:
         raise
 
@@ -54,7 +55,8 @@ class CameraOpenCV(CameraBase):
     '''
 
     def __init__(self, **kwargs):
-        # we will need it, because constants have different access paths between ver. 2 and 3
+        # we will need it, because constants have
+        # different access paths between ver. 2 and 3
         try:
             self.opencvMajorVersion = int(cv.__version__[0])
         except NameError:
@@ -91,9 +93,9 @@ class CameraOpenCV(CameraBase):
             # and get frame to check if it's ok
             frame = hg.cvQueryFrame(self._device)
             # Just set the resolution to the frame we just got, but don't use
-            # self.resolution for that as that would cause an infinite recursion
-            # with self.init_camera (but slowly as we'd have to always get a
-            # frame).
+            # self.resolution for that as that would cause an infinite
+            # recursion with self.init_camera (but slowly as we'd have to
+            # always get a frame).
             self._resolution = (int(frame.width), int(frame.height))
             # get fps
             self.fps = cv.GetCaptureProperty(self._device, cv.CV_CAP_PROP_FPS)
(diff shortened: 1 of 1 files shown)


udiboy1209:
This looks like a great change :+1: ! When I run the camera example, all I get is a white screen. But my camera flash switches on so I know that the camera is switching on. (Ubuntu 16, python 2.7 opencv3).

Also, there are some pep8 errors which you should fix. You can see the list of errors in the [Travis CI log](https://travis-ci.org/kivy/kivy/jobs/185025124)

udiboy1209:
Fixes #4697

kivy/core/camera/camera_opencv.py
```diff
@@ -97,12 +146,9 @@ def _update(self, dt):
 
     def start(self):
         super(CameraOpenCV, self).start()
-        if self._update_ev is not None:
-            self._update_ev.cancel()
-        self._update_ev = Clock.schedule_interval(self._update, self.fps)
+        Clock.unschedule(self._update)
```
dessant: Referencing the event instead of the method is much faster when unscheduling.
If the reason of the change was only to refactor, it should not be done in this case.

kivy/core/camera/camera_opencv.py
```diff
@@ -97,12 +146,9 @@ def _update(self, dt):
...
+        Clock.schedule_interval(self._update, self.fps)
 
     def stop(self):
         super(CameraOpenCV, self).stop()
-        if self._update_ev is not None:
-            self._update_ev.cancel()
-            self._update_ev = None
+        Clock.unschedule(self._update)
```
dessant: Same as https://github.com/kivy/kivy/pull/4828/files#r93962100.

dessant:
@amateusz, great contribution. If you have time to clean it up until next week, it could be part of the next release.

dessant:
With some other camera providers removed, this should be a release blocker.

hunternet93:
Very nice, hope this gets committed soon! However, there's a bug: when the OpenCV-reported FPS is used to schedule _update(), it doesn't convert frames/second to seconds, for example a 30fps source will update once every 30 seconds, causing a while box to be displayed until the first update. Dividing fps into 1 fixes the issue.

dessant:
@hunternet93, with opencv 3.2 the returned fps is `0.03333333333333333` on my part, could you give us more details to reproduce the bug? Opening a new issue would be the best way.

dessant:
@udiboy1209, same ^, we will need a way to reproduce it.

udiboy1209:
@dessant if you are talking about the white screen issue I posted above, I simply ran the camera example from kivy-examples. I will try again with #5011 and get back.

dessant:
@udiboy1209, does it work with opencv 2 on the same distro? Maybe the bug is only present on with opencv 3 on linux, cv3 works on win, and cv2 was confirmed to work on linux. https://github.com/kivy/kivy/pull/5011

@phunsukwangdu, is it possible that the frame rate (fps vs decimal) returned by opencv is dependent on the camera type/driver?

hunternet93:
@dessant It's possible that the driver was returning incorrect or unexpected values, I was using [v4l2loopback](https://github.com/umlaeute/v4l2loopback) on Linux, with the latest release of OpenCV on Python 3.

dessant:
It looks like the only reliable way to get a framerate is to measure it: https://www.learnopencv.com/how-to-find-frame-rate-or-frames-per-second-fps-in-opencv-python-cpp/.

We're hardcoding 30 fps when the fps cannot be retrieved, but that's not ideal either: https://github.com/kivy/kivy/pull/4828/files#diff-f184da55c1968a44465aa295d502abc5R120

I'll be available for reviewing if someone would like to work on a pr.

phunsukwangdu:
@dessant i am clueless regarding this.i will have look in this matter. :)

phunsukwangdu:
@dessant when i use my internal webcam it works fine!!.But when i use my logitech webcam it freezes.but y?.
https://github.com/opencv/opencv/issues/4355

PR #4836, opened 2016-12-21
Bubble reachable on android when touch in textinput is made near the ends.

saqib1707:
#4672


KeyWeeUsr:
Works nice even on the sides, not only top+bottom. Thanks :dango:

PR #4837, opened 2016-12-21
Restore gl/gles selection at compile-time.

matham:
Should fix some of the ios/android issues.


inclement:
I was able to get the Android build working with this branch, so seems good to me.

There were some issues though:

- I had to delete the GLES2 folder, I think because its contents don't match those in the NDK version. Is it supposed to work?
- With python2 I had to delete all the folders in `kivy/include` or I got `error: can't copy 'kivy/include/GL': doesn't exist or not a regular file`. I'm not sure what's wrong here or why it didn't happen with python3.
- I had to manually copy the `kivy/include` folder into the `build/lib.linux-x86_64-2.7` (or equivalent) folder.

All these can be worked around in python-for-android, especially as a first step, as they probably all arise somehow from python-for-android's hacks around the normal setup.py process. So unless they're meaningful and you want to fix them, I don't think they should delay things for now.

Once this is merged, I'll code these workarounds into the kivy recipe, and we can go from there for any more robust fixes.

matham:
@tito had added the gles and gl folder/headers. It was to make it easier so that we didn't have to hunt the OS for the location of the headers. I guess we thought they would be the same or always work. Since that doesn't seem to be the case I removed them and instead we'll rely again on OS provided headers.

That should resolve your first two issues. As soon as the build passes on apple/linux I'll merge since it works on Windows.

PR #4839, opened 2016-12-22
Typo fix

Leva7:
This dot seems out of place, so it should probably be removed
image


KeyWeeUsr:
Instead of a typo it rather looks as a dead `.. versionadded::` tag someone forgot to fix. 😋

Issue #3961, opened 2016-02-01
TextInput bubble actions can fail depending on widget positioning


dessant:
The bubble buttons will do nothing in the example below, but comment out BoxLayout `padding: '100dp'` and text selection and pasting from the bubble will work.

``` python
from kivy.app import App
from kivy.lang import Builder

kv = """
BoxLayout:
    padding: '100dp'
    TextInput:
        use_bubble: True
        text: 'long press to show the bubble, then click "Select All"'
"""


class TestApp(App):
    def build(self):
        return Builder.load_string(kv)

if __name__ == '__main__':
    TestApp().run()
```

dessant:
It works with 1.8.0.

udiboy1209:
Was fiddling around with this, found that bug only appears for large padding values. Works for small enough padding values like `10dp` or so.

akshayaurora:
Cause is the change of always_release: from True to False

dessant:
The button coordinates seem to be relative to the textinput, we could to the same transformation for the touch.

saqib1707:
@dessant was trying to solve this using correct approach . Found out that even if the touch is transformed in textinput.py , there is no way to call the on_touch_up of button.py from textinput.py having touch.grab_current = bubblebutton object . https://github.com/kivy/kivy/blob/master/kivy/uix/behaviors/button.py#L162-L163 .The grabbed touch is dispatched to the on_touch_up of button.py but making transformation of touch in button.py doesn't seems reasonable I guess.

dessant:
Fixed in https://github.com/kivy/kivy/issues/4844.

PR #4844, opened 2016-12-26
Fix#3961: BubbleButton of textinput becoming functional (not changing always_release)

saqib1707:


saqib1707:
#3961

dessant:
It mostly works, but there are some bugs, like crashing when the bubble is visible and you click near it, and the coordinate space of the touch is changed in a permanent way. This should fix the issues I could find:

```python
    def on_touch_up(self, touch):
        try:
            touch.push()
            self.transform_touch(touch)
            for child in self.content.children:
                if ref(child) in touch.grab_list:
                    touch.grab_current = child
            return super(TextInputCutCopyPaste, self).on_touch_up(touch)
        finally:
            touch.pop()
```

dessant:
@saqib1707, thanks for the update.

For all contributors: please do not squash the commits which have already been pushed to pr branches, the evolution of the contribution is lost and anyone revisiting the thread will miss some important context. We are able to squash while merging, if needed.

Issue #3799, opened 2015-11-24
Unable to get Window Position w/ WindowSDL2?


lhl:
Kivy doesn't seem to support getting the Window position (I'd like to save the final window position so I can open it back up at the same location).  I see that SDL has a [SDL_GetWindowPosition()](https://wiki.libsdl.org/SDL_GetWindowPosition) function - is that something that might be added to the WindowSDL2 class? (it looks like it'd need to be added to _window_sdl2.pyx since it needs to talk to the _WindowSDL2Storage win?)

I can probably submit a pull request, but figured I'd find out what the preferred method naming/way to add this (it looks like getting Window position is applicable to most of the backends?)

dessant:
`Window.pos` for getting and setting would be good. You could use Window.size as a template for implementation. This would mean that `Window.pos`, `Window.x` and `Window.y` would be added to the api.

PR #4851, opened 2016-12-28
Add Window position manipulation

KeyWeeUsr:
ref #3799

```
from kivy.app import App
from kivy.clock import Clock


class TestApp(App):
    def show(self, dt):
        pos = Window._get_window_pos()
        Window._set_window_pos(*[p + 1 for p in pos])

    def build(self):
        Clock.schedule_interval(self.show, 0)
TestApp().run()
```

For now getter and setter only. Not sure if it needs to be a separate thing from `Window.top` for example, or if it needs to _update_ those two properties.

@dessant @matham ? 😋

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -680,12 +680,32 @@ class WindowBase(EventDispatcher):
     def _set_cursor_state(self, value):
         pass
 
+    def _get_window_pos(self):
+        pass
+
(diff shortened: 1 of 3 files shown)


KeyWeeUsr:
Yeah, well... it's SDL2 property to be specific, but if it's necessary, we could have additional kivy property that would be just inverted `top` or `left` or something.

PR #4853, opened 2016-12-29
Add 'mp4' to audio_gstplayer.py

FeralBytes:
Since programs such as LMMS create '.mp4' files that are purely audio even though they should be 'm4a' I have added support for 'mp4' files it will work so long as they are audio only.


AndiEcker:
+1 (also for using LMMS ;)

2016-12-29 15:53 GMT+00:00 FeralBytes <notifications@github.com>:

PR #4858, opened 2016-12-30
Add is_shortened to Label

### LINKED ISSUE #3753: Add 'is_shortened' BooleanProperty to Label

KeyWeeUsr:
Taken from @dessant's PR, but made the bool working even with markup

ref #4589
closes #3753

### DIFF
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -605,14 +605,9 @@ class LabelBase(object):
         if strip:
             text = text.strip()
         if uw is not None and options['shorten']:
-            short_text = self.shorten(text)
+            text = self.shorten(text)
             chr = type(text)
-            flat_text = text.replace(chr('\n'), chr(' '))
-            if flat_text != short_text:
-                self.is_shortened = True
-                text = short_text
-            else:
-                self.is_shortened = False
+            self.is_shortened = chr('\n') not in text
         else:
             self.is_shortened = False
         self._cached_lines = lines = []
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -217,6 +217,7 @@ class LabelBase(object):
 
         self.options = options
         self.texture = None
+        self.is_shortened = False
         self.resolve_font_name()
 
     @staticmethod
@@ -604,7 +605,16 @@ class LabelBase(object):
         if strip:
             text = text.strip()
         if uw is not None and options['shorten']:
-            text = self.shorten(text)
+            short_text = self.shorten(text)
+            chr = type(text)
+            flat_text = text.replace(chr('\n'), chr(' '))
+            if flat_text != short_text:
+                self.is_shortened = True
+                text = short_text
+            else:
+                self.is_shortened = False
+        else:
+            self.is_shortened = False
         self._cached_lines = lines = []
         if not text:
             return 0, 0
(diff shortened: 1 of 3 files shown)


dessant:
Run the code below and press `Toggle shorten`, `is_shortened` will stay True. If `markup` is False, `is_shortened` correctly switches to False on `Toggle shorten`. The rest seems to work as expected.

```python
from kivy.app import App
from kivy.lang import Builder

kv = '''
BoxLayout:
    orientation: 'vertical'
    TextInput:
        id: input
        text: 'TEST TEST TEST TEST TEST'
    BoxLayout:
        Button:
            text: 'Toggle markup (currently {})'.format(label.markup)
            on_press: label.markup = not label.markup
        Button:
            text: 'Toggle shorten (currently {})'.format(label.shorten)
            on_press: label.shorten = not label.shorten
        Label:
            text: 'is_shortened is {}'.format(label.is_shortened)
    Label:
        id: label
        text_size: 100, None
        text: input.text
        shorten: True
        markup: True
'''


class TestApp(App):
    def build(self):


if __name__ == '__main__':
    TestApp().run()
```

kivy/core/text/__init__.py
```diff
@@ -604,7 +605,16 @@ def render(self, real=False):
         if strip:
             text = text.strip()
         if uw is not None and options['shorten']:
-            text = self.shorten(text)
+            short_text = self.shorten(text)
+            chr = type(text)
+            flat_text = text.replace(chr('\n'), chr(' '))
+            if flat_text != short_text:
```
matham: This seems expensive as it's done twice, once during the shorten and again for testing.

kivy/core/text/markup.py
```diff
@@ -620,12 +638,19 @@ def p_restricted(line, uw, c):
                 if word.lw:
                     line.append(word)
 
+        # if nothing to shorten
+        if uw >= sum([word.lw for word in line]):
```
matham: The same exact computation is done just two lines below, why do it twice?

kivy/core/text/markup.py
```diff
@@ -775,5 +816,10 @@ def p_restricted(line, uw, c):
         lw = sum([word.lw for word in line1])
         lh = max([word.lh for word in line1]) * line_height
         self.options = old_opts
-        return lw + 2 * xpad, lh + 2 * ypad, [LayoutLine(0, 0,
-            lw, lh, 1, 0, line1)]
+        if uw < sum([word.lw for word in line]):
```
matham: Again, this computation is already done. Why do it twice? Also, shouldn't it be line1?

KeyWeeUsr: Ah, yes, I missed the `1`. It's done multiple times, because there are multiple `return` statements in the method. Maybe I should move it right before the returns and use a single variable + check if it's computed. Could be better this way?

kivy/core/text/__init__.py
KeyWeeUsr: I think the reason behind this was to compare flattened and shortened text, but the already mentioned multiple returns would make it longer to write (for each return?) in the markup itself and maybe not even faster. Not sure how to handle this.

dessant:
With some other camera providers removed, this should be a release blocker.

kivy/core/text/markup.py
KeyWeeUsr: Ah, right, I'm blind.

KeyWeeUsr:
Still not sure how to handle the lines in `__init__`, any ideas?

Issue #3237, opened 2015-04-05
TextInput cursor shown outside of widget bounds


dessant:
``` python
from kivy.app import App
from kivy.lang import Builder

kv = """
AnchorLayout:
    TextInput:
        size_hint: None, None
        size: 300, 200
        text: "FILLER\\nFILLER\\nFILLER\\nFILLER\\nFILLER"
        font_size: 50
        focus: True
"""


class TestApp(App):
    def build(self):
        return Builder.load_string(kv)

if __name__ == '__main__':
    TestApp().run()
```
##

dessant:
Closed by https://github.com/kivy/kivy/pull/4905.

PR #4859, opened 2016-12-30
textinput cursor going outside bound problem solved

saqib1707:
Well the problem was because when the application is  getting started and text in the textinput exceeds the height of the textinput (leading to the cursor going out of bounds) , then the scroll_x as well as scroll_y are not updated. Removing that check (self._cursor == cursor) doesn't breaks anything except that it leads to execution of below code https://github.com/kivy/kivy/blob/master/kivy/uix/textinput.py#L2542-L2572 when scroll_x and scroll_y haven't changed.

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -2536,8 +2536,6 @@ class TextInput(FocusBehavior, Widget):
         cr = boundary(pos[1], 0, len(l) - 1)
         cc = boundary(pos[0], 0, len(l[cr]))
         cursor = cc, cr
-        if self._cursor == cursor:
-            return
 
         self._cursor = cursor
 


saqib1707:
#3237

akshayaurora:
@saqib1707 could you please post a sample script that reproduces this problem,?

saqib1707:
Sample script is here -- #3237

phunsukwangdu:
Removing check (self._cursor == cursor) will not be a good option as just below this method cursor is defined as alias property.The alias property has getter and setter function.Removing that check statement will lead to problems for cursor.
#4905 This pr overcome this situation and rules out the bug.

saqib1707:
@phunsukwangdu  if self._cursor != cursor then anyway the cursor has to be updated  and both our changes are doing the same  but  when self._cursor == cursor then the only difference between your change and mine is the execution of one more extra line self._cursor = cursor.
Basically I don't know which change will be relatively faster but can u explain or (much better give a testcase) where it will lead to problems for the cursor. ?
Thanks.

phunsukwangdu:
@saqib1707 Yes both cases will not the effect the cursor but the pr #4905 has edge over this solution in execution time. :) I have tested your pr too and is working fine. :)

saqib1707:
@phunsukwangdu Well since you have gone too deep in the discussion of execution time of an assignment statement versus that of a comparison statement the timings are roughly the same.Simply u cannot distinguish.I checked it out and it is around (on average) 1 us for both statements individually.
Another thing when user opens up a textinput and he is typing text then in your case it is executing two statements comparison followed by assignment whereas in my case it is only assignment (which is most often happening since user has to type text).
Well let the members decide what to do with these pr's and not waste our time in these discussions.

phunsukwangdu:
@saqib1707 If you carefully see your arguments you have proved my point.In your pr AliasProperty(cursor) setter function will always be true (ignoring first comparison) .
I believe in making Kivy a better place and always ready to burn midnight oil for it.I do appreciate your work and yes i will surely abide by the members.

Zen-CODE:
@phunsukwangdu. I think the merge of https://github.com/kivy/kivy/commit/b36ddd52e3f231c30212a8242be3f31663713dda fixed this issue? Can we close this?

phunsukwangdu:
yes sure,credits goes to @saqib1707  :) for his works.

Zen-CODE:
Thanks both :-)

PR #4884, opened 2017-01-04
setting minimum_height to height in clipboard.py

bhaveshAn:
examples/miscellaneous/clipboard.py started working after this correction

### DIFF
--- a/examples/miscellaneous/clipboard.py
+++ b/examples/miscellaneous/clipboard.py
@@ -11,7 +11,7 @@ Builder.load_string('''
     orientation: 'vertical'
     BoxLayout:
         size_hint_y: None
-        height: self.minimum_height
+        height: self.height
         Button:
             text: 'Paste raw clipboard'
             size_hint_y: None


KeyWeeUsr:
In 1.9.1+ this example works correctly. Any particular example that doesn't work would be nice.

Btw, what OS and Kivy version do you have? What about providers?

bhaveshAn:
Working on OS Windows 10 and kivy 1.9.1
Clipboard provider winctypes

KeyWeeUsr:
Apparently there was no `minimum_height` in `BoxLayout` yet. I wonder how it worked before...

Sorry for misunderstanding and thanks for noticing! 👍

bhaveshAn:
Np @KeyWeeUsr

examples/miscellaneous/clipboard.py
```diff
@@ -11,7 +11,7 @@
     orientation: 'vertical'
     BoxLayout:
         size_hint_y: None
-        height: self.minimum_height
+        height: self.height
```
dessant: The property references itself and the value will just stay `100`. If compatibility with older versions is needed, GridLayout could be used.

Issue #4867, opened 2017-01-02
Possible bug in actionview at maximize/minimize window


aerogelio:
I am adding and deleting ActionButtons to an ActionView  programmatically using add_widget and remove_widget functions, but at some point when I maximize or minimize the window, all ActionButtons removed are added again. If I do not maximize or minimize the window all work fine.

I was checking the ActionView code and found that it overwrites add_widget for stores the views on property _list_action_items but I did not find where to remove them from _list_action_items.

It appears that when removing a child with remove_widget( child ) and also with _list_action_item.remove( child ) the problem is solved, but I'm not sure and I do not like to get involved with the property directly.

KeyWeeUsr:
I think the widget worked normally before. A reproducible example would be really nice btw :)

aerogelio:
Here:
https://github.com/aerogelio/avposbug

I try:
1.- Add five widgets
2.- Remove five widgets
3.- Add five widgets
4.- Remove five widgets
5.- Maximize
6.- Minimize
7.- The effect should be visible.

If anything else happens tell me, for upload images and videos.

KeyWeeUsr:
It's somehow bound to the width of widget/window, probably triggered when the amount of visible widgets is recalculated. It's probably only the list issue.

PR #4891, opened 2017-01-05
Actionview window maximize/minimize bug removed

saqib1707:
Well the bug was because the removed or cleared items are not removed from _list_action_items resulting in addition of removed/cleared items on maximizing/minimizing window.Adding a function remove_widget in ActionView class and removing the items from the above mentioned list solves the bug.
Fixes - #4867

### DIFF
--- a/kivy/uix/actionbar.py
+++ b/kivy/uix/actionbar.py
@@ -486,24 +486,50 @@ class ActionView(BoxLayout):
             group.use_separator = value
         self.overflow_group.use_separator = value
 
+    def remove_widget(self, widget):
+        super(ActionView, self).remove_widget(widget)
+        if isinstance(widget, ActionOverflow):
+            for item in widget.list_action_item:
+                self._list_action_items.remove(item)
+
+        if widget in self._list_action_items:
+            self._list_action_items.remove(widget)
+    """
+    def clear_widgets(self, children=None):
+        '''
+        if not children:
+            children = self.children
+        remove_widget = self.remove_widget
+        for child in children[:]:
+            remove_widget(child)
+        '''
+        super(ActionView, self).clear_widgets()
+        self._list_action_items = []
+    """
     def _clear_all(self):
+        lst = []
+        for item in self._list_action_items:
+            lst.append(item)
         self.clear_widgets()
         for group in self._list_action_group:
             group.clear_widgets()
 
         self.overflow_group.clear_widgets()
         self.overflow_group.list_action_item = []
+        self._list_action_items = lst
 
     def _layout_all(self):
         # all the items can fit to the view, so expand everything
         super_add = super(ActionView, self).add_widget
         self._state = 'all'
         self._clear_all()
-        super_add(self.action_previous)
+        if self.action_previous.parent is None:
+            super_add(self.action_previous)
         if len(self._list_action_items) > 1:
-            for child in self._list_action_items[1:]:
-                child.inside_group = False
-                super_add(child)
+            for child in self._list_action_items[:]:
+                if not isinstance(child, ActionPrevious):
+                    child.inside_group = False
+                    super_add(child)
 
         for group in self._list_action_group:
             if group.mode == 'spinner':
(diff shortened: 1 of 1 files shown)


kivy/uix/actionbar.py
```diff
@@ -486,24 +486,39 @@ def on_use_separator(self, instance, value):
...
         self._state = 'all'
         self._clear_all()
-        super_add(self.action_previous)
+        if self.action_previous.parent is None:
+            super_add(self.action_previous)
         if len(self._list_action_items) > 1:
-            for child in self._list_action_items[1:]:
-                child.inside_group = False
-                super_add(child)
+            for child in self._list_action_items[:]:
```
KeyWeeUsr: Why the copy?

kivy/uix/actionbar.py
```diff
@@ -486,24 +486,39 @@ def on_use_separator(self, instance, value):
...
+        lst = []
+        for item in self._list_action_items:
+            lst.append(item)
         self.clear_widgets()
         for group in self._list_action_group:
             group.clear_widgets()
 
         self.overflow_group.clear_widgets()
         self.overflow_group.list_action_item = []
+        self._list_action_items = lst
```
KeyWeeUsr: I'm confused. What does it actually do? In the beginning of the method you append all items in `self._list_action_items` to that list, then clear all widgets in groups from the list and then return the same list? Isn't the list unchanged during the whole method?

kivy/uix/actionbar.py
```diff
@@ -486,24 +486,39 @@ def on_use_separator(self, instance, value):
...
 
     def _layout_all(self):
         # all the items can fit to the view, so expand everything
         super_add = super(ActionView, self).add_widget
         self._state = 'all'
         self._clear_all()
-        super_add(self.action_previous)
+        if self.action_previous.parent is None:
```
KeyWeeUsr: `if not self.action_previous.parent:`

kivy/uix/actionbar.py
```diff
@@ -523,7 +538,8 @@ def _layout_group(self):
         super_add = super(ActionView, self).add_widget
         self._state = 'group'
         self._clear_all()
-        super_add(self.action_previous)
+        if self.action_previous.parent is None:
```
KeyWeeUsr: `if not ...`

kivy/uix/actionbar.py
```diff
@@ -544,7 +560,8 @@ def _layout_random(self):
         hidden_items = []
         hidden_groups = []
         total_width = 0
-        super_add(self.action_previous)
+        if self.action_previous.parent is None:
```
KeyWeeUsr: `if not...`

kivy/uix/actionbar.py
```diff
@@ -600,7 +616,8 @@ def _layout_random(self):
                 over_add(child)
 
             overflow_group.show_group()
-            super_add(overflow_group)
+            if self.overflow_group.parent is None:
```
KeyWeeUsr: `if not...`

kivy/uix/actionbar.py
saqib1707: Yeah this is not required anymore. I added this because while making  changes the list items were disarranged a little..I just checked right now it works fine without this change.

kivy/uix/actionbar.py
saqib1707: @KeyWeeUsr the function _clear_all is called from the three functions of layout when it has to layout the items in the resized window.Due to the clear_widget function all the items in the _list_action_items are removed resulting in nothing added to the list. (Just remove the line self._list_action_item = lst  and then run the code https://github.com/aerogelio/avposbug -->  add around 10 items using add button --> then maximize and minimize the window --> result is all items getting removed because _list_action_items was emptied by clear widgets ).

KeyWeeUsr: Makes sense now. `lst = self._list_action_items[:]` would be shorter & [faster](http://stackoverflow.com/a/14821420/5994041) though. :)

KeyWeeUsr:
Should be ok now. Thanks 👍

PR #4890, opened 2017-01-05
fix pep8 errors

bhaveshAn:
Fixed pep8 errors wherever found necessary.

PR #4892, opened 2017-01-05
Version statements removed in examples

bhaveshAn:
@KeyWeeUsr 
[Version tags suggestion](https://github.com/kivy/kivy/pull/4890)
Removed version statements from examples as for now. Other changes may be good according to pep8 guidelines but can harm the code base, so its best to ignore other changes for now.

### DIFF
--- a/doc/autobuild.py
+++ b/doc/autobuild.py
@@ -6,17 +6,6 @@ Be careful if you change anything in !
 
 '''
 
-ignore_list = (
-    'kivy._clock',
-    'kivy._event',
-    'kivy.factory_registers',
-    'kivy.graphics.buffer',
-    'kivy.graphics.vbo',
-    'kivy.graphics.vertex',
-    'kivy.uix.recycleview.__init__',
-    'kivy.setupconfig'
-)
-
 import os
 import sys
 from glob import glob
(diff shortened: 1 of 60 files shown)


KeyWeeUsr:
First, when making another branch, switch from current one to **master**, where the repo is in an untouched state (without your edits), otherwise you will drag all your previous commits from each branch you do `git checkout -b <branch>` in. See the "fix pep8 errors" commit.

Then the amount of files is kind of horrible if you think about what changes you do. It wasn't _one_ change in multiple files, thus more checking if you didn't break anything for us (Not a happy solution). We could probably hammer it to more appropriate changes if you read the comments with requested changes properly.

And of course the same PR is already opened with changes **_requested_**, which means **the PR isn't closed** or anything, **it just highlights** files that need a change. The big `X` in a circle might seem like a "bad sign". It's not and it's completely ok to fix stuff with another commits in the already opened PR (`git checkout pep8`).

Just a note, pep8 isn't really a happy way how to start contributing in large projects, because of the way some projects work e.g. in kivy some imports on the top are required. Then there's docstring for docs generation and many many more little things which if you change everything goes → 💣 

PEP8 is just a guideline / recommendation and it even mentiones [ignore cases](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds). :)

Closing in favor of #4890

bhaveshAn:
@KeyWeeUsr 
Thanks for your note.

PR #4895, opened 2017-01-07
VideoGstplayer buffering and update check interval

ignus2:
As the commit message sais


dessant:
@ignus2, would it make sense to get the video framerate as reported by the provider and set the texture update interval according to that value?

What is the purpose of the buffer change in the gst provider?

ignus2:
I think checking every frame is better, as in that case you don't rely on some reported value, moreover the timer is not that accurate. Especially if you have vsync locked at 30fps, and video is also 30 fps, it can cause jitters if not checked every frame.
The buffering change is to try to mitigate jitter (vsync 30 and 30fps video case).

PR #4896, opened 2017-01-07
virtualenv install from source on Windows

ignus2:
This makes it possible to install/compile kivy from source inside a virtualenv on Windows.
Before compilation however, 2 python lib files must be moved out of the way in the global Python installation, otherwise MinGW will pick them up before it's own versions (unfortunately) and fail during linking, as those are compiled for MSVC.

So rename the files:
c:\Python34\libs\python3.lib 
c:\Python34\libs\python34.lib 
to
c:\Python34\libs\__python3.lib 
c:\Python34\libs\__python34.lib 
otherwise it will conflict during compilation with mingwpy's libs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ import os
 from os.path import join, dirname, sep, exists, basename, isdir
 from os import walk, environ
 from distutils.version import LooseVersion
+from distutils.sysconfig import get_python_inc
 from collections import OrderedDict
 from time import sleep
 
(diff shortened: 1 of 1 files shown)


matham:
About the renaming of files, if someone else can confirm and if there's no code fix it should be added to the docs. When this problem occurs, does it occur if you install mingpy both globally and into the venv, or only when it's installed to one of them?

ignus2:
I believe it only affects full-virtualenv installation, meaning mingwpy also in venv. I like to do everything in venvs always, so the base python installation never gets polluted, so I only use it this way.

matham:
Looking at the folder, mingwpy just adds a libpythonXX.dll.a in `/libs` which seems to fix to make it not use the .lib file. So maybe just try copying the .a file into that folder as opposed to removing the .lib file?

ignus2:
You could, I don't prefer, as that way the global python installation is polluted with a certain version of mingwpy, which can mess up other venvs. Renaming the file is different, and if you never use MSVC, it's not needed.

PR #4897, opened 2017-01-07
Focus fix #4816

### LINKED ISSUE #4816: Question: Focus Behavior
**Question**: Should this method be made public? `_get_focus_next`
https://github.com/kivy/kivy/blob/master/kivy/uix/behaviors/focus.py#L463

Reason:
User could navigate using events like `on_release` or `on_press` via the clicks/touch to change the focus of the different widgets. Sort of like navigating or user could also use this in pagination.

Use this method like this: `instances.get_focus_next('focus_next')` | `instances.get_focus_next('focus_previous')`

**Example to show how it could be used.**
screen shot 2016-12-17 at 12 37 06 pm

Editing the focus_behavior.py in the examples. https://github.com/kivy/kivy/blob/master/examples/widgets/focus_behavior.py

    from kivy.app import App
    from kivy.uix.behaviors import FocusBehavior
    from kivy.graphics import Color, Rectangle

    class FocusWithColor(FocusBehavior):
        ''' Class that when focused, changes its background color to red.
        '''

        _color = None
        _rect = None

        def __init__(self, **kwargs):
            super(FocusWithColor, self).__init__(**kwargs)
            with self.canvas:
                self._color = Color(1, 1, 1, .2)
                self._rect = Rectangle(size=self.size, pos=self.pos)
                self.bind(size=self._update_rect, pos=self._update_rect)

        def _update_rect(self, instance, value):
            self._rect.pos = instance.pos
            self._rect.size = instance.size

        def on_focused(self, instance, value, *largs):
            self._color.rgba = [1, 0, 0, .2] if value else [1, 1, 1, .2]

    class FocusButton(FocusWithColor, Button):
        pass

    class FocusLabel(FocusWithColor, Button):

        def on_release(self):
            FocusApp.count = FocusApp.instances.index(self)

    class FocusApp(App):

        instances=[]
        count = 0

        def go_right(self, *args):
            temp = self.instances[self.__class__.count]._get_focus_next('focus_next')
            temp.focus = True
            self.__class__.count += 1
            if self.__class__.count >= len(self.instances):
                self.__class__.count = 0

        def go_left(self, *args):
            temp = self.instances[self.__class__.count]._get_focus_next('focus_previous')
            temp.focus = True
            self.__class__.count -= 1
            if self.__class__.count < 0:
                self.__class__.count = len(self.instances)-1

        def build(self):
            root = BoxLayout(padding=[10, 10], spacing=5)
            root.add_widget(FocusButton(text='Go->.', size_hint_x=0.4,
                                       on_release=self.go_right))
            for i in range(5):
                temp = FocusLabel(text=str(i))
                self.instances.append(temp)
                root.add_widget(temp)

            for i, instance in enumerate(self.instances):
                if i == len(self.instances)-1:
                    instance.focus_next = self.instances[0]
                else:
                    instance.focus_next = self.instances[i+1]

            root.add_widget(FocusButton(text='<-Go', size_hint_x=0.4,
                                       on_release=self.go_left))

            return root

    if __name__ == '__main__':
        FocusApp().run()

phunsukwangdu:
#4816 implementing the discussed pr. Plz correct me if i am wrng somewhere.
note:- Not in merge able state. Further works and tests to be done.

### DIFF
--- a/kivy/uix/behaviors/focus.py
+++ b/kivy/uix/behaviors/focus.py
@@ -488,18 +488,16 @@ class FocusBehavior(object):
                 return None
 
     def get_focus_next(self):
-        '''This method provides the next focuses(ie self.focus_next)
-        and ponders over two cases if one hits focusable we walk through
-        focus_next.If one hits unfocusable we walk the widget tree
-        keeping various cases in mind in both of the cases.
+        '''Returns the next focusable widget using either :attr:`focus_next`
+           or the :attr:`children` similar to the order when tabbing forwards
+           with the ``tab`` key.
         '''
         return self._get_focus_next('focus_next')
 
     def get_focus_previous(self):
-        '''This method provides the previous focuses(ie self.focus_previous)
-        and ponders over two cases if one hits focusable we walk through
-        focus_previous.If one hits unfocusable we walk the widget tree
-        keeping various cases in mind in both of the cases.
+        '''Returns the previous focusable widget using either
+           :attr:`focus_previous` or the :attr:`children` similar to the
+           order when ``tab`` + ``shift`` key are triggered together.
         '''
         return self._get_focus_next('focus_previous')
 
--- a/kivy/uix/behaviors/focus.py
+++ b/kivy/uix/behaviors/focus.py
@@ -459,6 +459,11 @@ class FocusBehavior(object):
             focusable.focus = False
 
     def get_focus_next(self):
+        '''This method provides the next focuses(ie self.focus_next)
+        and ponders over two cases if one hits focusable we walk through
+        focus_next.If one hits unfocusable we walk the widget tree
+        keeping various cases in mind in both of the cases.
+        '''
         current = self
         walk_tree = 'walk'
         focus_dir = 'focus_next'
@@ -488,6 +493,11 @@ class FocusBehavior(object):
                 return None
 
     def get_focus_previous(self):
+        '''This method provides the previous focuses(ie self.focus_previous)
+        and ponders over two cases if one hits focusable we walk through
+        focus_previous.If one hits unfocusable we walk the widget tree
+        keeping various cases in mind in both of the cases.
+        '''
         current = self
         walk_tree = 'walk_reverse'
         focus_dir = 'focus_previous'
--- a/kivy/uix/behaviors/focus.py
+++ b/kivy/uix/behaviors/focus.py
@@ -1,19 +1,23 @@
 '''
 Focus Behavior
 ==============
+
 The :class:`~kivy.uix.behaviors.FocusBehavior`
 `mixin <https://en.wikipedia.org/wiki/Mixin>`_ class provides
 keyboard focus behavior. When combined with other
 FocusBehavior widgets it allows one to cycle focus among them by pressing
 tab. In addition, upon gaining focus, the instance will automatically
 receive keyboard input.
+
 Focus, very different from selection, is intimately tied with the keyboard;
 each keyboard can focus on zero or one widgets, and each widget can only
 have the focus of one keyboard. However, multiple keyboards can focus
 simultaneously on different widgets. When escape is hit, the widget having
 the focus of that keyboard will de-focus.
+
 Managing focus
 --------------
+
 In essence, focus is implemented as a doubly linked list, where each
 node holds a (weak) reference to the instance before it and after it,
 as visualized when cycling through the nodes using tab (forward) or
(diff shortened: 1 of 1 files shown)


kiok46:
@phunsukwangdu adding to @KeyWeeUsr 's comment on #4816  I think you should change the name of the methods from `go_left` and `go_right` to maybe `focus_next` and `focus_previous` or even `next` and `previous`  since this is focus behaviour so, go_left and go_right might be misleading.

kivy/uix/behaviors/focus.py
```diff
@@ -489,6 +489,12 @@ def _get_focus_next(self, focus_dir):
             else:
                 return None
 
+    def go_left(self,*args):
```
kiok46: please, don't forget pep8. :)

matham: Yes, please read https://kivy.org/docs/contribute.html. Also, the function doesn't return anything.

matham:
`focus_next`/`focus_previous` are already property names. `next`/`previous` is too vague and will likely clash with other classes. I think `get_focus_next`/`get_focus_previous` is probably best.

kiok46:
@matham  +1

phunsukwangdu:
@kiok46 thx for review @matham thx for anticipation.I will surely browse the link once again.I have many copies of repo on laptop.
In one of them i had done made hook but was facing some issues.That is why i restrained my self to use pep8 script from pip in my upcoming commits therefore i mentioned that pr was not in merge able state.

phunsukwangdu:
I have tested the pr and is in mergeable state. :)

kivy/uix/behaviors/focus.py
```diff
@@ -460,10 +458,39 @@ def _handle_post_on_touch_up(touch):
                 continue
             focusable.focus = False
 
-    def _get_focus_next(self, focus_dir):
+    def get_focus_next(self):
```
KeyWeeUsr: documentation is kind of important when making methods visible to people

kivy/uix/behaviors/focus.py
```diff
@@ -460,10 +458,39 @@ def _handle_post_on_touch_up(touch):
...
+            if isinstance(current, FocusBehavior):
+                if current is self:
+                    return None
+                if current.is_focusable and not current.disabled:
+                    return current
+            else:
+                return None
 
+    def get_focus_previous(self):
+        current = self
```
KeyWeeUsr: same here

KeyWeeUsr:
Otherwise, it seems nice :)

phunsukwangdu:
@KeyWeeUsr sure i will work them out in my next commit. :) I feel delighted for being guided and mentored. :)

kivy/uix/behaviors/focus.py
```diff
@@ -460,10 +458,49 @@ def _handle_post_on_touch_up(touch):
...
+        '''This method provides the next focuses(ie self.focus_next)
+        and ponders over two cases if one hits focusable we walk through
+        focus_next.If one hits unfocusable we walk the widget tree
+        keeping various cases in mind in both of the cases.
+        '''
         current = self
-        walk_tree = 'walk' if focus_dir is 'focus_next' else 'walk_reverse'
+        walk_tree = 'walk'
+        focus_dir = 'focus_next'
+        while 1:
```
dessant: Putting this code block in an internal method would be good, there is no reason to duplicate it. https://github.com/kivy/kivy/pull/4897/files#diff-0b541aab11285191c153cab88b7ea002R504

phunsukwangdu: @dessant I will do in that way :) .
The problem arised was naming the methods get_focus_next and get_focus_previous and making previous method public.I found this way much better.I have tested it ,worked out example on it.It is working fine.

matham: Code should not be duplicated because it leads to maintenance issues. The way it was before, except with an added return and documentation is how it should be. That required no code duplication.

phunsukwangdu: @matham  sure i see the point and appreciate it.I have made commits and done it in the previous way.:)

kivy/uix/behaviors/focus.py
```diff
@@ -489,6 +487,22 @@ def _get_focus_next(self, focus_dir):
             else:
                 return None
 
+    def get_focus_next(self):
+        '''This method provides the next focuses(ie self.focus_next)
```
matham: There's a general style of the kivy docs that we should try to stick to. Mainly, it should be more succinct and use internal linking. There are also some grammar issues. E.g. a better version would be

```
Returns the next focusable widget using either :attr:`focus_next` or the :attr:`children` similar to the order when tabbing forwards with the ``tab`` key.
```

phunsukwangdu: @mantham sure, i didn't pick up on that , my mistake.I will surely act on them and get them rectified.

matham:
👍

Issue #2172, opened 2014-05-08
Checkbox radio occasionally shows 2 active.


cropleyb:
This is intermittent on my system.
##

PR #4898, opened 2017-01-07
Multiple Radiobutton selected simultaneously.

saqib1707:
Well there were two bugs basically - first when a number of checkboxes are initiated under a group with active= True, all become active and secondly trying to control the grouped checkboxes via code (for example on_press of a button) results in multiple checkboxes being active as mentioned here #2172 .

I have used a new check_variable so that the checkbox multi-active bug is removed. Before considering this PR for merge one thing that is creating problem are these two test samples .
1. https://gist.github.com/saqib1707/78c38909988bd729d26826ab814056be
2. https://gist.github.com/saqib1707/c4381e815e0e37be36e8dc2193271e6a

In first case , of the two buttons which are made active , the first one(kv lang) (instead of second) remains in the 'down' state. while in the second case(usual python code)  the last checkbox is active.I am not getting where the difference is coming.Can anyone help ?

Fixes : #2172

### DIFF
--- a/kivy/uix/behaviors/togglebutton.py
+++ b/kivy/uix/behaviors/togglebutton.py
@@ -78,9 +78,10 @@ class ToggleButtonBehavior(ButtonBehavior):
     :attr:`allow_no_selection` is a :class:`BooleanProperty` and defaults to
     `True`
     '''
-    check_variable = BooleanProperty(False)
+
     def __init__(self, **kwargs):
         self._previous_group = None
+        self.check_variable = BooleanProperty(False)
         super(ToggleButtonBehavior, self).__init__(**kwargs)
 
     def on_group(self, *largs):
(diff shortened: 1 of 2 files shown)


udiboy1209:
I am testing this with the example @dessant posted in a [comment on 2172](https://github.com/kivy/kivy/issues/2172#issuecomment-59456288).

For the top radio group, the bug exists initially for one try, but then goes away i.e. Clicking the button for the first time makes both radio buttons active. It goes away once one of the radio buttons is clicked on. 

The second radio group has both radio buttons active initially, which also needs to be fixed acc. to the issue. 

I will check your test cases after dinner.

saqib1707:
@udiboy Actually it was working perfectly previously . I might have made some mistakes in the end. Let me add a commit to fix this.

dessant:
@udiboy1209, is the current state ok?

udiboy1209:
@dessant yes both your and saqi1707's examples are working okay. This looks good to merge.

PR #4900, opened 2017-01-08
Fix#4838: font_size is used for keyboard text size.

### LINKED ISSUE #4838: A way to set the VKeyboard font size could be added
I'm not sure how to go about this small but probably useful change. Currently, no clean way seems to exist for setting the font size of the keys in a VKeyboard. Indeed, the `draw_keys` method of the `VKeyboard` class has the following code:

    w, h = self.size
    # ...
    font_size = int(w) / 46

But the rendered keyboard is not pleasing to the eye as rendered on half of the screen of a 7'' touchscreen running on a Raspberry Pi. I changed the denominator from 46 to 32 and the keys look much better. So my question is, how could this customization be cleaner?

saqib1707:
Fix #4838 As suggested by @KeyWeeUsr , font_size(earlier exposed but was not used ) is used as vkeyboard text size without any dependence on size of window.

### DIFF
--- a/kivy/uix/vkeyboard.py
+++ b/kivy/uix/vkeyboard.py
@@ -229,6 +229,8 @@ class VKeyboard(Scatter):
     It should be kept within a limit otherwise text either goes outside
     the bounds of the key or becomes too small to read.
 
+    .. versionadded:: 1.9.2
+
     :attr:`font_size` is a :class:`~kivy.properties.NumericProperty` and
     defaults to 20.
     '''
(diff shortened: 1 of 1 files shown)


kivy/uix/vkeyboard.py
```diff
@@ -224,6 +224,15 @@ class VKeyboard(Scatter):
     to [2, 2, 2, 2]
     '''
 
+    font_size = NumericProperty(20.)
+    '''font_size, specifies the size of the text of virtual keyboard.
+    It should be kept within a limit otherwise text either goes outside
+    the bounds of the key or becomes too small to read.
+
+    :attr:`font_size` is a :class:`~kivy.properties.NumericProperty` and
+    defaults to 20.
```
Zen-CODE: Can we add here:

.. versionadded:: 1.9.2

Zen-CODE:
Thanks :-)

PR #4901, opened 2017-01-08
focus till StopIteration

bhaveshAn:
[#4816](https://github.com/kivy/kivy/issues/4816)
I think the focus has the significance till it stops iterating between widgets on pressing tab and shift+tab.
The  _get_focus_next method returning none for *current is self* which is not good according to me. So the method should return none only at *current is StopIteration*.
If this is true than this method will not be required to make public as well.
Please tell if any other changes needed as well.


matham:
The code is correct the way it currently is. Focus cycling should stop either when we hit the widget again, because otherwise we'll keep walking focus_next/previous in an infinite loop or if it's `StopIteration`, as documented.

I also don't understand how this will affect having to make the method public. That should be required either way.

PR #4905, opened 2017-01-09
Removed textinput cursor bug #3237

phunsukwangdu:
The bug was that TextInput cursor was showing outside of widget bounds.On reproducing the result with the following testcode i found that scroll view was not getting updated that is why cursor was not visible in viewport.New lines (\n) posed the problems here.
I even experimented with /t string literals and it was working fine.Below testcode signifies it.I would be happy to attach screenshots if required. #3237
[Testcode 1](https://gist.github.com/phunsukwangdu/b7810dedbd02a2ad8a3092da05836af1)
[Testcode 2](https://gist.github.com/phunsukwangdu/3c7e69e121bbf2d45ae914bbaa2863ee)


Zen-CODE:
Nice one. Thanks :-)

Issue #4908, opened 2017-01-10
Kivy.utils intersection Bug.


phunsukwangdu:
There is a interesting bug in Kivy.utils  intersection part.
In python interpreter 
>>> x=[1,2,3,4,5]
>>> y=[1,2,3,1,2]
>>> from kivy.utils import intersection
>>> intersection(x,y)
[1, 2, 3]
>>> intersection(y,x)
[1, 2, 3, 1, 2]

But we know intersection(x,y)==intersection(y,x) but same is not the case above.

PR #4909, opened 2017-01-10
Removed intersection bug and added more functionality

phunsukwangdu:
Removed intersection bug and union & symmetric difference were also added. #4908


dessant:
As you see the tests fail most likely because `set` is an unordered data structure. New features are usually added when the need arises for them, I do not see a need for including union and simmetric difference in `kivy.utils` at the moment.

Issue #4786, opened 2016-12-12
Feature Request: Range Slider Widget


crashMOGWAI:
Suggest extending the basic slider and adding a range slider. I think it is common enough to add a built in. 
Example: https://refreshless.com/nouislider/

saqib1707:
@crashMOGWAI  Apart from adding a range-slider , feature of showing min_value, max_value and current_value can be added so that user don't have to explicitly add them and can have a choice whether to show these values in his app by setting (may be) a boolean variable.

Julian-O:
Based on comments on the PR, it seems this isn't warranted as being part of Kivy, but belongs in the garden as an add-on.

Closing as not planned.

PR #4911, opened 2017-01-12
Range slider feature extending the slider

saqib1707:
I have modified slider to add a range-slider feature which gives the user an option to set a range for a particular property instead of setting a single value to a property.Well I have the manipulated the rangeslider style feature inside the <Slider> Rule of style.kv. An example of rangeslider is [here](https://refreshless.com/nouislider/) . As mentioned here [#4786comment](https://github.com/kivy/kivy/issues/4786#issuecomment-269769998) , I am also working on adding an option for the user to  show values like max, min and the current value below the slider. So should this feature be there or something should be left for the user to manipulate??.. Suggestions are welcome.

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -59,52 +59,32 @@
             rgba: root.value_track_color if self.value_track and self.orientation == 'horizontal' else [1, 1, 1, 0]
         Line:
             width: self.value_track_width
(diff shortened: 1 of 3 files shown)


matham:
Sorry that it took so long to respond. Unfortunately, Kivy has diverged and this cannot be merged as is. Also, there are multiple things that would need to be changed before it can be merged, Please let us know if you are still interested in this.

For starters, this would need to be a completely different widget than slider because they don't completely fit together.

Also, this is probably a better fit for garden. If you're interested, please feel free to re-open this in garden.

Issue #4915, opened 2017-01-13
Compound selection nonstandard behavior


dessant:
Tested with [this](https://github.com/kivy/kivy/blob/master/examples/widgets/compound_selection.py) example. Steps to reproduce:

* click first item
* Shift+click fifth item
* Ctrl+click first, second and third item
* Shift+click fourth item

Kivy:
![kivy](https://cloud.githubusercontent.com/assets/7513068/21932997/1971fb56-d9ac-11e6-946c-d52446df4c30.gif)

Windows 10 file explorer:
![windows](https://cloud.githubusercontent.com/assets/7513068/21933010/29b2aaec-d9ac-11e6-8195-20a0ce9cc014.gif)

PCManFM:
![pcmanfm](https://cloud.githubusercontent.com/assets/7513068/21933015/36a6da20-d9ac-11e6-91b1-15eb4d14fedc.gif)

Nautilus:
![nautilus](https://cloud.githubusercontent.com/assets/7513068/21933073/7c522fd4-d9ac-11e6-88a8-b12fd225cff6.gif)

malverick:
I want to work on this issue. I will send the pr solving the issue within 2 days.

saqib1707:
@dessant since I had resolved a similar issue a few days back.I have fresh everything in mind and want to solve this.Can u please assign this to me otherwise there would be ambiguity?

dessant:
It's ok to signal on 1-2 issues that you are currently working on them, but what is happening on the issue tracker has a taste of territory marking. Please do not overdo the 'i intend to work on this issue' comments.

Since both of you are prospective gsoc students, here is a small competition for you guys:

@malverick, @saqib1707, to take part, you must submit a pr fixing this issue at 14:00:00 UTC, 14 January 2017 (+- 1 minute). No new commits are allowed after submitting the pr. If one of you submits it sooner or later, you both lost.

If both parties submit in time, you have 4 hours to test and review the other pr, if you find any issues with it, show a clear way for reproducing it. The review and potential issues should be posted in a **single** comment on the other student's pr thread.

The pr author has 2 additional hours to respond to the review in a **single** comment.

The competition closes 6 hours after the prs have been submitted.

If there are more posts authored by you on the pr threads than the number detailed above, you lost. If you edit the review or response comments, you lost.


We will declare a winner after this is done.

[...]

PR #4916, opened 2017-01-14
Compound selection behaviour issue fix. #4915

sumitmadhwani:
Fixes https://github.com/kivy/kivy/issues/4915.


malverick:
Sorry due to some network problems I was unable to provide the description of my commit.
But there was nothing in the rules about commenting after submitting the pr except responding the opposition's review in a single comment.
@dessant the issue was that after deselecting the node it was not resolving the last node. It should keep track of last selected or deselected node so that when `_select_range` gets called it should select from last node to current selected node.

saqib1707:
Well I was testing this pr but I found out that the bug is still there. 
1. Trying to reproduce as mentioned [here](https://github.com/kivy/kivy/issues/4915#issue-200642134) leaves two nodes selected.
2.  a) select a range of nodes using shift (0 to 9 in this [example](https://github.com/kivy/kivy/blob/master/examples/widgets/compound_selection.py) 
     b) Then with ctrl pressed deselect 8 and 9.
     c) then again shift pressed click on 20 (say). The anchor node changes to 8 but it should remain at 0.
There are more such small bugs may be because of changes done to deselect_node.
I think u missed out on the main reason that the main anchor node was not getting updated. but on the whole nice effort @malverick .

malverick:
@saqib1707 I request you that the steps which you have mentioned in your second point and the steps which @dessant mentioned [here](https://github.com/kivy/kivy/issues/4915), to execute on windows file explorer.
I think you have misunderstood dessant 's [issue](https://github.com/kivy/kivy/issues/4915). The kivy image is the bug and Windows 10 file explorer, PCManFM, Nautilus images are showing how kivy's compound selection behavior should work. Therefore two nodes are left selected.
Coming to your second point, these steps in kivy were not executing in the same way as they were in windows. So I additionally solved that issue. If you follow the same steps in my PR, you will see that compound selection behavior in kivy is working in same way as in windows file explorer.
Other small bugs which you are referring to are actually not the bugs if you see them according to windows platform.
And in my PR, anchor node is getting updated whenever any node is selected or deselected. 
I am taking windows as base platform because dessant issue was regarding the three platforms windows, PCManFM and Nautilus.

KeyWeeUsr:
link #4915
(ignore that, I'm just linking for more comfortable navigating through this)

Zen-CODE:
This PR breaks on the compound widget example with:
```
 "File "/usr/local/lib/python2.7/dist-packages/kivy/uix/behaviors/compoundselection.py", line 670, in select_node
     self._reslove_last_node()
 AttributeError: 'SelectableGrid' object has no attribute '_reslove_last_node'
```
It needs to be reworked and re-opened for consideration.

Thanks

kivy/uix/behaviors/compoundselection.py
```diff
@@ -662,6 +667,7 @@ def select_node(self, node):
             nodes.append(node)
         self._anchor = node
         self._last_selected_node = node
+        self._reslove_last_node()
```
Zen-CODE: Errors. Should be "self.resolve_last_node()"

kivy/uix/behaviors/compoundselection.py
```diff
@@ -681,6 +687,12 @@ def deselect_node(self, node):
             This method must be called by the derived widget using super if it
             is overwritten.
         '''
+        self._last_selected_node = None
+        self._reslove_last_node()
```
Zen-CODE: Errors. Should be "self.resolve_last_node()"

kivy/uix/behaviors/compoundselection.py
malverick: done

kivy/uix/behaviors/compoundselection.py
malverick: done

kivy/uix/behaviors/compoundselection.py
```diff
@@ -492,15 +492,20 @@ def _select_range(self, multiselect, keep_anchor, node, idx):
         last_idx = self._anchor_idx
 
         if last_node is None:
-            last_idx = end
-            last_node = sister_nodes[end]
+            self.clear_selection()
```
matham: Looking at how windows does it, it does not clear the current selection. I.e. try ctrl selecting 1, then 3 and then shift-ctrl select 5. It will not clear the selection of 1.

kivy/uix/behaviors/compoundselection.py
```diff
@@ -662,6 +667,7 @@ def select_node(self, node):
             nodes.append(node)
         self._anchor = node
         self._last_selected_node = node
+        self.reslove_last_node()
```
matham: I don't see `reslove_last_node` defined anywhere.

matham: Also, I'm not sure why you're calling this as it has no side effects?

kivy/uix/behaviors/compoundselection.py
```diff
@@ -681,6 +687,12 @@ def deselect_node(self, node):
             This method must be called by the derived widget using super if it
             is overwritten.
         '''
+        self._last_selected_node = None
+        self.reslove_last_node()
```
Zen-CODE: Still not right. It's "self._resolve_last_node()". This does not give me much faith that you are running and testing your changes?

PR #4921, opened 2017-01-14
Added an allow_screensaver property for Window

inclement:
Currently SDL2 automatically disables the screensaver, but has functions for enabling or disabling manually. Kivy doesn't do anything other than the default, so this affects all Kivy apps. It isn't very noticeable on the desktop, but probably isn't desirable as a default, and on mobile devices it prevents them from sleeping.

This PR adds a property to Window that exposes the SDL2 setting. As such, it will only work with the SDL2 provider, but since it's not the only one that doesn't seem too controversial. Other providers could potentially support it too.

I considered just disabling this behaviour on Android, but since it does affect desktop and iOS as well, and the SDL2 management of it is quite nice, it seemed better to resolve it with an option.

Resolves https://github.com/kivy/python-for-android/issues/969.

### DIFF
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -185,6 +185,8 @@ class WindowSDL(WindowBase):
         self.bind(minimum_width=self._set_minimum_size,
                   minimum_height=self._set_minimum_size)
 
+        self.bind(allow_screensaver=self._set_allow_screensaver)
+
     def _set_minimum_size(self, *args):
         minimum_width = self.minimum_width
         minimum_height = self.minimum_height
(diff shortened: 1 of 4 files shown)


akshayaurora:
👍

kivy/core/window/window_sdl2.py
```diff
@@ -195,6 +197,10 @@ def _set_minimum_size(self, *args):
                 'Both Window.minimum_width and Window.minimum_height must be '
                 'bigger than 0 for the size restriction to take effect.')
 
+    def _set_allow_screensaver(self, *args):
+        allow_screensaver = self.allow_screensaver
+        self._win.set_allow_screensaver(allow_screensaver)
```
dessant: Could these lines be merged?

PR #4927, opened 2017-01-17
Fix sdl2 exit on Android home button press

germn:
This PR is about to fix https://github.com/kivy/python-for-android/issues/978

If our app doesn't support sleep mode, behavior on home button press should be same as [on pressing back button](https://github.com/kivy/kivy/blob/master/kivy/core/window/__init__.py#L1400), while now [it's not](https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L608) (`self.close()` miss). This leads to crush.

While this PR fixes error, I think it can be improved:
1) Above the `on_close` handling [there's](https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L603) also condition when `app` is `None`. When can it happen? Should this block also contain `self.close()` call?
2) Can anyone say what exactly should be fixed [here](https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L613)?

### DIFF
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -210,6 +210,8 @@ class WindowSDL(WindowBase):
             if not app:
                 Logger.info('WindowSDL: No running App found, exit.')
                 stopTouchApp()
+                # self.close() is missing ?
+                # see discussion: https://github.com/kivy/kivy/pull/4927
                 return 0
 
             if not app.dispatch('on_pause'):
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -216,6 +216,7 @@ class WindowSDL(WindowBase):
                 Logger.info(
                     'WindowSDL: App doesn\'t support pause mode, stop.')
                 stopTouchApp()
+                self.close()
                 return 0
 
             self._pause_loop = True
@@ -608,6 +609,7 @@ class WindowSDL(WindowBase):
         if not app.dispatch('on_pause'):
             Logger.info('WindowSDL: App doesn\'t support pause mode, stop.')
             stopTouchApp()
+            self.close()
             return
 
         # XXX FIXME wait for sdl resume


KeyWeeUsr:
Will the app exit correctly with that fix, or will it hang like before https://github.com/kivy/python-for-android/pull/923 ?

germn:
@KeyWeeUsr kivy/python-for-android#923 fixes closing of the app, what happens usually on back button press. 

[This](https://github.com/kivy/kivy/pull/4927) PR have nothing to do we back button, it just makes home button (when `on_pause` returns `False`) to behave exactly like back button.

I tested PR with kivy/p4a masters (that includes fix you're talking about) and it fixes home button handling for me. With this PR app exit correctly with both close and home buttons.

KeyWeeUsr:
I wonder if it should be included even [here](https://github.com/germn/kivy/blob/d13dfbc67ee1fbe7cde910c49776933643ed1b83/kivy/core/window/window_sdl2.py#L607) as the `self.close()` is basically SDL2 window teardown, which means that if the app is paused correctly, but something happens to it, the same crash like with the `on_pause: return False` will be present too.

germn:
> I wonder if it should be included even here

I don't know and wonder this too. Any idea how can I reproduce this situation when `not app`?

KeyWeeUsr:
`App._running_app = None` might do the trick as it's a [class](https://github.com/germn/kivy/blob/d13dfbc67ee1fbe7cde910c49776933643ed1b83/kivy/app.py#L464) variable for example with this piece:

* trigger `on_pause` [automatically](https://github.com/germn/kivy/blob/d13dfbc67ee1fbe7cde910c49776933643ed1b83/kivy/core/window/window_sdl2.py#L508) with:

       Config.set('kivy', 'pause_on_minimize', 1)  #somewhere on top of the file

* remove App instance

        def on_pause(self):
            App._running_app = None

germn:
@KeyWeeUsr situation can't be tested way you proposed since we should set `_running_app` to `None` before `on_pause` dispatched (or, more precisely, before `do_pause` called).

I tried this test (with `self.close()` added to place we talk about):

```
def on_start(self):
    App._running_app = None
    self.root_window.do_pause()
```

In this case application crashes (while without `self.close()` not). I still not sure if it's final answer, so I added comment about situation in code.

I propose to merge PR since it fixes problem and people [ask](https://github.com/kivy/python-for-android/issues/978#issuecomment-301956472) about it.

PR #4937, opened 2017-01-19
Fix Window position from Config

KeyWeeUsr:
Before Window is actually created, fetch values from Config (`if not self.initialized`) and set it to a private property. That property is used in setter+getter, so it gets updated.

After the Window is initialized, just fetch values from SDL2.

For testing purposes see the moving Window and console output for each step (uncommenting). Should be fine now 🍡 

```
from kivy.clock import Clock
from functools import partial
# position: screen center

# Config.set('graphics', 'position', 'custom')
# position: 0, 0

# Config.set('graphics', 'top', 100)
# position: 0, 100

# Config.set('graphics', 'left', 100)
# position: 100, 100

# Window.left = 200
# position: 200, 100

# Window.top = 200
# position: 200, 200

def printf(*_):
    print(Window.left, Window.top)
Clock.schedule_interval(printf, 0.5)
def move(*_):
    Window.left += 10
    Window.top += 10
Clock.schedule_interval(move, 1)
runTouchApp()
```

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -687,14 +687,16 @@ class WindowBase(EventDispatcher):
         pass
 
     def _get_left(self):
-        return self._get_window_pos()[0]
+        left = self._get_window_pos()[0]
+        return left if left != -10000 else Config.getint('graphics', 'left')
 
     def _set_left(self, value):
         pos = self._get_window_pos()
         self._set_window_pos(value, pos[1])
 
     def _get_top(self):
-        return self._get_window_pos()[1]
+        top = self._get_window_pos()[1]
+        return top if top != -10000 else Config.getint('graphics', 'top')
 
     def _set_top(self, value):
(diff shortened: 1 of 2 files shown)


dessant:
For reviewers, the issue was introduced in https://github.com/kivy/kivy/commit/bae5bf23dcccf23bd3b5e65b6a01c74e9633bc90, and it can be reproduced with the following example:

```python
Config.set('graphics', 'position', 'custom')
# Config.set('graphics', 'left', 100)
# Config.set('graphics', 'top', 100)

from kivy.app import App


class TestApp(App):
    def build(self):
        return Button()


if __name__ == '__main__':
    TestApp().run()
```

dessant:
@KeyWeeUsr, if the refactor didn't change any of your intended behaviors, +1 for merging, it works nicely on my part.

KeyWeeUsr:
++, nice refactor 😋

Issue #4322, opened 2016-05-26
Don't set logging.root


CreamyCookie:
I just started using the logging module and I was wondering for quite a while why `logger.debug` didn't print anything. Luckily I found out that it's because kivy changes the root logger of the logging module. Please make `kivy.logger` an optional thing.

dessant:
Related discussion: https://github.com/kivy/kivy/issues/4113.

misl6:
Fixed via #7988

Issue #4933, opened 2017-01-18
When I create a new thread, cpu and memory usage suddenly rose to 100%.


falseen:
I tried to create a new thread in kivy to perform other tasks, I built a new button, and bind the start function, in this function used threading to start my program. But once the implementation of this thread, cpu and memory suddenly rose to 100%, until the crash. But when it is running alone, there is no problem. I do not understand the problem lies in which, please help me!

ps:  I tried to run the program is  `local` of `shadowsocks`
It's home page is : https://github.com/shadowsocks/shadowsocks/tree/master

dessant:
Based on what you described, it's unlikely you're experiencing a kivy bug, check out our support channels, and make sure you have prepared an example.

https://github.com/kivy/kivy#support

falseen:
I found the reason! In my process there is a code on the logging, the content is `logging.getLogger (''). Handlers = []`, each time running here, cpu and memory will suddenly rise to 100% until crash. As long as this line commented out, there is no problem. But kivy will be being given, the content is related with the logging, I think this should be kivy the logging module and the logging module of my program had a conflict.

My procedure directly on the set of logging, not derived from the new object, these settings may affect the normal operation of Kivy bars. I am looking for a solution ....

dessant:
@falseen, if that's indeed the case, it may be interesting. Could you share a runnable example?

falseen:
@dessant  Sure! 
It's on my github home page, you can download it here. [shadowsocks-kivy](https://github.com/falseen/shadowsocks-kivy)
I just started to create it, there are many places not perfect.

The program will start running by clicking the "start" button, and the  `logging.getLogger (''). Handlers = []` is located at line 301 o f `/shaowsocks-kivy/kivy/shadowsocks/shell.py`.

My environment：

> windows7 64bit
> python34 64bit
> kivy 1.9.1

dessant:
I was hoping for a short, runnable example, your repo also contains binary blobs, so I will not be able to run it, shadowsocks is a pip install away.

Anyway, maybe the way kivy overwrites the root logger conflicts with shadowsocks, if you boil it down to a simple example, please take a look at this issue and see if it applies: https://github.com/kivy/kivy/issues/4322.

falseen:
I seem to find the real reason.
 Run `logging.getLogger('').handlers = []`, then run `VERBOSE_LEVEL`, the error will appear.

A simple example ：
```python
from kivy.app import App
from kivy.lang import Builder
[...]

PR #4940, opened 2017-01-20
do not replace the root logger unless the KIVY_REPLACE_ROOT_LOGGER en…

dessant:
…v var is set

### DIFF
--- a/doc/sources/guide/environment.rst
+++ b/doc/sources/guide/environment.rst
@@ -71,6 +71,11 @@ KIVY_NO_FILELOG
 KIVY_NO_CONSOLELOG
     If set, logs will be not print to the console
 
+KIVY_REPLACE_ROOT_LOGGER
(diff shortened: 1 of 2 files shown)


dessant:
Closes https://github.com/kivy/kivy/issues/4322, and it causes an infinite recursion if the env var is not set. :|

```python
from os import environ
# environ['KIVY_REPLACE_ROOT_LOGGER'] = ''

import logging

from kivy.app import App


class TestApp(App):
    def build(self):
        return Button(on_press=lambda *x: logging.warning('test'))


if __name__ == '__main__':
    TestApp().run()
```

dessant:
This fixes the recursion: https://github.com/kivy/kivy/issues/4933#issuecomment-274269290, but the app exit may be logged twice as a result in certain cases, so the pr needs further work.

PR #4947, opened 2017-01-25
Create main.kv

cocodevienne:
This is the counterpart to a main.py I will be also propsing. Simple Display is a working App that I've made.


dessant:
Same as https://github.com/kivy/kivy/pull/4948#issuecomment-275220272.

PR #4948, opened 2017-01-25
Create main.py

cocodevienne:
The main.py of the simple display app! Please include this in the examples because when I started there was no clearly outlined way to make something happen on a button press using a function. This app shows the proper way beautifully! Credits for showing me this go to @inclement @tshirtman .

### DIFF
--- a/examples/simple display/main.py
+++ b/examples/simple display/main.py
@@ -1,0 +1,59 @@
+# AN APP BY THE ONE AND ONLY,
+# THE LEGEND,
+# COCO DE VIENNE.
+
+try:
+    from kivy.app import App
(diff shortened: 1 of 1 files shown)


dessant:
@cocodevienne, thanks for contributing, but we are not in need of app examples of this kind. The function you're describing could be presented in much simpler ways, and there are enough examples showcasing this part of the API, including the docs.

```python
from kivy.app import App


class TestApp(App):

    def callback(self, *args):
        print('Coco de Paris est plus magnifique!')

    def build(self):
        return Button(on_press=self.callback)


if __name__ == '__main__':
    TestApp().run()
```

PR #4950, opened 2017-01-26
Fix animation transition around the unit circle in Android compass example

Lauszus:
For example if the previous angle was 269 and the new angle is -84, then the current code would rotate the needle in a clockwise direction. This fix would add 360 to -84, so the new angle would be 276, which would result in the correct behaviour of the needle rotating anti-clockwise.


Lauszus:
@tshirtman no worries :)

Issue #4872, opened 2017-01-02
Broken / not implemented multiple joystick support?


KeyWeeUsr:
I'm not really sure what's hapenning here mainly because I get different `stickid` if I have multiple joysticks, yet the only input I get is from _the last one plugged in_ into my machine. I've tested with vJoy and one real device, yet either I get only the real device after vJoy + real, or I get only the vJoy input after read + vJoy.

I tried to find something more about it, yet the only thing I found was about [linux](http://stackoverflow.com/q/29589982/5994041) and I have Windows.

I tried to fix that according to the SDL documentation with an obvious fix (`SDL_NumJoysticks()` needs to be included in `kivy/lib/sdl2.pxi`):

    cdef int joy_i
    for joy_i in range(SDL_NumJoysticks()):
        SDL_JoystickOpen(joy_i)

but still nothing. Am I doing something wrong?

> with xinput xbox-like usb logitech pad Local multiplayer works great in boardz.

Any ideas @chozabu ?

ref #2564

FeralBytes:
@KeyWeeUsr I have a strong interest in getting controller support but for Steam Controllers; do you need help with the multiple controller issues? I just helped out Muppen64 with some controller debugging; so it is still fresh in my mind.

KeyWeeUsr:
Well, I know about these [mappings](https://github.com/gabomdq/SDL_GameControllerDB), so basically if SDL recieves the input from the controller, it's only about the correct mapping.

Then again, according to snippets for multiple controllers I thought it's Just Work(TM), but as already mentioned, it opens only the last plugged in joystick. If you have any experience related to using multiple joysticks, advices welcome. 🐧

yglazner:
I'm also in need for multi-gampad support, any progress on the issue?

dessant:
@yglazner, please use the reaction buttons if you'd like to vote up an issue in the future, these comments do not add anything to the discussion, and all subscribers get notified.

KeyWeeUsr:
@yglazner According to a small talk I had on their IRC (freenode#SDL) it should work. Suggested approach was with [`SDL_JoyDeviceEvent`](https://wiki.libsdl.org/SDL_JoyDeviceEvent) but that kind of doesn't work for me and still only a single device is usable. Maybe it's just me though, I have only a single **physical** device. If you have 2+, feel free to test and share your experience. It's better than nothing.

[...]

PR #4952, opened 2017-01-26
Add multiple joysticks support

KeyWeeUsr:
Doesn't work for me with a single physical device and vJoy, but apparently works for **physical** devices. ref #4872 

@yglazner please compile from this branch like a casual installation for your platform and confirm if it works for you - 2+ devices connected at the same time give values with [this](https://github.com/kivy/kivy/blob/master/examples/miscellaneous/joystick.py) or a similar example just with a different **id**, for example if a button is pressed on each of the devices, the value in console is the same with **id** of the stick being the only thing different.

Or if you have different devices (different brand/model), then try to get the same axis/button/hat that returns the same value and compare ids.

    pip install https://github.com/KeyWeeUsr/kivy/archive/multi_joystick.zip

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -148,7 +148,11 @@ cdef class _WindowSDL2Storage:
             self.ctx = SDL_GL_CreateContext(self.win)
             if not self.ctx:
                 self.die()
-        SDL_JoystickOpen(0)
+
+        # Open all available joysticks
+        cdef int joy_i
+        for joy_i in range(SDL_NumJoysticks()):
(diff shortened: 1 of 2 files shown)


KeyWeeUsr:
@yglazner ping?

yglazner:
I haven't found time to compile it from source ... I did test the code I've posted above and it works for my 5 game-pads ...

KeyWeeUsr:
Ok, then vJoy seems to be broken, not sdl2 actually. I'll compile it again soon and merge as the code is the same. A test would be nice though. :)

KeyWeeUsr:
I've found super-cheap joysticks today, tested with 3 and all worked correctly. 💃

Issue #4941, opened 2017-01-21
MSVC Install / compile fails on x64 with venv


Badg:
### Versions

* Python: 3.5 x64
* OS: Windows 7 x64
* Kivy: stable 1.9.1; also, master dev branch
* Kivy installation method: MSVC compile instructions [here](https://kivy.org/docs/installation/installation-windows.html)

### Description

Attempting to compile/install Kivy fails with linking errors to glew dependencies (stable 1.9.1), or from missing ```gst/gst.h``` (1.9.2dev0 master branch). I followed the install docs exactly with one minor difference:

```
    "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x64
```

(note the x64 argument; otherwise it fails after trying to build for the wrong architecture target).

At first I thought this might have been related to #4216, but on further investigation it seems unrelated. Modifying path as discussed in docs within #4694 has no effect.

### Code and Logs

Building with stable results in the following error:

```
       Creating library build\temp.win-amd64-3.5\Release\Users\BADGER~1\AppData\
Local\Temp\pip-build-kpsgjh0k\kivy\kivy\graphics\svg.cp35-win_amd64.lib and obje
ct build\temp.win-amd64-3.5\Release\Users\BADGER~1\AppData\Local\Temp\pip-build-
kpsgjh0k\kivy\kivy\graphics\svg.cp35-win_amd64.exp
    svg.obj : error LNK2001: unresolved external symbol __glewIsRenderbuffer
    svg.obj : error LNK2001: unresolved external symbol __glewIsFramebuffer
    svg.obj : error LNK2001: unresolved external symbol __glewGetRenderbufferPar
ameteriv
    svg.obj : error LNK2001: unresolved external symbol __glewFramebufferTexture
2D
    svg.obj : error LNK2001: unresolved external symbol __glewFramebufferTexture
3D
    svg.obj : error LNK2001: unresolved external symbol __glewRenderbufferStorag
e
    svg.obj : error LNK2001: unresolved external symbol __glewCheckFramebufferSt
atus
    svg.obj : error LNK2001: unresolved external symbol __glewDeleteRenderbuffer
s
    svg.obj : error LNK2001: unresolved external symbol __glewDeleteFramebuffers

    svg.obj : error LNK2001: unresolved external symbol __glewFramebufferRenderb
uffer
    svg.obj : error LNK2001: unresolved external symbol __glewGenFramebuffers
[...]

PR #4956, opened 2017-01-26
Msvc venv install

Badg:
Fixes for #4941


setup.py
```diff
@@ -570,26 +570,46 @@ def determine_gl_flags():
...
+        # Not on OSX, and/or no sdl2_flags defined
+        sdl2_paths = []
+        sdl2_path_candidates = []
+        # Check in the executable's include dir
+        sdl2_path_candidates.append(
+            join(dirname(sys.executable), 'include', 'SDL2'))
+        # For every path in the %INCLUDE% (Windows), add the SDL2 subdir
+        sdl2_path_candidates.extend(
+            [join(p, 'SDL2') for p in
+             environ.get('INCLUDE', '').split(os.pathsep)]
```
matham: Why did you add this `INCLUDE`?

setup.py
```diff
@@ -570,26 +570,46 @@ def determine_gl_flags():
 
 def determine_sdl2():
     flags = {}
+    # This is redundant; the only time determine_sdl2 is called is preceded by
+    # verification that this condition is already satisfied
     if not c_options['use_sdl2']:
         return flags
 
+    # Check the environment for a Kivy SDL2 path (all platforms) or for Include
+    # and Lib (intended to catch Windows installation, see #4941)
```
matham: Generally it's not a good idea to add comments that just repeat what we can understand from looking at the code. I think most of these comments above and below can be removed. 

The idea of comments is to give the high level intent of the coder when it's not obvious from just looking at the code by an experienced coder. Generally python code is clear enough to need very few comments to an experienced python developer. So comments should only be added when that higher level intent is not clear. I would say that only the comment `# no pkgconfig info, or we want to use a specific sdl2 path, so perform manual configuration` is adding that as it explains why we are doing this stuff. Pretty much all the other comments should be removed.

You can see https://en.wikipedia.org/wiki/Comment_(computer_programming)#Code_description.

doc/sources/installation/installation-windows.rst
```diff
@@ -257,6 +296,18 @@ If the compilation succeeds without any error, Kivy should be good to go. You
 can test it with running a basic example::
 
     python share\kivy-examples\demo\showcase\main.py
+    
+.. note::
```
matham: I think this note is not needed as your code should fix this issue.

doc/sources/installation/installation-windows.rst
```diff
@@ -209,17 +210,25 @@ kivy with git rather than a wheel there are some additional steps:
...
+
+#. Set the environment variables.
+
+   .. warning::
    
-   Visual Studio is very big so you can also use the smaller,
-   `Visual C Build Tools instead
-   <https://github.com/kivy/kivy/wiki/Using-Visual-C---Build-Tools-instead-of-Visual-Studio-on-Windows>`_.
+      If compiling without the optional GSTREAMER dependency, set
+      ``USE_GSTREAMER=0`` instead of ``USE_GSTREAMER=1``.
```
matham: This is not needed as the default doesn't use gstreamer.

doc/sources/installation/installation-windows.rst
```diff
@@ -209,17 +210,25 @@ kivy with git rather than a wheel there are some additional steps:
...
+
+   **mingwpy:**
```
matham: The following applies both to msvc and mingwpy so this header should not be here.

doc/sources/installation/installation-windows.rst
```diff
@@ -230,6 +239,36 @@ kivy with git rather than a wheel there are some additional steps:
      export USE_GSTREAMER=1
 
    These variables must be set everytime you recompile kivy.
+   
+   **MSVC:**
```
matham: This whole msvc section should not be needed if you installed visual studio, which we officially support. If using the build tools, we don't officially support it so it should be added to the wiki instead, which is why I removed those instruction yesterday and moved it to the linked wiki. It's best if we only officially support the visual studio as that should work out of the box without anything else required. Python will automatically find it.

setup.py
Badg: I disagree what constitutes "obvious" in the context of a highly convoluted setup, but I will respect the maintainers' wishes and remove comments (when I get a chance to do so, as this already chewed up half of my afternoon, because the code was utterly unintelligible to a professional dev but casual contributor)

setup.py
Badg: Because without it, Kivy cannot find the SDL2 headers on Windows using virtual environments (which is the whole point of the PR).

doc/sources/installation/installation-windows.rst
Badg: Which bit is unnecessary? The docs immediately above that say to compile with ``USE_GSTREAMER=1``, which results in a build error; see [here](https://github.com/kivy/kivy/issues/4941#issuecomment-274259328).

setup.py
matham: I agree that the code is not particularity clear in terms of the larger contexts, but the local code should be clear. The type of comments I wanted to avoid is where they repeat info already in the code. 

E.g. `# SDL2 path was not contained in the environment; attempt to infer it` can be seen from the line above it. That line `sdl2_path = environ.get('KIVY_SDL2_PATH', None)` says the exact same things as the text. Similarly, `# Not on OSX, and/or no sdl2_flags defined` can be gottten by looking at the line above `if platform == 'darwin' and sdl2_flags:`. It's similar with the other comments.

setup.py
matham: But who defines this include? Is it the user that will point to the include directory? Shouldn't `join(dirname(sys.executable), 'include', 'SDL2'))` have already taken care of those? And why can't they use `KIVY_SDL2_PATH`.

doc/sources/installation/installation-windows.rst
Badg: I don't have the time right now to re-verify this locally, but I can say unequivocally that this did not work for me yesterday. Without a doubt, the CD + INCLUDE is required for the venv; that's what tells it to look for the headers locally instead of system-wide. Setup.py is LBYL, so otherwise it ignores SDL. See #4941.

doc/sources/installation/installation-windows.rst
matham: Oh, right sorry.

doc/sources/installation/installation-windows.rst
matham: But I thought that's what your fix does. It uses sys.executable, which I assumed is changed for venv to find the right header files? Is that not the case?

setup.py
Badg: They could use ``KIVY_SDL2_PATH``, if the documentation is changed to reflect that; at that point, ignore the whole PR because aside from code cleanup to reduce cyclomatic complexity, it will be functionally identical.

You have to ```set``` the include, as per preexisting docs. The ```join``` doesn't pick up the location, because it's not the same; that will look for the header in ```/venv/Scripts/include/SDL2``` which does not exist.

matham: I tried to speed read the issue as there was other stuff going on there. It wasn't clear to me how much the issue was the usage of build tools rather than visual studio which we don't officially support or venv. Doesn't sys.executable take care of the venv header issue? If not, we should find what the correct path is and use that.

doc/sources/installation/installation-windows.rst
Badg: No, the code was already using sys.executable, which doesn't work, see note above. It's checking the ```%INCLUDE %``` that actually does anything here; which is itself a shortcut to setting the SDL header location manually, which bypasses the entire SDL preparation logic (and is therefore more reliable, which is why I doc'd it as a note).

setup.py
Badg: I was never using the build tools, I was using MSCV, and following the pre-existing compile directions exactly. Where is build tools coming from? From the comment in the docs? That was there before this PR; I had nothing to do with it.

Venv and system python have different directory structures; that's the root cause of this whole problem. Because the install directions (used to, and in this PR, again do) specify that the %INCLUDE% has to be set, I made the decision to (more reliably) pull the SDL location from it than from inferring directory structure of the venv, in the hopes of a more robust solution.

doc/sources/installation/installation-windows.rst
matham: Right, I thought it wasn't used and that you had ordered it so that it would be used.

Virtualenv does indeed do this, it's just that the path is slightly different. To get the correct path in a virtual env you have to do `join(dirname(dirname(sys.executable)), 'include')`. That should fix the problem. So I'd try both  `join(dirname(dirname(sys.executable)), 'include')` and  `join(dirname(sys.executable), 'include')`.

setup.py
matham: See my other comment about how to find the include directory in a venv. About build tools, I'm referring to the previous instructions that I moved to the [wiki](https://github.com/kivy/kivy/wiki/Using-Visual-C---Build-Tools-instead-of-Visual-Studio-on-Windows). That seems to be what you followed (I'm assuming?).

However, that instructions is not really how we'd officially recommend you do it. Instead, you should download normal visual studio and that's it. Everything is the same as with mingwpy, which is why I updated the instructions to reflect that. There's no need for launching vcvarsall.bat or anything like that. Your problem comes from the venv include path issue. But the issue covered many things so I assumed initially it was due to those instructions (which is why I actually updated it and moved it to the wiki).

Badg: Okay, I think I've more or less cleared up the confusion on my end. A great deal of it was due to the old installation instructions.

I'll update the PR when I get a chance; it may be a few days though. I'm planning on:

1. Searching ```<sys.executable>/include``` then ```<sys.executable>/../include``` instead of ``%INCLUDE%``
2. Updating docs accordingly
3. Changing comments as requested
4. Keeping note about explicit setting for the ```SDL2``` path to help avoid support tickets for non-standard installs

matham: Yes, that sounds great! Although, the note about `KIVY_SDL2_PATH` probably can be shortened a bit and made more generic.

dessant:
This other pr might be related: https://github.com/kivy/kivy/pull/4896.

Badg:
@dessant : it's definitely addressing a similar issue, but maybe not the same (discussed somewhat tersely [here](https://github.com/kivy/kivy/issues/4941#issuecomment-275300805), though the install differences may render that discussion irrelevant). Regardless, I'm going to incorporate those changes into the search path I'm using. Using ```sys.prefix``` instead of awkwardly going to the parent directory is particularly clever; I didn't even realize ```sys.prefix``` existed.

Badg:
Okay, updated the PR as per discussion above. Also merged changes from #4896, so this PR now supersedes it. That said, while I've verified it works within an MSVC build, but don't have access to a mingw one, so I don't know **for sure** that it works there.

dessant:
Thanks for your work, but i will make the call now before you spend more time on this and close the pr, it's not clear why each change is made, the scope seems to be shifting, comments confuse rather than adding value.

We welcome you to open new atomic prs, explaining the reason for every change.

matham:
I guess in particular, it's not good to merge changes from other PRs, while those PRs are still open. Especially if it wasn't tested.

matham:
Also, there seems to be a misunderstanding about mingw; when we use or refer to using mingw(py), it's not that python is built using mingw (although is also possible, but rare), but rather that the current environment is using mingw(py), even though python is built using msvc.

matham:
I merged https://github.com/kivy/kivy/pull/4896 as that should fix the problem as you mentioned.

Badg:
@matham #4896 breaks the build on windows. You get linking errors and cannot install at all (at least from virtualenv, also from appveyor).

matham:
Should be fixed with https://github.com/kivy/kivy/commit/57d41c920997ae2b373b0bfa57862845d99337fb I believe.

PR #4961, opened 2017-01-28
Update video.py

doriclazar:
Small error in a code.
"value" is passed as an argument, but (not yet defined) "video" used.


inclement:
Thanks!

PR #4964, opened 2017-01-29
Reflect OpenWeatherMap's new API key requirements

Crowbrammer:
(Should say 'requires an API _key_'.)


dessant:
It could be enough to replace the url with https://httpbin.org/headers and adapt the callback function.

Crowbrammer:
That'd be great! :) And thanks for being active in the IRC. It means a lot.

dessant:
Fixed in https://github.com/kivy/kivy/pull/5061.

PR #4970, opened 2017-01-31
Win 3.6 whel links.

matham:
Still need to sync with the website.


KeyWeeUsr:
Oh, so they added 3.6 support on appveyor already, nice!

PR #4972, opened 2017-02-04
Fix OSC.py for py3 and py2

FeralBytes:
I fixed this already once but seems it never got accepted or was reversed and thus is broke again. Tested on py3 so please let the Travis test complete to ensure all is good for py2.

### DIFF
--- a/kivy/lib/osc/OSC.py
+++ b/kivy/lib/osc/OSC.py
@@ -103,6 +103,8 @@ class OSCMessage:
 
     def rawAppend(self, data):
         """Appends raw data to the message.  Use append()."""
+        if not isinstance(data, bytes): 
+            data = data.encode('utf-8')
         self.message = self.message + data
 
     def getBinary(self):


FeralBytes:
@inclement you have been using py3 and OSC; but you did not encounter this error in Py3? I also had a few other places in OSC.py that I needed to fix; so I am surprised. I will be investigating more in the coming days; but wanted to hear from you too; to help make sense of it. Because I would think Pyonic running in Python3 mode would certianly encounter these issues with OSC.

FeralBytes:
@inclement Now I C; you are using your own version of OSC; not Kivy's. https://github.com/inclement/Pyonic-interpreter/tree/master/pyonic/osc
Shouldn't this be part of a PR to fix OSC for Py2 & Py3: https://github.com/inclement/Pyonic-interpreter/commit/42717866522c51eb0413c7a2bc1194bea9f9f6b1
Sorry just asking; because it looks like you did the same thing I did just a few days ago.

FeralBytes:
@inclement also I really am not trying to call you out; was just unsure if my changes were really needed when I realized Pyonic was using OSC. If you do not have time. I will be making a fuller PR for OSC py3 &py2 compatable changes once I am done with germn testing the ffpyplayer_new_toolchain.

inclement:
I made only a minimal set of changes for Kivy's osc to work with
python3, they aren't robust or generally useful as it will still crash
easily, e.g. if given non-bytestrings. I avoid these by sanitising the
input first in Pyonic, but a proper fix would want to make the osc
module do it itself.

FeralBytes:
OK; but you are OK if i still make the changes in small doses; not a whole sale fix; because I still prefer small PR's. The PR's I make will push us in the right direction; they just might not get us all of the way there until later. The biggest area in OSC in which I did not touch was the bundle code; so I will have to create some code to test that area in a while. In a few days I will have some time off. Then I intend to more throughly test this PR and create the others.

inclement:
For something like fixing osc, doing a single PR is fine. It probably
won't end up being that large.

FeralBytes:
Closing in favor of a more complete PR.

PR #4973, opened 2017-02-04
[WIP] Manylinux wheels

KeyWeeUsr:
I'm trying to create manylinux wheels with Travis as shown in [this example](https://github.com/pypa/python-manylinux-demo), but the problem is basically that the docker image uses/is **`CentOS 5.x`**(basically old Red Hat distro), thus:

* no `SDL2` related packages (we can't use SDL<2)
* didn't find GLES & mesa packages
* missing khrplatform-devel
* missing mtdev (probably not necessary)

I've been looking for the packages at http://rpm.pbone.net but the only thing it found were quite new packages for Fedora 25 and those don't like being installed on such an old platform, so to make these wheels work we'll need first the GLES packages. If there's some source, I might be able to compile it, but free Travis builds are limited to 1 hour (iirc), so compiling everything from source probably isn't the best solution.

The next thing is Python 2.6. I'm not certain, but in the docker image there should be even python 2.7 and 3.x too, but I think we might need to create our own docker file (judging from the [pygame solution](https://bitbucket.org/pygame/pygame/src/407caa445ee0/manylinux-build/?at=default)).

What might fix this thing are SDL2 manylinux wheels (@matham ?), but that could be overkill if Kivy dependencies are available at repositories for modern gnu/linux distros and we'd probably still need opengl deps and other required deps (therefore we'd end up making wheels for the whole dependency tree just to compile Kivy).

If someone feels adventurous and would like to speed it up, I'll leave the edits from other maintainers allowed, as always. 😋


dessant:
As discussed on irc, there have to be kivy.deps.* wheels built separately and then pip installed in the docker container before creating the main kivy wheel, we must not use or depend on distro libs for our dependencies, kivy.deps.* wheels will be built by compiling sdl2 ourselves.

KeyWeeUsr:
Well, it is getting forward a bit by bit, however there are packages needed by SDL2 and others. Those are nowhere to find:
* [SDL2](https://github.com/kivy/kivy/pull/4973/files#diff-59eb4e6626fc3fc86d6b8dbc0d3fa82bR57)
* [pulseaudio](https://travis-ci.org/kivy/kivy/jobs/209712944#L524)

However, SDL2 seems satisfied with [libasound](https://travis-ci.org/kivy/kivy/jobs/209712944#L1719), [OpenGL GLX](https://travis-ci.org/kivy/kivy/jobs/209712944#L1766) instead of OpenGL ES, is nuts about [dbus](https://travis-ci.org/kivy/kivy/jobs/209712944#L1778) although it's installed and apparently uses [X11](https://travis-ci.org/kivy/kivy/jobs/209712944#L1833) instead of OpenGL although it reports GLX is ok.

From compilation [this](https://travis-ci.org/kivy/kivy/jobs/209712944#L1976) is a complete wtf to me and it fails SDL2 compilation right in the beginning, so Kivy has no SDL2 to work with.

KeyWeeUsr:
Well, currently it's rather a proof of working wheel(s), but it still needs some things done hopefully by the end of the week. The `docker` folder is +/- completely forked from the pypa's manylinux repo (MIT, I have it mentioned in readme that's WIP locally for now) except the Docker files that were customized for our needs mainly because missing crucial libs, so the OS (and basically the Docker image) was changed to CentOS 7, therefore it's not an official `manylinux1` wheel although it should work just fine on newer versions of GNU/Linux distros that have glibc >= 2.17.

Still even with that some of the libs are from rather unofficial package repos, so we will need some backup of those in case the repos die (well, until compiling from source is done for such libs).

Still missing stuff:

* switching to compiling from source (mainly SDL2)
* separating the libs into similar structure we have on Windows (e.g. `kivy.deps.sdl2`) - we could have `kivy.libs.sdl2` so that it isn't mixed with the Windows packages + the folder itself that contains the `.so` files is called `.libs`
* fixing tests for GStreamer
* uploading to server
* adding the rest of the Travis builds
* probably even something else

KeyWeeUsr:
GStreamer remaining! More fun soon 😎 

image

Also, I probably have an idea how to kick the damn thing to support even 32bit wheels, which would be amazing as my little traveling netbook is 32bit. That's being said, I'll maybe do something about it after I make GStreamer work and/or after all the tests start to pass.

dessant:
Fixes https://github.com/kivy/kivy/pull/4950.

matham:
Closing because we now use pypa's recently published manylinux2010 docker to make wheels.

PR #4976, opened 2017-02-06
begin new gsoc page for 2017

Kovak:
Does anyone else have project ideas for GSoC? Due date is the 7th for psf. I think we are planning on doing plyer project again, but do not know who wants to mentor.

### DIFF
--- a/doc/sources/gsoc.rst
+++ b/doc/sources/gsoc.rst
@@ -1,4 +1,4 @@
-Google Summer of Code - 2016
+Google Summer of Code - 2017
 ============================
 
 Introduction
@@ -136,7 +136,7 @@ wheelhouse of the average Pythonista.
     from p4a/kivy-ios to plyer and implementing some or all
     of the new facades to be decided with the student.
     
-  - **Mentors**: ?
+  - **Mentors**: Akshay Arora, Sebastian Armin
   - **Requirements**: Access to Linux, Windows, OS X, iOS device,  
     Android device.
   - **Task level**: Intermediate
--- a/doc/sources/gsoc2016.rst
+++ b/doc/sources/gsoc2016.rst
@@ -136,7 +136,7 @@ wheelhouse of the average Pythonista.
     from p4a/kivy-ios to plyer and implementing some or all
     of the new facades to be decided with the student.
     
-  - **Mentors**: ?
+  - **Mentors**: Akshay Arora, Ryan Pessa
   - **Requirements**: Access to Linux, Windows, OS X, iOS device,  
     Android device.
   - **Task level**: Intermediate
@@ -152,7 +152,7 @@ with C code, and using Cython to build your own C extensions these projects
 may appeal to you.
 
 
-**Kivent: Chipmunk 7 Integration**
+**Kivent: Tiled Integration**
 
   Description:
     KivEnt is a modular entity-component based game engine built on top of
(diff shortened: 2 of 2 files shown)


yaki29:
@Kovak @inclement There are still many recipes that are in old_toolchain but not in the new one. Some of them are even very useful(like django). Also there must be a testapp for every recipe so that any newbie can try that recipe without any problem means he don't have to search internet learn that framework or library and make an app then try p4a on that just to test how it works on that library or framework and what are the limitations(those testapps may also consist of buildozer.spec file just in case someone wants to use buildozer instead of p4a). Third, there are still many recipes which are to be made that people use on daily basis and want to use p4a to make an app. When they ask on irc they get to know that either make your own recipe or don't use it. Fourth and the last one, there are recipes even in master that works only with old versions of that library (best example is opencv, opencv3 has come and it is way too different from opencv2, more and more people are bending towards opencv3 but p4a-master opencv recipe still crashes while building opencv3 application. That's why each recipe must be tested and must go with the latest versions. So, to summarize I just want to say that 1) all old_toolchain recipes must be made in python to include in master branch, 2) There must be a testapp for each recipe ( makes it easy to test recipes), 3) More latest and everyday used libraries must be included in master branch, 4) All present libraries must be verified by testing them with the latest versions( they must support new versions of the libraries). 
I want reviews on this(I want this to be added in gsoc 2017). Correct me if I am wrong somewhere. Changes and suggestion are most welcome.

phunsukwangdu:
Yes i do have a idea and drafted a rough temprory proposal for it.
https://docs.google.com/document/d/1gXcsH0vTwcvhxhaosnDtYv2XFTSTrigALw8wGxN9O9w/edit?usp=sharing

inclement:
@yaki29 These are good ideas, but to be honest I'm not very enthusiastic about such a project for gsoc. While it would begreat to have more recipes and examples, I'm not sure this is actually a good gsoc target for multiple reasons - it's not on its own that large a task (whereas gsoc is a pretty long period), it isn't really an interesting coding problem, and in the end it's probably relatively unimportant - a big reason for these recipes not being ported yet is that nobody seems to be using them. It actually looks like the new toolchain now has a lot more recipes than the old one, but they're different because they reflect what people needed at the time.

Of course, it's still nice to have more recipes, and there are other exceptions to this like the opencv version problem. However, for gsoc this has the same issue that just fixing these for now isn't a big project, and also wouldn't fix things in the long term. It would be more interesting to make changes to allow flexible recipes that potentially could try and pull the current version, but this would be a much more complex project involving really understanding and reimagining parts of the python-for-android internals.

I should also note that I'm cautious about committing my own time to gsoc, as I wouldn't want to be busy or unable to provide full support over the period.

Thank you for the suggestion, it's good thinking and not a bad idea, but unfortunately it's probably not something we'll be targeting for a gsoc project.

yaki29:
@inclement Thanks for your reply. Like you shared the idea of flexible recipes which can find latest version itself and make it work on its own. It doesn't increases the complexity of the project and doesn't actually need reimagining parts of p4a internals. What it needs is knowledge of web scraping tools which are to be included in recipe itself so that it can find what is latest activity that has happened on the page and if something new has happened is it new version or something else, if not it will download the last latest version it finds. But it needs some series thinking.
If this was in your mind you didn't discuss it with me and directly claimed here that it is too complex.
This was already in my mind that's why I was working on BeautifulSoup4 recipe so that p4a contains at least the recipes needed for that. I was also thinking of adding something to buildozer or p4a which can identify the custom version user wants to set ( like 
requirements= package_one==1.9.4,package_two==1.4.5). This would let the user set requirements according to his own need. It comes in handy when some users' system doesn't support particular version your recipe is downloading but it supports some other version. What you would suggest is go to the p4a in the build directory find its recipe and change the version to what you want. That is some real difficult task for a newbie.

And, so what you won't be able to give full support, what we need to work is internet and laptops :yum:  . As you said you are little busy during the week you can give time during the weekend (well that was none of my business but still).

In the last I just want to say that discuss your ideas with people who are working on that may be something awesome comes out.
Again thanks for you reply. Correct me if I'm wrong somewhere and suggestions are most welcome.

PR #4987, opened 2017-02-10
Split examples into separate wheel for windows

matham:
PR should be squashed.

You can see results at https://drive.google.com/drive/folders/0B1_HB9J8mZepOV81UHpDbmg5SWM.


KeyWeeUsr:
No auto install (require in setup.py) of the examples with kivy wheel? If so, then some docs would be nice too 😋

PR #4989, opened 2017-02-10
Make cc0 video smaller

KeyWeeUsr:
The video is now 1.65MB (previously 4.36MB), with a smaller resolution + bitrate changed to 2000kbps. Also I corrected the timing of subtitles which seemed a bit off as the video is ~7sec long.

Changing the video to a lower bitrate might help with the size more and would probably kick even under 1MB, but the resolution is really small now and the frames are changing to pixelated something with compression artifacts. I'm not sure if I should try to push it more and make the quality worse.

### DIFF
--- a/examples/widgets/cityCC0.jsa
+++ b/examples/widgets/cityCC0.jsa
@@ -1,9 +1,9 @@
 [
-	{"start": 0, "duration": 2, "text": "This is an example of annotation"},
-	{"start": 2, "duration": 2, "text": "You can change the background color", "bgcolor": [0.5, 0.2, 0.4, 0.5]},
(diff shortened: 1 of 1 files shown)


dessant:
If we have only one video file, it might be ok to keep it as is, this new version is really low quality. I'm also ok witch merging if others want it, it just seems like it would offer a not so flashy experience in a video playback example.

FeralBytes:
Would be nice if the video had a disclaimer like "Low Resolution Video Test".

dessant:
Yeah, but that wouldn't make the experience better.

dessant:
Let's close this. We could use the original mp4 file (2.2MB) when the ffpyplayer recipe for android gets merged.

https://videos.pexels.com/videos/time-lapse-video-of-tall-buildings-894

PR #5015, opened 2017-02-22
Fix to_scroll() on RecycleViews

### LINKED ISSUE #4922: Why unbind _trigger_layout for RecycleBoxLayout?
Referring to __init__ function in RecycleBoxLayout, where we currently do funbind('children', self._trigger_layout)

Not sure on why this is necessary, as this breaks ScrollView's scroll_to. The code was introduced fairly recently in https://github.com/kivy/kivy/pull/4633 as a response to https://github.com/kivy/kivy/issues/4619

I'm trying to implement goto_view on RecycleView currently, hence why I came across this. I'm not sure whether the funbind is required, or whether we should instead check for _trigger_layout being callable in ScrollView.

### LINKED ISSUE #5014: RecycleView needs a function to scroll directly to a widget
When calling `scroll_to` on a RecycleView apps crash with the exception:

```
...
   File "/usr/lib64/python3.5/site-packages/kivy/uix/scrollview.py", line 899, in scroll_to
     if self._viewport._trigger_layout.is_triggered:
 AttributeError: 'function' object has no attribute 'is_triggered'
```

`scroll_to` assumes `_trigger_layout` is a ClockEvent object, however RecycleLayout sets it to a function instead.

I'll be submitting a PR with a potential solution momentarily.

hunternet93:
This allows `scroll_to` to work with RecycleView, fixing bug #5014. This may also resolve #4922 , although I'm not entirely sure.

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -130,7 +130,7 @@ from functools import partial
 from kivy.animation import Animation
 from kivy.compat import string_types
 from kivy.config import Config
-from kivy.clock import Clock
+from kivy.clock import Clock, ClockEvent
 from kivy.factory import Factory
 from kivy.uix.stencilview import StencilView
 from kivy.metrics import sp, dp
@@ -896,7 +896,7 @@ class ScrollView(StencilView):
 
         # if _viewport is layout and has pending operation, reschedule
         if hasattr(self._viewport, 'do_layout'):
-            if self._viewport._trigger_layout.is_triggered:
+            if type(self._viewport._trigger_layout) is ClockEvent and self._viewport._trigger_layout.is_triggered:
                 Clock.schedule_once(
                      lambda *dt: self.scroll_to(widget, padding, animate))
                 return


kivy/uix/scrollview.py
```diff
@@ -896,7 +896,7 @@ def scroll_to(self, widget, padding=10, animate=True):
 
         # if _viewport is layout and has pending operation, reschedule
         if hasattr(self._viewport, 'do_layout'):
-            if self._viewport._trigger_layout.is_triggered:
+            if type(self._viewport._trigger_layout) is ClockEvent and self._viewport._trigger_layout.is_triggered:
```
matham: This should be `isinstance(self._viewport._trigger_layout, ClockEvent) and ...`

hunternet93:
True, this doesn't really fix the problem. My use case was scrolling down widget-by-widget as part of keyboard navigation, so it sorta worked, but a function to scroll by RV data index would be much better. I'll take a look and see if I can figure that out, rather than this quick fix that doesn't really help.

PR #5036, opened 2017-03-03
Add mingw angle support

matham:
Test still seem to fail.


tito:
What is that not merged? Only doc change right?

matham:
I compile angle on mingw and it worked on my machine. But on appveyor it failed. I think it was missing a dll. I could never figure out which one though even though I tried with depends exe.

matham:
We can try using msvc angle on < 3.5, but I'm not sure that's a good idea due to mixing compilers etc.

Issue #5044, opened 2017-03-08
TextInput crashes on android


Fogapod:
### Versions

* Python: 2.7
* OS: android
* Kivy: master

### Buildozer .spec
* android.api = 14
* android.minapi = 8
* android.sdk = 23
* android.ndk = 9c

### Description

Sometimes app get crashed when I type numbers into textinput field. [The same problem on StackOverflow](http://stackoverflow.com/questions/42640986/buildozer-kivy-textinput-crashes-on-numeric-input-in-android)

### Code and Logs

```python
File "/home/foga/Main/Bot/.buildozer/android/platform/build/dists/textinput_tester/private/lib/python2.7/site-packages/kivy/core/window/window_sdl2.py", line 580, in _mainloop
    AttributeError: 'LogFile' object has no attribute 'encoding'

```

Fogapod:
#5045 this solution helped me

Fogapod:
#5043

Fogapod:
The same issue

Issue #5003, opened 2017-02-19
Examples: Keyboard & KeyboardListener


KeyWeeUsr:
### Versions

* Python: 2.7, 3.5
* OS: Win8.1, 64bit
* Kivy: master, wheels

### Description

Example `examples/keyboard/main.py` dead for py3(.5.3).

> UnicodeDecodeError: 'charmap' codec can't decode byte 0x88 in position 3960

Similar for `examples/widgets/keyboardlistener.py` after pressed modifier (Ctrl, Shift)

> UnicodeEncodeError: 'charmap' codec can't encode character '\u0131' in position 12 ...

akshayaurora:
Both work on python 3.6.0 on OS X,  could you upgrade your python version and test again ?

KeyWeeUsr:
Yep, dead with the same error on 3.6 too. I run in `cmd.exe` and although they changed the encoding in [pep528](https://www.python.org/dev/peps/pep-0528/) and 529, this didn't fix it, so the problem is probably somewhere in the conversion unicode↔bytes

```
 Traceback (most recent call last):
   File "share\kivy-examples\keyboard\main.py", line 222, in <module>
     KeyboardDemo().run()
   File "_\lib\site-packages\kivy\app.py", line 802, in run
     root = self.build()
   File "share\kivy-examples\keyboard\main.py", line 216, in build
     self.sm.add_widget(KeyboardScreen(name="keyboard"))
   File "share\kivy-examples\keyboard\main.py", line 163, in __init__
     self._add_keyboards()
   File "share\kivy-examples\keyboard\main.py", line 169, in _add_keyboards
     layouts = list(VKeyboard().available_layouts.keys())
   File "_\lib\site-packages\kivy\uix\vkeyboard.py", line 367, in __init__
     self._load_layouts()
   File "_\lib\site-packages\kivy\uix\vkeyboard.py", line 433, in _load_layouts
     basename(splitext(fn)[0]))
   File "_\lib\site-packages\kivy\uix\vkeyboard.py", line 440, in _load_layout_fn
     json_content = fd.read()
   File "_\lib\encodings\cp1250.py", line 23, in decode
     return codecs.charmap_decode(input,self.errors,decoding_table)[0]
 UnicodeDecodeError: 'charmap' codec can't decode byte 0x88 in position 3960: character maps to <undefined>
```

matham:
Have you tried running with `PYTHONLEGACYWINDOWSSTDIO=1` set in the env?

[...]

PR #5045, opened 2017-03-08
Fix TextInput crash on android

Fogapod:


Fogapod:
#5044

KeyWeeUsr:
This is a bad fix, it basically reverts a fix for #5003 (https://github.com/kivy/kivy/commit/92464860152abdbc8b9623eb76b0d558124d51cc) on android. I'm not sure if it's "needed" to behave that way, but having a workaround like this makes it only inconsistent. Rather try to find an appropriate `encoding` attribute for that to apply.

Issue #5043, opened 2017-03-08
window_sdl2.py: TextInput crashes on numeric input in Android


ScissorHIll:
### Versions

* Python: 2.7.6
* OS: Ubuntu 14.04
* Kivy: 1.9.2
* Kivy installation method: kivy-daily PPA

### Description

Trying to use TextInput on Android. Works perfectly on Desktop, but app crashes when attempting to input anything on Android.

### Code and Logs
Example taken from the [Kivy Basic documentation](https://kivy.org/docs/guide/basic.html#customize-the-application):

```python
from kivy.app import App
from kivy.uix.gridlayout import GridLayout
from kivy.uix.label import Label
from kivy.uix.textinput import TextInput


class LoginScreen(GridLayout):

    def __init__(self, **kwargs):
        super(LoginScreen, self).__init__(**kwargs)
        self.cols = 2
        self.add_widget(Label(text='User Name'))
        self.username = TextInput(multiline=False)
        self.add_widget(self.username)
        self.add_widget(Label(text='password'))
        self.password = TextInput(password=True, multiline=False)
        self.add_widget(self.password)


class MyApp(App):

    def build(self):
        return LoginScreen()


if __name__ == '__main__':
    MyApp().run()
```

The app runs on Android, the keyboard also comes up. But crashes on typing anything. Output of `$adb logcat | grep python` (last few lines only):

```
03-07 10:19:48.972  7842  7857 I python  : [INFO   ] [GL          ] Texture max units <16>
03-07 10:19:48.976  7842  7857 I python  : [INFO   ] [Window      ] auto add sdl2 input provider
03-07 10:19:48.978  7842  7857 I python  : [INFO   ] [Window      ] virtual keyboard not allowed, single mode, not docked
03-07 10:19:48.981  7842  7857 W linker  : /data/data/org.test.textinput_tester/files/app/lib/python2.7/site-packages/jnius/jnius.so: is missing DT_SONAME will use basename as a replacement: "jnius.so"
03-07 10:19:49.379  7842  7857 I python  : [WARNING] [Base        ] Unknown <android> provider
03-07 10:19:49.379  7842  7857 I python  : [INFO   ] [Base        ] Start application main loop
03-07 10:19:49.383  7842  7857 I python  : [INFO   ] [GL          ] NPOT texture support is available
03-07 10:19:49.386  7842  7857 I python  : [ERROR  ] [Base        ] Failed to import "android" module. Could not remove android presplash.
[...]

PR #5048, opened 2017-03-08
- there is no sys.stdout.encoding when piped or frozen

mahomahomaho:
sometimes sys.stdout.encoding is None (eg. when piped or when frozen with pyinstaller). in such case pressing any key crashes app.


KeyWeeUsr:
@matham @mahomahomaho what about:

    if hasattr(sys.stdout, 'encoding'):
        encoding = sys.stdout.encoding
    else:
        encoding = 'utf8'

ref #5043

mahomahomaho:
It won't work in my case, because sys.stdout has attribute 'encoding', but it's None

PR #5049, opened 2017-03-08
fix ActionDropDown.on_touch_down: return True

ecdsa:
this fixes event propagation when a dropdown menu is pressed.
on_touch_down should return True if the parent returns True.

### DIFF
--- a/kivy/uix/actionbar.py
+++ b/kivy/uix/actionbar.py
@@ -264,6 +264,7 @@ class ActionDropDown(DropDown):
         if super(ActionDropDown, self).on_touch_down(touch):
             if self.auto_dismiss:
                 self.dismiss()
+            return True
 
 
 class ActionGroup(ActionItem, Spinner):


KeyWeeUsr:
Can't reproduce https://github.com/kivy/kivy/commit/574780c9986818e76ee8370e2bc4d100e7416824 so quickly, but I'll try to reproduce it with older commits. It may have something to do with the change in https://github.com/kivy/kivy/commit/0b50ee2f1a18b9babd6c585f393b9fa7156ea8d6 - or not, but that fixed two touches, which might explain the dismissing.

The reproducing is basically about trying to change `auto_dismiss` in the `actionbar.py` example at the bottom from both Python and KV and the buttons have to behave correctly - dispatching events once, not closing the dropdown when `auto_dismiss` isn't active, etc.

ecdsa:
I have a drop down menu with several buttons in my app. without the "return True" statement, a touch event is sent to the dropdown, but also to the widgets hidden under the dropdown.

KeyWeeUsr:
I have not said no to the PR and I know why it needs to get back, but there's a testing needed because that bug hides very well. 🙍 I checked Windows with master + this and it **_seems_** to be ok again, but there remains to find the bug again to check what fixed it and check if it doesn't go wrong on Android too (well, it should behave the same, but anyway...).

KeyWeeUsr:
I couldn't reproduce it even without reverting old changes in `dropdown.py` and `actionbar.py` both on Windows and Android which is weird. Anyway, we know where to look if something breaks with `ActionDropDown` again.

Thanks! 👼

PR #5052, opened 2017-03-10
Fix ffpyplayer crash on Windows

germn:
Using ffpyplayer on Windows may lead to crush since some of `img.to_memoryview()` returning values may be `None`:

>    File "C:\Users\gmn\Dropbox\gmn\code\python\LLP\core\video_ffpyplayer.py", line 72, in _redraw
>      self._tex_u.blit_buffer(du, colorfmt='luminance')
>    File "kivy\graphics\texture.pyx", line 934, in kivy.graphics.texture.Texture.blit_buffer (kivy\graphics\texture.c:13449)
>  IndexError: Out of bounds on buffer access (axis 0)

This PR fixes it.

Video that leads to crash:
https://yadi.sk/i/au0NqZp43EsxA5
Tested on Windows 10, crash happens not always, on first `_redraw` call.

### DIFF
--- a/kivy/core/video/video_ffpyplayer.py
+++ b/kivy/core/video/video_ffpyplayer.py
@@ -199,11 +199,12 @@ class VideoFFPy(VideoBase):
         if self._texture:
             if self._out_fmt == 'yuv420p':
                 dy, du, dv, _ = img.to_memoryview()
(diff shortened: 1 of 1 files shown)


matham:
I don't understand why it would ever be None, but I guess it can't hurt..

PR #5053, opened 2017-03-10
ffpyplayer video: update frame/position on seek if video paused

germn:
This PR is about to fix:
https://github.com/matham/ffpyplayer/issues/21

@matham, code works, what do you think about it?

### DIFF
--- a/kivy/core/video/video_ffpyplayer.py
+++ b/kivy/core/video/video_ffpyplayer.py
@@ -260,23 +260,23 @@ class VideoFFPy(VideoBase):
                     relative=False)
                 self._next_frame = None
 
-            # If player is paused, but frame outdated,
-            # we need to unpause ffpyplayer to get actual frame.
-            if self._next_frame is None and self._ffplayer.get_pause():
-                self._ffplayer.set_volume(0)  # Try to do it silently.
+            # Get next frame if paused:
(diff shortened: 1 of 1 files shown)


kivy/core/video/video_ffpyplayer.py
```diff
@@ -260,9 +260,37 @@ def _next_frame_run(self):
...
 
-            t1 = time.time()
-            frame, val = ffplayer.get_frame()
-            t2 = time.time()
+            # Get next frame if paused:
+            if self._next_frame is None and ffplayer.get_pause():
+                ffplayer.set_volume(0.0)  # Try to do it silently.
+                ffplayer.set_pause(False)
+                try:
+                    to_skip = 6
```
matham: The problem last time I looked at it was that it's not 6 frames, it can be more or less, so you may end up seeking too far or too little. Last time I tried to see if I can fix this in ffpyplayer directly, but didn't get to it. I guess as a hack it may work for now though?

Also, instead of checking if `self._next_frame` is None, you should explicitly check if there was a seek earlier by e.g. setting a seek flag above in `if seek_queue:`.

germn:
Added explicit checking on seeking when paused, added commentary for magic number.

I agree solution is far from ideal, but, I assume, still better then no solution at all :)

PR #5054, opened 2017-03-12
Fix inconistent naming, where kv files are not unloaded

Zen-CODE:
This fixes an issue where Builder.unload_string fails. The Builder.load_file function adds the full path before adding it to the 'files' property (line 285).
```
   filename = resource_find(filename) or filename
```
This is not done when unloading, so the filename is not found in 'files'and nothing is unloaded.

### DIFF
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -319,6 +319,8 @@ class BuilderBase(object):
             if y[2] != filename:
                 templates[x] = y
         self.templates = templates
+
+        filename = resource_find(filename) or filename
         if filename in self.files:
             self.files.remove(filename)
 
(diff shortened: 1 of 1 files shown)


rafalo1333:
@Zen-CODE I am not sure about changing "Lang" -> "Builder" in Logger message. Some time ago we made the KV Lang logs consistent by using common "Lang" descriptor everywhere. Reverting this one is not a wise choice in my opinion.

Zen-CODE:
Okay, removed that. There is quite a lot of duplication between lang and builder, and I'm not sure why that is. I added that as the message can occur in 2 places, so is ambiguous. 

But yes, a PR should address 1 issues :-)

akshayaurora:
@Zen-CODE 👍

PR #5056, opened 2017-03-14
Avoid an android crash introduced in #5003 (git: 92464860).

learnleapfly:
On android, sys.stdout.encoding does not exist (sys.stdout is redirected - of type LogFile - which has no `encoding` attribute).

### DIFF
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -577,7 +577,9 @@ class WindowSDL(WindowBase):
                     try:
                         kstr_chr = unichr(key)
                         try:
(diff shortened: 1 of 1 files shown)


matham:
There are pep8 errors that need to be fixed before it can be merged.

llfkj:
After scouring the python2 and python3 documentation, I'm pretty sure this is the only "right" way to query for `sys.stdout.encoding`. (Whether doing this query is the "right thing" is a different topic). This patch will at least fix the crash-on-android issue that many have encountered. 

For more, see my longer comment on the  [related p4a issue (1018)](https://github.com/kivy/python-for-android/issues/1018)

PR #5067, opened 2017-03-19
Logger: purge_logs(): Replace hardcoded value *maxfiles* with config setting *log_maxfiles*.

ErwinRieger:
The number of kept logfiles in FileHandler.purge_logs() is hardcoded to 100 (*maxfiles*). This pull request adds a new config setting *log_maxfiles* that is used instead of this hardcoded value. Default value is the previous value of 100. Bumped KIVY_CONFIG_VERSION up to 19.

### DIFF
--- a/kivy/logger.py
+++ b/kivy/logger.py
@@ -59,7 +59,6 @@ import os
 import sys
 import kivy
 from kivy.compat import PY2
-from kivy.config import Config
 from random import randint
 from functools import partial
 
@@ -117,6 +116,7 @@ class FileHandler(logging.Handler):
         if randint(0, 20) != 0:
             return
 
+        from kivy.config import Config
         maxfiles = Config.getint('kivy', 'log_maxfiles')
 
         if maxfiles <= 0:
@@ -150,6 +150,7 @@ class FileHandler(logging.Handler):
 
     def _configure(self, *largs, **kwargs):
         from time import strftime
+        from kivy.config import Config
         log_dir = Config.get('kivy', 'log_dir')
         log_name = Config.get('kivy', 'log_name')
 
@@ -222,6 +223,7 @@ class FileHandler(logging.Handler):
         if FileHandler.fd is None:
             try:
                 self._configure()
+                from kivy.config import Config
                 Config.add_callback(self._configure, 'kivy', 'log_dir')
                 Config.add_callback(self._configure, 'kivy', 'log_name')
             except Exception:
--- a/kivy/logger.py
+++ b/kivy/logger.py
@@ -37,6 +37,7 @@ The Logger can be controlled via the Kivy configuration file::
     log_enable = 1
     log_dir = logs
     log_name = kivy_%y-%m-%d_%_.txt
+    log_maxfile = 100
 
 More information about the allowed values are described in the
 :mod:`kivy.config` module.
@@ -115,8 +116,8 @@ class FileHandler(logging.Handler):
         if randint(0, 20) != 0:
             return
 
-        # Use config ?
-        maxfiles = 100
+        from kivy.config import Config
+        maxfiles = Config.getint('kivy', 'log_maxfiles') - 1
 
         print('Purge log fired. Analysing...')
         join = os.path.join
@@ -131,16 +132,15 @@ class FileHandler(logging.Handler):
             # sort by date
             lst = sorted(lst, key=lambda x: x['ctime'])
 
-            # get the oldest (keep last maxfiles)
-            lst = lst[:-maxfiles]
-            print('Purge %d log files' % len(lst))
+            print('Purge %d log files' % (len(lst) - maxfiles))
 
-            # now, unlink every file in the list
-            for filename in lst:
+            # unlink the oldest files in the list,
+            # keep last maxfiles
+            for i in range(len(lst) - maxfiles):
                 try:
-                    unlink(filename['fn'])
+                    unlink(lst[i]['fn'])
                 except PermissionError as e:
-                    print('Skipped file {0}, {1}'.format(filename['fn'], e))
+                    print('Skipped file {0}, {1}'.format(lst[i]['fn'], e))
 
         print('Purge finished!')
 
--- a/kivy/logger.py
+++ b/kivy/logger.py
@@ -59,6 +59,7 @@ import os
 import sys
 import kivy
 from kivy.compat import PY2
+from kivy.config import Config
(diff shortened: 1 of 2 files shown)


kivy/logger.py
```diff
@@ -115,8 +116,8 @@ def purge_logs(self, directory):
         if randint(0, 20) != 0:
             return
 
-        # Use config ?
-        maxfiles = 100
+        from kivy.config import Config
```
KeyWeeUsr: I wonder if it wouldn't be better to move all `from kivy.config import Config` in that .py file to the top (there are 3).

kivy/logger.py
```diff
@@ -115,8 +116,8 @@ def purge_logs(self, directory):
...
+        maxfiles = Config.getint('kivy', 'log_maxfiles') - 1
```
KeyWeeUsr: I'm not sure if `- 1` is a good idea as the rest of the code expects `100` and not `99` as a value for the current behavior (`if len(lst) > maxfiles:`, `lst = lst[:-maxfiles]`, etc) . Probably just a matter of taste though.

kivy/logger.py
ErwinRieger: OK, i will add the "￼if not maxfile > 0: return" line and replace all Config imports with a single one...

kivy/logger.py
ErwinRieger: The question is the meaning of *log_maxfiles*. With the current implementation, you get one logfile if you specify log_maxfiles=1 in config.ini (purge removes all old logfiles and the logger then adds a new one). Without the "-1" you get two logfiles with the log_maxfiles=1 setting - which one is better?

KeyWeeUsr: I myself feel that the `log_maxfiles` should represent a number for the real state _before_ actually creating a new log i.e. keeping the variable kind of "local" to the `purge_logs()`, independent on the rest of the system. Imagine a case when someone wants to purge the logs manually(`purge_logs()` isn't a private method), this could lead to a confusion as for "why is there a file missing?".

Imagine a value from config equal to `10` and a total number of files equal to `100`. In case when `maxfiles = <config>` this is the output:
* total: 100
* erased: 90
* remaining: 10

in case when `maxfiles = <config> - 1` this is the output:
* total: 100
* erased: 91
* remaining: 9

ErwinRieger: This is a good point, we have to consider direct calls to `purge_logs()`. I am good with this and will remove the *-1* term. So the behaviour of *log_maxfiles* will be:

- `log_maxfiles <= 0`: disable purging, keep all logfiles.
- `log_maxfiles > 0`: purge logfiles and keep the most log_maxfiles recent logfiles. After that a new logfile will be added by the logger module so you will end up with `log_maxfiles+1` files.

kivy/logger.py
ErwinRieger: Replacing several 'import kivy.Config' statements with a single one at the top was not possible:
`I/python  ( 5561): Traceback (most recent call last):
I/python  ( 5561):   File "main.py", line 10, in <module>
I/python  ( 5561):     from kivy.support import install_twisted_reactor
I/python  ( 5561):   File "/data/data/de.ibrieger.ibr/files/app/lib/python2.7/site-packages/kivy/__init__.py", line 38, in <module>
I/python  ( 5561):     from kivy.logger import Logger, LOG_LEVELS
I/python  ( 5561):   File "/data/data/de.ibrieger.ibr/files/app/lib/python2.7/site-packages/kivy/logger.py", line 62, in <module>
I/python  ( 5561):     from kivy.config import Config
I/python  ( 5561):   File "/data/data/de.ibrieger.ibr/files/app/lib/python2.7/site-packages/kivy/config.py", line 320, in <module>
I/python  ( 5561):     from kivy import kivy_config_fn
I/python  ( 5561): ImportError: cannot import name kivy_config_fn
`

KeyWeeUsr:
@dessant should `log_maxfiles = 0` be handled the same way like `log_maxfiles = -1` i.e. disable removing, or should it be _remove all_ flag? I'm not quite sure about this, but I think it might be a separate behavior.

The import thing, well, there's this cat & mouse race for the `config.ini` (it's not created/located yet), so the imports might be troublesome. Let's ignore them or something important breaks.

I tried it with this:
```python
Config.set('kivy', 'log_maxfiles', 1)
from kivy import kivy_home_dir as hd
from kivy.logger import Logger
from os.path import join
Logger.handlers[1].purge_logs(join(hd, Config.get('kivy', 'log_dir')))
```

it doesn't work with config edited in a file (for me) most likely because the `FileHandler` is initialized before the code is even read (`100` files kept). I'm not sure if it's a bug per-se (i.e. if it can be fixed), though if not, it'd nice to have some note in the documentation like it is for the keyboard [here](https://github.com/kivy/kivy/blob/master/examples/keyboard/main.py#L134)(no clue why it isn't in the Config doc as it behaves similarly → [`Config.write()` + restart](https://github.com/kivy/kivy/blob/master/examples/keyboard/main.py#L144)).

Anyway, it looks really nice! :)

ErwinRieger:
Maybe it didn't removed logs because of the 5% random stuff ('if randint(0, 20) != 0: return').

KeyWeeUsr:
@ErwinRieger Nah (I put some statement before `return`), it's because of `<=` in `if maxfiles <= 0:` as I foolishly suggested in the beginning and missed the functionality for the `0` itself:

> ... a simple if not maxfile > 0: return (because -1 is still true)...

To introduce the `0` part just remove the `=` from the mentioned line and correct the `lst` part above `print('Purge....` to this:

    lst = lst[:-maxfiles] if maxfiles else lst

therefore:
* -1 → ignored
* 0 → all removed, new log dropped in the folder
* 1+ → count of _log_maxfiles_ remains in the folder + one new log dropped in

Also, for the documentation: instead of ["Note:"](https://github.com/kivy/kivy/pull/5067/files#diff-dc82eaf94f876f667675d3493cc95dd1R95) try to use e.g. [RST editor](http://rst.ninjs.org/) and add `.. note::` tag. [It'll look prettier](https://github.com/ErwinRieger/kivy/blob/04ff945b9764a0eefae77de30c2935206b5a2c0f/kivy/config.py#L168). That's probably everything before merge.

KeyWeeUsr:
Tested with those values, everything looks nice, thanks! 🍡

ErwinRieger:
Works for me too. Thanks.

PR #5075, opened 2017-03-21
Add link to examples wheel

### LINKED ISSUE #5008: Examples: Missing link after separation
### Versions

* OS: Windows+ ?

### Description

After the examples wheel separation, there's a missing link between the main wheel and the examples. This can be either solved as a Kivy wheel being dependent on the examples wheel, or with a mention in the documentation, though I'm not sure where.

There was a [`change`](https://github.com/kivy/kivy/blob/master/setup.py#L898) in the setup.py file, which means this might affect even PPA if it uses it and will affect everything that uses `pip install <url>`.

@matham ping 🐱

KeyWeeUsr:
Is this the good place for the link?

Closes #5008


matham:
Long term though this wheel should serve also linux etc. Because it's pure python an contains our example so it can be used on any os.

KeyWeeUsr:
Yeah, it could be reflected if e.g. ppa or homebrew packages (or wheels, someday hopefully) were changed to ignore the examples folder (it's default for win now, I think). Then we could probably move the link to the repo readme.

Issue #5040, opened 2017-03-05
Mesh triangle_fan doesn't render properly a big shape


KeyWeeUsr:
### Versions

* Python: 2, 3
* OS: Windows
* Kivy: master
* Kivy installation method: wheels

### Description

Line renders the points correctly, Mesh renders them too if drawn in a separate triangles, but when I try to render it as a single shape, for some reason it somehow "skips" the points. Test it with modifying the indices and you'll see when you get to the line-like shapes of the Mesh that it sees the points, but renders wrong thing. For example `indices: range(int(len(app.poly) / 12.0))`.

![image](https://cloud.githubusercontent.com/assets/8825439/23587617/b1e4786a-01b0-11e7-8f2f-a1f2d715407b.png)

![image](https://cloud.githubusercontent.com/assets/8825439/23587619/bc0d407e-01b0-11e7-822e-8309b5a5a90a.png)

### Code and Logs

```python
from kivy.app import App
from kivy.uix.floatlayout import FloatLayout
from kivy.lang import Builder
Builder.load_string('''
<Test>:
    Scatter:
        canvas:
            Color:
                rgb: 1, 0, 0
            Line:
                points: [p - 300 for p in app.poly]
    Scatter:
        canvas:
            Color:
                rgb: 1, 0, 0
            Mesh:
                vertices: [_ for p in [[app.poly[i] - 300, app.poly[i+1] - 300, 0, 0] for i in range(0, len(app.poly), 2)] for _ in p]
                indices: range(int(len(app.poly) / 2.0))
                mode: 'triangle_fan'
''')
class Test(FloatLayout): pass
class TestApp(App):
    poly = [
       857.4467, 594.8027, 861.5295, 595.6330, 870.0411, 604.6976,
       869.0722, 607.7422, 873.7778, 611.1328, 875.9230, 609.1261,
       882.3585, 610.7176, 888.5172, 607.8806, 896.4751, 612.5167,
       905.8095, 612.5172, 908.8542, 609.6482, 907.8003, 606.0179,
       910.8450, 600.2212, 908.6200, 596.3568, 915.0608, 595.3028,
       917.4614, 598.0548, 922.1457, 595.9469, 929.1134, 597.5278,
       929.4648, 602.0949, 932.0996, 603.6759, 932.5095, 611.1705,
       937.1352, 614.6251, 942.9904, 614.3324, 945.1682, 616.1405,
[...]

PR #5079, opened 2017-03-26
Add shape collision example

KeyWeeUsr:
Pretty much ~100 lines of commented stuff in there, but should give a really good look on how powerful and readable Kivy + Python is with multitouch and collisions. This example provides 3 classes:

* BaseShape, which is just an abstracted behavior for moving everything through `on_pos`
* RegularShape, a real, regular shape with a center == widget's center and increasing the amount of edges up to circle-like shape within the widget's size as a bounding box. Doing it manually was quicker than trying to use `Line` or `Ellipse` as I still needed to construct a `Mesh`, thus I'd end up with constructing twice that much stuff.
* MeshShape, a real, irregular and _convex_ shape based on an inputted array of points. The basic problem was that no editor is capable of creating a useful, out-of-box working set of points i.e. if I put there _raw_ points made with Pen in Gimp/Illustrator, it wouldn't display the points correctly, thus a correction with pushing the bottom-left part of the points (imagine it like a 2D bounding box) to the widget's origin was necessary

In a real application the `MeshShape` could be scaled to a capability of handling multiple subparts of a huge mesh like in my silly issue (#5040), therefore it wouldn't be convex-only, however that would take a long while to make it (if someone feels adventurous, though I might do it in future).

However, I can't make `BaseShape.move_collider()` to work, because for some reason the points are still the same. Probably I made some silly mistake somewhere, but I just don't see it. Help welcome? 😄

### DIFF
--- a/examples/miscellaneous/shapecollisions.py
+++ b/examples/miscellaneous/shapecollisions.py
@@ -168,7 +168,7 @@ class BaseShape(Widget):
 
         # ignore if no polygon area is set
         poly = self.poly
-        if not self.poly:
+        if not poly:
             return False
 
         n = self.poly_len
--- a/examples/miscellaneous/shapecollisions.py
+++ b/examples/miscellaneous/shapecollisions.py
@@ -10,7 +10,7 @@ from kivy.clock import Clock
 from kivy.uix.label import Label
 from kivy.uix.widget import Widget
 from kivy.core.window import Window
-from kivy.graphics import Color, Mesh, Line
+from kivy.graphics import Color, Mesh, Point
 from kivy.uix.floatlayout import FloatLayout
 from kivy.properties import (
     ListProperty,
@@ -79,6 +79,7 @@ class BaseShape(Widget):
         '''Move debug collider when the shape moves.
         '''
         points = self.debug_collider.points[:]
+
         for i in range(len(points)):
             if not i % 2:
                 points[i] += offset_x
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
For some reason `Line(circle=...)` won't move with touch, but classic `Line` or `Point` will. `Point` is prettier anyway:

image

matham:
I wonder if this may be too complex for an example. It has lots of parts with custom algorithms. But perhaps that's needed so people have examples of a more complex app? Also, we have garden.collider for non-standard shape collision detection.

KeyWeeUsr:
From algorithms, there's PIP only, the rest is Pythagorean theorem and playing with max/min (edges) of an array of points. It might seem maybe too much if you mean the mesh points correction, but that's kind of necessary and not too obvious, especially for the beginners (definitely wasn't obvious to me 😆 ) for creating a custom 2D shapes, which might be later used in e.g. kivent (didn't see anything like that in it).

If not, point me to the specific lines, maybe I might be able to simplify or remove it ^^

matham:
That was kind-of my point, an example should probably not include algorithms as that takes away a bit from the purpose of showing how to use kivy itself, rather than also having to understand algorithms used there, which at that point may belong on the wiki instead. However, it's probably fine for a few examples. Or maybe we should have a folder for more advanced examples.

That said I also don't us to be responsible for maintaining these more advanced examples, specifically by having people (or e.g. potential students) opening PRs to fix whatever subtlety you may have missed. But I guess that's not that strong of reason to reject this.

KeyWeeUsr:
Yeah, I agree the maintaining could lead to problems if someone decides to implement completely new stuff into the example, though if such thing would be rejected (I don't see the rest of the examples "reworked" anyway), then it might work i.e. fixes only.

Anyway, I'm also ok with the wiki if it's a better idea.

matham:
I think this is fine. Maybe just add a notice that it's just meant as a demo. Although I'm not really sure it's needed. Upto to you.

KeyWeeUsr:
I found a little bug with always true for `shape_collide` as `return` from a function means true, though otherwise it should be ok.

PR #5088, opened 2017-03-31
More auto scale options

Kovak:
This makes the auto_scale option more useful. Essentially it allows us to mix the behavior of an Image and a BorderImage when it comes to rendering, based on dimension as well as if the condition that the source image is larger than the area we are rendering into.

This let's us achieve effects such as a rounded button that scales across its width like a BorderImage but maintains the circular nature of its ends instead of having them become a square with rounded corners.

[Explanatory gif](https://drive.google.com/file/d/0B5Gkp8zHsG-PLTFiV3VQR0lOYU0/view?usp=sharing): This uses 'y_full_x_lower'

### DIFF
--- a/kivy/graphics/vertex_instructions.pyx
+++ b/kivy/graphics/vertex_instructions.pyx
@@ -851,8 +851,29 @@ cdef class BorderImage(Rectangle):
             Border information in the format (top, right, bottom, left).
             Each value is in pixels.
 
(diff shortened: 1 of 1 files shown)


dessant:
@Kovak, are you ok with merging?

Kovak:
I am, everything should be good on this one.

Issue #3910, opened 2016-01-12
The user's guide tells the user to install pygame


inclement:
https://kivy.org/docs/installation/installation.html

Julian-O:
Docs don't do that any more. Closing as Fixed.

PR #5097, opened 2017-04-03
doc: Mention pygame in installation dependencies

SanketDG:
Pygame is mostly required for window display in kivy.

Fixes https://github.com/kivy/kivy/issues/3910


KeyWeeUsr:
I believe the point of that issue is to replace Pygame stuff with SDL2 as it's _the_ Window provider used nowadays for Kivy and not to include a mention about Pygame in that particular place, because in the installation commands it's already used (`python-pygame`). Reopen if you think otherwise though.

SanketDG:
@KeeWeeUsr I see.

But I am surprised, that `pip` is not favoured in the documentation, instead distro packages are used which are often old(maybe not for pygame)

PR #5103, opened 2017-04-07
Update text_layout.pyx

ladyruby723:


KeyWeeUsr:
Hehe, tripping words, oh my... thanks for the catch :yum:

Issue #735, opened 2012-10-11
Is Chinese support in TextInput?


elaijuh:
Latest version still no support for IME? I am working on WIN7 not able to input Chinese. I wonder if it works on android? Is there any solution to this?

Issue #434, opened 2012-03-25
TextInput doesn't accept input in certain languages from the keyboard


akshayaurora:
To reproduce

1) launch any example that uses TextInput,
2) Change you keyboard Layout to a language like Chinese that require non standard input like IME
3) click to focus the TextInput Widget and try to type, nothing get's displayed.

yingshaoxo:
Actually interesting, I found that kivy `Textinput` component can't input any languages except English in android. (I'm using `Gboard` as my input method in android)

I was using `python-for-android` to build my application.

Here is the APK that I built, you can also found the source code here:
https://github.com/yingshaoxo/kivy-diary/releases

___

I have to say, at the linux desktop, It all works. I have no idea about how it can't be done with the android platfrom.

PR #5109, opened 2017-04-09
Add textedit event for text editing by IME

Adachinski:
Added code to return _windowSDL2Storage.poll () 'textedit' event to notify the text being edited when typing in IME.

In this code, It's get to editted text from the parameter of SDL_TextEditingEvent.
https://wiki.libsdl.org/Tutorials/TextInput

The objective is to make a basis for displaying text being edited in IME with Textinput (or its subclass).
(Improve the situation like #434 and #735)

*Sample code that dispatches textedit on WindowSDL
https://github.com/Adachinski/kivy/tree/textedit_patch_smple

*Sample code to display the text being edited on the label
https://github.com/Adachinski/kivy_textedit_sample

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -412,6 +412,9 @@ cdef class _WindowSDL2Storage:
         elif event.type == SDL_TEXTINPUT:
             s = event.text.text.decode('utf-8')
             return ('textinput', s)
+        elif event.type == SDL_TEXTEDITING:
+            s = event.edit.text.decode('utf-8')
+            return ('textedit', s)
         else:
             #    print('receive unknown sdl event', event.type)
             pass


wangeek:
We really need the `SDL_TEXTEDITING` event exposed, to support non-English language (Chinese, Japanese etc) editing. Currently Kivy is limited to "western" locales only. So I suggest this issue a higher priority.

KeyWeeUsr:
Thanks, sorry for the delay. Seems fine to me as the [`SDL_TEXTEDITING`](https://github.com/Adachinski/kivy/blob/43d0dd6c2c18f189ee1addd194553daa39ba2a04/kivy/lib/sdl2.pxi#L146) is already available in our `sdl2.pxi`.

wangeek:
@KeyWeeUsr Thanks for your work. 

Would you please merge the other commit ([TextInput Appendix](https://github.com/okajun35/kivy/commit/918fb0b8cc3b90d2c1bf9ad139f707ef022e6bb7#diff-5f05a9fe836a8283afe4a523b2038a32L611)) listed above from @okajun35  ? The original commit seems kinda incomplete, need a bit adjusting in `kivy/core/window/window_sdl2.py`.

Maybe we need another pull request to do that?

PR #5111, opened 2017-04-10
Add test for Fbo.get_pixel_color

KeyWeeUsr:
I'm not quite sure how does `nose` work with tests, but it works locally if you run it with python as a `main.py`. Let's see what Travis says about it.

Duration is ~0.05s which is probably acceptable if I create an app, bind to window and draw on canvas+fbo.

* `set_env()` is important to actually run anything after `App.run()` is executed
* `on_window_flip()` is important because `Fbo.get_pixel_color()` returns `(0, 0, 0, 0)` otherwise probably because there isn't anything on the fbo yet (probably not even on the canvas itself)


KeyWeeUsr:
Apparently it likes the test anyway although it's not a `TestCase`, but it failed with py2 (and would probably later) as it compares the rendered colors. Hm, fail... I'll find a better approach.

KeyWeeUsr:
* Passed: https://travis-ci.org/kivy/kivy/jobs/220459997#L1747
* Failed: https://travis-ci.org/kivy/kivy/jobs/220460002#L3760

The failed tests are on OSX, which I think [doesn't even create window environment](https://github.com/kivy/kivy/blob/master/.travis.yml#L86) in Travis, so it's quite obvious to expect this to fail. On the other hand... how to fix the test, so that it passes? :D

PR #5117, opened 2017-04-13
Fix issue when using multiple RecycleViews with different view classes.

charrea6:
This change fixes an issue where a class variable was being used across multiple instances of RecycleDataAdapter. This meant that stale data was being used on the first update of the next RecycleView to be created.

### DIFF
--- a/kivy/uix/recycleview/views.py
+++ b/kivy/uix/recycleview/views.py
@@ -139,10 +139,12 @@ class RecycleDataAdapter(EventDispatcher):
     with this instance.
     '''
 
-    # internals
-    views = {}  # current displayed items
-    # items whose attrs, except for pos/size is still accurate
-    dirty_views = defaultdict(dict)
+    def __init__(self):
+        super(RecycleDataAdapter, self).__init__()
+        # internals
+        self.views = {}  # current displayed items
+        # items whose attrs, except for pos/size is still accurate
+        self.dirty_views = defaultdict(dict)
 
     _sizing_attrs = {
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
@matham what do you think? It shouldn't need class variables, right?

kivy/uix/recycleview/views.py
```diff
@@ -139,10 +139,12 @@ class RecycleDataAdapter(EventDispatcher):
     with this instance.
     '''
 
-    # internals
-    views = {}  # current displayed items
-    # items whose attrs, except for pos/size is still accurate
-    dirty_views = defaultdict(dict)
```
matham: Don't delete the class declarations, it's required for documentation purposes. And it's always good to declare class variables in their class definitions.

kivy/uix/recycleview/views.py
```diff
@@ -139,10 +139,12 @@ class RecycleDataAdapter(EventDispatcher):
...
+    def __init__(self):
+        super(RecycleDataAdapter, self).__init__()
```
matham: This is missing kwargs propagation.

kivy/uix/recycleview/views.py
```diff
@@ -139,10 +139,12 @@ class RecycleDataAdapter(EventDispatcher):
...
+        # internals
+        self.views = {}  # current displayed items
```
matham: Rather than removing class declarations, keep them, but adding the per instance dicts like you did here should fix the original issue.

PR #5120, opened 2017-04-14
Fix py3 dict_keys in LoadIdentity getter

KeyWeeUsr:
For reference see https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects, this will clearly break on Py3 and will work correctly on Py2, thus converting the keys() to list first, then indexing.


KeyWeeUsr:
I put checks to Cython as @tito did in https://github.com/kivy/kivy/commit/a5716c464953ba463cbae2b1d03bb247fe2a101f and the rest is like in the whole repo i.e. calling `list(dict_keys)` anyway for both versions.

PR #5126, opened 2017-04-17
Fix modifier key codes for single

### LINKED ISSUE #5062: Press on key numpad_0 passes 'alt' in modifiers list
### Versions

* Python: 2.7 
* OS: Windows7 
* Kivy: v1.9.2.dev0
* Kivy installation method: pip and nightly wheel - git-3e1ea73, 20170315

### Description
When pressing numpad_0 key, 'alt' is passed in modifiers list. 
Single press on numpad_0 prints: 
```('Scancode:', 98L, 'Keycode:', 256, 'Modifiers:', ['alt'])```
but it should print:
```('Scancode:', 98L, 'Keycode:', 256, 'Modifiers:', [])```

This is also present on Ubuntu and Windows7 with Kivy 1.9.1.
Issue could be connected with https://github.com/kivy/kivy/issues/4008

### Code and Logs

Example code to reproduce:
```python
from kivy.app import App


class KeyboardApp(App):

    def build(self):
        Window.bind(on_key_down=self._on_keyboard_down)

    def _on_keyboard_down(self, window, keycode, scancode, codepoint=None,
                          modifiers=None, **kwargs):
        print('Scancode:', scancode, 'Keycode:', keycode, 'Modifiers:', modifiers)


if __name__ == '__main__':
    KeyboardApp().run()
```

manthansharma:
Single pressed modifiers key are now include in self._modifiers and also update super to `lmeta` and `rmeta` to support cmd key in OSX.

Source for SDL Keycode: https://wiki.libsdl.org/SDLKeycodeLookup
Source for key_map code: https://gist.github.com/manthansharma/fa7c0483779519945583c46be42be60a

Output:
`('The key', (304, 'shift'), 'have been pressed')
 - text is u'\u0130'
 - modifiers are ['shift']

('The key', (305, 'lctrl'), 'have been pressed')
 - text is u'\u0131'
 - modifiers are ['ctrl']

('The key', (256, 'numpad0'), 'have been pressed')
 - text is u'\u0100'
 - modifiers are []
`

Fixes #5062


kivy/core/window/window_sdl2.py
```diff
@@ -691,15 +698,14 @@ def _update_modifiers(self, mods=None, key=None):
                 modifiers.add('ctrl')
             if mods & (KMOD_RMETA | KMOD_LMETA):
                 modifiers.add('meta')
-
         if key is not None:
-            if key in (KMOD_RSHIFT, KMOD_LSHIFT):
+            if key in (self.key_map[SDLK_SHIFTL], self.key_map[SDLK_SHIFTL]):
```
KeyWeeUsr: right shift?

manthansharma:
@KeyWeeUsr made the requested changes and isn't pygame deprecated for kivy or have to made same changes for pygame window provider.

KeyWeeUsr:
Well, yes and no. Pygame still works although if it's used in p4a, it's mostly used with stable, but it should work even with master if I'm right therefore changes in this part might break it completely (no input=no fun). That's why the `needs-testing` tag. 😄 

If someone confirms we don't care about Pygame anymore, we'd just need to check the `meta` stuff with iOS/OSX as that changed, but let's say it works as you've already tested. Although I quite wonder, why `temp_key` instead of `key`?

manthansharma:
Ok, will wait for pygame confirmation and

`temp_key` helps convert numpad keycode to of 0-9 keys and as `key` is return in later part. So, can't replace key with new value.

KeyWeeUsr:
```
The key (264, 'numpad8') have been pressed
 - text is '8' <- +1, good
 - modifiers are []
The key (265, 'numpad9') have been pressed
 - text is None <- only this one seems left out, hm?
 - modifiers are []
The key (319, 'menu') have been pressed <- nice!
 - text is None
 - modifiers are []

The key (266, 'numpaddecimal') have been pressed
 - text is None
 - modifiers are []
The key (267, 'numpaddivide') have been pressed
 - text is None
 - modifiers are []
The key (268, 'numpadmul') have been pressed
 - text is None
 - modifiers are []
The key (269, 'numpadsubstract') have been pressed
 - text is None
 - modifiers are []
The key (270, 'numpadadd') have been pressed
 - text is None
 - modifiers are []
The key (266, 'numpaddecimal') have been pressed
 - text is None
 - modifiers are []
```

Please check the `numpad9` as for what happened with it. Also, if you're willing to take it further, you might be able to add the appropriate symbols for numpad keys i.e. `/*-+,` during the fixing of `numpad9`.

kivy/core/window/__init__.py
```diff
@@ -62,11 +62,12 @@ class Keyboard(EventDispatcher):
         # specials keys
         'backspace': 8, 'tab': 9, 'enter': 13, 'rshift': 303, 'shift': 304,
         'alt': 308, 'rctrl': 306, 'lctrl': 305,
-        'super': 309, 'alt-gr': 307, 'compose': 311, 'pipe': 310,
+        'lmeta': 309, 'rmeta': 310, 'alt-gr': 307, 'compose': 311,
```
KeyWeeUsr: If by `rmeta` you mean the right `alt` key i.e. "Alt Gr", then it's already there as `alt-gr` and therefore it's already there and you just broke `pipe` key name. About the `super` → `lmeta` change, this will break the key for everyone who used `if key == 'super'` as people most likely won't use the `int` value directly.

If you didn't mean by `rmeta` the "Alt Gr" key, please show how did you get it (pic of keyboard key or something), because when I press "Alt Gr", it returns `alt-gr` and its values, not `rmeta`.

kivy/core/window/window_sdl2.py
```diff
@@ -15,21 +15,22 @@
...
 from kivy.core.window import WindowBase
 from kivy.core.window._window_sdl2 import _WindowSDL2Storage
-from kivy.input.provider import MotionEventProvider
 from kivy.input.motionevent import MotionEvent
+from kivy.input.provider import MotionEventProvider
+from kivy.logger import Logger
 from kivy.resources import resource_find
 from kivy.utils import platform, deprecated
-from kivy.compat import unichr
-from collections import deque
```
KeyWeeUsr: Why the sudden shifting of lines? It's not like we need them in that specific order(please revert the change). 😋

manthansharma: Sorry, for this it happens due to auto formatting in pycharm ide.

manthansharma:
@KeyWeeUsr `rmeta` and `lmeta` are the commands keys is mac keyboard and 'super' key in windows keyboard.




and there is no such keys as `pipe` in both **pygame** constants and **sdl2** and in both `310` references as `meta` or `gui`.

KeyWeeUsr:
I feel rather puzzled by that particular change and although I felt like preserving `super` might be a good idea, it appears that the problematic keys i.e.:

* super
* pipe

were introduced not that long before `1.9.0`(#3109) i.e. they were introduced in that stable which could work for us quite well in renaming them to [their _proper_ names](https://www.libsdl.org/release/SDL-1.2.15/docs/html/sdlkey.html)(can't find sdl2) as the current stable is `1.9.1`.

If we were to replace the keys, [this file](https://github.com/kivy/kivy/blob/5913d861f45119dfff0b6cb5d004e6f79858563c/kivy/input/providers/hidinput.py#L202) would need a change as the keys were introduced together with hidinput stuff, however the problem here is [`super`](https://github.com/kivy/kivy/search?utf8=%E2%9C%93&q=super&type=) as it's really unfortunate naming colliding with Python language.

@tito should we go for the proper naming or for backwards compatibility if it's not that long since the keys have been introduced (~1.7y)?

Unfortunately we can't have both as using the same name for a single numeric value will return only a single value with [keycode_to_string](https://github.com/kivy/kivy/blob/master/kivy/core/window/__init__.py#L187) . There are keyboards with two windows keys, so something has to be done. Maybe we could name them `super` and `super-right`( 😕 )  ?

manthansharma:
@KeyWeeUsr, what's need to be updated in this PR and sorry, for leaving this PR in between.

KeyWeeUsr:
@manthansharma I checked it again, the Pygame part should be safe enough as the keycode is passed to our Window, therefore we are creating the key labels (`super`, `meta`, etc...). About the naming, well... let's go with the correct names (`lmeta`, `rmeta`, etc...) and label this PR as an API-break. It fixes way more than it breaks, so we shouldn't try hard to keep everything compatible especially if it's just two strings.

I don't like the try-except as you already know it will be a `KeyError`. Use `if <key> in <map>` instead:

                try:
                    temp_key = self.convert_key_map[key]
                except KeyError:
                    temp_key = key

What **needs** to be changed to reflect the changes is:
https://github.com/kivy/kivy/blob/c93841d6c66626fbdfd5c963b440c817c4b65eb4/kivy/input/providers/hidinput.py#L202

        0x56: ('pipe', ),           # this
        0x1d: ('ctrl', ),
        0x7D: ('super', ),          # this
        0x38: ('alt', ),
        0x39: ('spacebar', ),
        0x64: ('alt-gr', ),
        0x7e: ('super', ),          # this
        0x7f: ('compose', ),

PR #5134, opened 2017-04-22
Update readme for building docs

tony:
Fix indentation / code blocks. Tabs to spaces. Refactor sections.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -2,47 +2,48 @@ Kivy - Documentation
 ====================
 
 You can access the latest documentation on the web:
-  * http://kivy.org/docs
 
+* http://kivy.org/docs
 
-How to build the documentation
-------------------------------
+Contributing
+------------
 
-You need to install:
+If you intend on editing and contributing documentation, assure the kivy source
+code is up to date before proceeding. If your documentation is outdated, it
+could result in merge conflicts.
 
-  * Python Sphinx
-    - With pip:
-    ```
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
The point of the weird-ish headers' formatting is because we use [reStructuredText](http://docutils.sourceforge.net/docs/user/rst/quickref.html) which works both ways as Markdown and as ReST which is used even in the main README.md in the root of the docs. Please preserve it.

The rest of the changes is good, so if you could rewrite it with preserved syntax(you can test on [ninjs](http://rst.ninjs.org/)), I'll merge it. ^^

tony:
@KeyWeeUsr better?

doc/README.md
```diff
@@ -2,47 +2,42 @@ Kivy - Documentation
...
+  ``pip install sphinxcontrib-actdiag sphinxcontrib-nwdiag``
+    
+  Or just use the provided *doc-requirements.txt*:
+    
+  ``pip install -r doc-requirements.txt``
+  
+Building the documentation
+--------------------------
 
-  * Latest kivy
```
KeyWeeUsr: This part is essential when building the docs when a user wants to contribute, otherwise it may result in outdated parts or even collisions.

tony:
@KeyWeeUsr better?

KeyWeeUsr:
Thanks! 👋

PR #5135, opened 2017-04-22
Docs: Add py3 object.__init__() reference to properties

KeyWeeUsr:
I'm not sure if it's _really_ needed, but it represents a feature Kivy properties silently do and it can help beginners understand how Kivy properties work _the Python way_ i.e. not just "It works(tm)" part of the codebase.

As a bonus we'd have a reference to the Py3 error explained even here which is probably a good to have part of the docs as I've seen it mentioned even on #kivy.


Zen-CODE:
We hit this issue big time when starting with 3. I think it's really valuable to give concrete examples of how to handle this one. Nice one :+1:

PR #5145, opened 2017-05-01
Add Slider.sensitivity

KeyWeeUsr:
Bases on #3275 idea although I'm not quite sure if this is what @tito meant. On the other hand,
if it's meant like there should be something more tweakable in python so that kv can change it directly (i.e. not via `sensitivity: 'cursor'` only) then there could be e.g. this:

    #py:
    sensitivity = auto/cursor/manual(?)
    cursor_collide_area = ListProperty([?])
    elif self.sensitivity == 'cursor':
        root.cursor_collide_area # collide with that & grab

    #kv:
    Image:
        on_pos: root.cursor_collide_area = [*self.pos, *self.size] if root.sensitivity == 'cursor' else root.cursor_collide_area

    # widget kv:
    Slider:
        sensitivity: 'manual'
        cursor_collide_area: [0, 0, 400, 300] # half-window from origin

but that seems a little bit uglier (two properties for a single feature) although it seems to be a quite usable solution too (and probably even "too tweakable").

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -67,10 +67,10 @@
             points: self.center_x, self.y + self.padding, self.center_x, self.value_pos[1]
         Color:
             rgb: 1, 1, 1
-        Rectangle:
-            pos: (self.value_pos[0] - self.cursor_width / 2, self.center_y - self.cursor_height / 2) if self.orientation == 'horizontal' else (self.center_x - self.cursor_width / 2, self.value_pos[1] - self.cursor_height / 2)
-            size: self.cursor_size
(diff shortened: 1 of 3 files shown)


dessant:
We could use `handle` in place of `cursor`.

KeyWeeUsr:
The unittest for the default and `sensitivity='handle'` behavior works correctly I believe both locally and on Travis+Appveyor. The rest of the functionality is basically what was requested in the previous issue/PR so this should be ok to merge as it works with widgets only (not with canvas instructions directly).

Any requested changes?

dessant:
Versionadded syntax, the rest looks good, though I have not tested.

Issue #5146, opened 2017-05-01
CameraOpenCV class doesn't invert the camera's inherent frame rate


jkankiewicz:
### Versions

* Python: v2.7.13
* OS: Fedora 25
* Kivy: v1.9.2.dev0, git-ead4655, 20170501
* Kivy installation method: python setup.py install

### Description

I'm trying to write an augmented reality application but the frame rate of the kivy.uix.camera.Camera class is the inverse of the OpenCV camera provider's default frame rate e.g. on my computer, it captures only one texture every 30 seconds instead of one texture every 30th of a second. This is obviously much too slow to be usable for an augmented reality application.

### Code and Logs

```python
from kivy.config import Config
Config.set('kivy', 'log_level', 'debug')

from kivy.app import App
from kivy.core.window import Window
from kivy.uix.camera import Camera

class TestApp(App):
    def build(self):
        return Camera(size=Window.size, resolution=Window.size, play=True)

if __name__ == '__main__':
    TestApp().run()
```

dessant:
Here is a summary of the issue: https://github.com/kivy/kivy/pull/4828#issuecomment-283382305.

PR #5147, opened 2017-05-01
#5146: Invert the OpenCV camera's inherent frame rate.

jkankiewicz:


dessant:
This change may fix your use case, but it breaks others, please check https://github.com/kivy/kivy/issues/5146#issuecomment-299722233. The camera driver could also return `0.03333333333333333` as fps.

Issue #5083, opened 2017-03-27
Mtdev on touch down does not rotate coordinates after graphics rotation=90 (in portrait mode) [Graphics][core][platform-rpi]


audryste:
### Versions

Kivy      v1.9.2-dev0
Python v2.7.9 (default, Sep 17 2016, 20:26:04)
OS: pipaOS, RPi
Kivy installation method: apt-get install python2-kivypie

### Description
on_touch_down  does not rotate coordinates after setting [graphics] rotation=90 (portrait mode)

I simply added logging on mtdev provider and touch tracer example and I got following.

I used IR touchscreen. I cannot use rpi display rotate as it has impact on performance.

### Code and Logs

```
<MTDMotionEvent id=493 pos=(0.649190, 0.536790) device=zztouch>
 Touchtracer on_touch_down **701.124912259 1030.63692129**

 <MTDMotionEvent id=493 pos=(0.650685, 0.530229) device=zztouch>
 Touchtracer on_touch_move **572.646870327 670.684530168**

```
This one is correct:  Touchtracer on_touch_move **572.646870327 670.684530168**

Any ideas how to fix it? Only on touch down coordinates are not rotated.

audryste:
Found the problem seems modules/touchring.py messes it up:
you need to remove //comment out this line 60: 
#touch.scale_for_screen(win.width, win.height)

audryste:
Can anyone apply the changes for this bug?
file:
https://github.com/kivy/kivy/blob/master/kivy/modules/touchring.py

function:
def _touch_down(win, touch):

line: touch.scale_for_screen(win.width, win.height)

GaurangPoolCar:
Hi @audryste 

I am also having same problem. I really love Kivy, we want to use Kivy with raspberry PI as Kiosk app. 

When I rotate screen to 270 , coordinates are not rotating, but on startX coordinates match with touch.

PR #5149, opened 2017-05-01
Master

audryste:
https://github.com/kivy/kivy/issues/5083


KeyWeeUsr:
`merge 1,506 commits into stable from master` eh, what?

PR #5152, opened 2017-05-02
Fixed crash on python3, due to items not being subscriptable

JackED42:
I found a crash on Kivy with Python3 when attempting to paste into a TextField. This change seems to fix the problem.

Thanks

Jack


KeyWeeUsr:
Nice, thanks! I didn't notice this one at all when I was look for similar bugs. Also, be sure to check the Kivy+SDL2 as we're slowly leaving Pygame with SDL1 alone and use only SDL2 directly even on mobile platforms.

PR #5153, opened 2017-05-02
Fix KV include for quoted paths

KeyWeeUsr:
This should be a nice change. Generally, it works even _without_ this patch if a user leaves out the quotes, but it basically looks like this:

    #:include [force] C:\\my path
    ------------------------^ wat? works because of eval, but otherwise makes no sense

It's rather expected from a user to think that "yay, it uses some arg parser, I'll quote it just to be safe"

    #:include [force] "C:\\my path"
    ------------------^-----------^ well, nope

On Windows you can't have a file with quotes in it, on Mac I think it's forbidden too, but GNU/Linux distros can do "fancy" stuff even without it, so... I have 3 alternatives:

* let's play on lisp -> a single quote on ref[0] is enough to trigger stripping everything
* check if it ends with the same quotes and then strip the quotes anyway
* check for ref[0] and ref[-1], then return ref[1:-1] (strip 2 chars)
<br>

    1 #:include [force] "C:\\my path
    2 #:include [force] """"C:\\my path""""""
    3 #:include [force] "C:\\my path" # no multiple quotes allowed

### DIFF
--- a/kivy/lang/parser.py
+++ b/kivy/lang/parser.py
@@ -422,6 +422,10 @@ class Parser(object):
                     ref = ref[6:].strip()
                     force_load = True
 
+                # if #:include [force] "path with quotes around"
+                if ref.startswith('"') or ref.startswith("'"):
+                    ref = ref.strip("'").strip('"')
+
                 if ref[-3:] != '.kv':
                     Logger.warn('Lang: {0} does not have a valid Kivy'
                                 'Language extension (.kv)'.format(ref))
--- a/kivy/lang/parser.py
+++ b/kivy/lang/parser.py
@@ -423,7 +423,7 @@ class Parser(object):
                     force_load = True
 
                 # if #:include [force] "path with quotes around"
-                if ref.startswith('"') or ref.startswith("'"):
+                if ref[0] == ref[-1] and ref[0] in ('"', "'"):
                     ref = ref.strip("'").strip('"')
 
                 if ref[-3:] != '.kv':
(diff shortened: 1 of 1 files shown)


kivy/lang/parser.py
```diff
@@ -422,6 +422,10 @@ def execute_directives(self):
                     ref = ref[6:].strip()
                     force_load = True
 
+                # if #:include [force] "path with quotes around"
+                if ref[0] == ref[-1] and ref[0] in ('"', "'"):
+                    ref = ref[1:-1]
```
matham: What about triple quotes :D
Although to be honest supporting triple quotes is probably not that important.

KeyWeeUsr: well, exactly that's why the three alternatives i.e. the second one (ccb1f7d) doesn't care about how many of them you have and just strips all, but anyway, triple quotes rather make sense when making strings multiline which isn't possible with #:include as the parser goes line-by-line so unless we'd go for a completely new way of handling directives, I don't see it hapenning otherwise than just this:

    #:include [force] """C:\my path"""

and slicing them off. That seems rather ugly though.

matham: But what was wrong with https://github.com/kivy/kivy/pull/5153/commits/ccb1f7da31e79856e3fb0da4706305857973cea4 ? That seems like the perfect solution. Triple quotes, although ugly is allowed on one line which that commit handles well, but not the latest commit.

KeyWeeUsr: I did the commits as 3 alternatives i.e. to choose from. I thought it was wrong because it allows multiple quotes without handling the amount of quotes _and_ if they are closed, thus it might cause confusion as for why does it even work if someone forgets to add a quote and it'd be this:

    #:include [force] """C:\my path""

KeyWeeUsr:
Now it strips only the python supported quotes 👓 

    "path"
    ""path"" <- still annoyed me and I couldn't find a smart way to remove it :D
    """path"""

Issue #3261, opened 2015-04-09
reStructuredText links not working with Kivy v1.9.1-dev


Davideddu:
Using RST implicit links to headings doesn't work with Kivy's RST implementation. The links are colored in orange as usual, but clicking on a link does nothing.

Here is a sample (working) RST document with implicit links: https://gist.github.com/Davideddu/c8fdede9b98597972923
(note that they work on Gist)

Test it with kivy/examples/RST_Editor and Kivy master.

PR #5165, opened 2017-05-10
Trying to allow external URLs on anchors on RstDocument using webbrowser

chirale:
Adding external urls support to RstDocument as a proof of concept, to be fixed. By now it's based on label so if two anchors with the same name are active the browser will open the same page (the first) for both. Some good ideas are needed to make it works properly.

### DIFF
--- a/kivy/uix/rst.py
+++ b/kivy/uix/rst.py
@@ -83,6 +83,7 @@ from docutils.parsers.rst import roles
 from docutils import nodes, frontend, utils
 from docutils.parsers.rst import Directive, directives
 from docutils.parsers.rst.roles import set_classes
+import webbrowser
 
 
 #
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
ref #3261

kivy/uix/rst.py
```diff
@@ -645,9 +646,14 @@ def goto(self, ref, *largs):
...
         ax = ay = None
         for node in self.anchors_widgets:
-            if ref in node.anchors:
-                ax, ay = node.anchors[ref]
-                break
+            try:
+                if ref in node.anchors:
+                    ax, ay = node.anchors[ref]
+                    break
+            except AttributeError:
```
KeyWeeUsr: Maybe switching to something more precise such as [`hasattr`](https://docs.python.org/3.6/library/functions.html#hasattr) would be more efficient. It'll even make the code shorter (because `break`). Just a suggestion ^^

Issue #1287, opened 2013-06-13
reStructuredText substitution doesn`t work in RstDocument


lukaszkoziara:
Substitution doesn`t work in text in RstDocument

I try use substitution for non breaking spaces in RstDocument: 

```
self.add_widget(AnswerRstDocument(text="""
.. |nbsp| unicode:: 0xA0
   :trim:
.. |test| replace:: tttt

some text some |test| text |nbsp| some text some text
""", size_hint_x=10, bar_width = 0))
```

but the following example is render to:

``````
ttttsome text some test text nbsp some text some text```
``````
##

KeyWeeUsr:
Added in 5167

PR #5167, opened 2017-05-10
Add support for RST replace

KeyWeeUsr:
Adds support for `.. |ref| replace:: something` RST syntax, test with:

```python
from kivy.app import App
from kivy.lang import Builder
class My(App):
    def build(self):
        self.string = '''
.. |uni| unicode:: 0xe4
.. |nbsp| unicode:: 0xA0
.. |text| replace:: is
.. |hop| replace:: replaced
.. _hop: https://github.com

Replace this -> |uni| |nbsp| |text| |hop|_ <- that
'''
        return Builder.load_string('''
RstDocument:
    text: app.string
''')
My().run()
```

ref #1287

### DIFF
--- a/kivy/uix/rst.py
+++ b/kivy/uix/rst.py
@@ -839,6 +839,7 @@ class _Visitor(nodes.NodeVisitor):
         self.text_have_anchor = False
         self.section = 0
         self.do_strip_text = False
+        self.substitution = {}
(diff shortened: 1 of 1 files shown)


tito:
OK for me, not tested, but code is OK. Feel free to merge

PR #5169, opened 2017-05-11
Window Opacity Feature

them0ntem:
Get/Set the opacity for a Window.

Added two new functions:
1. SDL_SetWindowOpacity: https://wiki.libsdl.org/SDL_SetWindowOpacity
2. SDL_GetWindowOpacity: https://wiki.libsdl.org/SDL_GetWindowOpacity

### Screenshot from OSX
screen shot 2017-07-12 at 3 32 54 am

- [x] SDL2
- [x] WindowsX11
- [x] Example

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -796,14 +796,16 @@ class WindowBase(EventDispatcher):
     '''
 
     opacity = AliasProperty(_get_opacity, _set_opacity)
-    '''Opacity for a window.
+    '''Opacity of the window. Accepts a value between 0.0 (transparent) and
+    1.0 (opaque).
 
-    .. note:: It's an SDL2 property with value between 0.0(transparent),
-     1.0(opaque).
+    .. note::
+        This feature requires the SDL2 window provider.
 
     .. versionadded:: 1.10.1
 
-    :attr:`opacity` is an :class:`~kivy.properties.AliasProperty`.
+    :attr:`opacity` is an :class:`~kivy.properties.AliasProperty` and defaults
+    to `1.0`.
     '''
 
     @property
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -1082,6 +1082,10 @@ class WindowBase(EventDispatcher):
         '''Set the opacity for a window'''
         pass
 
+    def get_window_opacity(self):
+        '''Get the opacity for a window'''
+        pass
+
     def _update_childsize(self, instance, value):
         self.update_childsize([instance])
 
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -1078,6 +1078,10 @@ class WindowBase(EventDispatcher):
         '''Flip between buffers'''
         pass
 
+    def set_window_opacity(self, opacity):
+        '''Set the opacity for a window'''
+        pass
+
     def _update_childsize(self, instance, value):
         self.update_childsize([instance])
 
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -755,6 +755,18 @@ class WindowBase(EventDispatcher):
         pos = self._get_window_pos()
         self._set_window_pos(pos[0], value)
 
+    def _get_window_opacity(self):
+        pass
+
+    def _set_window_opacity(self, opacity):
+        pass
+
+    def _get_opacity(self):
+        return self._get_window_opacity()
+
+    def _set_opacity(self, opacity):
+        self._set_window_opacity(opacity=opacity)
+
     top = AliasProperty(_get_top, _set_top)
     '''Top position of the window.
 
@@ -781,6 +793,17 @@ class WindowBase(EventDispatcher):
 
     :attr:`left` is an :class:`~kivy.properties.AliasProperty` and defaults to
     position set in :class:`~kivy.config.Config`.
+    '''
+
+    opacity = AliasProperty(_get_opacity, _set_opacity)
+    '''Opacity for a window.
+
+    .. note:: It's an SDL2 property with value between 0.0(transparent),
+     1.0(opaque).
+
+    .. versionadded:: 1.10.1
+
+    :attr:`opacity` is an :class:`~kivy.properties.AliasProperty`.
     '''
 
     @property
@@ -1076,14 +1099,6 @@ class WindowBase(EventDispatcher):
 
     def flip(self):
         '''Flip between buffers'''
-        pass
-
-    def set_window_opacity(self, opacity):
-        '''Set the opacity for a window'''
-        pass
-
-    def get_window_opacity(self):
-        '''Get the opacity for a window'''
         pass
 
     def _update_childsize(self, instance, value):
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -419,6 +419,14 @@ cdef class _WindowSDL2Storage:
     def flip(self):
         SDL_GL_SwapWindow(self.win)
 
+    def set_window_opacity(self, opacity):
+        return SDL_SetWindowOpacity(self.win, opacity)
+
+    def get_window_opacity(self):
+        cdef float opacity
+        SDL_GetWindowOpacity(self.win, &opacity)
+        return opacity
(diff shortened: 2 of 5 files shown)


kivy/core/window/__init__.py
```diff
@@ -1078,6 +1078,14 @@ def flip(self):
         '''Flip between buffers'''
         pass
 
+    def set_window_opacity(self, opacity):
```
KeyWeeUsr: This could be an `AliasProperty` named `opacity`, which would be definitely prettier than having setter and getter + it doesn't match the rest of the file. See [this](https://github.com/kivy/kivy/blob/master/kivy/core/window/__init__.py#L758).

kivy/core/window/_window_sdl2.pyx
```diff
@@ -419,6 +419,14 @@ cdef class _WindowSDL2Storage:
     def flip(self):
         SDL_GL_SwapWindow(self.win)
 
+    def set_window_opacity(self, opacity):
+        return SDL_SetWindowOpacity(self.win, opacity)
```
KeyWeeUsr: There's no need for a return in this function and it might even return garbage or crash the app with SDL2 returns such as [this](https://hg.libsdl.org/SDL/file/b6315b6bc32f/src/video/SDL_video.c#l2264).

kivy/core/window/window_sdl2.py
```diff
@@ -392,6 +392,12 @@ def _get_window_pos(self):
     def _set_window_pos(self, x, y):
         self._win.set_window_pos(x, y)
 
+    def set_window_opacity(self, opacity):
+        return self._win.set_window_opacity(opacity)
```
KeyWeeUsr: Again unnecessary `return`. Also see how the rest of such functions are named and used in the file.

dessant:
Thanks for the new feature. 👍 
Having a Window.opacity property may be a better fit, with a getter and setter, which in turn call the sdl functions.

dessant:
I'd vote for even removing the example from this pr, it's way more complicated than it needs to be for this simple api endpoint.

rnixx:
or simplify the example ;)

kivy/core/window/__init__.py
```diff
@@ -783,6 +795,17 @@ def _set_top(self, value):
     position set in :class:`~kivy.config.Config`.
     '''
 
+    opacity = AliasProperty(_get_opacity, _set_opacity)
+    '''Opacity for a window.
+
+    .. note:: It's an SDL2 property with value between 0.0(transparent),
+     1.0(opaque).
+
+    .. versionadded:: 1.9.1
```
rnixx: 1.10.0 is already released.

manthansharma: to which should i set

rnixx: 1.10.1 I guess

kivy/lib/sdl2.pxi
```diff
@@ -586,6 +586,8 @@ cdef extern from "SDL.h":
     cdef SDL_bool SDL_HasScreenKeyboardSupport()
     cdef SDL_bool SDL_IsScreenKeyboardShown(SDL_Window *window)
     cdef void SDL_GL_GetDrawableSize(SDL_Window *window, int *w, int *h)
+    cdef int SDL_SetWindowOpacity(SDL_Window * window, float opacity)
```
KeyWeeUsr: I'm not sure how this even compiles for you as the calls don't appear in the [SDL.h](https://hg.libsdl.org/SDL/file/77da466731bc/include/SDL.h) file (you use `cdef extern from "SDL.h":`)and that's probably why gcc doesn't like it ([implicit function declaration](https://travis-ci.org/kivy/kivy/jobs/231139054#L1439)). Try to pull the calls from [SDL_video.h](https://hg.libsdl.org/SDL/file/77da466731bc/include/SDL_video.h) instead.

KeyWeeUsr:
Also, I agree with dessant on this example, it's just too complicated for an incrementing of a variable. Basically you could bind a `Slider` value as a root widget to the `Window.opacity` with min and max values of the `Window.opacity` and that's pretty much it. No labels or anything necessary and it probably wouldn't even require Python code (just kv).

So, either remove the unnecessary stuff from it, or remove the whole example because it pushes attention away from the real purpose of the example with all str to float conversion and other stuff. :)

manthansharma:
@KeyWeeUsr facing same problem of ([implicit function declaration](https://travis-ci.org/kivy/kivy/jobs/231543798#L1274-L1280)) after moving calls from `SDL.h` to `SDL_video.h`

KeyWeeUsr:
@manthansharma I hoped it'd work as the declaration is in the video header, but apparently it doesn't like it anyway. Well, to debug this use some GNU/Linux distro or [mingwpy](https://anaconda.org/carlkl/mingwpy) on Windows with GCC.

KeyWeeUsr:
Ah, of course it raises the decl error! The symbol **isn't there at all**(like anywhere), because the build used [SDL < `2.0.5`](https://travis-ci.org/kivy/kivy/jobs/231139054#L487) and in the docs there's [this pretty little line](https://wiki.libsdl.org/SDL_SetWindowOpacity):

> This function is available since SDL 2.0.5

So, theoretically linux build should be fine, but the problem here is [this repo](https://launchpad.net/~zoogie/+archive/ubuntu/sdl2-snapshots) which we basically use for SDL2 builds + there's even an ordinary [Ubuntu repo](https://packages.ubuntu.com/yakkety/libsdl2-2.0-0) that also has `< 2.0.5`.

Maybe we should just make our own SDL2 snapshots and stop depending on others especially with such a critical dependency? The latest SDL2 snapshot is really old (3yo) and the latest stable SDL2 is 2.0.5 which is almost 1yo. Also, there's rather inconsistency with us using different versions for installation and for test builds.

Here is [the source for the launchpad repo](https://github.com/ZoogieZork/SDL2-Package) we use in Travis and as it's just MIT, we might be able to just make a new account and bump the versions (if we're lucky).

manthansharma:
Oh, and I was trying to debug this issue on latest build, and can't we use the latest sdl2 build directly like `OSX` https://github.com/kivy/kivy/blob/master/.travis.yml#L57-L71

KeyWeeUsr:
The "why" is mentioned somewhere after [this](https://botbot.me/freenode/kivy-dev/2017-06-30/?msg=87988595&page=1) line in the logs. Basically, each distro has it's own way of packaging (and custom interval of upgrading own package repos) which we use with the linux build. The OSX is used that way only because SDL2 released the package for OSX.

What we "might" do is to compile SDL2 from released source for our Travis builds, which will probably remove the implicit declaration issue, however we can't expect users to compile from source on their distro. It's almost the same level of silliness like expecting Windows users to compile from source, but on GNU/Linux the compiler actually likes you. :D

I'll try check it again on Windows(someone could try OSX) and if there's no problem, we'll need to wait for the linux to either upgrade their packages to a required version, or figure out something else.

manthansharma:
@KeyWeeUsr Thanks for the logs, helps in understanding `why`.

I have tested the feature in OSX, I need someone to test it on windows and what about WindowsX11 how to proceed with that?

KeyWeeUsr:
@manthansharma if you can, [compile SDL2](https://wiki.libsdl.org/Installation) from [release source](https://libsdl.org/release/SDL2-2.0.5.tar.gz) with [these deps](https://hg.libsdl.org/SDL/file/default/docs/README-linux.md) and see if it still works.

manthansharma:
@KeyWeeUsr do I need to compile for OSX or linux? as I have access to Mac only for now.

KeyWeeUsr:
Ah, ok, someone else will need to compile for linux to try if the code change works then :)

manthansharma:
I can try compiling on Linux on VM but will take some time as i need to download the distro for linux.

KeyWeeUsr:
Screenshot for linux from @manthansharma :
image

mak8kammerer:
Good day!
I am ready to continue working on this feature using the developments from this PR. Should I open a new PR?

misl6:
Hi @mak8kammerer !

Yep, feel free to open a new PR, but please include a comment regarding this one, so we can close it if yours get merged.

misl6:
Replaced by #8296

Issue #5179, opened 2017-05-15
Redisstore working only in Python 2.7


gregers666:
### Versions

* Python: 3.5
* OS: Mint 18
* Kivy: 1.10
* Kivy installation method: pip

### Description

redisstore.py working only in Python 2.7 due to encoding problems.

There is suggested solution attached.

Edit: Attached zip removed.

dessant:
Please point to the line which causes the issue and paste the fixed line here. Also share a traceback of the error you get.

gregers666:
diff redisstore_py2.py redisstore_py3.py 
63a64
>         newresult={}
69,70c70,73
<             result[k] = loads(result[k])
<         return result
---
>             newk = k.decode()
>             newresult[newk] = loads(result[k].decode())
>         del(result)
>         return newresult
76c79
<         for k, v in values.iteritems():
---
>         for k, v in values.items():
81,86d83
<     def store_delete(self, key):
<         key = self.prefix + '.d.' + key
<         if not self.r.exists(key):
<             raise KeyError(key)
<         return self.r.delete(key)
< 
90a88
> 
95c93
<             skey = self.prefix + '.d.' + key
---
>             skey = self.prefix + '.d.' + key.decode()
96a95,96
>             svalues = filter(None, svalues)
>             svalues = [i.decode() for i in svalues]
99a100
> 
102c103
<             yield key, self.r.hgetall(skey)
---
>         yield key, self.r.hgetall(skey)


redisstore_py2.py", line 76, in store_put
     for k, v in values.iteritems():
 AttributeError: 'dict' object has no attribute 'iteritems'

dessant:
Thanks, it could be wrapped in code tags so it's readable. https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code

dessant:
Fixed in https://github.com/kivy/kivy/pull/5194.

KeyWeeUsr:
@gregers666 Please try with the linked change if it works for you.

PR #5194, opened 2017-05-26
Fix py2/py3 iteritems

KeyWeeUsr:
I wonder how is it possible that the graphics part even works with iteritems as it should throw an `AttributeError` on py3. Or `bind` and `unbind`. Although, maybe it was fixed during the translation to C by Cython itself?

Anyway, it should fix the #5179


dessant:
Yes, iteritems is handled by Cython. The change in the pyx modules will likely make things slower.

KeyWeeUsr:
Okay, that should be an easy fix then 🐱

PR #5195, opened 2017-05-27
Add on_load to AsyncImage

KeyWeeUsr:
When we bind to `texture` or `texture_size` it's still a bad workaround for the action when an image is downloaded and the texture updated because:

* texture is updated anyway *during* the downloading (the circle gif)
* texture_size might be the same

this event makes it clear when the image is ready for use.


KeyWeeUsr:
Proper naming from the `ProxyImage` as there already is the `on_load` method.

PR #5206, opened 2017-06-01
Touchscreen fixes

audryste:


iCmdr:
Thanks for your PR! I tested your changes and got an exception on reconnect. The mtdev_open function call in __INIT__ in kivy/lib/mtdev.py on device initialization throws an exception ("Unable to open device").

 Exception in thread Thread-2:
   File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
     self.run()
   File "/usr/lib/python2.7/threading.py", line 763, in run
     self.__target(*self.__args, **self.__kwargs)
   File "/usr/local/lib/python2.7/dist-packages/kivy/input/providers/mtdev.py", line 288, in _thread_run
     _device = Device(_fn)
   File "/usr/local/lib/python2.7/dist-packages/kivy/lib/mtdev.py", line 148, in __init__
     raise Exception('Unable to open device')
 Exception: Unable to open device

Your code is similar to mine, yet I didn't have the same problem. After some debugging I found that the reason for this is that the idling time of 0.05 per loop is not long enough. The os.path.exists function detects the presense of the device faster than its ready to be connected to. Thus I would suggest waiting like 1 second per loop. It should also decrease CPU load. Currently, after the exception has been thrown in __INIT__, there won't be any more reconnects.

audryste:
iCmdr,  What touchscreen you are using?
Maybe it depends from the touchscreen manufacturer? 
If it is so, you should add sleep of 1 second right after mtdev file input is detected and leave 0.05 s of polling. In this case, it will work 100 % of the time.

I am just waiting while input file appears on the system. Just make sleep one more second right after it is got detected. 
For my touchscreen it worked perfectly 24 hours, with no crashes or what so ever.

iCmdr:
I am using a Waveshare Capacitive IPS touchscreen (http://www.waveshare.com/wiki/7inch_HDMI_LCD_(C)). The touch controller works over USB. I can run further tests later this evening.

pietervandermeer:
this patch fixes all my problems with disconnecting usb keyboard and usb sound card.. is this already in the main branch?

KeyWeeUsr:
@audryste Sorry for the delay and thanks for the reconnecting + other fixes! 🍡

PR #5212, opened 2017-06-02
fix #5184: ScrollView bar_margin affects also touch position

### LINKED ISSUE #5184: `bar_margin` only affecting the appearance of scrollbar in `ScrollView`
### Versions

* Python: 3.6.1
* OS: Arch Linux
* Kivy: 1.10.0

### Description

I am trying to use the `bar_margin` keyword arg of the `ScrollView` class to change the position of the scrollbar when `scroll_type=['bars']`. When I set `bar_margin`, the scrollbar is correctly rendered in a different position, but then it cannot be dragged using the mouse. Instead, the user still needs to click in the original spot as if `bar_margin=0` in order to drag the scrollbar.

### Code and Logs

```python

layout = GridLayout(cols=1, spacing=10, size_hint_y=None)
# Make sure the height is such that there is something to scroll.
layout.bind(minimum_height=layout.setter('height'))
for i in range(100):
    btn = Label(text=str(i), size_hint_y=None, height=40)
    layout.add_widget(btn)
root = ScrollView(size_hint=(1, None), size=(Window.width, Window.height), scroll_type=["bars"], bar_margin=40, bar_width=20)
root.add_widget(layout)

runTouchApp(root)

```

ismailof:
Fix for the #5184 issue

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -634,22 +634,19 @@ class ScrollView(StencilView):
         ud = touch.ud
         scroll_bar = 'bars' in scroll_type
 
-        # check if touch is in bar_x(horizontal) or bay_y(bertical)
-        ud['in_bar_x'] = ud['in_bar_y'] = False
+        # check if touch is in bar_x(horizontal) or bar_y(vertical)
         width_scrollable = vp.width > self.width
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
Thanks :)

PR #5218, opened 2017-06-09
Add support for footnotes to RST

KeyWeeUsr:
Test with this and don't forget to click on both footnote numbers/symbols and on the ref numbers/symbols. It has to scroll between them properly.

```python
from kivy.app import App
from kivy.lang import Builder
class My(App):
    def build(self):
        self.string='''
Test title
==========

Footnote reference: [8]_

space

space

Footnote [#]_

space

space

autonumbered: [#]_

space

space

Footnote [*]_

space

space

autosymbol: [*]_

space

space

Text text text text text text text

text text text text text text text [8]_

space

space

.. [8] Footnote ref
.. [#] 1Footnote autonum
.. [#] 2Footnote autonum
.. [*] 1Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym
.. [*] 2Footnote autosym etc
'''
        return Builder.load_string('''
RstDocument:
    text: app.string
''')
My().run()
```

Also a note about the `brute_refs`: I tried to find something less shitty and with the parser directly, but all I got back was either an empty strng or an empty list, therefore you can deduce why such name. It basically does the same thing like a visitor, but we *need* to do it **before** we even use our Visitor class, so that we have the references and backreferences set up. If someone has a better idea, feel free to share. 😄

### DIFF
--- a/kivy/uix/rst.py
+++ b/kivy/uix/rst.py
@@ -840,7 +840,6 @@ class _ToctreeVisitor(nodes.NodeVisitor):
             self.push(section)
         elif cls is nodes.title:
             self.text = ''
-
         elif cls is nodes.Text:
             self.text += node
 
@@ -868,6 +867,8 @@ class _Visitor(nodes.NodeVisitor):
 
         # store refblock here while building
         self.foot_refblock = None
+
+        # store order for autonum/sym footnotes+refs
         self.footnotes = {
             'autonum': 0,
             'autosym': 0,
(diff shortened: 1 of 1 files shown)


tito:
Thanks for the PR :)

kivy/uix/rst.py
```diff
@@ -327,6 +327,21 @@ def run(self):
...
+    size_hint_y: None
+    height: self.minimum_height
+
+<RstFootName>:
+    markup: True
+    valign: 'top'
+    size_hint: 0.2, 1
+    color: (0, 0, 0, 1)
+    bold: True
+    text_size: self.width - 10, self.height - 10
```
tito: Maybe uses dp()

PR #5220, opened 2017-06-15
Redeclare Svg.reload as throwing an exception.

robertwb:
The .pxd file must match the .pyx file lest cimporters don't check for an exception.


KeyWeeUsr:
Should be fine, thanks :)

PR #5232, opened 2017-06-23
Clean hanging code

KeyWeeUsr:
Basically the default behavior for a Python function is returning a `None` if not told otherwise, so there's no need for the hanging `None`, nor for hanging `return` at the end of some functions + it doesn't match the rest of the codebase (either `return` or `return <value>` or nothing).

The `Python.h` extern is from older @matham's code that apparently isn't around (or at least the imported stuff isn't used), ref #2566

I'm not sure about the X11 Cython code, because it doesn't import anything from the C code apparently, only makes it required at the compile time I guess? But then the `window_x11_keytab.c` isn't required this way, so... That's @tito's code, so I'm not sure about it. ref https://github.com/kivy/kivy/commit/adb5a850c52ef7127201f78bac16f55df8c9ca32

Also, in py2 `print()` prints a tuple (`()`), which will break the output (the GLES file) and I've seen it even in the `report.py`.

### DIFF
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -13,17 +13,6 @@ handlers.
 
 __all__ = ('EventDispatcher', 'ObjectWithUid', 'Observable')
 
-
-cdef extern from "Python.h":
-    ctypedef int (*visitproc)(PyObject *, void *)
-    ctypedef int (*inquiry)(PyObject *)
-    ctypedef int (*traverseproc)(PyObject *, visitproc, void *)
-    ctypedef struct PyTypeObject:
-        traverseproc tp_traverse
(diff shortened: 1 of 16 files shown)


kivy/_event.pyx
```diff
@@ -13,17 +13,6 @@ handlers.
 
 __all__ = ('EventDispatcher', 'ObjectWithUid', 'Observable')
 
-
-cdef extern from "Python.h":
```
dessant: Can this go?

dessant: @matham, Can this go?

matham: yes.

PR #5233, opened 2017-06-23
svg improvements

tshirtman:
started with adding support for Q/T commands, and seeing some simple things didn't work, i added a few fixes

- empty filling weren't transparent
- paths were closed automatically all the time
- stroke width didn't seem supported
- only pixel and mm units where supported, but not even everywhere

all of that should be fixed, and without breaking the more complex cases like tiger, not sure about performance differences, but the tiger at least is not slower for me (could be faster, but i didn't check old timing again)

### DIFF
--- a/kivy/graphics/svg.pxd
+++ b/kivy/graphics/svg.pxd
@@ -55,6 +55,7 @@ cdef class Svg(RenderContext):
     cdef double last_cy
     cdef Texture line_texture
     cdef StripMesh last_mesh
+    cdef bint closed
 
     cdef void reload(self) except *
     cdef parse_tree(self, tree)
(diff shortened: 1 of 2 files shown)


Zen-CODE:
Great job. It fixes quite a few anomalies we picked up in rendering. We are looking at moving as many of our image over to SVG as possible for obvious reasons, so these improvements are really useful. Thanks massive for this :-)

Zen-CODE:
ps. Please don't delete this branch yet. We may want to merge it with 1.10.0 for deploying...

tshirtman:
oh, i didn't expect it to be merged yet, last time i checked, it broke a few things (as well as fixing numerous, of course). people might be surprised by the line width change, the tiger is not broken iirc, but some other models like the cloud with a smile look different iirc.

I guess i'll have to send another batch of fixes asap, but not sure when i can.

Anyway, people who depend on svg should test it thoroughly asap so the issues are known.

Zen-CODE:
With out test batch of svg's, its still a great improvement over the
existing implementation. As the docs label it experimental anyway, i dont
see any reason to hold these changes back:-)

Zen-CODE:
A thought. You could still keep committing on this branch and we can test and merge commits as they occur. We are currently doing heavy testing on SVG rendering so can follow and test quite thoroughly...

Issue #4777, opened 2016-12-08
Local Config file misplaced


KeyWeeUsr:
For some reason when you set local configuration and save it (to a `<app>.ini`), the file is misplaced on Android and maybe even on iOS, which makes it harder to remove the app - the file will persist and will annoy users, because the config is pulled from that file, therefore if a user makes a mistake in the config, the config would need to be removed manually.

ref https://github.com/kivy/python-for-android/issues/953

inclement:
This behaviour is probably also the only thing in Kivy that needs the external storage read/write permission. With this fixed (which I believe it should be), Kivy apps wouldn't need any permissions by default.

KeyWeeUsr:
@inclement [this line](https://github.com/kivy/kivy/blob/5913d861f45119dfff0b6cb5d004e6f79858563c/kivy/app.py#L657) and the line under (ios) will cause the issue. I don't have a clue how iOS handles permissions, but this requires the Android `WRITE_EXTERNAL_STORAGE` permission and if android is going to change, iOS should aquire the same/similar structure.

If not change, then at least a mention in the documentation that such thing _requires_ additional stuff such as android permission for writing to the SDcard.

However, I notices something weird and that's missing `Mac` platform or better said folder. @akshayaurora could you please look at the missing part and confirm if iOS requires similar thing to android write permission for such configuration?

Windows places the file next to the `main.py` file, therefore it behaves as expected, not as plain `Config.write` (which sometimes writes to `%appdata%`).

Zen-CODE:
@KeeWeeUsr. The iOS path looks fine. This '~/Documents/' folder is this-app-only, write-able and pretty much guaranteed available by iOS without special permissions, so should not present a problem. 

Although strictly speaking Apple recommends this only for files which a user should see, the file system is so hidden on iOS its' a very academic distinction.

https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html

KeyWeeUsr:
@Zen-CODE Thanks for the info, I thought it's just some general folder and it _needs_ additional permissions. I presume the folder **is deleted** after the app is uninstalled, right? Otherwise the issue persists i.e. file not removed and old config used.

I'm not sure if that should be thrown at the dev to handle him/herself, but it doesn't seem like the case to me. Kivy shouldn't just throw stuff everywhere and expect dev to go play hide&seek. 😋

[...]

PR #5237, opened 2017-06-25
Change android App.config path

KeyWeeUsr:
More about why is in the linked issue (#4777), but this change also does something "bad" in the sense that an ordinary user won't be able to edit and most likely even view the config file. On the other hand, the file will get removed by uninstalling the application and probably will be safe for updating the Kivy app (I think the `<app>/files/app` folder isn't removed with an update. Not sure though).

This change removes the necessity of having a `WRITE_EXTERNAL_STORAGE` permission for the default Kivy app and if required, the dev can change the folder manually by overriding the method if the dev would like to create the file in `/sdcard/<somewhere>` with adding the permission by own needs.

I think `sys.executable` returns such a path because of [this line](https://github.com/kivy/python-for-android/blob/c45bd81ea15b345815450dc2fd14ff884fb12fbb/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java#L351) which is used in `start.c` to set the Python paths.

I'm not sure, is this considered an API break if we change the path only and not the call directly? I mean, a dev should use the API we provide and not the raw path, so this should be safe, I guess?

### DIFF
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -666,7 +666,6 @@ class App(EventDispatcher):
             activity = autoclass('org.kivy.android.PythonActivity').mActivity
             defaultpath = join(
                 activity.getFilesDir().getAbsolutePath(),
-                'app',
                 '.%(appname)s.ini'
             )
 
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -679,6 +679,9 @@ class App(EventDispatcher):
                     'App: get_application_config sdcard path is deprecated, '
                     'moving to the new one - %s.' % new_path
                 )
+                if not exists(dirname(new_path)):
+                    # recursive mkdir, just in case
+                    os.makedirs(dirname(new_path))
                 copyfile(old_path, new_path)
 
         elif platform == 'ios':
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -307,14 +307,13 @@ Here is a simple example of how on_pause() should be used::
 
     Both `on_pause` and `on_stop` must save important data because after
     `on_pause` is called, `on_resume` may not be called at all.
-
 '''
 
 __all__ = ('App', )
 
 import os
 from inspect import getfile
-from os.path import dirname, join, exists, sep, expanduser, isfile
+from os.path import dirname, join, exists, sep, expanduser, isfile, abspath
 from kivy.config import ConfigParser
 from kivy.base import runTouchApp, stopTouchApp
 from kivy.compat import string_types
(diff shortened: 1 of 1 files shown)


hottwaj:
Thanks for looking into this.  One issue with merging this is that anyone who released an android app using a prior version of Kivy which they subsequently update to a new Kivy version will find that already installed android apps will lose all settings when updated.  

Maybe a method could be provided that lets legacy apps copy across old settings to the new location?

KeyWeeUsr:
@hottwaj That's of course clear to me and it's the main reason why I haven't merged it and only proposed the change. Everything that depends on this location, in the sense that some files are pulled from there and not just saved, will be broken with just this fix. I just brought it up, so that it triggered some discussion as for "what next".

kivy/app.py
```diff
@@ -654,7 +661,10 @@ def get_application_config(self):
         '''
 
         if platform == 'android':
-            defaultpath = '/sdcard/.%(appname)s.ini'
+            defaultpath = join(
+                dirname(abspath(sys.executable)),
```
inclement: Isn't `sys.executable` just `''` with python2 or `android_python` on Android? It would probably be better to just check the environment variables (`ANDROID_PRIVATE`) etc. to find the app's own directory.

inclement:
Great! I think the target directory should be a little different, but this is an excellent change.

Per the discussion elsewhere, I had thought it should be fine to first check if an old config file already exists in /sdcard, to avoid breaking existing apps. That seems like it should be safe, but I'm not sure. Do you think it would be fine?

kivy/app.py
KeyWeeUsr: works just fine with Pyonic on python 2 :) the env var is set in `PythonActivity.java` + `start.c`, so we actually just need to make sure we don't change the core that launches the app (and if we change, it won't launch), so we should be quite safe with this one, because it'd be probably the first thing to fix.

KeyWeeUsr:
Hm, seems like a good idea. Maybe just check if something is in that place and pull it to the new one (+ remove?). Also, it's apparently the only thing that [requires sdcard](https://github.com/kivy/kivy/search?utf8=%E2%9C%93&q=sdcard&type=), or at least that's so obvious, that it uses it.

The rest, such as using pyjnius for storage would mean the user knows what storage (s)he wants to access, therefore knows the permissions (i.e. not our problem to fix).

KeyWeeUsr:
I patched it, so that it detects the file on sdcard if available and moves to the new location. Then it ignores the old location completely. I guess it should be fine?

kivy/app.py
```diff
@@ -654,7 +662,25 @@ def get_application_config(self):
         '''
 
         if platform == 'android':
-            defaultpath = '/sdcard/.%(appname)s.ini'
+            from jnius import autoclass
+            activity = autoclass('org.kivy.android.PythonActivity').mActivity
+            defaultpath = join(
+                activity.getFilesDir().getAbsolutePath(),
+                'app',
```
inclement: Perhaps we should omit the 'app', that way it isn't in the main application files directory (which is fine), and also will work if anyone happens to use it with old p4a versions.

inclement:
Seems good to me. I'll test it to check, but other than that I'm all for making the change.

kivy/app.py
inclement: Updated comment: I'm pretty sure we should omit the `'app'`. This should mean that the config does not get overwritten on app update. This is actually quite important.

KeyWeeUsr: @inclement I'm on it.

KeyWeeUsr:
@inclement ping

Issue #5238, opened 2017-06-25
Vector.distance method: math domain error


taylorjdawson:
### Versions

* Python: 3.6.1
* OS: Windows 10 Education; Version 1703
* Kivy: 1.10.0
* Kivy installation method: pip

### Description

The app will crash when rapidly clicking around inside of the window. I tracked down the issue. It is with the [`Vector.distance()`](https://github.com/kivy/kivy/blob/master/kivy/vector.py#L254) function when the distance is calculated. The issue is the `math.sqrt()` function.
I had the distance method print the values that were causing the error. They are different everytime.

These values will cause the `math.sqrt()` function to throw the `ValueError: math domain error` exception:
```python
import math
x2 = 0.48125; x1 = 0.48; y2 = 0.4716666666666667; y1 = 0.475
print(math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2))
```

For some reason the `math.sqrt()` function is unable to handle these inputs.

Steps to Reproduce:
1. Run the code shown below
2. Rapidly click around inside of the window that is displayed
3. The app will crash and you will get the error seen below in the Traceback

### Code and Logs

Code:
```python
from kivy.app import App
from kivy.uix.label import Label

class MyApp(App):

    def build(self):
        return Label(text='Hello world')

if __name__ == '__main__':
    MyApp().run()
```

Traceback:

```python
 Traceback (most recent call last):
   File "C:/Users/Levi/PycharmProjects/TheOfficeQuoteDatabase/test_kizy.py", line 12, in <module>
     MyApp().run()
   File "C:\Users\Levi\AppData\Local\Programs\Python\Python36-32\lib\site-packages\kivy\app.py", line 828, in run
     runTouchApp()
   File "C:\Users\Levi\AppData\Local\Programs\Python\Python36-32\lib\site-packages\kivy\base.py", line 504, in runTouchApp
     EventLoop.window.mainloop()
   File "C:\Users\Levi\AppData\Local\Programs\Python\Python36-32\lib\site-packages\kivy\core\window\window_sdl2.py", line 663, in mainloop
     self._mainloop()
   File "C:\Users\Levi\AppData\Local\Programs\Python\Python36-32\lib\site-packages\kivy\core\window\window_sdl2.py", line 405, in _mainloop
     EventLoop.idle()
   File "C:\Users\Levi\AppData\Local\Programs\Python\Python36-32\lib\site-packages\kivy\base.py", line 342, in idle
     self.dispatch_input()
[...]

PR #5239, opened 2017-06-25
Fix Issue 5238

taylorjdawson:
Use ** .5 to calculate the square root instead of math.sqrt() to avoid ValueError: math domain error which causes the app to crash.
Issue source: https://github.com/kivy/kivy/issues/5238

### DIFF
--- a/kivy/vector.py
+++ b/kivy/vector.py
@@ -251,7 +251,7 @@ class Vector(list):
         14.035668847618199
 
         '''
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
It seems like the problem was on a single machine for some unexplained reason. Author and I couldn't reproduce the error on different machines. Closing until more details are provided.

PR #5241, opened 2017-06-26
Add async support to kivy

matham:
This adds async support to kivy. It is still experimental, but it should work perfectly and I think is in a mergeable state after testing by others.

Due to some compat methods, the codebase will work in < py3.5 but none of the async methods will be usable. To use the async parts you need python 3.5+. I tested with python 3.5.2 on windows (it "should" not be OS specific).

There are two possible approaches to add async support to kivy:

* Use kivy's event loop as an backend for an async event loop. That means that all the async task scheduling done externally go through the kivy event loop. This is the approach tshirtman has taken, which I found after doing this (https://github.com/kivy/kivy/pull/4525). 

  The problem with this approach is that the event loop is only as good as kivy's event loop, which you're stuck with since everything runs on top of it. But, I think it's a good idea to also have this approach. Still, that PR would require this PR as this PR converts kivy's Clock and sleeping methods into async. Otherwise, the underlying event loop would be very limited when it can execute the coroutines as kivy likes to sleep a lot :)
* Convert kivy's `run` method which runs the app into a async coroutine and make sure that all the methods it calls all the way down to the methods that sleep when idling are similarly coroutines. This is the approach of this PR. 

  What this means is that kivy is executing it's callbacks during each frame as it currently does. However, when it sleeps, the underlying eventloop running the kivy event loop is free to execute other async coroutines.

  In other words, the kivy event loop that is executing kivy clock events are run at a specific fps like before which executes all the clock events in a sync manner, which is required for e.g. animations and other things that require events be executed in association with graphics updates. However, during the downtime of the kivy event loop when it would normally sleep other coroutines can be executed. In other words, kivy's run method is a coroutine running as equal with any other coroutine and can get suspended only during its sleepy time.

One more thing has been added to make this usable and friendly. `EventDispatcher.async_bind` has been added. It returns a async iterable which can be used in a async function to wait in a non-blocking manner for properties to be updated and events to occur. E.g.

```py
async for event in my_button.async_bind('on_release'):
    # button has been pressed
    ... 

# or 
async for instance, value in textinput.async_bind('text'):
    print('text is :', value)
    ...
```

Look under `examples/async` for examples of how to use this, as well as with threads.


matham:
Hmm, didn't realize `yield from` is a syntax error which cripples the whole file in py2. I guess to allow the files to still be used in py2 I'd need to get get creative with eval or split the `yield from` code into a py3 only file :(

rnixx:
@matham - python 3 only file sounds not that bad. maybe symlinking it at setup time?

rnixx:
@matham at least it sounds less creepy than eval calls

matham:
I'm also wonder if it's time for a py3only branch with maybe a release of two until we support py3 only (but I don't want to make this PR too much a discussion about a py3 switch).

rnixx:
@matham just my 2c, but maintaining separate code bases for py2 and 3 might end up in a lot of unnecessary work due to keeping bug fixes in sync and managing multiple releases. That binds resources which might be better spent somewhere else.

What about putting python3 only code in a separate package until official python2 support gets dropped and merge it then to main line?

matham:
@rnixx yeah, that's what I'm thinking of doing. Create kivy/py3 and have e.g. kivy/py3/app.py, kivy/py3/clock.py etc. that we import into e.g. kivy/app.py only in py3. That should work generally I think. But we probably should come up with a py2 EOL plan.

rnixx:
you can also consider creating a dedicated kivy3 repo, which patches the original modules until python 2 support gets dropped. This way later replacement in the main line causes less noise with import locations, etc.

rnixx:
regarding EOL, this should be in sync with official python 2 EOL

matham:
Ok, I fixed the py2 issue by create modules side by side with original modules, e.g. clock/clock_async as recommended by @tito.

cstatz:
The clock_properties_async example is not working. It complains about missing attributes set and clear in the ClockEvent issued from utils_async.py:
 AttributeError: 'kivy._clock.ClockEvent' object has no attribute 'set'
 AttributeError: 'kivy._clock.ClockEvent' object has no attribute 'clear'

matham:
Yes, I didn't update clock_properties_async yet. Use the other example file. That should work.

matham:
I fixed a few things and add more examples. It should be in pretty good shape. I just need to add docs and test it fully in a real app.

jakogut:
> and test it fully in a real app.

I'm game to give it a shot.

tgoodlet:
Definitely works in my app. It'd be great to see this merged in soon :+1:

jakogut:
I can verify that this PR works in my app, and doesn't break anything.

matham:
Yeah, tests would be nice. I was also waiting to add some docs, but didn't find time so far.

chrysn:
I'm testing this from the perspective of someone well familiar with asyncio but new to kivy.

I appreciate the approach of running on an existing asyncio main loop; IMO using one's existing main loop as an asyncio loop should be reserved to high-performance corner cases.

Some assorted review notes:
* The KIVY_EVENTLOOP documentation only mentions "async" and "default", but the trio examples apparently use something different.
* There should be an example in classical asyncio (even if trip becomes more than a niche thing, the standard library's asyncio module is the more universally recognized). I can provide the 70-line telnet example I've used for testing if desired.
* The "Async app" documentation makes it sound like `App.run()` would, given KIVY_EVENTLOOP=async, run the program on the asyncio loop. That did not work for me, I had to use `asyncio.get_event_loop().run_until_complete(MyApp().async_run())` -- not unexpected or unfamiliar, but inconsistent with the documentation.

Once I found the last point, I could easily get an asyncio networked application running; this seems to be a fine asyncio port.

KeyWeeUsr:
@chrysn I haven't really looked at this PR, however, please do post the `telnet` example. The more examples for async, the smoother will be the transition for the users in the future (e.g. user finds this PR, gets a rough idea, etc). :)

ysangkok:
I tried this in Electrum, seems to work, but all our IO needs to be rewritten to not block, so it's not a good demo. But if I did that, I think it'd work well. I used Trio, didn't test with asyncio. https://github.com/ysangkok/electrum/tree/single-threaded

metamarcdw:
@ysangkok, I'm working on a different project that has some similarities to Electrum. All IO is currently non-blocking using asyncio. I would love to work together and potentially use that as a demo.

https://github.com/metamarcdw/nowallet

matham:
I haven't gotten a chance to finish this up yet, but will eventually. Right now I'm mainly stalling hoping we'll drop py2 support before I merge it so that I can avoid having the duplicate files. 

Also, @dolang is working on testing, and perhaps will add pytest support so that will make testing using trio with the pytest-trio plugin much easier.

Finally, regarding trio vs asyncio, it's true that the stdlib asyncio is the default. However, that has significant problems which is solved by trio and similarly curio (although I don't know about the latter). You can read more about it [here](https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/). In fact, yury, the person working on much of the stdlib asyncio has said that he's planning to integrate trio concepts [into asyncio for 3.8](https://mail.python.org/pipermail/python-ideas/2018-June/051224.html). That's why, in my implementation I was trying to discourage the use of asyncio and pushing trio.

With that said, I do think I'll make sure it works properly with asyncio, when I work on it again.

matham:
We don't have a ticket, but rather we are discussing it here:
https://groups.google.com/forum/m/#!topic/kivy-dev/kXnDut1e8dI.

Regarding what needs to happen, the decisions need to be made as to when
and how python 2 support is dropped. There are still some obstacles as you
can see in that thread, specifically with some of our sister projects.

chrysn:
The thread reads to me like this PR will be merged eventually, and all that stands in its way is Py2 support.

If that's the state of things, I'm perfectly happy with it staying as an open PR for as long as it takes and depending directly on the underlying branch in projects that explore async kivy. I'd suggest, though, that the branch be occasionally rebased onto master (or merged into a hypothetical next-after-py2-drop branch) to stay up to date with kivy, as I had to start cherry-picking commits like 05fc61424.

Thanks for your work on this!

matham:
Yeah, the PR will be merged, it's just a matter of dropping py2 support in master and the addition of docs/tests.

Given that, meaning that this will not be merged before we drop py2 support, I can refactor the PR to not create the side by side files because I wouldn't need to accommodate py2 when this gets merged. I guess I'll do that at some point.

RockyRx:
When will this be merged?

matham:
As I mentioned, it depends on us dropping py2 support. We don't expect that to happen for another couple of months, at least as there are other stuff that needs to happen before we can drop py2 (https://github.com/kivy/kivy/wiki/Kivy-Python-2-Support-Timeline). So maybe sometime at the beginning of next year at the earliest?

Fak3:
I've just tried this, and it works!

My main grief is that kivy's callback methods like `Button.on_release()` are sync. It would be more convenient to have async way to handle input events. Right now I can only overcome this limitation with boilerplate code like this:

```python

class MyButton(Button):
    def on_release(self, *a):
        """ Unfortunately it is sync method. Let's call a real async handler """
        # Assume App.nursery was assigned on app startup
        App.nursery.start_soon(self.async_on_release)
            
    async def async_on_release(self):
        """ This is real async on_release routine """
        url = 'https://apod.nasa.gov/apod/image/1705/Arp273Main_HubblePestana_3079.jpg'
        async for chunk in (await asks.get(url, stream=True)).body:
            print(len(chunk))
        self.text = 'done!'
            
```

Fak3:
> Could you save a ref to the button and write it as

The problem is that in non-trivial gui app the handler logic gets spread across many places, and very hard to follow.

So I was wondering if we can add support for async handlers. I managed to achieve it by adding a simple check `inspect.iscoroutinefunction()` to the `_event.pyx`.

Please look at this commit. I have added the example to demonstrate how simple the code becomes with this approach:

https://github.com/Fak3/kivy/commit/b10344a6eb07eba061c75d9644077861d7777a49#diff-3d2b16b26abff5cbc5b14251f83f4daf

My commit if very rough proof-of-concept, but I was wondering if we can add this behavior to kivy.

tgoodlet:
@matham any chance we can get this rebased onto `1.10.1`? 
I'm looking to use `kivy.clock.triggered` in vain :cry:

Fak3:
This PR includes the `async_bind()` convenience method, but i believe it is better be done in a separate PR.
I think there can be another API for event listening async-way, which uses `await` instead of `async for`. So instead of (currently suggested here):
```python
async for event in my_button.async_bind('on_release'):
    print(event)
```
I could look like this:
```python
while True:
    event = await my_button.listen('on_release')
```
This API could be more flexible than for-loop, but it is surely debatable. Anyway i'd suggest to split this into another PR, and have this one focused only on core of async support.

Chaoste:
I just faced an issue which might be because it is not the latest version. The basic drag 'n drop example is not working - the related event is simply not called ("on_dropfile") but the SDL2 example in C works fine and it also works with a clean kivy version. Maybe installing from this branch somehow messes some permissions up.

goodboy:
@matham another feature I want to [disable cursor blinking](https://kivy.org/doc/stable/api-kivy.uix.textinput.html#kivy.uix.textinput.TextInput.cursor_blink) in the `TextInput` widget.

Only in `1.10.1` though :crying_cat_face:

matham:
Now that we dropped python 2 support, we can add support for async. Closing this PR isn favor of https://github.com/kivy/kivy/pull/6368.

Issue #5242, opened 2017-06-26
Kivy can crash when a TextInput.is_focusable is set to False and focus is switched to another widget at same time


Bakterija:
* Python: All
* OS: Ubuntu 16.04
* Kivy: 1.10 and 1.10.1-dev0
* Kivy installation method: pip

Tested with this [test app](https://gist.github.com/Bakterija/81fa1b11785cfb28279c65c2660ef512)

Error log:
```
[INFO   ] [Logger      ] Record log in /home/atis/.kivy/logs/kivy_17-06-26_71.txt
[INFO   ] [Kivy        ] v1.10.1.dev0, git-f91378d, 20170626
[INFO   ] [Python      ] v2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609]
[INFO   ] [Factory     ] 194 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_sdl2, img_pil, img_gif (img_ffpyplayer ignored)
[INFO   ] [OSC         ] using <multiprocessing> for socket
[INFO   ] [Window      ] Provider: sdl2(['window_egl_rpi'] ignored)
[INFO   ] [GL          ] Using the "OpenGL" graphics system
[INFO   ] [GL          ] Backend used <gl>
[INFO   ] [GL          ] OpenGL version <3.0 Mesa 12.0.6>
[INFO   ] [GL          ] OpenGL vendor <Intel Open Source Technology Center>
[INFO   ] [GL          ] OpenGL renderer <Mesa DRI Intel(R) HD Graphics 520 (Skylake GT2) >
[INFO   ] [GL          ] OpenGL parsed version: 3, 0
[INFO   ] [GL          ] Shading version <1.30>
[INFO   ] [GL          ] Texture max size <8192>
[INFO   ] [GL          ] Texture max units <32>
[INFO   ] [Window      ] auto add sdl2 input provider
[INFO   ] [Window      ] virtual keyboard not allowed, single mode, not docked
[INFO   ] [Text        ] Provider: sdl2
[INFO   ] [Clipboard   ] Provider: gtk3(['clipboard_dbusklipper'] ignored)
xclip version 0.12
Copyright (C) 2001-2008 Kim Saunders et al.
Distributed under the terms of the GNU GPL
[INFO   ] [Cutbuffer   ] Provider: xclip
[INFO   ] [CutBuffer   ] cut buffer support enabled
[...]

PR #5243, opened 2017-06-26
Fix for crash when setting is_focusable property in issue #5242

Bakterija:
Issue #5242

### DIFF
--- a/kivy/uix/behaviors/focus.py
+++ b/kivy/uix/behaviors/focus.py
@@ -375,14 +375,11 @@ class FocusBehavior(object):
         fbind('focus_previous', self._set_on_focus_previous)
 
     def _on_focusable(self, instance, value):
-        if not value and self.focus:
-            self.focus = False
-            self._unbind_keyboard()
-        elif self.disabled or not self.is_focusable:
+        if self.disabled or not self.is_focusable:
             self.focus = False
 
     def _on_focus(self, instance, value, *largs):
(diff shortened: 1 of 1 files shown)


kivy/uix/behaviors/focus.py
```diff
@@ -375,7 +375,10 @@ def __init__(self, **kwargs):
...
 
     def _on_focusable(self, instance, value):
-        if self.disabled or not self.is_focusable:
+        if not value and self.focus:
+            self.focus = False
+            self._unbind_keyboard()
+        elif self.disabled or not self.is_focusable:
             self.focus = False
 
     def _on_focus(self, instance, value, *largs):
```
matham: Ideally, setting `self.focus = False` should be enough as that should release everything in `_on_focus`. The problem is that if you set first `is_focusable` to False, `_on_focus` will not do anything because of the `if self.is_focusable and self.keyboard_mode == 'auto':`. 

I think a better fix would then be to change 

```py
if self.is_focusable and self.keyboard_mode == 'auto':
     if value:
         self._bind_keyboard()
     else:
        self._unbind_keyboard()
```
to
```py
if self.keyboard_mode == 'auto':
     if not value:
         self._unbind_keyboard()
     elif self.is_focusable:
        self._bind_keyboard()
```

Bakterija: That indeed is a fine way to do it, i've updated code in my patch.

PR #5244, opened 2017-06-26
Fix sdl2 Window.on_textinput event not working properly

Bakterija:
OS: Windows and Ubuntu
Python: All

Hello.
A few days ago i talked about sdl on_textinput events dissapearing permanently after a TextInput widget gains focus. 
Now i have a pr with a good [example app](https://gist.github.com/Bakterija/bbe89c1b95857255e71abc215591783b)

The window patch in it fixes the issue and doesn't cause any problems on desktop or android, or at least i couldn't see any.

### DIFF
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -714,8 +714,7 @@ class WindowSDL(WindowBase):
 
     def release_keyboard(self, *largs):
         super(WindowSDL, self).release_keyboard(*largs)
-        if self._system_keyboard.widget:
-            self._win.hide_keyboard()
+        self._win.hide_keyboard()
         self._sdl_keyboard = None
         return True
 
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -714,7 +714,8 @@ class WindowSDL(WindowBase):
 
     def release_keyboard(self, *largs):
         super(WindowSDL, self).release_keyboard(*largs)
-        self._win.hide_keyboard()
+        if self._system_keyboard.widget:
+            self._win.hide_keyboard()
         self._sdl_keyboard = None
         return True
 


kivy/core/window/window_sdl2.py
```diff
@@ -714,7 +714,8 @@ def request_keyboard(self, callback, target, input_type='text'):
 
     def release_keyboard(self, *largs):
         super(WindowSDL, self).release_keyboard(*largs)
-        self._win.hide_keyboard()
+        if self._system_keyboard.widget:
+            self._win.hide_keyboard()
         self._sdl_keyboard = None
```
matham: I'm not sure why this would work. `widget` seems to be a private property of the keyboard. Why would we only release the keyboard when there's a virtual keyboard?

Bakterija: Ok, i read some more source code, and this is a wrong way to do it, system keyboard widget is  None even on android. The example is still good imo, it shows where the problem is. 
The problem is that self._win.hide_keyboard() calls SDL_StopTextInput() in the sdl2 cythonized module. [That function](https://wiki.libsdl.org/SDL_StopTextInput) disables textinput events completely.

Bakterija: @matham i have used kivy apps without that for weeks now and didn't see any bad effects, is there a good reason for disabling textinput events?

matham: I don't really know why that line is there. @tito added it when he wrote the backend https://github.com/kivy/kivy/blame/master/kivy/core/window/window_sdl2.py#L717. Maybe it can be removed.

Bakterija: I am actually not sure why i put the system keyboard widget there, i must have been looking at the wrong code when i was rushing to complete my widget. It works because it always returns None

Bakterija:
I changed the sdl2 window to what it was before, let's just wait for a response from @tito .

Bakterija:
I don't think that i can help here, fixing the bind requires a decision from core devs.
I have decided to make an issue instead.

PR #5246, opened 2017-06-27
fix emacsbehavior inserting text

### LINKED ISSUE #5245: Shortcuts for EmacsBehavior insert text
### Versions

* Python: 3.6.1
* OS: linux
* Kivy: 1.10.1.dev0 7daea785f5dfe5b54a3ced7510d87ad6a711f0e9
* Kivy installation method: `pip install --user --editable .`

### Description
When using EmacsBehavior with TextInput shortcuts with alt insert the pressed keys are into the text. E.g. when pressing alt+f the cursor is moved to the end of the word *and* an f is inserted. This does not happen with ctrl shortcuts.

### Code and Logs

The error can be reproduced by the `examples/widgets/codeinput.py` example. Alternativly here is a minimal example which reproduces the error:

```python
Config.set('kivy', 'log_level', 'debug')

from kivy.app import App

from kivy.uix.behaviors import EmacsBehavior

class CodeInput(EmacsBehavior, TextInput):
    pass

class TestApp(App):
    def build(self):
        codeinput = CodeInput()
        codeinput.text = "Multiline Text\n" * 10
        return codeinput

if __name__ == '__main__':
    TestApp().run()

```

syntonym:
fixes #5245 

I think SDL2 does not forward textinput events when ctrl is pressed, but does if alt is pressed.

### DIFF
--- a/kivy/uix/behaviors/emacs.py
+++ b/kivy/uix/behaviors/emacs.py
@@ -106,6 +106,12 @@ class EmacsBehavior(object):
             super(EmacsBehavior, self).keyboard_on_key_down(window, keycode,
                                                             text, modifiers)
 
+    def keyboard_on_textinput(self, window, text):
+        if (self._alt_l or self._alt_r):
+            if text in self.bindings['alt']:
+                return
+        super(EmacsBehavior, self).keyboard_on_textinput(window, text)
+
     def delete_word_right(self):
         '''Delete text right of the cursor to the end of the word'''
         if self._selection:


KeyWeeUsr:
Seems like a weird way for fixing this issue. I mean, if it's a problem with the default keyboard and its modifiers, this isn't the place for such fixes. If it's just emacs behavior thingy, I think it should be fixed in the keybindings directly. Also, I tried the fix and it doesn't PR the issue for me.

syntonym:
Seems more like a issue with SDL on my platform, see #5245

PR #5252, opened 2017-07-01
Add support for shaped windows

KeyWeeUsr:
The funny feeling when you implement own feature request after a year. 😆 

I tested it only on Windows for now, so feel free to go and try on other systems. I'm quite interesting on the behavior on Android although I'm not sure if it'd work out of the box. I think the image resizing might be necessary.

image

And yes, the transparent areas don't collide with the Kivy window, therefore if you click on something through the Kivy window, the touch will also go through (and the Kivy window will disappear from the front).


manthansharma:
Tried on OSX but same window shape for all the scenario.

screen shot 2017-07-12 at 3 22 52 am
screen shot 2017-07-12 at 3 23 05 am
screen shot 2017-07-12 at 3 23 08 am

KeyWeeUsr:
Just a note about SDL2:

* X11 will need a `--enable-video-x11-xshape` flag in [SDL2 configure](https://hg.libsdl.org/SDL/file/912b770ded2a/configure#l20862) before compilation, so it might not work with the default library from some distro.
* OSX will need [Cocoa](https://en.wikipedia.org/wiki/Cocoa_(API)) and `--enable-video-cocoa` flag ([config](https://hg.libsdl.org/SDL/file/912b770ded2a/configure#l21071)) at configuring time also.
* Windows just works because it's damn old API from Win 98 or something, so it's not really something that wouldn't be disabled as a lot of stuff uses it even now.

KeyWeeUsr:
This code was tested by @Bakterija on linux, it [definitely works on Mac](http://www.cocoawithlove.com/2008/12/drawing-custom-window-on-mac-os-x.html), however the SDL2 code might need to be compiled manually i.e. not downloaded from some package repo, brew or anything like that. For example on [debian](https://hg.libsdl.org/SDL/file/e3797888c6f1/debian/rules#l9)-like packages they disable shared X11 libs which might result in not supporting the shaped windows, although I haven't found anything to support this claim (I haven't found anything about excluding XShape files in the `configure`).

The previous behavior is untouched and the `Config` value defaults to disabled shaped windows, so I don't see why it shouldn't be merged as it's sometimes an asked feature on IRC and it has been in SDL2 since version `1.3`.

hackalog:
What does "it definitely works on Mac, however the SDL2 code might need to be compiled manually i.e. not downloaded from some package repo, brew or anything like that" mean here? This would seem to imply the Kivy OSX instructions (https://kivy.org/docs/installation/installation-osx.html) are no longer correct?

KeyWeeUsr:
@hackalog No, it means that the SDL2 library distributed via `brew` or some other package manager might not be configured to actually use the [`SDL_cocoashape.m`](https://hg.libsdl.org/SDL/file/tip/src/video/cocoa/SDL_cocoashape.m) (i.e. it won't even try to compile and link it), therefore the provided [example](https://github.com/kivy/kivy/blob/master/examples/miscellaneous/shapedwindow.py) might not work at all - you'd see only a basic window without an ability to use the shapes.

fabiobhl:
What do you mean by: "OSX will need Cocoa and --enable-video-cocoa flag (config) at configuring time also."
Do I have to change a config file manually or what should I do to get it working on OSX?

KeyWeeUsr:
It means that the SDL2 library has to be compiled with Cocoa support which might not be the default for you if you download from some package manager or something. You should make sure you compile SDL2 with the required flags for shaped window (those that include `SDL_cocoashape` and related files. :)

bballjbv1:
So how do you implement a basic drag feature once the alpha or image shape is set for the window? I can't seem to find out how to move the fixed window around the desktop once its been shaped.

Fionnoch:
Are there any examples on how to implement this? I am using windows and have been unable to make my kivy app transparent as in the first posts photos. Any guidance would be hugely appreciated

hardik-rajpal:
Is it possible to lock the kivy window in the foreground, while maintaining transparency? I want the window to stay visible but also allow clicks through the transparent area.

PR #5260, opened 2017-07-05
Disable emacs bindings for Alt-Gr (Ctrl+Alt) key

KeyWeeUsr:
ref #5245 #5246 

I'm not sure if it's the right way, but at least by default AltGr keybindings don't work for me on Emacs, so I guess it's the right fix?


inclement:
The code looks like a reasonable way to do it to me, although I haven't tested it.

PR #5265, opened 2017-07-07
Remove dead code for SDL2 windowresized event

bionoid:
I don't know the original intent here, but this is never called due to
same condition in elif block above. Maybe the canvas.ask_update should
be included, maybe it was intended for handling a different event...


dessant:
It seems a redraw is also triggered by the compound statement above the removed one:
https://github.com/kivy/kivy/blob/b02d30590804ba0474e55f4ee44d1f9a1a031aa0/kivy/core/window/__init__.py#L1279

Issue #4224, opened 2016-04-26
Remove deprecated video and audio providers: pygst, pyglet and pygi


Zonzely:
hello!
I run a demo named **Camera Example**  in kivy website and get error ---**No module named 'pygst'**.
the link of demo  is [https://kivy.org/docs/examples/gen__camera__main__py.html](url).
my OS is  ubuntu15.04  and i use python3. the detailed error is following:

```
zz@zz-Aspire-E1-471G:~/data/code/kivy_practice/camera$ python3 test.py [INFO   ] [Logger      ] Record log in /home/zz/.kivy/logs/kivy_16-04-26_47.txt
[INFO   ] [Kivy        ] v1.9.1
[INFO   ] [Python      ] v3.4.3 (default, Mar 26 2015, 22:03:40) 
[GCC 4.9.2]
[INFO   ] [Factory     ] 179 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_gif, img_sdl2, img_pil (img_ffpyplayer ignored)
[INFO   ] [OSC         ] using <multiprocessing> for socket
[INFO   ] [Window      ] Provider: sdl2(['window_egl_rpi'] ignored)
[INFO   ] [GL          ] OpenGL version <b'3.0 Mesa 10.5.9'>
[INFO   ] [GL          ] OpenGL vendor <b'Intel Open Source Technology Center'>
[INFO   ] [GL          ] OpenGL renderer <b'Mesa DRI Intel(R) Ivybridge Mobile '>
[INFO   ] [GL          ] OpenGL parsed version: 3, 0
[INFO   ] [GL          ] Shading version <b'1.30'>
[INFO   ] [GL          ] Texture max size <8192>
[INFO   ] [GL          ] Texture max units <16>
[INFO   ] [Window      ] auto add sdl2 input provider
[INFO   ] [Window      ] virtual keyboard not allowed, single mode, not docked
[CRITICAL] [Camera      ] Unable to find any valuable Camera provider at all!
pygst - ImportError: No module named 'pygst'
  File "/usr/lib/python3/dist-packages/kivy/core/__init__.py", line 59, in core_select_lib
    fromlist=[modulename], level=0)
  File "/usr/lib/python3/dist-packages/kivy/core/camera/camera_pygst.py", line 15, in <module>
    import pygst

opencv - ImportError: No module named 'cv'
  File "/usr/lib/python3/dist-packages/kivy/core/__init__.py", line 59, in core_select_lib
[...]

PR #5267, opened 2017-07-08
Remove unregistered camera_gi provider

bionoid:
gi and pygst providers were removed (#4224) but camera_gi still remains in the repository. 

It was apparently disabled in favor of pygst (also removed) in https://github.com/kivy/kivy/commit/de22f3aef91edad371ab4ea38dff7fd68212e5e4 (2013), and https://github.com/kivy/kivy/commit/61fd64d98854804a8af1d893d5a5ac69b63fd210 removes the last relic in ``__init__.py``. 

I guess this should just be removed, and Kivy no longer supports camera via gstreamer at all? 

**Edit:** or is it intended that you import this explicitly if you want to use it? It seems like pygst and the others would have been left as well if that was the philosophy, plus the file has no code updates since late 2013.

(reported by user C_D on IRC)


dessant:
Yes, the gi camera provider was left there because there was no other alternative left for gstreamer.
There is also a leftover at https://github.com/kivy/kivy/blob/68793c0669ef14c78768a3560e2012449c966693/kivy/__init__.py#L223.

It could be tested if it still works with a recent Gstreamer version, if not, let's remove it. Otherwise it could be listed again in `camera/__init__.py`.

LukeBiddle:
Just an FYI, I've been using camera_gi successfully with gstreamer version 1.10.5 (latest available for Fedora 25) for a couple of months now. It's been perfectly solid for me, and I've found the code is easy to extend as needed.

tito:
Yes, don't remove it :)

Issue #5269, opened 2017-07-09
Deceptive pixel format logic in _image_sdl2.pyx load_from_surface()


bionoid:
### Description

The intended happy path of converting to BGR888/ABGR8888 if ``image.format.BytesPerPixel`` is 3 or 4 is not reached. Instead, [this is the default execution path](https://github.com/kivy/kivy/blob/1.10.0/kivy/core/image/_img_sdl2.pyx#L76-L87). It seems like BPP=3 will not do any conversion, or incorrectly as ABGR if Rshift>Bshift. 

### Code and Logs

```python
fmt = 'rgb'
if fmt not in ('rgb', 'rgba'): # <------- suspect "not"
    if fmt == 'rgb':
```

https://github.com/kivy/kivy/blob/1.10.0/kivy/core/image/_img_sdl2.pyx#L63-L64

tito:
I was drunk? O_o

bionoid:
@tito do you remember the purpose of the Rshift>Bshift test? Handle bgr, or some endian-issue, or something else? 

I've rewritten this function, will finalize & submit a pr, maybe easier to discuss that

tito:
I think it was an endian issue on iOS/android.

PR #5274, opened 2017-07-13
Refactor _img_sdl2 load_from_surface

bionoid:
See #5269

* Should now always return rgb or rgba pixel data (as noted in the original FIXME comment, the old code could return mismatching data)
* Saves 1 byte per pixel for some <16 bit formats, previously converted to RGBA even when no alpha
* We can save 1 byte per pixel for indexed format without alpha by figuring out the problem with the palette code (see FIXME). For now they are converted to rgba.

I think this is better but it certainly needs review/testing. Not sure about the original intention of the shift test, I suspect it is no longer needed.


kivy/core/image/_img_sdl2.pyx
```diff
@@ -42,53 +42,58 @@ def save(filename, w, h, fmt, pixels, flipped):
...
+    elif image.format.format == SDL_PIXELFORMAT_RGB24:
+        fmt = 'rgb'
+    elif image.format.Amask:
+        fmt = 'rgba'
+        target_fmt = want_rgba
+    elif image.format.palette != NULL:
+        # We want RGB24 here, and rgba only if the palette has alpha
+        fmt = 'rgba'
+        target_fmt = want_rgba
+# FIXME: ... but this breaks compile, "SDL_Color has no member unused"
```
KeyWeeUsr: It's a problem with `unused` naming because the Cython parts used only `rgb` for some reason, but leaving alpha completely probably broke the compilation, therefore `cdef Uint8 unused`. See [this line](https://github.com/kivy/kivy/pull/5252/files#diff-d18c5eedfc97dd5c3934914bc1e8bc8aR92) of my shaped window PR to see the fix. It should work just fine after it.

kivy/core/image/_img_sdl2.pyx
```diff
@@ -42,53 +42,58 @@ def save(filename, w, h, fmt, pixels, flipped):
...
-            image2 = SDL_ConvertSurfaceFormat(image, pf.format, 0)
+        fimage = image
+        if target_fmt != 0:
+            image2 = SDL_ConvertSurfaceFormat(image, target_fmt, 0)
             if image2 == NULL:
-                return
-
-            fimage = image2
-        else:
-            if (image.format.Rshift > image.format.Bshift):
```
tito: Are you sure about removing this part? It need to be tested on both desktop and iOS+Android.

tito: Actually, the previous comment is self-explainatory:

> the format might be 3 or 4, but it doesn't mean it's rgb/rgba.
> It could be argb, bgra etc. it needs to be detected correctly.

Which is exactly the purpose of the R>B, to test the order i think (and not little/big). But i may be wrong.

tito: And please readd the fixme to have a comment/fixme for hardware that support natively argb/bgra, in case of for later contribution.

bionoid: I don't **think** it's needed, but not 100% sure, it absolutely needs testing. The replacement is the use of a variable want_rgba depending on platform endianness.

Manually building the SDL_PixelFormat structure like this also seems weird to me (it is documented as read-only). Why not use SDL_ConvertSurfaceFormat with the correct SDL_PIXELFORMAT_* (pr does this), or SDL_AllocFormat, or SDL_MasksToPixelFormatEnum? It doesn't seem to take platform endianness into account at all, so my impression is that if it fixes one case, it must break another.

bionoid: > And please readd the fixme to have a comment/fixme for hardware that support natively argb/bgra, in case of for later contribution.

I added bgr+bgra (supported by ImageData), and a comment that I think covers this. It doesn't seem like a FIXME with the format detection now (hopefully) fixed, so I went with NOTE. 

> Which is exactly the purpose of the R>B, to test the order i think (and not little/big).

Agree, seems most likely, and I think the pr is "more correct" for handling this as well as endian-ness. Quick look at the current implementation:

* All < 24bpp formats are converted to ABGR8888
  * We want RGBA8888 on big-endian here, right?
  * [See the ***32 aliases](https://github.com/spurious/SDL-mirror/blob/release-2.0.5/include/SDL_pixels.h#L265-L275) which needs a recent sdl2.

* 24/32bpp formats are returned as-is, with format 'rgb' or 'rgba', **unless** Rshift>Bshift is true.
  * Example 1: ARGB2101010 is incorrectly returned as 'rgba' (unless the Rshift>Bshift test is true, in which case it is converted to ABGR8888 - probably, see below)
  * Example 2: RGB888 (which is **32-bit** SDL_PACKEDORDER_XRGB) will be returned as 'rgba'. Not 100% sure how that works out byte-wise, but seems broken

So about the Rshift>Bshift test:

* When true the pixel format is converted via custom SDL_PixelFormat. As I pointed out above, I am confused about this approach.
  * What are the consequences of pf.[RGBA]shift = 0? Of BytesPerPixel = 0? Is it luck or planning that the called function gives the desired result in this case??
  * Seems like it will convert to ABGR8888 in practice (unless the 
custom pixel format serves a purpose I don't understand)
  
The update should do:

* 24-bit RGB and BGR used as-is (RGB24, BGR24)
* 32-bit RGBA and BGRA used as-is (want_rgba, want_bgra)
* Everything else is converted to want_rgba (if alpha) or RGB24 (no alpha)

bionoid:
I've done some tests, and confirmed that some formats are returned incorrectly by the existing implementation. But there is another issue caused by SDL surface pitch implementation. Refresher: pitch is the number of bytes used to store a single row of pixels. So with 32bits/4 bytes per pixel, a 100x100 image has pitch 100*4 = 400, whereas an 8-bit color format has pitch 100. 

SDL pitch-aligns pixel data to 4 bytes, so when using a 32-bit format, that works perfectly, but 24-bit formats not so much. The simplest example of this is to load 1x1 png without alpha, it gives you \xRR\xGG\rBB\x00 (pitch=4, not 3). Same for 2x1 (pitch=8), 3x1 (pitch=12) and finally a perfect match at 4x1 pitch=12. Only this perfectly-aligned surface is safe to copy pixels from if you expect RGBRGBRGB continuously.

I rigged a test that loads a png, converts the surface to many pixel formats, and load them with the function under scrutiny. I then compared the pixel output to manually input pixel data. It's hacky, will try to make it a real test at some point. But the results are clear -- the old load_from_surface fails. 

Below is listed the results of pixel formats. "_source_" is the surface returned by load_from_filename, this surface is converted to a long list of pixel formats and then loaded again.

# Existing implementation failures

```
 b'SDL_PIXELFORMAT_BGR24': {'fail': 9, 'pass': 0},
 b'SDL_PIXELFORMAT_BGR888': {'fail': 9, 'pass': 0},
 b'SDL_PIXELFORMAT_BGRA8888': {'fail': 9, 'pass': 0},
 b'SDL_PIXELFORMAT_BGRX8888': {'fail': 9, 'pass': 0},
 b'SDL_PIXELFORMAT_RGB24': {'fail': 8, 'pass': 1},
 b'SDL_PIXELFORMAT_RGB565': {'fail': 7, 'pass': 2},
```

# PR failures

```
 b'SDL_PIXELFORMAT_RGB565': {'fail': 6, 'pass': 3},
 ```
 
 This format fails due to precision loss
 
 ```
1x1green.png: b'SDL_PIXELFORMAT_RGB565' FAILED ===> rgba
1x1green.png: b'SDL_PIXELFORMAT_RGB565' Got: b'\x00\xfe\x00\xff'
1x1green.png: b'SDL_PIXELFORMAT_RGB565' Want: b'\x00\xff\x00\xff'
```

bionoid:
It seems like pitch-aligned pixel data is correctly supported after all, but [converting from bgr -> rgb](https://github.com/kivy/kivy/blob/master/kivy/graphics/img_tools.pxi#L47-L58) doesn't take it into account (will include pitch-alignment bytes as pixel data). [That is called before](https://github.com/kivy/kivy/blob/master/kivy/graphics/texture.pyx#L919-L921) handling pitch alignment further down the function.

bionoid:
Summary of the situation:

* The previous commit that returned only pitch-aligned textures has been reverted, it was a misunderstanding rgd pitch alignment. The internal bgr -> rgb conversion seems like it needs adjustment, as mentioned in previous comment.

* The primary failure mode of original implementation is with images loaded in BGR24 format (BMP2 and BMP3 at least). SDL_PIXELFORMAT_BGR24 should have been returned as 'bgr', but instead:
  * It is initially tagged as fmt='rgb' since bpp=3
  * It then ends up in the conversion process to BGRA8888 (via shift test)
  * But the fmt is not updated to reflect conversion, so it is returned as 'rgb' with rgba pixels
  * For example, loading 1x1 green pixel in BMP2 format gives you ``\x00\xFF\x00\xFF``. The final byte is alpha; you would expect it to be a ``\x00`` pitch alignment byte. 

* The secondary failure mode is wrt binary alpha
  * Loading PNG with binary alpha becomes SDL_PIXELFORMAT_RGB24
  * This would be returned "correctly" as 'rgb', but the color key is ignored
  * We should convert this to rgba.
  
* Some other surface formats also loaded incorrectly, but it doesn't seem like IMG_Load actually uses them under normal operations (so it does not affect image loading). Either way, this should be fixed now.

This should be safe to merge, I have tested it with generated image files and verified that the pixel data is returned as expected.

bionoid:
I should add that this still fails for loading PNG24/PNG48 with binary transparency **and with** background color. Transparency is ignored in this case, I'm not sure why (this didn't work before either). 

PR fixes loading of PNG24/PNG48 binary transparency **without** background color

bionoid:
Closing this, reopening a clean PR

PR #5280, opened 2017-07-15
Add note to docs regarding App.on_config_change

bionoid:
The ``on_...`` name and App being an EventDispatcher leads you to believe this is an event name, the docs even described it as such. But it's not actually an event, [the method is just called outright](
https://github.com/kivy/kivy/blob/1.10.0/kivy/app.py#L1019-L1020).  

It doesn't pass the app instance, so converting to dispatch is a breaking change

I guess this is the correct fix?


bionoid:
My bad, that is not a breaking change I think

PR #5281, opened 2017-07-15
Make App.on_config_change an event

bionoid:
As noted in #5280, the docs are misleading here. It seems okay to just make it an event?


matham:
I kinda dislike having events just for the kivy app settings built into App, but I guess it's already there, just not in event form so I guess that's ok.

Issue #5277, opened 2017-07-13
Kivy prevents  kivy.support.install_twisted_reactor in Python 3


rjungbeck:
### Versions

* Python: 3.6.1
* OS: any
* Kivy: 1.10.0
* Kivy installation method: pip
* Twisted: 17.5.0

### Description
Kivy raises an exception  after finding out, that it is running under Python 3.
But the required twisted functionality (_threadselect) seems to be available now.

At least my app works (when doing a really ugly one time  monkey patch for  sys.version_info)

```` python
realVersionInfo=sys.version_info

class DummyVersionInfo(object):
    def __getitem__(self, index):
        sys.version_info=realVersionInfo
        return 2
sys.version_info=DummyVersionInfo()

try:
    from kivy.support import install_twisted_reactor
    install_twisted_reactor()
except:
    pass
````

KeyWeeUsr:
ref #5025 and https://github.com/kivy/kivy/issues/4275#issuecomment-301281689

There was a new Twisted [release](https://github.com/twisted/twisted/releases) on 11th June, so it might even work. Please try to run their test suite if everything works (even if that monkeypatch if you want to) and if there are no errors we can remove the NYI check.

Is this monkeypatch needed only on the Kivy side (for installing the reactor) or is it necessary even at runtime? If the latter, it may and most likely even will break things.

rjungbeck:
I am using that twisted release (17.5.0) but  can only check my app (which includes a simple web server).

My app works  with the monkey patch (that only circumvents your check). It is only needed exactly once  during the install_twisted_reactor_call(). The app  does not work when I modify sys.version_info for any longer (like the full duration of install_twisted_reactor)

KeyWeeUsr:
Ok, so then we could probably add it to our file if it's still some kind of issue with twisted stable version. Or maybe we're missing the new name for the module or something.

Anyway, you don't need to test your app, there's a folder called [`test`](https://github.com/twisted/twisted/tree/trunk/src/twisted/test) in the package together with [instructions](https://github.com/twisted/twisted#unit-tests) for how to use it. I'm not sure if it's [included](https://github.com/twisted/twisted/blob/trunk/src/twisted/python/_setup.py) in the package by default though, so you might need to clone their stable version and test then from source.

KeyWeeUsr:
Closed with #5285

PR #5026, opened 2017-02-27
Update Twisted Framework Example to Py3

manthansharma:
Update Twisted Framework integration example to py3 with maintaining support to py2. 

- `_threadedselect.py` has been ported to py3 will release with next update. https://github.com/twisted/twisted/pull/717

PR #5285, opened 2017-07-17
fix install_twisted_reactor for python3 (_threadedselect is now inclu…

tshabs:
…ded in twisted for py3)

see also:

- https://github.com/kivy/kivy/issues/5277
- https://github.com/kivy/kivy/pull/5026

### DIFF
--- a/kivy/support.py
+++ b/kivy/support.py
@@ -166,11 +166,6 @@ def install_twisted_reactor(**kwargs):
         put the twisted distribution (and zope.interface dependency) in your
         application directory.
     '''
-    import sys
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
@tshabs does it work for you without the version fix or is it still necessary?

tshabs:
@KeyWeeUsr I don't need to do any of this to make it work.  Works perfectly on 3.5.2 & Twisted 17.5.0

KeyWeeUsr:
Tested, both client & server worked correctly. Thanks for the patch! 🍡

PR #5300, opened 2017-07-23
Refactor _img_sdl2.pyx load_from_surface (cleanup)

### LINKED ISSUE #5269: Deceptive pixel format logic in _image_sdl2.pyx load_from_surface()

bionoid:
Closes #5269. This supersedes PR #5274, squashed to avoid commit spam + some minor cleanup (release gil during conversion, change order of some tests, added comments)

Summary of changes:

* Fix loading of SDL_PIXELFORMAT_BGR24 surfaces (BMP2, BMP3 images)
* Fix loading of (some) PNG formats with binary alpha
* Fix byte order for big-endian arch (untested)
* Some other surface formats were incorrectly handled, but not used by SDL_Image

Known issues:

* Some PNG48 images with binary transparency are still not loading correctly
* 8/16-bit grayscale PNGs are not supported. This seems like an issue with SDL_Image, it returns unexpected SDL_PIXELFORMAT_RGB565 instead of expected grayscale palette.

Details:

* Log of PR against test images: [link](https://gist.github.com/bionoid/ecb09b6db036faac77e4618fa9d82b2f) (100% pass)
* Log of original implementation with same images: [link](https://gist.github.com/bionoid/1738bce4170df332350a530b7a36553d) (AssertionError: _img_sdl2: 100 of 1832 tests failed, these are all BMP2/BMP3 images. Plus unreported failures, see below)
  * Some of the tested ``_BINARY`` (binary alpha) png files loaded here are reported with 'pass' but they don't "really" pass. The test doesn't mandate the returned pixel format, it just verifies that the pixels are consistent with the returned format. So these are  correctly converted to RGB, **but we lost alpha**
  * Hence, the PR test lists all ``_BINARY`` images as 'rgba', while 116 of them are incorrectly listed as "pass" with "rgb" above.

Tested formats (ImageMagick):

* Opaque: TIFF, BMP2, BMP3, PNG, GIF87
* Binary: GIF, PNG8, PNG24, PNG48, BMP
* Full Alpha: PNG32, PNG64, TGA

Notes:

* All formats above were tested with fully opaque pixels
* "Binary" and "Full Alpha" were tested with binary transparency
* "Full Alpha" were tested with the same images as for opaque, but semitransparent.
* TIFF is only tested as opaque, because the test doesn't support premultiplied alpha
* JPG not tested because pixel values are unpredictable (lossy compression)

### DIFF
--- a/kivy/core/image/_img_sdl2.pyx
+++ b/kivy/core/image/_img_sdl2.pyx
@@ -58,57 +58,83 @@ def save(filename, w, h, fmt, pixels, flipped):
     IMG_SavePNG(image, c_filename)
     SDL_FreeSurface(image)
 
-
+# NOTE: This must be kept up to date with ImageData supported formats. If you
+# add support for converting/uploading (for example) ARGB, you must ensure
+# that it is returned unmodified below to avoid converting to RGB/RGBA.
+#
+# FIXME:
+# - Some PNG48 images with binary transparency load incorrectly
+# - (Some?) 8/16 bit grayscale PNGs load incorrectly (likely SDL_Image problem)
 cdef load_from_surface(SDL_Surface *image):
     cdef SDL_Surface *image2 = NULL
     cdef SDL_Surface *fimage = NULL
-    cdef SDL_PixelFormat pf
+    cdef int want_rgba = 0, want_bgra = 0, target_fmt = 0, n = 0
     cdef bytes pixels
 
-    try:
-        if image == NULL:
-            return None
+    if image == NULL:
+        Logger.warn('ImageSDL2: load_from_surface() with NULL surface')
+        return None
 
-        fmt = ''
-        if image.format.BytesPerPixel == 3:
-            fmt = 'rgb'
-        elif image.format.BytesPerPixel == 4:
-            fmt = 'rgba'
+    # SDL 2.0.5 now has endian-agnostic 32-bit pixel formats like RGB24,
+    # but we can't count on that yet (RGB24 is available since 2.0)
+    if SDL_BYTEORDER == SDL_BIG_ENDIAN:
+        want_rgba = SDL_PIXELFORMAT_RGBA8888
+        want_bgra = SDL_PIXELFORMAT_BGRA8888
+    else:
+        want_rgba = SDL_PIXELFORMAT_ABGR8888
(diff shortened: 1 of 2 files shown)


tito:
I didn't check the full code of the PR, but is there anything against a merge?

bionoid:
Testing on windows/macos/android/ios remains to be done, although I am fairly confident it will work as they are all little-endian in practice (right?). Big-endian would be nice to test of course, but how? PowerPC? Can a bi-endian ARM be used, if so how? Does anyone have a working system? Is it possible to run kivy in big-endian qemu?

One issue does need clarification; the default conversion to RGB24 if no alpha channel is detected. That is technically best in terms of memory performance, but not obviously the best overall. SDL2 will pitch-align it to 4 bytes, which needs an OpenGL extension for unpacking (or we'll do it in software). But it'll likely be unpacked as 32bpp in the end, so the question is, would it be better to default to RGBA conversion, even if the image has no alpha information? I don't know...

All in all, I consider it a low-risk merge. As noted in previous issues, there are a several problems with the original implementation (the shift test, non-standard SDL_PixelFormat, no colorkey detection, dead code paths from the weird not, ...), and the ImageLoader test case (although far from perfect) speaks in favor of this implementation..

Unless 1.10.1 is planned for release shortly, merging is probably a good way to get it test it more widely?

akshayaurora:
@bionoid any updates on this? is this ready to merger ?

dolang:
Note to myself: try this together with #5785. I think the core dump happened in the SDL2 image provider when doing the screenshot test.

PR #5301, opened 2017-07-23
img_tools.pxi: Support pitch alignment in bgr->rgb conversion

bionoid:
Adds support for pitch alignment in BGR pixel data. I think this is okay, but I didn't actually test it in-place. 

It handles a couple of hand-crafted 3x1, 3x2, 4x1, 4x2 with and without alignment.


bionoid:
Does this need to support 2- and 8-byte alignment as well? SDL2 won't give us that, but I don't know about others.

PR #5302, opened 2017-07-23
Docs: Fetch cython version from setup.py

KeyWeeUsr:
For `1.10.1` is necessary at least `0.25` because of #5220 + we shouldn't write various versions for each install file if we already define the range of supported cython versions in the `setup.py`. Also, the changed flag is preferred (`--upgrade`/`-U` vs `--ignore-installed`/`-I`) as we already use it in some other places.

### DIFF
--- a/doc/sources/conf.py
+++ b/doc/sources/conf.py
@@ -218,6 +218,9 @@ latex_toplevel_sectioning = 'part'
 
 from kivy import setupconfig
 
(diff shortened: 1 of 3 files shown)


KeyWeeUsr:
Later if we'd want to switch to RTD, we might need to update CSS a little, but in current docs it looks the same as `.. code-block::` and the alternatives.

image

PR #5303, opened 2017-07-23
Add support for converting ARGB -> RGBA

bionoid:
Is the ``&src_buffer[1]`` in memcpy okay wrt bytes/memoryview?

### DIFF
--- a/kivy/graphics/img_tools.pxi
+++ b/kivy/graphics/img_tools.pxi
@@ -21,7 +21,7 @@ cdef inline convert_to_gl_format(data, fmt):
     cdef char [::1] view
     cdef int datasize
     cdef str ret_format
-    cdef int i
+    cdef int i, k
     cdef char c
 
     # if native support of this format is available, use it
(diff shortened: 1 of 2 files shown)


bionoid:
Unless I am missing something, ``read_pixel`` doesn't support pitch alignment for 24bpp formats either

https://github.com/kivy/kivy/blob/1.10.0/kivy/core/image/__init__.py#L901

bionoid:
* Does anyone know how to properly test for ``GL_UNSIGNED_INT_8_8_8_8_REV`` support? (which is required for ARGB/ABGR to render as BGRA/RGBA respectively) 
  * It seems like this is part of OpenGL 1.2, but not ES2.0

* Is there something else missing for uploading ARGB/ABGR to GPU? I mean beyond detecting _REV support and using the correct glTexImage2D ``type`` argument?

tito:
@bionoid Check https://github.com/kivy/kivy/blob/master/kivy/graphics/opengl_utils.pyx#L88 - This is where we implement specific check to enable/disable part of Kivy graphics. Just create a new CAP, implement it, then use it.

See how we use this:
- https://github.com/kivy/kivy/blob/master/kivy/graphics/texture.pyx#L492
- https://github.com/kivy/kivy/blob/master/kivy/graphics/texture.pyx#L433
- https://github.com/kivy/kivy/blob/master/kivy/graphics/texture.pyx#L974

kivy/core/image/__init__.py
```diff
@@ -6,6 +6,10 @@
 :class:`~kivy.graphics.texture.Texture`. The raw image data can be keep in
 memory for further access.
 
+.. versionchanged:: 1.10.1-dev0
```
tito: remove the -dev0, if merged, it will be already ok for the next release

tito:
As for "&src_buffer[1]", we do memcpy already the same way, offsetting by 1 don't seem like a big deal.

bionoid:
It's been a while, I forgot the exact problem here. I think it ties in to platform endianness in some way, I don't think there is an actual `GL_UNSIGNED_INT_8_8_8_8_REV` capability to test for. At least I couldn't find it.

matham:
https://github.com/kivy/kivy/pull/5969

PR #5304, opened 2017-07-26
Add ImageLoader test case

bionoid:
**Summary of changes**

Add test case for image loaders, skipped by default. The test works by generating a wide range of image files with predictable pixel values (encoded in filename). These images are loaded using available image providers, and the resulting pixel data is compared to a prediction made from the filename. Only lossless formats are tested.

* Add ``tests/test_imageloader.py`` and ``tools/create-testimages.sh``
* Add ``testimages`` rule to Makefile (populates kivy/tests/testimages)

**Running the test**

1. Install ImageMagick (convert utility must be in path) + pillow, pygame, ffpyplayer, sdl2 to run all tests
2. ``make testimages``
3. Run ``kivy/tests/test_imageloader.py`` as a script to enable debug output to console (py3 recommended, missing some pretty printing for pixel data). Generates lots of output. You can specify providers as arguments to limit which tests are run, ``SDL2``, ``Pygame``, ``FFPy``, ``PIL``, ``GIF``, ..

**Test results**

Numbers will not be reproduced exactly due to randomly generated images, different library versions, etc. Passed = exact match with prediction, skipped = supported extension but loading failed, fail = image loaded, but pixel data mismatch.

* **ImageLoaderSDL2:** 1664 passed, 0 skipped, 182 failed
  * With #5300 applied: 100% passed
* **ImageLoaderFFPy:** 2329 passed, 0 skipped, 336 failed
  * Only issue I've spotted is +1 from expected value, seems like precision/rounding/conversion, F0 becomes F1, DD->DE etc. PNG32 passes and PNG64 fails for the same pixel data in at least one case.
* **ImageLoaderGIF:** 68 passed, 0 skipped, 101 failed
  * I think incorrect byte order in one case, and one seems plain wrong (1x1 fully opaque colored pixel returned as \x00\x00\x00\x00), needs research
* **ImageLoaderPIL:** 1833 passed, 403 skipped, 130 failed
  * Various problems, needs research
* **ImageLoaderPygame:** 1378 passed, 169 skipped, 130 failed
  * Ignores colorkey for some binary transparency (maybe other problems)

**Notes**

* The binary alpha tests allow white and black with 00 alpha as a fully transparent pixel. This could cause some tests to pass incorrectly (esp for small image dimensions). It should be caught by larger images, and I'm not sure if it even happens
* Many of the image loaders return pitch=0
  * Probably only affects rgb/bgr, but is there a reason for using 0 over w*4 for rgba for example?
  * I don't know the reasoning or full extend of the consequences here
  * The test assumes pitch 0 == no pitch alignment (raw bytes)
* magick PCX are unreadable by gimp and kivy providers, not included
* magick DDS (+DXT1, DXT5) are unreadable by ImageLoaderDDS, not included
* magick XV unreadable, not included
* img_sdl2 does not have 'gif' format listed as supported. I guess this is a deliberate priority choice? It's tested explicitly
* PIL lists SGI format support, but fails to load magick SGI files. FFPy reads them, so I left it enabled. Should probably 
* TIFF transparency is not tested, because it uses premultiplied alpha which the test does not support
* JPG is not tested, because lossy compression = unpredictable pixel values

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,7 +35,7 @@ kivy/lib/gstplayer/_gstplayer.c
 kivy/core/camera/camera_avfoundation.c
 kivy/tests/build
 kivy/tests/results
-kivy/tests/testimages
+kivy/tests/image-testsuite
 kivy/garden
 .last_known_portable_deps_hash
 deps.zip
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ kivy/lib/gstplayer/_gstplayer.c
 kivy/core/camera/camera_avfoundation.c
 kivy/tests/build
 kivy/tests/results
+kivy/tests/testimages
 kivy/garden
 .last_known_portable_deps_hash
 deps.zip
--- a/Makefile
+++ b/Makefile
@@ -83,9 +83,9 @@ hook:
 	cp kivy/tools/pep8checker/pre-commit.githook .git/hooks/pre-commit
 	chmod +x .git/hooks/pre-commit
 
-testimages:
-	mkdir -p "${KIVY_DIR}tests/testimages"
-	-kivy/tools/create-testimages.sh "${KIVY_DIR}tests/testimages"
+image-testsuite:
+	mkdir -p "${KIVY_DIR}tests/image-testsuite"
+	-${KIVY_DIR}tools/image-testsuite/imagemagick-testsuite.sh "${KIVY_DIR}tests/image-testsuite"
 
 test:
 	-rm -rf kivy/tests/build
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,10 @@ hook:
 	cp kivy/tools/pep8checker/pre-commit.githook .git/hooks/pre-commit
 	chmod +x .git/hooks/pre-commit
 
+testimages:
+	mkdir -p "${KIVY_DIR}tests/testimages"
+	-kivy/tools/create-testimages.sh "${KIVY_DIR}tests/testimages"
+
 test:
 	-rm -rf kivy/tests/build
 	$(NOSETESTS) kivy/tests
--- a/kivy/tests/test_imageloader.py
+++ b/kivy/tests/test_imageloader.py
@@ -392,6 +392,10 @@ class ImageLoaderTestCase(unittest.TestCase):
         loadercls = LOADERS.get('ImageLoaderTex')
         ctx = self._test_imageloader(loadercls)
 
+    def test_ImageLoaderImageIO(self):
+        loadercls = LOADERS.get('ImageLoaderImageIO')
+        ctx = self._test_imageloader(loadercls)
+
     def test_missing_tests(self):
         for loader in ImageLoader.loaders:
             key = 'test_{}'.format(loader.__name__)
--- a/kivy/tests/test_imageloader.py
+++ b/kivy/tests/test_imageloader.py
@@ -1,0 +1,358 @@
+import os
+import re
+import unittest
+from pprint import pprint
+from kivy.core.image import ImageLoader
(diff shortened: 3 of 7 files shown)


akshayaurora:
@bionoid 🥇 awesome freakin job man.

bionoid:
> we need ImageMagick for that, right? From their travis.yml [...]

It's not a runtime requirement, only a helper to generate image files in advance of the test (they can be downloaded as a zip worst case). This is more of a "report/research tool for advanced users" than an "automated test", I don't think it makes sense to run in travis or by end users, since everything fails. SDL2 passes with the other PR applied, but it's not supposed to, because some grayscale pngs are broken upstream (not provoked by this test). If it's an expected failure, and nobody looks at the report, why spend the cycles?

> In my opinion it'd be better to just add final images + the script that generates them only and not actually run the test with installing ImageMagick

It is intended for debugging/testing/developing image providers/format support in kivy core, so the audience is fairly narrow. There are ~3k/15mb images currently, and it should grow larger (more formats, other encoders, not limited to 1 in w/h, etc). Surely we all agree 15mb of evolving binary data shouldn't live in the main repo. 

So we'd have to pick "just a few" images, but that creates a lot more problems than it solves, since the test's relevance is to a large extent determined by the size of input. From my perspective, this test can evolve into or inspire "something else" that deserves binaries added to the repository. But I don't think it's justifiable now.

KeyWeeUsr:
Makes sense to me now (although I still don't like the wagon of potential deps :D ), but if it's not intended for CI + end-user testing, then I'd completely move it to `tools` as a separate folder with some useful readme. From what I've understood, it's a helper tool for making more providers therefore the file itself would do nothing good in the `tests` folder in the first place unless there are images to test + the upstream fixes their stuff (i.e. like you said "it's an expected failure"). Similar to gles_compat or benchmarking, packaging, etc, I think this would be properly classified as a tool with a good test.

Instead of using it directly from `tests` folder and make it available as a common test you could just run `nose` in the folder you'd put into `tools` and add that command to the `make testimages` target, or just `python test_imageloader.py` as you have `unittest.main()` already in there. Maybe the rest of the team has different opinion, but when I run the test suite, I don't expect that this or that will tell me `skipped` for whatever reason(well, maybe except a missing provider, but still...). Don't get me wrong, I like the PR. I just don't want to see a good test die when we can't generate the input because this or that thing that has nothing to do with the framework itself broke.

bionoid:
Your objection seems like a sign of the obvious truth -- this code is by no means optimally factored for what it does. It doesn't cover as much ground as it should, for example in terms of sensible test cases for end-user systems, but it provides the basis for it (you'd want to reuse most of the low-level stuff). 

Personally I prefer the current location, because it is a kivy core test **spiritually** ;) And I think that is the direction it will evolve. But I'm completely fine with moving it out too, if that is desired

bionoid:
Summary of recent commit:

* Add gimpfu plugin for creating image testsuite
* Add statistics per format, loader, extension etc, ``| grep REPORT``
* Fixed prediction of transparent pixels
* Improved console output
* Made a directory in kivy/tools for this
* Updated test protocol + move docs README.md
* Renamed the target directory and the make rule to "image-testsuite", also renamed the .sh

```
REPORT ImageLoaderFFPy total: ok=16790, skip=885, fail=1553
REPORT ImageLoaderGIF total: ok=153, skip=0, fail=273
REPORT ImageLoaderPIL total: ok=14156, skip=3634, fail=196
REPORT ImageLoaderPygame total: ok=9001, skip=5376, fail=140
REPORT ImageLoaderSDL2 total: ok=14558, skip=590, fail=444
```

(with sdl2 pr applied)

akshayaurora:
@bionoid thanks for all the hard work ! 💯

PR #5305, opened 2017-07-27
img_pygame: Fix loading of binary alpha formats

bionoid:
Fixes Pygame imageloader for binary alpha formats, passes imageloader test 100% after this. It just forces  the conversion process if a colorkey is present, which gives us an alpha channel


akshayaurora:
@bionoid thanks 👍

PR #5311, opened 2017-07-30
#5146: Measure the OpenCV camera's frame rate and invert it.

jkankiewicz:
From #5147:
> The fps returned by the driver should not be trusted, a better approach is to measure the frame rate when the stream begins, then limit the texture refresh rate according to the measured frame rate.

### DIFF
--- a/kivy/core/camera/__init__.py
+++ b/kivy/core/camera/__init__.py
@@ -59,6 +59,7 @@ class CameraBase(EventDispatcher):
         self.stopped = kwargs.get('stopped')
         self._resolution = kwargs.get('resolution')
         self._index = kwargs.get('index')
+        self._frame_sample_size = kwargs.get('frame_sample_size')
         self._buffer = None
         self._format = 'rgb'
         self._texture = None
--- a/kivy/core/camera/camera_opencv.py
+++ b/kivy/core/camera/camera_opencv.py
@@ -8,8 +8,12 @@ OpenCV Camera: Implement CameraBase with OpenCV
 
 __all__ = ('CameraOpenCV')
 
+import functools
+import math
+import operator
 import time
 from collections import Counter
+from fractions import Fraction
 from kivy.logger import Logger
 from kivy.clock import Clock
 from kivy.graphics.texture import Texture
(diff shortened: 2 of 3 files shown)


jkankiewicz:
This is intended to resolve #4828 and #5146.

tito:
I just discovered the issue with your PR.

Measuring the FPS because you don't trust opencv is just a little bit too far? From what i read, if the FPS is < 1, as you say, it's a driver issue. Why trying to measure it yourself?
Maybe the fast and working fix could be that if fps < 1: fps = 1 / fps. In your bug report, it would be 30 FPS.

The real way to fix it would be fix is to report it into OpenCV team or the camera driver itself. But in my opinion, this PR is going to far, and it is not reliable either (if Python is slow, thread stuck, etc, FPS will be incorrect.)

tito:
And that's what i did already, except i put 30, instead of self.fps. Why your initial patch was not good?

jkankiewicz:
> The real way to fix it would be fix is to report it into OpenCV team or the camera driver itself.
I agree that that would be ideal but [the Satya Mallick article](https://www.learnopencv.com/how-to-find-frame-rate-or-frames-per-second-fps-in-opencv-python-cpp/) @dessant referenced seems to indicate that OpenCV will always force the application to measure the camera's frame rate itself.
> Why your initial patch was not good?
@tito, There seems to be a difference of opinion between yourself and @dessant regarding how this should be handled.

dessant:
@tito, besides the different value types, which could be handled by Kivy, some webcam drivers will not report the frame rate at all, or return a lower value.

This feature could also be behind an option, for applications where you do not control the hardware.
And otherwise always measure the frame rate when OpenCV does not return it.

jkankiewicz:
> This feature could also be behind an option, for applications where you do not control the hardware.
And otherwise always measure the frame rate when OpenCV does not return it.

@dessant, I've already added the `frame_sample_size` property for specifying the number of frames to sample for the frame rate measurement. I could make it skip the measurement process altogether and instead trust the frame rate value returned by OpenCV when that property is assigned `None`.
How does that sound?

dessant:
@jkankiewicz, I think that would be a good solution, but it should default to None for now.

annemariet:
Hi! I'm having the fps bug as well. What's missing from the current PR? Can I be of some help?

seanodonnell:
I'm bitten by this as well. Is there any update on @jkankiewicz work? I would be happy to pitch in if any help is needed.

Issue #4648, opened 2016-09-30
Video can seek only to position that multiple of five seconds


germn:
**Code:**

```
import kivy
kivy.require('1.9.1')

from kivy.app import App
from kivy.uix.videoplayer import VideoPlayer


class MyVideoPlayer(VideoPlayer):
    def __init__(self, **kwargs):
        self.source = 'Tavern Brawl.mp4'  # Some video here.
        self.state = 'play'
        super().__init__(**kwargs)


class MyApp(App):
    def build(self):
        return MyVideoPlayer()


if __name__ == '__main__':
    MyApp().run()

```

**Video file I used:** 
[https://yadi.sk/i/WGeoND21vw8yY](https://yadi.sk/i/WGeoND21vw8yY)

**How to reproduce:**
Try to click on player time bar. Starting with fifth second position video position can be changed only to multiple of five seconds: _0:01-0:05_, _0:10_, _0:15_, _0:20_, _0:25_, etc. I can't seek to _0:17_, for example.

This behavior related to [seek](https://kivy.org/docs/api-kivy.uix.video.html#kivy.uix.video.Video.seek) function in general.

**Kivy debug log:**
http://pastebin.com/gVWtfB1Z

**System:**
Windows 10 x64
Python 3.4.4
Kivy 1.9.1

matham:
Try using ffpyplayer? In general this is because when the video is compressed key frames were probably added every 5 seconds. So it may not be able to seek between key frames. I didn't know gst suffers from this, but the solution then is to quickly seek until the correct time one you jumped to the closest 5 second mark.

germn:
Thanks for answer!

> Try using ffpyplayer?

I've just started with Kivy and didn't use anything then standard video widget, but I'll try this (if I'll understand how).

> the solution then is to quickly seek until the correct time one you jumped to the closest 5 second mark

I didn't understand this part. How can I «seek until correct time»? `seek` function doesn't help:

```
self.video.seek(0.07)  # I'm getting at 5 sec. position
self.video.seek(0.08)  # 5 sec.
self.video.seek(0.09)  # 5 sec.
self.video.seek(0.10)  # 5 sec.
self.video.seek(0.11)  # 5 sec.
self.video.seek(0.12)  # 5 sec.
self.video.seek(0.13)  # 5 sec.

self.video.seek(0.14)  # 10 sec.
```

BaggioMarco:
[...]

PR #5313, opened 2017-08-01
ffpyplayer video: better video seek

germn:
Right now changing position of video with ffpyplayer is a bit buggy. Audio stream will seek correctly, while video stream will be slowed or speeded for few seconds before both streams became synchronized.

This PR fixes it making seeking much nicer. While ffpyplayer [doc](https://matham.github.io/ffpyplayer/player.html#ffpyplayer.player.MediaPlayer.seek) says this method is "inaccurate", on practice it works better given us very close position to wanted and synchronized streams momentarily.

I tested it on Windows and Android on different videos for a long time.


matham:
The problem with this is that some videos only have I-frames at regular intervals of e.g. 5 or 10 seconds. When `accurate` is False, it'll only seek to the closest I-frame which means that you may be as much as 5s or 10s away from your seek point. Try it with videos with large distances between their I-frames.

germn:
@matham, yes, I know it, but I think it's better to use inaccurate seek on practice. When we seek accurately we will wait same 5-10 second seeing slowed/speeded video stream. It's confusing, I'm not sure it worth it. 

It's debatable issue, but I can notice that `gstreamer` also [uses](https://github.com/kivy/kivy/issues/4648) inaccurate seeking and we'll will at least get similar behavior for different providers.

Long story short, I like inaccurate seeking much more. Feel free to accept or reject this PR.

**P.S.** Sorry if I'm being annoying, couldn't you also take a look when you have time at [this](https://github.com/kivy/kivy/pull/5053) ffpyplayer-related PR? I use it for a long time and it works fine for me.

matham:
Ok, a better solution then is to add a `def seek(self, percent, precise=True):` and save the `precise` flag in the queue as a tuple along with `percent` and use that. That way, the current behavior remains default but you can overwrite it.

Remember to add docs for the meaning and implication of using the precise flag though.

germn:
`precise` flag added.

kivy/core/video/video_ffpyplayer.py
```diff
@@ -255,11 +255,13 @@ def _next_frame_run(self):
         while not self._ffplayer_need_quit:
             seek_happened = False
             if seek_queue:
-                vals = seek_queue[:]
```
matham: The reason for copy and then delete is because of threading issues. If you get the last value and then delete all elements, because this runs in a second thread, in between the get and delete a new seek may have been added from the other thread. So you end up losing that seek request. The way it was before (copy and then delete the number of copied elements) is thread safe so please change it back to that.

PR #5320, opened 2017-08-08
Proof of Concept: Pango + fontconfig/freetype2 text provider

bionoid:
I added comments and left some unused code for easier review

* Pango can do lots of stuff, this is very basic. Approach, direction, etc needs to be discussed.

* Pango does not have an API for loading a TTF file from disk, so this "solution" will only work with a fontconfig backend, [link](https://bugzilla.gnome.org/show_bug.cgi?id=347237). and some [practical details](http://mces.blogspot.no/2015/05/how-to-use-custom-application-fonts.html) 

  * Win32 / coretext will therefore require the font to be added to the system, and looked up via a font description. Alternatively use fontconfig+ft2 for these platforms as well?

* Pango markup is (ab)used for bold/underline/... using attrs and pango_layout_set_text() would probably be faster

* There is no integration with kivy for new features - "text_direction" and "text_language" properties should be added, maybe also a "font_features" for css-style features (supported by pango). Plus suitable config options, etc.

* Some label options may not be correctly supported, or at all. I did some tests, but probably not enough..

  * I can't find any relevant way to implement font_hinting/font_kerning properties like SDL2 does. Some commented code shows the approach for *selecting font files* based on this, but not adjust how it's rendered once a font is selected. It may need support for grouping related font files elsewhere in Kivy, or some other bigger change. Or it may not be worth the trouble..

  * outline_color/outline_width not supported by Pango, must be hacked somehow (not on my agenda)

  * Probably other things I forgot ...
  

Tested on Python 2.7.13 and 3.5.3 with Cython 0.26 on debian 9.


kivy/core/text/_text_pango.pyx
```diff
@@ -0,0 +1,372 @@
...
+cdef class ContextContainer:
+    cdef PangoContext *context
+    cdef PangoFontMap *fontmap
+    cdef PangoFontDescription *fontdesc
+    cdef PangoFontMetrics *metrics
+    cdef PangoLanguage *metrics_lang # do not free, see docs
+    cdef PangoLayout *layout
+    cdef FcConfig *fc_config
+
+    # Called explicitly on malloc fail to release as fast as possible
```
tito: This can be dangerous as you don't have any __cinit__, and so value are not initialized at all. Then __dealloc__ can crash

kivy/core/text/_text_pango.pyx
```diff
@@ -0,0 +1,372 @@
...
+# Add a contextcontainer to cache + keep max 64 open fonts
+cdef inline _add_pango_cache(unicode fontid, ContextContainer cc):
+    global kivy_pango_cache, kivy_pango_cache_order
+
+    cdef unicode popid
+    while len(kivy_pango_cache_order) >= 64:
+        popid = kivy_pango_cache_order.pop(0)
+        # FIXME: I think this is safe? but leaving out for now
+        #kivy_pango_cache[popid].__dealloc__()
+        del kivy_pango_cache[popid]
```
tito: Calling yourself __dealloc__ is not safe. Del is :)

kivy/core/text/_text_pango.pyx
```diff
@@ -0,0 +1,372 @@
...
+
+    # Creat a new context
+    cdef ContextContainer cc = ContextContainer()
+
+    # Create blank FcConfig (fontconfig), and load the TTF file
+    cc.fc_config = FcInitLoadConfig()
+    cdef bytes filename = options['font_name_r'].encode('UTF-8')
+    if FcConfigAppFontAddFile(cc.fc_config, <FcChar8 *>filename) == FcFalse:
+        Logger.warn("_text_pango: Error loadinging font '{}'".format(filename))
+        cc.__dealloc__()
```
tito: Nop, don't call it yourself. Just don't use it anymore, and leave it to GC later (cython will call the __dealloc__ itself). CF http://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#finalization-method-dealloc

tito: The fix is just "return" (None implicit, so cc will not be used anymore therefore goes into garbage collection)

kivy/core/text/_text_pango.pyx
```diff
@@ -0,0 +1,372 @@
...
+    cdef unsigned long maxpos = final_w * final_h
+
+    # Sanity check that we don't go out of bounds here, should not happen
+    if offset_fixed + ((h-1)*offset_yi) + ((h-1)*w) + w - 1 > maxpos:
+        Logger.warn('_text_pango: Ignoring out of bounds blit: final={}x{} '
+                    'x={} y={} w={} h={} maxpos={}'.format(
+                    final_w, final_h, x, y, w, h, maxpos))
+        return
+
+    with nogil:
```
tito: I wonder the impact of nogil at this point. Just a reflexion, i know we do it from times to times, but i'm thinking that by the time we need to acquire, would we wait more than just keep it?

tito: Maybe you could put the whole function nogil, but then no Logger

tito:
Actually, i don't approve the PR (my bad) just because of the __dealloc__. This must be fixed before merging.

kivy/core/text/_text_pango.pyx
bionoid: The pointers are guaranteed to be NULL by Cython - cinit not needed here

http://cython.readthedocs.io/en/latest/src/reference/extension_types.html

"By the time __cinit__() is called… All C-level attributes have been initialized to 0 or null."

bionoid:
Thanks for review @tito and happy new year! 

In my opinion this is **not** mergeable, there are too many issues to clarify. Specifically we need to explore the FontConfig stuff more, it seems extremely hacky to issue FcInitLoadConfig and a separate FcConfig for each font. **Surely** there must be a better way than this, and the substitute callback didn't work for me, it needs research.

Also I don't think your assessment of the dealloc use is correct. As I commented above, there is no need for cinit since the pointers are guaranteed to be NULL. Just imagine that I cloned the dealloc method to a duplicate `def premature_dealloc(self):`. This would be called on malloc failure instead of an explicit call to `__dealloc__`. It is exactly the same, just saving the duplicated method, and the code doesn't do any of the things your doc link advice against (afaik). The point of this is just to release memory ASAP, so that there is room to even make it to the next GC cycle (though, I don't know how this works under the hood, it is possible that a return would dealloc it immediately). Anyway, I'll change it if that's desired.

> I wonder the impact of nogil at this point. Just a reflexion, i know we do it from times to times, but i'm thinking that by the time we need to acquire, would we wait more than just keep it?

I have no idea about this; from what I have seen in the wild, it seems nogil is commonly used for less demanding code blocks than the blit. Ideally we should benchmark it etc of course, but not on my agenda at the moment

bionoid:
Note, I force-pushed a rebased branch here. Recent commits add some basic integration with Kivy for features. The major outstanding issues as I see it:

* Font fallback. Not obvious how to deal with it. We really do have it for "free" currently (by using pango's high-level APIs), the problem is that only one font can be loaded per context (if we want to be sure that we reference the correct custom loaded font). There are several options for dealing with this, but not clear which is the best tradeoff.

* Bidirectional text. This "works" now (afaik, can't really test it), but Kivy doesn't know that the direction is switched (which implies a switch of alignment). So RTL text is incorrectly left-aligned.

* Use of pango markup, this needs to use attrs directly. Currently it's like quadruple work, first parsing kivy markup, rebuilding the string as pango markup, which is then parsed into an attr list. This does a lot of unneccesary work. 

* Unicode, I'm pretty sure there are issues with py2

* Library versions; there is at least one issue with freetype2, I use `FT_Bitmap_Init()` but in older versions it's `FT_Bitmap_New()` (now deprecated). Not sure how to correctly determine which one to use.

* Maybe the provider should be called `text_pangoft2`, since it's not clear if it can be practically extended for other backends.

And I probably left out a bunch of stuff, but..

bionoid:
Reshaping, font fallback and alignment is now working to some degree, with brand new caveats and pitfalls!

* [TextInput horizontal alignment demo](https://i.imgur.com/rnmeDN7.gif) and [the demo code](https://gist.github.com/bionoid/01483540f2cc767eecd6ea7c5aac797a)
* [Font fallback and alignment demo](https://i.imgur.com/iouCwsg.gif), and [the demo code](https://gist.github.com/bionoid/5cb8ca938d0e1b995fc66e1497225455)

I could really use some help testing + identifying and ironing out issues. @akshayaurora @Kovak and @matham - I'm pinging since you volunteered to mentor the task for GSoC - can you please find some time to test/leave  feedback?

**Build instructions**

1. Install prerequisites
    * **apt install libfreetype6-dev libpango1.0-dev libpangoft2-1.0-0**
    * or otherwise make sure these are available, pangoft2 must be visible to pkg-config for the build to work

2. Grab the code from my kivy fork
    * OPTION A) `git clone https://github.com/bionoid/kivy.git`
        * After this, `cd kivy; git checkout text_pango`
    * OPTION B) `git remote add bionoid https://github.com/bionoid/kivy.git`
        * After this, `git fetch bionoid; git checkout bionoid/text_pango`
    
3. Build Kivy from the `text_pango` branch

I didn't actually do this from scratch lately, but I don't think you need anything else special..

tshirtman:
I just did some testing (on ubuntu). The instructions works perfectly, and the result is very nice, (i tested hebrew and arabic, though i can't read them, the result looks like what google translates does, so it must be correct :)).

I'll try to build on windows to test there.

Congrats on all that work.

bionoid:
@tshirtman thanks for testing, appreciate the feedback!

I've just pushed a commit that adds a public API for font contexts, probably the docs are a little inconsistent at the moment. [Here is an example of using the new FontContextManager API](https://gist.github.com/bionoid/b900fcc61a1d05a3ce9d5d0e6d94f3bf)

png

> and the result is very nice

Well it doesn't look good at small font sizes, at least on my system sdl2 looks better at the default size with roboto. I'm not sure exactly why, it may be down to configuration or the way blending is done in sdl2, or something else.. please confirm if this is the same for you

> though i can't read them, the result looks like what google translates does

At least for Arabic script + Farsi, @RevengeComing has confirmed that it works :) There are **lots** of issues remaining though, like Pango is not used for line breaks, strong/weak cursor will be a hack if it's even possible, etc. For a proper solution. the entire core needs to be refactored, and TextInput heavily modified or rewritten from scratch. 

> I'll try to build on windows to test there.

I am not sure that is a worthwhile goal at this stage, you mean building with ft2+fontconfig right? AFAIK this will not let us use system fonts, so it's a bit of a kludge, although it would enable fallback among custom fonts and reshaping, so still a win... Ideally it should have support for Pango's win32 backend, but at this point I think it makes more sense to add cairo or xft backend (since both use fc/ft2), that will clarify a lot of the issues regarding abstraction before dealing with the required AddFontMemResourceEx hacks and mingw+msvc etc... 

Building it for fc+ft2 on windows doesn't seem entirely straight forward either, though some meson stuff has been added recently, that might make it easier (I didn't look into it at all)... Here is some info: https://wiki.gnome.org/Projects/GTK+/Win32/MSVCCompilationOfGTKStack - it'd need FreeType2, FontConfig, Pango and their dependencies

Cheaterman:
Tested and confirmed working on Python2, including changing the font inside the Label using `[font]` markup from English (ascii) to localized (Arabic) font. Congratz @bionoid :-)

bionoid:
@tshirtman I have dug around a little regarding cross-platform support.. a lot of stuff is not very clear at all,. but I am pretty sure Cairo is the path of least resistance to support Windows/MacOS. It mostly leaves the same question that's been sitting around since day one, "how do you load an application-specific font and then refer to it in a font description?"... PR currently uses FreeType2 + a hack that relies on FontConfig, but it can be made to work on all platforms with ft2 only. It's not ideal, but other options are more difficult and platform-specific. With Cairo, the platform-specific code will be the build scripts + one call to `AddFontResourceEx` (win32) or `CTFontManagerRegisterFontsForURL` (macos), that's in theory of course.

* Desktop platforms will use pangocairo and require Pango, Cairo and FreeType2
* Mobile platforms will use pangoft2 and require FontConfig and FreeType2
  * No system font support, unless we can read the ttfs from filesystem (probably no?), or some enterprising soul wants to contribute new pango backends

So my current plan is to add a Cairo backend, and decouple the family name detection from FontConfig. 

That should put all supported platforms within reach :)

PR #5326, opened 2017-08-11
Default attachment widget for dropdown

Konevka:
The widget argument in DropDown open method becomes optional. Added default behaviour attaches the DropDown object to the parent if no widget is specified.

### DIFF
--- a/kivy/uix/dropdown.py
+++ b/kivy/uix/dropdown.py
@@ -237,6 +237,11 @@ class DropDown(ScrollView):
         if self.attach_to is not None:
             self.dismiss()
 
+        # no parent or widget given, cannot open the dropdown
+        if widget is None and self.parent is None:
+            raise DropDownException(
+                'Cannot open a dropdown list on a hidden widget or without a parent')
+
         # automatically attach to parent if no attach_to given
         if widget is None and self.parent is not None:
             widget = self.parent
@@ -253,7 +258,8 @@ class DropDown(ScrollView):
         self._reposition()
 
         # attach ourself to the main window
-        self._win.add_widget(self)
+        if self.parent is None:
+            self._win.add_widget(self)
 
     def dismiss(self, *largs):
         '''Remove the dropdown widget from the window and detach it from
@@ -380,7 +386,8 @@ if __name__ == '__main__':
             item = Button(text='hello %d' % i, size_hint_y=None, height=44)
             item.bind(on_release=lambda btn: dp.select(btn.text))
             dp.add_widget(item)
-        dp.open(button)
+        button.add_widget(dp)
+        dp.open()
 
     def touch_move(instance, touch):
         instance.center = touch.pos
--- a/kivy/uix/dropdown.py
+++ b/kivy/uix/dropdown.py
@@ -227,7 +227,7 @@ class DropDown(ScrollView):
         if value is not None:
             self.container.bind(minimum_size=self._reposition)
 
-    def open(self, widget):
+    def open(self, widget=None):
         '''Open the dropdown list and attach it to a specific widget.
         Depending on the position of the widget within the window and
         the height of the dropdown, the dropdown might be above or below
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
The patch unfortunately breaks the basic example at the bottom of the file - it doesn't remove the dropdown. Try to make this little change and you'll see for yourself. Also, maybe try to check even with something you've tested before (please paste an example if possible).

```python
if __name__ == '__main__':

    def show_dropdown(button, *largs):
        dp = DropDown()
        dp.bind(on_select=lambda instance, x: setattr(button, 'text', x))
        for i in range(10):
            item = Button(text='hello %d' % i, size_hint_y=None, height=44)
            item.bind(on_release=lambda btn: dp.select(btn.text))
            dp.add_widget(item)
        button.add_widget(dp)
        dp.open()

    def touch_move(instance, touch):
        instance.center = touch.pos

    btn = Button(text='SHOW', size_hint=(None, None), pos=(300, 200))
    btn.bind(on_release=show_dropdown, on_touch_move=touch_move)

    runTouchApp(btn)
```

image

Although `DropDown` is kind of `Popup`-ish (not supposed to be added as a child anywhere), this behavior could actually work if the dismissing was properly fixed.

image

On the other hand, we're getting into a situation where `Spinner` might be a better choice for you as that widget has the proper dismissing. Also, backporting features from `Spinner` to `Dropdown` isn't that much desired as that's actually why those widgets are separated.

Konevka:
Thanks for the review. I made changes so that the previous behaviour is preserved, and when the DropDown is explicitly added on a parent, it still functions properly.

The way I see it, the existence of a parent for a dropdown is an intuitive assumption, as typical DropDown function is to show a range of values when a widget (presumably parent) is selected, and - when one of the values is selected - to change the value/text of a widget (again, presumably the parent). 

About the spinner, which is a good suggestion:
I did not realise spinner was essentially a button with a dropdown in one. That is a simple solution. Still, the name 'spinner' makes it hard to identify as such. What I expect to find under the name 'spinner' would be either a value selection on a spinning scale (like date in iOS), or some element with circular appearance (like the loading animation of round rotating arrow, dots or bars).
It would be great if DropDown documentation linked to the spinner as a shortcut solution. It would be easier to find, and probably more often used.

KeyWeeUsr:
I tested the changes and it works fine from the Python side. However it kind of doesn't from the KV side. It's not a fault of this PR, but from the design of `DropDown`, `ModalView` and `Popup` widgets. Those aren't meant to be added as a child anywhere, therefore if you add it to *any* widget you'll immediately see the `DropDown`.

This can be fixed by a silly `on_parent` + `DropDown.dismiss()`, but in my opinion this introduces an idea that those widgets actually *are* meant to be added as a children. I'll leave the PR open for now, so perhaps other core devs will join the discussion too. Widgets such as `DropDown` use the `Window` as a parent, because they are meant to be **above** of every other widget. Pseudo Z-axis if you will, because Kivy doesn't have a concept of Z-axis (well, at least by default). Some people don't like it, but it keeps things simple.

If a widget supposed to be above all already existing widgets is added as a child to an ordinary widget, it gets a parent and can't be added to the `Window` unless explicitly removed from the current parent. This is basically the silly `on_parent` + `dismiss()`. Also, the instances of these widgets are (or at least should be imo) disposable. For example on each `open()` the instances are new:

```python
from kivy.lang import Builder

#:import Factory kivy.factory.Factory
<Penguin@Button>:
    size_hint_y: None
    on_parent: self.text = str(self)

<FlyingCarpet@Popup>:
    title: str(self)
    Button:
        text: 'Aladout'
        on_release: root.dismiss()

<MrPopper@DropDown>:
    Penguin:
        on_release: root.dismiss()
    Penguin:
        on_release: root.dismiss()
    Penguin:
        on_release: root.dismiss()

BoxLayout:
    Button:
        text: 'Aladin'
        on_release: Factory.FlyingCarpet().open()
    AnchorLayout:
        Button:
            size_hint_y: None
            text: 'Pop!'
            on_release: Factory.MrPopper().open(self)
'''))
```

If you think of these type of widgets only as of some kind of container, you'll get the idea why it's easier to use it this way i.e. `Popup` is just a "template" you can add other widgets into and a `DropDown` you can imagine as a handler for a [this kind](http://i.stack.imgur.com/ti2pG.png) of menus.

Issue #5327, opened 2017-08-13
Kivy won't build with Cython 0.26


inclement:
Trying to build gives an error like the following:

```
  RAN: '/usr/bin/cython ./kivy/lib/vidcore_lite/egl.pyx'

  STDOUT:

Error compiling Cython file:
------------------------------------------------------------
...
    cdef:
        bcm.DISPMANX_DISPLAY_HANDLE_T disp
        bcm.DisplayHandle D
    disp = bcm.vc_dispmanx_display_open( device )
    if disp == 0:
        raise bcm.BCMDisplayException("Couldn't open handle to display")
                ^
------------------------------------------------------------

kivy/lib/vidcore_lite/egl.pyx:608:17: cimported module has no attribute 'BCMDisplayException'

Error compiling Cython file:
------------------------------------------------------------
...
    cdef:
        bcm.DISPMANX_UPDATE_HANDLE_T hdl
        bcm.UpdateHandle U
    hdl = bcm.vc_dispmanx_update_start( priority )
    if hdl == 0:
        raise bcm.BCMDisplayException("Couldn't open handle to update-start")
                ^
------------------------------------------------------------

kivy/lib/vidcore_lite/egl.pyx:619:17: cimported module has no attribute 'BCMDisplayException'


  STDERR:

```

inclement:
I think this doesn't affect most builds because it only happens if vidcore_lite is cythonised. It does affect android builds, since everything is cythonised automatically.

KeyWeeUsr:
What about version `0.25.2`?

Also, try to add `cdef` on [this line](https://github.com/kivy/kivy/blob/4bd65154e2d52b3f9ca71d8fbf59a77d1620d41a/kivy/lib/vidcore_lite/bcm.pyx#L57) and/or `import bcm` after [this line](https://github.com/kivy/kivy/blob/4bd65154e2d52b3f9ca71d8fbf59a77d1620d41a/kivy/lib/vidcore_lite/egl.pyx#L4) because it seems like a really strange "error".

inclement:
I've already made a PR with the latter solution, which seems to work.

11USERNAME11:
So, how to fix this error?

KeyWeeUsr:
@11USERNAME11 It should be already fixed in the master branch.

OptimusGREEN:
downgrading python to 0.25 solved this for me.

`pip install 'cython == 0.25'`

KeyWeeUsr:
[...]

PR #5328, opened 2017-08-13
Fix build with cython 0.26

inclement:
I think this is necessary for the build to work with cython 0.26, probably because the cimport only imports from the bcm.pxd file while egl.pyx needs some things from bcm.pyx. I guess this is a change in the new cython version.

It might be worth checking the details, I've tested this only during the Android build, and it resolves the issue there.


inclement:
Resolves https://github.com/kivy/kivy/issues/5327.

Issue #5333, opened 2017-08-16
ActionBar throws exception when resized


tiktok7:
### Versions
* Python: 3.4.3
* OS: Linux-3.13.0-24-generic-x86_64
* Kivy: 1.10.1.dev0
* Kivy installation method: manual (git repo)

### Description
Steps to recreate the problem
1. Run `python3 ./kivy/uix/actionbar.py`
2. Resize the width down until an exception occurs

### Code and Logs
In `ActionView._layout_random()`, when the width contracts, newly hidden entries are added to `hidden_groups` which is in turn added to `hidden_items`. All hidden items are then added to `self.overflow_group` (so the Help menu item is no longer displayed unless the ellipsis is clicked). Unfortunately this means that the `ActionView.remove_widget()` later fails because it tries to remove the group from the action items list even though it's not contained in it. 

I'm not entirely sure what the correction should be but I verified the problem with the following change (which allows the same code to handle resizes without the same crash happening)

```
diff --git a/kivy/uix/actionbar.py b/kivy/uix/actionbar.py
index 5e4f400..c8161d7 100644
--- a/kivy/uix/actionbar.py
+++ b/kivy/uix/actionbar.py
@@ -491,7 +491,8 @@ class ActionView(BoxLayout):
         super(ActionView, self).remove_widget(widget)
         if isinstance(widget, ActionOverflow):
             for item in widget.list_action_item:
-                self._list_action_items.remove(item)
+                if item in self._list_action_items:
+                    self._list_action_items.remove(item)
 
         if widget in self._list_action_items:
             self._list_action_items.remove(widget)
```

#### Trace
```
[DEBUG  ] [Window      ] Resize window to (296, 600)
[INFO   ] [Base        ] Leaving application in progress...
 Traceback (most recent call last):
   File "actionbar.py", line 779, in <module>
     runTouchApp(float_layout)
   File "/home/will/Projects/kivy/kivy_git/kivy/kivy/base.py", line 502, in runTouchApp
     EventLoop.window.mainloop()
   File "/home/will/Projects/kivy/kivy_git/kivy/kivy/core/window/window_sdl2.py", line 660, in mainloop
     self._mainloop()
   File "/home/will/Projects/kivy/kivy_git/kivy/kivy/core/window/window_sdl2.py", line 405, in _mainloop
     EventLoop.idle()
   File "/home/will/Projects/kivy/kivy_git/kivy/kivy/base.py", line 346, in idle
     Clock.tick_draw()
[...]

PR #5335, opened 2017-08-16
issue #5333 - actionbar throws exception when resized

tiktok7:
prevent attempts to remove items that are not present in the target list (ActionView._list_action_items)
This is a minimal fix for Issue #5333


KeyWeeUsr:
Thanks! :)

PR #5346, opened 2017-08-22
Add unittest for ActionBar

KeyWeeUsr:
Still 2 test cases missing and I'll try to switch it to the `GraphicUnitTest` shape, however when I tried the first time it didn't return correct coordinates for me and the touch wasn't dispatched for some reason. Now I think it should be just a matter of a quick rewrite after adding the two other test cases.

### DIFF
--- a/kivy/tests/test_uix_actionbar.py
+++ b/kivy/tests/test_uix_actionbar.py
@@ -1,0 +1,254 @@
+import unittest
+
+import os
+import sys
+import time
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
As mentioned in 81b578d, Nose fetches a `GraphicUnitTest` object on OSX even if we mock it if the name is "Test" or perhaps has a "Test"/"test" prefix. Not nice.

PR #5349, opened 2017-08-24
Fixed 'is an' -> 'is a' in label docstrings

Fogapod:


KeyWeeUsr:
This syntax of RST:

    :class:`~kivy.properties.OptionProperty`

returns a hyperlink [`OptionProperty`](https://kivy.org/docs/api-kivy.properties.html#kivy.properties.OptionProperty). According to the Oxford dictionary the word [`option`](https://en.oxforddictionaries.com/definition/option) is written with indefinite article `an`, therefore it's not a fix.

PR #5351, opened 2017-08-25
Feature Add Nested Scrollview

### LINKED ISSUE #4444: Nested ScrollView touch transformation is broken for horizontal scrollbars
If you have a nested horizontal ScrollView which uses bars as scroll type, inside a vertical ScrollView, then the touch transformation is broken for the horizontal scroll-bar inside the horizontal ScrollView. It is only noticeable though, if the vertical ScrollView is not starting at `x=0` in window coordinates.

I have added a code snippet to reproduce the error (see below). It is based on kived's nested ScrollView example ([from here](https://gist.github.com/kived/2aa1b6814979340598a7)).

What I also found out was:
- The issue does not occur if the horizontal ScrollView is using content as scroll type.
- The issue does not appear (as already mentioned above) if the vertical ScrollView is starting at `x=0` in window coordinates
- Also the issue does not appear on Kivy 1.8.0. On 1.9.0 and 1.9.1 it appears. (Thanks @bionoid for hinting me to test the different versions.)

```
import kivy

from kivy.app import App
from kivy.lang import Builder

import sys

if 'horizontal' in sys.argv:
    sys.argv.remove('horizontal')
    fmt = {
        'horizontal': 'vertical',
        'x': 'y',
        'y': 'x',
        'width': 'height',
        'height': 'width',
        'col': 'row',
        'row': 'col',
    }
else:
    fmt = {
        'horizontal': 'horizontal',
        'x': 'x',
        'y': 'y',
        'width': 'width',
        'height': 'height',
        'col': 'col',
        'row': 'row',
    }

BoxLayout:
    orientation: 'horizontal'
    BoxLayout:
        orientation: 'horizontal'
        size_hint_x: None
        width: 150
        canvas.before:
            Color:
                rgba: 0, 0, 1, 1
            Rectangle:
                pos: self.pos
                size: self.size
        Label:
            text: "I mess up the nested horizontal scrollbar."
            size_hint_x: None
            width: self.parent.width
            text_size: self.parent.width, None
            max_lines: 5
    BoxLayout:
        orientation: '{horizontal}'
        BoxLayout:
            orientation: '{horizontal}'
            size_hint_{x}: None
            {width}: 0
            Widget:
                id: quarter
            Widget
            Widget
            BoxLayout:
                orientation: '{horizontal}'
                Widget:
                    id: eighth
                Widget
        ScrollView:
            canvas.before:
                Color:
                    rgba: 1, 1, 1, 0.3
                Rectangle:
                    pos: self.pos
                    size: self.size
            scroll_type: ['bars']
            bar_width: 50
            GridLayout:
                {col}s: 1
                size_hint_{y}: None
                {height}: self.minimum_{height}
                {row}_default_{height}: quarter.{height}
                {row}_force_default: True
                Widget
                BoxLayout:
                    orientation: '{horizontal}'
                    canvas.before:
                        Color:
                            rgba: 1, 0, 0, 0.5
                        Rectangle:
                            pos: self.pos
                            size: self.size
                    Widget
                    ScrollView:
                        scroll_type: ['bars']
                        bar_width: 50
                        GridLayout:
                            {col}s: 1
                            size_hint_{y}: None
                            size_hint_{x}: 2.0
                            {height}: self.minimum_{height}
                            {row}_default_{height}: eighth.{height}
                            {row}_force_default: True
                            canvas.before:
                                Color:
                                    rgba: 1, 1, 1, 1
                                Rectangle:
                                    pos: self.pos
                                    size: self.size
                            Widget
                            Label:
                                text: 'Hi!'
                                canvas.before:
                                    Color:
                                        rgba: 0, 0, 1, 0.5
                                    Rectangle:
                                        pos: self.pos
                                        size: self.size
                            Widget
                    Widget
                Widget
'''.format(**fmt))

class TestApp(App):
    def build(self):
        return root

if __name__ == '__main__':
    TestApp().run()
```

sugarlata:
This feature adds elegant handling of nested scrollviews. Overscroll data of nested scrollview get thrown to the parent. Also adds in a new property scroll_nested, which defaults to False.

Testing needs to be done, as far as I can tell it is working, and fixes #4444, while not removing other functionality.

Some documentation might need to be amended also to reflect the scroll_nested property.


sugarlata:
Just checked the test case in #4444. Seems like the issue is still active. (Was fixed for the test case I was using)
I know my way around scrollview now though so will take a stab at fixing.

KeyWeeUsr:
Why implement a new property that enables nesting instead of making it doing so automatically? I mean, I can't imagine myself actually doing this:

    ScrollView:
        scroll_nested: True
        ScrollView:
            scroll_nested: True
            ScrollView:
                scroll_nested: True
                ...

sugarlata:
Problem I ran into was that nested views needed to have the effect class use ScrollEffect, instead of DampedScroll, this was easiest implemented through a new property. Furthermore, it seemed the simplest solution to the problem. Another solution could be to traverse children to see if there are any scrollviews nested, and set the effect class then. Will get back to you.

sugarlata:
@KeyWeeUsr 
Have updated the code as per your suggestions, removing the need for a new declared property, instead the scrollview looks for children, and if there are, changed the private property.

Also corrected some issues with the scroll bars. From what I could tell the bindings in the style.kv file weren't working properly. I removed these binding and created replacements in the scrollview.py file. This has corrected the issue.

Still some testing needed for multiple nestings, i.e. SV in SV in SV, or Multiple nested SVs in one SV.

KeyWeeUsr:
I have an idea, but I'm not really sure if you're in favor of it. I noticed this part:

> Still some testing needed for multiple nestings, i.e. SV in SV in SV, or Multiple nested SVs in one SV.

Recently I've made some changes in the part that's used directly for the UI unittesting of (not only) Kivy widgets - [`common.py`](https://github.com/kivy/kivy/blob/master/kivy/tests/common.py). If you're willing to, you can inspire yourself with [`test_uix_slider.py`](https://github.com/kivy/kivy/blob/master/kivy/tests/test_uix_slider.py) and [`test_uix_actionbar.py`](https://github.com/kivy/kivy/blob/master/kivy/tests/test_uix_actionbar.py)(mainly this) and make a test(s) for the functionality of this PR.

I made some progress with [`test_uix_scrollview.py`](https://github.com/kivy/kivy/blob/master/kivy/tests/test_uix_scrollview.py) for basic SV testing, however I don't think that's enough and adding changes targeting the nesting would definitely make the widget more friendly for future changes as I see you make quite a lot of them. IMO this PR has quite a potential.

Feel free to contact me or anyone from the core team at #kivy-dev if you have questions about the unittesting. I mostly tend to appear there between 6pm to 2am CEST (or at least stalk the [logs](https://botbot.me/freenode/kivy-dev)).

Issue #5341, opened 2017-08-20
Kivy on raspberrypi wrong libs linked


dbgasaway:
### Versions

* Python: 3.5
* OS: Linux raspberrypi (Raspbian Stretch)
* Kivy: 1.10.0
* Kivy installation method: pip

### Description

Apparently the ability to link to the broadcom EGL/GLES libs on rpi using -lEGL and -lGLES was a mistake, as these drivers aren't completely interchangeable with the mesa ones that other applications use and link to the same way. Some comments here:
https://bugreports.qt.io/browse/QTBUG-62216
https://github.com/raspberrypi/firmware/issues/625#issuecomment-230356111
As a result, the libraries /opt/vc/lib/lib{EGL,GLESv2}.so were removed in Stretch, and the default versions became the ones provided by mesa, with the renamed versions continuing to exist.

As a result you get confusing errors like "libEGL warning: DRI2: failed to authenticate" as seen here:
https://raspberrypi.stackexchange.com/questions/61078/qt-applications-dont-work-due-to-libegl
because if you have libegl1-mesa installed, then -lEGL links to the mesa one and not the broadcom one, which fails as they're partly incompatible. Unfortunately, despite being a quick google search away, the solution in the link isn't ideal, and doesn't work after ldconfig overrides the relevant link.

Note that rpi-update does fix this, as it puts these libraries back in /otp/vc/lib/, but according the above links that's not a good solution.

### Steps to reproduce
Upgrade a raspbian jessie image to stretch (I don't know if there are any official stretch images yet).
Install Kivy in a way that involves compiling from source, like pip
Install and run any Kivy application that uses EGL (any fullscreen app?)

### Proposed fix
At https://github.com/kivy/kivy/blob/master/setup.py#L601, change EGL and GLESv2 to brcmGLESv2 and brcmEGL. These are the new library names, and should prevent mesa conflicts.

##### Potential pitfalls
It's possible the renamed libs don't exist in earlier raspbian releases. My up-to-date jessie pi has it, but I don't have access to anything older to test. Maybe the release can be detected dynamically, if backwards compatibility that far in the past is important?

MikahB:
There is indeed an official Stretch image now (both full and Lite) for the RPi - it's the default offered on the download site at the moment [https://www.raspberrypi.org/downloads/raspbian/](url)

[...]

PR #5362, opened 2017-09-01
Raspbian stretch egl library fix

rasmuspeders1:
Proposed fix for replacing the names of libraries EGL and GLESv2 to brcmEGL and brcmGLESv2 which seems like the correct way to get kivy to use the closed source broadcom ELG GLESv2 libs on raspian stretch.

Only replaces the names if platform is rpi and linux dist is debian 9.1 or newer.

Proposed fix inspired from this issue report: https://github.com/kivy/kivy/issues/5341

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -598,7 +598,12 @@ def determine_gl_flags():
             '/opt/vc/include/interface/vcos/pthreads',
             '/opt/vc/include/interface/vmcs_host/linux']
         flags['library_dirs'] = ['/opt/vc/lib']
-        flags['libraries'] = ['bcm_host', 'EGL', 'GLESv2']
+        from platform import linux_distribution
+        dist = linux_distribution()
+        if if dist[0] == 'debian' and float(dist[1]) >= 9.1:
+            flags['libraries'] = ['bcm_host', 'brcmEGL', 'brcmGLESv2']
+        else:
+            flags['libraries'] = ['bcm_host', 'EGL', 'GLESv2']
     elif platform == 'mali':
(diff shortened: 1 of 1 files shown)


dbgasaway:
Is release 9.1 the correct release to compare to? These libraries are actually available under the new names in 8.0, at least on my pi, and it's possible someone with 9.0 could have these problems as well (if that's even possible to have as a version on a pi).

Also, the current code (in my understanding) checks if it's an rpi by checking for the existence of the file /opt/vc/include/bcm_host.h on line 129 at https://github.com/rasmuspeders1/kivy/blob/6043834d4532b67cab6e01783883467f0db96b99/setup.py#L129. Maybe the best solution for detecting the new library names, using a similar technique, would be to check for the libbrcmGLESv2 and libbrcmEGL files in /opt/vc/lib/? Especially considering that it's already manually adding the /opt/vc/lib/ include directory, and it avoids checking a high-level feature like a Debian release version.

rasmuspeders1:
I agree @dbgasaway. The idea withe simply checking the distribution and version was to affect only this and newer distros.
I like your idea with checking if libbrcmEGL and libbrcmGLESv2 exists better. This could potentially solve the issue on different and older distros as well, provided they follow the same naming convention as Raspbian. And it should be at least as safe as my original proposed fix.

I will push a change with this different solution to the pull request.

rasmuspeders1:
The pull request is updated. It has been tested successfully on a rpi with rasbian stretch.

tito:
Looks good, thanks you!

PR #5363, opened 2017-09-01
[WIP] Various improvement for speeding up Widget instanciation

tito:
The changes:
- cythonized ObservableDict and ObservableList (+22% faster for Widget
instanciation, +8.74% for Label, +2.80% for Button)
- created a kivy.lang._speed and moved some part into it. Right now,
only create_handler got optimized as much as possible. There is room for
property/fbind optimization if somehow we can pass C function to it.
update_intermediates is not cythonized. (22% (no changes) faster for
Widget, 14% for Label, 12% for Button)

The test script is available here: https://gist.github.com/tito/9714e53fd76fe08153f11d5aef1b80e0
My speed test is how much widget i can create within 2 seconds.

Here is the number:
- Original version (v1.10.1.dev0 at 1st september 2017): 27554 Widgets, 6278 Labels, 3290 Buttons
- After both modifications: 33328 Widgets, 6909 Labels, 3539 Buttons

(full document for reference: https://docs.google.com/spreadsheets/d/1qliYdec3SPJd8qNJI8GDZtjcCR_SVEMkGgr6Fm_oeFU/edit?usp=sharing)

### DIFF
--- a/kivy/_event.pxd
+++ b/kivy/_event.pxd
@@ -8,13 +8,13 @@ cdef class Observable(ObjectWithUid):
     cdef object __fbind_mapping
     cdef object bound_uid
 
-
 cdef class EventDispatcher(ObjectWithUid):
     cdef dict __event_stack
     cdef dict __properties
     cdef dict __storage
     cdef object __weakref__
     cdef public set _kwargs_applied_init
+    cdef object _proxy_ref
     cpdef dict properties(self)
 
 
--- a/kivy/lang/__init__.py
+++ b/kivy/lang/__init__.py
@@ -857,9 +857,8 @@ will first be unloaded and then reloaded again. For example:
 '''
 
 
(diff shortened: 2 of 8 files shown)


KeyWeeUsr:
The numbers look amazing, but it killed quite a lot tests that use `do_layout()` though, see the [build](https://travis-ci.org/kivy/kivy/jobs/270913691). I'm not sure if you moved the `ObservableReferenceList.prop` somewhere, but that's the error apparently :/

It's rather mystery that it doesn't affect other tests (and I've used layouts in some), therefore perhaps it affects only the `StackLayout`?

tito:
I was waiting the tests to see if the PR was ok or not. I'll look at it later, maybe the PR is not a good thing to do :)

tito:
This is not mergable anymore, but the same work can be redone if necessary.

PR #5368, opened 2017-09-05
Unittesting features

KeyWeeUsr:
This PR disables creating of the screenshots unless explicitly asked for it via env var, which lets our `common.py` module to be *the* way for unittesting without workarounds such as creating a folder for the screenshots, cleaning it later and making the tests to behave "correctly", so that the `GraphicUnitTest.render()` doesn't raise an error if you render a single frame at the end after `EventLoop.idle()` the second time you run the test (visible on plain `python test_uix_slider.py`).

I also added a method for moving the frames in more visible way instead of forcing the user to go and dig through the internals. And then there's a simple `MotionEvent` that represents a single touch with methods highly similar to those we use in the `Widget` (obvious to use just by the name). In the future (dunno when) I'd like to add more events, so that we can write tests related to desktop platforms only for example, or so that we can write a tests for providers (if we can expect the data retrieved from the provider to be sane, we can then just test our classes with some constant input).

### DIFF
--- a/kivy/tests/common.py
+++ b/kivy/tests/common.py
@@ -283,3 +283,18 @@ class GraphicUnitTest(_base):
         root.mainloop()
 
         return self.retval
+
+    def advance_frames(self, count):
+        '''Render the new frames and:
+
+        * tick the Clock
+
+        * dispatch input from all registered providers
+
+        * flush all the canvas operations
+
+        * redraw Window canvas if necessary
+        '''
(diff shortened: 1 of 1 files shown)


tito:
IMO, on this part, you could commit without any PR. OK to merge

KeyWeeUsr:
Cool. The simple guide how to make tests will be available soon-ish :)

PR #5370, opened 2017-09-06
Fix all ScreenManagers sharing the same transition

Cheaterman:
All ScreenManagers share the same SlideTransition instance by default, which can cause some timing issues and other weirdness (and is more generally a bad idea). This PR fixes the issue.

### DIFF
--- a/kivy/uix/screenmanager.py
+++ b/kivy/uix/screenmanager.py
@@ -906,7 +906,7 @@ class ScreenManager(FloatLayout):
     to None.
     '''
 
-    transition = ObjectProperty(SlideTransition(), baseclass=TransitionBase)
+    transition = ObjectProperty(baseclass=TransitionBase)
     '''Transition object to use for animating the transition from the current
     screen to the next one being shown.
 
@@ -966,6 +966,7 @@ class ScreenManager(FloatLayout):
     def __init__(self, **kwargs):
         super(ScreenManager, self).__init__(**kwargs)
         self.fbind('pos', self._update_pos)
+        self.transition = SlideTransition()
 
     def _screen_name_changed(self, screen, name):
         self.property('screen_names').dispatch(self)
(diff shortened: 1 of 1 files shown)


kivy/uix/screenmanager.py
```diff
@@ -964,6 +964,7 @@ def _get_screen_names(self):
     '''
 
     def __init__(self, **kwargs):
+        self.transition = SlideTransition()
```
pythonic64: I would suggest using: 
```
if 'transition' not in kwargs:
    kwargs['transition'] = SlideTransition()
```
This way if `transition` keyword is used in creating of `ScreenManager` instance then default `SlideTransition` instance doesn't get created.

Cheaterman: Agreed my good friend, thanks for the idea! :-)

kivy/uix/screenmanager.py
```diff
@@ -964,6 +964,8 @@ def _get_screen_names(self):
     '''
 
     def __init__(self, **kwargs):
+        if 'transition' not in kwargs:
+            self.transition = SlideTransition()
```
KeyWeeUsr: I'd perhaps choose to rewrite only the `kwargs` part and not put it directly into the instance because the `transition` is a Kivy property and those are evaluated from `kwargs` in the `super()` call to any `EventDispatcher` instance afaik. However when tested, the assignment triggers `on_transition` correctly, sooo.. still looks good to me :)

Cheaterman: Yes, that's what he suggested, but I felt like we didn't need that much indirection to simply set a variable, and I discussed with tshirtman in meatspace and we reached the same conclusion, so there you have it :-) if you still feel setting kwargs is a better idea, I can amend my PR of course!

KeyWeeUsr: Well, it doesn't break the `on_*` event, so it should be ok.

KeyWeeUsr:
Thanks 🍡

PR #5372, opened 2017-09-07
test_video.py: Fix misleading class name

bionoid:


KeyWeeUsr:
Copied from `test_animation.py` apparently. Thanks 😸

Issue #5369, opened 2017-09-06
inspector not working in kivy-1.10.0 Popup


rjungbeck:
Opening inspector (Ctrl+E) on a Popup window shows the fields from the underlying window (and not from the Popup). 

This makes using inspector on Popups (which could be the bigger part of an application) impossible.

bionoid:
I cannot reproduce this on Windows, can you be more specific about the context in which this fails? I tested with this code:

```python
from kivy.base import runTouchApp
from kivy.lang import Builder

runTouchApp(Builder.load_string('''
#:import F kivy.factory.Factory

<TestPopup@Popup>:
	size_hint: .5, .5
	Button:
		on_press: root.dismiss()

Button:
	on_press: F.TestPopup().open()
'''))
```

rjungbeck:
The context is a bit more complicated:

`````
from kivy.base import runTouchApp
from kivy.lang import Builder

runTouchApp(Builder.load_string('''
#:import F kivy.factory.Factory


<TestPopup2@Popup>:
    title: "test2"
    size: 300, 300
    size_hint: None, None
    Button:
        size: 100, 100
        size_hint: None, None
        on_press: root.dismiss()
        
<TestPopup@Popup>:
    title: "test"
    size: 300, 300
    size_hint: None, None
    Button:
        size: 200, 200
        size_hint: None, None
        on_press: F.TestPopup2().open()

BoxLayout:
    size: 800,480
    size_hint: None, None

    Label:
        size: 200, 500
        size_hint: None, None
        text: "Root"
        
    Button:
        size: 200, 500
        size_hint: None, None
        on_release: F.TestPopup().open()
    
    
    '''))

`````
Try to activate the inspector on the button on the second popup (where the button overlap with the underlaying popups button)  It activates the wrong field..

(Test on Mac)

rjungbeck:
From looking at the source:
* Each additional Popup (which is a ModalView) is added as the first child widget to the parent window (in uix/modalview.py 180)
* But the search in modules/inspector.py 391 reverses this order before searching (with pick) and takes the first match.

So although the 2nd Popup is displayed on top, the widget on the underlaying popup is found.

So one of these seems to be wrong .

But which one?

KeyWeeUsr:
It's definitely a problem in the Inspector module as the `ModalView` / `Popup` should by default add itself as a child to the `Window` on top of everything (i.e. as the *first* child).

KeyWeeUsr:
Fixed in master branch.

PR #5378, opened 2017-09-11
Add test for Inspector module, fix children order for ModalView

KeyWeeUsr:
ref #5369 

The PR still needs the last part of unittest i.e. testing multiple `ModalView`s or `Popup`s opened and check if they stack properly as LIFO.

### DIFF
--- a/kivy/modules/inspector.py
+++ b/kivy/modules/inspector.py
@@ -388,7 +388,7 @@ class Inspector(FloatLayout):
         # modalviews before others
         win_children = self.win.children
         children = chain(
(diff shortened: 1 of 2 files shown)


KeyWeeUsr:
meh, the tests will crash because of Travis 😢 

> The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
>
> The job has been terminated

KeyWeeUsr:
Tests pass, the additional change in `inspector.py` closes #5369.

PR #5387, opened 2017-09-18
_text_sdl2.pyx: Don't clear pixel memory twice

bionoid:
[SDL2 does this already](https://github.com/SDL-mirror/SDL/blob/release-2.0.0/src/video/SDL_surface.c#L97-L98)


KeyWeeUsr:
Thanks! :)

Issue #5388, opened 2017-09-21
SDL_Dropfile event is drop as it s send to the app while the app is still in _loop_pause mode


khertan:
### Versions

* Python: 2.7.1
* OS: iOS
* Kivy: 1.10.0
* Kivy installation method: buildozer ios debug

### Description

When registering an iOS app for a URI scheme, when opening that URI Scheme in iOS 11 (with an QRCode for example), iOS call the openURL method which is well implemented in SDL2 (>=2.0) and generate an SDL_DROPFILE event. But as the app is in background while scanning the qrcode, kivy drop the event

### Code and Logs

in kivy/core/window/window_sdl2.py

    def _mainloop(self):                                                                           
        EventLoop.idle()                                                                           
                                                                                                   
        # for android/iOS, we don't want to have any event nor executing our                       
        # main loop while the pause is going on. This loop wait any event (not                     
        # handled by the event filter), and remove them from the queue.                            
        # Nothing happen during the pause on iOS, except gyroscope value sent                      
        # over joystick. So it's safe.                                                             
        while self._pause_loop:                                                                    
            self._win.wait_event()                                                                 
            if not self._pause_loop:                                                               
                break                                                                              
            self._win.poll()

PR #5389, opened 2017-09-21
Don t drop SDL_Dropfile event while in pause #5388

khertan:
We need to catch the dropfile event sent by SDL even in Pause for inter app communication with scheme URI (qrcode from camera from example on iOS) #5388

### DIFF
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -465,7 +465,15 @@ class WindowSDL(WindowBase):
             self._win.wait_event()
             if not self._pause_loop:
                 break
-            self._win.poll()
+            event = self._win.poll()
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
I can't really test it because I don't have Apple device, but if someone can confirm it works (@khertan, an example to reproduce would be really really nice) it should be ok to merge. Seems like a good change and perhaps there's a use for it even on Android, @inclement?

khertan:
Basically I follow https://github.com/kivy/kivy/wiki/Deep-Linking-with-iOS-and-Android.

But I should provide a sample :)

Benoît HERVIER - http://khertan.net

KeyWeeUsr:
Thanks! 🍡

PR #5390, opened 2017-09-21
Fix exception declaration in base class, as subclasses in svg.pxd require it.

scoder:
This will become an error in Cython 0.27, as it would depend on the type that Cython thinks it is calling (base or subtype) whether exceptions will be handled by the caller or not.

### DIFF
--- a/kivy/graphics/instructions.pxd
+++ b/kivy/graphics/instructions.pxd
@@ -94,7 +94,7 @@ cdef class Canvas(CanvasBase):
     cdef float _opacity
     cdef CanvasBase _before
     cdef CanvasBase _after
-    cdef void reload(self)
+    cdef void reload(self) except *
     cpdef clear(self)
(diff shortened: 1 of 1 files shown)


scoder:
Looks like further changes in other .pxd files are also needed (`fbo.pxd`, maybe others). Basically, if you want to raise exceptions from a method in a subtype, then the base type (and all its subtypes) must allow that.

KeyWeeUsr:
The current change throws [an error](https://travis-ci.org/kivy/kivy/jobs/278124193#L1222) during compilation unfortunately(as you've mentioned), so I can't merge it in its current state, but afaik, this change is harmless (ref #5220) except that it required switching to higher version of Cython.

If you're willing, please make similar changes to the files Travis build complains about, so that it can be merged. Thanks :)

scoder:
Closing in favour of #5406.

PR #5392, opened 2017-09-21
Fix setup.py under python2

inclement:
The setup.py currently has an `except FileNotFoundError`, but if an exception does occur then this crashes because FileNotFoundError was introduced only in Python 3. Using OSError instead seems to be the right way to catch the error in both versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,11 @@ def get_version(filename='kivy/version.py'):
         GIT_REVISION = check_output(
             ['git', 'rev-parse', 'HEAD']
         ).strip().decode('ascii')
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
Unfortunately Windows is kind of special and on `2.7.11` returns an `IOError` instead which escapes the `FileNotFoundError`, `OSError`. My bad, I should have doublechecked it before.

```python-traceback
  File "<stdin>", line 1, in <module>
IOError: [Errno 2] No such file or directory: '/wop/wopwop/wop_.txt'
```

PR #5396, opened 2017-09-23
Actually display multitouch emulation if sim set to True.

matham:
It didn't display it when the sim profile was set to True, even though it should have. Instead it never displayed the sim touch.

### DIFF
--- a/kivy/input/providers/mouse.py
+++ b/kivy/input/providers/mouse.py
@@ -267,12 +267,17 @@ class MouseMotionEventProvider(MotionEventProvider):
             self.current_drag = None
 
         cur = self.current_drag
-        if (cur and (self.disable_multitouch or 'multitouch_sim' not in
-                     cur.profile or not cur.multitouch_sim)) or\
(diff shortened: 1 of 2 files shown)


KeyWeeUsr:
@matham do you have any simple example/instructions to test it with?

matham:
Yeah, set multitouch on demand in config and then right click in a normal widget vs scatter. Previously it would never show the red dot (but would still emulate it so it'd be invisible), now it does it for scatter only.

KeyWeeUsr:
Reproduced, feel free to merge. I'm trying to create a simple unittest for that. :)

KeyWeeUsr:
I wonder why is it a different behavior for `on_demand` simulator and the default one i.e. why the default one works correctly right after pressing and the `on_demand` only after releasing? Shouldn't they match?

matham:
The default is verrrrry greedy and always puts the dot. With on_demand, at the time of press we don't know whether it'll be a mulittouch sim touch, so we wait until release.

PR #5399, opened 2017-09-24
Add unittest for Mouse multitouch simulator

KeyWeeUsr:
Tests the "red dot" existence, position, movement and disappearing. Takes care of testing the fix in #5396 too. Cases are for the default (always on) multitouch, `multitouch_on_demand` and `disable_multitouch`.


matham:
The name fix is good. This should be noted though in release notes as breaking change though.

matham:
Ahh, never mind. The typo was only internal - I thought it was also miss-spelled in the config.

PR #5402, opened 2017-09-25
Fix: there

kwadrat:


KeyWeeUsr:
Thanks, but please make sure you add more changes and a proper commit message in the future. :)

kwadrat:
Good idea, thank you! I haven't seen more misspellings and therefore I have sent a single word correction today.

PR #5406, opened 2017-09-25
Fixes for Cython 0.27

KeyWeeUsr:
Based on #5390 and trying to fix remaining issues. While it seems to be ok on py2, py3 for whatever reason tries to say `kivy.graphics.instructions` isn't available ([it is actually ô_ô](https://travis-ci.org/KeyWeeUsr/kivy/jobs/279673529#L1201)). Perhaps [the error](https://travis-ci.org/KeyWeeUsr/kivy/jobs/279673529#L2500) means I'm not on the right track?

### DIFF
--- a/kivy/factory.py
+++ b/kivy/factory.py
@@ -136,7 +136,11 @@ class FactoryBase(object):
         # No class to return, import the module
         if cls is None:
             if item['module']:
-                module = __import__(name=item['module'], fromlist='.')
+                module = __import__(
+                    name=item['module'],
+                    fromlist='*',
+                    level=0  # force absolute
+                )
                 if not hasattr(module, name):
                     raise FactoryException(
                         'No class named <%s> in module <%s>' % (
--- a/kivy/graphics/fbo.pxd
+++ b/kivy/graphics/fbo.pxd
@@ -27,4 +27,4 @@ cdef class Fbo(RenderContext):
     cdef int apply(self) except -1
     cdef void raise_exception(self, str message, int status=?)
     cdef str resolve_status(self, int status)
-    cdef void reload(self)
+    cdef void reload(self) except *
--- a/kivy/graphics/fbo.pyx
+++ b/kivy/graphics/fbo.pyx
@@ -360,7 +360,7 @@ cdef class Fbo(RenderContext):
             self.flag_update_done()
         return 0
 
-    cdef void reload(self):
+    cdef void reload(self) except *:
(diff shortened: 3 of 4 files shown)


scoder:
Note how it says `kivy.graphics.instructions.`, with a dot at the end. Maybe that is the problem?

scoder:
Note that in CPython 3.5 and later, Cython now follows [PEP 489](https://www.python.org/dev/peps/pep-0489/) to improve the behaviour of module code. It is now very similar to Python modules, which also impacts relative imports, for example. Might be related.

scoder:
FYI, 0.27.1 will disable PEP-489 support again. It has proven to cause too much instability on user side.

KeyWeeUsr:
@scoder thanks for notifying! However, I guess you'd like to get it going asap? If that's the case, it's quite an issue for us. :)

scoder:
> I guess you'd like to get it going asap?

I guess you mean that we'd re-enable PEP-489? At some point, sure, but probably as an opt-in, or at least with safer defaults. Module reloading, for example, might not be allowed without user interaction.

You can try the latest Cython master branch, it's essentially 0.27.1 already.

scoder:
Note that I would still recommend using proper relative imports (`from . import xyz`) in your modules, and preferably also `from __future__ import absolute_import`. It's cleaner and also faster, because it avoids relative directory searches for globally installed modules.

KeyWeeUsr:
@scoder I agree we need to fix it, I was just confirming the tests pass again i.e. that it's the PEP489 we had an issue with (I somewhat doubt it).

@matham I changed `fromlist` to `fromlist='*'` which apparently fixed our issue with the new stable Cython version. We tend to limit exported symbols with `__all__`, however it's ignored all the same, therefore it shouldn't do *much* harm. I'm rather curious what exactly did it change for `Factory` because I'm apparently missing the result as the imports work the same:

    Layout = __import__(name='kivy.uix.boxlayout', fromlist='*').Layout
    Layout = __import__(name='kivy.uix.boxlayout', fromlist='.').Layout
    import kivy.uix.boxlayout as b; Layout = b.Layout

although `Layout` isn't present in the `__all__`.

KeyWeeUsr:
Anyone anything against? [Tested with buildozer](https://github.com/kivy/kivy/issues/5432#issuecomment-336650458) and seems to work fine.

inclement:
\o/

PR #5407, opened 2017-09-26
Downgrade Cython to 0.26.1 for builds

KeyWeeUsr:
Until #5390/#5406 fixes the stuff for the newest stable version of Cython we need to use the previous version (it won't compile) otherwise we can't really run the tests.


dessant:
It's ok, just make sure to switch it back. It always uses the latest Cython to act as a canary.

PR #5415, opened 2017-10-05
[RFC] KivyWM

jakogut:
This pull request is a work-in-progress demonstration of Kivy adapted to function as a compositing X11 window manager. The intent is to create a channel of communication with the Kivy development team and users, to evaluate the utility of the feature, as well as shape design choices going forward.

Below is a screenshot of Kivy running in Xephyr, with a GridLayout for the root widget. Added to the GridLayout is a Label widget, and an XWindow widget displaying glxgears.

screenshot from 2017-10-04 16-09-21

Here is a video demonstration: https://www.youtube.com/video_id=tkuvaVhVNus

Currently, the only supported window provider is X11, so Kivy must be built with `USE_X11=1`, and the example program must be run with `KIVY_WINDOW=x11`.

There's been some initial work on making the SDL2 window provider work with KivyWM, but it's unfinished.

Areas that need improvement:

- Input handling
- Scaling windows based on widget size
- Extension loading (TFP)
- Xfixes/XDamage for updating textures
- Documentation
- Compatibility with window providers other than X11 (SDL2)
- Compatibility with display servers other than X
- Window/Display server agnostic method of binding a texture to a pixmap
    (May be worth considering integrating a UI toolkit that supports this natively, such as COGL, or Clutter)

### DIFF
--- a/kivy/graphics/cgl.pxd
+++ b/kivy/graphics/cgl.pxd
@@ -22,10 +22,33 @@ cdef extern from "gl_redirect.h":
     ctypedef signed long int    GLintptr
     ctypedef signed long int    GLsizeiptr
 
+    # X11
     cdef struct _XDisplay:
         pass
 
     ctypedef _XDisplay Display
+
+    cdef struct XVisualInfo:
+        pass
+
+    ctypedef int XID
+    ctypedef XID Bool
+    ctypedef XID Window
+    ctypedef XID Drawable
+    ctypedef XID Pixmap
+
+    # GLX
+    cdef struct __GLXFBConfigRec:
+        pass
+
+    ctypedef __GLXFBConfigRec GLXFBConfig
+
+    cdef struct __GLXcontextRec:
+        pass
+
+    ctypedef __GLXcontextRec *GLXContext
+    ctypedef XID GLXPixmap
+    ctypedef XID GLXDrawable
 
     int GL_DEPTH_BUFFER_BIT
     int GL_STENCIL_BUFFER_BIT
--- a/kivy/graphics/cgl.pxd
+++ b/kivy/graphics/cgl.pxd
@@ -22,6 +22,10 @@ cdef extern from "gl_redirect.h":
     ctypedef signed long int    GLintptr
     ctypedef signed long int    GLsizeiptr
 
+    cdef struct _XDisplay:
+        pass
+
+    ctypedef _XDisplay Display
 
     int GL_DEPTH_BUFFER_BIT
     int GL_STENCIL_BUFFER_BIT
@@ -378,6 +382,7 @@ cdef extern from "gl_redirect.h":
 
     int GL_FRAMEBUFFER_UNDEFINED_OES
 
+# GL
 ctypedef const GLubyte* (__stdcall *GLGETSTRINGPTR)(GLenum) nogil
 ctypedef GLboolean (__stdcall *GLISBUFFERPTR)(GLuint buffer) nogil
 ctypedef GLboolean (__stdcall *GLISENABLEDPTR)(GLenum cap) nogil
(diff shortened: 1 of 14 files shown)


kivy/graphics/cgl.pxd
```diff
@@ -639,6 +675,8 @@ ctypedef struct GLES2_Context:
     void (__stdcall *glVertexAttribPointer)(GLuint, GLint, GLenum, GLboolean, GLsizei, const void *) nogil
     void (__stdcall *glViewport)(GLint, GLint, GLsizei, GLsizei) nogil
 
+    GLX_Context *glx
```
KeyWeeUsr: I'm not really sure, but if we use GLES2 on mobile devices, is this a good approach? I mean, does it compile on phones if this change is present?

kivy/graphics/cgl.pyx
```diff
@@ -71,10 +77,16 @@ cdef void cgl_set_context(GLES2_Context* ctx):
...
 
+cdef void glx_init() except *:
+    cgl.glx.glXBindTexImageEXT = <PFNGLXBINDTEXIMAGEEXTPROC>glXGetProcAddress("glXBindTexImageEXT")
+    cgl.glx.glXReleaseTexImageEXT = <PFNGLXRELEASETEXIMAGEEXTPROC>glXGetProcAddress("glXReleaseTexImageEXT")
 
 cdef void cgl_init() except *:
     Logger.info('GL: Using the "{}" graphics system'.format(
         'OpenGL ES 2' if kivy_opengl_es2 else 'OpenGL'))
+
+    glx_init()
```
KeyWeeUsr: I think this will cause troubles on non-X11 platforms.

kivy/lib/sdl2.pxi
```diff
@@ -4,15 +4,15 @@
 #whose terms are available in the LICENSE file or at http://www.ignifuga.org/license
 
 
-cdef extern from "SDL_joystick.h":
+cdef extern from "SDL2/SDL_joystick.h":
```
KeyWeeUsr: Without fixing `setup.py` to find it under this path everywhere by default I don't think this will be ok, nor that it's a good way as it's an API-break. Or rather a dependency-break for anyone using Kivy for longer period and compiling manually.

kivy/graphics/cgl.pxd
jakogut: You are correct, this breaks the build on non-Linux/x11 systems, and it's something that needs changed. I haven't figured out a good solution yet, because the window provider, graphics API and display server all change the implementation of TFP. Some platforms (Android/iOS) likely can't support it in any form.

jakogut:
> Still, I think adding a new core widget that evidently supports only X11 (therefore no Wayland, no OSX, no Win, etc) is a bad idea.

I agree that the API shouldn't be X specific (like it is now), but as far as a backend, X11 is one of the most portable display servers in existence. Any macOS support would surely require replacing the default shell with an X server, and I question how many people would want to do that, rather than run Linux. Windows may be possible, I'm aware of a Blackbox port that replaces the Windows shell, but I'm not sure of the specifics.

I think the three backends that matter are X (GLX or EGL), Wayland (EGL), and Windows (WGL).

> About tfp.pyx, that's rather a too narrow approach to the idea. You probably can get the data in/out as some non-Pixmap container. Perhaps XBitMap.

Xcomposite redirects drawing to offscreen pixmaps, at which point the reason to use TFP or equivalent is for zero-copy GL texture creation from a window's contents. I think I'll have to do some research on Wayland to better understand how to do this portably.

KeyWeeUsr:
> I agree that the API shouldn't be X specific (like it is now), but as far as a backend, X11 is one of the most portable display servers in existence.

The point is about making a new kivy *core widget*, not about the work in the first place. I haven't said I'm against it (the whole work), I said I'm against including it as a *widget* to the core, because afaik everything that's in the `uix` folder works anywhere Kivy itself works no matter what backend is chosen (well, perhaps except mock).

If you ask *where* to put it, well, that's not clear to me yet, because it's X11 specific thing, however it'd be a nice thing to have. Maybe a good approach would be including it as a Window feature like it's for moving the window or other SDL2 specific parts. If no good place is found, then you could add it to the [Kivy Garden](http://kivy-garden.github.io).

What matters the most here is not breaking the already existing graphics core with putting raw platform specific symbols to random places(with no safety switch). You said to enable this feature Kivy needs to be built with `USE_X11=1`, therefore please look at @tito's [SO answer](https://stackoverflow.com/a/5162028/5994041) related to this "issue" (basically using DEF(ine) macros). I haven't touched it for a while, but the basic example is here for [python side](https://github.com/kivy/kivy/blob/f5161a2485db8c15520b8370211fd579cac5c3bc/kivy/core/window/__init__.py#L28) and for [cython side](https://github.com/kivy/kivy/blob/082c4544415ab6bd976d9dda9c342bae109e7cbb/kivy/graphics/cgl_backend/cgl_sdl2.pyx#L10)(don't forget [`config.pxi`](https://github.com/kivy/kivy/blob/082c4544415ab6bd976d9dda9c342bae109e7cbb/kivy/graphics/cgl_backend/cgl_sdl2.pyx#L6))

> Xcomposite redirects drawing to offscreen pixmaps, at which point the reason to use TFP or equivalent is for zero-copy GL texture creation from a window's contents.

I'm not familiar with X11 related stuff too much, however I'm not sure if using X11 specific containers is the best approach if we don't use it anywhere else in Kivy. Perhaps something showing the differences between XPixMap and our stuff (maybe pixmap → raw → Kivy texture conversion) would be useful when deciding whether to use this or that.

Zen-CODE:
For my 2 cents, KeeWeeUsr is in on point here. It's a nice addition, but does not belong in the core, which needs to stay platform-agnostic and minimal. Put in in the Kivy-Garden, so it's easy for people to use. If enough people use it or want it as a core widget, then it can be added later...

PR #5421, opened 2017-10-08
Fix host/port handling in UrlRequest

### LINKED ISSUE #2575: [UrlRequest]  ':' inside Url will cause invalid literal error when it isn't marking the port number
If you have a Url like this:

url =" https://myAppID:javascript-key=myJavaScriptKey@api.parse.com/1/classes/Door"

Urlrequest will cast following error:

invalid literal for int() with base 10: "javascript-key=myJavaScriptKey@api.parse.com"

This makes it so a lot of REST APIs can't be called from kivy not just from the parse.com site

This is caused because Urlrequest will see the ':' and will expect a port number. 

See lines 242-247 in Urlrequest:

 # correctly determine host/port
        port = None
        host = parse.netloc.split(':')
        if len(host) > 1:
            port = int(host[1])
        host = host[0]
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5259483-urlrequest-inside-url-will-cause-invalid-literal-error-when-it-isn-t-marking-the-port-number?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

KeyWeeUsr:
Closes #2575 

Note that this does NOT handle the authentication, just appends `user:pass` part to the `host` if applicable, therefore the host + port part is handled correctly, however, when the credentials are specified, it will most likely fail (no test url to check unfortunately) if it's not handled via `HTTPConnection` or `socket` (probably not).

### DIFF
--- a/kivy/network/urlrequest.py
+++ b/kivy/network/urlrequest.py
@@ -245,6 +245,31 @@ class UrlRequest(Thread):
         if self in g_requests:
             g_requests.remove(self)
 
+    def _parse_url(self, url):
+        parse = urlparse(url)
+        port = parse.port
+
+        # append user + pass to hostname if specified
+        if parse.username and parse.password:
+            host = '@'.join([
+                ':'.join([
+                    parse.username or '',
+                    parse.password or ''
+                ]),
+                parse.hostname
+            ])
+
+        elif parse.username and not parse.password:
+            host = '@'.join([
+                parse.username,
+                parse.hostname
+            ])
+
+        elif not parse.username and not parse.password:
+            host = parse.hostname
+
+        return host, port, parse
+
     def _fetch_url(self, url, body, headers, q):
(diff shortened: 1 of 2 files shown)


tshirtman:
http://httpbin.org/ can be used to check http protocl, it has enpoints for authentication

KeyWeeUsr:
Thanks for the link, I could implement basic auth with that. I removed those manual checks for URL parsing as we use the defaults from `urllib` now (CPython has unittests for that) + added cases for auth through header and via this syntax:

    http://user:pass@domain.xyz

PR #5439, opened 2017-10-15
OSX Install Instruction Update - Cython explicit version

ParokshaX:
- kivy does not compile with the recent Cython 0.27.x in OSX, there is compilation failure
- Added instruction to install the latest supported version which is 0.26.x

### DIFF
--- a/doc/sources/installation/installation-osx.rst
+++ b/doc/sources/installation/installation-osx.rst
@@ -19,7 +19,7 @@ You can install Kivy with Homebrew and pip using the following steps:
 
     2. Install Cython and Kivy using pip::
 
-        $ pip install -U Cython
+        $ pip install Cython==0.26
         $ pip install kivy
 
     - To install the development version, use this in the second step::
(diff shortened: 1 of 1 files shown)


dessant:
@KeyWeeUsr, 0.26.1 is the newest version which works with kivy stable?

ParokshaX:
@dessant I also see a related commit that explicitly specifies the Cython version [5c4b8ed14e272c182b2d](https://github.com/kivy/kivy/commit/5c4b8ed14e272c182b2d097a07973ead462213ae)

dessant:
@ParokshaX, yes, your change is good, I was just wondering if Cython devs are planning to fix the regression, or if it was intentional.

KeyWeeUsr:
@dessant The latest one I tried with the stable is mentioned [here](https://kivy.org/docs/installation/installation-linux.html#common-dependencies), however I guess `0.26.1` is the latest good one for stable. The [commit](https://github.com/kivy/kivy/pull/5220) that actually introduced incompatibility was introduced *after* the release and then the Cython `0.27` was released which raises compilation errors about the `reload()` declarations.

@ParokshaX The related commit is there only because #5406 isn't merged yet. I don't have OSX/iOS, however it was confirmed to work on Windows+Linux+Android, so it should be safe to merge. However, if you can, please test the PR, so that I can merge without worrying too much about breaking OSX 😄

PR #5445, opened 2017-10-20
Resize treeview collapse. closes #5426

### LINKED ISSUE #5426: TreeView: collapse button really small on android
The problem (with screenshot) is described here:
https://stackoverflow.com/questions/41177706/kivy-treeview-collapse-button-really-small-on-android

"I'm using TreeView to build an app, and it renders fine when testing on my laptop but the little buttons to collapse/uncollapse tree nodes are tiny on my phone (see the tiny arrow next to Group 1 in the image). I've looked for an attribute or property in TreeView that I could set, but nothing seems to help."

The reason for that is a bug in style.kv:

```
 232         Rectangle:
 233             source: 'atlas://data/images/defaulttheme/tree_%s' % ('opened' if self.is_open else 'closed')
 234             size: 16, 16
 235             pos: self.x - 20, self.center_y - 8
```
Change this code to:
```
 232         Rectangle:
 233             source: 'atlas://data/images/defaulttheme/tree_%s' % ('opened' if self.is_open else 'closed')
 234             size: sp(16), sp(16)
                     # or better:    size: self.height / 2, self.height / 2  
 235             pos: self.x - sp(20), self.center_y - sp(8)
```
solves the problem

Zen-CODE:
Sizing the collapse button dynamically does make more sense.

https://github.com/kivy/kivy/issues/5426

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -231,7 +231,7 @@
             rgba: 1, 1, 1, int(not self.is_leaf)
         Rectangle:
             source: 'atlas://data/images/defaulttheme/tree_%s' % ('opened' if self.is_open else 'closed')
-            size: 16, 16
+            size: self.height / 2.0, self.height / 2.0
             pos: self.x - 20, self.center_y - 8
     canvas.after:
         Color:


kivy/data/style.kv
```diff
@@ -231,7 +231,7 @@
             rgba: 1, 1, 1, int(not self.is_leaf)
         Rectangle:
             source: 'atlas://data/images/defaulttheme/tree_%s' % ('opened' if self.is_open else 'closed')
-            size: 16, 16
+            size: self.height / 2.0, self.height / 2.0
             pos: self.x - 20, self.center_y - 8
```
KeyWeeUsr: I guess changing `size` only isn't enough if you introduce `self.height / 2.0`.

Zen-CODE: Should be? Size is a ReferenceListProperty for width and height, so changing size would also change these two properties. Do you have reason to believe otherwise?

KeyWeeUsr: My edit didn't pass O_o ...anyway, I'm talking about `pos` because currently there's just fixed amount subtracted :)

KeyWeeUsr:
Thanks :)

PR #5451, opened 2017-10-23
Revert "Downgrade Cython to 0.26.1 for builds"

KeyWeeUsr:
Reverts kivy/kivy#5407 because #5406 was merged.

Perhaps reverting/switching to latest Cython might be a good idea for https://github.com/kivy/kivy/commit/6b9c10c41957809a220444da0f08830d0ba3e53f too. If the versions on iOS/OSX match the rest of the builds switching to syntax mentioned in #5302 would be a good idea to keep stuff in one place.


KeyWeeUsr:
Seems to work fine with CIs and tests.

PR #5455, opened 2017-10-25
Add TextInput cursor blinking control

### LINKED ISSUE #5449: the textinput widget's cursor_blink property can't be set
* Python: 3.6
* OS: Windows 7
* Kivy: 1.10

According to the docs the TextInput widget's cursor_blink property is automatically set and can't be changed. (And the docs are right, 'cos I tried setting it to False.)

I have told Windows not to blink the cursor, but Kivy ignores this even though all other Windows applications respect it.

Some people cannot work with blinking cursors, so it ought to be possible to switch it off -- or at least to respect the user preferences.

See http://www.jurta.org/en/prog/noblink

KeyWeeUsr:
Moves the original functionality of `cursor_blink` to `_cursor_blink`. Perhaps if we should preserve the "public" property, we might rename it to something more obvious such as `cursor_invisible` or something similar.

"invisible" just because of the current behavior, otherwise we might invert it and use `cursor_visible` instead.

Closes #5449

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -184,7 +184,10 @@
             size: self.size
             source: self.background_active if self.focus else (self.background_disabled_normal if self.disabled else self.background_normal)
         Color:
-            rgba: (self.cursor_color if self.focus and not self.cursor_blink else (0, 0, 0, 0))
+            rgba:
+                (self.cursor_color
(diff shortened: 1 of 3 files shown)


tito:
Good to me, but you missed to document the change with .. versionchanged::

PR #5466, opened 2017-10-31
Introduce no support for Cython 0.27 - 0.27.2

KeyWeeUsr:
Resolved in https://github.com/cython/cython/issues/1968, however we probably shouldn't support those versions directly as they contain the unpatched `Coroutine.c` which doesn't build Kivy with MinGWPy/MinGW (GCC) for Python 3.4 on Windows.

Trying temporarily with Cython's master branch now and hopefully there will be a release soon.

Also, I think we might move [this chapter](https://kivy.org/docs/installation/installation-linux.html#cython)(or section or...) out from the linux installation page to let's say a separate page e.g.:

    https://kivy.org/docs/installation/cython.html
    https://kivy.org/docs/installation/cython_support.html
    https://kivy.org/docs/installation/deps/cython.html

The last alternative seems even nice and we might go with this URL style even for other deps e.g. SDL2 (mixer issue(s) currently), GStreamer or the recent RPi drivers. I think it'd make it easier for editing for the users such as:

##### cython.html

    * 0.28 yay for Kivy X.Y.Z
    * 0.27.2 nope with 3.4 + mingw + win for Kivy X.Y.Z (#issue)
    * -- || --
    * ...

### DIFF
--- a/doc/sources/conf.py
+++ b/doc/sources/conf.py
@@ -224,10 +224,10 @@ from kivy import setupconfig
 replacements = {
     'cython_install': 'Cython==' + setupconfig.CYTHON_MAX,
     'cython_note': (
-        'This version of **Kivy requires at least Cython version {0}**, '
-        'and has been tested through {1}. Later versions may work, '
+        'Kivy {0} **requires at least Cython version {1}**, '
+        'and has been tested through {2}. Later versions may work, '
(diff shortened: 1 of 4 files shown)


KeyWeeUsr:
New Cython stable (https://github.com/cython/cython/releases/tag/0.27.3), less troubles! ^^

Issue #180, opened 2011-06-27
Rotation window + shader = fail


tito:
If you rotate the window and use a shader, it will fail: the rotation is not applied to the shader.
We should have an uniq way for using the parent matrix for model and projection from shader in graphics.
##

tshirtman:
5 years passed… here is an example to reproduce, at least, and yes, it's not yet fixed.

`python2 examples/shader/plasma.py -c graphics:rotation:90`

vs 

`python2 examples/shader/plasma.py`

![shader_rotation](https://cloud.githubusercontent.com/assets/22759/14055236/73c939b0-f2e0-11e5-9bb6-9b85e310a61a.png)

multiple issues visible:
- The shader doesn't cover the correct part of the window, if you resize, you see it's size is [windows.height, windows.width] instead of the opposite.
- The animation looks exactly the same as the un-rotated version, instead of rotated.

IgorCode:
I'm seeing the behaviour that @tshirtman described without using shader (that I know of). Everytime I pause and return to app I get this, that is if I'm holding the device in different orientation from the one app was meant to be in (i.e. device is in portrait mode, app is in landscape mode).

parabolize:
In a effort to understand kivy a little better I explored this problem a bit. I will try to describe what is happening. If you see any inaccuracies please correct me.
The short description is rotations, translations and other manipulations of the parent RenderingContext arn't applied to any custom vertex or fragment shader you create unless you implement what is normally done in the default.vs and default.fs.

The longer explanation:
I will use the examples/shader/plasma.py as the example. 
kivy.core.window.Window is a giant, magical singleton that is normally created in App.run() after App.build() returns your root Widget. Window collects the graphics:rotation:90 argument and uses it along with other information to create two matrices and attaches them to its RenderContext as 'projection_mat' and 'modelview_mat'. This is done in two methods Window._get_size() and Window.update_viewport(). The RenderContext passes these matrices to its Shader. The Shader passes them to cgl. The default vector shader kivy/data/glsl/default.vs multiplies these against each vector. default.vs passes frag_color and tex_coord0 to default.fs.
[...]

PR #5470, opened 2017-11-03
Added 'frag_modelview_mat' uniform to address #180

parabolize:
A model view matrix for the fragment shader that can be manipulated
independently from the vertex model view matrix.

### DIFF
--- a/examples/shader/shadertree.py
+++ b/examples/shader/shadertree.py
@@ -12,15 +12,13 @@ With this way, you can apply cool effect on your widgets :)
 
 '''
 
-from kivy.app import App
 from kivy.clock import Clock
+from kivy.app import App
+from kivy.uix.floatlayout import FloatLayout
+from kivy.core.window import Window  # side effects needed by Shader
+from kivy.properties import StringProperty, ObjectProperty
 from kivy.graphics import (RenderContext, Fbo, Color, ClearColor, ClearBuffers,
         Rectangle)
-from kivy.properties import StringProperty, ObjectProperty
-from kivy.uix.floatlayout import FloatLayout
-
-# imported for side effects needed by Shader
-import kivy.core.window
 
 import itertools
 
--- a/examples/shader/shadertree.py
+++ b/examples/shader/shadertree.py
@@ -12,27 +12,21 @@ With this way, you can apply cool effect on your widgets :)
 
 '''
 
-from kivy.clock import Clock
 from kivy.app import App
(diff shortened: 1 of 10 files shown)


examples/shader/shadertree.py
```diff
@@ -12,27 +12,21 @@
...
-from kivy.core.window import Window
+from kivy.clock import Clock
+from kivy.graphics import (RenderContext, Fbo, Color, ClearColor, ClearBuffers,
+        Rectangle)
 from kivy.properties import StringProperty, ObjectProperty
-from kivy.graphics import RenderContext, Fbo, Color, Rectangle
+from kivy.uix.floatlayout import FloatLayout
 
+# imported for side effects needed by Shader
+import kivy.core.window
```
KeyWeeUsr: I'm not sure, but why `from kivy.core.window import Window` isn't good (i.e. the old version)?

Also, (as you aren't the only one with such PR) I don't really think using [this feature](https://www.jetbrains.com/help/pycharm/optimizing-imports.html) that alphabetically sorts all imports is actually useful as it introduces unnecessary noise to the diff. If you could, it'd be lovely to get a diff without this noise, please. Just a silly wish ^-^

examples/shader/shadertree.py
```diff
@@ -109,17 +103,18 @@ class ShaderWidget(FloatLayout):
...
-            self.fbo = Fbo(size=Window.size, use_parent_projection=True)
+            self.fbo = Fbo(size=self.size)
+            self.fbo_color = Color(1, 1, 1, 1)
+            self.fbo_rect = Rectangle()
 
-        # Set the fbo background to black.
         with self.fbo:
-            Color(0, 0, 0)
-            Rectangle(size=Window.size)
+            ClearColor(0, 0, 0, 0)
```
KeyWeeUsr: Hehehe, `Color()` + `Rectangle()` was a nice trick when using `Window.size`. This should work fine under the `Widget` + `Fbo` directly I think.

KeyWeeUsr:
Tested with:

    python plasma.py                              # works
    python plasma.py -c graphics:rotation:90      # works
    python rotated.py                             # works
    python rotated.py -c graphics:rotation:90     # works
    python shadertree.py                          # works
    python shadertree.py -c graphics:rotation:90  # works

Closes #180

KeyWeeUsr:
Thanks for the fix of a very old issue, much appreciated! 🍡

Issue #5454, opened 2017-10-25
ShapedWindow doesn't work without button click


NafisAbdullahKhan:
### Versions

* Python: 3.6.3
* OS: Windows 10 64 bit
* Kivy: 1.10.1.dev0
* Kivy installation method: wheel

### Description

I was using [this](https://github.com/kivy/kivy/blob/master/examples/miscellaneous/shapedwindow.py) example and wanted to create a shaped window from the starting without any button click. But apparently it only works when I place `win.shape_mode = 'default'` inside any button event (pressed or released) and perform the click event with mouse. I also tried with `trigger_action` to trigger button event automatically but it hangs the window. Same thing happens if I try with threading.Timer to hit `win.shape_mode = 'default'` again after few seconds.

KeyWeeUsr:
I'm not quite sure what are you talking about. Could you elaborate? Here's a very simple example of how to run an already shaped window.

```python
from kivy.config import Config
Config.set('graphics', 'shaped', 1)

from kivy.app import App
from kivy.uix.button import Button
from kivy.core.window import Window
from kivy.properties import StringProperty


class ShapedWindow(App):
    def build(self):
        self.default_shape = Config.get('kivy', 'window_shape')
        Window.shape_mode = 'colorkey'

        return Button()


if __name__ == '__main__':
    ShapedWindow().run()
```

NafisAbdullahKhan:
Well, looks like with `Window.shape_mode = 'colorkey'` it's working. But what about shape_mode 'default' or 'binalpha'? I was hoping to make it work without excluding any color; with an Alpha image.

KeyWeeUsr:
@NafisAbdullahKhan it actually works, but it doesn't load properly unless your `Window` is the same size as the image you want to use for the shape (otherwise you get **errors** in the console). For example you can get this state after changing the image to `alpha_shape` and clicking on the `default` button. The same applies for `binalpha`. 

The first error you will see is after you press the `alpha_shape` button, that's because the `Window`'s size wasn't `(800, 600)` i.e. the size of `default_shape`. That's why you would probably see some 512x512 `Window` that isn't affected by any shape although from the first look it "should" (no, it really shouldn't).

These are the steps you need to use to actually shape the `Window` (I'm going to put them to docs asap):

* set `Window` size to the size of the shape (with `Config`!)
* set `Window` shape to some picture you want to use
* set the `shape_mode` for the `Window`

[...]

PR #5474, opened 2017-11-04
shape_mode set to default initially when window is created

NafisAbdullahKhan:
shape_mode set to "default" initially when window is created. Otherwise it had to be manually triggered (set). See this issue: #5454


KeyWeeUsr:
I don't see what are you trying to accomplish, the point about that issue was setting the Window size equal to shape image size. The shape mode itself is by default set to `'default'` from SDL2 side.

```python
Config.set('graphics', 'shaped', 1)

from kivy.app import App


class ShapedWindow(App):
    def build(self):
        print(Window.shape_mode)
        # >>> 'default'
        return Button()


if __name__ == '__main__':
    ShapedWindow().run()
```

PR #5479, opened 2017-11-08
 add title_markup for ActionPrevious

brami:
This add the BooleanProperty title_markup to ActionPrevious.

### DIFF
--- a/kivy/uix/actionbar.py
+++ b/kivy/uix/actionbar.py
@@ -211,14 +211,14 @@ class ActionPrevious(BoxLayout, ActionItem):
        defaults to ''.
     '''
 
-    title_markup = BooleanProperty(False)
+    markup = BooleanProperty(False)
     '''If True, the text will be rendered using the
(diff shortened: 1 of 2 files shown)


brami:
Sorry. I don't understand the error in continuous-integration/appveyor/pr ...

How can I help?

kivy/uix/actionbar.py
```diff
@@ -211,6 +211,17 @@ class ActionPrevious(BoxLayout, ActionItem):
        defaults to ''.
     '''
 
+    title_markup = BooleanProperty(False)
```
KeyWeeUsr: Instead of `title_markup` use just `markup` like with the `Label` or `Button`. It'll be more obvious and user-friendly for devs to just try whether `markup` property "just works" when used somewhere, where obviously `Label` (or likes) is used. :)

brami: done

KeyWeeUsr:
Thanks!

Issue #5469, opened 2017-11-02
TextInput selection disappearing on <enter> key


KeyWeeUsr:
### Versions

* Python: 3.5
* OS: Windows 8.1
* Kivy: master
* Kivy installation method: wheel

### Description

1. Write some text e.g. "somelongtext"
2. press `<enter>`
3. write some text again, preferably something short e.g. "text"
4. press `<enter>`
5. `<shift> + <up>` (aka select the text)
6. press 7x `<enter>` and look at the selection disappearing

### Code and Logs

```python
from kivy.app import runTouchApp
from kivy.uix.textinput import TextInput
runTouchApp(TextInput())
```

BoyanH:
What is the desired functionality? I would rather expect the selected text to be replaced with a new line when pressing enter

KeyWeeUsr:
@BoyanH yeah, that should be the behavior. More precisely the contents of the selection should completely disappear and then the new line inserted at the end of the previous selection on the first enter key and then 6 times new line.

Try with multiple simple text editors and you should get the result I'm talking about.

emomicrowave:
@BoyanH and I looked it over and we found the problem. There wasn't any check for what happens if the `enter` key is pressed, but when there is a text selection. I made a pull request with the changes.

KeyWeeUsr:
A test for the fix would be very nice to see. I see some other changes mentioned in the PR, so perhaps adding the test for the whole rewritten selection behavior might be better.

What error did you get? I can run the test case with this simple call:

    python test_uix_textinput.py

if it's about where to place the test, well since it works with the keys, I'd add it under the already existing `TextInputGraphicTest.test_text_validate()` which might even inspire you (or perhaps you will have a better approach). :)

emomicrowave:
That's the error I get when I run the file this way.

```
  File "test_uix_textinput.py", line 48, in test_text_validate
     self.assertTrue(ti.text_validate_unfocus)
 AttributeError: 'TextInput' object has no attribute 'text_validate_unfocus'
```

KeyWeeUsr:
That just means you probably don't have up-to-date version of Kivy, because that attribute is just another Kivy property here:

https://github.com/kivy/kivy/blob/21a68562be135338c6aaef0f3a3f6641f4473af4/kivy/uix/textinput.py#L2488

The test runs fine on Travis, Appveyor and locally with the latest master, therefore I don't think it's some weird bug with the graphic unittest.

emomicrowave:
[...]

PR #5482, opened 2017-11-12
Textinput - _key_down() method now more readable, separated deletion and insertion / action keys handling

BoyanH:
As we fixed issue #5469 with @emomicrowave, we noticed the _key_down() method needs some refactoring (because of repetitive self.delete_selection() calls) so we tried to make a clearer separation between deleting text when selected and other key handling.

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -2247,9 +2247,22 @@ class TextInput(FocusBehavior, Widget):
 
     def _key_down(self, key, repeat=False):
         displayed_str, internal_str, internal_action, scale = key
+
+        # handle deletion
+        if self._selection and internal_action in (None, 'del', 'backspace', 'enter'):
+            self.delete_selection()
+        elif internal_action == 'del':
+            # Move cursor one char to the right. If that was successful,
+            # do a backspace (effectively deleting char right of cursor)
+            cursor = self.cursor
+            self.do_cursor_movement('cursor_right')
+            if cursor != self.cursor:
+                self.do_backspace(mode='del')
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
@BoyanH Could you please rebase the branch against master so that it contains only your commits? Thanks! :)

BoyanH:
@KeyWeeUsr Sorry for the mess I did, but I don't think that would be a good idea. I think I need to open a further issue.

I wanted to write some tests today, but I saw you already did, so I rebased this branch to test everything and wrote some further tests, fixed a small mistake I did.

Thing is, I found out that, on selection, when the shift is released and a cursor in the other direction is pressed, the cursor doesn't jump to the other side of the selection.

I would like to commit the test and solution (hopefully), in 30 minutes or so. Then you could look at it more easily and if you want to, I could create a cleaner branch.

KeyWeeUsr:
The branch contains changes that either are already included in the master branch or aren't yours, therefore it's quite a good idea to rebase the branch (`git rebase master`) :)

Re the new found selection issue, I think it should be in a separate PR together with appropriate tests, so that if somebody wants to go back in the history doesn't need to guess what happened where. This PR was about refactoring the conditions (ref the original post), therefore leave the PR as it was in the original state and further changes open in another pull request(s).

BoyanH:
Solves #5490

BoyanH:
@KeyWeeUsr thanks for the suggestions, I rebased this branch on master now and removed the further tests and fixes that don't belong here

KeyWeeUsr:
Thanks for PR and splitting it to parts! 🍡

PR #5509, opened 2017-12-02
Explain units and datatype of dt in kivy.clock

tonyfinn:
The callback is documented as returning the time since the last callback, but it's not actually documented anywhere what datatype or units this comes in. Update the docs to explain that.

### DIFF
--- a/kivy/clock.py
+++ b/kivy/clock.py
@@ -5,7 +5,8 @@ Clock object
 The :class:`Clock` object allows you to schedule a function call in the
 future; once or repeatedly at specified intervals. You can get the time
 elapsed between the scheduling and the calling of the callback via the
-`dt` argument::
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
> You can get *the time elapsed between the scheduling and the calling of the callback*

Thanks for the PR, however I wonder... how is it not explained?

tonyfinn:
If it's been 1503 milliseconds since the last call, any of the following values could be expected with the existing explanation:

1503 (milliseconds, int)
1 (seconds, int)
2 (seconds, int, rounded)
1.503 (seconds, float)

Can reword to have less redundancy though, will during the week. Just something that tripped me up as I was expecting the first value when it was actually the 4th.

tito:
It is already explain just in the paragraph before. So no. You got the issue yes, but the documentation is still accurate.

PR #5518, opened 2017-12-07
less renaming

DavidCaughell:
not sure why core_platform was written that way

### DIFF
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -323,13 +323,10 @@ from kivy.logger import Logger
 from kivy.event import EventDispatcher
 from kivy.lang import Builder
 from kivy.resources import resource_find
-from kivy.utils import platform as core_platform
+from kivy.utils import platform as platform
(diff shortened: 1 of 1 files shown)


inclement:
I'm not sure either, but the fix is certainly welcome. Thanks!

dessant:
It's most likely to avoid confusion with the stdlib [platform](https://docs.python.org/3.6/library/platform.html). Maybe at a certain point both were used in the module.

kivy/app.py
```diff
@@ -323,15 +323,12 @@ def on_resume(self):
 from kivy.event import EventDispatcher
 from kivy.lang import Builder
 from kivy.resources import resource_find
-from kivy.utils import platform as core_platform
+from kivy.utils import platform as platform
```
rnixx: Won't be ``from kivy.utils import platform`` enough?

DavidCaughell: Thanks.  :)

PR #5526, opened 2017-12-10
refactor unicode_textinput.py example

Fogapod:
Fix: loading font with `.` in name will not cause exception
Fixed spelling of Russian example sentence

### DIFF
--- a/examples/widgets/unicode_textinput.py
+++ b/examples/widgets/unicode_textinput.py
@@ -4,10 +4,12 @@ from kivy.app import App
 from kivy.lang import Builder
 from kivy.properties import StringProperty, ObjectProperty
 from kivy.core.text import Label as CoreLabel
+from kivy.utils import reify
 from kivy.uix.boxlayout import BoxLayout
 from kivy.uix.floatlayout import FloatLayout
 from kivy.uix.spinner import SpinnerOption
 from kivy.uix.popup import Popup
+
 import os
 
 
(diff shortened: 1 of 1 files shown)


AndreMiras:
Great! But why all the double quotes changes?
Also if you could come with a unit test, that would be awesome!
Disclaimer: I'm just a random contributor, not a maintainer.

Fogapod:
> Great! But why all the double quotes changes?

I changed quotes because they were different with no sense, inserted double quotes according to [PEP257](https://www.python.org/dev/peps/pep-0257/).

> Also if you could come with a unit test, that would be awesome!

Unit tests for example?

AndreMiras:
> I changed quotes because they were different with no sense, inserted double quotes according to PEP257.
I see the point, thank you.
Sometimes in open source projects, you have more chances of getting approved doing one thing at a time, even though the quote thingy is a very simple change. So the idea could be to do a dedicated pull request for it for example.

> Unit tests for example?
Oh yes, I didn't realise this file was an example file.

Now that I realised In this is a code example, even my comment about the quotes in different PR makes less sense. Code examples are not that critical.

Thanks for the pull request and for the little chat :)

tito:
@Fogapod Same, why changing the quotes... Don't try to mix things like that with your actual fix. It doesn't help, and we can pick to merge only the fix.

Always best to seperate PR.

tito:
And Kivy source code use ''', not """. Please reopen the PR with only one commit related to the fix you actually want to merge :) Without breaking the current kivy style.

(and feel free to open another PR to change the ''' to """ for -all- the source code, not just the one you see.)

Fogapod:
@tito created new PR

Issue #5534, opened 2017-12-13
FileNotFoundError when input/event doesn't present device/capabilities/abs


thecodenomad:
To reproduce the bug:

Provide a /sys/class/input/event that doesn't have a folder structure containing 'device/capabilities/abs'

Logs and HelloWorld.py:

FileNotFoundError: [Errno 2] No such file or directory: '/sys/class/input/event21/device/capabilities/abs'

https://gist.github.com/thecodenomad/f1e6c4c669dc423aab5e057d6cffa9a0

PR #5535, opened 2017-12-13
Fix FileNotFoundError when sys path doesn't exist

thecodenomad:
This is intended to fix https://github.com/kivy/kivy/issues/5534

You get a FileNotFoundError when trying to access  '/sys/class/input/event<some integer>/device/capabilities/abs' when the full path doesn't exist. 

This may be an input device, but capabilities aren't listed so the fix just defaults "0".

### DIFF
--- a/kivy/input/providers/probesysfs.py
+++ b/kivy/input/providers/probesysfs.py
@@ -82,7 +82,12 @@ else:
 
         def get_capabilities(self):
             path = os.path.join(self.path, "device", "capabilities", "abs")
-            line = read_line(path)
+            line = "0"
+            try:
+                line = read_line(path)
+            except OSError:
+                return []
+
             capabilities = []
             long_bit = getconf("LONG_BIT")
             for i, word in enumerate(line.split(" ")):


KeyWeeUsr:
Ref https://github.com/kivy/kivy/commit/e18cd352b3aa966167a9a80b0b61437b1577d6ab#diff-2eeaed663bd0d25b7e608891384b7298 try to use `OSError` instead.

Also, shouldn't you instead just return `[]` from the exception? What's the point of going through the computation and return `[False]`?

thecodenomad:
Looks like it's getting denied due to other files I didn't modify. How is this situation typically handled?

thecodenomad:
Updated the fix based on comments.

tito:
Thanks you!

kivy/input/providers/probesysfs.py
```diff
@@ -82,7 +82,12 @@ def name(self):
 
         def get_capabilities(self):
             path = os.path.join(self.path, "device", "capabilities", "abs")
-            line = read_line(path)
+            line = "0"
+            try:
+                line = read_line(path)
+            except OSError:
```
AndreMiras: Is this not throwing a `IOError` (python2) or `FileNotFoundError` (python3) instead?

KeyWeeUsr: I think it returns OSError for Windows, but I'm not sure if OSError is the parent or IOError, sooo... test or check with cpython source I'd say :) It should handle FNF in either case for py3 iirc.

KeyWeeUsr: @AndreMiras This resolves FNF → https://github.com/python/cpython/blob/621cebe81b1e6c8de10425955bf532d31ee4df42/Objects/exceptions.c#L1256 (`OSError`), for the rest just append IOError if it still brags :)

AndreMiras: Oh didn't know about that thingy. Thanks

PR #5537, opened 2017-12-16
Make `disabled` an AliasProperty

tshirtman:
This allows rembembering the previous disabled state, which is useful
when using disabled at multiple levels.

Before that, disabling then un-disabling a widget, would un-disable all
the child widgets, which is a surprising and most of the time undesired
behavior.

Potential downsides are:

- you can't change the disabled state of a children of a disabled
  widget anymore, which may or may not be desired.
- the property isn't identified as a boolean anymore, which makes the
  inspector less useful.

I think it can be considered a bugfix, but it's arguably a slight api change, if so, the `versionchanged` tag should be updated accordingly.

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -315,6 +315,10 @@ class Widget(WidgetBase):
         callbacks to properties or events, as in the Kv language.
     '''
 
+    _disabled_count = NumericProperty(0)
+    '''internal, used to keep track of the current disabled state
+    '''
+
     __metaclass__ = WidgetMetaclass
     __events__ = ('on_touch_down', 'on_touch_move', 'on_touch_up')
     _proxy_ref = None
@@ -328,6 +332,7 @@ class Widget(WidgetBase):
             self._context = get_current_context()
 
         no_builder = '__no_builder' in kwargs
+        self._disabled_value = False
         if no_builder:
             del kwargs['__no_builder']
         on_args = {k: v for k, v in kwargs.items() if k[:3] == 'on_'}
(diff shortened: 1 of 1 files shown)


tshirtman:
Also, despite the `force_dispatch` option, it seems disabling an already disabled (through parent widget) widget doesn't work.

A test program can be found here.

https://gist.github.com/tshirtman/3db1281dc945f53a9ca9513c5252ad7f

```python
from kivy.app import App
from kivy.lang import Builder


KV = '''
BoxLayout:
    BoxLayout:
        orientation: 'vertical'
        Button:
            text: 'toggle fl'
            on_press: fl.disabled = not fl.disabled
        Button:
            text: 'toggle bl'
            on_press: bl.disabled = not bl.disabled
        Button:
            text: 'toggle b1'
            on_press: b1.disabled = not b1.disabled
        Button:
            text: 'toggle b2'
            on_press: b2.disabled = not b2.disabled
    FloatLayout:
        id: fl
        BoxLayout:
            id: bl
            pos_hint: {'center': (.5, .5)}
            size_hint: .5, .5
            Button:
                id: b1
            Button:
                id: b2
'''


class Test(App):
    def build(self):


if __name__ == '__main__':
Test().run()
```

tito:
It seems logic to not be able to change the disable state of a children if a parent is disabled.
Second is not really an issue imo.

My only feedback: can't _disabled_count be a simple python attribute, not a kivy property? Every new property increase the time to create the widget so...

tito:
Otherwise, OK to merge after the change

tshirtman:
Should be fine now :)

kivy/uix/widget.py
```diff
@@ -1305,7 +1303,32 @@ def on_opacity(self, instance, value):
     See :class:`~kivy.graphics.Canvas` for more information about the usage.
     '''
 
-    disabled = BooleanProperty(False)
+    def get_disabled(self):
+        return self._disabled_count > 0
+
+    def set_disabled(self, value):
```
SomberNight: Consider casting value to `bool`.

PR #5539, opened 2017-12-17
Window info

jakogut:
This PR implements a method for getting platform specific window
information from Kivy window providers.

Platform specific window system information is useful for
programatically acquiring a handle to the Kivy window.

When unable to retrieve windowing subsystem information, the function
returns `None`. Currently, only the SDL2 window provider actually
implements the function, using `SDL_GetWindowWMInfo`. The implementation
only returns information for Linux (X11/Wayland) and Windows.

Window system info is encapsulated in the `core.window.window_info` module, allowing other providers to create and return window info objects as necessary, using a standard interface.

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -131,6 +131,9 @@ class Keyboard(EventDispatcher):
         #: VKeyboard widget, if allowed by the configuration
         self.widget = kwargs.get('widget', None)
 
+    def get_window_info():
+        pass
+
     def on_key_down(self, keycode, text, modifiers):
         pass
 
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -10,6 +10,12 @@ from kivy.graphics.cgl import cgl_get_backend_name
 
 from cpython.mem cimport PyMem_Malloc, PyMem_Realloc, PyMem_Free
 
+IF UNAME_SYSNAME == 'Linux':
+    from window_info cimport WindowInfoX11, WindowInfoWayland
+
+IF UNAME_SYSNAME == 'Windows':
+    from window_info cimport WindowInfoWindows
+
 cdef int _event_filter(void *userdata, SDL_Event *event) with gil:
     return (<_WindowSDL2Storage>userdata).cb_event_filter(event)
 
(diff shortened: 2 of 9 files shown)


jakogut:
One problem that needs addressed is that the build will currently fail on Linux if development headers for X11 or Wayland are not available, and will fail on Windows if some headers (provided by the platform SDK?) are not installed.

I'm open to suggestions for a solution. I think in the ideal case, missing headers would simply disable the feature, and calling `Window.get_window_info()` would always return `None`.

jakogut:
@KeyWeeUsr Thanks for the feedback. I've added a USE_WAYLAND flag. Additionally, the X11 and Wayland structures are toggled on individually at compile-time based on their respective flags, which both default off.

In addition to not requiring both sets of development files for building on Linux, I think this is a cleaner solution, as it allows support on other platforms, such as macOS and Cygwin.

KeyWeeUsr:
@jakogut Now it seems pretty clean, thanks for all the changes! 🍡  I've restarted jobs for Mac and hopefully it'll pass even there (or someone could confirm it compiles there at least). Seems like a really good addition to me. :)

jakogut:
@KeyWeeUsr Thanks for the approval!

It seems AppVeyor has identified a build issue on Windows. 

> cythoning kivy\core/window/_window_sdl2.pyx to kivy\core/window\_window_sdl2.c
building 'kivy.core.window._window_sdl2' extension
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\x86_amd64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -Ikivy\include -IC:\Python36-x64\include -IC:\Python36-x64\include\SDL2 -I/usr/local/include/SDL2 -I/usr/include/SDL2 -IC:\Python36-x64\include -IC:\Python36-x64\include "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE" "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\ATLMFC\INCLUDE" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\winrt" /Tckivy\core/window\_window_sdl2.c /Fobuild\temp.win-amd64-3.6\Release\kivy\core/window\_window_sdl2.obj
_window_sdl2.c
kivy\core/window\_window_sdl2.c(5757): error C2039: 'hinstance': is not a member of '<unnamed-tag>'
C:\Python36-x64\include\SDL2\SDL_syswm.h(201): note: see declaration of '<unnamed-tag>'
python :  error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\BIN\\x86_amd64\\cl.exe' failed with exit status 2

After doing a little digging, it appears the `hinstance` variable that's part of the `win` structure was introduced in SDL2 version 2.0.6, and the version that's packaged for `kivy.deps.sdl2_dev` doesn't have it. I removed it to fix the build issue.

Here's the source for this information: https://wiki.libsdl.org/SDL_SysWMinfo

After this change, it compiles in my Windows environment again.

KeyWeeUsr:
@jakogut Perhaps after the testing adding a simple unittest might be a good idea. I think there isn't really much to test (because different platforms/providers), however stuff such as handle, display/window name, etc could be just checked if they are not `NULL` or some other silly value. This way the window_info part would be at least triggered while running the tests. :)

tito:
+1 for the PR, very good idea

jakogut:
I wrote a simple test to check that the window info object is either None, or is populated with non-zero data. I tested on Linux with both X11 and Wayland. I attempted to test on Windows, but received an import error I believe is unrelated to this branch. With any luck, AppVeyor will show this test passing.

jakogut:
Looks like the Windows test passed. I'm removing the WIP tag.

tito:
@jakogut About "One problem that needs addressed is that the build will currently fail on Linux if development headers for X11 or Wayland are not available" => Will this fail on RPI platform?

jakogut:
@tito I don't have an RPi to test with, but that issue is fixed. Previously, this branch added code that required X11 and Wayland headers when compiling on a Linux platform.  @KeeWeeUsr pointed out that this was a bad idea, and it was changed to use the USE_X11 and USE_WAYLAND flags respectively to enable features that require those headers.

In effect, Kivy will compile the same way it always has. If a user wants X11 or Wayland window info, they have to install the appropriate development files, and enable the appropriate flags during the build. Otherwise, `Window.get_window_info()` will return `None`.

tito:
Thanks you for this PR!

FI-Mihej:
There is a forgotten `return` for Windows platform (broken as result). Fixed in https://github.com/kivy/kivy/pull/7619

PR #5556, opened 2018-01-01
Fixed loading fonts with dot in name, fixed spelling in Russisn examle

Fogapod:
Reopenning #5526  without style changes


tito:
Thanks you!

PR #5564, opened 2018-01-05
Update basic.rst

yhcharles:
overload -> override
https://stackoverflow.com/questions/12893907/are-polymorphism-overloading-and-overriding-similar-concepts


KeyWeeUsr:
Thank you!

PR #5567, opened 2018-01-07
Correct typo 'setdefaults' to 'setdefault'

ismailof:
In file `/kivy/uix/effectwidget.py`, there is this line:
  `kwargs.setdefaults("with_stencilbuffer", True)`
which throws an AttributeError `('dict' object has no attribute 'setdefaults')`

This PR corrects the method `setdefaults `to `setdefault`.


inclement:
Thanks!

PR #5577, opened 2018-01-16
 window_x11: fix python3 TypeError

jakogut:
Encode Window.title as byte-string before attempting to set the x11 window
title. This avoids "TypeError: expected bytes, str found" when using the
x11 window provider with Python 3.


KeyWeeUsr:
Thanks! 🍡

Issue #4492, opened 2016-07-26
ZeroDivisionError in mtdev.py


davelawr:
running kivy 1.9.2-dev0 on raspberrypi my application occasionally produces this exception:

```
Exception in thread Thread-1:                                                                                  
 Traceback (most recent call last):                                                                             
   File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner                                       
     self.run()                                                                                                 
   File "/usr/lib/python2.7/threading.py", line 763, in run                                                     
     self.__target(*self.__args, **self.__kwargs)                                                               
   File "/usr/local/lib/python2.7/dist-packages/kivy/input/providers/mtdev.py", line 305, in _thread_run        
     range_max_minor)                                                                                           
   File "/usr/local/lib/python2.7/dist-packages/kivy/input/providers/mtdev.py", line 214, in normalize
     return (value - vmin) / float(vmax - vmin)
 ZeroDivisionError: float division by zero
```

seems to be related to keyboard input.

full kivy 'debug' leading up to the exception attached.

[kivylog.txt](https://github.com/kivy/kivy/files/383938/kivylog.txt)

tshirtman:
from your log
```
[INFO   ] [MTD         ] </dev/input/event0> range position X is 0 - 0
[INFO   ] [MTD         ] </dev/input/event0> range position Y is 0 - 0
[INFO   ] [MTD         ] </dev/input/event0> range touch major is 0 - 0
[INFO   ] [MTD         ] </dev/input/event0> range touch minor is 0 - 0
[INFO   ] [MTD         ] </dev/input/event1> range position X is 0 - 0
[INFO   ] [MTD         ] </dev/input/event0> range pressure is 0 - 255
[INFO   ] [MTD         ] </dev/input/event2> range position X is 0 - 800
[INFO   ] [MTD         ] </dev/input/event1> range position Y is 0 - 0
[INFO   ] [MTD         ] </dev/input/event0> axes invertion: X is 0, Y is 0
[INFO   ] [MTD         ] </dev/input/event2> range position Y is 0 - 480
[INFO   ] [HIDMotionEvent] using <SIGMACHIP USB Keyboard>
[INFO   ] [MTD         ] </dev/input/event1> range touch major is 0 - 0
[INFO   ] [MTD         ] </dev/input/event0> rotation set to 0
[INFO   ] [MTD         ] </dev/input/event2> range touch major is 0 - 0
[...]

PR #5593, opened 2018-02-01
Fix float division by zero

skallet:
Fix for the issue mentioned in #4492.

### DIFF
--- a/kivy/input/providers/mtdev.py
+++ b/kivy/input/providers/mtdev.py
@@ -220,7 +220,8 @@ else:
                     queue.append((action, touch))
 
             def normalize(value, vmin, vmax):
-                return (value - vmin) / float(vmax - vmin)
+                divisor = float(vmax - vmin) if vmax > vmin else 1
+                return (value - vmin) / divisor
 
             # open mtdev device
             _fn = input_fn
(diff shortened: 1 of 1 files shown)


kivy/input/providers/mtdev.py
```diff
@@ -220,7 +220,8 @@ def process(points):
                     queue.append((action, touch))
 
             def normalize(value, vmin, vmax):
-                return (value - vmin) / float(vmax - vmin)
+                divisor = float(vmax - vmin) if vmax > vmin else 1
```
KeyWeeUsr: As it hasn't returned absolute values before explicitly, I believe this is wrong (what about the negative values?). I'm not sure what are those values, but something like `... if vmax != vmin else 1` should be good at the first look, however if there is a possibility at least one of the values is a float, I'd rather like to see more hm, float-friendly handling i.e. let the error appear, but handle it directly:

    try:
        return (value - vmin) / float(vmax - vmin)
    except ZeroDivisionError:  # it's both in py2 and py3
        return (value - vmin)

skallet: Good point, I modified code.

akshayaurora:
@skallet thanks 👍

PR #5612, opened 2018-02-11
raise exception when trying to add Widget with a parent to Window

### LINKED ISSUE #5508: No exception is thrown if root widget already has a parent
### Versions

* Python: 3.6.1
* OS: Windows
* Kivy: 1.10

### Description

Using a root widget that already has a parent doesn't throw an exception.

### Code and Logs

```python
from kivy.factory import Factory as F

A = F.BoxLayout()
B = F.Button()
A.add_widget(B)

runTouchApp(B)
```

With BoxLayout for A this floods console with clock iteration errors (using Widget for A doesn't)

tshirtman:
fixes #5508

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -1240,6 +1240,13 @@ class WindowBase(EventDispatcher):
 
     def add_widget(self, widget, canvas=None):
         '''Add a widget to a window'''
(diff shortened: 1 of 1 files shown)


akshayaurora:
@tshirtman 👍

Issue #5450, opened 2017-10-23
Needs a proper way to reorder on the Z-index


JellyWX:
Kivy currently has no proper way to reorder elements on the Z-index. The only way is by sorting the child list and then using `remove_widget(i); add_widget(i)`. This is not ideal for most situations, since it means that for an instant the widget's parent becomes `NoneType` and the program crashes if you're doing positioning relative to parent size.
TL;DR: Please add a proper way to order element on the z-index, possibly just a `.to_front()` would work nicely.

tito:
When you talk about z-index, it's the index of the widget in the its parent children list right?

JellyWX:
yes, basically. however, simply reordering the children list doesnt work.

kuguma:
here is workaround:

`
class Widget(WidgetBase):
...
    def to_front(self, widget):
        target = None
        if widget.canvas in self.canvas.children:
            target = self.canvas.children
        elif widget.canvas in self.canvas.after.children:
            target = self.canvas.after.children
        elif widget.canvas in self.canvas.before.children:
            target = self.canvas.before.children
        else:
            return
        top = len(target)-1
        i = target.index(widget.canvas)
        target[top],target[i] = target[i],target[top]
...
`

JellyWX:
looks very neat, havent got a set up to properly test it however. would be nice if it could somehow fit into a PR maybe, but good for future reference. Thank you

Nasjoe:
Hi ! Can you give us an example with two widgets, like an image ? Many Thanks !

JellyWX:
https://github.com/JellyWX/sample-ordering-kivy

sorry for slow reply!

⁣Get BlueMail for Android ​

PR #5620, opened 2018-02-16
[WIP] Widget.swap_widget(), a hack to reorder children without transitioning parent via None

bionoid:
### Description

This is a "KISS" approach to #5450. It kinda sucks to tag objects with a hostile attribute, maybe it should be a BooleanProperty, or set a class attribute to avoid the hasattr? Not sure.. Other options are to make `add_widget` and `remove_widget` wrappers for lower-level code that can also swap, or possibly add a `swap` argument to their signature (but I think this will break a lot of code that overrides it, maybe even in core). 

I landed on the proposed implementation; it seems a little overkill to add a property to Widget just to have this flag alive across a few infrequent function calls, right? And equally overkill to incur extra function call by making add/remove into wrappers..

One thing to note, this suppresses the `on_parent` event of the swapped widgets, but not the `on_children` event of the container. That seems acceptable to me, but in theory user code could depend on these events occuring in parallell.

Share your opinions, please? Suggestions for a better approach?

### Usage example (must merge PR to test)

```python
from kivy.app import App
from kivy.lang import Builder

KV = '''
#:import F kivy.factory.Factory

<TestBase@ToggleButtonBehavior+Label>:
    bgcolor: [0, 0, 0, 0]
    font_size: '25sp'
    on_press: app.press_target(self)
    on_parent: print('on_parent: %r now a child of %r' % (self, self.parent))
    canvas.before:
        Color:
            rgba: root.bgcolor
        Rectangle:
            pos: self.pos
            size: self.size

<TestOne@TestBase>:
    bgcolor: [1, 0, 0, self.state == 'down' and 1 or .3]

<TestTwo@TestBase>:
    bgcolor: [0, 0, 1, self.state == 'down' and 1 or .3]

<Headline@Label>:
    bgcolor: [0, 0, 0, 0]
    font_size: '15sp'
    size_hint_y: None
    height: '40dp'
    canvas.before:
        Color:
            rgba: root.bgcolor
        Rectangle:
            pos: self.pos
            size: self.size
BoxLayout:
    orientation: 'vertical'
    Headline:
        text: 'Container ONE:'
        bgcolor: [1, 0, 0, 1]
    GridLayout:
        id: ONE
        cols: 4
        on_parent:
            for n in range(8): self.add_widget(F.TestOne(text='ONE %d' % (n, )))
    Headline:
        text: 'Container TWO:'
        bgcolor: [0, 0, 1, 1]
    GridLayout:
        id: TWO
        cols: 4
        on_parent:
            for n in range(8): self.add_widget(F.TestTwo(text='TWO %d' % (n, )))
'''

class SwapApp(App):
    tgt = Factory.ObjectProperty(None, allownone=True)

    def press_target(self, wid):
        if not self.tgt:
            self.tgt = wid
        else:
            self.tgt.parent.swap_widget(self.tgt, wid)
            self.tgt = None
            for w in self.root.ids.ONE.children:
                w.state = 'normal'
            for w in self.root.ids.TWO.children:
                w.state = 'normal'

    def build(self):

SwapApp().run()
```

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -526,7 +526,7 @@ class Widget(WidgetBase):
                 'Widget instances cannot be added to themselves.')
         parent = widget.parent
         # Check if the widget is already a child of another widget.
-        if parent:
+        if parent and not hasattr(widget, '_kivy_hostile_swap_flag'):
             raise WidgetException('Cannot add %r, it already has a parent %r'
                                   % (widget, parent))
         widget.parent = parent = self
@@ -583,7 +583,9 @@ class Widget(WidgetBase):
             self.canvas.after.remove(widget.canvas)
         elif widget.canvas in self.canvas.before.children:
             self.canvas.before.remove(widget.canvas)
-        widget.parent = None
+        # swap_widget hack to avoid parent transition via None
+        if not hasattr(widget, '_kivy_hostile_swap_flag'):
+            widget.parent = None
 
     def clear_widgets(self, children=None):
         '''
@@ -601,6 +603,53 @@ class Widget(WidgetBase):
         remove_widget = self.remove_widget
         for child in children[:]:
             remove_widget(child)
+
+    def swap_widget(self, a, b):
+        '''Swaps a child widget for another widget. This avoids transitioning
+        the parent property via None when possible. Note that the `children`
+        property of this class still emits all events.
+
+        :Parameters:
+            `a`: :class:`Widget`
+                Widget to swap out; this must be an existing child widget.
+            `b`: :class:`Widget`
+                Widget to replace `a`. If it has an existing parent, it
+                will be removed from it.
+
+        .. versionadded:: 1.10.1
+        '''
+        if a is b:
+            return
+        children = self.children
+        cidx = self.children.index
+        a_is_my_child, a_idx = a in children and (1, cidx(a)) or (0, None)
+        b_is_my_child, b_idx = b in children and (1, cidx(b)) or (0, None)
+        if not a_is_my_child:
+            raise WidgetException(
+                'swap_widget() "a" widget %r is not in self.children.' % (a, ))
+        add_widget = self.add_widget
+        remove_widget = self.remove_widget
+        # If b has no parent, just do plain remove/add
+        b_parent = b.parent
+        if not b_parent:
+            remove_widget(a)
+            add_widget(b, index=a_idx)
+            return
+        # b has a parent, raise the flag for add_widget/remove_widget
+        a._kivy_hostile_swap_flag = True
+        b._kivy_hostile_swap_flag = True
+        remove_widget(a)
+        if b_is_my_child:
+            add_widget(a, index=b_idx)
+            remove_widget(b)
+            add_widget(b, index=a_idx)
+        else:
+            extern_b_idx = b_parent.children.index(b)
+            b_parent.remove_widget(b)
+            b_parent.add_widget(a, index=extern_b_idx)
+            add_widget(b, index=a_idx)
+        del a._kivy_hostile_swap_flag
+        del b._kivy_hostile_swap_flag
 
     def export_to_png(self, filename, *args):
         '''Saves an image of the widget and its children in png format at the


bionoid:
I added a `move_widget` method as well, to handle "bring to front" / "send to back" without inserting a dummy widget at the desired index. Alternatively, swap signature could be `def swap_widget(self, a, b=None, index=None)` (and require one of the arguments)... or we could make b=None move to index 0, or something. But the index change is not a swap operation, so it seems weird to combine them.

Maybe `swap_child` and `move_child` are better names?

[Updated test example code](https://gist.github.com/bionoid/2bbcd9b05f1d5f1a7591da991b0d5d50) for `move_widget`

rnixx:
To me, dedicated ``swap_widget`` and ``move_widget`` seems to be the better solution than one signature with a hard to explain behavior. Also ``*_widget`` is better than ``*_child`` in analogy to ``add_widget`` and ``remove_widget``.

rnixx:
Referring your doubts about the hostile attribute. What about keeping this information on a threading.local() instance?

bionoid:
> To me, dedicated swap_widget and move_widget seems to be the better [...]

I agree, no better ideas so far... but maybe there is a superior API to be discovered :)

> Also *_widget is better than *_child in analogy to add_widget and remove_widget.

I thought `swap_widget` was pretty clear, but `move_widget` is really more like `set_child_index`, if I were to describe what it does. This lead me to question the initial choice. If anyone has suggestions for better names, granularity, etc please speak up!

.. The "proper" solution is probably more along the lines of adding `swap_out=None` and `swap_in=None` arguments to add/remove_widget signatures. But I think that has to be a 2.0-severity API break??


> What about keeping this information on a threading.local() instance?

Yeah, maybe, seems similar to using a class attribute, sans future kivy thread safety?

```python
class Widget(...):
    _being_swapped = set()
```

Now could `.update((a, b))` before and remove after. In theory it can probably grow past two widgets, since events are dispatched during the process. Is there any benefit to wrapping this in thread.local? I didn't try using a set like this, I just assumed it was going to be slower, but that may not be true/signifficant in the real world..

rnixx:
> but move_widget is really more like set_child_index, if I were to describe what it does

Then maybe just ``set_widget_index`` if it's what actually happens?

> The "proper" solution is probably more along the lines of adding swap_out=None and swap_in=None arguments to add/remove_widget signatures.

I'm not really convinced on this one. It probably just causes a lot of work in kivy core and code depening on the existing API for very little benefit. Then better keep discussing the newly introduced function names :)

I have not studied the code in detail, but the places where you mark widgets with ``_kivy_hostile_swap_flag``, and delete it afterwards, look like a perfect candidate for a context manager.

e.g:

```python
with swap_flag(a, b):
    # do something
```

and inside ``swap_flag``, widgets might be kept in a ``threading.local``, or even a global singleton (but then it needs to be syncronized), which can be used then to check whether we're currently swapping.

So the context manager may look similar to:

```python
class swap_flag(object):
    _data = threading.local()
    _data.widgets = set()

    def __init__(self, *widgets):
        self._widgets = widgets
        for w in widgets:
            self._data.widgets.add(w)

    def __enter__(self):
        # if no threading.local, acquire lock here
        pass

    def __exit__(self, type_, value, traceback):
        for w in self._widgets:
            self._data.widgets.remove(w)
        # if no threading.local, release lock here

    @classmethod
    def on(cls, ob):
        return ob in cls._data.widgets
```

And the places where you need to check for swapping then look like:

```python
if swap_flag.on(widget):
    # do something
```

This way you keep the information away from the widget entirely and have an IMHO pretty nice API as well.

bionoid:
Thanks, yeah `set_widget_index` seems good, avoids the implication of "position" from the word "move". Will update later!

> I'm not really convinced on this one. It probably just causes a lot of work in kivy core [...]

You are right, it just seemed like a natural combination of the existing code and desired API; like `parent.add_widget(new, swap_out=old)` instead of specifying the index. But surely it would become a mess over time :)

> [...] look like a perfect candidate for a context manager.

Honestly I struggle to see the "objective" benefit over the proposed solution. Is there a thread safety problem with the PR, or does it solve another problem the PR doesn't?

One of the reasons I landed on the hostile attribute, is that it (seems to) have the least impact on `remove_widget`. We need to test for the flag on every run (this is not true of `add_widget`, we check flag only if the widget already has a parent, and were doing this test already). If we ignore `swap_widget` for a moment, and just consider the final lines of `remove_widget` in PR:

```python
    if not hasattr(widget, '_kivy_hostile_swap_flag'):
        widget.parent = None
```

This is basically why I landed on a hostile attribute - `hasattr` is at least optimized and reasonably fast (maybe it can be further optimized as `flag in widget.__dict__` or something?).. Of course I didn't benchmark it, but seems like your example above would do a dot lookup for `.on`, then call `on`, then resolve set membership. I am pretty sure it can't beat hasattr... ??

.. I am not the ultimate judge of how important it is to conserve cycles in remove_widget of course, but from my perspective, the hostile attribute seems like the best compromise so far. My biggest objection is philosophical; it is "action at a distance". But the distance between these methods is fairly short, and it's obvious what the code does.. at this stage, anyway

rnixx:
> Honestly I struggle to see the "objective" benefit over the proposed solution. Is there a thread safety problem with the PR, or does it solve another problem the PR doesn't?

There is no thread safety problem with your solution. I originally thought you were uncontent with the fact you introduced a hostile attribute with your solution. Thus I thought you want to get rid of it.

The proposed context manager just seemed to be a more elegant syntax, it easily would be possible to manage the hostile attribute inside the context manager as well.

> I am pretty sure it can't beat hasattr... ??

I'm not 100% sure on this. Depends whether a try/except is involved in ``hasattr``. Asking small iterables for containment is faster than catching an exception.

bionoid:
> Thus I thought you want to get rid of it.

Ah, yes it felt that way, but I was not sure what to do with it.. :) but it turns out you are right about the hasattr catching an exception, so I replaced it by a set in class scope. [Here is some benchmark code](https://gist.github.com/bionoid/480de9c890629a35b58c3e05845ee38f) of questionable quality, and its output on my system:

```
--------- CHECK FLAG ---------
check false: hasattr(n, "myflag"): 3.018791730606311
check true: hasattr(n, "myflag"): 0.9397991329916247
check flase: "myflag" in n.__dict__: 0.4747848827474246
check true: "myflag" in n.__dict__: 0.4816195456050334
--------- SET MEMBERSHIP ---------
check false: n.wrapper("myflag"): 1.372010989652261
check true: n.wrapper("myflag"): 1.3710775905929324
check false: "myflag" in n.data: 0.40444975934843264
check true: "myflag" in n.data: 0.4060080636529477
raise flag: n.data.update(("myflag", )): 1.3366089785189175
(apples-to-oranges) lower false flag: n.data.discard(("myflag", )): 0.8306539042
502052
(apples-to-oranges) lower true flag: n.data.discard(("myflag", )): 2.07874099334
42156
--------- RAISE FLAG ---------
raise flag, n.myflag = True: 0.4131688150289268
raise flag, n.__dict__["myflag"] = True: 0.5558823934779955
--------- LOWER FLAG ---------
lower flag, del n.myflag: 0.8853279814152994
lower flag, del n.__dict__["myflag"]: 1.0342680324458087
```

given this, `set` looks like the best compromise to me now..

rnixx:
Is there something missing in the implementation or why is this PR still WIP?

bionoid:
> Is there something missing in the implementation or why is this PR still WIP?

Yes - integration with other widgets besides those tested (this is what the FIXME alludes to). 

Like if you attempt to swap with a ScreenManager, it will only have the widget in `.children` if it is the currently active screen. Could adapt the code to rely only on `.parent`, and wrap attempts to get the index in a try/except. Or introduce an `allow_external_swap` type of BooleanProperty and set it to false by default in ScreenManager/recycle stuff/other widgets that override normal add/remove behavior in a problematic way.

There is also currently an issue with specifying index in add_widget calls; not all overrides allow it (including ScreenManager iirc). Should these signatures be given an index argument (to be ignored), or should swap_widget be adapted to cause minimal damage? (ie by never issuing `add_widget(n, index=None)` or `0`)..

tito:
Widget._active_widget_swaps and the add_widget changes doesn't make it work in all widgets right now.

IMO, handling z-index is sort-of the same issue as the one i want to work on soon: hidden attribute (hidden=True / visible=False / visibility=hidden whatever, you get the idea).

z-index have only a impact on the order in children, while hidden also have an impact on either the widget exist in children or not.

I think if hidden implementation would be there, you would have a place to store all the "linked" children somewhere, and then the real ".children" attribute that actually have the current list of children to be drawed, ordered.

Therefore, working on a special "children" attribute (list that auto order according to z-index), and reflect the canvas ordering (i've never seen a widget adding a instruction itself to the parent canvas, but all our widgets add their own instructions in the canvas, so we could assume this to rebuild parent canvas).

Otherwise, we'll tend to check too many attribute just to see if we can actually use the widget or not.

bionoid:
Having a hidden-type of attribute would be nice  of course. Personally I consider the z-order issue secondary, I haven't really been annoyed in that situation. My personal pet peeve is that we have no way to change parent without intermediate None value. It's sort of the same problem, but not quite.

tito:
This is too much a hack that IMO should not be as it is in Kivy.

PR #5626, opened 2018-02-20
Fix typo in docs.

germn:
`log_maxfiles` is [correct](https://github.com/kivy/kivy/blob/97d675e0669f600f62b074a6cc9efb409091ced0/kivy/logger.py#L120)


akshayaurora:
@germn thanks 👍

PR #5635, opened 2018-02-25
Add bidirectional communication support to OSC

Cheaterman:


kivy/lib/osc/oscAPI.py
```diff
@@ -113,9 +113,12 @@ def _queue_message(self, message):
 
 def init() :
     '''instantiates address manager and outsocket as globals
+        returns local port of outsocket for bidirectional communication
     '''
     global outSocket
     outSocket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+    outSocket.bind(('0.0.0.0', 0))
```
tito: No way to make it conditionnal?

Making a server by default listening on a random port to all IP if you use OSC is a little bit too open :)
Imo, init could have a listen_host, listen_port, both set to None, but if set, then listen for it.

Cheaterman: I think that sounds good - the real reason I'm binding the socket here is because I need the outgoing port to be set when I use getsockname() on the next line (and later), I thought we could simply bind to 127.0.0.1 instead of 0.0.0.0 (to avoid changing the API) but then I wouldn't be able to receive remote data from that socket. Let's do this :-)

Cheaterman: Done :-)

Cheaterman: (BTW I didn't use your naming convention because the rest of the file is camelCased :-/ )

PR #5647, opened 2018-03-09
Disable welcome message for new issues

bionoid:
I find this annoying and propose to disable it. I just don't think this adds any value at all.

The bot's auto-response doesn't make up for lacking human resources to respond to and deal with issues. I'd argue that it makes it worse.

Every message lands in repository watcher's inbox, including mine, I'd rather not have it.

Again feel free to close, just voicing my opinion.

### DIFF
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -1,8 +1,8 @@
 # Configuration for new-issue-welcome - https://github.com/behaviorbot/new-issue-welcome
 
 # Comment to be posted to on first time issues
-newIssueWelcomeComment: |
-  👋 Thanks for opening your first issue here! Be sure to follow the issue template!
+# newIssueWelcomeComment: |
+#  👋 Thanks for opening your first issue here! Be sure to follow the issue template!
 
 # Configuration for new-pr-welcome - https://github.com/behaviorbot/new-pr-welcome
 


suroh:
I whole heartedly agree!

inclement:
I don't feel that strongly about it, partly because I disabled most of the notifications, but I agree more than I disagree. Especially since the instruction to use the issue template is often ignored, and I'm not sure if the template is that useful anyway.

bionoid:
> partly because I disabled most of the notifications

Can I ask why you do that? I am slightly surprised to be honest. One of the reasons I want this message gone is precisely because it can encourage people who actually follow development to disable notifications. That means even less chance of getting a response if you are looking for help to isolate a bug or progress a pull request to completion.

inclement:
I disabled the Kivy repository messages even before this message was added, because I was getting more than I wanted to read anyway, and because I've mostly focused my efforts on python-for-android for quite a while. I do keep a closer watch on the p4a-relevant repositories.

I also still keep an eye on the commit summaries from the irc bot, and regularly interact with issue from there (including this one).

bionoid:
This seems like a straightforward merge to me, but it's been sitting here for 3 months.

I don't really see the point in having it sit around, anyone can make the change if the decision is made, so I'm closing it.

For future consideration, my core arguments are:

* There is no message content, it's a clapping hand icon congratulating the user on having filled in a form

* Thousands of copies of this pointless message are distributed to repository watchers over a year. It screws up at least some e-mail conversation tracking, making it appear like you have a conversation with welcome[bot]

* It deteriorates the website UX, since issues get a comment it looks like someone has responded to it, but you won't know it's the bot until you click the issue.

PR #5649, opened 2018-03-09
Try first utf-8 encoding by default on reading .kv files. Fixes #5154

### LINKED ISSUE #5154: Loading UTF8 KV file with Japanese (multi-byte?) text crashes Kivy under Windows
### Versions

* Python: 3.4.4
* OS: Windows 10 Home
* Kivy: 1.9.1
* Kivy installation method: followed https://kivy.org/docs/installation/installation-windows.html#installation

### Description

Loading a UTF8 KV file with Japanese (and perhaps other multi-byte?) text crashes Kivy under Windows. The example below uses the text field, but even a comment in Japanese will cause the crash. Changing the KV file encoding to Shift-JIS (approximately cp932) fixes the issue. 

Full details can be found in this user support group post: https://groups.google.com/forum/#!topic/kivy-users/dlR0f6tmANw

The essence is that, this fact combined with the fact that Linux-type systems support UTF8 KV files and NOT Shift-JIS KV files makes cross-platform development very difficult. 

Issue #4003 mentions a display glitch, but I believe a crash is more serious and may merit being dealt with. 

### Code and Logs

The following will reproduce the issue; the commented out alternatives are to show the options that work properly (without crashing).

utftest.py
```python
# -*- coding: utf-8 -*-

# test using kivy KV File in UTF8 with Japanese text

from kivy.app import App
from kivy.lang import Builder

# # This will load properly (ignoring the fact that a font set needs to be loaded separately for proper display)
# Builder.load_string('''
# <UtfTestScreen>
#     Label:
#         # font_name: '../SourceHanSerifJP/SourceHanSerifJP-Medium.otf'
#         id: inst_txt
#         text: u'テストテキスト'
#
# ''')


class UtfTestScreen(BoxLayout):
    def __init__(self, **kwargs):
        super(UtfTestScreen, self).__init__(**kwargs)

        # # Test using Japanese text from py file to kv specified parts
        # self.ids.inst_txt.text = 'テストテキスト'


class UtfTestApp(App):
    def build(self):
        return UtfTestScreen()


if __name__ == "__main__":
    UtfTestApp().run()

```

utftest.kv **(saved in UTF8 encoding)**
```
<UtfTestScreen>
    Label:
        # font_name: '../SourceHanSerifJP/SourceHanSerifJP-Medium.otf'
        id: inst_txt
        text: u'テストテキスト'

```

caused an error as below:
```
   File "utftest.py", line 34, in <module>
     UtfTestApp().run()
   File "C:\Python34\lib\site-packages\kivy\app.py", line 801, in run
     self.load_kv(filename=self.kv_file)
   File "C:\Python34\lib\site-packages\kivy\app.py", line 598, in load_kv
     root = Builder.load_file(rfilename)
   File "C:\Python34\lib\site-packages\kivy\lang.py", line 1831, in load_file
     data = fd.read()
 UnicodeDecodeError: 'cp932' codec can't decode byte 0x86 in position 141: illegal multibyte sequence

Process finished with exit code 1
```

jsbueno:
Add optional 'encoding' parameter loading .kv files, defaulting
to None. If not passed, also tries native system encoding
if file fails loading as utf-8.


tito:
Tests are broken with the encoding. Dunno why the error is not explicit.

matham:
We can delay this until we drop python2 support and then simplify it a little. Also, I would remove the warning as it doesn't add much and would log a warning everytime it used the default.

tshirtman:
python2 was dropped in master, so it's likely worth revisiting this one.

kivy/lang/builder.py
```diff
@@ -285,21 +285,63 @@ def create_from(cls, builder):
...
+                'File "{0}" failed to open with  "utf-8" codec. '
+                'Trying native system encoding now. Note that '
+                'this may cause text differences when the project '
+                'is run in other systems'.format(filename)
+            )
+            encoding = sys.getdefaultencoding()
+            with open(filename, 'r') as fd:
+                data = fd.read()
+
+             # remove bom ?
```
tshirtman:
```suggestion
            # remove bom ?
```

pythonic64:
Closing in favor of https://github.com/kivy/kivy/pull/6880 which is merged in master.
@jsbueno Thank you for your contribution. Your commits are preserved in rebased  https://github.com/kivy/kivy/pull/6880.

Issue #5646, opened 2018-03-08
CI builds have been failing for a year


bionoid:
### Description

It seems like the most recent successful CI build was in January 2017. In my opinion it does more harm than good at this stage. I propose to disable failing builds until the issues have been resolved.

KeyWeeUsr:
The Travis builds were mostly working except some PEP8 issues, however Appveyor builds are half-successful with the jobs because of Angle running on < Py 3.5 though it works (or worked at least) just fine when on a PC with correctly installed drivers.

Then there is the issue with nightly wheels on Appveyor because they made the scheduled builds disabled by default.

bionoid:
Aha, I don't know about Travis/Appveyor, can it be fixed in a reasonable timeframe? I just think the value is gone when CI doesn't tell you about the state of the pull request. Right off the bat it's discouraging to anyone that stops by to have a look, and it's confusing to people that open the first PR, and it's confusing to me although I've been around a while. 

From my perspective, it seems important to fix ASAP. If a resolution is not on the horizon, I would prefer to disable stuff until a clean PR doesn't result in a reported failure.

Feel free to close of course, just voicing my opinion here

Zen-CODE:
I'll second bionoid here. It defeat the point of having them if they permanently fail: a failure should be alarming. I'll look into this. Some builds are working, so it may be a matter of small tweaking...

Zen-CODE:
Okay, I've managed to get 4 of the 6 builds passing with in Travis.

https://travis-ci.org/kivy/kivy

The remaining errors are bizarre. I can' reproduce them and running the tests on my machine works fine. About to experiment with different cython versions, but any insight would be help. Pretty weird. The inspector test fails with
```
   File "/home/travis/build/kivy/kivy/kivy/modules/inspector.py", line 382, in on_window_children
     win.add_widget(self)
   File "/home/travis/build/kivy/kivy/kivy/core/window/__init__.py", line 1247, in add_widget
     (widget, widget.parent)
 kivy.uix.widget.WidgetException: Cannot add <kivy.modules.inspector.Inspector object at 0x2b8dae99ace0> to window, it already has a parent <kivy.core.window.window_sdl2.WindowSDL object at 0x2b8d9941c660>
 Exception ignored in: 'kivy.properties.observable_list_dispatch'
 Traceback (most recent call last):
   File "kivy/properties.pyx", line 571, in kivy.properties.Property.dispatch
[...]

Issue #5645, opened 2018-03-07
Closing inspector throws exception after merging #5612


bionoid:
### Versions

* Python: 3.6.4
* OS: Linux
* Kivy: master

### Description

Closing inspector throws an exception after merging #5612. Seems like the issue is that [when calling remove_widget here](https://github.com/kivy/kivy/blob/1.10.0/kivy/modules/inspector.py#L381), removing from children list (which triggers this callback) happens before clearing .parent property

Edit to add: The exception is caught, doesn't crash the application

### Code and Logs

`python any.py -m inspector`, press ctrl-e and hit "x" to close the popup, result:

```
 Traceback (most recent call last):
   File "kivy/properties.pyx", line 571, in kivy.properties.Property.dispatch
   File "kivy/_event.pyx", line 1214, in kivy._event.EventObservers.dispatch
   File "kivy/_event.pyx", line 1120, in kivy._event.EventObservers._dispatch
   File "/home/kivy/kivy/kivy/modules/inspector.py", line 382, in on_window_children
     win.add_widget(self)
   File "/home/kivy/kivy/kivy/core/window/__init__.py", line 1247, in add_widget
     (widget, widget.parent)
 kivy.uix.widget.WidgetException: Cannot add <kivy.modules.inspector.Inspector object at 0x7f5f8645f1e8> to window, it already has a parent <kivy.core.window.window_sdl2.WindowSDL object at 0x7f5f9654a730>
 Exception ignored in: 'kivy.properties.observable_list_dispatch'
 Traceback (most recent call last):
   File "kivy/properties.pyx", line 571, in kivy.properties.Property.dispatch
   File "kivy/_event.pyx", line 1214, in kivy._event.EventObservers.dispatch
   File "kivy/_event.pyx", line 1120, in kivy._event.EventObservers._dispatch
   File "/home/kivy/kivy/kivy/modules/inspector.py", line 382, in on_window_children
     win.add_widget(self)
   File "/home/kivy/kivy/kivy/core/window/__init__.py", line 1247, in add_widget
     (widget, widget.parent)
 kivy.uix.widget.WidgetException: Cannot add <kivy.modules.inspector.Inspector object at 0x7f5f8645f1e8> to window, it already has a parent <kivy.core.window.window_sdl2.WindowSDL object at 0x7f5f9654a730>
```

KeyWeeUsr:
The same thing makes some tests in CI fail.

tshirtman:
[...]

PR #5654, opened 2018-03-13
Remove widget from pre-existing parent. references #5508, #5612, #5645

Zen-CODE:
To my mind, this approach makes more sense the throwing the error. 

a. It fixes/avoids this issue, which does seem relatively artificial in the first place: https://github.com/kivy/kivy/issues/5508

b. More often  than not, this will be desired behavior. i.e. it just works.

c. It should fix various other issues. e.g. https://github.com/kivy/kivy/issues/5645, https://github.com/kivy/kivy/issues/5646

d. I can't really see many downsides?

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -1241,11 +1241,7 @@ class WindowBase(EventDispatcher):
     def add_widget(self, widget, canvas=None):
         '''Add a widget to a window'''
         if widget.parent:
-            from kivy.uix.widget import WidgetException
-            raise WidgetException(
-                'Cannot add %r to window, it already has a parent %r' %
-                (widget, widget.parent)
-            )
+            widget.parent.remove_widget(widget)
 
         widget.parent = self
         self.children.insert(0, widget)


bionoid:
> a. It fixes/avoids this issue, which does seem relatively artificial in the first place

I opened that because a user ran into the issue; obviously because they didn't understand the situation, but still something that should not have the result it did: [link](https://www.reddit.com/r/kivy/comments/7gsb95/passing_the_text_of_a_button/dqn874r/)

When I opened the issue, I imagined a simple check at startup, `if root.parent: raise WidgetException()`, ie not generally testing everything added to window. This solution looks reasonable to me, but so did inclement's original :)

akshayaurora:
"Thou shall fail when necessary".

If we silently recover from a obvious user error, the user will end up with the wrong behavior and no way to debug it.

Failing appropriately raising the exception would be the right way here.

Zen-CODE:
@akshayaurora 
But is it necessary top fail? Here, it is being added to a new parent so removal from old parents is implied. I don't see it's necessary to fail when the intention of add_widget is explictely to add it to a new parent.

But assuming we leave it. How do I fix the travis builds, which raise that error in a way I cannot replicate. I'll try the latest cython version, but otherwise, why would the travis test fail when mine passes?

bionoid:
It's not completely clear to me *why,* but [changing this test](https://github.com/kivy/kivy/blob/1.10.0/kivy/modules/inspector.py#L378-L379) to the below avoids the crash:

```python
if self.avoid_bring_to_top or self not in children:
    return
```

This may be a flaw in the inspector implementation? It is doing something recursive with window children, I haven't traced through the details. The same approach is also used in [joycursor and console modules](https://github.com/kivy/kivy/search?utf8=%E2%9C%93&q=on_window_children&type=) and may need fixing as well.

PR #5658, opened 2018-03-17
Prevent inspector from adding itself to the window

Zen-CODE:
Based on https://github.com/kivy/kivy/issues/5645, I managed to reproduce
the Inspector "already has a parent" error but it happens immediately I pres "Ctrl + e" to open
the inspector but only when a popup has been created.
To debug this, I monitored the 'on_parent' event of the inspector. It turns
our the Inspector binds to the 'children' event of the window and then tries
to add itself when whenever th 'children' Window event` fires. 

As the ModalView adds itself directly to the Window, the event fires and the 
Inspector adds itself even when it has not been activated. This is clearly
not desired behavior.

THe PR adds a check to the Inpector to detect that it does indeed want to 
re-insert itself into the children.

@bionid. Please confirm it fixes https://github.com/kivy/kivy/issues/5645 as I
could only reproduce this when creating the popup, not on closing the
inspector as your ticket indicates.  It also seems to fix the error in the travis builds, so those should pass if this is merged :-)

### DIFF
--- a/kivy/modules/inspector.py
+++ b/kivy/modules/inspector.py
@@ -375,7 +375,7 @@ class Inspector(FloatLayout):
         return ret
 
     def on_window_children(self, win, children):
-        if self.avoid_bring_to_top:
+        if self.avoid_bring_to_top or not self.activated:
             return
         self.avoid_bring_to_top = True
         win.remove_widget(self)


bionoid:
Yup, confirmed, this works!

I guess you should apply it to console and joycursor modules as well

Zen-CODE:
I looked at that those but they don't have an 'activated' property, so this
fix is not directly applicable.

Can they be switched on and off like the inspector? If not, they are
probably fine add is...


Richard

ZenCODE

bionoid:
I just noticed those modules appeared to do the same thing

> they don't have an 'activated' property, so this fix is not directly applicable.

Am I missing something here?

* https://github.com/kivy/kivy/blob/1.10.0/kivy/modules/console.py#L783
* https://github.com/kivy/kivy/blob/1.10.0/kivy/modules/joycursor.py#L83

Zen-CODE:
Indeed not. Not sure how I missed those when searching, but nice catch. Thanks.

So,  when using the console, it appears immediately when the popup does. These changes fix that.

When using the joycursor, simply deactivating it throws:
```
   File "kivy/properties.pyx", line 571, in kivy.properties.Property.dispatch
   File "/usr/local/lib/python2.7/dist-packages/kivy/modules/joycursor.py", line 127, in on_window_children
     win.add_widget(self)
   File "/usr/local/lib/python2.7/dist-packages/kivy/core/window/__init__.py", line 1247, in add_widget
     (widget, widget.parent)
 kivy.uix.widget.WidgetException: Cannot add <kivy.modules.joycursor.JoyCursor object at 0x7f5838e13ec0> to window, it already has a parent <kivy.core.window.window_pygame.WindowPygame object at 0x7f583c1c9670>
 Exception kivy.uix.widget.WidgetException: WidgetException('Cannot add <kivy.modules.joycursor.JoyCursor object at 0x7f5838e13ec0> to window, it already has a parent <kivy.core.window.window_pygame.WindowPygame object at 0x7f583c1c9670>',) in 'kivy.properties.observable_list_dispatch' ignored
[INFO   ] [JoyCursor   ] joycursor deactivated
```
These change fix than too.

Zen-CODE:
Hmm. There is an annoying issue with the travis builds where it seems to hang after the graphical unit tests. Just doing some commits to experiment and see if I can find the issue....

akshayaurora:
@Zen-CODE is this still WIP ?

Zen-CODE:
I think it's ready to be merged. It does fix the issues mentioned above.

The reason I was holding it back is that I hit an issue in the travis builds. In earlier commits, it passed all travis tests. Then suddenly, it start hanging on the last 'test_widget_popup' test in kivy/tests/test_module_inspector.py'. It works fine on my machine, but when push to github, it does not get past the last 'self.render(self.root)' call on line 296. The test then times out.

But since, I've tested with and without the changes. Both never get past that line. So these fixes do not introduce this issue at least. But it's strange that they did pass initially....

Zen-CODE:
@bionid. Is that what you meant? 
https://github.com/kivy/kivy/pull/5658/commits/11e9307cf87f9003104a235bc19d2b799862011f

Now we seem to get a a new set of errors...

bionoid:
@Zen-CODE well it needs to refer to the [xvfb process that is started here](https://github.com/kivy/kivy/blob/1.10.0/.travis.yml#L86). I'm not sure the exact syntax to stop it correctly, so I left out the rest of the arguments as an exercise for the reader ;-)

As a test (or solution..) you could try

```
kill -9 `cat /tmp/custom_xvfb_99.pid`
```

Zen-CODE:
That does not seem to help. I must admit that this is real guesswork for me. How travis does the unit test and background environment setup is unknown to me and not something I can investigate without pushing.  It also strange that this issue only happens on that particular test, when it does not appear to do anything really different? As this issue seems independent of this PR (the same happens on master), should be have investigate that as a separate issue? Should I not remove the chanage to .travis.yml and let it be merged without those changes?

Zen-CODE:
This seems like a hack route. I wish there were some way of finding why it does not exit cleanly on the travis builds but is fine on a running locally. It's probably related to the need for this PR. Something in that last 'test_widget_popup' test is different but I cannot see what.... 

The mocks around the app event loop make it difficult to debug or follow. Not sure what to do from here. It solves the original errors, so is maybe worth merging? Or how to identify to reason for the hang?

bionoid:
@Zen-CODE I have no idea about the details concerning travis, but the fix looks good & resolves the issue, from my perspective good to merge... But this pr is 19 commits for +8-3 lines, better to squash before merge

Zen-CODE:
Will do...:-)


Richard

ZenCODE

PR #5659, opened 2018-03-17
add an "auto" parameter to postproc:calibration

tshirtman:
This allows to manage touch calibration non-fullscreen windows in a
relatively automatic way, which reacts to window's size and position
change.

Assumes that the device's input area exactly covers the display area.

This still require the configuration to indicate the display/input area size,
because we don't have a portable way of detecting that, but libraries
like screeninfo could be used for that.


kivy/input/postproc/calibration.py
```diff
@@ -94,7 +106,32 @@ def process(self, events):
...
             event.ud['calibration:frame'] = frame
+
+        for event in to_remove:
+            events.remove(event)
+
         return events
+
+    def auto_calibrate(self, sx, sy, size):
+        from kivy.core.window import Window as W
+        WIDTH, HEIGHT = size
```
LukeBiddle: @tshirtman What do you think about using SDL_GetCurrentDisplayMode to fetch the screen size dynamically? 
I found another PR ([here](https://github.com/kivy/kivy/pull/4286/commits/f143512ce3de02f698d72b78d193f56b0d1abe68)) that added the functions to retrieve it, when I patched that into this PR and used those values it worked fine. Then a user only needs to add auto=true to their config, makes it a bit easier.

Issue #4483, opened 2016-07-21
Memory leak executing export_to_png fbo with stencilbuffer


Farkal:
For some reasons i need to stream my kivy app in a web page so i use the export_to_png function (it would be nice to have the same interface for PyGame and SDL2 because i can save image on the format that i want with PyGame by putting the extension on the filename parameter but with SDL2 event if i put an extension on filename it will only save on png) each 20ms to save an image (and then i send it through socket to my server and display it on webpage with js code).

The problem is that my memory consumption increase each time i use this function (i try on Fedora 24 with SDL2 and on CentOS 7.2 with PyGame) so after some experimentations in the kivy code if find that it was linked with the creation of **Fbo** object in _kivy/uix/widget.py_ line 608

`fbo = Fbo(size=self.size, with_stencilbuffer=True)`

If i remove the parameter _with_stencilbuffer=True_  everything is ok (the image is saved) and the memory leak disappear (on CentOs it completely disappear but on Fedora the memory consumption still increase a little).

So i think there is a problem with the **stencilbuffer**

riataman:
I just ran into this in the raspberry pi (egl_pi backend).

tito:
Fixed by #5664 - please reopen if it's not the case.

Issue #5581, opened 2018-01-24
Raspberry Pi OpenGL issue with screenmanager


frankgould:
### Versions
* Python:  2.7.13
* OS:  Raspian Jessie
* Kivy:  v1.10.1.dev0, git-8776f28, 20180123
* Kivy installation method:  Manual

[Trace-startup-GLerror.txt](https://github.com/kivy/kivy/files/1660682/Trace-startup-GLerror.txt)

### Description
I am migrating an operational kivy app using the same resources now on Raspberry Pi that is failing to display images in screen manager. I am trying to find a solution that other open trouble report appear to be regarding the same issues (see below). I asked on the kivy forum but was told to use KivyPie, which I did but got the same results. My configuration is NOOBS, manually installed Kivy for Raspian Jessie, on a Raspberry Pi 3. I have also imported opencv to manipulate images.

[Raspberry Pi 3] - ScreenManager and the 12 transitions #4662
Raspberry Pi 3 "Shader didnt link, check info log." #4542

In the log after starting my app, the following sections appear regarding OpenGL and Shader (see attached 'python main.py -c kivy:log_level:trace' log). It appears these are working and setup.py reports the library files are in the /opt/vc/lib/ folder.
  [INFO   ] [GL          ] Using the "OpenGL ES 2" graphics system
  [DEBUG  ] [GL          ] glShaderBinary is not available
  [INFO   ] [GL          ] Backend used <gl>
  [INFO   ] [GL          ] OpenGL version <OpenGL ES 2.0>
  [INFO   ] [GL          ] OpenGL vendor <Broadcom>
  [INFO   ] [GL          ] OpenGL renderer <VideoCore IV HW>
  [INFO   ] [GL          ] OpenGL parsed version: 2, 0
  [INFO   ] [GL          ] Shading version <OpenGL ES GLSL ES 1.00>
  [INFO   ] [GL          ] Texture max size <2048>
  [INFO   ] [GL          ] Texture max units <8>
  [DEBUG  ] [Shader      ] Fragment compiled successfully
  [DEBUG  ] [Shader      ] Vertex compiled successfully

After my app starts displaying images through screenmanager, the following error appears in the same log (this error means invalid operation):
  glGetError 0x502

After 7 of these invalid operation error messages, the following error appears (glGetError 0x505 means out-of-memory):
   Exception: FBO Initialization failed: Incomplete attachment (36054)
[...]

PR #5664, opened 2018-03-21
Fixes renderbuffer leaking when creating Fbo

tito:
If a FBO was created with stencil buffer but no depth buffer (like what ShaderTransition does today), it doesn't release the Renderbuffer created for the stencil.

The same renderbuffer is used for both stencil and depth, but only the depth id was checked for the release. This PR fixes the leak.

Ref #5581


bionoid:
Nice catch, maybe this fixes #4483?

PR #5679, opened 2018-03-28
only run shaders if the widgets actually changed

### LINKED ISSUE #3605: EffectWidget performance issue
With this code I get a steady cpu usage of 7-10%, commenting out https://github.com/kivy/kivy/blob/34bf7ef6778d06b7cee34be384b0334fff27f7b7/kivy/uix/effectwidget.py#L670 reduces it to 0-2%.

Dr. @inclement noted that `the time thing can just be removed (or more precisely, made cleverer), and the rest bound to its size`.

```python
from kivy.app import App
from kivy.lang import Builder

kv = """
BoxLayout:
    EffectWidget:
    EffectWidget:
    EffectWidget:
    EffectWidget:
    EffectWidget:
    EffectWidget:
    EffectWidget:
    EffectWidget:
    EffectWidget:
    EffectWidget:
"""


class TestApp(App):
    def build(self):

if __name__ == '__main__':
    TestApp().run()

```

jabdoa2:
fix #3605

### DIFF
--- a/kivy/uix/effectwidget.py
+++ b/kivy/uix/effectwidget.py
@@ -119,6 +119,7 @@ You can also set the glsl by automatically loading the string from a
 file, simply set the :attr:`EffectBase.source` property of an effect.
 
 '''
+from kivy.graphics.instructions import Callback
 
 from kivy.clock import Clock
 from kivy.uix.relativelayout import RelativeLayout
@@ -653,6 +654,7 @@ class EffectWidget(RelativeLayout):
 
         self.canvas = RenderContext(use_parent_projection=True,
                                     use_parent_modelview=True)
+        self._callbacks = {}
 
         with self.canvas:
             self.fbo = Fbo(size=self.size)
@@ -669,8 +671,6 @@ class EffectWidget(RelativeLayout):
 
         super(EffectWidget, self).__init__(**kwargs)
 
-        Clock.schedule_interval(self._update_glsl, 0)
-
         fbind = self.fbind
         fbo_setup = self.refresh_fbo_setup
         fbind('size', fbo_setup)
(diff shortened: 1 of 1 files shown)


jabdoa2:
Looks like this does not work with pixelate filter. Maybe not the best solution then?

jabdoa2:
Is there interest to get this merged? Pixelate should be easy to fix (it depends on the previous behavior).

kivy/uix/effectwidget.py
```diff
@@ -683,17 +683,11 @@ def __init__(self, **kwargs):
...
-        '''(internal) Passes new time and resolution uniform
-        variables to the shader.
-        '''
-        time = Clock.get_boottime()
-        resolution = [float(size) for size in self.size]
-        self.canvas['time'] = time
-        self.canvas['resolution'] = resolution
+    def _propagate_updates(self, *largs):
+        """Propagate updates from widgets."""
+        del largs
```
inclement: Why `del largs`?

jabdoa2: "Callback" requires those largs but we do not use them. This is to prevent any linter from reporting unused parameters (not sure if those are active for kivy currents).

inclement: I see, could you add an `# unused` comment to clarify what's going on? It would also be good to rename `largs` to `_`, which will actually keep pylint happy on its own since that's a convention it expects for unimportant arguments.

tshirtman:
I'm going to test it on various things i did with EffectWidget, but from looking at the code, the main effect is that we don't have a time value by default anymore, as it would force a refresh every frame, if we want one, we have to add one. This makes sense as unnecessarily redrawing a large widget through multiple FBOs for no other reason that time changed can be wasteful in a lot of cases, and adding a time uniform when needed shouldn't be too hard. Still, if only one effect in the chain depends on time, then all the ones after it will have to update, is that change going to propagate automatically, or do we need something extra?

PR #5680, opened 2018-03-28
fix osc/tuio for python3

tshirtman:


tito:
ok to merge?

tshirtman:
i'm actually planning on completly removing this old and crappy vendored module and switching to oscpy if there is no objection with that (https://pypi.python.org/pypi/oscpy), i'm still testing/improving it (i have to fix windows support that i accidentally broke in last version by adding unix socket support). Since less people are probably using tuio these days, it could probably be made an optional dependency (like enchant). OSCPy has nearly full test coverage on python2 and 3, and its performances are way superior to osc.py/oscapi, switching kivy to it is trivial (i did it in a local branch).

tshirtman:
closing in favor of #5716

PR #5693, opened 2018-04-11
PiCamera-based camera provider for Raspberry Pi

whitelynx:
Added a new camera provider based on [PiCamera](https://picamera.readthedocs.io/en/release-1.13/index.html), for use with CSI-based cameras on the Raspberry Pi. (not for USB cameras)

### DIFF
--- a/kivy/__init__.py
+++ b/kivy/__init__.py
@@ -221,7 +221,7 @@ kivy_options = {
         'avplayer'),
     'image': ('tex', 'imageio', 'dds', 'sdl2', 'pygame', 'pil', 'ffpy', 'gif'),
     'camera': ('opencv', 'gi', 'avfoundation',
-               'android'),
+               'android', 'picamera'),
     'spelling': ('enchant', 'osxappkit', ),
     'clipboard': (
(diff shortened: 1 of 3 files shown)


tito:
Nothing to say about the PR, thanks for all!

tito:
Can you also check the documentation part about the requirements needed for the camera on RPI ? (Numpy + picamera)

bionoid:
Seems like numpy is only used to chop off some "bonus" pixels..? Assuming it's not a requirement of PiCamera in the first place, you could maybe avoid it with a "master" texture (at the full/rounded size), and `master_texture.get_region(0, 0, cropped_w, cropped_h)` for drawing or a Cython snippet?

Related: PR #5210 - seems like it does similar things without external dependencies

whitelynx:
I do like the fact that #5210 doesn't require `numpy`... I'd be fine with that one replacing mine, if someone more comfortable with the C/Cython side of things could review it.

In the meantime, I'll look at using your suggestion for cutting out the `numpy` dep in this one :)

tito:
meh :/ i didn't knew about #5210, why this was wasn't merged :(

PR #5696, opened 2018-04-12
add layout.add_widgets()

ire4ever1190:
added layout.add_widgets() so you can add many widgets to the layout at once
```
layout = BoxLayout(orientation='vertical')
lbl0 = Label(text = "Foo")
lbl1 = Label(text = "Bar")
layout.add_widgets(lbl0, lbl1)
```


Zen-CODE:
This has been brought up before and declined before as this can also be easily and concisely achieved by
```
[layout.add_widget(w) for w in [lbl0, lbl1]]
```
Although it seems like a nice and easy addition, for consistency, similar methods would have to be added to all the various widgets and layouts. That's a lot of additional code for and room for error for changes that do not really add anything.

But thanks for the PR anyway :-)

PR #5703, opened 2018-04-20
Fixed format string mistake in Error Message

Bounder:
While debugging a issue with creating a Kivy exe using PyInstaller I found this bug where it adds the 's' to file names in the error message. Just a small bug, but annoying.


Zen-CODE:
Thanks :-)

Issue #2636, opened 2014-11-02
Callback instruction not available from within KV language


SenshiSentou:
For a specific Widget I need to change the blending function multiple times within the same pass. The docs mention the `Callback` object for this, and examples I found use it as such:

```
def __init__(...):
    with self.canvas.before:
        Callback(self.set_blend_func)

def set_blend_func(*args):
    glBlendFunc(gl.GL_SRC_ALPHA, gl.GL_DST_COLOR)`
```

However, this doesn't provide any control over when to switch within the canvas.before block.

Is there any way we can use the Callback instruction in our .kv files, like so for instance?

```
<MyWidget>:
    Rectangle:
        [...]
    Callback:
        call: self.set_blend_func()
    Rectangle:
        [...]
    Callback:
        call: self.set_other_blend_func()
    Rectangle:
        [...]
```

Thanks,
Patrick
##

matham:
https://github.com/kivy/kivy/pull/5878

PR #5711, opened 2018-04-25
Make 'kivy.graphics.instruction.Callback' available from within KvLanguage

gottadiveintopython:
Currently `kivy.graphics.instruction.Callback` is not available in KvLanguage. This PR allows you to do like this:

```python3
from kivy.lang import Builder


root = Builder.load_string(r'''
FloatLayout:
    canvas:
        Callback:
            func: lambda __: print('test')
''')


runTouchApp(root)
```

Since I'm not familiar with cython, the code might be not efficient.

#2636


tshirtman:
hey, that does look neat, though it would be better if it could be used like events in widgets, (i.e, avoid the lambda, just put the code directly).

gottadiveintopython:
That sounds good. I'll try that but yeah kivy/lang/*.py looks complex. It'll take long time.

gottadiveintopython:
I re-created this PR. #5876
sorry I'm still bad at git command.

PR #5715, opened 2018-04-28
Let Layout.add_widget use the ``canvas`` argument

LogicalDash:
I had occasion to turn a Widget subclass of mine into a Layout, which caused TypeError as Layout's add_widget never got the ``canvas`` argument that base Widget has. Layout.add_widget isn't that different from Widget.add_widget though, so I just added it. Might help somebody?


Zen-CODE:
Think it's also more expected behaviour. There should be real reasons for omitting arguments to inherited method. If these exist, they should documented or explained in the code?

Thanks :+1:

PR #5716, opened 2018-04-28
replace vendored lib/OSC and lib/oscAPI with oscpy

tshirtman:
This fixes issues with tuio on python3, and replace an old hacky vendored lib, with a more modern alternative, that is tested on both python versions, and has good coverage, as well as better performances.

The impact on kivy tuio provider is minimal.


tito:
With Python 3.6, i needed to change 
```
TuioMotionEventProvider.register('/tuio/2Dcur', Tuio2dCurMotionEvent)
TuioMotionEventProvider.register('/tuio/2Dobj', Tuio2dObjMotionEvent)
TuioMotionEventProvider.register('/tuio/2Dblb', Tuio2dBlbMotionEvent)
```
with
```
TuioMotionEventProvider.register(b'/tuio/2Dcur', Tuio2dCurMotionEvent)
TuioMotionEventProvider.register(b'/tuio/2Dobj', Tuio2dObjMotionEvent)
TuioMotionEventProvider.register(b'/tuio/2Dblb', Tuio2dBlbMotionEvent)
```
to make tuio accepting the messages

setup.py
```diff
@@ -1103,7 +1103,9 @@ def get_extensions_from_sources(sources):
             'Topic :: Software Development :: User Interfaces'],
         dependency_links=[
             'https://github.com/kivy-garden/garden/archive/master.zip'],
-        install_requires=['Kivy-Garden>=0.1.4', 'docutils', 'pygments'],
+        install_requires=[
+            'Kivy-Garden>=0.1.4', 'docutils', 'pygments', 'oscpy'
```
tito: I wonder if that could be optionnal now? I know you use it in your daily job, just wondering.

tito:
Otherwise +1 for this PR. It saved my week.

setup.py
tshirtman: Sure, i guess it would be more convenient for platforms like android/ios where it's not useful.

PR #5727, opened 2018-05-03
input: probesysfs: remove getconf dependency

### LINKED ISSUE #5598: Probesysfs provider requires getconf
### Versions

* Python: 3.6.4
* OS: Linux
* Kivy: 1.10.0
* Kivy installation method: setuptools

### Description

Kivy's probesysfs provider requires getconf, provided by glibc, to get the platform's LONG_BIT value.

This dependency precludes the use of other C libraries, such as musl, as well as platforms that choose not to install getconf.

jakogut:
Remove the dependency on getconf to get LONG_BIT value, in order to
improve portability to Linux systems without Glibc and/or getconf.

My testing with this patch is, so far, limited to the existing unit tests, and using a mouse with the upstream examples. I have not tried a touchscreen, or different architectures with this patch yet, and I'm marking it WIP until I can get more testing data with it.

This closes #5598

### DIFF
--- a/kivy/input/providers/probesysfs.py
+++ b/kivy/input/providers/probesysfs.py
@@ -47,6 +47,7 @@ if 'KIVY_DOC' in os.environ:
     ProbeSysfsHardwareProbe = None
 
 else:
+    import ctypes
     from re import match, IGNORECASE
     from glob import glob
     from subprocess import Popen, PIPE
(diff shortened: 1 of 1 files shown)


tshirtman:
I happen to have a touchscreen on my laptop, that works with probesysfs, and this patch works here, and i don't see why it wouldn't. Unless you have a plan to do more test, i'm ok to merge this.

jakogut:
Thanks for testing it. I've removed the WIP tag, I just didn't want this to be merged before testing it with a touchscreen.

tshirtman:
Thank you!

PR #5749, opened 2018-05-13
Add decorator for Clock.create_trigger()

ismailof:
This PR adds the decorator `@triggered` to `Clock`. It schedules the given function through `Clock.create_trigger()`, the same way that `@mainthread` uses `Clock.schedule_once()`, but with the ability to set the `timeout` and `interval` parameters.

It simplifies the use of triggered events on regular methods, without complicating the code using dt, lambda, partial, etc. The decorated method can be called several times within the timeout, even with different arguments, but only the last one will be triggered.

It can be very helpful when in a complex system, you have several events or functions calling an expensive method with an updating parameter, but only the last one is needed.

### DIFF
--- a/kivy/clock.py
+++ b/kivy/clock.py
@@ -803,6 +803,52 @@ def mainthread(func):
     return delayed_func
 
 
+def triggered(timeout, interval=False):
+    '''Decorator that will trigger the call of the function at the specified
+    timeout, through the method :meth:`CyClockBase.create_trigger`. A new call
+    to the function before the timeout will interrupt the previous call and
+    restart the timeout, so only one call is executed during this timeout.
+
+    It can be helpful when an expensive funcion (i.e. call to a server) can be
+    triggered by different methods. Setting a proper timeout will delay the
+    calling and only one of them wil be triggered.
+
+        callback(id=1)
+        callback(id=2)
+
+        @triggered(timeout, interval=False)
+        def callback(self, id):
+            print('The callback has been called with id=%d' % id)
+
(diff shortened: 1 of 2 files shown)


bionoid:
Quoting from the docs in PR:

> A new call to the function before the timeout will interrupt the previous call and restart the timeout, so only one call is executed during this timeout.

This seems inaccurate to me - calls to an already-scheduled trigger does **not** restart the timer, it will execute when the timeout has elapsed from the initial call. It doesn't seem like there is any code in the PR that does this.

It would be a nice feature to add, I've needed it from time to time, for example to keep a menu open for N seconds after the most recent interaction. It's trivial to implement of course, but it would be nice to have something like `Clock.create_trigger(retrig=True)` or a `retriggered` decorator

ismailof:
@tshirtman Sure! I'll commit the default timeout=0 right away to have it done. I will need some time to come with a proper test since I do not know well the test engine, but I'm on it.

About the @bionoid proposition, I have to say that I though that the timeout was indeed reset on retriggers, but I think it is a matter of another PR and discussion between main devs. If such behaviour is added to the `create_trigger()` functionality it will be trivial to add it to the decorator too.

ismailof:
I found no test cases for `create_trigger()` itself, so I added some along with the decorator function. 
I couldn't figure it out how to deal with tests involving timeouts, such as triggering some callback twice inside or outside the timeout, but I hope it is enough for now.

tshirtman:
Thank you!

PR #5753, opened 2018-05-16
audio_avplayer.py

dasgauranga2:
it is just a small change

### DIFF
--- a/kivy/core/audio/audio_avplayer.py
+++ b/kivy/core/audio/audio_avplayer.py
@@ -21,10 +21,11 @@ class SoundAvplayer(Sound):
         # taken from https://goo.gl/015kvU
         return ("aac", "adts", "aif", "aiff", "aifc", "caf", "mp3", "mp4",
                 "m4a", "snd", "au", "sd2", "wav")
+                
 
-    def __init__(self, **kwargs):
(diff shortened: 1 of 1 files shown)


tshirtman:
Can you explain why this change is needed? what does it fixes? this looks like an API break and would make the API works different based on the platform, that would be the wrong thing to do.

KeyWeeUsr:
`SoundAvPlayer` inherits from `Sound` which [inherits from `EventDispatcher`](https://github.com/kivy/kivy/blob/337f371b49e2d7938cfee92ca35a1a4811b4caec/kivy/core/audio/__init__.py#L92).

`__cinit__` I guess shouldn't be even accessible from Python side and `__init__` in there [accepts only `**kwargs`](https://github.com/kivy/kivy/blob/337f371b49e2d7938cfee92ca35a1a4811b4caec/kivy/_event.pyx#L233), therefore I believe this change makes no sense and the author probably just e.g. tried to inherit from `SoundAvPlayer` and threw some arguments at that instead of keyword arguments, thus this small change that would break the cascade to `EventDispatcher`.

No usage of raw args instead of kwargs in other providers + the author hasn't responded yet, closing.

PR #5760, opened 2018-05-20
Add build requirements file according to PEP 518

dolang:
**Note:** Should probably not be merged before v1.10.1 so we can see what effect this has.
It shouldn't cause any problems in any case, but you never know.

[PEP 518](https://www.python.org/dev/peps/pep-0518/) specifies a way to add build requirements to a project with a new file `pyproject.toml`. Pip from v10 on knows how to handle this.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["Cython==0.28.2", "setuptools", "wheel"]


tshirtman:
NOTE: the failing travis build is a false positive.

bionoid:
Is `Cython==0.28.2` the best option here? I don't know how this works exactly, but I guess specifying setuptools means that setup.py will be executed by pip anyway? If so, it might be better to let setup.py check for a compatible version. Maybe leaving it out altogether, or specifying `Cython` or `Cython>=0.27.3` or something else is a better tradeoff?

It just seems a bit off to require 0.28.2 specifically when Kivy builds with 0.23+ except a few incompatible versions. People may want to stick with 0.25.x or whatever for their projects

bionoid:
> that might cause fewer surprises if the Cython guys break something in their next version, whenever that may be.

Another option is  `Cython<=0.28.2`, that'll let setup.py catch incompatible 0.27.x or <0.23. But Cython is moving quite fast; 0.26, 0.27 and 0.28 series were all released after Kivy 1.10. I'm not convinced blocking newer versions is the best option either. Using just `Cython` as a requirement seems most sensible to me right now -- I think the [unsupported version warning from setup.py](https://github.com/kivy/kivy/blob/1.10.0/setup.py#L183-L196) is good enough to help users understand when a newer Cython is the cause of a failing build.

dolang:
Almost forgot, but I wanted to test this before making any changes and before it gets merged.

It's actually no problem the way it is now. I've preinstalled `Cython==0.27` on a new clone, then added the `pyproject.toml`. Looks like pip is happy to accept that without doing any modifications.  There's a short and opaque
```
Installing build dependencies ... done
```
then it continues as usual. I then reran it with `-v` and it looks like it grabs the specified version and uses that for the build only. I will have to look into if it's possible to override/ignore that completely.

**TL;DR:**
* Doesn't negatively impact or change an existing environment
* Specifying another version/range is still up for discussion

dolang:
Looks like it's possible to disable the build dependency with `pip install --no-build-isolation ...` although that's a bit cumbersome:
```
Disable isolation when building a modern distribution. Build
dependencies specified by PEP 518 must be already installed
if this option is used.
```
https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-no-build-isolation

The `-b` option ([docs](https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-b)) might also play into this, but I haven't tested that.

matham:
I'll close for now as the cython issue needs to be resolved. It can be opened once there's a fix for that.

PR #5765, opened 2018-05-22
Update documentation for Clock.triggered decorator

bionoid:
I think this is a more accurate description of trigger behavior


akshayaurora:
@bionoid thanks 👍

PR #5779, opened 2018-05-31
Add 'on_pre_open' and 'on_pre_dismiss' events to ModalView

pythonic64:
I find `on_pre_open` useful for a case when keyboard event listener must be disabled before popup animation start.


pythonic64:
Closing in favor of modified pull request where `on_pre_open` is dispatched immediately after window is found and set.

PR #5781, opened 2018-06-01
Add 'on_pre_open' and 'on_pre_dismiss' events to ModalView

pythonic64:
I find `on_pre_open` useful for a case when keyboard event listener must be disabled before popup animation start.

### DIFF
--- a/kivy/uix/modalview.py
+++ b/kivy/uix/modalview.py
@@ -85,11 +85,20 @@ class ModalView(AnchorLayout):
     '''ModalView class. See module documentation for more information.
 
     :Events:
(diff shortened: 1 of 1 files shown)


tshirtman:
Thanks!

pythonic64:
You're welcome, but before merge `versionchanged` should have been updated to `1.10.0`.

tshirtman:
i assume you mean `1.11.0`, if so, you are right, i'll fix that.

pythonic64:
Yes, I meant `1.11.0`, the current dev version.

PR #5785, opened 2018-06-03
[WIP] Remove SDL image dependence from SDL2 Window

dolang:
**WIP**, or rather, proof of concept. Needs review.

SDL2 currently depends on the SDL image project, even when PIL is used as the image provider. There are 3 API calls in `_window_sdl2.py` which make this a necessity.

This is to remove that dependency by replacing those calls with the `core.image` API.

The `surface_from_core_image()` is adapted from `_img_sdl2.pyx`. I've been able to test both the icon code and the screenshot code, but I'm unable to enable a shaped window (with or without my modifications). Also, I'm not sure if the `SDL_FreeSurface()` is required there.

Second commit fixes a minor API inconsistency.

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -3,6 +3,7 @@ include "../../include/config.pxi"
 
 from libc.string cimport memcpy
 from os import environ
+from kivy.core import image
 from kivy.config import Config
 from kivy.logger import Logger
 from kivy import platform
@@ -406,13 +407,19 @@ cdef class _WindowSDL2Storage:
         elif mode == 'colorkey':
             sdl_window_mode.mode = ShapeModeColorKey
 
-        sdl_shape = IMG_Load(<bytes>shape.encode('utf-8'))
-        if not sdl_shape:
+        try:
+            core_image = image.Image.load(shape, nocache=True)
+        except:
+            sdl_shape = NULL
             Logger.error(
                 'Window: Shape image "%s" could not be loaded!' % shape
             )
+        else:
+            sdl_shape = surface_from_core_image(core_image)
 
         result = SDL_SetWindowShape(self.win, sdl_shape, &sdl_window_mode)
+        if sdl_shape:
+            SDL_FreeSurface(sdl_shape)
 
         # SDL prevents the change with wrong input values and gives back useful
         # return values, so we pass the values to the user instead of killing
@@ -437,8 +444,11 @@ cdef class _WindowSDL2Storage:
     # twb end
 
     def set_window_icon(self, filename):
-        icon = IMG_Load(<bytes>filename.encode('utf-8'))
-        SDL_SetWindowIcon(self.win, icon)
+        core_image = image.Image.load(filename, nocache=True)
+        print(core_image.nocache, core_image.size)
+        cdef SDL_Surface *icon_surface = surface_from_core_image(core_image)
+        SDL_SetWindowIcon(self.win, icon_surface)
+        SDL_FreeSurface(icon_surface)
 
     def teardown_window(self):
         if self.ctx != NULL:
@@ -634,17 +644,9 @@ cdef class _WindowSDL2Storage:
         SDL_GL_SwapWindow(self.win)
 
     def save_bytes_in_png(self, filename, data, int width, int height):
-        cdef SDL_Surface *surface = SDL_CreateRGBSurfaceFrom(
-            <char *>data, width, height, 24, width * 3,
-            0x0000ff, 0x00ff00, 0xff0000, 0
-        )
-        cdef bytes bytes_filename = <bytes>filename.encode('utf-8')
-        cdef char *real_filename = <char *>bytes_filename
-
-        cdef SDL_Surface *flipped_surface = flipVert(surface)
-        IMG_SavePNG(flipped_surface, real_filename)
-        SDL_FreeSurface(surface)
-        SDL_FreeSurface(flipped_surface)
+        loader = next(loader for loader in image.ImageLoader.loaders
+                      if loader.can_save() and 'png' in loader.extensions())
+        loader.save(filename, width, height, 'RGB', data, flipped=True)
 
     def grab_mouse(self, grab):
         SDL_SetWindowGrab(self.win, SDL_TRUE if grab else SDL_FALSE)
@@ -686,3 +688,14 @@ cdef SDL_Surface* flipVert(SDL_Surface* sfc):
         memcpy(&rpixels[pos], &pixels[(pxlength - pos) - pitch], pitch)
 
     return result
+
+cdef SDL_Surface* surface_from_core_image(core_image):
+    # create an SDL_Surface from a core image's texture pixels. Use
+    # SDL_FreeSurface(surface) to deallocate
+    width, height = core_image.size
+    cdef int pitch = width * 4
+    cdef char *c_pixels = core_image.image.texture.pixels
+    cdef SDL_Surface *icon_surface
+    icon_surface = SDL_CreateRGBSurfaceFrom(c_pixels, width, height, 32,
+        pitch, 0x00000000ff, 0x0000ff00, 0x00ff0000, 0xff000000)
+    return icon_surface
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -3,8 +3,8 @@ include "../../include/config.pxi"
 
 from libc.string cimport memcpy
 from os import environ
-from kivy.core import image
 from kivy.config import Config
+from kivy.core.image import Image, ImageLoader
 from kivy.logger import Logger
 from kivy import platform
 from kivy.graphics.cgl import cgl_get_backend_name
@@ -408,18 +408,21 @@ cdef class _WindowSDL2Storage:
             sdl_window_mode.mode = ShapeModeColorKey
 
         try:
-            core_image = image.Image.load(shape, nocache=True)
-        except:
-            sdl_shape = NULL
+            image = Image.load(shape, nocache=True)
+        except Exception:
             Logger.error(
-                'Window: Shape image "%s" could not be loaded!' % shape
+                'Window: Shape image "%s" could not be loaded!' % (shape,)
             )
-        else:
-            sdl_shape = surface_from_core_image(core_image)
+            return
+
+        sdl_shape = surface_from_core_image(image)
+        if sdl_shape == NULL:
+            Logger.error(
+                'Window: Shape image "%s" does not have a texture' % (shape,)
+            )
+            return
 
         result = SDL_SetWindowShape(self.win, sdl_shape, &sdl_window_mode)
-        if sdl_shape:
-            SDL_FreeSurface(sdl_shape)
 
         # SDL prevents the change with wrong input values and gives back useful
         # return values, so we pass the values to the user instead of killing
@@ -444,9 +447,21 @@ cdef class _WindowSDL2Storage:
     # twb end
 
     def set_window_icon(self, filename):
-        core_image = image.Image.load(filename, nocache=True)
-        print(core_image.nocache, core_image.size)
-        cdef SDL_Surface *icon_surface = surface_from_core_image(core_image)
+        try:
+            image = Image.load(filename, nocache=True)
+        except Exception:
+            Logger.warn(
+                'Window: Icon "%s" could not be loaded' % (filename,)
+            )
+            return
+
+        cdef SDL_Surface *icon_surface = surface_from_core_image(image)
+        if icon_surface == NULL:
+            Logger.warn(
+                'Window: Icon "%s" does not have a texture' % (filename,)
+                )
+            return
+
         SDL_SetWindowIcon(self.win, icon_surface)
         SDL_FreeSurface(icon_surface)
 
@@ -644,8 +659,12 @@ cdef class _WindowSDL2Storage:
         SDL_GL_SwapWindow(self.win)
 
     def save_bytes_in_png(self, filename, data, int width, int height):
-        loader = next(loader for loader in image.ImageLoader.loaders
-                      if loader.can_save() and 'png' in loader.extensions())
+        try:
+            loader = next(l for l in ImageLoader.loaders
+                          if l.can_save() and 'png' in l.extensions())
+        except StopIteration:
+            Logger.error('Window: No image provider can save PNG files')
+            return
         loader.save(filename, width, height, 'RGB', data, flipped=True)
 
     def grab_mouse(self, grab):
@@ -690,12 +709,34 @@ cdef SDL_Surface* flipVert(SDL_Surface* sfc):
     return result
 
 cdef SDL_Surface* surface_from_core_image(core_image):
-    # create an SDL_Surface from a core image's texture pixels. Use
-    # SDL_FreeSurface(surface) to deallocate
+    # Create an SDL_Surface from a core image's texture pixels. Use
+    # SDL_FreeSurface(surface) to deallocate.
+    #
+    # Returns NULL if the core image has no texture pixels, or if
+    # there's a problem with the creation of an SDL Surface.
+    cdef int width, height
     width, height = core_image.size
     cdef int pitch = width * 4
-    cdef char *c_pixels = core_image.image.texture.pixels
-    cdef SDL_Surface *icon_surface
-    icon_surface = SDL_CreateRGBSurfaceFrom(c_pixels, width, height, 32,
-        pitch, 0x00000000ff, 0x0000ff00, 0x00ff0000, 0xff000000)
-    return icon_surface
+    cdef bytes pixels
+    cdef char* sdl_error
+    try:
+        pixels = core_image.image.texture.pixels
+    except AttributeError:
+        return NULL
+
+    if pixels == None:
+        return NULL
+
+    cdef Uint8 *c_pixels = pixels
+    cdef SDL_Surface *surface
+    surface = SDL_CreateRGBSurface(0, width, height, 32,
+        0x00000000ff, 0x0000ff00, 0x00ff0000, 0xff000000)
+    if surface == NULL:
+        sdl_error = SDL_GetError()
+        Logger.warn(
+            "Window: Can't create a surface from a core image: %s" % sdl_error
+        )
+        return NULL
+
+    memcpy(surface.pixels, c_pixels, pitch * height)
+    return surface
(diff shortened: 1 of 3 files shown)


kivy/core/window/_window_sdl2.pyx
```diff
@@ -437,8 +444,11 @@ cdef class _WindowSDL2Storage:
     # twb end
 
     def set_window_icon(self, filename):
-        icon = IMG_Load(<bytes>filename.encode('utf-8'))
-        SDL_SetWindowIcon(self.win, icon)
+        core_image = image.Image.load(filename, nocache=True)
+        print(core_image.nocache, core_image.size)
```
dolang: Note to myself: remove when this gets approved.

kivy/core/window/_window_sdl2.pyx
```diff
@@ -406,13 +407,19 @@ cdef class _WindowSDL2Storage:
         elif mode == 'colorkey':
             sdl_window_mode.mode = ShapeModeColorKey
 
-        sdl_shape = IMG_Load(<bytes>shape.encode('utf-8'))
-        if not sdl_shape:
+        try:
+            core_image = image.Image.load(shape, nocache=True)
+        except:
```
KeyWeeUsr: Please don't. There are way better ways to handle errors and if you necessarily need `try-except`, then handle the error explicitly by using it in the `except` clause.

kivy/core/window/_window_sdl2.pyx
```diff
@@ -686,3 +688,14 @@ cdef SDL_Surface* flipVert(SDL_Surface* sfc):
         memcpy(&rpixels[pos], &pixels[(pxlength - pos) - pitch], pitch)
 
     return result
+
+cdef SDL_Surface* surface_from_core_image(core_image):
+    # create an SDL_Surface from a core image's texture pixels. Use
+    # SDL_FreeSurface(surface) to deallocate
+    width, height = core_image.size
```
KeyWeeUsr: I think these could be `cdef int`s since it's already a `cdef`ed function anyway.

kivy/core/window/_window_sdl2.pyx
dolang: It's a proof of concept, I didn't bother structuring it properly, yet. I've only tried to recreate the previous logic flow so far.

So yes, agree completely.

dolang:
Thanks for having a look. I did try on Windows and Linux without success. I'll have a look at the linked PR and try again, I guess.

But if you think there's nothing wrong with the general approach to this, I'll probably have a refined version until next weekend.

dolang:
This version is much better, fixed a critical bug among other things.

However, there are still a few open questions:
* It's close to current logic, but that has a memory leak. So this version has one, too.
* Not sure anymore if the images should be loaded with `nocache=True` here. Original reason for that was:
  * to not keep potentially big images (shaped window) in RAM
  * the icon would not get loaded several times anyway; seems like that can happen, however
* I am not entirely sure about the `ImageLoader` logic, so I wrote it in a way that is not assuming a loader to have a valid texture with pixel data.
* 32bit RGBA is assumed in `surface_from_core_image`, so this won't work with e.g. just RGB images.
* Cython unit tests?

kivy/core/window/_window_sdl2.pyx
```diff
@@ -686,3 +707,36 @@ cdef SDL_Surface* flipVert(SDL_Surface* sfc):
...
+        return NULL
+
+    cdef Uint8 *c_pixels = pixels
+    cdef SDL_Surface *surface
+    surface = SDL_CreateRGBSurface(0, width, height, 32,
+        0x00000000ff, 0x0000ff00, 0x00ff0000, 0xff000000)
+    if surface == NULL:
+        sdl_error = SDL_GetError()
+        Logger.warn(
+            "Window: Can't create a surface from a core image: %s" % sdl_error
```
dolang: Not entirely sure how `char *` interact with Python strings, yet.

dolang:
Tested on Linux Mint 19 (ubuntu bionic): 

Works about as well as current implementation (read: shapedwindow is a bit flaky), but doesn't crash. Works definitely better on Windows (where i tested this first). However, it only works if it's set to `KIVY_IMAGE=pil`. If not (i.e. SDL2 image provider), I get a core dump when taking a screenshot. Not yet sure why that happens.

In any case, this needs to be **thoroughly tested** on all target systems before it can be merged.

kivy/core/window/_window_sdl2.pyx
```diff
@@ -686,3 +707,36 @@ cdef SDL_Surface* flipVert(SDL_Surface* sfc):
...
+        pixels = core_image.image.texture.pixels
+    except AttributeError:
+        return NULL
+
+    if pixels == None:
+        return NULL
+
+    cdef Uint8 *c_pixels = pixels
+    cdef SDL_Surface *surface
+    surface = SDL_CreateRGBSurface(0, width, height, 32,
```
bionoid: Seems like you could use `SDL_CreateRGBSurfaceFrom` here instead of the additional memcpy

kivy/core/window/_window_sdl2.pyx
```diff
@@ -686,3 +707,36 @@ cdef SDL_Surface* flipVert(SDL_Surface* sfc):
...
+        0x00000000ff, 0x0000ff00, 0x00ff0000, 0xff000000)
+    if surface == NULL:
+        sdl_error = SDL_GetError()
+        Logger.warn(
+            "Window: Can't create a surface from a core image: %s" % sdl_error
+        )
+        return NULL
+
+    memcpy(surface.pixels, c_pixels, pitch * height)
+    return surface
```
bionoid: This looks like it leaks memory? Is there a problem using try/finally in this situation?

```python
try:
    return surface
finally:
    SDL_FreeSurface(surface)
```

kivy/core/window/_window_sdl2.pyx
```diff
@@ -686,3 +707,36 @@ cdef SDL_Surface* flipVert(SDL_Surface* sfc):
...
+    width, height = core_image.size
+    cdef int pitch = width * 4
+    cdef bytes pixels
+    cdef char* sdl_error
+    try:
+        pixels = core_image.image.texture.pixels
+    except AttributeError:
+        return NULL
+
+    if pixels == None:
```
bionoid: Maybe use `if not pixels:` instead of testing for None explicitly? Just a safety measure, in case for whatever reason you get an empty byte array or whatever from the coreimage.

kivy/core/window/_window_sdl2.pyx
```diff
@@ -686,3 +707,36 @@ cdef SDL_Surface* flipVert(SDL_Surface* sfc):
...
+    # Create an SDL_Surface from a core image's texture pixels. Use
+    # SDL_FreeSurface(surface) to deallocate.
+    #
+    # Returns NULL if the core image has no texture pixels, or if
+    # there's a problem with the creation of an SDL Surface.
+    cdef int width, height
+    width, height = core_image.size
+    cdef int pitch = width * 4
+    cdef bytes pixels
+    cdef char* sdl_error
```
bionoid: Probably avoid the cdef here,

```python
the_error = <bytes>SDL_GetError()
```

The resulting bytes object is garbage collected by the interpreter, as opposed to the char pointer

dolang:
@bionoid Sorry, it's already been so long since I've worked on it that I'm kinda fuzzy on the details now, i.e. the reason why I did things the way I did, there.

I still haven't heard back from @KeyWeeUsr, so in the meantime this is on hold, i.e. I'm not going to work on it.

I did to put it on top of your improved SDL2 image PR, but I still got a crash. Also, if you actually want to try it yourself, you should try using PIL as the image provider. Otherwise, I'd be happy to chat on IRC as well.

KeyWeeUsr:
Ah, I misclicked, sorryyyy :cry: going to test it now

kivy/core/window/_window_sdl2.pyx
AndreMiras: My 2¢ to be honest I prefer explicit rather than implicit, I rather not covering the case and let it crash to later (unit test and) fix it, than covering some cases "just in case" even though they may never happens. And this exactly for the reason you evoked "whatever reason you get an empty byte array or whatever". I think if we are not sure of what we get and if we don't make it explicit we'll pay for it later. But I believe it's also a matter of taste and point of view..
Same thing with catching `Exception`, I'm not so much of a big fan.
Other matter of taste: I would use a more Pythonic way `if pixels is None`.

kivy/core/window/_window_sdl2.pyx
dolang: This one [I've actually looked up](https://wiki.libsdl.org/SDL_GetError) sometime this month. There's no need to memory manage the `char*` returned from `SDL_GetError()`.

However, it isn't great and `<bytes>` isn't great either because I'm doing some string interpolation with its value below, and for py3 it should be decoded instead.

It actually should be decoded either way if you wanted to do proper Unicode handling in the whole Kivy codebase, but that's another can of worms.

kivy/core/window/_window_sdl2.pyx
dolang: Alright, I've looked at it again.

I started out with an `SDL_CreateRGBSurfaceFrom()` call but was unable to make that work. My previous attempt looked like this:
```cython
    cdef char *c_pixels = core_image.image.texture.pixels
    cdef SDL_Surface *icon_surface
    icon_surface = SDL_CreateRGBSurfaceFrom(c_pixels, width, height, 32,
        pitch, 0x00000000ff, 0x0000ff00, 0x00ff0000, 0xff000000)
    return icon_surface
```

`SDL_CreateRGBSurface()` allocates the necessary buffer and `memcpy()` below copies the pixels into it, so that should work correctly. Maybe I'm not getting something, but how would `SDL_CreateRGBSurfaceFrom()` help here? I tried that, where's the error?

kivy/core/window/_window_sdl2.pyx
dolang: Indeed this leaks memory, as mentioned in my [earlier comment](https://github.com/kivy/kivy/pull/5785#issuecomment-395694274). But it's not possible to just call `SDL_FreeSurface()` there. Or at least I was unable to make that work IIRC.

dolang:
Alright, I really appreciate that you guys took the time to look at it, but as you might've guessed I was kind of unhappy with the outcome.

The [5 bullet points](https://github.com/kivy/kivy/pull/5785#issuecomment-395694274) are still unanswered, and I don't think of the code as final either. That's why I'm not sure how much these code reviews are going to help in the end.

KeyWeeUsr:
I'll try to answer some points:

* Cython unit tests?

Pure Cython (basically C) unittests, hehe, good luck with integrating that into Kivy (Win+Lin+Mac & Py2+3 & different .SOs/.DLLs, ...) if you want to poke the low-level stuff and it'll be probably hell maintaining it. Not only we'd need to compile the extensions, we'd also need to compile & link the tests properly and then perhaps if we used some fixed version for the provider libs we might be lucky and it would work. Otherwise, just look at the `tests` folder in the repo, there are casual unittests for Cython extensions too e.g. [`test_clock.py`](https://github.com/kivy/kivy/blob/master/kivy/tests/test_clock.py). Testing pure Cython stuff might be easy probably via implementing `cpdef`ed methods calling the critical `cdef`ed ones just for the sake of being called in a unittest, however I don't see testing pure `cdef`ed methods (or their parts) with some `test_something.pyx` realistically usable here.

* It's close to current logic, but that has a memory leak. So this version has one, too.

I'm not sure what I'm supposed to say. Basically if you can keep it simple and without leaks, that's the welcomed result, no? :yum: 

* Not sure anymore if the images should be loaded with nocache=True here. Original reason for that was: to not keep potentially big images (shaped window) in RAM; the icon would not get loaded several times anyway; seems like that can happen, however

For shapes I'd say use the cache, because someone might decide to swap multiple ones e.g. some launcher with 300x300, then switch to fullscreen with no shape, then perhaps something a'la World of Warcraft launcher with transparent parts. Just pulling random stuff out of nowhere, but if there isn't a massive improvement in disabling cache for shapes, I would rather use the cache.

dolang:
Thanks! This is the kind of context/information I was looking for. 

I'm new to many things here (Kivy internals, Cython), but not all of them (did some low-level programming before, but a while ago) and that makes it hard to put certain things into perspective.

> Basically if you can keep it simple and without leaks, that's the welcomed result, no?

Sure. But let's take that leaking code as an example; I left it like that because the PR isn't complete anyway and having stopped at that point makes it easier to look at the diff.

Of course, leaks are generally bad, but from my perspective there could be several reasons for why the code is like that:
- the Surface gets loaded only once anyway and ...
  - the underlying library has an issue with freeing that memory
  - maybe there's some multiplatform/-provider compatibility problem (which I wouldn't be able to test everywhere even if I wanted to)
- it's simply an oversight and should be fixed
- it was introduced with some unrelated or tangentially related change, i.e. it's a regression
- some other reason i haven't even thought about (maybe another library needs binary compatibility with something?)

-----

> ... Testing pure Cython stuff might be easy probably via implementing cpdefed methods calling the critical cdefed ones just for the sake of being called in a unittest, however I don't see testing pure cdefed methods (or their parts) with some test_something.pyx realistically usable here.

Ok, good to hear that from someone who has experience with Cython programming. We've been talking about that a bit on IRC as well, and I'd like to do a few experiments there.  E.g. there's a [Cython plugin for pytest](https://pypi.org/project/pytest-cython/).

> ... Not only we'd need to compile the extensions, we'd also need to compile & link the tests properly and then perhaps if we used some fixed version for the provider libs we might be lucky and it would work. ...

And yes, I'm aware it's tricky. :wink: 

-----

Anyway, there's no rush to this PR now. I thought if it were easy enough to do, it could've simplified some packaging for the release (see kivy/kivy-sdk-packager#41).

dolang:
**Side note:** To me there's also the question of why some of these things are very C-like instead of using Cython to get some better abstraction/encapsulation. I don't know enough about Cython or SDL yet, but I hope you get the idea ([RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization)):

```cython
cdef class Surface:

    def __cinit__(self, ...):
        self.surface = SDL_CreateRGBSurface(...)
        if self.surface == NULL:
            raise MemoryError(SDL_GetError())


    # various SDL_Surface operations here
    # probably want some generic SDL_GetError() handler method, too


    def __dealloc__(self):
        if self.surface != NULL:
            SDL_FreeSurface(self.surface)
            # and maybe some SDL_GetError() logic again
```

KeyWeeUsr:
I just play with stuff and notice when and why it breaks :) The problem isn't missing an experience with Cython programming, the problem is that unittesting Cython parts that compile directly to C and are callable only from C is nasty because you aren't unittesting Python there, [this is a whole different world](https://www.youtube.com/watch?v=zW_HyDTPjO0)(basically `cpdef` for testing is nice, `cdef` and C unittests are ugly, but he used CFFI instead). With Python you get some proper binaries (the interpreter) and a language that's safe to use, testing with that is easy.

With unittests for Cython you are basically doing **unittests for C** and the more fancy stuff you do (bits and memory operations, etc), the more is your Cython unittest depending on:

* Cython version
* Cython doing its stuff properly and us using it properly (ref `cdef void reload(...)` quirk for `svg.pyx`/`svg.pxd` in 1.9.x - 1.10.x)
* C implementation (MSVC(not sure how to call it), glibc, musl, ...)
* C/C++ libraries linked for the test, e.g. we still haven't figured out why `angle` doesn't work on Appveyor, yet it works locally and that's "just" an OpenGL provider/interface/...
* etc

not only you need to care about building Kivy properly, you especially need to care for the tests because it's not useful to have tests which provide incorrect output. Afaik there isn't a proper unittesting framework for Cython, therefore we'd need to make a custom one. (common function here, common class there and boom, you have a framework, [ref](https://github.com/cython/cython/blob/master/runtests.py)) Cython itself is [having fun with various stuff](https://github.com/cython/cython/tree/master/tests) for their tests and I believe we'll be bombed even more with issues since we use more libraries.

IMO let's completely forget about all the stuff callable only from C and take it as something that works for now... let's make a proper suite for Python parts and [the issues we've encountered](https://github.com/kivy/kivy/issues?utf8=✓&q=label%3Aneeds-unittest)(also feel free to mark some other issues with this label) **first** and *then actually care about the C stuff* in depth, because only then we'll have the **API we provide** tested which means we can safely break a part from the core if necessary and have a way to get it right instead of later doing `git bisect`.

Also... we use `nose` to run tests that use `unittest` module from standard Python library. That plugin has no docs what so ever and does basically less than [`runtests.py`](https://github.com/cython/cython/blob/master/runtests.py) which is actually a part of Cython, therefore if we should use something, we should use what Cython does.

> ... why some of these things are very C-like instead of using Cython to get some better abstraction/encapsulation. ...

Encapsulation is fine, however, we do encapsulation [for *our* objects](https://github.com/kivy/kivy/blob/082c4544415ab6bd976d9dda9c342bae109e7cbb/kivy/graphics/vertex.pyx)(`VertexFormat`, `Buffer`, ...) mainly, because we know how they behave or there is one/two pointers ([1](https://github.com/kivy/kivy/blob/d8ef8c2834293098bc404c0432049b2761f9b721/kivy/core/audio/audio_sdl2.pyx#L114), [2](https://github.com/kivy/kivy/blob/7991cef4184b26894139d68c561dc79bedde7736/kivy/core/text/_text_sdl2.pyx#L27)) that we basically "propagate" to the level of Python objects i.e. keep the memory allocated for the whole life of the object, but sometimes it's just easier (or proper/faster/...) to do this:

```python
cdef foo(...):
    # a lot of stuff
    cdef Something * something = SDL_Something(...)
    # stuff
    free(something)
    # a lot of stuff
```

for example even because of [the order of allocation](https://stackoverflow.com/a/9069486/5994041)([nicer example](https://stackoverflow.com/a/7241413/5994041), [nicer explanation](https://stackoverflow.com/a/7241505/5994041))

> You have to free() the allocated memory in exact reverse order of how it was allocated using malloc().

Just try to mimic OOP with structs and e.g. `mystruct_init()` and `mystruct_free` where you `free()` in the wrong order or n-dim arrays or something, lovely stuff :D

For me personally freeing multiple stuff has sometimes been a sick fun too when playing with C - sometimes the order mattered and sometimes not depending on which compiler and its version I used with the exact same C code, so... better be safe and free asap (also, we aren't Java, free RAM != **we** have to allocate everything) than wrap every SDL_* thing through Cython and rely on GC to take care. Anyway, everything depends on the particular code and what works better for such use-case. [`_SurfaceContainer`](https://github.com/kivy/kivy/blob/7991cef4184b26894139d68c561dc79bedde7736/kivy/core/text/_text_sdl2.pyx#L27) for example.

PR #5797, opened 2018-06-10
fix syntax table for emacs kivy-mode

e-matteson:
"<" and ">" were considered part of symbols, but are now punctuation. This means
selecting the symbol-at-point will now correctly select "Foo" instead of "\<Foo\>".

"_" was part of a word, but is now part of a symbol. This means you can move by
word inside of snake_case symbols, like in other emacs modes.


inclement:
Thanks!

Issue #5558, opened 2018-01-02
release source package should contain cythonized source.


tshirtman:
Installation from source (usually linux), can be painful to users, because of Cython being a moving target, and can be slow to build on some platform (e.g: rapsberry pi), if you have to build multiple versions to find a working one, pain starts to grow. We can't add Cython to dependency (and this fix the version here), because it's not needed at runtime, just at build time, it would be wasteful, and potentially wouldn't even solve the issue, because pip doesn't know it have to install it *before* installing kivy.

However, the Cython project advise shipping C code produced by Cython in packages, to avoid the problem entirely: [distributing cython modules](http://docs.cython.org/en/latest/src/reference/compilation.html#distributing-cython-modules)

I don't think we want to add that to the repository, but we certainly do want to add it to the source package we build for pypi.

This action would be part of the release process, probably just a:

    cythonize **/*.pyx

before the

     python setup.py sdist

for release.

And we could remove the requirement to install Cython for people installing releases (of course, it would still be required for building from master, as having the C source in master would be nightmarish if different devs have different versions of Cython, outputing different C code…)

Any downside to consider?

rnixx:
Actually compiling the cythonized c files is the slow part, not the process of cythonizing itself. Thus I'm not convinced about the performance benefit.

Though I see the benefit of getting rid of cython dependency for installing.

Another problem expecially on rpi is the need of increasing swap file size in order to prevent strange compiler fails by running out of memory, but that's some kind of OT here.

What I'd be interested in is playing with the compile flags to skip debug symbols and other stuff making the build process slower and bloats the compiled *.so files.

KeyWeeUsr:
@rnixx I remember I had to compile Cython first, then Kivy when on RPi and the Cython compilation was sloooow (RPi Zero), so it's quite a benefit. I wonder if we can just distribute the C files (and e.g. distutils can handle them) or if we need to create a Makefile(or aternative) for just the C files.

While I pick a comment from a slightly related issue, I think we might use the existing `setup.py` and create some "options" for distributing as mentioned https://github.com/kivy/kivy/issues/5375#issuecomment-331983558 for example:

    pip install kivy  # source
    pip install kivy[cythonized]  # C files
    etc

or perhaps just the good old `python setup.py build_ext <something>` would be enough. We might not need pip here at all to create C files, however it could help if we decided to distribute both source and cythonized source.

rnixx:
[...]

PR #5808, opened 2018-06-16
create C source in sdist, so cython is not needed for build

tshirtman:
This should do away with the cython requirements to build from source distribution

addresses #5558

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,9 @@ matrix:
       os: linux
       dist: trusty
     - language: python
+      python: 3.5
+      env: RUN=sdist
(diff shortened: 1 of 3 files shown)


tshirtman:
The merge has been reverted, because while it works fine when the sdist is installed on the same platform it's been built (e.g linux on both sides) it doesn't if, for example, you install that same sdist on a windows machine.

This is because the C files generated by cython depends on the config.pxi file, which depends on the platform and choice made by the user. If we want a portable sdist with C files in it, we need to rework that so these options are used at compile time, not at the cythonization stage.

PR #5814, opened 2018-06-18
[WIP] ActionBar overhaul; includes ActionOverflow fix for #5754

### LINKED ISSUE #5754: ActionOverflow disappears when resizing the window
### Versions

* Python: Python 3.6.4 :: Anaconda, Inc.
* OS: macOS 10.13.4
* Kivy: 1.10.1.dev0 

### Description

The ActionOverflow button (3 vertical dots on the right of ActionBar) disappears when the window is resized and the new width is more than the button size smaller than the old width. It reappears when the window is resized to be a little bigger. See attached [mp4 file](https://github.com/kivy/kivy/files/2016571/Actionoverflow_dissapears.mp4.zip).

### Minimal example
The option button on the top-right disappears when the window is resized and the new width is smaller than the old width minus the button width.
```python
from kivy.app import App
from kivy.lang import Builder

kv = """
<MyRoot>:
    orientation: 'vertical'
    ActionBar:
        ActionView:
            ActionPrevious:
                title: ''
                with_previous: True
            ActionOverflow:
                ActionButton:
                    text: 'Preferences'
                    on_press: app.open_settings()
    Label:
        text: 'placeholder'
"""
Builder.load_string(kv)


class MyApp(App):
    def build(self):
        return MyRoot()


class MyRoot(BoxLayout):
    pass


if __name__ == '__main__':
    MyApp().run()
```
### Attachments
Video showing the bug: [3.5MB](https://github.com/kivy/kivy/files/2016571/Actionoverflow_dissapears.mp4.zip) (11s)

dolang:
**WIP** not quite ready for a merge.

- Update the documentation and corresponding images
- Some cosmetic changes
- Improve API consistency and add tests
- Fix layout logic &mdash; or at least improve it; probably needs a bit more work still
  - Partially fixes #5754

Ready for a first review.


kivy/data/style.kv
```diff
@@ -620,7 +620,6 @@
 
 <ActionPrevious>:
     size_hint_x: 1
-    minimum_width: layout.minimum_width + min(sp(100), title.width)
```
dolang: Setting the `minimum_width` achieves nothing, it's counterproductive even:
the event gets triggered twice for some reason and it ends up at 56 instead
of the expected 156.  Also, the docs say these properties are supposed
to be read-only (see `BoxLayout`).

dolang:
So I've managed to stabilise the width computations of the `ActionBar` and it should no longer come to different conclusions about how much width the children take. I've left 2 `print()` in, add them to the
original version and play around with it a bit to compare.

That's the good news. The bad news is that I'm pretty sure all the work I did to fix this module amounts to an API break in one way or another.

Also, I'm still not entirely happy with it:

- Need to figure out how to make the initial size computation consistent, too.
- I'm unsure about `minimum_width`. The way it's set up right now, the computation simply can't work on the `ActionLabel`, but maybe it shouldn't? Is that something people use?  If so, one potential solution might be to add a "spacer" Widget, instead of letting `ActionLabel` take up the remaining space.

PR #5826, opened 2018-06-24
Drop support for py3.3, which is EOL

dolang:
Python 3.3 has reached end of life (since 2017-09-29).

See:
- [Python 3.3.0 Release | Python.org](https://www.python.org/download/releases/3.3.0/)
- [Python Insider: Python 3.3.7rc1 is now available prior to Python 3.3.x end-of-life](https://blog.python.org/2017/09/python-337rc1-is-now-available-prior-to.html)


akshayaurora:
@dolang 👍

PR #5834, opened 2018-06-30
Fix missing requirements for Python 3.6 64bit

KeyWeeUsr:
sdist might need to be first, but without deps it will fail on missing Cython, thus no wheels for 3.6 64b

ref: https://ci.appveyor.com/project/KivyOrg/kivy/build/1.0.2460/job/u6bsnlrwwe5n5yy2#L202

Perhaps `kivy.deps.*` might need to be moved to the original place, I'm not sure.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -135,9 +135,7 @@ build_script:
 
     Check-Error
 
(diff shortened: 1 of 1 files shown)


matham:
We don't want any of the kivy.deps.xxx to be installed when we do the sdist, otherwise they may get into the sdist. So we should only install the non kivy related packages before the sdist.

PR #5841, opened 2018-07-06
fix issues with `disabled` aliasproperty

### LINKED ISSUE #5838: Disabled property doesn't work for views used by RecycleView
### Versions

* Python: 2.7
* OS: Ubuntu 14.04
* Kivy: 1.10.1
* Kivy installation method: pip install git+https://github.com/kivy/kivy.git@1.10.1

### Description

Instances of `SomeView` class should be disabled when `disabled` property is set to `True` in `refresh_view_attrs` method, but that's not the case with Kivy 1.10.1 version. Related to https://github.com/kivy/kivy/pull/5537/

### Code and Logs

```python
from kivy.app import App
from kivy.lang import Builder
from kivy.uix.recycleview.views import RecycleDataViewBehavior


class SomeView(RecycleDataViewBehavior, Button):

    def refresh_view_attrs(self, rv, index, data):
        self.disabled = True
        self.text = 'Disabled is {}'.format(self.disabled)


kv = '''
RecycleView:
    effect_cls: 'ScrollEffect'
    scroll_type: ['bars']
    bar_width: '15dp'
    scroll_wheel_distance: '50dp'
    viewclass: 'SomeView'
    RecycleGridLayout:
        cols: 1
        default_size_hint: 1, None
        default_size: None, dp(50)
        size_hint_y: None
        height: self.minimum_height
'''


class DisabledViewApp(App):

    def build(self):
        rv = Builder.load_string(kv)
        rv.data = [{}] * 30
        return rv


if __name__ == '__main__':
    DisabledViewApp().run()
```

tshirtman:
fixes #5838

`disabled` alias property wasn't saving disabled state when moving a
widget from one parent to another, this was notably visible in `RecycleView`.

Also, part of the fix required to fix `AliasProperty`'s `dispatch` method
to dispatch the correct value.

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -531,7 +531,7 @@ class Widget(WidgetBase):
                                   % (widget, parent))
         widget.parent = parent = self
         # Child will be disabled if added to a disabled parent.
-        widget._disabled_count = self._disabled_count
+        widget.inc_disabled(self._disabled_count)
 
         canvas = self.canvas.before if canvas == 'before' else \
             self.canvas.after if canvas == 'after' else self.canvas
@@ -584,6 +584,7 @@ class Widget(WidgetBase):
         elif widget.canvas in self.canvas.before.children:
             self.canvas.before.remove(widget.canvas)
         widget.parent = None
+        widget.dec_disabled(self._disabled_count)
 
     def clear_widgets(self, children=None):
         '''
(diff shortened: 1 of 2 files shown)


kivy/uix/widget.py
```diff
@@ -1319,19 +1320,19 @@ def set_disabled(self, value):
             else:
                 self.dec_disabled()
```
pythonic64: Maybe you should `return True` here when new value for `self._disabled_value` is set? Because of `set` method in `AliasProperty`

PR #5848, opened 2018-07-09
Document the `data` parameter for add_json_panel()

### LINKED ISSUE #5847: The "data" parameter for add_panel_json() is undocumented
https://github.com/kivy/kivy/blob/fb8c0b9c5c367514f1c6949f86efbc0cae207c2c/kivy/uix/settings.py#L967

The "data" parameter can be used to provide JSON data, instead  of using a file, which I figured out by reading the source. It should be documented in the method description.

Gustra:
fixes #5847

The `data` parameter was undocumented.


tshirtman:
Thanks!

PR #5856, opened 2018-07-12
Update AliasProperty to cache value only if "cache" argument is set to True

pythonic64:
Fixes the case when `bind` is None or empty list/tuple and `cache` is `True`. 
If `force_dispatch` is True, it will always dispatch current value on each setter call.

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -1386,7 +1386,8 @@ cdef class AliasProperty(Property):
     If you don't find a Property class that fits to your needs, you can make
     your own by creating custom Python getter and setter methods.
 
-    Example from kivy/uix/widget.py::
+    Example from kivy/uix/widget.py where x` and `width` are instances of
+    `NumericProperty`::
 
         def get_right(self):
             return self.x + self.width
@@ -1394,21 +1395,38 @@ cdef class AliasProperty(Property):
             self.x = value - self.width
         right = AliasProperty(get_right, set_right, bind=['x', 'width'])
 
+    If `x` were to be an instance level attribute and not Kivy property then
+    you have to return `True` from setter to dispatch value of `right`::
+
+        def set_right(self, value):
+            self.x = value - self.width
+            return True
+
+    If your want to cache the value returned by getter then pass `cache=True`.
+    This way getter will only be called if new value is set or one of the
+    binded properties changes. In both cases new value of alias property will
+    be cached again.
+
+    To make property readonly pass `None` as setter. This way `AttributeError`
+    will be raised on every set attempt::
+
+        right = AliasProperty(get_right, None, bind=['x', 'width'], cache=True)
+
     :Parameters:
         `getter`: function
-            Function to use as a property getter
+            Function to use as a property getter.
         `setter`: function
-            Function to use as a property setter. Properties listening to the
-            alias property won't be updated when the property is set (e.g.
-            `right = 10`), unless the `setter` returns `True`.
+            Function to use as a property setter. Callbacks listening to the
+            alias property won't be called when the property is set (e.g.
+            `right = 10`), unless the setter returns `True`.
         `bind`: list/tuple
-            Properties to observe for changes, as property name strings
+            Properties to observe for changes, as property name strings.
+            Changing values of this properties will dispatch value of the
+            alias property.
         `cache`: boolean
-            If True, the value will be cached, until one of the binded
-            elements changes. If `bind` collection is empty and `cache` is
-            True, `setter` must return True in order to cache new value and
-            trigger its dispatch.
-        `rebind`: bool, defaults to False
+            If `True`, the value will be cached, until one of the binded
+            elements changes or if setter returns `True`.
+        `rebind`: bool, defaults to `False`
             See :class:`ObjectProperty` for details.
 
     .. versionchanged:: 1.9.0
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -1416,7 +1416,7 @@ cdef class AliasProperty(Property):
         `getter`: function
             Function to use as a property getter.
         `setter`: function
-            Function to use as a property setter. Callbacks listening to the
+            Function to use as a property setter. Callbacks bound to the
             alias property won't be called when the property is set (e.g.
             `right = 10`), unless the setter returns `True`.
         `bind`: list/tuple
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -1386,8 +1386,8 @@ cdef class AliasProperty(Property):
     If you don't find a Property class that fits to your needs, you can make
     your own by creating custom Python getter and setter methods.
 
-    Example from kivy/uix/widget.py where x` and `width` are instances of
-    `NumericProperty`::
+    Example from kivy/uix/widget.py where `x` and `width` are instances of
+    :class:`NumericProperty`::
 
         def get_right(self):
             return self.x + self.width
@@ -1452,8 +1452,12 @@ cdef class AliasProperty(Property):
             self.use_cache = 1
 
     def __read_only(self, _obj, _value):
-        raise AttributeError('"{}.{}" property is readonly'
-                             .format(type(_obj).__name__, self._name))
+        raise AttributeError(
+            '"{}.{}" property is readonly'.format(
+                type(_obj).__name__,
+                self._name
+            )
+        )
 
     cdef init_storage(self, EventDispatcher obj, PropertyStorage storage):
         Property.init_storage(self, obj, storage)
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -1452,7 +1452,8 @@ cdef class AliasProperty(Property):
             self.use_cache = 1
 
     def __read_only(self, _obj, _value):
-        raise AttributeError('property is read-only')
+        raise AttributeError('"{}.{}" property is readonly'
+                             .format(type(_obj).__name__, self._name))
 
     cdef init_storage(self, EventDispatcher obj, PropertyStorage storage):
         Property.init_storage(self, obj, storage)
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -1404,8 +1404,10 @@ cdef class AliasProperty(Property):
         `bind`: list/tuple
             Properties to observe for changes, as property name strings
         `cache`: boolean
-            If True, the value will be cached, until one of the binded elements
-            will changes
+            If True, the value will be cached, until one of the binded
+            elements changes. If `bind` collection is empty and `cache` is
+            True, `setter` must return True in order to cache new value and
+            trigger its dispatch.
         `rebind`: bool, defaults to False
             See :class:`ObjectProperty` for details.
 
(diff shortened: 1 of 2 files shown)


pythonic64:
@tshirtman, can you review this pull request?

tshirtman:
Yes, I've had a look earlier and it seems correct, I need to test it properly, but looks good, sorry for the delay.

tshirtman:
Actually, do you feel up to converting these into actual tests ? that would be awesome :), if not i'll try to do it at some point after merging, but it would be better if it's in the same PR.

pythonic64:
Tests would be nice, but sadly I don't have time to write them. These were just some manual test with `print` calls in setter and getter methods. 

I have pull request which adds `cache=True` for most of usage of `AliasProperty` throughout Kivy and I'll push that after this one gets merged.

KeyWeeUsr:
@pythonic64 If you don't want to write the tests, that's fine too, but please just share an example with expected results first and someone will find the time to convert it to a proper test I believe :) If not, I'll add it to my to-do list :yum: It's harder to reproduce what the original intention was if there is no example though, so... :)

tshirtman:
so, here is one test that doesn't pass on master but does on this branch

(kivy/tests/test_properties.py)
```
    def test_aliasproperty_cache(self):
        from kivy.properties import AliasProperty

        called = 0

        def _get(wid):
            nonlocal called
            print(called)
            assert called < 2
            called += 1

        def _set(wid, value):
            return True

        alias = AliasProperty(_get, _set, cache=True)
        alias.link(wid, 'alias')
        called = 0
        alias.set(wid, 1)
        alias.get(wid)
```

I'll look into doing more of them tomorrow.

pythonic64:
@KeyWeeUsr, it's not that I don't want to write the tests, I just don't have the time as I have other work. I talked with @tshirtman about this issue with `AliasProperty` and he is familiar with it. Also, he said that it's ok for me to do the pull request, which is a follow-up of https://github.com/kivy/kivy/issues/5838 and that he will review it. 

Regarding the test cases, they are pretty straightforward. There are eight combinations as `cache` can be `True/False`, `bind` can be a collection or `None` and `force_dispatch` can be `True/False`. 

If `cache` is `True`, value will be saved in `ps.value` and every next attempt to get value will return `ps.value`, otherwise `cache` is `False` and value will never be saved and `getter` method will be called every time when value is requested.

If `bind` is a collection, dispatch of value will occur only on change, but if `cache` is `True` value will be saved in `ps.value` and again that value will be returned every time when is requested. 

At last, if `force_dispatch` is True, current value will be always dispatched on each `setter` call.

KeyWeeUsr:
@pythonic64 Don't get me wrong. :D If you don't want to write/don't have time to write/whatever, that's your thing. IMO you aren't obliged to anything and you don't need to explain it. We're obviously happy for PRs, yet I tend to at least ask what's the expected result of the PR/issue if I notice it (well, except being annoying about pep8 & stuff).

I particularly care mainly about the information about the issue/PR not being lost because I've found some older issues in `kivy/kivy` repo that I couldn't reproduce and were closed because nobody cared to share what's actually wrong except that "it doesn't work". This way (even without tests or PR) there will be at least something in the issue and if someone finds the time to check the [`needs unittest`](https://github.com/kivy/kivy/labels/needs-unittest) tag or if more familiar with the part of the codebase, then s/he will be able to convert it to a program or investigate further :)

Thanks for the write up of the combinations! That's basically what I had in mind :dango:

tshirtman:
So, sorry for taking so long to answer, but i've been trying on and off to work on tests for this, and they don't pass, so i must be confused about something *or* the code is not correct, so in doubt, here they are:

```
    def test_aliasproperty_cache_true_no_bind(self):
        from kivy.properties import AliasProperty

        called = [0]

        def _get(wid):
            called[0] += 1

        def _set(wid, value):
            return True

        alias = AliasProperty(_get, _set, cache=True)
        alias.link(wid, 'alias')
        called = [0]
        assert called[0] == 0
        alias.set(wid, 1)
        assert called[0] == 1
        alias.get(wid)
        assert called[0] == 1

    def test_aliasproperty_cache_true_bind(self):
        from kivy.properties import NumericProperty, AliasProperty
        from itertools import count
        import q
        called = count()

        def _get(wid):
            next(called)
            return 1

        def _set(wid, value):
            return True

        test_bind = NumericProperty()
        test_bind.link(wid, 'test_bind')

        alias = AliasProperty(_get, _set, bind=['test_bind'], cache=True)
        alias.link(wid, 'alias')

        wid.test_bind = 0
        assert str(called) == 'count(0)'
        alias.get(wid)
        assert str(called) == 'count(1)'
        wid.test_bind += 1
        assert str(called) == 'count(2)'
        alias.get(wid)
        assert str(called) == 'count(2)'

```

```
gabriel@gryphon:~/d/kivy> pytest kivy/tests/test_properties.py
========================================================== test session starts ==========================================================
platform linux -- Python 3.6.3, pytest-3.7.4, py-1.6.0, pluggy-0.7.1
rootdir: /home/gabriel/dev/kivy, inifile:
collected 22 items                                                                                                                      

kivy/tests/test_properties.py .F....................                                                                              [100%]

=============================================================== FAILURES ================================================================
_________________________________________ PropertiesTestCase.test_aliasproperty_cache_true_bind _________________________________________

self = <kivy.tests.test_properties.PropertiesTestCase testMethod=test_aliasproperty_cache_true_bind>

    def test_aliasproperty_cache_true_bind(self):
        from kivy.properties import NumericProperty, AliasProperty
        from itertools import count
        import q
        called = count()
    
        def _get(wid):
            next(called)
            return 1
    
        def _set(wid, value):
            return True
    
        test_bind = NumericProperty()
        test_bind.link(wid, 'test_bind')
    
        alias = AliasProperty(_get, _set, bind=['test_bind'], cache=True)
        alias.link(wid, 'alias')
    
        wid.test_bind = 0
        assert str(called) == 'count(0)'
        alias.get(wid)
        assert str(called) == 'count(1)'
        wid.test_bind += 1
>       assert str(called) == 'count(2)'
E       AssertionError: assert 'count(1)' == 'count(2)'
E         - count(1)
E         ?       ^
E         + count(2)
E         ?       ^

kivy/tests/test_properties.py:593: AssertionError
=========================================================== warnings summary ============================================================
kivy/tests/test_properties.py::TestProperty
  cannot collect test class 'TestProperty' because it has a __init__ constructor

-- Docs: https://docs.pytest.org/en/latest/warnings.html
============================================ 1 failed, 21 passed, 1 warnings in 0.88 seconds ============================================

```

I'm pretty sure count here should be 2, because the property is bound and so `_get` should be called to re-evaluate, even if `cache` is true, to return an updated value.

pythonic64:
@tshirtman I think you need to add `alias.link_deps(wid, 'alias')` after `alias.link(wid, 'alias')`, like it's done [here](https://github.com/kivy/kivy/blob/master/kivy/tests/test_properties.py#L152).

tshirtman:
Oh, indeed, that's helpful

```python
    def test_aliasproperty_cache_true_bind(self):
        from kivy.properties import NumericProperty, AliasProperty
        from itertools import count
        called = count()

        def _get(wid):
            next(called)
            return 1

        def _set(wid, value):
            return True

        wid.__class__.test_bind = test_bind = NumericProperty(100)
        test_bind.link(wid, 'test_bind')
        test_bind.link_deps(wid, 'test_bind')

        wid.__class__.alias = alias = AliasProperty(_get, _set, bind=['test_bind'], cache=True)
        alias.link(wid, 'alias')
        alias.link_deps(wid, 'alias')

        wid.test_bind = 0
        assert str(called) == 'count(1)'
        alias.get(wid)
        assert str(called) == 'count(1)'
        wid.test_bind += 1
        assert str(called) == 'count(2)'
        alias.get(wid)
        assert str(called) == 'count(2)'
```

```
$ pytest kivy/tests/test_properties.py
========================================================== test session starts ==========================================================
platform linux -- Python 3.6.3, pytest-3.7.4, py-1.6.0, pluggy-0.7.1
rootdir: /home/gabriel/dev/kivy, inifile:
collected 22 items                                                                                                                      

kivy/tests/test_properties.py ......................                                                                              [100%]

=========================================================== warnings summary ============================================================
kivy/tests/test_properties.py::TestProperty
  cannot collect test class 'TestProperty' because it has a __init__ constructor

-- Docs: https://docs.pytest.org/en/latest/warnings.html
================================================= 22 passed, 1 warnings in 0.56 seconds ================================================
```

looking into the other cases, thanks. (i think only checking dispatch with/without force is missing).

kivy/properties.pyx
```diff
@@ -1386,27 +1386,47 @@ cdef class AliasProperty(Property):
...
+
     :Parameters:
         `getter`: function
-            Function to use as a property getter
+            Function to use as a property getter.
         `setter`: function
-            Function to use as a property setter. Properties listening to the
-            alias property won't be updated when the property is set (e.g.
-            `right = 10`), unless the `setter` returns `True`.
+            Function to use as a property setter. Callbacks listening to the
```
tshirtman: I guess if you change `Properties` to `Callback` then it makes sense to change `listening` to `bound`.

```suggestion
            Function to use as a property setter. Callbacks bound to the
```

pythonic64:
@KeyWeeUsr can you take a look at this pr? 

There is a small typo in docs, but I'll fix it along with your recommendations if any.

kivy/properties.pyx
```diff
@@ -1386,27 +1386,47 @@ cdef class AliasProperty(Property):
     If you don't find a Property class that fits to your needs, you can make
     your own by creating custom Python getter and setter methods.
 
-    Example from kivy/uix/widget.py::
+    Example from kivy/uix/widget.py where x` and `width` are instances of
```
KeyWeeUsr: Missing backtick for `x`.

kivy/properties.pyx
```diff
@@ -1386,27 +1386,47 @@ cdef class AliasProperty(Property):
...
+    `NumericProperty`::
```
KeyWeeUsr: Use this:

    :class:`NumericProperty`

to get the reference for `NumericProperty` in the docs.

kivy/properties.pyx
```diff
@@ -1432,7 +1452,8 @@ cdef class AliasProperty(Property):
             self.use_cache = 1
 
     def __read_only(self, _obj, _value):
-        raise AttributeError('property is read-only')
+        raise AttributeError('"{}.{}" property is readonly'
+                             .format(type(_obj).__name__, self._name))
```
KeyWeeUsr: I feel like a linter :D sorry, but this syntax is (imo) just plainly ugly and only a backslash away from a shell syntax. If you separate the lines you can actually get even nicer errors from the traceback if anything occurs for example:

    raise AttributeError(
        '...'.format(
            type(_obj).__name__,
            self._name
        )
    )

and if you can't access `__name__` or `_name` or the format is wrong, you get the right line instead of a line with multiple accesses or calls. And it's more readable if you just scroll down the code on a phone (more space around specific objects, eyes will see the desired objects sooner).

pythonic64:
@KeyWeeUsr, I've updated `AliasProperty` with your suggestions. 
I think you can merge this request now 😃

KeyWeeUsr:
Thank you both :-)

PR #5866, opened 2018-07-19
Add support for cross-compiling for the raspberry pi

foozmeat:
Without these patches its not possible to cross-compile kivy to include on a raspberry pi OS image.

This adds the following

* `KIVY_CROSS_PLATFORM` to override the platform
* `KIVY_CROSS_SYSROOT` to prefix the include and lib paths

The changes to videocore simply switch the paths to relative.


chintal:
This would be a welcome change to get me to try out buildroot, since my precise requirement is getting a kivy based application to run on an epi. Is there anything I can do to help get this merged?

tshirtman:
As this introduce some environment settings, it would indeed be nice to document them somewhere, but i don't think it's a reason to block the addition, which has apparently no way to disturb things outside of the rpi platform, so as long as it's been tested there (with and without cross-compile) i'd be ok to merge.

setup.py
```diff
@@ -640,18 +646,31 @@ def determine_gl_flags():
...
+            brcm_lib_files = (
+                cross_sysroot + '/usr/lib/libbrcmEGL.so',
+                cross_sysroot + '/usr/lib/libbrcmGLESv2.so')
+
         if all((exists(lib) for lib in brcm_lib_files)):
-            print(
-                'Found brcmEGL and brcmGLES library files'
-                'for rpi platform at /opt/vc/lib/')
+            print('Found brcmEGL and brcmGLES library files'
+                  'for rpi platform at ' + dirname(brcm_lib_files[0]))
```
pythonic64: One space character is missing at then end of first part of the string, so this now prints "... library filesfor rpi..." instead of "... library files for rpi..."

tito: PR ? :)

PR #5867, opened 2018-07-22
Added a scale argument to Widget.export_to_png.

inclement:
Seems to work.

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -602,7 +602,7 @@ class Widget(WidgetBase):
         for child in children[:]:
             remove_widget(child)
 
-    def export_to_png(self, filename, *args):
+    def export_to_png(self, filename, *args, scale=1):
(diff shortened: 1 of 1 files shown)


kivy/uix/widget.py
```diff
@@ -602,7 +602,7 @@ def clear_widgets(self, children=None):
         for child in children[:]:
             remove_widget(child)
 
-    def export_to_png(self, filename, *args):
+    def export_to_png(self, filename, *args, scale=1):
```
tshirtman: I don't think that's python2 compatible
```
Python 2.7.15rc1 (default, Apr 15 2018, 21:51:34) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> def test(a, b, *args, c=1):
  File "<stdin>", line 1
    def test(a, b, *args, c=1):
                          ^
SyntaxError: invalid syntax
```

PR #5868, opened 2018-07-22
anim_kwargs in PageLayout

JakubBlaha:
Added `anim_kwargs` property to `PageLayout`. You are now able to customize the `Animation` on page switching.


tshirtman:
Please fix style violations, double empty lines doesn't belong between methods, and the last line of the file is supposed to have an end of line character (probably an editor misconfiguration).

other than that it seems fine.

PR #5870, opened 2018-07-23
Provider matching for input postproc calibration

LukeBiddle:
Allows users to match calibration rules to any and all devices of a particular input provider. This is useful when the name of the device is not known in advance. For example, if probesysfs is used to match an mtdev touchscreen:

`[input]`
`mtdev_%(name)s = probesysfs,provider=mtdev`

then the touchscreen can be calibrated using:

`[postproc:calibration]`
`(mtdev) = xratio=0.3333,xoffset=0.3333`

The new convention is to use parentheses to denote a provider name instead of a device.

I mainly made this change to work with https://github.com/kivy/kivy/pull/5659 to apply auto calibration to a resizable app no matter what device it is installed on.

I have only tested this PR with an mtdev touchscreen, but it should work for any input that might need calibrating.

### DIFF
--- a/kivy/input/postproc/calibration.py
+++ b/kivy/input/postproc/calibration.py
@@ -26,6 +26,22 @@ Then, you can use the calibration postproc module::
 Now, the touches from the left screen will be within 0-0.3333 range, and the
 touches from the middle screen will be within 0.3333-0.6666 range.
 
+You can also match calibration rules to devices based on their provider type.
+This is useful when probesysfs is used to match devices. For example::
+
+    [input]
+    mtdev_%(name)s = probesysfs,provider=mtdev
+
+To apply calibration to these devices, you can match with the provider name
+enclosed by parentheses::
+
+    [postproc:calibration]
+    (mtdev) = xratio=0.3333,xoffset=0.3333
+
+Matching devices like this means the device's path doesn't need to be
+configured ahead of time. Note that with this method, all mtdev inputs will
+have the same calibration applied to them. For this reason, matching by
+provider will typically be useful only when expecting one input device.
 '''
 
 __all__ = ('InputPostprocCalibration', )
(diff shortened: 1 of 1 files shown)


kivy/input/postproc/calibration.py
```diff
@@ -74,6 +98,33 @@ def __init__(self):
                 params[key] = float(value)
             self.devices[device_key] = params
 
+    def _get_provider_map(self):
+        """Iterates through all registered input provider names and finds the
+        respective MotionEvent subclass for each. Returns a dict of MotionEvent
+        subclasses mapped to their provider name.
+        """
+        provider_map = {}
+        for input_provider in MotionEventFactory.list():
```
LukeBiddle: This is only called once in __init__, and I was unsure if MotionEventFactory.list() might ever change its output after this. Can anyone confirm?

kivy/input/postproc/calibration.py
```diff
@@ -74,6 +98,33 @@ def __init__(self):
...
+            if not hasattr(providers, input_provider):
+                continue
+
+            p = getattr(providers, input_provider)
+            for m in p.__all__:
+                event = getattr(p, m)
+                if issubclass(event, MotionEvent):
+                    provider_map[event] = input_provider
+
+        return provider_map
```
LukeBiddle: This is what provider_map looks like:

```
{
<class 'kivy.input.providers.tuio.Tuio2dCurMotionEvent'>: 'tuio',
<class 'kivy.input.providers.tuio.Tuio2dObjMotionEvent'>: 'tuio',
<class 'kivy.input.providers.mtdev.MTDMotionEvent'>: 'mtdev',
<class 'kivy.input.providers.hidinput.HIDMotionEvent'>: 'hidinput',
<class 'kivy.input.providers.linuxwacom.LinuxWacomMotionEvent'>: 'linuxwacom',
<class 'kivy.input.providers.leapfinger.LeapFingerEvent'>: 'leapfinger'
}
```

kivy/input/postproc/calibration.py
tshirtman: Touch providers are supposed/assumed to be registered before all of this already, while it would be cool to be allow to enable/disable them at runtime, this is currently not really possible anyway, so no worries here.

PR #5874, opened 2018-07-24
KV Evolution proposal

matham:
KV Evolution proposal
======================

To reviewers, please read the whole proposal (once all sections are complete), because although long and in depth, it does provide a full specification.

Following is a proposal for the evolution of KV, including it's integration into pure python and an associated compiler. For end users, writing KV code will seamlessly integrate into the rest of their python code, while automatically binding KV rules inline. To remove any ambiguity of where binding occurs and how to effect it, the goal of this project was to make binding syntactically clear, with the minimal amount of line noise and to remove any ambiguity in how KV rules are applied - it is all fully speced out.

This proposal is completely independent of traditional KV, in fact the amount of changes to Kivy is a few lines of code - the new code resides in its own module under `kivy/lang/compiler`. The new tools can be used simultaneously with traditional KV, with zero interactions between them. This means, that although I hope the proposal is merged into Kivy and that future KV will be written with the new syntax, it doesn't have to be and can still be supported as a side tool with little effect on the rest of Kivy. 

But I am very excited for this new tool because I think it will solve many longstanding issues with KV, such as rule order instantiation, knowing when rules are instantiated, and unbinding. With this proposal, KV is integrated with python, so we can easily theme, by overwriting the functions that apply the KV rules. We can also unbind all, or specific  rules at will. KV rules can also be named. And much more. On a somewhat complex KV widget (7 KV rules), there was a  `2.6x` speed improvement with the compiler. Performance comparison is expected to increase with increasing KV complexity because bindings are now batched across a context.  There is also almost `200` unittests written, testing all aspects of the new syntax.

I consider the implementation complete and ready to be used, although I will of course welcome any discussion on potential improvements. This project only supports python 3.5+ and I don't ever expect to add python 2 support as the amount of effort to do that, will be pretty much the same as starting from scratch. If people like this proposal, consider it my way of motivating people to move to python 3. If merged, this code, however, will not prevent Kivy from being used in python 2.

Syntax
--------

### Intorduction

Finally, the fun part - the new syntax. Consider a traditional KV rule:
```
<MyWidget>:
    x: self.y
```
We write this as
```python
class MyWidget(Widget):
    @KV()
    def apply_kv(self):
        with KVCtx():
            self.x @= self.y
```
I'll go over all the components. `@KV()` returns a decorator which compiles the function and returns a new function that performs the bindings. We'll go over the decorator options later. `apply_kv` can be any function and may contain any python code (except as rejected by the compiler if it interferes with the bindings, which should be rare - the decorator will raise an exception at import time, when the function is compiled for that case, so it'll be obvious).

All KV rules must be under a `KVCtx` and a `KVCtx` may contain many KV rules, which are executed independently. However, all the initial bindings within a `KVCtx` are batched and are executed when the context exits.

A rule is created using `expression_left @= expression_right`. `expression_left` and `expression_right` can be any legal python and will bind to `expression_right`, such that whenever any Kivy properties in it changes, `expression_left = expression_right` is executed.

Here are more examples of the function:
```python
def apply_kv(self):
    self.widget = widget = Widget()
    offset = self.offset
    with KVCtx():
        self.x @= widget.width + self.y + offset
```
In the example above, the value of `offset` is captured and when the rule is ever triggered, it will use that initial value. The same for `self` and `widget`. But of course, `widget.width` and `self.y` is re-evaluated every time. 

Another example:
```python
def apply_kv(self):
    self.widget = widget = Widget()
    with KVCtx():
        offset = self.offset
        self.x @= widget.width + self.y + offset
```
or even
```python
def apply_kv(self):
    self.widget = widget = Widget()
    with KVCtx():
        offset = self.offset
        widget.width @= self.y + offset
```
Notice how the left side can be anything? Like in the previous examples, `self`, `widget`, and `offset` are captured and their initial values will be used whenever the rule is triggered.

As you can see, the right and left sides of the `@=` can be anything, and that line is executed initially when it is encountered, and then once the KVCtx exits, the bindings will be executed and from then on, whenever the right hand property values change, the line will be executed (without the `@` sign of course). Naturally, the bindings capture the value of any locals and globals at the time the rule is encountered. In fact, to keep the bindings consistent, the compiler will raise an exception if a local or global variable used in a rule is changed between the rule definition and the context exiting. `This makes sure that code is easy to understand and reason about.`

So this example:
```python
def apply_kv(self):
    widget = Widget()
    with KVCtx():
        self.x @= widget.width
        widget = Widget()
```
will raise a parser exception because when the bindings are executed when the context exits, the `widget` is not the same as when it was encountered and used to initialize `self.x`. The exceptions are very descriptive so it's easy to fix when encountered.

KV Rules
--------

### Explicit Rules And Events

KV rules can be created more explicitly, which provide a lot more control over their execution. The example above:
```python
def apply_kv(self):
    with KVCtx():
        self.x @= self.y
```
can be also written as:
```python
def apply_kv(self):
    with KVCtx():
        with KVRule():
            self.x @= self.y
```
The benefits is that anything under a `KVRule` is considered as **one** rule and is executed as such. Consider the following:
```python
def apply_kv(self):
    with KVCtx():
        with KVRule():
            self.width @= self.height + 1
            self.name = 'johnny'
            self.x @= self.y
```
Now, whenever either `self.height` or `self.y` change, all three lines will be executed sequentially as if it's one code block, which it is.

A `KVRule` also accepts positional arguments, and will bind to whatever is provided there, this is an alternative for using the `@` syntax and allows binding to kivy **events** as follows:
```python
def apply_kv(self):
    with KVCtx():
        with KVRule(self.on_touch_down, self.width, self.y, self.widget.height):
            self.x @= self.y + self.compute()
```
In the rule above, the rule will be bound to all of the provided args! That's how you can bind to kivy events, just provided it as an argument. Notice how `self.y` is provided both as an argument and using the `@=` syntax - that's okay, but it could also have been written as:
```python
def apply_kv(self):
    with KVCtx():
        with KVRule(self.on_touch_down, self.width, self.y, self.widget.height):
            self.x = self.y + self.compute()
```
But either way is fine and duplications are not a problem.

Also, the above could have been written as 
```python
def apply_kv(self):
    with KVCtx():
        with KVRule('self.on_touch_down', 'self.width', 'self.y', 'self.widget.height'):
            self.x = self.y + self.compute()
```
the compiler parses them identically, whether it's in string form or code form, but code form is easier as it's more IDE friendly due to auto-completion :).

In fact, you should know that the parser looks for a with statement whose context manager is literally written as `KVRule` and then modifies that line to create a rule in the compiled code. So the line `with KVRule(...)` is treated magically. That and the `with KVCtx()` line and the `@=` and `^=` lines are the only objects that modify the standard python syntax and do something that is unexpected in normal python.

### Canvas and Clock Rules
In traditional KV, canvas rules are treated differently and the rules are only executed at the end of each frame - this helps with graphical performance. We have a similar construct, as opposed to rules created with the `@=` syntax or `with KVRule()`, which is the same as `with KVRule(delay=None)`, rules created using `^=` and `with KVRule(delay='canvas)`, will schedule the rules to be executed with the other graphics instructions, rather than immediately when triggered.

For example:
```python
def apply_kv(self):
    with self.canvas:
        rect = Rectangle()
    with KVCtx():
        rect.x ^= self.x
```
or:
```python
def apply_kv(self):
    with self.canvas:
        rect = Rectangle()
    with KVCtx():
        with KVRule(delay='canvas'):
            rect.x ^= self.x
```
The options between canvas and normal rules regarding binding options etc. are the same. Naturally, one rule cannot mix both canvas and normal rules, so within a rule either `@=` or `^=` must be used.

We also allow for rules to create a Clock trigger, so that when the rule bindings are triggered, instead of executing immediately, it triggers the clock event with provided delay. This allows KV rules to be batched per frame or targeted for the future. To use, you must use a explicit rule and provide a number to delay - which will be delay used when creating the clock event. E.g.
```python
def apply_kv(self):
    with KVCtx():
        with KVRule(delay=0):
            self.x @= self.x
```
will delay and batch all the rule updates for the next clock frame. The syntax is otherwise idnetical to normal KV rules.

### Loops, conditionals, and other constructs
Whenever a rule is encountered within a `KVCtx`, that rules bindings will occur when the context exits. This would be the case even if the condition was False. COnsequently, a rule is now allowed under any conditional code and the parser will raise an exception. **However**, you can wrap the rule under a new `KVCtx` and place that under the conditional. That makes **all** the contents under the nested `KVCtx` completely isolated from the outer context it may appear in.

For example, the following examples will raise a parser exception:
```python
def apply_kv(self):
    with KVCtx():
        if self.height:
            self.x @= self.x
```
and
```python
def apply_kv(self):
    with KVCtx():
        for i in range(2):
            self.x @= self.x
```
and even
```python
def apply_kv(self):
    with KVCtx():
        try:
            print('hello')
        except Exception as e:
            self.x @= self.x
```
and so on.

However, it can be converted as follows and is now perfectly fine:
```python
def apply_kv(self):
    if self.height:
        with KVCtx():
            self.x @= self.x
```
and
```python
def apply_kv(self):
    for i in range(2):
        with KVCtx():
            self.x @= self.x
```
and even
```python
def apply_kv(self):
    try:
        print('hello')
    except Exception as e:
        with KVCtx():
            self.x @= self.x
```
In all cases, the `KVCtx` will only be executed if the `KVCtx` is executed. Although the code will of course always be compiled ahead of time when the function is originally compiled.

The following is also allowed:
```python
def apply_kv(self):
    with KVCtx():
        self.height @= self.width + 45
        if self.height:
            with KVCtx():
                self.x @= self.x
        self.name @= self.widget.your_name
```
In the above example, the rule in the inner context is completely isolated and independent from the outer context. Similarly, from the outer context's POV, it sees only two rules, the `height` and `name` rules.

For the loop example, each iteration creates a new and independent context, which are independent of each other. So the following works as expected:
```python
def apply_kv(self):
    src_widgets = [Widget(), Widget()]
    target_widgets = [Widget(), Widget()]
    
    for i in range(2):
        with KVCtx():
            target_widgets[i].x @= src_widgets[i].x
```
i.e. each widget's `x` is bound to the corresponding widget's `x`. This is also perfectly legally written as:
```python
def apply_kv(self):
    src_widgets = [Widget(), Widget()]
    target_widgets = [Widget(), Widget()]
    
    for target, src in zip(target_widgets, src_widgets):
        with KVCtx():
            target.x @= src.x
```

### Rule largs

It is sometimes desirable to get the arguments that was used to trigger the rule. E.g. when binding to `widget.on_touch_down`, the triggering touch will be dispatched. Following is how to access it in a rule:
```python
def apply_kv(self):
    with KVCtx():
        with KVRule() as my_rule:
            print(my_rule.largs)
            self.x @= self.x
```
When a explicit rule is created and assigned to something, the compiler will automatically assign the captured args used when triggering the rule to the `largs` attribute of the rule. Keep in mind that `largs` may be empty, e.g. if it's dispatched during binding, so be sure to check if it's an empty tuple first.

### Rule Unbinding, named rules

A `KVRule` has some attributes and methods of interest. Specifically, `KVRule` has an `unbind_rule` method, which when called will unbind the rule and it will not be triggered again. However, it may **only** be called once the rule is fully created after the `KVCtx` has exited as in the following example:
```python
def apply_kv(self):
    with KVCtx():
        with KVRule() as my_rule:
            self.x @= self.x
    my_rule.unbind_rule()  # goodbye rule
```
Similarly, the `KVCtx` has a `unbind_all_rules` methods, which unbinds all the rules in the context. E.g.:
```python
def apply_kv(self):
    with KVCtx() as ctx:
        with KVRule():
            self.x @= self.x
        self.width @= self.height
    ctx.unbind_all_rules()  # goodbye all the rules
```

Every rule that is created under a context is added to that context's `rules` list in sequential order. `KVCtx` also has a `named_rules` dictionary, for rules that are explicitly named. E.g.:
```python
def apply_kv(self):
    with KVCtx() as ctx:
        with KVRule(name='my_rule'):
            self.x @= self.x
        self.width @= self.height
    assert len(ctx.rules) == 2
    assert len(ctx.names_rules) == 1
    assert ctx.names_rules['my_rule'] is ctx.rules[0]
```
As can be seen from the asserts, the rules can be named and then accessed by name. They can also be accessed by index ordered by the order in which it was created.

Further specification and options
=================================

Background
-------------

The KV compiler uses AST parsing an manipulation to generate the compiled function. Given a function, if we have not already compiled a kvc file for this function, or if it's stale (source has since been changed or compile flags are different), it loads the source of the function using the inspect module and parses its AST. It then  modifies the AST and dumps it to a new `kvc` file under the `__kvcache__` directory in the same folder as the original function file. 

The compiled file is then loaded and reused, making use of a runtime cache to speed up disk access. The existence of such a file will make debugging much easier as ew can just inspect it and step through it with the debugger. Although we currently do not provide any user side details on the compiled function structure.

To make sure the context the compiled function runs in is identical to the context of the original un-compiled function, every time the function is called, we set the locals/globals of the original function to the one of the compiled function.

Initially I looked into using f-strings, but it became clear that will not work.

Justification
--------------

Although using AST transformations is niche, it is justified and part of the python standard API. So it should work in e.g. pypi, although it is untested. In fact there are other projects with such a need who make use of inline AST modifications. Examples of popular projects doing that are: [pony](https://github.com/ponyorm/pony), [xpyth](https://github.com/hchasestevens/xpyth), [google pasta](https://github.com/google/pasta), and most famously [pytest](https://docs.pytest.org/en/latest/) and [ipython](https://ipython.org/). You can read more about the AST [ecosystem](https://ep2018.europython.eu/conference/talks/exploring-the-python-ast-ecosystem).

Python is also notoriously conservative when it comes to addding new syntax, e.g. there was only one minor AST change that we needed to add to support 3.7, after targeting 3.5. In addition, to prevent sudden breakage, we keep a whitelist of known AST nodes, so that when new appear the compiler will refuse to compile until the new nodes are accounted for.

I have also made liberal use of asserts everywhere in the compiler. And in general, the compiler stage is not at all optimized, but that is fine for now.

I'm hoping to add more comments to the code, but pretty much all of the algorithms I used are based on graph search or manipulations, so it's correctness can be proven. I intend to eventually add more comments with such info into the code, if accepted.

Proxying and garbage collection
-------------------------------

The idea behind using a proxy reference to a widget, is that if a widget that doesn't get garbage collected holds a reference to another widget, the other widget will also not die. We solve this by providing a parameter to the decorator which tells it to make the long living widget hold only a proxy reference to the other widgets. E.g. consider:
```python
class MyWidget(Widget):
    @KV(proxy='app')
    def apply_kv(self):
        app = App.get_running_app()
        with KVCtx():
            self.x @= app.x
```
This bind to `app`, however, because app will never die, `MyWidget` will also never die. However, in the decorator we specified `proxy='app'` - that makes the compiler use a proxy for all the references `app` holds on to. By default `proxy` is `False`, but it can be set to `True` and all will bindings will only store proxies. Or, it can be a glob style string or list of glob style strings which match the widgets that may only hold proxies. See the documentation of `KV` for more details.

A note of caution, if all the bindings use proxy references, **no one** will hold on to the KV rules and the rules will stop doing anything as soon as garbage collection runs. The solution is to store a reference to each `KVCtx` - as long as that is alive, the rules are also alive.

Rebind
-------

Usage is similar to `proxy`. It indicates the widgets to which we rebind. E.g. for
```python
class MyWidget(Widget):
    @KV()
    def apply_kv(self):
        with KVCtx():
            self.x @= self.widget.x
```
if `rebind` is True, the default, then when `self.widget` changes, we rebind the rule to be triggered when then widget's `width` changes.

By default `rebind` is `True`, but it can be set to `False` and no intermediate widgets will be rebind. Or, it can be a glob style string or list of glob style strings which match the widgets that should be rebind. See the documentation of `KV` for more details. E.g. `rebind='*widget''` will rebind to `self.x @= self.widget.x`.

As opposed to traditional KV, we do not inspect the property to see if it set to rebind, because we feel that violates action at a distance, and whether a widget is rebind should be specific to the rules, and should have to be known when a property is declared. E.g. should a widget's `parent` be rebind? The answer is that we don't know and should therefore be a decision made for each compiled function.

Binding on context enter vs exit
--------------------------------

In previous examples we assumed that the bindings occur when the `KVCtx` exits. Although recommended, this is but one option. By setting `bind_on_enter` in the decorator to `True`, the bindings will actually occur upon context enter. Although fully functional, this is not recommended as it is not intuitive because the bindings would happen before the rules are executed.

Nonetheless, it is offered as an option for the following reason. When binding occurs upon context exit, the bindings occur **after** the rules are executed. This means that some rules could now be stale, because when variables were set during rule execution, bindings were not in place. But, if the bindings had occurred before the rule execution, this is not a problem.

We solve this problem for the binding on exit configuration by providing the `exec_rules_after_binding` option in the `KV` decorator. If `True`, it will cause the rules to be executed again after all the bindings. This is not normally needed and reduces performance, so it defaults to `False`. But, the `bind_on_enter=True` could be used rather, rather then the default `bind_on_enter=False` combined with `exec_rules_after_binding=True`, when needed as the former is more performant than the latter case.

Globals and locals capturing within KV decorated function
---------------------------------------------------------

To be filled in

Expressions that get bound
---------------------------

To be filled in

KV Decorable functions
-----------------------

The KV decorator does not support the decoration of closure functions. That means it can only decorate functions at the top level of a module. E.g.:
```python
class MyWidget(Widget):
    @KV()
    def build_kv(...):
        pass
```
is allowed. As is:
```python
@KV()
    def build_kv(...):
        pass
```

but the following is not allowed:
```python
class MyWidget(Widget):
    def apply_kv(self):
        @KV()
        def build_kv(...):
            pass
```

KV restrictions
---------------

Following are things that will cause the compiler to raise an exception.
* A return statement within a `KVCtx`.
* A `KVRule` not under a `KVCtx`.
* A `KVRule` that is under conditionally executing code such as an `if` or `for` loop, unless it's wrapped with a `KVCtx`.
* Overwriting a read only captured variable (see other sections).
* A `KV` decorated function can have one, and only one decorator - the `KV` decorator.
* Nesting a `KVRule` or `KVCtx` under another `KVRule`.
* A `KVRule` may not contain a function or class definition because it's unclear the meaning of such as defintion in a rule that is repeatedly executed. Also, it creates issues with ensuring the consistency of local variables.
* The `global` or `nonlocal` keyword may not occur within a function that is compiled by the `KV` compiler.
* The `del` statement is illegal within a KV rule.
* Mixing canvas and normal or clock KV syntax within a rule.

Manual KV compiler
---------------------

In addition to the KV compiler described above and below, we also provide a manual compiler for debugging and unofficial usage and in order to have a complete implementation. Following is a comparison using the KV compiler and manual compilation:
```python
class ManualKVWidget(Widget):
    def apply_kv(self):
        ctx = KVParserCtx()

        def manage_val(*largs):
            self.x = self.y + 512
        manage_val()

        # will bind to `self.y`.
        rule = KVParserRule('self.y + 512')
        rule.callback = manage_val
        rule.callback_name = manage_val.__name__
        ctx.add_rule(rule)

        KV_apply_manual(ctx, self.apply_kv, locals(), globals())
```
vs. the automatic compiler, which would look like as follows:
```python
class ManualKVWidget(Widget):
    @KV()
    def apply_kv(self):
        with KVCtx():
            self.x @= self.y + 512
```
Both examples effect a compilation, but the decorator version happens automatically and is more efficient.

Requirements for property binding
----------------------------------

The following are required methods to exist on a object, in order for binding to occur on the object's property. We `fbind` and `fbind_proxy` return `Falsy` if it fails and does not raise an error - we therefore do not check at all whether the property is a Kivy property before binding. Duck typing FTW.
* Must have the following functions
    * `fbind` or `fbind_proxy` or both, depending on the `proxy` setting.
    * `unbind_uid`. If the former exists, `unbind_uid` will be assumed to exist and not checked before usage.
    
For code correctness, we also assume that calling `fbind` will have no side effects.

Some simple rules and style guide
---------------------------------

To be filled in
* All KV rules within a context are *always* bound and cannot therefore be under any conditional or loop etc.
* To place a KV rule under a conditional or loop, wrap it with a new context and place the context under the conditional.
* By default no proxies are used. If proxies are used and all objects are proxied, the rules will die immediately as no one will have a reference to them. So make sure to only proxy objects that are not needed to keep the rule alive. Additionally, as long as someone has a direct ref to the context, the contexts and its rules will not die.
* The start and end of a KVCtx is special: that's where the bindings occur, depending on the setting, so whatever variable is referenced by a rule, that variable must exist at the start or end of the context depending on the setting. Similarly, the value of the variables that are bound, are the values of the variables at that fence point. If the variables are changed during the rule, during the initialization, all the code is executed, it's just the binding that captures it.
* clearly we can use locals to change a local, but we disallow it on principle the easy way
* kv rules are not finalized until the context exists. But `largs` is always available.
* locals cannot be None is used in binding.
* Use proxy for widgets that never die, to prevent them from keeping other widgets from dying. But, make sure to store the KVCtx somewhere if all widgets store only proxies, otherwise, if no one holds on to the KV rules, it will be garbage collected.
* During binding, local and global variables that are potentially bound are assumed to not be None as for performance reasons it's not practical to check each local whether it's None. But deeper variables, e.g. `widget` in `self.widget.x` will be inspected and only bound if not None.

Explain that the `with KVRule` and `with KVCtx` are magic, not real context manager. Actually entering will raise error.

style guide: try to keep kv code from mixing with other code
save the ctx

Theming Demo
=============

Following is a demo of how theming may work with this proposal. Consider the following two classes, the first may be considered to have a default theme, the second wants to overwrite the theme.
```python
class MyWidgetTheme(Widget):

    def __init__(self, **kwargs):
        super(MyWidgetTheme, self).__init__(__no_builder=True, **kwargs)
        self.build_kv()

    @KV()
    def build_kv(self):
        with KVCtx():
            self.x @= self.y

class MyWidgetThemeMaterialDesign(MyWidgetTheme):
    @KV()
    def build_kv(self):
        with KVCtx():
            self.x @= self.width
```
When run, we get the following:
```python
>>> w = MyWidgetTheme()
>>> print(w.x, w.y, w.width)
0 0 100
>>> w.y = 46
>>> print(w.x, w.y, w.width)
46 46 100

>>> themed_widget = MyWidgetThemeMaterialDesign()
>>> print(themed_widget.x, themed_widget.y, themed_widget.width)
100 0 100
>>> themed_widget.y = 43
>>> print(themed_widget.x, themed_widget.y, themed_widget.width)
100 43 100
>>> themed_widget.width = 25
>>> print(themed_widget.x, themed_widget.y, themed_widget.width)
25 43 25
```

As can be seen, the two "themed" widgets, each apply their own rule. In other words, we can use the full power of python inheritance with KV rules for theming.

Example of full widget with traditional KV vs the proposed syntax.
===================================================================

Following is the traditional description of `Label`'s KV rule
```
<Label>:
    canvas:
        Color:
            rgba: 1, 1, 1, 1
        Rectangle:
            texture: self.texture
            size: self.texture_size
            pos: int(self.center_x - self.texture_size[0] / 2.), int(self.center_y - self.texture_size[1] / 2.)
```
Here it is with the proposed new syntax.
```python
@KV()
def apply_kv(self):
    with self.canvas:
        Color(1, 1, 1, 1)
        rect = Rectangle()
    
    with KVCtx():
        rect.texture @= self.texture
        rect.size @= self.texture_size
        rect.pos @= int(self.center_x - self.texture_size[0] / 2.), int(self.center_y - self.texture_size[1] / 2.)
```
Following is the traditional description of `Button`'s KV rule
```
<-Button,-ToggleButton>:
    state_image: self.background_normal if self.state == 'normal' else self.background_down
    disabled_image: self.background_disabled_normal if self.state == 'normal' else self.background_disabled_down
    canvas:
        Color:
            rgba: self.background_color
        BorderImage:
            border: self.border
            pos: self.pos
            size: self.size
            source: self.disabled_image if self.disabled else self.state_image
        Color:
            rgba: 1, 1, 1, 1
        Rectangle:
            texture: self.texture
            size: self.texture_size
            pos: int(self.center_x - self.texture_size[0] / 2.), int(self.center_y - self.texture_size[1] / 2.)
```
Here it is with the proposed new syntax.
```python
@KV()
def apply_kv(self):
    with self.canvas:
        color = Color()
        border = BorderImage()
        Color(1, 1, 1, 1)
        rect = Rectangle()

    with KVCtx():
        self.state_image @= self.background_normal if self.state == 'normal' else self.background_down
        self.disabled_image @= self.background_disabled_normal if self.state == 'normal' else self.background_disabled_down
        color.rgba @= self.background_color

        border.border @= self.border
        border.pos @= self.pos
        bordersize @= self.size
        border.source @= self.disabled_image if self.disabled else self.state_image

        rect.texture @= self.texture
        rect.size @= self.texture_size
        rect.pos @= int(self.center_x - self.texture_size[0] / 2.), int(self.center_y - self.texture_size[1] / 2.)
```
As one can, they look very similar, but the python syntax can be more easily integrated into the rest of the widget's python code.

Things to work out
===================

* To compile a function, we need to be able to load the source using the inspect module. Are the situations where the source code is not available? If we cannot find source code, should we just provide a way/place to look for previously compiled files and use that instead? How exactly will that work?
* I chose to save each compiled function into its own file. I don't really see issues with that, and there's also no clear reasonable way currently to combine all the compiled functions of a file into a single file. So unless this becomes an issue, single file per function seems reasonable.
* If we were to use this for theming, we will need to come up with standard function names that apply the KV rules, and perhaps a variable that keeps the KV context for ease of rebinding.


matham:
So first, thanks for your comments :) Following is my responses:

> Tremendous work on this! In your mind, do you see this as a replacement of KV lang or complementary? It's a little bit sad to see the proposal going away from the KV lang itself, but there is a lot of interesting stuff in it!

My feeling and intention on this is almost very similar to the python/cython split. Traditional KV should be used for writing quick apps, or simple rules when you don't really need extra power or when the issues of mixing it in with python is not really a problem. But if you want more power or integration with python or for performance reasons, you'd switch to writing it in the new syntax, see at the end below for more of what I mean. Or, after a quick prototype, you switch to using the new syntax. Really, it'd be up to how people want to use it, if they want to use it all.

And honestly, while I would be a little sad, I'm willing to drop it if people don't think it's going to be helpful. It was fun doing it and I loved learning about AST manipulation in python and graph manipulations, but I won't push it if it's not useful or just adds more difficulty in usage.

> I really like the way you addressed the loop and conditionnal. That's something we lack of from a long time. However, my general feeling right now is (imo obviously):

Thanks :)

> all your example miss the necessary import to be able to use Factory registered widgets

My goal here was actually to try to be as close to python as possible and do the least of amount of magic under the hood so I didn't want to add e.g. automatic keywords, like app or factory. I was even hesitating adding the `^` syntax for canvas, rather than making the explicit option the only option, but then added it for practical reasons. but I figured if it's needed, people would import `Factory` at the top of their python file and use it from there. In fact as you'll see further, I do think all the styling should happen in a separate file, where they can just import stuff at the top.

With regards to app specifically, I was hoping with the proposed syntax, people will do it the more traditional python way, and in `__init__`, they would do e.g. `self.app = App.get_running_app()`, so in kv rules they could just then do e.g. `self.x @= self.app.x`. Or even better, because they could mix it with python, they may not need a reference to `app` at all because there would not be need for quick access to a global `app` singleton. But whatever works!

> it look harder to get to the point of writing the actual rule of a widget (needed import, function declaration, context)

That is true, but if you look at the examples at the end of the proposals, of how button and label look, for widgets with more rules, it's not that much extra. But also, that was a little the point for me, to make things more explicit so you have more control over the order of things and more in line with general python. I wanted this to be as pythonic as I could make it.

> everything is now flattened, and splitted. The KV lang allow to instantiate and configure the widget in a Tree logic. Going back to Python, you declare first, then assign rules later

This one, though does not really have to be that way. That's just how i wrote the examples, but it does not have to be written that way at all because it's python so you can literlaly write it anyway you like. I do agree though that KV provides a nice visual way describing the widget tree, which this doesn't provide. But, consider how I may re-write the label example:

```python
@KV()
def apply_kv(self):
    with self.canvas:
        with KVCtx():
            self.state_image @= self.background_normal if self.state == 'normal' else self.background_down
            self.disabled_image @= self.background_disabled_normal if self.state == 'normal' else self.background_disabled_down

            color = Color()
            color.rgba @= self.background_color

            border = BorderImage()
            border.border @= self.border
            border.pos @= self.pos
            bordersize @= self.size
            border.source @= self.disabled_image if self.disabled else self.state_image

            Color(1, 1, 1, 1)
            rect = Rectangle()
            rect.texture @= self.texture
            rect.size @= self.texture_size
            rect.pos @= int(self.center_x - self.texture_size[0] / 2.), int(self.center_y - self.texture_size[1] / 2.)
```

or you can even write it like so:
```python
@KV()
def apply_kv(self):
    with KVCtx():
        self.state_image @= self.background_normal if self.state == 'normal' else self.background_down
        self.disabled_image @= self.background_disabled_normal if self.state == 'normal' else self.background_disabled_down

        with self.canvas:
            color = Color()
        color.rgba @= self.background_color

        with self.canvas:
            border = BorderImage()
        border.border @= self.border
        border.pos @= self.pos
        bordersize @= self.size
        border.source @= self.disabled_image if self.disabled else self.state_image

        with self.canvas:
            Color(1, 1, 1, 1)
            rect = Rectangle()
        rect.texture @= self.texture
        rect.size @= self.texture_size
        rect.pos @= int(self.center_x - self.texture_size[0] / 2.), int(self.center_y - self.texture_size[1] / 2.)
```

> It raises also some question like:
> how to do plug new rules on a existing widget (like reset and rewrite new rule for Button)?
is it possible to separate the rules from the widget declaration itself?

So this is actually the thing I like most about the proposed syntax. There's nothing in the compiled function that says it must be part of a widget. In fact, you can write a top level function like this:
```python
@KV()
def apply_kv(widget, widget2):
    with KVCtx():
        widget.width @= widget.height
```
and then call it as follows:
```python
>>> widget, widget2 = Widget(), Widget()
>>> apply_kv(widget, widget2)
```

In other words, how do you add new rules to an existing widget? As follows:
```python
class MyWidget(Widget):
    @KV()
    def apply_kv(self):
        with KVCtx():
            ...

class MyWidgetInherited(MyWidget):
    @KV()
    def apply_kv(self):
        super(MyWidgetInherited, self).apply_kv()
        # make new bindings
        with KVCtx():
            ...
```
or
```python
class MyWidget(Widget):
    @KV()
    def apply_kv(self):
        with KVCtx():
            ...

class MyWidgetInherited(MyWidget):
    @KV()
    def apply_kv2(self):
        self.apply_kv()
        # make new bindings
        with KVCtx():
            ...
```

Basically, you can selectively call whatever function you want, with whatever args and have that apply the rules with the proposed syntax. But this is what I'm excited about, because we can create themed widgets that way. Create a widget like normal, but on a separate, but associated python file, almost like we do with a a side by side .kv file, we create a style for it or multiple styles, and the end user decides which style they want to use, by co-inheriting from the main widget, and the associated style class. Or they can inherit and then just overwrite the style function, and implement their own. The options are limitless and you can use the full power of python inheritance patterns etc.

> One of the usage done today is that it's possible to compile all your app/python with Cython, and you pack it with the data. Using AST at the core prevent all possible move with Cython.

This is obviously the biggest remaining issue I will need to work out. I will have to look into how cython deals with decorators and perhaps we can provide a way to include all the compiled files in the cython exe, so that it can load it from there. Is there any other sourceless distributions you're aware of?

Also, in case it's not clear, the AST stuff happens only the first time the function is compiled, once compiled, the compiled function is just loaded instead from the `__kvcache__` directory and no AST is involved, only the inspect module in order to find the original location where the function was declared so we can find its associated `__kvcache__`. So really, it's just a matter at that point of providing a standard place from where to load the substitute compiled function, if the source is not available - it could probably even be from a zip file, as python lets you import modules from a compressed zip (because compiled files are larger also, naturally).

> I'm all in if KV lang stay. If your end goal is to remove it at some point, i don't think it is a good idea. Yes it is more explicit and now possible to do things we can't on KV lang, but i feel it is on the basic stuff a step back. KV lang have limitation, but make it easier to write basics. So i guess it depend of your goal, i can go on the specific of your proposal and implementation if it's complementary :)

So as I said, I do think it should be complementary, and I would not remove traditional KV syntax. However, I think we should say that traditional KV syntax will not be updated (i.e. we can close issues that request anything that isn't currently within KV, or request changes to KV) to allow more complex pattens and that we let KV do what it does best - clear and simple rules describing a widget tree. Anything beyond that, e.g. if you want more complex bindings, or whatever, you should switch to the proposed syntax rather than trying to stuff it into KV. In fact if I could, I would remove the ability to rebind within KV because that already added more complexity to native KV than I think was good for it. native KV shines best, when it clearly and cleanly describes a widget tree.

However, for the proposal to be useful with regards to theming and customization, I think all widgets within Kivy itself, e.g `Button`, should make use of the new proposed syntax rather than native KV so that (1) it's faster, but (2) more importantly, so that we can use the theming strategies I explained above. I.e. by standardizing a `apply_kv` function name for each widget, we can provide a default theme, but let end users mix it up by just inheriting and overwriting the `apply_kv` function.

In fact, rather than `syle.kv`, we can have a `style.py` file, where each widget's style is described. E.g. you may have a `Label` inherit from `LabelStyle`, defined in `style.py`, and `LabelStyle` inherit from some common base `Style` class. And we can offer multiple style classes, or whatever we choose to do.

Not directly related, but I did want to point out something performance-wise: the more rules you can stuff into a `KVCtx`, the faster the compiled code will be because it can batch more rules and reduce binding redundancy.

matham:
@Cheaterman said, 
> Only one thing I'm not a fan of, is using a decorator AND a context manager
> Like, if I could get the decorator to implicitly give me a KVCtx for my whole function it would be nice

I started out with an implicit context, but then dropped it because that was too magical. Also, I wanted to make it super clear syntactically when the bindings happen - and it's when the context exits. If that wasn't there, users may forget where in the function bindings occur and that could lead to mistakes. So I think we should always be required to have KV rules under an explicit KV context. I.e overwriting variables required for binding will raise a parser error, if it;s done before the bindings, i.e. before the context exits. With an implicit context, that would not be clear.

There's also the issue of how do you give the user access to an implicit context? I thought of passing it as the first variable to the function, but then that felt too hacky because then we have to make sure the user provided a variable for it, name it properly and etc. so I decided not to do that but require explicit contexts.

tito:
The way you could bind together to unrelated widget is appealing: what about memory management? If both are deleted but a rule exist between them, will that prevent both of them to be collected?

In your examples with the multiple `with self.canvas`, the repetition of `color.PROP` to do the rule assignement, imo, the proposal doesn't look simple as it is much more explicit, if you think about all the stuff you need to import just to write a rule, this is not helping to keep things easy to read.

What your proposal does it giving tools to overcome some of limitations of the languages. And because it's AST implemented, it doesn't have impact on the current code that's great! Saying the KV lang is feature-freeze because of this proposal, no :(

About extending a Button, i still don't really see how i can add and/or overwrite the current rules declared in a Button, without creating my own class, and working for all the next instanciation. aka:

```
<Button>:
    text_size: self.width - dp(10), None
```

You talked about performance in execution. What about the parsing at startup? I tried to read the code, but didn't really understood what was saved in the kvc and if it's redoing an AST parse.

Because it's AST based, i doubt it would work also if the app is packaged like in a zip, or packaged in onefile with pyinstaller. Do you think that your implementation could work with "dis" ? Or working in a such way would require a compilation of the into a kvc prior to the packaging, and then put the __kvcache__ as part of the package?

About conditionnal, if you wrote:
```
def apply_kv(self):
    with KVCtx():
        self.height @= self.width + 45
        if self.height > 500:
                self.x @= self.anotherwidget.x
```
You told that this is not something that would work. The correct way is:
```
def apply_kv(self):
    with KVCtx():
        self.height @= self.width + 45
        if self.height > 500:
                self.x @= self.anotherwidget.x if self.height > 500 else self.x
```

Since you're working on the AST level, would that be possible not to execute a rule if a condition is not met? And then prevent any assignation? Cause things could be then more easier:
```
def apply_kv(self):
    with KVCtx():
        if self.state == 'normal':
            self.state_image @= self.background_normal  # this rule is bound but with a cond (self.state == "normal")
        else
            self.state_image @= self.background_down # this rule is bound but with a cond (not (self.state == "normal"))
```

In general, doing pure KV in Python but with limitation about what we can do, can be even more confuse at the beginning. KV lang is explicit in a way you can't really write python except expression. And few rules like no id for canvas element, so you can't reference them.
I'm still puzzled about pure-python syntax and AST conversion, and wonder if KV lang could be translated to your approach to benefit of implementation? Cause it look like now there is 2 implementations. This is not helping to keep things consistent and up to date :)

Generally, in your previous attempt at the KV compiler (big subject, and same level of complexity), what was wrong with it that put you in a state you wanted to redo KV lang in pure Python? Could the KV lang compiler be now a KV -> pure python implementation, then using your AST transformer?

KeyWeeUsr:
Repasting from [irc#kivy-dev](https://botbot.me/freenode/kivy-dev/):

@KeyWeeUsr
> Great work matham! I'm a lil bit picky about totally unimportant stuff, but could you please change Ctx names to Context if it's a symbol's name and not just a variable? Perhaps even try to run the code with pylint and check if the names are correct pascalcase, snakecase, etc ^^

@matham
> regarding KVCtx vs KVContext, yes I agree that's terse, but I assumed most people would prefer to write the shorter name. But I absolutely don'y mind changing it to be longer if that's what you'll think is ok
I could just see the objection ahead of time about having to type it out, so I shortened it
but peeps, please comment on the actual issue if you have questions or comments so that we can keep track of it more easily :)

matham:
Previous KV Compiler
---------------------

> Generally, in your previous attempt at the KV compiler (big subject, and same level of complexity), what was wrong with it that put you in a state you wanted to redo KV lang in pure Python? Could the KV lang compiler be now a KV -> pure python implementation, then using your AST transformer?

I suppose I should explain how I got here. When I initially worked on the [compiler](https://github.com/kivy/kivy/pull/3456/) after your initial implementation, my main motivation was performance. That's just because at the time, that is what I found most interesting to work on. So I reimplemented the compiler, gotten to the state where it generated bindings and had a mostly working system.

Exiting KV issues
-----------------

Except, I immediately ran into decisions about the language that I had to make. Parsing the binding part was easy (that was mostly done by the KV parser), the hard part was deciding where and when to bind. We have a bunch of rules, and the question was how can we batch them to optimize them, otherwise, the compiler wouldn't be particularly usefule. You have to worry about initialization, then binding and then making sure all the rules are up to date, without constant re-triggering of rules. Then you need to worry about how to handle rules that overwrite each other because of inheritance and be able to deal with clearing existing inherited rules, garbage collection, and a bunch of other stuff. It turned into a mess of arbitrary decisions, and I was also not getting feedback, so I stopped.

But the whole time, it anyway felt hacky to me, because it didn't resolve any of the longstanding issues encountered in KV. I felt that any compiler worth its salt will need to provide ways of resolving those issues. Specifically:

* Variable initialization was always a fight between KV and python. E.g. previously KV would always overwrite variables passed in python. Even now, it tries to not do that and respect the value passed from python, but it doesn't always.
* Similarly, we have no control over widget creation and the exact stages of rule application from within python, for widgets that are created in KV. This is often most encountered with the `ids` problem, where newbies quickly learn that `ids` are not available until the widgets is fully constructed and you typically have to delay everything by a clock cycle, if the widget is created in KV. This makes it clear that there needs to be a way to hook into the rule application system.
* Control over individual rules and the ability to disable them. Currently, you can clear the rules of a class overall fine. Then we added the ability to clear a specific property value, but not really - inherited rules are still applied, but then cleared.
* The exact order that rules are applied is not specified in the language spec so it's never clear to end users. This often means that can run into problems where a variable is None, because the order that the rules are applied is unclear and there's no way to force a rule to be applied before another (for someone not knowing the internals). This generally means that there are a lot of `val: 55 if self.widget is None else self.widget.x` type rules to be on the safe side.
* The above issue, is a main reason why rebind became necessary. Although I was the original proponent of this, I have since come to view it as a problem, because it violates the action at a distance rule. Specifically, when you create a property, you shouldn't have to decide right there if rebind is should be True, because at that point, there's no KV involved. The place to decide whether a property should be rebind, is at the location where the property is bound in the KV rule. For example, should `parent` be set to `rebind=True`? It's impossible to answer this, because this decision should not be made there when the property is declared, but at the place it is used.
* This leads to the next issue, how to customize rules. Specifically it was [desired](https://github.com/kivy/kivy/wiki/KEP001:-Instantiate-things-other-than-widgets-from-kv) to customize properties created in KV, but there was no clear way to do that.
* Of course the biggest issue was theming, which in order to support, we'd probably need to solve all of the above issues.
* It also bothered me that KV bindings occurred externally to python in a way that cannot be easily documented. And in particular, by looking at a Widget's python code, you cannot tell what happens in response to what, which again violates the action at a distance principle. You have to keep switching to KV to be able to tell what method is triggered by what code, and what it does in response.

Now, non of these are deal breakers and these are not issues that one encounters in simple KV programs, because they don't really need any of the above, but it definitely becomes a problem as one gets involved deeper. Sure there are workarounds, but it at times feels like you're fighting it, rather that it working with you.

The death of the compiler
---------------------------

This all led me to stop working on the compiler, because I didn't see how that resolved any of this and I started caring about these kinds of issues more than just optimizing the code speed. My last iteration of the compiler did provide a clear spec for how the rules are applied, but it was mostly a post-hoc explanation of how KV was implemented, rather than a spec that was designed ahead of time. I felt there were far too many special cases in KV, which led to unexpected behavior and so I stopped. Periodically, I would think about it, but could never come up with a way to resolve all of these issues.

Evolution
------------------

Since then my motivations has changed a lot. While I do enjoying optimizations, I have come to value correct design decisions as much more important, especially in API and language design. I have spent a lot of time on the python mailing lists observing how a language is designed and I feel like that has changed my views a lot about potential language pitfalls and what is important. I've also since encountered trio, which had immensely impressed me with its design choices, in particular the idea of using syntactical constructs to point out where important things happen in the code, while otherwise, getting out of the way.

So, when it occurred to me to use a decorator with `AST` manipulations, I realized how it could solve pretty much all the issues I mentioned about traditional KV. MY motivation was not at all for performance, although that of course was interesting, it was primarily about how to fully spec out a version of the KV language with little ambiguity that fit into python as much as possible with the least amount of **magical** differences between normal python. In other words, my concern was primarily about integrating it into python so that it was syntactically clear to users when magic was happening, but otherwise being as close to normal python as possible allowing the full usage of typical python constructs without getting in the way.

As I mentioned, parsing the bindings with AST is the easiest part, and I in fact completed that first within a couple of days and that code is pretty much unchanged still (it's the first commit). The part that took the most time, was figuring out the simplest syntax, which satisfied my stated design goals, including all the issues I mentioned earlier. I feel like I achieved that. A major reason is that by simply remaining with python's legal syntax, we inherit all the nice design decisions by the python team - we just add a little spice. But, it did come at the cost of splitting from traditional KV and integrate into python.

Conclusion
------------

I hope this clarifies my motivations and why I undertook this project. And as I said multiple times, I think the traditional KV syntax is great for specifying a GUI structure and for dealing with simple and clean rules. But for anything beyond that, I think the proposed syntax should be used instead.

As for whether this compiler could be used for compiling traditional KV. I hope it's clear from the above that the proposed compiler's primary innovation is not simply for creating bindings, but doing it within the context of python given all the constraints. That's where most of the work occurred. So, while the easy part of AST parsing can be mostly reused, that component doesn't solve the difficult issues listed above. The hard part will be figuring out how and when to do the bindings in an efficient way, given the KV language, and that is still unresolved.

matham:
@tito, regarding your question of how to use this compiler to compile traditional KV. After talking today, I was thinking more on that and consider such a KV rule:
```
<MyWidget@Widget>:
    name: 'whatever'
    x: self.height + 55
    Label:
        text: label.text
    Label:
        id: label
        text: root.name
```

We can easily write a compiler that converts this to the following:
```python
class MyWidget(Widget):
    @KV()
    def apply_kv(root):
        super(MyWidget, root).apply_kv()
        __label0 = Factory.Label()
        label = Factory.Label()

        with KVCtx():
            root.name @= 'whatever'
            root.x @= root.height + 55
            __label0.text @= label.text
            label.text @= root.name
```

If we were to do this, this would satisfy most of my issues with KV not having an exact specification, because now KV would have an equivalent python transformation, and I think the python syntax is well specified, so therefore, KV would be well specified. In other words, we'd be able to specify exactly the order in which rules are executed.

There are a number of problems here, which is similar to what I faced when writing the original compiler though:
* **Rule negation.** I.e.
  ```
  <...>:
      -height: self.height
  ```
  we would need to add some additional code that somehow knows what properties something is bound to and unbind that. Hmm, maybe we can do that by naming the rules and accessing it from the `KVCtx` that is stored somewhere.
  Change `Widget` to contain the following property `kv_rules = DictProperty({})`, or do `kv_rules = {}` and in `Widget`'s `__init__` do `self.kv_rules = {}`. I prefer the former so that the dict is created ahead of time and we don't have to worry about the order of calling super by subclasses. Anyway, we have a base rule:
  ```
  <MyWidget@Widget>:
      x: self.y
  ```
  which is transformed into
  ```python
      @KV()
      def apply_kv(root):
          super(MyWidget, root).apply_kv()
          with KVCtx() as root.kv_rules['MyWidget']:
              with KVRule(name='x'):
                  root.x @= root.y
  ```
  Then, we may have:
  ```
  <OtherWidget@MyWidget>:
      -x: self.width
  ```
  which is now transformed into
  ```python
  class OtherWidget(Factory.MyWidget):
      @KV()
      def apply_kv(root):
          super(MyWidget, root).apply_kv()

          for __previous_ctx_rules in root.kv_rules.values().named_rules:
              if 'x' in __previous_ctx_rules:
                  __previous_ctx_rules['x'].unbind_rule()

          with KVCtx() as root.kv_rules['OtherWidget']:
              with KVRule(name='x'):
                  root.x @= root.y
  ```
  The proposed compiler already creates the rules and adds them to `named_rules` if a rule is named. So this would just be making use of that feature.

  But this would only work for root rules, what about rules defined on children widgets? Should we just outlaw them? Alternatively, we can make a `KVCtx` for each child and store it in the child - that should work well.

  The only problem, again, is that the rule is still applied first, before clearing it. But that isn't different than current KV. However, if we use the compiler, there would be no way in the future to change this because the compiler has no provision for pooling rules across the inheritance hierarchy. We can also consider dropping individual rule negation - I believe that's a recent addition that I may have added, so maybe we don't even need it.
* **Whole rule negation.** I.e. consider:
  ```
  <-MyWidget@SomeWidget>:
      ...
  ```
  that's not something we have to worry about though, because the builder takes care of this and will not apply previous rules if we hit such an instruction. But, I think i'd prefer that in `apply_kv`, we just don't call `super` if that is in the rule deceleration. That will prevent the previous rules from being applied at all and is more in line with how it's done in python (also, the bulder rule search is slowwwwww).
* **Property creation.** The fact that we need to check every rule `x: self.y`, whether `x` is a Kivy property and create if not kills performance and adds complexity to any compiler. I suggest we change that and require that if a Kivy property doesn't exist and you want to create it, you specify it syntactically. I.e. if `x` is not a kivy property and you want it to make one for you, you have to write e.g. `!x: self.y` or some other symbol. Then we can transform the following:
  ```
  <MyWidget@Widget>:
      !x: 55
      !z: 'name'
      !val: self.height
      other_thing: self.object
  ```
  into:
  ```python
      x = NumericProperty(55)
      z = StringProperty('name')
      val = ObjectProperty(None)
      @KV()
      def apply_kv(root):
          super(MyWidget, root).apply_kv()
          with KVCtx() as root.kv_rules['MyWidget']:
              with KVRule(name='x'):
                  root.x @= 55
              ...
  ```
* **Rebind** Now, rebind is a problem because the compiler doesn't at all look at each property whether to rebind, but treats it as a global variable for the whole set of rules being compiled. Our options are to run rebind ON or OFF for all the objects. We can also add a syntactical way of marking that an object should be rebind using the glob syntax used in the compiler. E.g.
  ```
  <MyWidget@Widget>:
      #:rebind *object, self.some_widget, ...
      height: self.width
      other_thing: self.object.size
  ```
* **Proxying** Here's another situation that the compiler treats things differently. We say which widgets are long lived and make sure that those widgets don't hold direct references. So e.g. for `x: app.height`, we would need some way to mark app as being long lived and that it should not hold a direct reference to the widget containing the rule.

  Our options are: turn on proxying for all the widgets. But then, the bindings will just die because no one will hold on to it. Except, because we save the rules into the context stored in the root widget's `kv_rules`, the root widget will keep everything alive. So as long as the root widget is alive, the rules stay alive. I think this is a pretty good solution.
* **Overwriting defaults** The only major thing, is overwriting default rules. That I cannot see of a clean way of dealing with it. I.e. the following situation:
  ```
  <Label>:
      x: self.y
  ```
  I can see a two options:
  * Make that illegal, and instead break style.kv into smaller parts, and make them optional where a rule from style.kv is loaded only if a substitute is not provided (registered?), so we'd never need to overwrite defaults, as you could just specify alternate rules instead. I would be in favor of this, because that makes things more consistent for me.
  * Make the builder compile such rules as just functions, that it applies to the the Widget when built. This will add messiness because instead of having a single `apply_kv` function that gets attached to each widget and that people can overwrite from python and call super if they want or whatever, now the builder is doing things again in the background. If not for this option, there would be no need for the builder at all, because creating a widget would just call `apply_kv` and that creates all the rules. We should discuss more about this though (we can maybe add this builder call into `apply_kv`, at the end).
* All other things, should work out of the box with the compiler.

If ya'll think this is a good idea, I can write up a full proposal stating KV as a transformation into python with clearly applied rules and order of execution with no ambiguities. For me, this would basically satisfy most problems I see and would make me feel comfortable about the completeness and specification of KV. In particular, naming rules and having a standard `apply_kv`, would provide the access from python I would have liked to have, which would now be specified.

matham:
On IRC @tito said:
> matham: funny, rule negation is not existing on KV, but the discussion from yesterday make me believe it would nice to have :D
as if you declare a new rule on a prop that have already an cond expr, both would be always executed
matham: because you think per file, and yes it would be possible but not optimized
but, same as linking multiple .o into a executable, you may add one step of link to do this: if you know about everything, you have the power to optimize the best way

If we had individual rule negation where the previous rules that apply to a property is removed, including for children widgets in a rule, but even if we restricted rule negation to only be allowed on the root widget of a rule, there would be no reason for the compiler. Obviously, as I mentioned in a previous post you can make it work by applying the rule to a prop, and in a later rule class where this prop is negated we can unbind previous rules applied to the prop. But, the rule would have been applied at some point.

Rule negation with compiler
-------------------------------

Imagine what would be needed to make the compiler be able to clear a rule so that is never applied, if some subclass somewhere decides to negate the rule. Before every rule application, the compiler will have to check for some flag to see whether the rule should be applied. This would involve searching through some structure, so that only rules applied before the negation, but not after are cleared. Ignoring the performance, per-rule issues that this would introduce, it would also mean we cannot batch bind anymore because each rule may or may not be bound to. And without that, there's really no advantage to a compiler, as any code it generates would probably be slower than application of rules the current way which would be able to pool all rules before applying it and do it in a loop.

As for the suggestion that the compiler somehow will be able to link everything together, that would only work if the compiler, at the compilation stage had access to all the rules anywhere the widget participates in. Not only that, it would also need to know about rules applied to the widget when the widget is a child, because there could be a negation there. Furthermore, that means we would need to generate per situation rules and will need to replicate all the rules everywhere because if someone negates something somewhere, we now need to generate a different set of rules for it and any widget that inherits. That's why there's no linking stage that could link everything together because they are not independent. So the idea of linking I don't think applies here at all. All of this just adds complexity.

Furthermore, in KV, there's no linking stage when we know all the places a style.kv exists and a rule is declared there - that info is only known at runtime, meaning we would need to recompile everything at runtime. And if you do know all the files, or place everything in one file, why are we using rule negation in the first place? Just don't declare the rule in the root widget so it won't need to be overwritten.

Why have rule negation?
---------------------------

I'm not even sure why individual rule negation is useful - I have never actually used that. Or if it's useful, why we cannot live with the original application and then unbinding. If we are splitting up style.kv and making it optional, why would you really need to overwrite a rule so that it's never applied? You'd only need to do that if you don't control the original source.

So perhaps we should just not do a compiler and add rule negation by pooling all the rules and not applying the ones that are cleared - as suggested on IRC. But I think such a change would cause a bunch of unintended consequences. The fact we have no official kv model spec, but rely instead on the on the implementation to make the spec means that if we just suddenly change the implementation, things that worked before may not work anymore for no clear reason. I'll list some ways in which pooling rules, as suggested may break things:
* Say a parent defined a rule `some_prop: 55` and then in an inherited widget `-some_prop: None`. If `some_prop` is a KV created prop, what is the property created? A NumericProperty or a ObjectProperty?
* But this leads to the most obvious issue: what if the parent rule assumes some value for the prop that the rule sets, yet in a child widget it is overwritten by something unexpected? Like what if the parent widget actually uses the prop value in another rule, and now all of a sudden, the prop doesn't even exist anymore when the parent rules are applied because it was cleared by a child? And if you think negation would not be used without knowing the expectations of the parent rule, in that case, we're back to why have negation in the first place?
* Previously, the assumption was that each sequence of rule decelerations, when done will have applied all the rules. But all of sudden that won't be the case anymore and as you said on IRC, rules and behaviors go together, so clearing an individual rule may break the behavior. So in my view, individual rule negation should not be supported.

But this is the kind of issues I'm hoping to avoid in the future by using the compiler as a spec. Because otherwise, we'd go just into KV's implementation, change a few things and now the whole spec has changed (given that the implementation is the spec) without no one really seeing all the consequences. But, with the compiler's python transformation, all of a sudden there's a simple execution model, which lets us reason more easily what breaks or changes the spec given some proposed changes to KV because there's a python model to refer to. Of course, the compiler is not the only way to do that - we can do it by having comprehensive tests or something else. But, this proposal is about the compiler.

inclement:
I've been quite busy and have only just seen this, although I have now read most of the original post and discussion. My initial reactions are:

- In general, this is amazing. I'm 100% behind the idea of using ast manipulation to achieve kv benefits in pure python. Actually, I've thought for a while that if the python gui ecosystem were very competitive (like that of js, say) then there would undoubtedly be a cool new kivy-like library that does things this way.

- Just reading the examples, I find the code quite obtuse and hard to follow. I'm sure this is at least partly because it's new to me, and it doesn't really put me off. 

- The new methodology seems to be a great replacements for kv's flexibility (discussion about the syntax details notwithstanding), but the examples do seem to mostly lose the clarity of kv's simply-indented rule-based semantics. I'm finding it much harder to follow at a glance, more comparable to how it's hard to read pure-python gui code in kivy as it currently stands, even if the new code is more concise and clearer line by line. Again, I'm not sure how much this is because it's new to me, or if in practice you could structure things to work out fine.

Overall, I think this is a good direction, and I'll be interested to try some examples for myself to understand it better when I have the time.

matham:
@inclement thanks for reviewing!

One of the changes to the PR since I began was the idea of using the compiler as a spec for traditional KV, see two comments above yours for a description of that idea. That would allow us to use both traditional KV with all the benefits of a clear spec and compiler as well as the pure python version.

Regarding the clarity of python vs KV, I do agree that KV is much clearer, especially for simpler rules. But I also think we can probably come up with some conventions of writing the `apply_kv` function that would make things clearer, maybe. 

But one reason for the lack of visual clarity is that you need all the `self.xxx = `, rather than the KV colon syntax. Also there's of course the visual parent tree.

I had initially though of doing with widgets, what we do with graphics. I.e. so that you could do in python:

```py
@KV()
def apply_kv(self):
    with KVCtx():
        with MyWidget() as w:
            w.width @= w.height
            with Label() as label:
                label.text @= str(w.x)
```

The idea was that `Label` would now automatically be a added as a child to `MyWidget`. This is easily doable and can be done completely independent of the compiler. The problem though is that you quickly run out of horizontal space in the 80 char limit, so I'm not sure if this is suited for python code. But even worse, unless we have special syntax there, `MyWidget` would now automatically add any widget created, to itself, while `MyWidget` held the global context. So in some subfunction you could be creating a widget, which is now added to `MyWidget` automatically until the `MyWidget` exits.

I could not think of any way to syntactically indicate that a widget should be added to the parent.

matham:
Another thought is that instead of having each rule being written out e.g. as `self.width @= self.height` we could do `width @= self.height` and assume that `width` is set for the last context manager. E.g.
```py
@KV()
def apply_kv(self):
    with KVCtx():
        with MyWidget() as w:
            width @= w.height
            with Label() as label:
                text @= str(w.x)
```

But there are multiple issues here:
* Not only will `with KVCtx()` and `with KVRule()` be magical, now all `with` statements will be magical because the compiler will have to know which widget the rule is assigned to at compile time. 
* I find the confusion between local and rule variables to be too much with this syntax.
* It's also too magical and unexpected for writing it in python itself.

So I think that since this is python, it's not unreasonable to require that we write out each rule as `self.val @= self.other_val` and now allow skipping the first `self`.

Regarding the bigger issue of using `with` as a way to denote the children tree, this would work only if we can syntactically indicate in the `with` statement that the `with` is special. For `KVCtx` and `KVRule`, I literally use it by checking the name of the object in the `with` statement, but this is already dirty enough and will not work anyway for widgets which can have any name.

I cannot think of anything clean that is allowed by python syntactically that would indicate to us that the `with` object is a widget and needs to be added to its parent. Any ideas? Also, by using `with` for both `KVCtx` and `KVRule` and for widgets, it feels weird and I'd rather reserve `with` for the former rather than latter.

At some point I would try to mock up a fuller example example to provide a more practical example.

matham:
I now converted `filechooser.py` to use the proposed syntax. I'm attaching the diff and final `filechooser.py` file as a zip [filechooser.zip](https://github.com/kivy/kivy/files/2258423/filechooser.zip) as well as in [gist form](https://gist.github.com/matham/f04b5af6b763c406491ea84d60b0400d) so it's easier to preview.

This should give you a feel for how the proposed syntax would feel. You can apply the diff to the compiler branch and it'll work. There are some issues with the icon filechooser, but that seems to be present in master as well so I didn't bother fixing it.

Other than the syntax difference, the conversion, as you can see was pretty much copy and paste without any surprises, so it should be pretty easy to go between the forms. There was one modification I added, which has been my plan with the previous compiler as well, is that when we do `Widget(parent=some_widget)`, the widgets gets automatically added as child of `some_widget`. This is just a convenience so we don't have to type out `add_widget` all the time.

Regarding the idea to use `with` statements to construct the parent tree and automatically add children that way, I also tried converting the `apply_kv` method from `FileChooserProgress` to use [that approach](https://gist.github.com/matham/2005a18d8c94c9edd5a6890eef4f4d61#file-new_proposed-py) and you can see how it compares to the [current proposed approach](https://gist.github.com/matham/2005a18d8c94c9edd5a6890eef4f4d61#file-current_proposed-py). To me, the current approach looks more cleaner. Yes, the parent tree is not as clearly delineated as with KV, but the alternative approach using `with` wastes a lot of white space and reduces overall clarity even more IMO.

FluxIX:
I'm encouraged to see this proposal as I always found the traditional KV language to have too many special cases and the order of evaluation always seemed to cause me trouble. I think the proposal is well laid-out and the proposed syntax is cleaner to use with less ambiguity then the traditional syntax. As I have some experience with AST generation from writing domain-specific languages, I will examine the code changes and see what comments I have regarding them.

matham:
@FluxIX Thanks, I really appreciate it.

Right now, I'm making a second pass over the code fixing things and adding comments etc. so that it would be more understandable, so you may want to hold off on a review until that's done, hopefully within the week. I'm also working on adding a visualization of the AST binding graphs using graphviz to help everyone better understand what is going on in the code.

PR #5876, opened 2018-07-26
[WIP] Make kivy.graphics.Callback available from within Kv language

gottadiveintopython:
This PR is progrees of #5711. I re-created the repository so I re-create the PR. Sorry I'm bad at git command.

Hi, currently `kivy.graphics.Callback` is not available from within KvLanguage. This PR allows you to do like this:

```python3
from kivy.lang import Builder
from kivy.app import App


KV_STRING = r'''
#:import random random
#:set MESSAGE 'Kivy'

BoxLayout:
    Widget:
        canvas:
            Callback:
                on_call:  # any name that starts with 'on_' is OK
                    print('-----------------------------------')
                    print(app)
                    print(self)
                    print(root)
                    print(args)
                    print(cm)
                    print(random.random())
                    print(MESSAGE)
                    print(label)  # Can't use other widget's id. That's why this PR is WIP.
    Label:
        id: label
        text: 'Hello Kivy'
'''



class TestApp(App):
    def build(self):
        return Builder.load_string(KV_STRING)

TestApp().run()
```

```text
-----------------------------------
<__main__.TestApp object at 0x7f1777effb40>
<kivy.uix.widget.Widget object at 0x7f177855d868>
<kivy.uix.boxlayout.BoxLayout object at 0x7f1777effba8>
(<kivy.graphics.instructions.Callback object at 0x7f97505ff228>,)
<function cm at 0x7f17851d8ea0>
0.5298193401683202
Kivy
   File "kivy/graphics/compiler.pyx", line 143, in kivy.graphics.compiler.GraphicsCompiler.compile
   File "kivy/graphics/instructions.pyx", line 482, in kivy.graphics.instructions.Callback.apply
   File "/tmp/firefox/kivy/kivy/lang/builder.py", line 65, in custom_callback
     exec(__kvlang__.co_value, idmap)
   File "<string>", line 12, in <module>
 NameError: name 'label' is not defined
 Exception ignored in: 'kivy.graphics.instructions.InstructionGroup.build'
   File "kivy/graphics/compiler.pyx", line 143, in kivy.graphics.compiler.GraphicsCompiler.compile
   File "kivy/graphics/instructions.pyx", line 482, in kivy.graphics.instructions.Callback.apply
   File "/tmp/firefox/kivy/kivy/lang/builder.py", line 65, in custom_callback
     exec(__kvlang__.co_value, idmap)
   File "<string>", line 12, in <module>
 NameError: name 'label' is not defined
```

#2636

### DIFF
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -24,6 +24,7 @@ from kivy.compat import PY2, iteritems, iterkeys
 from kivy.context import register_context
 from kivy.resources import resource_find
 from kivy._event import Observable, EventDispatcher
+from kivy.graphics import Callback as CallbackInstruction
 
 __all__ = ('Observable', 'Builder', 'BuilderBase', 'BuilderException')
 
@@ -828,6 +829,10 @@ class BuilderBase(object):
             if name == 'Clear':
                 canvas.clear()
                 continue
+            if name == 'Callback':
+                for handler in crule.handlers:
+                    _create_callback_instruction(widget, handler, idmap)
+                continue
             instr = Factory.get(name)()
             if not isinstance(instr, Instruction):
                 raise BuilderException(
@@ -916,3 +921,10 @@ if 'KIVY_PROFILE_LANG' in environ:
         print('Profiling written at builder_stats.html')
 
     atexit.register(dump_builder_stats)
+
+
+def _create_callback_instruction(widget, prule, idmap):
+    idmap = copy(idmap)
+    idmap.update(global_idmap)
+    idmap['self'] = widget.proxy_ref
+    return CallbackInstruction(partial(custom_callback, prule, idmap))


matham:
I think the original formulation is preferable. Canvas instructions don't have the normal KV syntax of widgets, so canvas instructions don't support the `on_xxx` syntax as well as being unable to have an `id`. It would be weird if that's still the case, except for the one canvas instruction `Callback` that does support `on_xxx` events. It would make it a special case. The original formulation of specifying the function as an attribute of the canvas instruction is more in line with how KV works for graphics stuff.

Also, this `on_xxx` syntax is completely different than the rest of KV, because it's not actually binding an event.

gottadiveintopython:
Thanks for the review. Yes, I understand what you mean.
The reason why using 'on_' is [this](https://github.com/kivy/kivy/blob/master/kivy/lang/parser.py#L171-L184).
The parser call `compile(...., 'exec')` or `compile(...., 'eval')` depending on the name of the property.

 The original formulation is `compile(...., 'eval')`, so I think it's not easy to use.

matham:
I think we may have got our wires crossed :) By original implementation I meant https://github.com/kivy/kivy/pull/5711/.

gottadiveintopython:
yep I understand `the original formulation` == #5711 .

I mean #5711  can't do like this:

```text
        Callback:
            func:
                print(1234)
                print(5678)
```

so it's not convenient.

tshirtman:
Ok, after discussing with @matham on kivy-dev, i understand this might actually be worse.

The idea was that passing code to kv for event is usually done using literal blocks of code, that are evaluated on event dispatching, as opposed to setting a property, which is code that is evaluated when properties in the expression change, and the result is assigned to the expression.

But canvas instructions are not event dispatchers, and this is not really an event, but an attribute that we are setting, which creates a special case that might be harder to reason about the differences between widgets and canvas instructions in kvlang. For example, you think the fact that `label` is not defined is an issue, while it's actually normal considering how canvas is managed.

Sorry for the additional work i asked from you, i think we can reopen #5711 instead.

gottadiveintopython:
Thanks @tshirtman as well. I've read the discussion.
I still think it's possible to make `widget's ids` available from within `Callback`, maybe because I don't fully understand `kivy/lang/*.py`.

>  i think we can reopen #5711 instead.

I can't reopen it, probabely because I re-created the repository as I stated. So I'm gonna create a new branch.

sorry for late reply

matham:
It's not about what can be physically done in the lang/ code, because yeah, you can definitely change it to make it available. It's more about about what the language supports and the language doesn't support callbacks for canvas instructions because canvas instructions do not support dispatching. At least not currently anyway.

gottadiveintopython:
I see, I might try it.

thanks @tshirtman @matham 
It's 04:09 AM here. I'm gonna sleep.

matham:
It would certainly be interesting to make canvas instructions be a first class object in KV, giving it e.g. an `id` and then perhaps allowing dispatching by inheriting from `EventDispatcher`. However, I'm not sure there's a need. 

Generally, canvas instructions are lower level gl instructions written in cython. Meaning, they are not like widgets that are used to build higher level objects used in GUI construction and therefore not used in situations that you would often need binding or dispatching. In fact, I don't think it has ever come up before the need to allow dispatching/binding on instruction properties.

So I'm not sure there would be much benefit to adding that functionality to instructions. On the other hand, it would certainly add a lot more complexity and performance issues to canvas instructions, because `EventDispatcher` is not a simple class to inherit from.

The one improvement that could be made, is to allow canvas instructions to be given an `id`, but, that also has issues. Right now, canvas instructions cannot have an `id` and when `self` is used, it references the containing widget. But if suddenly canvas instructions can also have an `id`, it would be confusing to people because now it can have a `id`, yet using `self` will still not reference the instruction but rather the widget that contains it. So I think it just may be best to leave canvas as is, because that way people just learn that canvas instructions are special and unlike widgets.

gottadiveintopython:
What I mean by `I might try it.` is just trying to make other widget's id available from within Callback. I'm not trying to make `issubclass(Canvas, EventDisparcher)` true. Because I dont think there is a need as you stated.

Sorry for confusing you by my bad English :-)

gottadiveintopython:
Sorry I totally misunderstood the conversation above.

I just realized I can't do this:

```text
BoxLayout:
    Widget:
        canvas:
            Rectangle:
                size: label.size
    Label:
        id: label
```

I thought I can do that. That's why I was saying "Make other widget's id available from within Callback". But now, I know it's normal so I'm not going to do such a thing.
Sorry for bothering you. @matham @tshirtman

matham:
Oh, no worries :)

PR #5878, opened 2018-07-28
Make kivy.graphics.instructions.Callback available from within Kv lan…

gottadiveintopython:
…guage

Hi, currently `kivy.graphics.instructions.Callback` is not available from within Kv language. This PR allows you to do like this:

```text
FloatLayout:
    canvas:
        Callback:
            callback: lambda __: print('test')
```

#2636 #5711 #5876


kivy/graphics/instructions.pyx
```diff
@@ -456,7 +456,7 @@ cdef class Callback(Instruction):
         regarding that, please contact us.
 
     '''
-    def __init__(self, arg, **kwargs):
+    def __init__(self, arg=None, **kwargs):
```
matham: The fact that arg is now optional means that later in `apply` we need to check if it's not None and only execute the function if not None (`if self.func is not None and self.func(self):`).

gottadiveintopython: yep and if `self.func` is None `flag_update_done()` still need to be called, doesn't it?  I mean
 
```python
        if self.func is None or self.func(self):
            self.flag_update_done()
```

gottadiveintopython: and if `self.func` is None doesn't need to reset context. Am I correct?

matham: I don't think so. Because currently, if the `func` returns Falsey, then it indicates that no graphics was updated, hence why it diesn't call update. Similarly, if there's no func in the first place, then the we should assume we don't need to update.

Regarding the context reset, it still occurs even if the `func` returns Falsey so I'm asuming it should still happen. But I have not used `Callback` myself so I'm not really familiar with its usage.

Also, given that `arg` is now a optional keyword arg, we should rename it to `callback` because it wouldn't make sense to call `Callback(arg=...)`. But @tshirtman can chime in with what he thinks about these things.

gottadiveintopython: sorry I forgot about unittest.
I'm going to make it now

matham: Hmm, ok my bad. I misread `self.flag_update_done()` as `self.flag_update()` so you're right, without a callback, the update probably should be immediately considered done. @tshirtman is more familiar with the graphics internals so he can confirm.

gottadiveintopython: It's ok :)

matham:
Looking good, @tshirtman wanne double check I didn't miss anything?

kivy/tests/test_graphics.py
```diff
@@ -109,3 +109,19 @@ def test_identity_creation(self):
...
+
+    def test_from_kv(self):
+        from kivy.lang import Builder
+
+        root = Builder.load_string("""
+Widget:
+    canvas:
+        Callback:
+            callback: lambda __: setattr(self, 'callback_test', 'TEST')
+""")
```
tshirtman: nit picking: you could use textwrap.dedent here to keep the indentation consistent here.

gottadiveintopython: thanks :) I've fixed that.

PR #5880, opened 2018-07-29
Added changes in build_config method of App class

naman653:


AndreMiras:
Two commits for just a `pass`? Also the title is not so descriptive.
Could you please squash and also be more descriptive why this change is really needed?

naman653:
My sincere apologies @AndreMiras . I am new to github and thanks for telling me about my mistake. Any other advice is appreciated.

AndreMiras:
C'mon no need to apology :hugs: 
Read about commit squashing. And also try to be as much descriptive as possible in your pull request. It's hard for other to understand why some changes are needed, so that's what a description is for.

PR #5907, opened 2018-08-21
Better #4752 fix

tshirtman:


kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -336,7 +332,7 @@ cdef class Line(VertexInstruction):
         cdef double pcx, pcy, px1, py1, px2, py2, px3, py3, px4, py4, pangle, pangle2
         cdef double w = self._width
         cdef double ix, iy
-        cdef unsigned int piv, pii2, piv2
+        cdef unsigned int piv, pii2, piv2, skip = 0
```
KeyWeeUsr: If you do `skip = 0` here (and I haven't seen any resetting later) you're basically skipping only a single iteration and then everything falls into `if i - skip > 0 and self._joint != LINE_JOINT_NONE:` or similar branches. Is it correct?

tshirtman: skip is incremented every time we encounter two identical points, (line 349), these branch then need to remove this offset to consider the number of actually drawn points. I think it's correct.

tshirtman:
@KeyWeeUsr can you have another look?

kivy/graphics/vertex_instructions_line.pxi
KeyWeeUsr: I wasn't sure how initializing `skip` to `0` in one place and never resetting it helped. Yeah, it works correctly after a second look i.e.:

```
first,i=0,skip=0,false
duplicate,i=1,skip=1,false
duplicate,i=2,skip=2,false
second,i=3,skip=2,true
```

Just a note, not resetting `skip` might result in a very unlikely overflow, though I guess we'd sooner run out of memory?

tshirtman: If that's the concern `skip` is never going to bigger than the maximum number of points in a line (unsigned int too), so if we have more points than allowed, we have a bigger problem than `skip` overflowing.

PR #5914, opened 2018-08-25
[WIP] GPIOInputProvider and DialBehavior for rotary knobs

ismailof:
The purpose of this PR is to discuss about the expansion of HIDInputProvider (or new provider) to handle REL events produced by a discrete knob (through the linux rotary-encoder driver).

There's an ongoing conversation on [kivy-dev mailing list](https://groups.google.com/forum/#!topic/kivy-dev/dg8HG2ap6dM) about how to better integrate it on current kivy input system.

Of course names of objects, properties and events are provisional.
I've added a use case into 'examples/demo/dialpong' which adds knob control to the popular pong tutorial example (knob with axis 'x' for player1 and knob with axis 'y' for player 2)

### DIFF
--- a/kivy/input/providers/gpioinput.py
+++ b/kivy/input/providers/gpioinput.py
@@ -1,0 +1,270 @@
+# coding utf-8
+from __future__ import print_function, division
+
+'''
+Support for GPIO input from the linux kernel using gpio devtree
+==================================================
+
+To configure GPIOInput, add this to your configuration::
+
+    [input]
+    # devicename = gpioinput,/dev/input/eventXX,[accel_time=xx(ms)]
+    # example with Rotary Encoder for volume
+    volume = gpioinput,/dev/input/event2
+
+.. note::
+    You must have read access to the input event.
+
+.. versionadded:: NOT RELEASED
+
+'''
+
+__all__ = ('GPIOEventProvider', 'GPIOEvent')
+
+import os
+from collections import defaultdict
+
+from kivy.input.motionevent import MotionEvent
+from kivy.input.shape import ShapeRect
+Window = None
+
+
+class GPIOEvent(MotionEvent):
+
+    def depack(self, args):
+        super(GPIOEvent, self).depack(args)
+        for k, v in args.items():
+            self.__dict__[k] = v
+        if getattr(self, 'is_rotary', False):
+            self.profile.append('dial')
+
+    def __str__(self):
+        return '<%s id=%d device=%s>' \
+            % (self.__class__.__name__, self.id, self.device)
+
+
+if 'KIVY_DOC' in os.environ:
+    # documentation hack
+    GPIOEventProvider = None
+
+else:
+    import threading
+    import collections
+    import struct
+    import fcntl
+    from kivy.input.provider import MotionEventProvider
+    from kivy.input.factory import MotionEventFactory
+    from kivy.logger import Logger
+
+    #
+    # This part is taken from linux-source-2.6.32/include/linux/input.h
+    #
+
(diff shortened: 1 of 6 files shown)


kivy/input/providers/gpioinput.py
```diff
@@ -0,0 +1,270 @@
...
+
+            try:
+                while True:
+                    me = self.dispatch_queue.pop(0)
+                    Window.dispatch('on_motion', 'begin', me)
+                    Logger.trace('GPIOEvent: dispatching %r' % me)
+            except IndexError:
+                pass
+
+    MotionEventFactory.register('gpioinput', GPIOEventProvider)
```
matham: As far as I can see, only the `register` needs to be under `if 'KIVY_DOC' in os.environ`. Can you please place only that under the if statement. Eventually we will find a better solution than conditionally defining things based on `KIVY_DOC`, so it would be best if the minimally amount of things would be under it.

tito:
Where is this "on_dial" event dispatched? It appears nowhere in the PR

ismailof:
@tito You're right. The "catching" code of the events is in the mixin class `DialBehavior`, which I had inserted in my app code. I've uploaded it to the code here.
The use of a mixin class or inserting directly into `Widget` is part of what's been discussed.

scottdmorgan:
Has this been tested on Python 3.7?

scottdmorgan:
I manually installed the changes on top of my installation of Kivy 1.11.1 on Python 3.7 running on a Raspberry Pi. 

When I try to run the Pong demo the first issue I get in __main__.py is that it cannot import DialBehavior. I had to change line 10 of that file to this:
from kivy.uix.behaviors.dialbehavior import DialBehavior
Any idea why?

After I do that I get this error on line 27 of dialbehavior.py:
dial_axis = OptionProperty(AXIS[0], options=AXIS)
TypeError: 'dict_values' object is not subscriptable

Any ideas what might be wrong?

ismailof:
> Has this been tested on Python 3.7?

TBH, not even in Python 3. Sorry I forgot about this PR long time.
Unluckily, I used to try this on a Raspberry Pi I no longer have set-up, and I don't when/if I'll have the time to be back on this.

scottdmorgan:
I have modified this code to work on Python 3.7.3. The only changes required are as follows due to changes in how dictionaries work in Python 3:
1. gpioinput.py - line 174: change options.iteritems() to options.items()
2. dialbehavior.py - line 24: change AXIS_COORDS.values() to list(AXIS_COORDS.values())

When running dialbehavior.py standalone using "sudo python dialbehavior.py" I had to comment out line 3 (from kivy.core.window import Window) and move that line just under line 89. Running it this way brings up a slider at the bottom of the screen that moves left/right when you rotate the rotary encoder.

scottdmorgan:
I only have one rotary encoder connected to my Raspberry Pi but I see in the dialpong example it uses two encoders. I added this line to the Kivy config.ini to automatically find identify which /dev/input/eventXX number is assigned to the rotary encoder:
rotary = probesysfs,match=rotary,provider=gpioinput,param=

scottdmorgan:
One more thing I did was to modify __main__.py in the dialpong example. I changed the import of DialBehavior from: "from dialbehavior import DialBehavior" to "from kivy.uix.behaviors.dialbehavior import DialBehavior". This didn't run for me without specifying the path.

Also, to run the test code in dialbehavior.py that puts a slider at the bottom of the screen, I had to move
"from kivy.core.window import Window" to just below "import kivy" at line 90. Just a tip if anyone wants to use that to test out the implementation.

PR #5919, opened 2018-08-27
Let the Inspector browse into WeakProxy'd widgets

LogicalDash:
When a widget has an ObjectProperty and it's filled with another widget in kv-lang, this actually results in a ``kivy.weakproxy.WeakProxy`` to the object being created. The inspector doesn't quite know how to handle those, and will crash if you try to browse them directly. This PR changes the behavior of the inspector so that instead of inspecting the ``WeakProxy``, it will inspect the underlying widget, although it will still show up as a ``WeakProxy`` in the ``TreeView``.

### DIFF
--- a/kivy/modules/inspector.py
+++ b/kivy/modules/inspector.py
@@ -82,6 +82,7 @@ from kivy.properties import ObjectProperty, BooleanProperty, ListProperty, \
 from kivy.graphics.texture import Texture
 from kivy.clock import Clock
 from kivy.lang import Builder
+from kivy.weakproxy import WeakProxy
 
 
 Builder.load_string('''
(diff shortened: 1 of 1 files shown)


kivy/modules/inspector.py
```diff
@@ -530,7 +531,10 @@ def show_widget_info(self):
         keys = list(widget.properties().keys())
         keys.sort()
         node = None
-        wk_widget = weakref.ref(widget)
+        if type(widget) is WeakProxy:
+            wk_widget = widget.__ref__
```
matham: Shouldn't this just be `wk_widget = widget`? Because it's dereferenced by calling it, the same as `__ref__` and `__ref__` relies on internal details.

LogicalDash: WeakProxy doesn't seem to be callable...?

matham: Opps, I was thinking of `WeakMethod`. My bad.

PR #5922, opened 2018-08-28
Updated camera_opencv.py to use reshape(-1) instead of tostring()

farleylai:
Replaced frame.tostring() with frame.reshape(-1) to remove unnecessary overhead.
The following simple benchmark shows the overhead of tostring()/tobytes() are significant compared with reshape(-1).
```python
from time import time
import cv2 

cap = cv2.VideoCapture(0)
ret, frame = cap.read()

tic = time()
for _ in range(1000):
    string = frame.tostring()

toc = time()
print(f"frame.tostring() takes {toc - tic:.3f}s")


tic = time()
for _ in range(1000):
    bytes = frame.tobytes()

toc = time()
print(f"frame.tobytes() takes {toc - tic:.3f}s")


tic = time()
for _ in range(1000):
    string = frame.reshape(-1)

toc = time()
print(f"frame.reshape(-1) takes {toc - tic:.3f}s")
```

On a 6-core machine (i7-6850K/Ubuntu 16.04), the output results are as follows:
```
frame.tostring() takes 0.088s
frame.tobytes() takes 0.059s
frame.reshape(-1) takes 0.001s
```

### DIFF
--- a/kivy/core/camera/camera_opencv.py
+++ b/kivy/core/camera/camera_opencv.py
@@ -138,9 +138,11 @@ class CameraOpenCV(CameraBase):
             try:
                 self._buffer = frame.imageData
             except AttributeError:
-                # On OSX there is no imageData attribute but a tostring()
+                # On OSX/Linux there is no imageData attribute but a tostring()
                 # method.
-                self._buffer = frame.tostring()
+                # frame is already of type ndarray which can be reshaped to 1-d.
+                # Either frame.tostring() or frame.tobytes() is slow.
+                self._buffer = frame.reshape(-1)
             self._copy_to_gpu()
         except:
             Logger.exception('OpenCV: Couldn\'t get image from Camera')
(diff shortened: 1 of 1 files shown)


kivy/core/camera/camera_opencv.py
```diff
@@ -138,9 +138,11 @@ def _update(self, dt):
             try:
                 self._buffer = frame.imageData
             except AttributeError:
-                # On OSX there is no imageData attribute but a tostring()
+                # On OSX/Linux there is no imageData attribute but a tostring()
                 # method.
-                self._buffer = frame.tostring()
+                # frame is already of type ndarray which can be reshaped to 1-d.
+                # Either frame.tostring() or frame.tobytes() is slow.
+                self._buffer = frame.reshape(-1)
```
tshirtman: If the comments about tostring/tobytes are not relevant anymore, just remove them, and instead document why reshape does what we want (which maybe is what 143 is about).

farleylai: All the three choices tostring/tobytes/reshape do what you what but reshape should be chosen over the other two for efficiency. It is relevant for people unfamiliar with Numpy but irrelevant otherwise. There might be a reason why tostring() is used initially, which is unfortunately not explicitly commented though.

PR #5925, opened 2018-08-30
Fix Doc 'Input Management'

gottadiveintopython:
Hi, the documentation says:

> The touch coordinate is not translated to your widget space because the touch is coming directly from the Window. It's your job to convert the coordinate to your local space.

but that's not correct.

https://github.com/kivy/kivy/blob/267da41decb49b1eb43f15050719b418581980f7/kivy/base.py#L254-L261

```python
from kivy.lang import Builder


class Grabber(Factory.Widget):

    def on_touch_down(self, touch):
        touch.grab(self)
        return True

    def on_touch_move(self, touch):
        if touch.grab_current is self:
            print('on_touch_move()', touch.pos, 'grab')
            return True
        else:
            print('on_touch_move()', touch.pos)

    def on_touch_up(self, touch):
        if touch.grab_current is self:
            print('on_touch_up()', touch.pos, 'grab')
            touch.ungrab(self)
            return True
        else:
            print('on_touch_up():', touch.pos)


BoxLayout:
    Widget:
    RelativeLayout:
        Grabber:
''')

runTouchApp(root)
```

```text
on_touch_move() (72.0, 345.0)
on_touch_move() (72.0, 345.0) grab
on_touch_move() (70.0, 343.99999999999994)
on_touch_move() (70.0, 343.99999999999994) grab
on_touch_up(): (70.0, 343.99999999999994)
on_touch_up() (70.0, 343.99999999999994) grab
```

The coordinates of grab-touch is same as non-grabbed-touch.

### DIFF
--- a/doc/sources/guide/inputs.rst
+++ b/doc/sources/guide/inputs.rst
@@ -241,9 +241,6 @@ are some limitations to grabbing:
     - You might receive an event with a grabbed touch, but not from you: it can be
       because the parent has sent the touch to its children while it was in
       the grabbed state.
-    - The touch coordinate is not translated to your widget space because the
-      touch is coming directly from the Window. It's your job to convert the
-      coordinate to your local space.
 
 Here is an example of how to use grabbing::
 


tshirtman:
It's weird indeed, because the code has been doing this since the initial import, and that documentation was added later, maybe it was buggy at that point, or it was just an overlook. I'm nearly sure I've had to manually manage offsets on grabbed touches at some point though. I'll agree on this removal, at best it's a simplistic and misleading bullet point, so it's not useful to have.

gottadiveintopython:
>  I'm nearly sure I've had to manually manage offsets on grabbed touches at some point though.

DragBehavior is the one of that that being in that kind of situation. Scheduling a callback function and dispatching a touch event inside that callback function.

https://github.com/kivy/kivy/blob/038acbf6ac324c8ebca6107185d0a736c21b2fca/kivy/uix/behaviors/drag.py#L220-L234

well I feel `_do_touch_up()` also might need to transform the coordinate manually.

https://github.com/kivy/kivy/blob/038acbf6ac324c8ebca6107185d0a736c21b2fca/kivy/uix/behaviors/drag.py#L208-L218

I'll test it out later.

tshirtman:
Ah, you are right, it might be becaus the touch had initially been grabbed from a different widget, dragging a widget from/to a relative layout and re-parenting it. It was some custom code though, not using kivy's DragBehavior, so i'm not saying this code is wrong, still good to check though.

Zen-CODE:
Thanks :-)

Issue #5443, opened 2017-10-19
Chain more than 2 animations doesn't work


20Tauri:
### Versions

* Python: 3.4
* OS: Vista
* Kivy: 1.10
* Kivy installation method: pip

### Description

When I chain 3 (or more) animations, I've got an exception raised in "animate.py (see below)

### Code and Logs

```python
    line 270, in have_properties_to_animate
    if props and props['properties']:
    TypeError: 'bool' object is not subscriptable
```

This is caused by
```
        self._widgets[widget.uid] = True
```

I suggest to override method "have_properties_to_animate()" for both Sequence and Parallel

```
    def have_properties_to_animate(self, widget):
        return (self.anim1.have_properties_to_animate(widget) or
                self.anim2.have_properties_to_animate(widget) )
```

dessant:
Please also share a runnable example.

20Tauri:
The most trivial one is:

```python
from kivy.animation import Animation, AnimationTransition
from kivy.uix.widget import Widget

w = Widget()
anim = Animation(size_hint_y=1, t="in_circ", duration=0.5)
anim += Animation(size_hint_y=0, t="out_circ", duration=1.2)
anim += Animation(size_hint_y=0.5, t="out_circ", duration=1.2)
anim.start(w)
anim.have_properties_to_animate(w)
```

gottadiveintopython:
@20Tauri 
I think you should create a PullRequest.
Your code(have_properties_to_animate()) seems good for me.


(This is not related to the issue) I'm a bit curious about [animated_properties](https://github.com/kivy/kivy/blob/master/kivy/animation.py#L154).

```python3
Animation(x=100).animated_properties  # => {'x': 100}
(Animation(x=100) + Animation(x=0)).animated_properties  # => {}
(Animation(x=100) & Animation(x=0)).animated_properties  # => {}
```
But it probabely fine as it is.

20Tauri:
Still not work with 1.11, please integrate the @gottadiveintopython 's fixes (https://github.com/kivy/kivy/pull/5926)

PR #5926, opened 2018-08-31
[base/animation]fix kivy.animation.Sequence and kivy.animation.Parallel

### LINKED ISSUE #5204: kivy.animation.Sequence sends "on_complete" event while animation is still in progress
### Versions

* Python: 2.7
* OS: Linux Ubuntu 16.04
* Kivy: 1.10.1.dev0
* Kivy installation method: git clone and python setup.py install

### Description

The program uses Animation to pulse a Widget color from black to red and back again.  When the animation completes, the Widget is supposed to be black, but it is red.  Could be a canvas update problem, could be an animation bug.

### Code and Logs

```python
from kivy.app import App
from kivy.graphics import Color, Line, Rectangle

class Blinker(Widget):

        super(Blinker, self).__init__(*args, **kwargs)
        self.color = None
        self.animation = None
        self.draw()
        self.bind(size=self.draw, pos=self.draw, parent=self.draw)

    def draw(self, *args):
        self.canvas.clear()
        with self.canvas:
            self.color = Color(0, 0, 0, 1)
            Rectangle(pos=self.pos, size=self.size)        

    def repaint(self, *args):
        if self.color is not None:
            self.color.rgba = (0, 0, 0, 1)
            self.canvas.ask_update()
            print 'color.rgba is', self.color.rgba

    def start(self, *args):
        if self.color is not None:
            self.animation = Animation(r=1, duration=0.5) + Animation(r=0, duration=0.5)
            self.animation.repeat = True
            self.animation.bind(on_complete=self.repaint)
            self.animation.start(self.color)

    def stop(self, *args):
        if self.animation is not None:
            self.animation.stop(self.color)
            self.animation = None


class A(App):

    def stop_start_animation(self, *args):
        if self.blinker.animation is not None:
            self.blinker.stop()
            self.button.text = "Start"
        else:
            self.blinker.start()
            self.button.text = "Stop"

    def build(self):
        top = BoxLayout(orientation="horizontal")
        self.button = (Button(on_press=self.stop_start_animation, text="Start"))
        top.add_widget(self.button)
        self.blinker = Blinker()
        top.add_widget(self.blinker)
        return top

if __name__ == "__main__":
    A().run()
```
Logs:
```
(testkivy) wjanssen@archie:/local/kivy-1.10$ python ~/projects/cbmx/platform/t.py
[INFO   ] [Logger      ] Record log in /home/wjanssen/.kivy/logs/kivy_17-05-31_34.txt
[WARNING] [Modules     ] Module <clipboard> not found
[WARNING] [Modules     ] Module <cutbuffer> not found
[INFO   ] [Factory     ] 194 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_sdl2, img_pil, img_gif (img_ffpyplayer ignored)
[INFO   ] [Text        ] Provider: sdl2
[INFO   ] [Kivy        ] v1.10.1.dev0, git-b7906e7, 20170531
[INFO   ] [Python      ] v2.7.13 |Continuum Analytics, Inc.| (default, Dec 20 2016, 23:09:15) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)]
[INFO   ] [GL          ] Backend used <gl>
[INFO   ] [GL          ] OpenGL version <3.0 Mesa 12.0.6>
[INFO   ] [GL          ] OpenGL vendor <Intel Open Source Technology Center>
[INFO   ] [GL          ] OpenGL renderer <Mesa DRI Intel(R) HD Graphics 5500 (Broadwell GT2) >
[INFO   ] [GL          ] OpenGL parsed version: 3, 0
[INFO   ] [GL          ] Shading version <1.30>
[INFO   ] [GL          ] Texture max size <8192>
[INFO   ] [Window      ] auto add sdl2 input provider
[WARNING] [Modules     ] Module <clipboard> not found
[WARNING] [Modules     ] Module <cutbuffer> not found
[INFO   ] [GL          ] NPOT texture support is available
color.rgba is [0.0, 0.0, 0.0, 1.0]
color.rgba is [0.0, 0.0, 0.0, 1.0]
color.rgba is [0.0, 0.0, 0.0, 1.0]
```

gottadiveintopython:
fix #5204 #5443 #5929

### DIFF
--- a/kivy/animation.py
+++ b/kivy/animation.py
@@ -391,7 +391,30 @@ class Animation(EventDispatcher):
         return Parallel(self, animation)
 
 
-class Sequence(Animation):
+class CompoundAnimation(Animation):
+
+    def stop(self, widget):
+        self.anim1.stop(widget)
+        self.anim2.stop(widget)
+        props = self._widgets.pop(widget.uid, None)
+        if props:
+            self.dispatch('on_complete', widget)
+        super(CompoundAnimation, self).cancel(widget)
+
+    def stop_property(self, widget, prop):
+        self.anim1.stop_property(widget, prop)
+        self.anim2.stop_property(widget, prop)
+        if (not self.anim1.have_properties_to_animate(widget) and
+                not self.anim2.have_properties_to_animate(widget)):
+            self.stop(widget)
+
+    def cancel(self, widget):
+        self.anim1.cancel(widget)
+        self.anim2.cancel(widget)
+        super(CompoundAnimation, self).cancel(widget)
+
+
+class Sequence(CompoundAnimation):
 
     def __init__(self, anim1, anim2):
         super(Sequence, self).__init__()
@@ -418,26 +441,6 @@ class Sequence(Animation):
         self._register()
         self.anim1.start(widget)
         self.anim1.bind(on_complete=self.on_anim1_complete)
-
-    def stop(self, widget):
-        self.anim1.stop(widget)
-        self.anim2.stop(widget)
-        props = self._widgets.pop(widget.uid, None)
-        if props:
-            self.dispatch('on_complete', widget)
-        super(Sequence, self).cancel(widget)
-
-    def stop_property(self, widget, prop):
-        self.anim1.stop_property(widget, prop)
-        self.anim2.stop_property(widget, prop)
-        if (not self.anim1.have_properties_to_animate(widget) and
-                not self.anim2.have_properties_to_animate(widget)):
-            self.stop(widget)
-
-    def cancel(self, widget):
-        self.anim1.cancel(widget)
-        self.anim2.cancel(widget)
-        super(Sequence, self).cancel(widget)
 
     def cancel_property(self, widget, prop):
         '''Even if an animation is running, remove a property. It will not be
@@ -481,7 +484,7 @@ class Sequence(Animation):
         self.dispatch('on_progress', widget, .5 + progress / 2.)
 
 
-class Parallel(Animation):
+class Parallel(CompoundAnimation):
 
     def __init__(self, anim1, anim2):
         super(Parallel, self).__init__()
@@ -502,26 +505,6 @@ class Parallel(Animation):
         self._widgets[widget.uid] = {'complete': 0}
         self._register()
         self.dispatch('on_start', widget)
-
-    def stop(self, widget):
-        self.anim1.stop(widget)
-        self.anim2.stop(widget)
-        props = self._widgets.pop(widget.uid, None)
-        if props:
-            self.dispatch('on_complete', widget)
-        super(Parallel, self).cancel(widget)
-
-    def stop_property(self, widget, prop):
-        self.anim1.stop_property(widget, prop)
-        self.anim2.stop_property(widget, prop)
-        if (not self.anim1.have_properties_to_animate(widget) and
-                not self.anim2.have_properties_to_animate(widget)):
-            self.stop(widget)
-
-    def cancel(self, widget):
-        self.anim1.cancel(widget)
-        self.anim2.cancel(widget)
-        super(Parallel, self).cancel(widget)
 
     def on_anim_complete(self, instance, widget):
         self._widgets[widget.uid]['complete'] += 1
--- a/kivy/animation.py
+++ b/kivy/animation.py
@@ -492,7 +492,6 @@ class Sequence(CompoundAnimation):
             return
         if self.repeat:
             self.anim1.start(widget)
-            self.anim1.bind(on_complete=self.on_anim1_complete)
         else:
             self.dispatch('on_complete', widget)
             self.cancel(widget)
@@ -518,8 +517,7 @@ class Parallel(CompoundAnimation):
     def stop(self, widget):
         self.anim1.stop(widget)
         self.anim2.stop(widget)
-        props = self._widgets.pop(widget.uid, None)
-        if props:
+        if self._widgets.pop(widget.uid, None):
             self.dispatch('on_complete', widget)
         super().cancel(widget)
 
--- a/kivy/animation.py
+++ b/kivy/animation.py
@@ -413,6 +413,22 @@ class CompoundAnimation(Animation):
         self.anim2.cancel(widget)
         super(CompoundAnimation, self).cancel(widget)
 
+    def cancel_property(self, widget, prop):
+        '''Even if an animation is running, remove a property. It will not be
+        animated further. If it was the only/last property being animated,
+        the animation will be canceled (see :attr:`cancel`)
+
+        This method overrides `:class:kivy.animation.Animation`'s
+        version, to cancel it on all animations of the Sequence.
+
+        .. versionadded:: 1.10.0
+        '''
+        self.anim1.cancel_property(widget, prop)
+        self.anim2.cancel_property(widget, prop)
+        if (not self.anim1.have_properties_to_animate(widget) and
+                not self.anim2.have_properties_to_animate(widget)):
+            self.cancel(widget)
+
     def have_properties_to_animate(self, widget):
         return (self.anim1.have_properties_to_animate(widget) or
                 self.anim2.have_properties_to_animate(widget))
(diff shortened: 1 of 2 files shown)


matham:
I wonder if `has_animation_properties` is a better name? And maybe it's better if its a `@property`?

matham:
Oh, Never mind, the function already exists in master. I'll let someone with animation usage review it.

gottadiveintopython:
yep it override `Animation.have_prop...()`
Thanks for the review.

tshirtman:
I meant for the Sequence and Parallel to have a common method, (but i'm not sure which meaning of "like" you used here, if you meant "in the same way as", then ok). i.e:

```
class CompoundAnimation(Animation):
    def have_property_to_animate(self, widget):
        …

class Sequence(CompoundAnimation):
    …

class Parallel(CompoundAnimation):
    …
```

i think is slightly better than having the exact same method in two classes.

gottadiveintopython:
ah I understand what you mean.

`stop()`, `stop_property()`, `cancel()`, `have_properties_to_animate()` are completely same. So moving them to `CompoundAnimation`, the base class. I feel that doing that loose some flexibility. But anyawys I'm doing that now.

gottadiveintopython:
`Sequence` and `Parallel` are really buggy while `Animation` is not.
If I continue fixing thoe bugs, this PR will become not related to the issue(#5443).
So I'm probably going to make another PR since the issue is fixed.

tshirtman:
Yes, there are certainly bugs left in them, thank you for looking into it.

gottadiveintopython:
np :)
well I updated the PR. Can you review it?

gottadiveintopython:
thanks

kivy/animation.py
```diff
@@ -437,24 +412,67 @@ def stop_property(self, widget, prop):
...
             self.cancel(widget)
 
+    def have_properties_to_animate(self, widget):
+        return (self.anim1.have_properties_to_animate(widget) or
+                self.anim2.have_properties_to_animate(widget))
+
+    @property
+    def transition(self):
+        # This property is impossible to implement
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
gottadiveintopython: Yep I agree with `adding a message to the exception`.
And about AttributeError, I can't decide because for me both looks fine. But according to the [reference](https://docs.python.org/3/library/exceptions.html#NotImplementedError), AttributeError might be better.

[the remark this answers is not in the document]
gottadiveintopython: Because `NotImplementedError` means `real implementation required`, which is not our case.
so yeah AttibuteError is better I think.

gottadiveintopython:
When I ran [this code](https://gist.github.com/gottadiveintopython/ac5e4e1a33a93c31bc53d6875e2b035c) on master branch, `Parallel`s were never released. I assume that's the reason why `Parallel` doesn't override `cancel_property()`. That means I should revert 3608092fcefb6ce949a5ae92a6c347569b95f7f8.

gottadiveintopython:
This PR contains this hackey code:

```python
old_max_fps = None


def setUpModule():
    global old_max_fps
    old_max_fps = Clock._max_fps
    Clock._max_fps = 0.0


def tearDownModule():
    global old_max_fps
    Clock._max_fps = old_max_fps
```

Does anyone know proper way of changing `maxfps` temporarily?

kivy/animation.py
```diff
@@ -437,29 +415,78 @@ def stop_property(self, widget, prop):
...
+
+    if PY2:
+        @property
+        def animated_properties(self):
+            d = dict(self.anim1.animated_properties)
+            d.update(self.anim2.animated_properties)
+            return d
+    else:
+        @property
+        def animated_properties(self):
```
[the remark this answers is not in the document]
gottadiveintopython: ok I'll fix that

gottadiveintopython:
ok I'll make this PR more simple.
um should I reset commits? I mean `git reset --hard/soft xxxx`.
Since you already spent a time on reading this PR, I better keep current commits?

tshirtman:
It's ok to leave the commits, we can always squash & merge in the interface, if we decide the history is not important in the end. Thank you, just let us know when you are happy about the state, so i can review again :).

gottadiveintopython:
ok I got it.
I might reset commits to make this pr easier to understand though.

gottadiveintopython:
This should be ok. Can you review it?

>  it also contains tests that are expected to fail, with comments about things that are currently unclear, so i'm not sure what to make of it

sorry for that unreadable test code.
I created simple code: [this](https://gist.github.com/gottadiveintopython/cc23a1ecaa6aa5ffd8a2f60445478eaf) and [this](https://gist.github.com/gottadiveintopython/7605ac5e2d0eaebbc68f124619717087)
Those are not fixed by this PR and I'm not even sure that those should be fixed or not.

tshirtman:
Hm, I'm really sorry this lingered for so long, but the code looks good, it does still feel worth merging, especially people still hits the bugs with reusing composed animations from time to time.

tshirtman:
Hm, I'm really sorry this lingered for so long, but the code looks good, it does still feel worth merging, especially people still hits the bugs with reusing composed animations from time to time.

gottadiveintopython:
@tshirtman Hi, is it better to `git rebase master` and rewrite unit tests in pytest-style? or better to keep the current state?

tshirtman:
Hey, if you are willing to, rebasing/updating would be great.

gottadiveintopython:
ok, I'll do it then.

gottadiveintopython:
`Parallel` never fires `on_progress`, and I think it's corrent behavior. But it might confuse the user, so we may need to document it.

kivy/animation.py
```diff
@@ -455,11 +421,64 @@ def cancel_property(self, widget, prop):
...
+        return (self.anim1.have_properties_to_animate(widget) or
+                self.anim2.have_properties_to_animate(widget))
+
+    @property
+    def animated_properties(self):
+        # TODO: after kivy drops python2 support, this code can be optimized.
+        #       for example: using ChainMap
+        d = dict(self.anim1.animated_properties)
+        d.update(self.anim2.animated_properties)
+        return d
```
gottadiveintopython, line 434: Might be better to not use `ChainMap`. Because if we re-implement it like this:

```python
    @property
    def animated_properties(self):
        return ChainMap(self.anim2.animated_properties, self.anim1.animated_properties)
```

the user will be able to write `anim2.animated_properties`.

tshirtman: in both case it's an object that is returned to the user, that they can write to, but that doesn't mean they can write to animation.animated_properties, unless the property defines a setter.

tshirtman:
> `Parallel` never fires `on_progress`, and I think it's a corrent behavior. But it might confuse the user, so we may need to document it.
> 
> Or maybe `Parallel` should fire `on_progress` only when `Parallel.anim1` fires `on_progress`.

hm, if we can't then it's better not to document, i don't see how only firing for anim1 would be correct, ideally it would fire for both, and have the progress value take both durations into account (so if anim 1 is 3s and anim 2 is 2s, 1s into the first animation, we'd get progress at 0.2, 1/5th of the total), but i didn't check how hard it would be to implement that.

kivy/animation.py
gottadiveintopython: Oh sorry, I meant this part

> Lookups search the underlying mappings successively until a key is found. In contrast, writes, updates, and deletions only operate on the first mapping.

in the [doc](https://docs.python.org/3/library/collections.html#collections.ChainMap).

To prevent the user from writing to underlying dictionaries(`anim1.animated_properties` and `anim2.animated_properties`), we need to impelement it like either of

```python
@property
def animated_properties(self):
    return {**self.anim1.animated_properties, **self.anim2.animated_properties, }
```

or 

```python
@property
def animated_properties(self):
    return ChainMap({}, self.anim2.animated_properties, self.anim1.animated_properties)
```

tshirtman: ok, that makes sense indeed. I guess the second one is a bit more efficient, as it doesn't unnecessarily copy data around, while still allowing to update the values for this animation (as they'll be saved in the first dict, they'll override values for the same key in the other dicts), and also they'll reflect updates to the animated properties of the underlying animations, in case they happen (and unless an overriding value has been put in the first dict, of course).

gottadiveintopython:
> we'd get progress at 0.2, 1/5th of the total

Sounds like you are confusing `Sequence` with `Parallel`. But anyways It may be better to keep the current implementation. If we change it to fire `on_progress`, it probably impact the performance, because `on_progress` usually is fired very frequently.

> if we can't then it's better not to document

Ok, got it.

tshirtman:
> > we'd get progress at 0.2, 1/5th of the total
> 
> Sounds like you are confusing `Sequence` with `Parallel`. But anyways It may be better to keep the current implementation. If we change it to fire `on_progress`, it probably impact the performance, because `on_progress` usually is fired very frequently.
> 

:facepalm: you are right, maybe we could report based on the longest, but yeah, in any case, no need to think about that in this PR, it would be a new feature imho, to be evaluated independently.

> > if we can't then it's better not to document
> 
> Ok, got it.
:+1:

gottadiveintopython:
`EventCounter.assert_()` is my concern.

gottadiveintopython:
@tshirtman 
I believe the PR is mergeable. Could you check if it's ok or not?

> we could report based on the longest

Yeah I agree with that way if we implement it.

gottadiveintopython:
@tshirtman I mean Could you approve the PR.

tshirtman:
sorry i missed the update (again). I'll approve now.

PR #5927, opened 2018-09-02
Repair power consumption on android

nkiiiiid:
set play to False Can not reduce power consumption，i found self._camera.stop() doesn't release the camera object，the camera object still running in the background and consume a lot of power，change stop() to _release_camera()  can repair it.

### DIFF
--- a/kivy/uix/camera.py
+++ b/kivy/uix/camera.py
@@ -116,4 +116,7 @@ class Camera(Image):
         if value:
             self._camera.start()
         else:
-            self._camera._release_camera() 
+            if hasattr(self._camera,"_release_camera")
+                self._camera._release_camera()
+            else:
+                self._camera.stop()
--- a/kivy/uix/camera.py
+++ b/kivy/uix/camera.py
@@ -116,4 +116,4 @@ class Camera(Image):
         if value:
             self._camera.start()
         else:
-            self._camera.stop()
+            self._camera._release_camera() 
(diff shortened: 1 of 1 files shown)


kivy/uix/camera.py
```diff
@@ -116,4 +116,7 @@ def on_play(self, instance, value):
         if value:
             self._camera.start()
         else:
-            self._camera.stop()
+            if hasattr(self._camera,"_release_camera"):
```
matham: This should happen in the android camera implementation, rather than here. The `hasattr` is basically checking whether the camera is the android camera, which is not appropriate here at this high API level.

Also, can you play again after release the camera? If not, this change seems wrong. Finally, maybe we just need an explicit API to release the underlying camera which you can call directly from your code?

nkiiiiid: thanks for reply，i ignore the play attribute is used to switch status of camera，and i hasn't test play again after release the camera. but the power consument  exist on android after  use camera.stop() to release camera. as you say we just need an explicit API to release the underlying camera , so i hope the official can consider to provide a api to do it thus make kivy more easy for newbies to use . by the way,  my english is poor ,  many thanks.

linuxrootok: Hey guys, I have some words. @matham ,at first,I just wanna get a way to release the camera on my pixel2, because if it cannt be closed, it can't  save power consumption, and the energy lost pretty fast. However,as you said,if  change the code, the status of camera will no longer  be changed. 
In short, we can join another function e.g "on_xxx" for release camera,and "on_play" will works  well and we can fix these.
Best regards!

PR #5940, opened 2018-09-20
Set android camera to autofocus

Huyston:
This change makes the Camera widget on Android to autofocus. This avoids the blurry images that appear when using this widget.


Huyston:
I also changed the fragment shader to rotate the texture for android devices.

tito:
Except the pep8, looks good!

PR #5943, opened 2018-09-25
Fixed bounce

erm3nda:
Disabled bounce ability once paddle hits ball untill paddle doesn't collide.

I know that if you hit twice the ball very very quick two times it will bounce again, but its not likely to happen.

Anyway, this just solves the weird bounce behavior when you move up/down the paddle when hitting the ball. It's not intended to fix the whole bouncing stuff.

Regards.

### DIFF
--- a/examples/tutorials/pong/main.py
+++ b/examples/tutorials/pong/main.py
@@ -4,7 +4,7 @@ kivy.require('1.1.1')
 from kivy.app import App
 from kivy.uix.widget import Widget
 from kivy.properties import (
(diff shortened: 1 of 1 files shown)


erm3nda:
How do i "re-do" this to make travis try again?
The error says something about "No output". Seems is something not related with the code itself.

Zen-CODE:
Don't worry about the Travis fails: as you suspected, they are not because of your changes. But what is the exact issue this PR fixes? What is the exact "bounce" problem this fixes?

PR #5950, opened 2018-10-03
Added `smooth` property to `kivy.uix.scrollview.ScrollView`

JakubBlaha:
This property says whether smooth scrolling should be used or not when scrolling with the mouse-wheel.

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -443,11 +443,13 @@ class ScrollView(StencilView):
     '''
 
     smooth = BooleanProperty()
-    '''Whether smooth scrolling should be used or not when scrolling with the the mouse-wheel.
+    '''Whether smooth scrolling should be used or not when scrolling with the
+    the mouse-wheel.
 
     .. versionadded:: 1.11.0
 
-    :attr:`smooth` is a :class:`~kivy.properties.BooleanProperty` and defaults to True.
+    :attr:`smooth` is a :class:`~kivy.properties.BooleanProperty` and defaults
+    to True.
     '''
 
     # private, for internal use only
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -452,8 +452,8 @@ class ScrollView(StencilView):
 
     .. versionadded:: 1.11.0
 
-    :attr:`smooth_scroll_end` is a :class:`~kivy.properties.NumericProperty` and defaults
-    to None.
+    :attr:`smooth_scroll_end` is a :class:`~kivy.properties.NumericProperty`
+    and defaults to None.
     '''
 
     # private, for internal use only
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -442,6 +442,14 @@ class ScrollView(StencilView):
     defaults to ['content'].
     '''
 
+    smooth = BooleanProperty()
+    '''Whether smooth scrolling should be used or not when scrolling with the the mouse-wheel.
+
+    .. versionadded:: 1.11.0
+
+    :attr:`smooth` is a :class:`~kivy.properties.BooleanProperty` and defaults to True.
+    '''
+
     # private, for internal use only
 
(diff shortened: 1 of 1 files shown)


matham:
The velocity setting seems arbitrary. Generally smoothing would imply actively filtering the points with some filter.

JakubBlaha:
I am not sure what do you mean by that @matham . @tshirtman suggested me to change the velocity and position which I have done.

matham:
Sorry that I wasn't clearer. There are two issues here:
One, the number `10` seems arbitrary. I.e. why 10? It's not likely that 10 would translate the same way across different OSs and mice. So that should be user configurable.

Second, maybe you meant speed rather than smoothness? Because this just changes the speed, not the smoothness. To change the smoothness, you would need to run the distances/velocity through a smoothing filter, which isn't happening here.

So, if you did mean speed, then this should be globally configurable - i.e. a parameter that controls the mouse wheel speed, like `scroll_timeout` here https://github.com/kivy/kivy/blob/master/kivy/uix/scrollview.py#L145. It could be named e.g. `mousewheel_speed` and default to 1.

matham:
The `scroll_wheel_distance` tells you how far it should scroll. The `scroll_wheel_duration`, would specify the amount of time, and therefore speed it should take to get there. `scroll_wheel_distance` alone would not be enough to be able to set the velocity. 

Currently, you just multiply `scroll_wheel_distance` by 10 to get the velocity, but instead, you'd divide by `scroll_wheel_duration` if non-zero, otherwise it's set to zero. Then, you can set `scroll_wheel_duration` to `0.1` to get the same effect, but other people can customize it to their own mouse/screen.

I just mentioned the smoothing filter, because I wasn't sure what you were trying to achieve. To me, smooth, implies that you're smoothing the scrolling e.g. by averaging etc. But, given what you said, it's not really about smoothing but about controlling the speed, so nevermind that :)

tshirtman:
I did the suggestion to set velocity assuming it would be app-specific code, it's nice if it's an option others can use, but that indeed means giving some more thoughts about what flexibility will be required, and what a good cross-platform experience is going to be. If it adds inertia to the end of the last scroll motion, it does change the distance a single scroll will cover, so the name of the property should reflect that effect properly (smooth could mean a number of things, which could also affect the beginning of the motion, or what happens when user does multiple scroll events at an uneven rate, etc).

JakubBlaha:
So if I get it right, I only need to change the name of it? I don't think any platform issues could be present here.

matham:
I guess a name change and clarification about what specifically this does. Because now I'm also confused about what we're trying to achieve here :)

Can you maybe say in words exactly what effect you're going for? I.e. what the current behavior is and how your change will modify it? Sorry that this is getting so complicated. Also, maybe ping us on discord so we can talk it about more dynamically?

tshirtman:
The idea is that when you use a mousewheel, scrolling jumps directly from one position to the other, motion is instant at each scroll, the motion can end less abruptly by just setting the velocity to a fixed value at the same time as doing the jump, which gives a smoother settling to the new position, if you scroll multiple times, since the velocity is set to a fixed value, it only influences the effect of the last scroll.

matham:
Ok, that makes sense. Then yeah, I guess just name should be changed to be clearer. E.g. name it to `smooth_scroll_end = NumericProperty(None, allownone=True)` and in the docs suggest to use `10` as a value to smooth the scroll end (assuming I understood what you're trying to do). 

And if `None`, the default, it would not smooth, yet provide a way for users to customize the value. But feel free to choose a more descriptive name?

matham:
Looks great, but could you please also add a test for this? You can see https://github.com/kivy/kivy/blob/master/kivy/tests/test_graphics.py#L11 for inspiration.

JakubBlaha:
I don't really know what u wanna test ont this thing. I would say it's ok. To be honest, I don't want to write any test for it.

tshirtman:
closing since #6020 was merged.

PR #5954, opened 2018-10-04
Scale export to png

tshirtman:
fixes and superseed #5867


kivy/uix/widget.py
```diff
@@ -622,19 +622,29 @@ def export_to_png(self, filename, *args):
             extension in your filename.
 
         .. versionadded:: 1.9.0
+
+        :Parameters:
+            `filename`: str
+                The filename with which to save the png.
+            `scale`: float
```
tito: It missed the .. versionadded

Issue #4666, opened 2016-10-10
RecycleView should not use a global cache


ngoonee:
I'm getting weird errors in my application because I use many recycleviews with similar data structures (dict names etc match), and sometimes a 'wrong' cached view is picked up.

Problem happens within the same view, but is especially jarring when two different views start interfering with each other.

I'd like the cache to be owned by each instance of recycleview (is there really much need for multiple instances to access the same cache?), at least optionally. As for the picking up of 'wrong' views (yes I've checked to make sure dict values don't overlap) I'll just have to do more debugging, hard to identify the root cause.

matham:
I guess there's no real reason to make it global. That may also fix your view mixup.

ngoonee:
Unfortunately I do get view mixups even with only one recycleview, so I doubt it would fix it totally (may get better). Still investigating that, really.

ngoonee:
@matham are you the point person for Recycleview? Am imagining some changes which could be useful for my own project, and would like to run them by someone who can comment on whether its worth the effort to modify Recycleview itself rather than just subclassing for my own use (which I've already started doing). Especially in terms of roadmap for how recycleview will look like, feature set, etc.

matham:
See https://github.com/kivy/kivy/issues/5913.

Issue #5913, opened 2018-08-24
RecycleView steals data from neighbouring RecycleView


KeikiHekili:
### Versions

* Python: 2.7.14
* OS: OS X 10.11.6
* Kivy: 1.10.1
* Kivy installation method: pip

### Description

I have one RecycleView (RV1) above another (RV2), where based on the size of the elements, the RecycleView can't. I've noticed that if RV1 is scrolled to the bottom, and then RV2 is scrolled, RV2 is showing the elements that should be exclusive to RV1. The same is true if RV2 is scrolled to the bottom, and then RV1 is scrolled. In this case, RV1 shows RV2 data. In the below example, the data from RV1 is `['chocolate', 'vanilla', 'strawberry']` and the data for RV2 is `['apple', 'orange', 'banana']`. Scroll RV1 to the bottom and then start scrolling RV2, and bizarrely, RV2 will show `'apple'` and then `'vanilla'`, and then `'banana'`. 

### Code and Logs

```python
from kivy.base import runTouchApp
from kivy.lang import Builder

kv = '''
<Card@FloatLayout>:
    size_hint: None, None
    size: 250, 320
    myName: 'Chocolate'
    canvas.before:
        Color:
            rgba: 1, 1, 1, 1
        Rectangle:
            size: self.size
            pos: self.pos
    Label:
        size_hint: None, None
        size: 230, 25
        pos: self.parent.x+10, self.parent.y+295
        text: root.myName
        color: [0, 0, 0, 1]

BoxLayout:
    orientation: 'vertical'
    padding: 5
    spacing: 5
    FloatLayout:
        RecycleView:
            id: rv1
            size_hint: 25/79., 0.35
            pos_hint: {'x': 0.68, 'y': 0.5}
            viewclass: 'Card'
            data: [{'myName': i} for i in ['Chocolate', 'Vanilla', 'Strawberry']]
            RecycleBoxLayout:
                default_size: None, 320#dp(56)
                default_size_hint: 1, None
                size_hint_y: None
                height: self.minimum_height
                orientation: 'vertical'
                touch_deselect_last: False
        RecycleView:
            id: rv2
            size_hint: 25/79., 0.35
            pos_hint: {'x': 0.68, 'top': 0.495}
            viewclass: 'Card'
            data: [{'myName': i} for i in ['Apple', 'Orange', 'Banana']]
            RecycleBoxLayout:
                default_size: None, 320#dp(56)
                default_size_hint: 1, None
[...]

PR #5963, opened 2018-10-05
Fix 5913 recycle view steals data

CompassMentis:
Applied fix as described at https://github.com/kivy/kivy/issues/5913#issuecomment-416000967

Manually tested the fix - it does fix the issue - but did not create a regression test (couldn't find a sample, so would be too time consuming). Existing unit tests still work fine


CompassMentis:
Reversed as requested

#close #4666 
#ref #5963

Issue #5864, opened 2018-07-18
AsyncImage - Failed to load image


realslimshanky:
### Versions

* Python: v3.5.3
* OS: Debian GNU/Linux 9
* Kivy: v1.11.0.dev0, git-1828af3, 20180623
* Kivy installation method: `pipenv install --skip-lock git+http://github.com/kivy/kivy.git#egg=kivy-1.10.1`

### Description

I tried using `AsyncImage` to load `https://cdn-images-1.medium.com/max/800/1*RO3AhbppPUFJBZwt3gUW-A.png` which is failing with error `HTTP Error 403: Forbidden`

Possible Solution: Adding proper headers before making the request using `urllib` as being done [here](https://github.com/kivy/kivy/blob/master/kivy/loader.py#L317) in `loader` module

#### main.py
```
from kivy.config import Config
Config.set('kivy', 'log_level', 'debug')

from kivy.app import App

class TestApp(App):
    def build(self):
        self.load_kv('./main.kv')


if __name__ == '__main__':
    TestApp().run()

```

#### main.kv
```
BoxLayout
    AsyncImage
        source: 'https://cdn-images-1.medium.com/max/800/1*RO3AhbppPUFJBZwt3gUW-A.png'
        canvas:
            Color:
                rgb: [1, 1, 1]
            Rectangle:
                texture: self.texture
                size: self.width + 20, self.height + 20
                pos: self.x - 10, self.y - 10
```

#### python main.py (logs)
```
[INFO   ] [Logger      ] Record log in /home/realslimshanky/.kivy/logs/kivy_18-07-18_83.txt
[INFO   ] [Kivy        ] v1.11.0.dev0, git-0471549, 20180718
[INFO   ] [Python      ] v3.5.3 (default, Jan 19 2017, 14:11:04) 
[GCC 6.3.0 20170118]
[INFO   ] [Factory     ] 194 symbols loaded
[DEBUG  ] [Cache       ] register <kv.lang> with limit=None, timeout=None
[DEBUG  ] [Cache       ] register <kv.image> with limit=None, timeout=60
[DEBUG  ] [Cache       ] register <kv.atlas> with limit=None, timeout=None
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_sdl2, img_gif (img_pil, img_ffpyplayer ignored)
[DEBUG  ] [Cache       ] register <kv.texture> with limit=1000, timeout=60
[...]

PR #5964, opened 2018-10-05
Allow setting url agent for async image and urlrequest

tshirtman:
through kivy config

ref #5864


tshirtman:
need more testing.

Issue #5956, opened 2018-10-04
In Texture blit_buffer, cast from unsigned short (and int) don't work


maddyaby:
### Versions

* Python: 3.6
* OS: Windows
* Kivy: 1.10.1
* Kivy installation method: anaconda

### Description

I'm trying to blit a buffer to a texture with type uint16 / ushort / unsigned short. This results in an error on line 941 of kivy/graphics/texture.pyx. It seems as if the cast from unsigned short to short failures here:
```python
        cdef short [:] short_view
          ...
            elif glbufferfmt == GL_SHORT or glbufferfmt == GL_UNSIGNED_SHORT:
                short_view = pbuffer
```
I changed the source and rebuilt to handle the unsigned and signed cases separately with the proper C type and this seems to work.

The same problem seems to apply to the 32-bit int type in the lines following.

matham:
Would you like to submit a pull request?

maddyaby:
Sure! I can do that this weekend

matham:
https://github.com/kivy/kivy/pull/5980

PR #5980, opened 2018-10-08
Issue #5956: Fix casts in texture.blit_buffer for ushort and uint types.

maddyaby:
Issue: When trying to use ushort and uint types, the cast to short and int
ctypes fail and result in an exception.

Fix: Do a cast to unsigned short and unsigned int instead.

Testing: Tested with the same diff on the release version 1.10.1. Tested
blitting a buffer of type int16, uint16, int32 and uint32. All types now
work as expected with no exceptions.


matham:
That's perfect. Could you try adding a test case for this? See https://github.com/kivy/kivy/blob/master/kivy/tests/test_graphics.py#L11 for an example.

maddyaby:
Some things came up and I unfortunately don't have time to write a test case in any kind of timely manner. :(

PR #5982, opened 2018-10-09
Removed kivy.lib.osc from setup.py packages

misl6:
Fixes setup.py error
Related to: #5967


tshirtman:
Silly me, thank you for catching it!

PR #5983, opened 2018-10-09
Issue #5773: Ensure to dispatch on_transform_with_touch event when the angle change

### LINKED ISSUE #5773: Rotating Scatter does not dispatch on_transform_with_touch
### Versions

* Python 3.6.1 + Kivy 1.10 on Windows
* Python 3.6.4 + recent master on Linux

### Description

1. Rotating Scatter does not dispatch the on_transform_with_touch event

2. Translating Scatter with `do_translation=False` still dispatches the on_transform_with_touch event

### Code and Logs

```python
from kivy.lang import Builder

Scatter:
    size_hint: None, None
    pos: 200, 200
    do_scale: False
    do_translation: False
    on_transform_with_touch: print("!!! Transform")

    canvas:
        Color:
            rgba: 1, 0, 0, 1
        Rectangle:
            pos: 0, 0
            size: self.size
'''))
```

tito:
Fixes #5773

### DIFF
--- a/kivy/uix/scatter.py
+++ b/kivy/uix/scatter.py
@@ -476,6 +476,8 @@ class Scatter(Widget):
             return changed
 
         angle = radians(new_line.angle(old_line)) * self.do_rotation
+        if angle:
+            changed = True
         self.apply_transform(Matrix().rotate(angle, 0, 0, 1), anchor=anchor)
 
         if self.do_scale:


kivy/uix/scatter.py
```diff
@@ -476,6 +476,8 @@ def transform_with_touch(self, touch):
             return changed
 
         angle = radians(new_line.angle(old_line)) * self.do_rotation
+        if angle:
+            changed = True
```
matham: I'm not very familiar, but what if the previous angle was non-zero and now it is set to zero, shouldn't we still dispatch?

tito: Angle is more angle diff. The transformation are cumulative. If there is no angle, you just apply identity so there is no change so changed is False. If there is a rotation, then the transformation is marked as changed.

PR #5988, opened 2018-10-10
Removed cache print statements

Lauszus:
This was also done in: https://github.com/kivy/kivy/pull/4709/files, but was never merged.

### DIFF
--- a/kivy/cache.py
+++ b/kivy/cache.py
@@ -170,7 +170,7 @@ class Cache(object):
 
     @staticmethod
     def _purge_oldest(category, maxpurge=1):
-        print('PURGE', category)
+        Logger.debug('Cache: Remove oldest in %s' % category)
         import heapq
         time = Clock.get_time()
         heap_list = []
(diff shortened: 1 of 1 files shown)


kivy/cache.py
```diff
@@ -163,14 +163,18 @@ def remove(category, key=None):
...
                 Cache._objects[category] = {}
+                Logger.trace('Cache: Flushed category %s from cache' %
+                             category)
         except Exception:
             pass
 
     @staticmethod
     def _purge_oldest(category, maxpurge=1):
-        print('PURGE', category)
+        Logger.debug('Cache: Remove oldest in %s' % category)
```
matham: Shouldn't all of them be trace rather than debug?

Lauszus: I think that it makes sense from a developer perspective to use debug for this function and the ```_purge_by_timeout``` function, as it is useful to know when those two function are called when debugging the cache functionality without flooding the terminal with too much logging data.

However if the maintainers disagree, then I'll be happy to change it, as long the cache debugging is actually logged and not just printed to stdout.

PR #5997, opened 2018-10-17
add magical load_kv method

tshirtman:
This method looks up the name of the module that calls it, and load the
kv file with the same name but kv extension.

example usage:

in src/widgets/thing.py

    import load_kv

    class Thing(Widget):
        pass

    load_kv()

in src/widgets/thing.kv

    <Thing>:
        Label:
            text: 'magic!'

This also adds a packaging tool that will inline all such loaded kvlang
in the python code.

    python kivy/tools/packaging/bundle_kv.py -y

after this is called, all python code that contains the load_kv() call
will be edited *in place* to replace that call with
Builder.load_string("content of the kv file"), and the kv file *will be
deleted*, don't do this if you don't have uncommited code.

### DIFF
--- a/kivy/tools/packaging/bundle_kv.py
+++ b/kivy/tools/packaging/bundle_kv.py
@@ -1,0 +1,62 @@
+# after this is called, all python code that contains the load_kv() call
+# will be edited *in place* to replace that call with
+# Builder.load_string("content of the kv file"), and the kv file *will
+# be deleted*, don't do this if you don't have uncommited code.
+
+from tempfile import mkstemp
+from os.path import join
+from os import walk, unlink, close, write
+from shutil import move
+import sys
+
+if sys.version_major == 2:
+    input = raw_input
+
+
+if len(sys.argv) < 2 or sys.argv[1] not in ('-y', b'-y'):
+    answer = input(
+        'this will edit your python files in place, and remove your kv '
+        'files.\nAre you sure? [y(es)/N(o)]'
+    )
+    if answer.lower() not in ('y', 'yes'):
+        sys.exit()
+
+
+def get_kv_source(fn):
+    kv = fn[:-2] + 'kv'
+    with open(kv, encoding='utf8') as f:
+        source = f.read()
+        res = source.replace('\\', r'\\')
+    unlink(kv)
+    return res
+
+
+for root, dirnames, filenames in walk('src'):
+    for f in filenames:
+        fn = join(root, f)
+        if f.endswith('.py'):
+            tmp, tmpname = mkstemp()
+            with open(fn, encoding='utf8') as source:
+                found = False
+                for line in source:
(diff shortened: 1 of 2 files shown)


tshirtman:
These are fair points, and in fact, it reflects the fact that i basically extracted this from a pattern i have in a lot of my recent projects, because i assume it can be useful to others.

- Yes, the name could be better, and `Builder.load_mod_kv` is certainly an improvement. About __file__, i know i had issues with that at some point, iirc, that was after packaging with pyinstaller (and before i introduced the `bundle_kv` thing), that i think `__file__` had a different behavior. Or maybe it was because of a side effect of `condiment`, which i use a lot too, anyway, i should probably test that again.

- About bundle_kv.py, i do think it's fair to put it into tools/packaging, i did add the specific "-y" parameter/question just before doing the PR because i wanted to avoid people running it mindlessly and then realize their mistake. I do think it's a nice complement on the packaging side, because while you can cythonize cython code to hide it, kv code stays available on disk, this method automatically hides it too. Also, in platforms like android, having numerous files can really slow down app startup, and it can discourage people from segmenting their kvlang in multiple files like `load_kv` allows, for performance reasons, this way allows to have it both way, clean in development, and fast in production. 

About downsides of accumulating all kv files and loading them before hand, it doesn't necessarily work, because they might import classes in an order you didn't plan for, and create circular imports, which would mean your code works in development, and breaks after packaging. I also though there would be issues for defining kv rules for classes before they are defined, but apparently not (just tested), maybe that changed at some point, or maybe i never challenged this assumption.

If you think it's too brittle, we can improve it or move it elsewhere, but i do like that it goes hand in hand with the other method autoload kv.

matham:
I'm not completely opposed to `bundle_kv.py`, but I think if we merge it, it needs tests and very complete docs. Also, rather than just looking for `load_kv`, it would be nice if it also looks at `Builder.load_kv`, although I'm not completely sure about that. 

But also, it should adds a method to find all the kv files in the project and return the list, so that could be reused in packaging and used by the new code. And I think the code should also accept the directory or project name and look it up by importing  - right now it just uses `src`.

I'm still not convinced this belongs in a kivy, because I think the better alternative to recommend is that users load the kv file by importing them in main - automatically changing source code should be the last resort as a recommendation. I do that it and it works pretty well.

But, if others disagree, I'm fine with merging this if it's made more generic.

tshirtman:
Oh, right, i overlooked the `src` thing, when i copied it, i'll think that as well, (i agree with attaching it to `Builder` instead of the `utils` module).

tito:
I don't think this belong to Kivy itself either. Would it be better as a side project or the wiki?

tito:
I'm closing this PR. Either reopen if it's a full tool with doc for later support, and/or create an external tool that handle the use case

Issue #5984, opened 2018-10-09
Kivy requires Cython to be installed by the user in some situations, which makes depending on it as a package problematic


flaviusb:
### Versions

* Python: Python 3.6.5
* OS: Linux (4.12.12-gentoo x86_64)
* Kivy: 1.10.1
* Kivy installation method:  install_requires in setup.py
### Description

I am trying to create a package that uses kivy, that a user can install using pip, so that I can upload it to PyPI. This is a package that a user should be able to have as a dependency in their own package using the usual 'install_requires' method in setup.py, ideally using vanilla setuptools. At present, I have to install Cython separately in order to get Kivy to install, as Kivy does not correctly manage its dependencies.

### Code and Logs

After clearing out wheel cache, to replicate installing on a fresh system, and in the directory I am developing the package in:

```
$ virtualenv --no-site-packages install
Using base prefix '/usr'
New python executable in /home/flaviusb/code/rvit/install/bin/python3.6
Also creating executable in /home/flaviusb/code/rvit/install/bin/python
Installing setuptools, pip, wheel...done.
$ . install/bin/activate
(install) $ pip install -e .
Obtaining file:///home/flaviusb/code/rvit
Collecting kivy (from rvit==0.1)
  Using cached https://files.pythonhosted.org/packages/7d/8b/89d220b7f96dc2662b81319067f679b4cd73cda66f4aa850db5b6c6cfc7a/Kivy-1.10.1.tar.gz
    Complete output from command python setup.py egg_info:
    Using distutils
    
    Cython is missing, it's required for compiling kivy !
    
    
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-qf1pe7c9/kivy/setup.py", line 232, in <module>
        from Cython.Distutils import build_ext
    ModuleNotFoundError: No module named 'Cython'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-qf1pe7c9/kivy/


```

tshirtman:
Yes, i opened this one #5558 some time ago to try to find a solution to this, i tried to ship the code in the package to avoid the build time dependency on cython, but since the generated C code depends on the python version, we would need multiple source packages, and i didn't investigate yet if this was possible.

tito:
IMO, either there is a wheel so you don't need Cython, either you install source version therefore Cython is required? Trying to create a source version containing all potential python version seems counter-intuitive

[...]

PR #5998, opened 2018-10-18
Change check for Cython to attempt fallback to setuptools on supporte…

flaviusb:
…d platforms.

This attempts to solve Issue #5984 without causing more problems for Issue #5558.


matham:
It looks good, but am hesitant to merge without someone else looking it over because these parts of the code is so fragile and effects different systems that perhaps I'm not thinking of.

tito:
I'm wondering if that make Cython a requirements when using the wheels?

flaviusb:
@tito It should not do; as long as when you build the wheel, have_cython or skip_cython would be true. In this way, it is exactly like it was before.

matham:
I think @tito is right. This may force the installation of cython when
someone installs a kivy wheel file for e.g. windows. The only way yo test
is to build a wheel and install it with no cython installed.

flaviusb:
So, this does not force an install of cython when installing from a wheel, but it could definitely be put together better. Myself and a friend are working on a better version of this pull request that also clears up some of the fragile code, rather than just layering another set of checks on top.

matham:
Thanks.

If it's listed as a requirement for install, why doesn't it install cython
when it's a wheel just like e.g. it installs garden? Did you test making a
win wheel? Maybe it needs to be added to setup requirements instead?

flaviusb:
Because when you make a wheel, you install cython manually first, as is already the current practice for kivy. Then `have_cython` is `True` and `getting_cython` is `False`. So, `'cython'` is never added to `install_requires`.

flaviusb:
I tested making a Linux wheel, and can confirm that if you install cython manually and then build a wheel with this pull request, cython does not end up as a requirement for the generated wheel. There are some error messages generated though during the wheel generation process, and the whole thing is not ideal. So myself and a friend who is more of a python programmer than I am are working on a better version of this, which will hopefully be ready some time today or tomorrow.

flaviusb:
Right, we have managed to get rid of a lot of the redundant and fragile code that checked cython versions manually, and now generate bounds directly. We have tested building manually with cython manually installed (both versions that are allowed and versions that are disallowed), and without cython installed, and it does the right thing. The wheels generated do not depend on cython. All testing was on Linux, as we both only have linux machines available.

flaviusb:
Right, we have managed to get rid of a lot of the redundant and fragile code that checked cython versions manually, and now generate bounds directly. We have tested building manually with cython manually installed (both versions that are allowed and versions that are disallowed), and without cython installed, and it does the right thing. The wheels generated do not depend on cython. All testing was on Linux, as we both only have linux machines available.

flaviusb:
So, most of the travis builds actually build now. The only one that does not seems to be some kind of certificate/provisioning problem; the relevant error message is:
```
Download error on https://pypi.python.org/simple/cython/: [SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:645) -- Some packages may not be found!
```
I think it is out of scope of this patch to try to address that, though. Hopefully this will be able to be the final version of this pull request.

flaviusb:
One more commit, just adding a comment that clarifies some of the details of the changed KivyBuildExt initialisation. This should hopefully make the reasons for the changes more apparent in the code.

setup.py
```diff
@@ -14,17 +14,18 @@
 import os
 from os.path import join, dirname, sep, exists, basename, isdir
 from os import walk, environ
+from distutils.command.build_ext import build_ext
 from distutils.version import LooseVersion
 from distutils.sysconfig import get_python_inc
 from collections import OrderedDict
-from time import sleep, time
+from time import time
```
AndreMiras: do we still need the `time` import?

flaviusb: The `time` import is new, it was added an hour ago by someone else for epoch based timestamps on builds. I rebased to master only ten ish minutes ago.

flaviusb: Relevant pull request here: https://github.com/kivy/kivy/pull/6008

AndreMiras: Oh yes my bad I missed that `time()` call, thanks!

flaviusb:
>  Yes some of the tests are flaky I think. We should restart the failing build tasks manually to make sure everything is actually green

Ah, the way that particular build is provisioned needs to actually be changed in order to get it to be able to test. Previously it was 'green' because the tests to detect installing the wrong version of cython in setup.py were broken. But it is provisioned with the wrong version of cython, and it doesn't have updated certs or whatever are needed to be able to connect to pypi, which is correctly detected with the new setup.py.

flaviusb:
The CI for kivy has been consistently red for a while. This pull request cleans up some of that - comparing it to other pull requests we only have one Travis failure versus many, and it is clear that our one failure is from the preinstalled cython being 0.29 plus not being able to connect to pypi to get a supported version.

setup.py
```diff
@@ -14,17 +14,18 @@
...
 from subprocess import check_output, CalledProcessError
 from datetime import datetime
 
-if environ.get('KIVY_USE_SETUPTOOLS'):
```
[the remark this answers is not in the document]
flaviusb: Just checking, do you mean that that environment variable is used specifically for setting up a test matrix?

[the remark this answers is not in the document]
flaviusb: That is, if py4a uses it in the normal course of affairs, we don't need to check it - we can simply use the ImportError method to fallback to distutils, as this pull request does. If however you want to be able to deliberately turn off the use of setuptools in order to test that building using just distutils works, then we should have a `KIVY_USE_DISTUTILS` environment variable instead, and skip the attempt to import setuptools if that environment variable is set to `1`.

flaviusb:
> Watch out for stuff like this:
> 
> https://github.com/kivy/kivy-ios/blob/aa01292d81a02ca732d3cc3128bcef9b7823a521/recipes/kivy/__init__.py#L27
> 
> and make sure you check the recipe classes, so that nothing depends on Kivy setup.py env variable:
> 
> https://github.com/kivy/kivy-ios/blob/master/toolchain.py#L916
> 
> Alternatively in p4a:
> 
> https://github.com/kivy/python-for-android/blob/c12dbd693850f1124f6e96a22fe6c9fc30260728/pythonforandroid/recipes/kivy/__init__.py#L8
> 
> https://github.com/kivy/python-for-android/blob/c12dbd693850f1124f6e96a22fe6c9fc30260728/pythonforandroid/recipe.py#L946

Just doing a quick search over both repositories in github did not find the string KIVY_USE_SETUPTOOLS, so it looks like it isn't referenced in any recipies anymore.

setup.py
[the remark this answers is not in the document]
tito: From my searchs, only appveyor.yml uses it now. I don't think we need an env for forcing distutils either. Either you have setuptools and it's used prior distutils, either we fallback on distutils. LGTM

PR #6001, opened 2018-10-18
Add drag and drop behavior

matham:
Demo: https://www.youtube.com/watch?v=LQ4OUCckZqg

Drag and Drop
==============

This adds a drag and drop functionality to layouts and widgets. There are 2
primary and one internal component used to have drag and drop:

* The :class:`DraggableObjectBehavior`. Each widget that can be dragged needs
  to inherit from this. It has :attr:`DraggableObjectBehavior.drag_cls`, which
  is a string that indicates the :class:`DraggableLayoutBehavior` instance
  into which the widget can potentially be dropped.
* The :class:`DraggableLayoutBehavior`, which any layout that wants to accept
  drop behavior must inherit from. It has a
  :attr:`DraggableLayoutBehavior.drag_classes`, which determines the
  dragging widgets it accepts - their :attr:`DraggableObjectBehavior.drag_cls`
  must be listed in :attr:`DraggableLayoutBehavior.drag_classes`.
* Finally, :class:`DraggableController`which manages the drag and drop
  coordination. This is created automatically for each
  :class:`DraggableObjectBehavior.drag_controller` upon first drag, or it
  can be set by the user to a global controller to save memory.

Layout
-------

To use :class:`DraggableLayoutBehavior` with a layout,
:meth:`DraggableLayoutBehavior.compare_pos_to_widget` must be implement for each
layout. This determines where in the layout the widget is being previewed and
ultimitely dropped. But, it can be further customized for any type of layout
by overwriting the default
:meth:`DraggableLayoutBehavior.get_drop_insertion_index_move` and
:meth:`DraggableLayoutBehavior.get_drop_insertion_index_up`.

Similarly, :meth:`.handle_drag_release` should be implemented. It handles the
actual final drop into the layout. A good default implementation is::

    def handle_drag_release(self, index, drag_widget):
        self.add_widget(drag_widget, index)

Dragging Widget
-----------------

To use :class:`DraggableObjectBehavior`, one should implement
:meth:`DraggableObjectBehavior.initiate_drag` and/or
:meth:`DraggableObjectBehavior.complete_drag`. Typically, you would want to
remove the widget from it's current layout as it starts to be dragged, with
e.g. :

    def initiate_drag(self):
        self.parent.remove_widget(self)

Describing Which Widgets can be Dropped Into Which Layout
-------------------------------------------------------------

Given some :class:`DraggableObjectBehavior` and
some :class:`DraggableLayoutBehavior` instances. We can control which widget can
be dragged into which layout using their
:attr:`DraggableObjectBehavior.drag_cls` and
:attr:`DraggableLayoutBehavior.drag_classes`.
:attr:`DraggableObjectBehavior.drag_cls` must be listed in
:attr:`DraggableLayoutBehavior.drag_classes` for the widget to be draggable into
the layout.

Example
--------

Following is an example with :class:`kivy.uix.boxlayout.BoxLayout`:

    class DraggableBoxLayout(DraggableLayoutBehavior, BoxLayout):

        def compare_pos_to_widget(self, widget, pos):
            if self.orientation == 'vertical':
                return 'before' if pos[1] >= widget.center_y else 'after'
            return 'before' if pos[0] < widget.center_x else 'after'

        def handle_drag_release(self, index, drag_widget):
            self.add_widget(drag_widget, index)

    class DragLabel(DraggableObjectBehavior, Label):

        def initiate_drag(self):
            # during a drag, we remove the widget from the original location
            self.parent.remove_widget(self)

And then in kv:

    BoxLayout:
        DraggableBoxLayout:
            drag_classes: ['label']
            orientation: 'vertical'
            Label:
                text: 'A'
            Label:
                text: 'A'
            Label:
                text: 'A'
        DraggableBoxLayout:
            drag_classes: ['label']
            orientation: 'vertical'
            DragLabel:
                text: 'A*'
                drag_cls: 'label'
            DragLabel:
                text: 'A*'
                drag_cls: 'label'


tito:
I didn't check the code, but conceptually, you mixed 2 thing: the behavior (drag&drop) and one usage of it which is remove the object from one list and add it in another list.

This is one of the usage of the behavior, but not the only one. Think about dropping files to a bucket, categorize images just by drag&drop etc. Thoses cases require just the behavior and event, the rest is up to the user about what to do when the drop is accepted and happening.

It would be better to seperate the first behavior and have a class that implement one use case, which is the one you described: insert a widget at the right place when dropping on the receiver. `DraggableReceiverBehavior` ?

Also, i would move out the creation of the preview out of the controller to the `DraggableObjectBehavior` itself. I think it's best that each widget can have a different preview implementation rather than taking the decision to draw it in a Fbo. For example, if you have a "group" of widget or a selection that you want to accept drag&drop, you could customize the preview just by returning a instance of something else and not use the Fbo. And that instance could also react to some event like if it's the right place to drop or not.

If you seperate both, i think this is a very good initial implementation that would allow multiple other uses cases not tied to the one described here!

matham:
```
Conceptually, you mixed 2 thing: the behavior (drag&drop) and one usage of it which is remove the object from one list and add it in another list.
```

The code is actually agnostic of this particular case of removing from one and dropping in another. Take a look at the example https://github.com/kivy/kivy/blob/drag_n_drop/kivy/uix/behaviors/drag_n_drop.py#L679 and you'll see that this case needs to be implemented and is not the default.

More broadly, `DraggableObjectBehavior` receives a `initiate_drag` and `complete_drag` from the controller when a drag starts and finishes for the widget. By default it does **nothing**, but in the example that you saw, `initiate_drag` does `self.parent.remove_widget(self)` which removes it. But that is not the default.

Similarly, `DraggableLayoutBehavior` has a method `handle_drag_release` that needs to be overwritten and does **nothing** by default. In the example you saw it calls `self.add_widget(drag_widget, index)`. Its signature is `def handle_drag_release(self, index, drag_widget)`.

`index` is the value returned by `get_drop_insertion_index_move(x, y)` and `get_drop_insertion_index_up(x, y)`, which is called with the pos of the touch and returns the index in children where the widget/spacer would be dropped/shown.

These methods can and should be overwritten as needed, but by default under the hood they call `get_widget_under_drag` and `compare_pos_to_widget` to compute the index. This is the implementation for the example video that I showed:

```py
def compare_pos_to_widget(self, widget, pos):
    if self.orientation == 'vertical':
        return 'before' if pos[1] >= widget.center_y else 'after'
    return 'before' if pos[0] < widget.center_x else 'after'
```

This is how we know where to insert the widget/spacer. But by default, only the spacer would be shown, but the actual drop would do nothing.

Does this resolve the abstraction issue you mentioned? Do you think by default even the spacer should not be shown in the dropping layout? Maybe I misunderstood and you want to not have any `DraggableObjectBehavior` and just the `DraggableReceiverBehavior`? But then how would you track the widget dragged and dropped? Even when dropping files into a bucket, granted the spacer behavior would be different (e.g. there may not be a spacer and rather the whole layout lights up - but I think the spacer should still be used but be the size of the layout in e.g. a float layout), but the two basic behaviors would still be there (layout and widget dragged).

Would making the spacer optional help? I'm not really sure how that would work, because there would be no visual indication of dropping. Can you give a more concrete example?


```
Also, i would move out the creation of the preview out of the controller to the DraggableObjectBehavior itself. I think it's best that each widget can have a different preview implementation rather than taking the decision to draw it in a Fbo. For example, if you have a "group" of widget or a selection that you want to accept drag&drop, you could customize the preview just by returning a instance of something else and not use the Fbo. And that instance could also react to some event like if it's the right place to drop or not.
```

Are you objecting to the use of an Fbo, or to using the widget being dragged as the preview? If it's the latter, that's why `DraggableObjectBehavior.drag_widget` is there. By default it's `None` and the dragged widget is previewed on the Fbo. But if not None, its value would be previewed in the Fbo. This can also be easily overwritten. The controller calls `self.prepare_preview_widget(source.drag_widget or source)`.

If the problem is with the Fbo, I can add a property, e.g. `preview_widget` to `DraggableObjectBehavior` that is used if provided instead of the Fbo. In terms of reacting or looking differently if it's on top of something that accepts it, the widget being dragged is not currently aware of this kind of thing, all it knows is when a drag starts and stops. We would need to add stuff to make it reactive. Currently, only the receiving layout is reactive as it shows the spacer when the touch is on top of it.

Are you thinking that the layout should not respond to an object being on top of it with a spacer, but rather the dragged widget should change? If so, we'd need to modify the controller to handle the touch when it's over the layout and it could call a method on the dragged object to notify them. Is this what you meant with your first comments also? In that case, we could just add  methods e.g. `enter_drop` and `leave_drop` to `DraggableObjectBehavior`. But there would still be a a layout behavior for potential dropps and a draggable widget behavior of dragging.

PR #6026, opened 2018-10-29
Support building against mesa video core drivers.

rnixx:
* Remove unused c_options['use_mali']
* Add platform option to build against mesa video core drivers

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -139,8 +139,12 @@ if ndkplatform is not None and environ.get('LIBLINK'):
 kivy_ios_root = environ.get('KIVYIOSROOT', None)
 if kivy_ios_root is not None:
     platform = 'ios'
+# proprietary broadcom video core drivers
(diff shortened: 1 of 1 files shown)


tito:
Why removing the Mali option?

tito:
I would rename vc to rpivc if possible too

rnixx:
> Why removing the Mali option?
It's never used. Every mali related aspect is controlled with the ``platform`` variable in ``setup.py``

> I would rename vc to rpivc if possible too
``rpi`` platform checks for ``/opt/vc/...``, which is broadcom's video code drivers, so in theory not even raspberry pi related. Thus i thought I better name the new option ``vc`` for video core, in analogy to ``mali``, which also addresses a GPU family. I even was temped to rename ``rpi`` to ``bc`` for broadcom ;)

PR #6027, opened 2018-10-29
Fix 5146

seanodonnell:
This is intended to resolve https://github.com/kivy/kivy/issues/5146

There is an existing pull request here: https://github.com/kivy/kivy/pull/5311  which seems to have stalled. It involves actually measuring the cameras framerate. 

I think that pull request will be a better long term solution when finished. But in the interim, this one fixes current problems with no drawbacks that I can think of.


tito:
Good for me.

kivy/core/camera/camera_opencv.py
```diff
@@ -118,8 +118,10 @@ def init_camera(self):
             # get fps
             self.fps = self._device.get(PROPERTY_FPS)
 
-        if self.fps <= 0:
-            self.fps = 1 / 30.
+        if self.fps == 0 or self.fps == 1:
+            self.fps = 1 / 30
```
pythonic64: On Python2 `1 / 30 == 0`. You have to ensure float value, therefore: `1.0 / 30`

kivy/core/camera/camera_opencv.py
```diff
@@ -118,8 +118,10 @@ def init_camera(self):
...
+        elif self.fps > 1:
+            self.fps = 1 / self.fps
```
pythonic64: Same here, it should be `self.fps = 1.0 / self.fps`.

seanodonnell: Thanks, I'll put together a follow on pull request asap

seanodonnell:
follow on to fix the division bug here https://github.com/kivy/kivy/pull/6032

kivy/core/camera/camera_opencv.py
vitormhenrique: same for me, fine one mac, but on windows I needed the `self.fps = 1.0 / self.fps.`

seanodonnell: There was a follow on pull request from this one which made that change

seanodonnell: here it is https://github.com/kivy/kivy/pull/6032/files

PR #6036, opened 2018-11-01
website layout improvement

Nearoo:
Changes this

beofre

to this 

after

(guides: [before](https://imgur.com/6L7SwXr.png), [after](https://imgur.com/3xo8SlK.png) ).

Research suggest shorter lines improves readability. [[1]](https://en.wikipedia.org/wiki/Line_length). It is said that 75 would be optimal, however I find that a bit drastic. This is about 120.

### DIFF
--- a/doc/sources/.static/fresh.css
+++ b/doc/sources/.static/fresh.css
@@ -143,7 +143,7 @@ a:hover {
 
 
 div.sphinxsidebar {
-	width: 230px;
+	width: 15%;
 	float: left;
 }
 
@@ -214,8 +214,10 @@ div.sphinxsidebar li a:hover {
 }
 
 #content {
-	margin-left: 250px;
-	margin-top: -8px;
+	margin: auto;
+	width: 950px;
+	min-width: 50%;
+	max-width: 95%;
 }
 
 #content pre {
@@ -423,6 +425,7 @@ div.sphinxsidebar .api-index li a {
 
 /** TOC **/
 .toc {
+	width: auto;
 	background-color: #f5f5f5;
 	border: 1px solid #e0e0e0;
 	padding: 10px;
@@ -469,7 +472,6 @@ table.highlighttable td.linenos {
 /** version added in method/class/attribute/module **/
 span.versionadded {
 	font-size: 12px;
-	position: absolute;
 	right: 2.5em;
 }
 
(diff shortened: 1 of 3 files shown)


matham:
I like the look, but will let someone more familiar with js/css review.

tito:
I don't think it is ok.
The current layout is not optimal, but your changes make it less:
- putting all the title on bold breaks the padding intented for some block (see the menu at the right)
- the bold title is not consistent with the kivy website anymore
- there is blank padding at the left and right side of the content
- the content is now fixed size, so even if before it was too big to be read on large screen, yours cannot be read on tiny anymore.
- the sidebar is 15% width. meaning it will be unusable on tiny screen also (even if it's already unusable), but if you put %, put also min/max width. But IMO, it should stay the same width as today.

I'll be happy with just something that have max-width on the content since it's what you want to address. The others should not be changed, except if it include a proper discussion about accessibility, mobility etc.

Nearoo:
@tito Are you open to merging my pull request if I address your concerns?

KeyWeeUsr:
@Nearoo I'd like to see the results on phone resolutions more than on desktops. Still though even if it improves the readability, the white space is about the same width as 3x sidebar.

If there needs to be a change in the design instead of screenshots I'm rather for setting up a dummy website (own hosting, free hosting, ReadTheDocs, GitHub pages, xyz), so everyone can test if it breaks something. Just upload the content of `build` folder somewhere and remove it afterwards.

Nearoo:
@KeyWeeUsr Alright, will do. 

>  Still though even if it improves the readability, the white space is about the same width as 3x sidebar.

...just to be sure: Do you think that's a bad idea? Because I personally consider it _much_ more comfortable. I will upload my version this week so you can look at it.

KeyWeeUsr:
I don't think it's a bad idea to add *some* space for readability. I do think it's a bad idea to leave half of the website's width blank. If you read through the answers on [this question](https://meta.stackexchange.com/q/310908/323700) you'll get various views on the topic.

Especially [this one](https://meta.stackexchange.com/a/311254/323700) pretty much describes my opinion on the whitespace on websites. I don't use vertical monitors, but I use phones and if you do such thing for phones, the docs will be unreadable :dango:

Nearoo:
I agree on your note about phones. I have changed the website locally so it implements the `flex`-layout and the `@media` css tag for practice, which makes the website kinda responsive. Will commit those next week too, then we have a visual reference.

PR #6037, opened 2018-11-02
Cythonized a few important widgets, and lang

tshirtman:
Adding lang package and some widgets to the cythonization in setup.py, improve performances at the cost of making kivy a bit less easy to change (as you now need to rebuild everytime you change any of these, or just to delete the .so files when you are developping).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -800,7 +800,14 @@ sources = {
             'lib/libtess2/Source/tess.c'
         ]
     }),
-    'graphics/svg.pyx': merge(base_flags, gl_flags_base)
+    'graphics/svg.pyx': merge(base_flags, gl_flags_base),
+    'lang/builder.py': {},
+    'lang/parser.py': {},
+    'uix/videoplayer.py': {},
+    'uix/relativelayout.py': {},
+    'uix/floatlayout.py': {},
+    'uix/layout.py': {},
+    'uix/carousel.py': {},
 }
 
 if c_options["use_sdl2"]:


setup.py
```diff
@@ -800,7 +800,14 @@ def determine_sdl2():
...
     }),
-    'graphics/svg.pyx': merge(base_flags, gl_flags_base)
+    'graphics/svg.pyx': merge(base_flags, gl_flags_base),
+    'lang/builder.py': {},
+    'lang/parser.py': {},
+    'uix/videoplayer.py': {},
+    'uix/relativelayout.py': {},
+    'uix/floatlayout.py': {},
+    'uix/layout.py': {},
+    'uix/carousel.py': {},
```
brentpicasso: It looks like you cherry picked some layouts to cythonize, but not all of them, like boxlayout or anchorlayout?

tshirtman: Yes, it was mostly opportunistic, i don't think there is much preventing adding others.

tshirtman: This is a "dumb" cythonization, the kind that can sometime gets you a 300% performance increase, but rarely more, a more involved PR would make use of cython-specific syntax and make it mandatory to build these modules for usage, and potentially increase them much more.

tito:
This is a gist i did while ago: https://gist.github.com/tito/9092028
- scatter.pyo: 12KB
- scatter.so (pyx compiled with Os (small optimization, not even O2) and stripped): 146KB

You can imagine the impact if you do that for all the py :)

matham:
Not really sure about doing this to widgets. One of the major benefits of python is that you can easily modify and/or step through the debugger when issues arise. I do this all the time. Compiling to cython will make that harder/impossible. This is especially relevant for user facing widgets which are likely to be debugged in this way compared to core widgets. 

If we do add this, it should be optional and turned off by default. We can use a env variable to turn it ON. Also, how did you pick the widgets in the list? It seems kinda random :)

tito:
More than that, do we want to support that?
Why adding cythonize python here, and not think globally for an app?
The later is imo what the end user want to achieve. Doing it in Kivy is just a part of it.

KeyWeeUsr:
Same as @matham when something doesn't seem right or I want to debug/rewrite quickly, I change the source directly and I do not expect to have e.g. GDB or something else present on each machine I work on (I tend to code on phone too, so no luck there even if I'd wanted to try).

KV lang as an addition to SVG I understand, but the layouts choice seems random to me. I'm not even sure what would be the performance improvement for layouts and whether it's worth the effort to compile it (binary size, not being able to hotfix/debug/etc quickly) and lock yourself out. Carousel, hm... perhaps with many widgts? But that limits our use-case as a framework to a narrowly specified use-case.

IMHO those particular files aren't worth the effort and I'd rather expect `Scatter`, `ScrollView` and likes to be compiled because those are the files at least I tend to hotfix/play with less and personally I don't expect other people to mess with let's say `ScrollView` or `Scatter` internals on the fly. Still though it'd need a table with results before/after the change + testing it on mobile platforms where we care about the total application size (more).

tshirtman:
Yeah, certainly the selection is a bit random here, and much more though can be put into it, not expecting this to be merged as-is, I do think cythonizing lang makes sense and we had most performance gains from that, but i agree going for widgets is going to be a lot more painful, and yeah, the size concern is pretty big (here the content of uix after cythonizing *all* the python files in the repos).

```
du -csh kivy/uix/*.so
732K	kivy/uix/accordion.cpython-36m-x86_64-linux-gnu.so
1,1M	kivy/uix/actionbar.cpython-36m-x86_64-linux-gnu.so
352K	kivy/uix/anchorlayout.cpython-36m-x86_64-linux-gnu.so
680K	kivy/uix/boxlayout.cpython-36m-x86_64-linux-gnu.so
556K	kivy/uix/bubble.cpython-36m-x86_64-linux-gnu.so
96K	kivy/uix/button.cpython-36m-x86_64-linux-gnu.so
260K	kivy/uix/camera.cpython-36m-x86_64-linux-gnu.so
1,1M	kivy/uix/carousel.cpython-36m-x86_64-linux-gnu.so
216K	kivy/uix/checkbox.cpython-36m-x86_64-linux-gnu.so
508K	kivy/uix/codeinput.cpython-36m-x86_64-linux-gnu.so
1,1M	kivy/uix/colorpicker.cpython-36m-x86_64-linux-gnu.so
572K	kivy/uix/dropdown.cpython-36m-x86_64-linux-gnu.so
1,9M	kivy/uix/filechooser.cpython-36m-x86_64-linux-gnu.so
432K	kivy/uix/floatlayout.cpython-36m-x86_64-linux-gnu.so
996K	kivy/uix/gesturesurface.cpython-36m-x86_64-linux-gnu.so
1,1M	kivy/uix/gridlayout.cpython-36m-x86_64-linux-gnu.so
580K	kivy/uix/image.cpython-36m-x86_64-linux-gnu.so
524K	kivy/uix/label.cpython-36m-x86_64-linux-gnu.so
628K	kivy/uix/layout.cpython-36m-x86_64-linux-gnu.so
468K	kivy/uix/modalview.cpython-36m-x86_64-linux-gnu.so
496K	kivy/uix/pagelayout.cpython-36m-x86_64-linux-gnu.so
324K	kivy/uix/popup.cpython-36m-x86_64-linux-gnu.so
192K	kivy/uix/progressbar.cpython-36m-x86_64-linux-gnu.so
760K	kivy/uix/recycleboxlayout.cpython-36m-x86_64-linux-gnu.so
788K	kivy/uix/recyclegridlayout.cpython-36m-x86_64-linux-gnu.so
968K	kivy/uix/recyclelayout.cpython-36m-x86_64-linux-gnu.so
376K	kivy/uix/relativelayout.cpython-36m-x86_64-linux-gnu.so
2,1M	kivy/uix/rst.cpython-36m-x86_64-linux-gnu.so
508K	kivy/uix/sandbox.cpython-36m-x86_64-linux-gnu.so
1,1M	kivy/uix/scatter.cpython-36m-x86_64-linux-gnu.so
228K	kivy/uix/scatterlayout.cpython-36m-x86_64-linux-gnu.so
1,6M	kivy/uix/screenmanager.cpython-36m-x86_64-linux-gnu.so
1,9M	kivy/uix/scrollview.cpython-36m-x86_64-linux-gnu.so
1,6M	kivy/uix/settings.cpython-36m-x86_64-linux-gnu.so
428K	kivy/uix/slider.cpython-36m-x86_64-linux-gnu.so
416K	kivy/uix/spinner.cpython-36m-x86_64-linux-gnu.so
652K	kivy/uix/splitter.cpython-36m-x86_64-linux-gnu.so
576K	kivy/uix/stacklayout.cpython-36m-x86_64-linux-gnu.so
64K	kivy/uix/stencilview.cpython-36m-x86_64-linux-gnu.so
220K	kivy/uix/switch.cpython-36m-x86_64-linux-gnu.so
1,1M	kivy/uix/tabbedpanel.cpython-36m-x86_64-linux-gnu.so
5,2M	kivy/uix/textinput.cpython-36m-x86_64-linux-gnu.so
68K	kivy/uix/togglebutton.cpython-36m-x86_64-linux-gnu.so
888K	kivy/uix/treeview.cpython-36m-x86_64-linux-gnu.so
448K	kivy/uix/video.cpython-36m-x86_64-linux-gnu.so
1020K	kivy/uix/videoplayer.cpython-36m-x86_64-linux-gnu.so
1,4M	kivy/uix/vkeyboard.cpython-36m-x86_64-linux-gnu.so
1,3M	kivy/uix/widget.cpython-36m-x86_64-linux-gnu.so
40M	total
```

PR #6044, opened 2018-11-08
Limited FileChooserProgress text size to widget size

tshirtman:


kivy/data/style.kv
```diff
@@ -436,6 +436,7 @@
         pos_hint: {'x': .2, 'y': .6}
         size_hint: .6, .2
         text: 'Opening %s' % root.path
+        text_size: self.size
```
matham: Do we need to also set valign and halign to center? Otherwise it would just show on the left.

Zen-CODE: Indeed

    halign: 'center'

fits best with the centering of the rest of the layout...

Issue #5939, opened 2018-09-19
Example Camera app crash on start Android


Huyston:
### Versions

* Python: python3crystax
* OS: Android
* Kivy: v1.10.0
* Kivy installation method: buildozer

### Description

The example Camera App from https://kivy.org/doc/stable/examples/gen__camera__main__py.html doesn't work on Android.
The app crashes on startup.

### Code and Logs

```python
09-18 23:27:21.292 25017-25057/org.test.cameraexample I/python: Initialize Python for Android
    Changing directory to the one provided by ANDROID_ARGUMENT
    /data/user/0/org.test.cameraexample/files/app
    Preparing to initialize python
    crystax_python exists
    calculated paths to be...
    /data/user/0/org.test.cameraexample/files/app/crystax_python/stdlib.zip:/data/user/0/org.test.cameraexample/files/app/crystax_python/modules
    set wchar paths...
09-18 23:27:21.679 25017-25057/org.test.cameraexample I/python: Initialized python
    AND: Init threads
09-18 23:27:21.680 25017-25057/org.test.cameraexample I/python: testing python print redirection
09-18 23:27:21.683 25017-25057/org.test.cameraexample I/python: Android path ['.', '/data/user/0/org.test.cameraexample/files/app/crystax_python/stdlib.zip', '/data/user/0/org.test.cameraexample/files/app/crystax_python/modules', '/data/user/0/org.test.cameraexample/files/app/crystax_python/site-packages']
[...]

PR #6051, opened 2018-11-16
Update camera_android.py; fixes camera for Python 3

gbm001:
In Python 3, the buffer needs to an explicit bytestring rather than a string; hence changing:
buf = '\x00' * self._bufsize
to
buf = b'\x00' * self._bufsize

Fixes various issues:
in Kivy https://github.com/kivy/kivy/issues/5939
and in associated garden project ZbarCam https://github.com/kivy-garden/garden.zbarcam/issues/13

### DIFF
--- a/kivy/core/camera/camera_android.py
+++ b/kivy/core/camera/camera_android.py
@@ -126,7 +126,7 @@ class CameraAndroid(CameraBase):
         with self._buflock:
             self._buffer = None
         for k in range(2):  # double buffer
-            buf = '\x00' * self._bufsize
+            buf = b'\x00' * self._bufsize
             self._android_camera.addCallbackBuffer(buf)
         self._android_camera.setPreviewCallbackWithBuffer(self._preview_cb)
 


tito:
Is this still working with Python 2? I mean did you tried?

gbm001:
I haven't because I haven't got a Python 3 environment set up to test it (I am very new to Kivy). However, it shouldn't make any difference in Python 2.6 as b'xxxxx' is completely identical to 'xxxxx' in Python 2. It will cause a SyntaxError in Python 2.5 as the syntax was introduced in 2.6, but I believe Kivy only supports 2.7+?

tito:
@gbm001 Thanks you!

einav888:
Where can i update this? i cannot find the  camera_android.py file....
I'm trying to run this -> https://kivy.org/doc/stable/examples/gen__camera__main__py.html on Kivy Launcher. It runs perfectly on my computer, but crashes automatically on Kivy Launcher.

einav888:
I checked the camera_android.py file and it already was:      buf = b'\x00' * self._bufsize
I'm trying to run this -> https://kivy.org/doc/stable/examples/gen__camera__main__py.html on Kivy Launcher. It runs perfectly on my computer, but crashes automatically on Kivy Launcher.
Do you have any idea what the reason could be?

PR #6054, opened 2018-11-21
Add new "canonical" path for binary Mali driver

Cheaterman:
This improves detection on "standard" setups of the Mali driver for platforms such as the Odroid XU4.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -143,7 +143,8 @@ if exists('/opt/vc/include/bcm_host.h'):
     platform = 'rpi'
 if(
     exists('/usr/lib/arm-linux-gnueabihf/libMali.so')
-    or exists('/usr/local/mali-egi/libmali.so')
+    or exists('/usr/lib/arm-linux-gnueabihf/mali-egl/libmali.so')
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -141,7 +141,10 @@ def pkgconfig(*packages, **kw):
     platform = 'ios'
 if exists('/opt/vc/include/bcm_host.h'):
     platform = 'rpi'
-if exists('/usr/lib/arm-linux-gnueabihf/libMali.so'):
+if(
+    exists('/usr/lib/arm-linux-gnueabihf/libMali.so')
+    or exists('/usr/local/mali-egi/libmali.so')
```
jakogut: Shouldn't this path be `/usr/local/mali-egl/libmali.so` (instead of `mali-egi`)? Also, Arch Linux ARM installs this library to a different path. Would it make sense to allow the path to the libGL to be specified with an environment variable during the build?

Cheaterman:
(Probably squash the commits if/when merging, but first please let @jakogut and possibly other Mali users answer with their distribution's correct search location)

Issue #1401, opened 2013-07-31
TextInput Mask


vgaviraghi:
Would be interesting if the TextInput had the option to format the field through a mask, such as money values​​, zip code, CPF/CNPJ (Brazilian uses)

rodhoff:
Adding the MaskedTextInput class, wich inherits from TextInput, and has a 'mask' property. It allows to add values such as '###.###.###-##' in .kv. Pull request #6090 created for this.

PR #6090, opened 2019-01-02
Added MaskedTextInput class, with masking option, (# for numeric, A f…

rodhoff:
…or string)

This is in order to address properly Issue #1401 and have an out-of-the-box ready component for adding format-specific text. It understands simply literal sizing, with numbers (#) and chars (A), but it can be enhanced for further masking operations.

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -3353,6 +3353,47 @@ class TextInput(FocusBehavior, Widget):
     defaults to `True`.
     '''
 
+class MaskedTextInput(TextInput):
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
Sorry, but I don't think custom widget such as this should be part of the core widget set as already said https://github.com/kivy/kivy/issues/1401#issuecomment-48821969. Also you have non-ASCII characters in your string and the mask is basically hard-coded. You can try to add it to https://github.com/kivy-garden though if you fix the implementation.

PR #6095, opened 2019-01-10
Move android data_dir to write-able location. references #4777

### LINKED ISSUE #4777: Local Config file misplaced

Zen-CODE:


kivy/app.py
```diff
@@ -737,6 +740,29 @@ def directory(self):
...
+            file_p = cast('java.io.File', context.getFilesDir())
+            data_dir = file_p.getAbsolutePath()
+        elif platform == 'win':
+            data_dir = os.path.join(os.environ['APPDATA'], self.name)
+        elif platform == 'macosx':
+            data_dir = '~/Library/Application Support/{}'.format(self.name)
+        else:  # _platform == 'linux' or anything else...:
+            data_dir = os.environ.get('XDG_CONFIG_HOME', '~/.config')
+            data_dir = join(data_dir, self.name)
+        data_dir = expanduser(data_dir)
```
JonasT: the location of this line seems like a bug, not caused by this PR of course but maybe worth fixing in the same go? `expanduser` is not correct to do on the Windows and Android paths, and might actually lead to this function returning an entirely wrong path on Android **since it is possible to create a folder with `~` in the name on Linux** (and therefore most likely on Android)!

So this line should really be moved inside the cases where it is appropriate, which as far as I can tell is Linux, iOS, macOS and possibly Linux - although here it might also be wrong, depending on whether is `XDG_CONFIG_HOME` allowed to contain `~` which I can't tell off the cuff. Might be worth checking the freedesktop spec

Zen-CODE: Glad you mentioned that. I noticed that too but decided not to play with it. But I agree, now is probably a good time to change that. Will do...

JonasT: Nice :+1: I checked the new `expanduser` locations and from my understanding and knowledge, they are correct now. Very good! :tada:

Zen-CODE:
Yeah, you're right. But that's not such a straight forward one to fix. Despite the defaultpath being passed in as an optional, it seems to be flatly ignored and re-assigned??? The doc string seems entirely misleading...

JonasT:
@Zen-CODE I think the idea is that it defaults to be in the app data directory *usually*, while the (kivy-app-writing-) user can specify another location, and there was an attempt to fix the path for some special cases.

And the Windows fix looks indeed legitimate to me, while the iOS one is weird (maybe that location is good, but why override it?), and the Android one definitely looks like now outdated nonsense. I suggest to remove it.

So I'd change it to this:
```
        if platform == 'ios':
            defaultpath = '~/Documents/%(appname)s.ini'
        elif platform == 'win':
            defaultpath = defaultpath.replace('/', sep)
        return expanduser(defaultpath) % {
            'appname': self.name, 'appdir': self.directory}
```
And possibly the iOS one should be removed as well, but that should be decided by someone who has an idea how iOS looks like

JonasT:
@Zen-CODE with "I'd change it..." I meant to suggest to put it in addition into this pull request btw

Zen-CODE:
@JonasT. Okay, on further examination, the docs do say the defaultpath parameter only applies to Desktop OS's, which makes the code and docs much less confusing. Whew :-)

The iOS version seems fine, but missed a dot in the config name, which I've added back. I think it's overridden to prevent any changes. It HAS to be in the documents subfolder to be write-able, so it's not really open to much changing. The documentation clarification that the defaultpath changes are applicable only to desktop OS's would seen to confirm that.

So I've pushed the changes discussed. They should do the trick? Thanks again

JonasT:
> I think it's overridden to prevent any changes.

Okay makes sense. With that in mind, your new changes look good!

JonasT:
For what it's worth, I think `_get_user_data` home on Linux is either wrong, or the config should be in a different place (because the config should be either in `XDG_CONFIG_HOME` (or `~/.config/` if not set), or `XDG_DATA_HOME` (or `~/.local/share/` if not set). Putting config folders directly into the home is very outdated and very annoying on Linux and shouldn't be done

But of course, that's completely beyond the scope of the pull request. It might be some food for thought for future changes

JonasT:
@tshirtman would be really cool if this could go in some time soon, so we can merge the SDL2 p4a bump without breaking kivy! :heart:

JonasT:
Sorry, still had a "request changes" review left on it myself, apparently. All the things I requested were solved in a way I find useful though, so as far as I'm concerned this looks good :+1:

Zen-CODE:
MacOS (10.13), iOS (9.3.5) tested, working.

hackalog:
If a migration path is desired (and this would be a separate PR), we could check the old location for a config file and return the old location (with a big DEPRECATION warning) that this behavior will change on the next kivy major version crank.

JonasT:
@hackalog just keep in mind this will cause a PermissionError on API level >= 21, unless the application already has decided to get SD card permission on its own. So that should possibly be handled gracefully (by not blowing up)

PR #6097, opened 2019-01-11
Support of base64 pictures

kuzeyron:
My attempt to add support of base64 encoded pictures. Very useful features when you don't want to store locally these pictures. Also good when you work with remote apps in Kivy.

Demo app of examples on how to use it:
https://gist.github.com/kuzeyron/80416dcb10111ea858d03737f1e9423c

kivy.uix.image.Base64Image is ( WIP !! ) as we talked about with André!

### DIFF
--- a/kivy/uix/image.py
+++ b/kivy/uix/image.py
@@ -428,7 +428,7 @@ class Base64Image(Image):
                 image64 = source.split(';base64,')[1]
                 type64 = source.split('data:image/')[1] \
                                 .split(';base64,')[0]
-            except:
+            except IndexError:
                 raise Exception('Wrong type of base64 image.')
             base64l = base64.b64decode(image64)
             self._coreimage = image = CoreImage(
@@ -443,7 +443,7 @@ class Base64Image(Image):
 
     def _on_source_load(self, value):
         image = self._coreimage.image
-        if not image:
+        if image is None:
             return
         self.texture = image.texture
         self.dispatch('on_load')
--- a/kivy/uix/image.py
+++ b/kivy/uix/image.py
@@ -53,7 +53,8 @@ from kivy.properties import StringProperty, ObjectProperty, ListProperty, \
     AliasProperty, BooleanProperty, NumericProperty
 from kivy.logger import Logger
 from kivy.compat import PY2
-
+from io import BytesIO
+import base64
 # delayed imports
 Loader = None
 
@@ -400,3 +401,74 @@ class AsyncImage(Image):
             Loader.remove_from_cache(source)
 
         super(AsyncImage, self).reload()
+
+
+class Base64Image(Image):
+
+    __events__ = ('on_error', 'on_load')
+
+    def __init__(self, **kwargs):
+        self._coreimage = None
+        super(Base64Image, self).__init__(**kwargs)
+        self.fbind('source', self._load_source)
+        if self.source:
+            self._load_source()
+        self.on_anim_delay(self, kwargs.get('anim_delay', .25))
+
+    def _load_source(self, *args):
+        source = self.source
+        if not source:
+            if self._coreimage is not None:
+                self._coreimage.unbind(on_texture=self._on_tex_change)
+                self._coreimage.unbind(on_load=self._on_source_load)
+            self.texture = None
+            self._coreimage = None
+        else:
+            try:
+                image64 = source.split(';base64,')[1]
+                type64 = source.split('data:image/')[1] \
+                                .split(';base64,')[0]
+            except:
+                raise Exception('Wrong type of base64 image.')
+            base64l = base64.b64decode(image64)
+            self._coreimage = image = CoreImage(
+                BytesIO(base64l), ext = type64,
+                nocache=self.nocache, mipmap=self.mipmap,
+                anim_delay=self.anim_delay)
+
+            image.bind(on_load=self._on_source_load)
+            image.bind(on_error=self._on_source_error)
+            image.bind(on_texture=self._on_tex_change)
+            self.texture = image.texture
+
+    def _on_source_load(self, value):
+        image = self._coreimage.image
+        if not image:
+            return
+        self.texture = image.texture
+        self.dispatch('on_load')
+
+    def _on_source_error(self, instance, error=None):
+        self.dispatch('on_error', error)
+
+    def on_error(self, error):
+        pass
+
+    def on_load(self, *args):
+        pass
+
+    def _on_tex_change(self, *largs):
+        if self._coreimage:
+            self.texture = self._coreimage.texture
+
+    def texture_update(self, *largs):
+        pass
+
+    def reload(self):
+        if Loader:
+            source = self.source
+            if not self.is_uri(source):
+                source = resource_find(source)
+            Loader.remove_from_cache(source)
+
+        super(Base64Image, self).reload()
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -250,6 +250,9 @@ from kivy.context import get_current_context
 from kivy.weakproxy import WeakProxy
 from functools import partial
 from itertools import islice
+from io import BytesIO
+from PIL import Image
+import base64
 
 
 # References to all the widget destructors (partial method with widget uid as
@@ -688,15 +691,8 @@ class Widget(WidgetBase):
 
                 .. versionadded:: 1.11.0
         '''
-        try:
-            from PIL import Image
-        except ImportError as error:
-            raise Exception(error)
-        from io import BytesIO
-
-        import base64
         scale = kwargs.get('scale', 1)
-        
+
         if self.parent is not None:
             canvas_parent_index = self.parent.canvas.indexof(self.canvas)
             if canvas_parent_index > -1:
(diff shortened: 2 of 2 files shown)


kivy/uix/widget.py
```diff
@@ -660,6 +660,73 @@ def export_to_png(self, filename, *args, **kwargs):
...
+
+                .. versionadded:: 1.11.0
+        '''
+        try:
+            from PIL import Image
+        except ImportError as error:
+            raise Exception(error)
+        from io import BytesIO
+
+        import base64
```
AndreMiras: Question: why the lazy imports?

kivy/uix/image.py
```diff
@@ -400,3 +401,74 @@ def reload(self):
...
+                self._coreimage.unbind(on_texture=self._on_tex_change)
+                self._coreimage.unbind(on_load=self._on_source_load)
+            self.texture = None
+            self._coreimage = None
+        else:
+            try:
+                image64 = source.split(';base64,')[1]
+                type64 = source.split('data:image/')[1] \
+                                .split(';base64,')[0]
+            except:
```
AndreMiras: that exception catching is a bit too broad to me, are you looking for `IndexError` instead?

kivy/uix/image.py
```diff
@@ -400,3 +401,74 @@ def reload(self):
...
+                anim_delay=self.anim_delay)
+
+            image.bind(on_load=self._on_source_load)
+            image.bind(on_error=self._on_source_error)
+            image.bind(on_texture=self._on_tex_change)
+            self.texture = image.texture
+
+    def _on_source_load(self, value):
+        image = self._coreimage.image
+        if not image:
```
AndreMiras: minor: do you mean `if image is None:`?

kivy/uix/widget.py
```diff
@@ -660,6 +660,73 @@ def export_to_png(self, filename, *args, **kwargs):
...
+                empty or 'web'. Note: None is for returning the image in Kivy and to be shown.
+            `scale`: float
+                The amount by which to scale the saved image, defaults to 1.
+
+                .. versionadded:: 1.11.0
+        '''
+        try:
+            from PIL import Image
+        except ImportError as error:
+            raise Exception(error)
```
AndreMiras: Why are we rerasing to a less explicit type? Why even try/catching?

AndreMiras:
I made a first quick review. It looks good overall :clap: 
We still probably need to test it I guess.
Also could we clean the commit history a bit? We have like 10+ commits that say "Adding Arch Linux ARM". Maybe you should rebase and squash, what do you think?

kivy/uix/image.py
kuzeyron: Now when you say it!

kivy/uix/widget.py
kuzeyron: You mean why I use it only in the function and not for the entire widget?
If so: I don't think it's useful to call them for all the other widget functions.

kivy/uix/image.py
kuzeyron: This is the part I do not really understand how these rules are working. Someone needs to cure that!

kivy/uix/widget.py
kuzeyron: I wanted it to look good. And a few Google searches this seemed to be a good one to use.

kuzeyron:
@AndreMiras Absolutely! Be my guess and do that. I'm still too new with this so any help is appreciated!

kivy/uix/image.py
AndreMiras: OK, it was minor anyway, but I was talking about the Pythonic to check for `None` https://www.python.org/dev/peps/pep-0008/#programming-recommendations

kivy/uix/widget.py
AndreMiras: The idea is reraising needs to serve a purpose, like specializing the exception or logging info prior to raising or something.
If you find that link you came across in your Google searches, feel free to share it, I'm curious about it. I could be missing something

kivy/uix/image.py
kuzeyron: Aaa okay!

kivy/uix/widget.py
AndreMiras: Yes I mean why importing not in the top of the file? Usually this is where we put imports unless a specific reason, for instance to workaround dependency cycles, but there're other reasons.
Since lazy importing is the exception and not the rule, what I usually do is to put a comment explaining why I would do a lazy import in a specific place, so later maintainers won't move them by mistake unless it's for good reason

kivy/uix/widget.py
kuzeyron: Here it is (Stackoverflow)[https://stackoverflow.com/a/3131251]

AndreMiras: OK so here what they do is to fallback on another import if one doesn't exist, that's the reason they try/except. They don't try/except to just reraise, you see what I mean?
For instance this pattern is used for urllib import that works with both Python2 and Python3:
```python
try:
    from urllib.parse import urlparse
except ImportError:
    from urlparse import urlparse
```
Another example could be to import a non-mandatory module:
```python
try:
    import foo
except ImportError:
    foo = None

# then later in the code
if foo is not None:
    foo.whatever()
```
But again there's no reraise. One reason to reraise could be:
```python
try:
    import foobar
except ImportError:
    logger.error('I want dedicated logs for this error')
    raise
```
Or:
try:
    import foobar
except ImportError:
    raise FoobarModuleNotFound('whatever message')
```
This sometimes makes sense in specific cases, e.g. if you're dealing with `FoobarModuleNotFound` later on.

Does it make more sense?

Comment (kuzeyron) [reply kivy/uix/widget.py]: Aa okay! But you should be able to edit this because I allowed it when I made this PR.

**Comment by AndreMiras** [comment]
OK, I'll try to rewrite the history with a single commit on your name and I'll make another commit to implement my minor suggestions/fixes

File: kivy/uix/widget.py (hunk 1 above)
Comment (AndreMiras) [reply kivy/uix/widget.py]: addressed in daab224

File: kivy/uix/image.py (hunk 2 above)
Comment (AndreMiras) [reply kivy/uix/image.py]: addressed in daab224

File: kivy/uix/image.py (hunk 3 above)
Comment (AndreMiras) [reply kivy/uix/image.py]: addressed in daab224

File: kivy/uix/widget.py (hunk 4 above)
Comment (AndreMiras) [reply kivy/uix/widget.py]: addressed in daab224

**Comment by AndreMiras** [comment]
OK I rewrote the history, it's now 2 rather than 27 commits and I've addressed few comments.
I haven't tested it yet. Feel free to give it a try to make sure I didn't introduce any regression

**Comment by kuzeyron** [comment]
@AndreMiras I don't follow the process: Feel free to give it a try to make sure I didn't introduce any regression. My understand I need to do something. Is it enough with `git pull`? 
Github has too many functions so I feel like being stuck in the traffic.

**Comment by AndreMiras** [comment]
Yes that's what I mean, you can simply git pull and try out see if I didn't break anything from your original PR

**Comment by kuzeyron** [comment]
@AndreMiras Thank you will get back to you!

**Comment by kuzeyron** [comment]
@AndreMiras Tested with both `v2.7.15 (default, Jun 27 2018, 13:05:28)` and `v3.7.1 (default, Oct 22 2018, 10:41:28)` and runs like an old Corvette!

**Comment by AndreMiras** [comment]
Great! Well it looks good to me.
Now I'm thinking Couldn't we have achieved that base64 image thingy directly playing with the `CoreImage` class? I mean in your [code example](https://gist.github.com/kuzeyron/80416dcb10111ea858d03737f1e9423c), rather than going with:
```python
b64 = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAA \
    FCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAA \
    O9TXL0Y4OHwAAAABJRU5ErkJggg=="
b64_image = Base64Image(source=self.b64)
add_widget(b64_image)
```
Why couldn't we go with?:
```python
b64 = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAA \
    FCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAA \
    O9TXL0Y4OHwAAAABJRU5ErkJggg=="
data = io.BytesIO(b64)
b64_image = CoreImage(data, ext="png")
add_widget(b64_image)
```

kuzeyron:
@AndreMiras I prefer hiding the BytesIO. I don't want to confuse the users you know

kivy/uix/widget.py
```diff
@@ -250,6 +250,9 @@ def on_touch_down(self, touch, after=False):
 from kivy.weakproxy import WeakProxy
 from functools import partial
 from itertools import islice
+from io import BytesIO
+from PIL import Image
```
tito: Importing PIL in Widget is a absolute no-go!

kuzeyron: If you know another way around it, I would like to know.

kuzeyron:
@akshayaurora Is it possible for you guys to put that together? Prettier to have that built-in like the export_to_png().

tito:
@kuzeyron I'll redo this PR to make it work in the way better for Kivy/API, and update here.

PR #6119, opened 2019-02-01
Add support for opencv 4

adn05:
The opencv camera provider don't support opencv 4, so we get errors when trying to use Camera with opencv 4

### DIFF
--- a/kivy/core/camera/camera_opencv.py
+++ b/kivy/core/camera/camera_opencv.py
@@ -71,7 +71,7 @@ class CameraOpenCV(CameraBase):
 
     def init_camera(self):
         # consts have changed locations between versions 2 and 3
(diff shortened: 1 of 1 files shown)


kivy/core/camera/camera_opencv.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, **kwargs):
 
     def init_camera(self):
         # consts have changed locations between versions 2 and 3
-        if self.opencvMajorVersion == 3:
+        if self.opencvMajorVersion == 3 or self.opencvMajorVersion == 4:
```
inclement: Would you mind changing both of these to `if self.opencvMajorVersion in (3, 4)` etc., just for code neatness?

inclement:
Thanks!

PR #6120, opened 2019-02-01
Corrected textinput key input detection to only use on_textinput

inclement:
@JonasT pointed this mistake out when I made the original commit, but I misunderstood the code and failed to fix it. Sorry for doubting you @JonasT!


JonasT:
@inclement it's ok I also have vehemently defended blatantly wrong things often enough :joy:

JonasT:
(sorry this initially could have been read differently before the edit, I was really just poking fun at my own ability to have strong opinions even when sometimes I'm very wrong about it :grimacing: lol)

PR #6121, opened 2019-02-01
[widgets][VERY WIP] Start to implement themes

Sires0:
### What is it

This PR creates a new "fake module" called uixc, that will actually point to any other module on your disk, that's what I call of theme, this way you can customize widgets by downloading or creating themes(folders with the widgets kivy.uix has) and importing via `from kivy.uixc.label import Label` for example.

### How do I do it
Okay so my approach to implement themes is by using the `__path__` variable.

If you already know skip this paragraph. Every file that contains a `__path__` variable is already considerated a module, the `__path__` as you have guesses contains the path to the module files inside a list, like: ["/home/heyimcool/mypythonmodule"]. What I do here is detecting if config.ini has an entry called theme_path when loading the uixc module, if it is "/" keep `__path__` to the uix module, if not, set the `__path__` to the theme module.

Here themes would be folders inside ~/.kivy/themes(maybe someone can suggest a better path for putting themes but that one seems okay). Then in the config at section [kivy] you set theme_path = "/home/path/to/your/theme". This way the uixc file will point to your theme folder.

### Extras

For switching themes in runtime (I guess that) you have to import config, set theme_path to your other theme and reimport the uixc widgets.

Also for that a new config version was created, version 21.

Also it is nice for themes to just import the default widgets and customize them, this way if kivy fixes a bug or something in its widgets you don't need to update your code.

UIXC is for UIX Customized :) If you got a better name tell me, I'll gladly accept it.

There is a lot more to do, I'll write docs about it and possibly make theming kivy_garden a thing.

### Pros and Cons

Pros of this solution for theming(I'm not talking about the dillema, but just about overriding uix folders):
- Lightweight compared to some other attempts(like running a for overriding every widget)
- Highly customizable since you can code the widgets

Cons:
- Idk, if you find one please tell me :) I'll happilly add it

### See it yourself :)

If you want to have a quick demo, create a folder, ~~put an __init__.py and~~(the `__init__.py` is not called, don't worry about it) copy the button.py from uix folder, in the end add the line:
```python
text = StringProperty("lalala")
```
Now go to you kivy config and set the theme_path to the path of your folder, then go to kivy/kivy folder and there create a file with this code:
```python
from uix.button import Button as NButton
from uixc.button import Button as CButton


print(NButton().text, CButton().text)
```
You will see that the NButton(Normal button) will print nothing, but the button you imported via uixc will print "lalala".

### DIFF
--- a/kivy/uix/__init__.py
+++ b/kivy/uix/__init__.py
@@ -54,9 +54,3 @@ Kivy widgets can be categorized as follows:
 
 ----
 '''
-from kivy.config import Config
-
-
-__theme_path = Config.get("kivy", "theme_path")
-if __theme_path != "/":
-    __path__ = [__theme_path]
--- a/kivy/uixc.py
+++ b/kivy/uixc.py
@@ -1,4 +1,9 @@
 from os import getcwd
+from kivy.config import Config
 
 
-__path__ = [getcwd() + "/uix"]
+__theme_path = Config.get("kivy", "theme_path")
+if __theme_path == "/":
+    __path__ = [getcwd() + "/uix"]
+else:
+    __path__ = [__theme_path]


tshirtman:
I like that it's simple and doesn't add (much) complexity to kivy itself

The obvious drawback is that themes developers will have no "contract" with us, they will implement their own widgets (even if it copies our api at some point), and users will be able to use these, but there is no absolute portability between "themes", they have to be considered like completely different sets of widgets. An application developer develop for a set of widget, and if it's not the default one, they probably won't benefit from any fix we do to the kivy widgets, unless the theme maintainer ports it back.

Considering thing, it'll potentially create additional work for us to triage between bugs in kivy, and bugs in themes, especially if these limitations are not well understood by the users.

Additionally, i do wonder about the data part of kivy's uix, i.e, style.kv, and the defaulttheme's atlas, that are still going to be loaded by this PR even if the user selects a different theme.

Sires0:
On my idea I wanted theme developers would only make the widgets supported in kivy's uix :( 
Can't we just gently ask __Please do not develop other widgets than the ones at kivy.uix__ ?

About your additional question, it probably will, if people don't develop their own widgets from scratch instead of copying kivy ones then it will probably load the default theme. I can probably "fix" that by also, gonna code it so I can explain better.

matham:
I also like the simplicity of this.

However, I wonder what the benefit is of integrating this with kivy? Why not just have these widgets be like garden flowers? Or some other uix package that people import directly. E.g. `from kivy_widgets.label import ...`, like flat kivy did.

Sires0:
Well, in my utopic dreams we would gently ask: __DO NOT CREATE ANY WIDGET OTHER THAN THE ONES IN KIVY.UIXXXX__ and it would work :/ But if it doesn't we can just ask people to do not post issues while using themes. Oh, or were you talking about developing their own code for making widgets? If so we could try asking to make their theme widgets to inherit the original ones. I plan to make a theme template(a folder with some nice code following the idea of just inheriting the original widgets) that people should follow later with some instructions, I think it would help.

About the additional question, people can add their resource paths so that things load from there not from the default one. Also they can change the texture sources. Welp the files would still be loaded if one uses both uix and uixc widgets.

Sires0:
> I also like the simplicity of this.
> 
> However, I wonder what the benefit is of integrating this with kivy? Why not just have these widgets be like garden flowers? Or some other uix package that people import directly. E.g. `from kivy_widgets.label import ...`, like flat kivy did.

Not sure if I got what you said right but are you asking about a kivy garden flower per theme or one kivy garden package for loading themes? Not really sure

If you mean making a "flower"(sorry never heard this theme before xD I never really used kivy garden) per theme, there is an advantage about my method, it lets the user customize your app, like, I prefer adwaita-dark theme and other prefers kivy-md theme, this way they change the theme_path and all other kivy apps using uixc would integrate. Also if there is no theme available it falls back to kivy default widgets.

PR #6122, opened 2019-02-01
[core/image]Allow saving a core Image into BytesIO

tito:
This PR is a first step to get any image content as a BytesIO instead of just saving into a file.

It is seperated as:
- make the `Image` core provider check if a provider can save into `BytesIO`, and if the format is supported
- add `save()` fmt parameter to explicitly specify an image format (extension actually), needed if `BytesIO` is passed
- implement support for `BytesIO` in SDL2 Image provider using `SDL_RWops`
- add `Widget.export_as_image` and change `Widget.export_to_png` to use the first one
- update `Texture.save` to allow using a format (it's honestly just a shortcut to `Image(texture).save(...)`

### DIFF
--- a/kivy/tests/test_image.py
+++ b/kivy/tests/test_image.py
@@ -2,6 +2,7 @@ import unittest
 import io
 import os
 import tempfile
+from kivy import setupconfig
 
 
 class ImageTestCase(unittest.TestCase):
@@ -25,6 +26,11 @@ class ImageTestCase(unittest.TestCase):
 
     def test_save_into_bytesio(self):
         Image = self.cls
+
+        if setupconfig.PLATFORM == "darwin":
+            # XXX on OSX CI Builder, img_sdl2 is not used
+            # therefore the test below wont work yet with imageio only.
+            return
 
         # load kivy logo
         img = Image.load("data/logo/kivy-icon-512.png")


kivy/core/image/__init__.py
```diff
@@ -862,12 +878,24 @@ def save(self, filename, flipped=False):
...
         else:
             raise Exception('Unable to determine the format of the pixels')
-        return loader.save(filename, size[0], size[1], fmt, pixels, flipped)
+        return loader.save(filename, size[0], size[1], pixelfmt, pixels, flipped, fmt)
+
+    def _find_format_from_filename(self, filename):
+        ext = filename.rsplit(".", 1)[-1].lower()
+        if ext in (
+            'bmp', 'jpe', 'lbm', 'pcx', 'png', 'pnm',
+            'tga', 'tiff', 'webp', 'xcf', 'xpm', 'xv'):
```
AndreMiras: Minor taste suggestion: I find it more sexy to use sets for that type of things:
```python
        if ext in {
            'bmp', 'jpe', 'lbm', 'pcx', 'png', 'pnm',
            'tga', 'tiff', 'webp', 'xcf', 'xpm', 'xv'}:
```
Specially since it's `O(1)`

kivy/tests/test_image.py
```diff
@@ -18,3 +21,35 @@ def test_keep_data(self):
...
+        img = Image.load("data/logo/kivy-icon-512.png")
+        self.assertIsNotNone(img)
+
+        # try to save with missing format
+        try:
+            bio = io.BytesIO()
+            self.assertTrue(img.save(bio))  # if False, then there is no provider
+            self.fail('Image.save() with bytesio')
+        except Exception:
+            pass
```
AndreMiras: I quite don't get this block :thinking:

kivy/core/image/__init__.py
```diff
@@ -830,10 +833,23 @@ def save(self, filename, flipped=False):
...
 
+        .. versionchanged:: 1.11.0
+            Parameter `fmt` added to force the output format of the file
+            Filename can now be a BytesIO object.
+
         '''
+        if isinstance(filename, BytesIO):
+            is_bytesio = True
+            if not fmt:
+                raise Exception("You must specify a format to save into a BytesIO object")
```
AndreMiras: is this covered by the unit tests?

kivy/tests/test_image.py
KeyWeeUsr: `with self.assertRaises(exc_cls)`

kivy/core/image/__init__.py
AndreMiras: Ok thanks! This what I was suspecting. Then this is usually covered with `assertRaises` not try/catch. Because it makes things more obvious/explicit. I will take another look once on laptop

PR #6124, opened 2019-02-01
Removed doc note about Python 3 on Android being experimental

inclement:


JonasT:
@inclement what about a note that Python 2 will be officially unsupported (by the Python developers, not kivy/p4a) by the start of 2020? That seems like it would be worth mentioning

Issue #6132, opened 2019-02-04
no automatic dispatching of Window's left/top


tshirtman:
### Versions

* Python:
* OS:
* Kivy:
* Kivy installation method:

### Description

reacting to window's position changes is hard, comparatively to reacting to window's size, the alias properties work, but don't bind to the underlying state, so one has to check their values repeatedly.

### Code and Logs

```python
from kivy.core.window import Window as W
from kivy.app import App
from kivy.lang import Builder
from kivy.clock import Clock


KV = '''
#:import W kivy.core.window.Window
Label:
    text: 'top: {}\\nleft: {}\\nwidth: {}\\nheight: {}'.format(W.top, W.left, W.width, W.height)
'''

class WApp(App):
    def build(self):
        W.bind(top=self.log, left=self.log, size=self.log)
        Clock.schedule_interval(self.show_pos, 0)
        return Builder.load_string(KV)

    def show_pos(self, *args):
        print(W.left, W.top)

    def log(self, *args):
        print(args)

    def on_stop(self):
        print (
            'top: {}\nleft: {}\nwidth: {}\nheight: {}'
            .format(W.top, W.left, W.width, W.height)
        )

if __name__ == '__main__':
    WApp().run()
```

matham:
Fixed by https://github.com/kivy/kivy/pull/6133

PR #6133, opened 2019-02-04
Make top/left of window dispatch events on updates

tshirtman:
tentative fix for #6132 , but doesn't seem to work


matham:
What's the status of this?

akshayaurora:
@matham  did some changes tested with window_sdl2 on osx and It works as expected.

kivy/properties.pyx
```diff
@@ -1466,11 +1466,8 @@ cdef class AliasProperty(Property):
     cpdef link_deps(self, EventDispatcher obj, str name):
         cdef Property oprop
         for prop in self.bind_objects:
-            if prop in obj.__events__:
-                obj.fbind(prop, self.trigger_change, 0)
-            else:
-                oprop = getattr(obj.__class__, prop)
-                oprop.fbind(obj, self.trigger_change, 0)
```
akshayaurora: sry, for removing this, this was indeed needed.

kivy/core/window/__init__.py
```diff
@@ -1431,6 +1433,9 @@ def on_resize(self, width, height):
         '''Event called when the window is resized.'''
         self.update_viewport()
 
+    def on_move(self):
+        self._moved = not self._moved
+
```
matham: Why do you need the `_moved` property? Could you not just instead directly dispatch `top`  and `left` here (`self.dispatch('top')`).

Issue #6136, opened 2019-02-06
App crashes when trying to get Window info in master


OptimusGREEN:
### Versions

* Python: 2.7
* OS: Android
* Kivy: master
* Kivy installation method: pip

### Description

When trying to get Window info (Window.width etc...) during startup of an app on Android, if using the master branch of kivy then the app crashes as there is no window yet so a None Type error occurs.

This doesn't happen in 1.10.1, it works with no errors in the current stable version of kivy.

This error doesn't occur with master in my build env which is currently linux, only on android.

This issue was briefly discussed here https://github.com/kivy/python-for-android/issues/1655

1.10.1:
apk: [http://optimusgreen.com/dev_test/window_test_kivy1.10.1.apk](url)
log: [https://paste.city/141](url)

master:
apk: [http://optimusgreen.com/dev_test/window_test_kivy_master.apk](url)
log: [https://paste.city/142](url)

The below code is just an example app I made to show the difference in behaviour between 1.10.1 and master. I'm aware that Window.softinput_mode isn't doing anything here.

### Code and Logs

```
import kivy
from kivy.app import App
from kivy.core.window import Window
from kivy.uix.boxlayout import BoxLayout
from kivy.uix.label import Label



kivy.require('1.10.1')
Window.softinput_mode = 'below_target'

class RootWidget(BoxLayout):

    def __init__(self):
        super(RootWidget, self).__init__()

        label = Label(text="This is a test")
        self.add_widget(label)


class WindowTest(App):

    def build(self):
        root = RootWidget()
        return root


if __name__ == '__main__':
    WindowTest().run()
```

OptimusGREEN:
this pr fixes this for me so closing this issue
https://github.com/kivy/python-for-android/pull/1692

PR #6138, opened 2019-02-07
Fix android's sensor orientation

opacam:
Due to the recent sdl2 updates we loosed the `sensor` orientation in android platform.

Resolves: kivy/python-for-android#1688 (this was reported in python-for-android's git but it seems that it is a kivy's issue after all)

Note: I made this pull request starting at commit b47f669 of kivy's master branch, because we use that commit as a version in python-for-android's kivy recipe, as this is the last known kivy's version that works for python-for-android (see issue #6136). This way we can test this pr avoiding that issue.

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -117,9 +117,11 @@ cdef class _WindowSDL2Storage:
                 orientations = 'Portrait'
             elif env_orientations == 'landscape':
                 orientations = 'LandscapeLeft'
+            elif env_orientations == 'sensor':
+                orientations = 'sensor'
             else:
                 Logger.warning(('Could not satisfy Android orientation "{}", only '
-                               '{{portrait,landscape}} are currently supported. '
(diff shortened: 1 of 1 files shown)


opacam:
Yes, it works, I tested for python2 and python3 modifying p4a's test apps and the kivy recipe (l also tested in one of my apps which had that problem, python3).

Here are a `gist` with a patch to test this pr with some p4a's `testapps` (the ones with openssl and sqlite):

https://gist.github.com/opacam/8ffc6ae140bee1ef74bc1e5a96a9bc01

inclement:
Okay, that is the code I was referring to above, but I didn't realise that if it fails to set anything the default is already to use the sensor. That's okay then, thanks for the fix.

inclement:
I hope you don't find the welcome bot too patronising :<

JonasT:
@inclement @opacam I'm wondering, wouldn't `user` orientation also need this sort of fix? (I'm not using kivy so I can't test, but I wondered since `user` is what I usually use and seems the most natural to me since it essentially obeys whatever the user sets)

PR #6146, opened 2019-02-08
Fixes travis FBO tests

tito:
The current FBO tests were calling render() function, but Fbo.draw() was already working. In addition, the render() output potentally the window output to the Fbo, which screw it in the end.

Just fbo.draw() and check the content works as intented.


opacam:
I close this pull request, because @tito asked me to move this work into pull request #6148, so we can have this fix and the pep8 fixes I made in there together

PR #6148, opened 2019-02-09
Fix some travis errors

opacam:
- The current `FBO` tests were calling render() function, but `Fbo.draw()` was already working. In addition, the `render()` output potentially the window output to the Fbo, which screw it in the end. Just `fbo.draw()` and check the content works as intended (Thanks @tito).
  - Fix kivy's initialization error on graphics test for linux: `Unable to find any valuable Cutbuffer provider`
    (This is due to a missing dependency for linux `xclip` or `xsel`)
     See also: https://travis-ci.org/kivy/kivy/jobs/490796639
  - Fix pep8 E501: line too long (`filechooser` and `camera_opencv`)
  - Fix pep8 E121: continuation line under-indented for hanging indent (`setup.py`)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ install:
         sudo apt-get -y install libgstreamer1.0-dev gstreamer1.0-alsa gstreamer1.0-plugins-base;
         sudo apt-get -y install python-dev libsmpeg-dev libswscale-dev libavformat-dev libavcodec-dev libjpeg-dev libtiff4-dev libX11-dev libmtdev-dev;
         sudo apt-get -y install python-setuptools build-essential libgl1-mesa-dev libgles2-mesa-dev;
-        sudo apt-get -y install xvfb pulseaudio;
+        sudo apt-get -y install xvfb pulseaudio xsel;
         pip install --upgrade cython pillow nose coveralls docutils;
       fi;
       if [ "${RUN}" == "docs" ]; then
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -70,7 +70,7 @@ install:
         sudo apt-get -y install libgstreamer1.0-dev gstreamer1.0-alsa gstreamer1.0-plugins-base;
         sudo apt-get -y install python-dev libsmpeg-dev libswscale-dev libavformat-dev libavcodec-dev libjpeg-dev libtiff4-dev libX11-dev libmtdev-dev;
         sudo apt-get -y install python-setuptools build-essential libgl1-mesa-dev libgles2-mesa-dev;
-        sudo apt-get -y install xvfb pulseaudio;
+        sudo apt-get -y install xvfb pulseaudio xsel;
```
opacam: Here I use `xsel` but we can also use `xclip`. There is no special reason to use `xsel`, I used that because is the package that I use in my development machine...
should we use `xclip`?

opacam: I forgot, we also could drop these changes, I know that the test is ok without installing anything more...so actually we don't need that because does not solve any travis error, but I think that makes easier to find the real problem...
Here a test without `xsel` (scroll to end):
  - https://travis-ci.org/kivy/kivy/jobs/490796639

And here one with `xsel` (scroll to end):
  - https://travis-ci.org/kivy/kivy/jobs/490991404

PR #6156, opened 2019-02-12
Adding fixes to support ios camera

GoBig87:
Fixing issues with camera support for ios.

### DIFF
--- a/kivy/core/camera/camera_avfoundation.pyx
+++ b/kivy/core/camera/camera_avfoundation.pyx
@@ -62,7 +62,7 @@ class CameraAVFoundation(CameraBase):
         self._resolution = (width, height)
 
         if self._texture is None or self._texture.size != self._resolution:
-            self._texture = Texture.create(self._resolution)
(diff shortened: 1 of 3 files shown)


kivy/core/camera/camera_avfoundation.pyx
```diff
@@ -62,7 +62,7 @@ class CameraAVFoundation(CameraBase):
         self._resolution = (width, height)
 
         if self._texture is None or self._texture.size != self._resolution:
-            self._texture = Texture.create(self._resolution)
+            self._texture = Texture.create(self._resolution, colorfmt='bgra')
```
tito: Looks very easy to add it, however, the pixel format is different than the default. Is that also the case for OSX?
I think this part may break OSX, so it should be per platform

tshirtman:
This PR was broken :( see commit https://github.com/kivy/kivy/commit/3d14a1a8606d3bf81fdc0c5367d29cdf66f01b5c

PR #6163, opened 2019-02-13
Changed default orientation from LandscapeLeft to LandscapeRight

inclement:


inclement:
Deprecated by https://github.com/kivy/kivy/pull/6164

PR #6164, opened 2019-02-13
Removed default orientation hints on Android

inclement:
Adding these was a mistake, I didn't realise that SDL2 will use the AndroidManifest.xml orientation hints by default because that behaviour was obscured by Kivy's existing default hints.

Removing the use of P4A_ORIENTATION and using a blank hint by default should fix this. The user can still override it SDL2-style if they want to.

Fixes https://github.com/kivy/python-for-android/issues/1698

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -109,22 +109,10 @@ cdef class _WindowSDL2Storage:
         if USE_IOS:
             orientations = 'LandscapeLeft LandscapeRight Portrait PortraitUpsideDown'
 
-        # Override the Android orientation based on what p4a was built with, if available.
-        # This may be overridden again by $KIVY_ORIENTATION.
         if USE_ANDROID:
-            env_orientations = environ.get('P4A_ORIENTATION', orientations)
-
-            if env_orientations == 'portrait':
-                orientations = 'Portrait'
-            elif env_orientations == 'landscape':
-                orientations = 'LandscapeLeft'
-            elif env_orientations == 'sensor':
-                orientations = 'sensor'
-            else:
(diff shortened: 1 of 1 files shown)


JonasT:
Makes sense :+1: sorry I didn't suggest this at some point, it could have dawned to me this was the actual issue but it never did :grimacing: lol

opacam:
Thanks @inclement,

**Note:** if this does not work, take a look on this kivy/python-for-android#1708, because it fix the problem from p4a side (if this works, then we maybe should close the mentioned pr, for now...I put it in `WIP` status)

inclement:
Thanks for your efforts here @opacam, fortunately this way seems to work well.

inclement:
I've tested that this works, it looks good.

kivy/core/window/_window_sdl2.pyx
```diff
@@ -109,22 +109,10 @@ cdef class _WindowSDL2Storage:
...
-            elif env_orientations == 'sensor':
-                orientations = 'sensor'
-            else:
-                Logger.warning(('Could not satisfy Android orientation "{}", only '
-                               '{{portrait,landscape and sensor}} are currently supported. '
-                                'Defaulting to portrait').format(orientations))
-                orientations = 'Portrait'
+            # Do not hint anything: by default the value in the AndroidManifest.xml will be used
+            # Note that the user can still override this via $KIVY_ORIENTATION if they wish
+            orientations = ''
```
JonasT: Haven't tested it but @inclement has, and I use a similar approach in my non-kivy app and this looks like the right thing to do(TM) to me. :+1:

opacam: I've tested it and confirm that it works.

¡¡¡Good work, @inclement and @JonasT!!! :+1:

PR #6165, opened 2019-02-16
Include GStreamer in PyInstaller package

### LINKED ISSUE #6126: PyInstaller package does not include GStreamer (solved)
### Versions

* Python: 2.7.15
* OS: macOS 10.14.1 Mojave
* Kivy: 1.10.1
* Kivy installation method: `pip install kivy --no-binary kivy` in virtualenv
* GStreamer 1.14.4 (homebrew; gst-plugins-* also installed)

### Description

When packaging an app that uses `kivy.core.audio.SoundLoader`, PyInstaller does not include the GStreamer plugins or libraries they require in the package. As a result, trying to play a sound results in `GstPlayerException: Unable to create a playbin`. (The unpackaged app works fine).

I think there is code missing from the kivy PyInstaller hook that would find and include the plugin files. The runtime hook includes a directory 'gst-plugins' in the GST_PLUGIN_PATH, but that path isn't mentioned anywhere else.

I solved the issue by adding the following code to my spec file:

```py
def find_gst_plugin_path():
    p = subprocess.Popen(
        ['gst-inspect-1.0', 'coreelements'],
        stdout=subprocess.PIPE)
    (stdoutdata, stderrdata) = p.communicate()

    match = re.search(r'\s+(\S+libgstcoreelements\.\S+)', stdoutdata)

    if not match:
        raise Exception('could not find GStreamer plugins')

    return os.path.dirname(match.group(1))


def find_gst_binaries():
    plugin_path = find_gst_plugin_path()

    plugin_filepaths = glob.glob(os.path.join(plugin_path, 'libgst*'))

   # PyInstaller does not automatically include the libraries the plugins link to
    lib_filepaths = set()
    for plugin_filepath in plugin_filepaths:
        plugin_deps = bindepend.selectImports(plugin_filepath)
        lib_filepaths.update([path for _, path in plugin_deps])

    plugin_binaries = [(f, 'gst-plugins') for f in plugin_filepaths]
    lib_binaries = [(f, '.') for f in lib_filepaths]

    return plugin_binaries + lib_binaries

# SNIP
a = Analysis(['../testapp/main.py'],
             pathex=['.'],
             binaries=find_gst_binaries(),
# SNIP
```

### Code and Logs

Here's my test app:
https://github.com/brsaylor/soundloader-test

(It requires a command line argument with the sound file to play).

Deleting the `find_gst_binaries()` code demonstrates the issue.

brsaylor:
Closes #6126.

This changes `get_deps_minimal` in `kivy.tools.packaging.pyinstaller_hooks`  to include GStreamer plugins and libraries when `gstplayer` is given as a provider to include. It also changes `get_deps_all` to always include GStreamer, which seems to match the intent of that function.

Also fixes an issue with `GST_PLUGIN_PATH` in the runtime hook.

Tested only on MacOS with GStreamer installed using homebrew. There are a couple of things I'm not sure about:

- How does the `kivy.deps.gstreamer` package factor in? Seems like it's Windows-only; should we disable this fix on Windows?

- When I `pip install kivy`, the package gets installed with a `.dylibs` directory that includes `libgstreamer-1.0.0.dylib`. I got errors when the plugins were loading about a GST version mismatch. `pip install kivy --no-binary kivy` fixed this issue for me. I'm not sure about the implications of this.

Sample spec file:

```python
# -*- mode: python -*-

block_cipher = None
from kivy.tools.packaging.pyinstaller_hooks import get_deps_all, get_deps_minimal, hookspath, runtime_hooks

kivy_deps = get_deps_minimal(video=None, window=True, audio=['gstplayer'])
# kivy_deps = get_deps_all()

a = Analysis(['../testapp/main.py'],
             pathex=['.'],
             datas=[],
             hookspath=hookspath(),
             runtime_hooks=runtime_hooks(),
             win_no_prefer_redirects=False,
             win_private_assemblies=False,
             cipher=block_cipher,
             noarchive=False,
             **kivy_deps)
pyz = PYZ(a.pure, a.zipped_data,
             cipher=block_cipher)
exe = EXE(pyz,
          a.scripts,
          [],
          exclude_binaries=True,
          name='soundloader-test',
          debug=False,
          bootloader_ignore_signals=False,
          strip=False,
          upx=True,
          console=False )
coll = COLLECT(exe, Tree('../testapp/'),
               a.binaries,
               a.zipfiles,
               a.datas,
               strip=False,
               upx=True,
               name='soundloader-test')
app = BUNDLE(coll,
             name='soundloader-test.app',
             icon=None,
             bundle_identifier=None)
```

### DIFF
--- a/kivy/tools/packaging/pyinstaller_hooks/__init__.py
+++ b/kivy/tools/packaging/pyinstaller_hooks/__init__.py
@@ -64,10 +64,14 @@ import pkgutil
 import logging
 from os.path import dirname, join
 import importlib
+import subprocess
+import re
+import glob
 
 import kivy
 import kivy.deps
 from kivy.factory import Factory
+from PyInstaller.depend import bindepend
 
 from os import environ
 if 'KIVY_DOC' not in environ:
@@ -163,9 +167,10 @@ def get_deps_minimal(exclude_ignored=True, **kwargs):
 
     :returns:
 
-        A dict with two keys, ``hiddenimports`` and ``excludes``. Their values
-        are a list of the corresponding modules to include/exclude. This can
-        be passed directly to `Analysis`` with e.g. ::
+        A dict with three keys, ``hiddenimports``, ``excludes``, and
+        ``binaries``. Their values are a list of the corresponding modules to
+        include/exclude. This can be passed directly to `Analysis`` with
+        e.g. ::
 
             a = Analysis(['..\\kivy\\examples\\demo\\touchtracer\\main.py'],
                         ...
@@ -229,9 +234,17 @@ def get_deps_minimal(exclude_ignored=True, **kwargs):
 
     mods = sorted(set(mods))
 
-    if exclude_ignored and not any('gstplayer' in m for m in mods):
+    binaries = []
+    if any('gstplayer' in m for m in mods):
+        binaries = _find_gst_binaries()
+    elif exclude_ignored:
         excludes.append('kivy.lib.gstplayer')
-    return {'hiddenimports': mods, 'excludes': excludes}
+
+    return {
+        'hiddenimports': mods,
+        'excludes': excludes,
+        'binaries': binaries,
+    }
 
 
 def get_deps_all():
@@ -245,15 +258,17 @@ def get_deps_all():
 
     :returns:
 
-        A dict with two keys, ``hiddenimports`` and ``excludes``. Their values
-        are a list of the corresponding modules to include/exclude. This can
-        be passed directly to `Analysis`` with e.g. ::
+        A dict with three keys, ``hiddenimports``, ``excludes``, and
+        ``binaries``. Their values are a list of the corresponding modules to
+        include/exclude. This can be passed directly to `Analysis`` with
+        e.g. ::
 
             a = Analysis(['..\\kivy\\examples\\demo\\touchtracer\\main.py'],
                         ...
                          **get_deps_all())
     '''
     return {
+        'binaries': _find_gst_binaries(),
         'hiddenimports': sorted(set(kivy_modules +
                                     collect_submodules('kivy.core'))),
         'excludes': []}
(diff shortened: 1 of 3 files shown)


kivy/tools/packaging/pyinstaller_hooks/pyi_rth_kivy.py
```diff
@@ -5,8 +5,7 @@
 
 os.environ['KIVY_DATA_DIR'] = os.path.join(root, 'data')
 os.environ['KIVY_MODULES_DIR'] = os.path.join(root, 'modules')
-os.environ['GST_PLUGIN_PATH'] = '{};{}'.format(
-    sys._MEIPASS, os.path.join(sys._MEIPASS, 'gst-plugins'))
+os.environ['GST_PLUGIN_PATH'] = os.path.join(sys._MEIPASS, 'gst-plugins')
```
tito: Very nice!

kivy/tools/packaging/pyinstaller_hooks/__init__.py
```diff
@@ -64,10 +64,14 @@
...
 from os.path import dirname, join
 import importlib
+import subprocess
+import re
+import glob
 
 import kivy
 import kivy.deps
 from kivy.factory import Factory
+from PyInstaller.depend import bindepend
```
tito: Are you sure about this?
Because it is failing with travis currently:
https://travis-ci.org/kivy/kivy/jobs/494325672

brsaylor: The purpose of that import is to find all the libraries that the GStreamer plugins link to, which otherwise aren't included in the package.

I just committed a change to .travis.yml that gets those jobs working again - how does that look? I see that a couple of the jobs are still failing, but are also failing on master.

kivy/tools/packaging/pyinstaller_hooks/pyi_rth_kivy.py
matham: Why this change? On Windows the plugins are not in a separate directory so the root needs to be on the path. Was there an error on osx that made you change this?

brsaylor: Ah, I didn't realize the plugins were stored in the package root in Windows. If I remember correctly, the error I was seeing on macOS was due to the semicolon instead of a colon, and I followed this PR: https://github.com/pyinstaller/pyinstaller/pull/1963

matham: No worries. This was fixed in https://github.com/kivy/kivy/pull/6359/.

PR #6168, opened 2019-02-21
fix broken update to avfoundation

tshirtman:
can't test it at all, but should at least fix one issue.


OptimusGREEN:
This fixes the error I was getting during the build process here 
https://paste.city/162

tito:
SORRY, i merged, but though travis was ok.

PR #6172, opened 2019-02-23
Slave mode testing

KeyWeeUsr:
Add a public API for parts of Kivy main loop, fix slave mode and cleaning the context, add some example test with and without managing your own main loop.

### DIFF
--- a/doc/sources/guide/environment.rst
+++ b/doc/sources/guide/environment.rst
@@ -82,6 +82,16 @@ KIVY_NO_ARGS
 
     .. versionadded:: 1.9.0
 
+KIVY_SLAVE_MODE
+    If set, Kivy will not run its main loop, nor create the default Window
+    on using ``kivy.core.window`` import.
+
+    Check :class:`~kivy.core.window.WindowBase` documentation for `slave` mode
+    related methods and :mod:`~kivy.core.flow` to manually call parts of the
+    Kivy main loop yourself.
+
+    .. versionadded:: 1.11.0
+
 Restrict core to specific implementation
 ----------------------------------------
 
--- a/examples/application/slavemode/simpleapp.py
+++ b/examples/application/slavemode/simpleapp.py
@@ -1,0 +1,49 @@
+'''
+Simple Application with Clock to showcase slave mode, getting to the next
+frame and Clock ticking/scheduling.
+
+.. versionadded:: 1.11.0
+'''
+
+from kivy.app import App
+from kivy.uix.button import Button
+from kivy.clock import Clock
+
+
+class Simple(App):
+    '''
+    Main application class.
+
+    .. versionadded:: 1.11.0
+    '''
+
+    TEXT_DEFAULT = 'default'
+    TEXT_CHANGED = 'changed'
+    INTERVAL = 0.5
+
+    @staticmethod
+    def change_text(widget):
+        '''
+        Change the text attribute of a widget.
+
+        .. versionadded:: 1.11.0
+        '''
+        widget.text = Simple.TEXT_CHANGED
+
+    def build(self):
+        '''
+        Kivy build method.
+
+        .. versionadded:: 1.11.0
+        '''
+
+        widget = Button(text=Simple.TEXT_DEFAULT)
+        Clock.schedule_once(
+            lambda dt: self.change_text(widget),
+            Simple.INTERVAL
+        )
+        return widget
+
+
+if __name__ == '__main__':
+    Simple().run()
--- a/examples/application/slavemode/test_slave.py
+++ b/examples/application/slavemode/test_slave.py
@@ -1,0 +1,170 @@
+'''
+Test showing using a slave mode of Kivy main loop while testing
+a root widget from an application changing its text after some
+time period.
+
+.. versionadded:: 1.11.0
+'''
+
+import unittest
+from os import environ
+
+
+class SimpleTestCase(unittest.TestCase):
+    '''
+    Simple TestCase setting & unsetting slave mode and showcasing
+    testing with runTouchApp() + stopTouchApp() with a Window being
+    opened and closed and testing with a custom application main loop
+    and Kivy being run in the slave mode.
+
+    .. versionadded:: 1.11.0
+    '''
+
+    @classmethod
+    def setUpClass(cls):
+        '''
+        Set slave mode just for this test case.
+
+        Any tests running in normal mode will most likely have
+        global variables present, therefore it's reasonable
+        to assume failure while mixing the modes.
+
+        .. versionadded:: 1.11.0
+        '''
+        environ['KIVY_SLAVE_MODE'] = '1'
+
+    @classmethod
+    def tearDownClass(cls):
+        '''
+        Clean slave mode flag from the environment.
+
+        It should allow any normal mode tests to run properly even if
+        a slave mode test ran prior to normal ones due to slave mode
+        not importing the ``kivy.core.window`` module.
+
+        .. versionadded:: 1.11.0
+        '''
+        if 'KIVY_SLAVE_MODE' in environ:
+            del environ['KIVY_SLAVE_MODE']
+
+    def test_widget(self):
+        # pylint: disable=protected-access
+        '''
+        Test a root widget by using slave mode, spawning your own Window
+        manually and pushing the frames forward by yourself with the
+        EventLoop.idle().
+
+        .. versionadded:: 1.11.0
+        '''
+
+        from kivy.clock import Clock
+        from kivy.base import EventLoop
+        from kivy.app import runTouchApp, stopTouchApp
+        from simpleapp import Simple
+
+        # get OpenGL window
+        import kivy.core.window
+        kivy.core.window.Window = kivy.core.window.get_core_window()
+        EventLoop.ensure_window()
+
+        # get widget from App instance and run it
+        app = Simple()
+        widget = app.build()
+        runTouchApp(widget)
+
+        # manage the event loop
+        frames = 60
+        delta = 5
+        for frame in range(frames):
+            EventLoop.idle()
+
+            # on 30th frame, i.e. on 0.5s the Clock.schedule_once
+            # is triggered and the text has been changed since
+            # however it's not accurate to the single frame,
+            # therefore approximate with some delta
+            if frame < Clock._max_fps * app.INTERVAL - delta:
+                self.assertEqual(widget.text, app.TEXT_DEFAULT)
+            elif frame > Clock._max_fps * app.INTERVAL + delta:
+                self.assertEqual(widget.text, app.TEXT_CHANGED)
+
+        # destroy the window and stop eventloop
+        EventLoop.window.close()
+        stopTouchApp()
+
+    def mainloop_widget_hook(self, frame, fps, widget, app):
+        '''
+        Custom main loop hook used for testing.
+
+        .. versionadded:: 1.11.0
+        '''
+
+        delta = 5
+        if frame < fps * app.INTERVAL + delta:
+            self.assertEqual(widget.text, app.TEXT_DEFAULT)
+        elif frame > fps * app.INTERVAL - delta:
+            self.assertEqual(widget.text, app.TEXT_CHANGED)
+
+    def test_mainloop_widget(self):
+        # pylint: disable=no-else-return,protected-access,too-many-locals
+        '''
+        Test a root widget by using slave mode, spawning your own Window
+        manually and pushing the frames forward by yourself with the
+        EventLoop.idle() in your own main loop while polling for the Window
+        events both during the pause mode and active mode.
+
+        .. versionadded:: 1.11.0
+        '''
+
+        from kivy.clock import Clock
+        from kivy.base import EventLoop
+        from kivy.app import runTouchApp, stopTouchApp
+        from kivy.core.flow import FlowReturn, FlowBreak, FlowContinue
+        from simpleapp import Simple
+
+        # get OpenGL window
+        import kivy.core.window
+        kivy.core.window.Window = kivy.core.window.get_core_window()
+        EventLoop.ensure_window()
+
+        # get App instance and run it
+        app = Simple()
+        widget = app.build()
+        runTouchApp(widget)
+
+        # manage the event loop
+        #
+        # two while loops because of a Window main loop (polling one)
+        # is being wrapped with an event loop (e.g. for EventDispatcher)
+        current_frame = 0
+        while current_frame < 60:
+            while current_frame < 60:
+                current_frame += 1
+                EventLoop.idle()
+
+                flow = EventLoop.window.slave_poll()
+                flow_type = type(flow)
+                if flow_type is FlowReturn:
+                    return flow.return_value
+                elif flow_type is FlowBreak:
+                    break
+                elif flow_type is FlowContinue:
+                    continue
+
+                # test in custom main loop too
+                # note:
+                #     generally you would want some hook function
+                #     instead of just copy-pasting your test in here
+                #
+                # in case of premature errors / segfaults use pdb.set_trace()
+                self.mainloop_widget_hook(
+                    current_frame, Clock._max_fps,
+                    widget, app
+                )
+
+        # destroy the window and stop eventloop
+        EventLoop.window.close()
+        stopTouchApp()
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/kivy/base.py
+++ b/kivy/base.py
@@ -170,6 +170,7 @@ class EventLoopBase(EventDispatcher):
         self.quit = True
         self.stop()
         self.status = 'closed'
+        self.window = None
 
     def stop(self):
         '''Stop all input providers and call callbacks registered using
--- a/kivy/base.py
+++ b/kivy/base.py
@@ -20,6 +20,7 @@ __all__ = (
 )
 
 import sys
+from os import environ
 from kivy.config import Config
 from kivy.logger import Logger
 from kivy.utils import platform
@@ -438,6 +439,7 @@ def runTouchApp(widget=None, slave=False):
 
     '''
 
+    slave = bool(environ.get('KIVY_SLAVE_MODE', False)) or slave
     from kivy.input import MotionEventFactory, kivy_postproc_modules
 
     # Ok, we got one widget, and we are not in slave mode
--- a/kivy/core/flow/__init__.py
+++ b/kivy/core/flow/__init__.py
@@ -1,0 +1,71 @@
+'''
+Kivy main loop flow
+===================
+
+To direct Kivy main loop flow we use Python keywords such as ``return``,
+ ``continue`` or ``break``. In `slave` mode however you either copy
+verbosely the entire Window main loop or the EventLoop main loop which
+is almost only another ``while True`` wrapper or you call only specific parts.
+
+While doing the former, i.e. copying verbosely, you will be in constant
+struggle to maintain 1-to-1 consistency with the Kivy main loop and hope
+something doesn't change between the releases.
+
+With these classes we can control the flow with a small overhead and give you
+an API to manually trigger a non-blocking Kivy main loop (i.e. without
+``while True`` on our side) constructs by keeping the Python keywords under
+control via various returned objects as a mark for the futrher action within
+the first level of our ``while True`` loop, therefore keeping the polling
+or other behaviors linear and completely non-related to the loop manipulation
+via ``return``, ``continue``, ``break``, etc.
+
+You can do a main loop in your app like this::
+
+    # trigger the slave mode for the framework
+    # !!! has to be done before importing anything from Kivy
+    from os import environ
+    environ['KIVY_SLAVE_MODE'] = '1'
+    from kivy.core.flow import FlowReturn, FlowBreak
+
+    # make sure OpenGL window is available
+    EventLoop.ensure_window()
+    window = EventLoop.window
+
+
+    def my_main_loop():
+        # your custom main loop
+        while True:
+            flow = window.slave_poll()
+            if isinstance(flow, FlowReturn):
+                return flow.return_value
+            elif isinstance(flow, FlowBreak):
+                break
+            ...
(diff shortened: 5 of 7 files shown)


tshirtman:
Can we name it `KIVY_MANAGED` or something alike? i'm fine with the current name if we don't find any other suitable one, but i'm pretty sure we can, and it would be better not to cause unnecessary bad feelings to users. Besides, if the effect is mostly to avoid any border side effects by default, and having to ask for everything, maybe `KIVY_MANAGED` would be a more apt description of the effect, or we could even go for something more explicit like `KIVY_PURE_IMPORT` or `KIVY_NO_LOOP`, though i still think `KIVY_MANAGED` is the best option, i'm open to other suggestions.

KeyWeeUsr:
First of all, in writing I sound more harsh, so take this with a little bit of salt:

> ... it would be better not to cause unnecessary bad feelings to users ...

There has already been `slave` mode present. In fact, *since the beginning of the project* and there hasn't been any reported issue with it. If and if only there are people **closely related to that term** and *they* feel uncomfortable or anything, then we might perhaps consider changing the terminology if the impact is significant. Otherwise, if it's not broken don't fix it and especially if the term does not include you because you can do more harm than help if you do stuff instead of the targeted people (I don't know if you are related to the term or not).

Ref [Indian vs Native American](https://www.quora.com/How-do-Native-Americans-feel-when-referred-to-as-Indians-or-American-Indians)(written by an Indian). I know some of them and they in fact prefer being called Indian instead of Native xyz to point out the stupidity of the naming itself (because someone thought he's in India). A similar thing is with Indians from India being also called something different before colonization and yet renamed. And a little bit of the word origins:

> Word History: The derivation of the word slave encapsulates a bit of European history and explains why the two words slave and Slav are so similar; they are, in fact, historically identical. The word slave first appears in English around 1290, spelled sclave. The spelling is based on Old French esclave from Medieval Latin sclavus, "Slav, slave," first recorded around 800. Sclavus comes from Byzantine Greek sklabos (pronounced sklä′vōs) "Slav," which appears around 580. Sklavos approximates the Slavs' own name for themselves, the Slověnci, surviving in English Slovene and Slovenian. The spelling of English slave, closer to its original Slavic form, first appears in English in the 1500s. Slavs became slaves around the beginning of the ninth century when the Holy Roman Empire tried to stabilize a German-Slav frontier. By the 1100s, stabilization had given way to wars of expansion and extermination that did not end until 1410, when the Poles crushed the knights of the Teutonic Order at Grunwald in north-central Poland. · As far as the Slavs' own self-designation goes, its meaning is, understandably, better than "slave"; it comes from the Indo-European root *kleu-, whose basic meaning is "to hear" and occurs in many derivatives meaning "renown, fame." The Slavs are thus "the famous people." Slavic names ending in -slav incorporate the same word, such as Czech Bohu-slav, "God's fame," Russian Msti-slav, "vengeful fame," and Polish Stani-slaw, "famous for withstanding (enemies)."
[source](https://www.thefreedictionary.com/slave)

I personally am a Slav from [Slovakia](https://en.wikipedia.org/wiki/Slovakia), young country which pre-republican history is full of other nations using it or including it into their, let's say "ownership" therefore if anything, this should concern me too and other Slavs and basically every other nation because slavery has been common in many places. ***But it doesn't.*** This is just to point out that the word itself has **many** meanings and even if some of them are simply said "a property of someone", it **doesn't** mean the word itself is wrong, bad or anything like that. Slave != what happened to slaves or how they were treated or anything like that. I understand your concerns and I think they are wrong and you should detach feelings from science and technology and leave the technological terms be technological terms.

Master/Slave are specific terms for a behavior [in technology](https://en.wikipedia.org/wiki/Master/slave_(technology)). Technology is meant to be **used**, not **felt** and slave has a specific meaning in this specific context.

> if the effect is mostly to avoid any border side effects by default, and having to ask for everything, maybe `KIVY_MANAGED` would be a more apt description of the effect, or we could even go for something more explicit like `KIVY_PURE_IMPORT` or `KIVY_NO_LOOP`, though i still think `KIVY_MANAGED` is the best option, i'm open to other suggestions.

Import/no loop - no, that's only a single part of the behavior. If anything, what is done with the mode is completely controlling and owning what and where happens and the closest alternative term to that is perhaps marionette, but not managed. You are not "managing" anything, in fact, once you activate the mode, Kivy is broken, unsustainable and dead unless you add event loop, create a window, poll window for events, etc etc. In this mode you force Kivy internals to do irregular actions and whatever you do to your runtime is your problem, your actions - therefore you control and own, therefore the best term is in fact slave.

> ... force selection and do something like `KIVY_BASE_MODE` with various options like `managed` and `standalone` ...

Still no, see the section above. Kivy is basically dead with this mode. There is nothing to *select*, you have to go and call each and every function we do in the internals in the correct order to have the same runtime as we have by default. After adding the slave mode to the rest of the providers if needed / requested Kivy will fail on each and every global variable it uses and each specific part of the internals we would add the mode to e.g. let's say compiling KV by yourself or ticking the Clock or ... With the slave mode I can manually run Kivy to have the exact same runtime as it's by default and alternatively I can bork it so much it'll segfault after I frown at it. Same as the master can behave to a slave. It's a behavior, relationship, not a definition of evil.

Still though, I was considering term "marionette" (inspired by Mozilla and Geckodriver) as an alternative yet that's still not describing the behavior properly imo. Other combinations I was thinking about were parent/child, teacher/pupil, yet either it sounds completely stupid in this context or it's not an obvious term and I believe everyone in the technological field has a clear idea what slave term in this context stands for.

tshirtman:
I have to take it with quite a large amount of sugar, considering the weight these kind of discussions tend to carry.

Yes, there is already such a mode present, thankfully in a minor feature, that i did want to rename, or remove, as previously discussed, and that — i believe — you opposed, as an unimportant change. There are bugs in the project still the begining, and typos in the documentations, and we take them out when we find them, because they are errors, i consider the inclusion of the term slave to be a mistake, harder to resolve because part of the api, but still an error, that should be corrected if possible, adding more API using the same term is only making it harder to fix it, that's why i'm kindly asking to consider alternatives.

I also know about the origin of the term, and your country of birth/residence, the fact that you don't feel strongly about it is less relevant though, because the relationship that defined that term is pretty dated, and your historical proximity to this practice is certainly as good as mine. There are, on the other hand, people whose grand parents where slaves, that have access to our technologies, and i imagine some of them would care about seeing the term used for trivial things like this, and no, they won't necessarily say it, they may just leave silently. Anyway, i'm not talking in their name, i do think it's bad practice to use the bad word just because it's been used historically, especially if it can be a distraction. Yes i *can* detach words from their meaning, but it's not as easy for people for which that meaning is a living reality (i don't need to tell you slavery exists to this day, and that technically there are more of them than ever were, in absolute amounts).

Now, considering the historical usage of the term, it's true that it's been used far and wide, but people have been talking of this as a problem for a few years now, not all people of course, not even people you could expect to care, but certainly some of them. A lot of projects have done the work to find better alternatives, sometime it's really hard to come up with a good replacement term, sometime it's not, i'm merely asking for best effort, here. So, is the term really the most semantically appropriate?

A slave can still have agency, ability to act at their own will, although they certainly would face retribution, does that fit our model? Not at all, as not many programs or framework can pretend to such autonomy of will, drawing an equivalent between a human being subject to a grave loss of freedom, to a theoretical machine, when you start it in a mode that requires very involved maintenance to even be functional, seems quite inappropriate, technically. The term marionette, in this regard, is much more suited, not saying that i like it much, but it's technically better.

By managed i mean that you need to manage the framework when running in this state, as opposed to the normal mode where you let it manage the running of your code, here, you need to control a lot more of its behavior, one definition that google provided me of "manage" is "be in charge of (a company, establishment, or undertaking); administer; run.", which seems quite a fit to what this mode means for the user, they have to manage kivy, completely, not just giving it orders, but caring about its internal state and being in charge of all the necessary actions, which is the opposite of delegating them to a slave.

I also must have expressed myself pretty badly about the KIVY_BASE_MODE, this allows to run kivy in the mode this PR describe, or any other alternative mode, should we discover a need of them (i don't know, maybe some future async loop will require such an half managed state), if you selected the managed mode using the environment variable, kivy would run in the mode this PR describe, and if not, it would run in the "normal" mode, that is the current behavior, to be perfectly clear, KIVY_BASE_MODE could accept the "slave" mode, and work just like this PR describe, this has nothing to do with the term itself, and is just a suggestion to give more flexibility in case other potential "modes" are discovered later.

KeyWeeUsr:
> ... you opposed, as an unimportant change. 

No. I opposed it as a nonsensical change. Especially because no one that's the target of the slave word either cared to talk back or give a damn in the first place. You can't solve problems for someone else especially if that problem is not related to you. It's similar to children - if you move the obstacles, it won't serve the purpose and the child won't learn from it.

> There are bugs in the project still the beginning, and typos in the documentations, and we take them out when we find them, because they are errors ...

This is not an **error** this is a ***technological term.*** Neither history, nor proper description is an error.

> i consider the inclusion of the term slave to be a mistake, harder to resolve because part of the api, but still an error, that should be corrected if possible, adding more API using the same term is only making it harder to fix it, that's why i'm kindly asking to consider alternatives.

It's not a mistake it's [***the first commit***](https://github.com/kivy/kivy/commit/1f2fb6eb41651e4d10cb51c24a24af66be37606d) in this repository and has been present in PyMT for [**10 years**](https://github.com/tito/pymt/commit/1da2a37fcb6be1a58901e5194537f290441bc51e).

> I also know about the origin of the term, and your country of birth/residence, the fact that you don't feel strongly about it is less relevant though ...

> There are, on the other hand, people whose grand parents where slaves...

Ok, so there is history and history. And one slave != some other slave. No comment...

> ... and i imagine some of them would care about seeing the term used for trivial things like this, and no, they won't necessarily say it, they may just leave silently.

And that is not **our** problem. You can't include everyone, can't solve every problem a single person has due to the state of mind, heritage, body features or whatever else.

> ... it's bad practice to use the bad word ...

It's not a bad word, it's a word with a specific meaning in this context and with other specific meanings in different contexts. Here it's a technological term. Do you want to go word after a word in the Oxford dictionary because someone might not like it? Would you like to fly in a plane with safety instructions written in pretty words or in strict and direct ones everyone can understand in a split of a second?

> ... technically there are more of them [slaves] than ever were, in absolute amounts

I was looking at this sentence and wondering if I even should reply to that, but... Any scientific evidence or you just *feel* that there is more slaves today than in the past? Every nation had had slaves in one shape or another and if not, then that particular nation had been enslaved. That particular sentence is just a piece of crap without any numbers, sources, articles, etc. Logically if you start summing up some percentage of the world's population in a current year and go back to the primitive ages or basically ages when the human mind had evolved enough to treat someone as a subordinate/slave, you can't exceed the number of slaves in this year. Even if you do sum by unique individuals instead of population percentage. And in *absolute* amounts, you must seriously be joking. Even today you can't get a proper number of people living in some countries depending on the inclusion of the people (ref Sān for example) and their evidence by governments.

The current time isn't an apocalypse, not everyone is a slave in the same shape as in the past, nor you can be serious saying the current amount of slaves in the world might be **even equal**, not even higher than in the whole past. Not even if you somehow manage to apply Orwell's novel to the half of the current population. In the ancient Greece were slaves even between neighbor cities themselves and a label "slave" was passed even to children born in the slavery. Slavery != selling/providing whatever useful skill you have in an exchange for a form of value holder to trade it for basic goods to satisfying your basic needs and not die.

> A lot of projects have done the work to find better alternatives, sometime it's really hard to come up with a good replacement term, sometime it's not, i'm merely asking for best effort, here. So, is the term really the most semantically appropriate?

Yes, it is. It's as appropriate as using it for motherboard + hard drives which iirc was the first case when the technological terms master and slave have been used. The best effort is responding to a problem **if there actually is one** and if someone whose problem it is actually calls out for a change and considering the significance of the impact to that target group instead of trying to do stuff for others even though no one cared for 10 years and the term **did no harm** because it's not meant as a shaming word, but as a word describing a historically valid behavior that most likely everyone understands better than using an alternative word that describes the behavior less effectively.

> A slave can still have agency, ability to act at their own will, although they certainly would face retribution, does that fit our model? Not at all ...

Yes, it fits perfectly as you said because - widget behaviors, input management, providers inserted automatically, basically everything not under the `KIVY_SLAVE_MODE` flag is considered to have its own will, its own implementation and the only thing the slave mode does is taking some parts of the environment out - e.g. a slave without boots if you want to go there... It's still controlled by the master, still the actions are owned by the master and the master is responsible for them i.e. master == developer/application wrapping Kivy with slave mode. If I run Kivy in the slave mode, obviously the "retribution" part is that the SDL2 window won't be created and everything using OpenGL will bork. Same with the event loop. On the other hand, Clock will work at least partially, EventDispatcher might too, Widget behaviors and some other parts of the ecosystem will function normally too same as an imaginary slave being able to live, think, etc and to exit a building (s)he's kept in from a non-ground floor.

> By managed i mean that you need to manage the framework ...

Not exactly. You can poke the framework by calling the functions that would be called anyway, but you need to care for the providers. That is not something Kivy can't live without, however it significantly cripples the framework to the state it's not usable in its primary way unless you know when, where and what to call (e.g. create a Window). You can still use the rest of the packages and modules.

> they have to manage kivy, completely, not just giving it orders, but caring about its internal state and being in charge of all the necessary actions, which is the opposite of delegating them to a slave.

Quite the opposite, because Kivy isn't a wrapper for the providers. Kivy is much more, just using the providers themselves. In fact, Kivy itself as a framework can be perfectly considered to be a master and its providers slaves. Also, [not every slave had a bad life](https://www.quora.com/Did-slaves-live-a-good-life-after-freedom). This doesn't mean I'm pro-slavery, only responding to the "but caring about its internal state" part.

> ... KIVY_BASE_MODE, this allows to run kivy in the mode this PR describe, or any other alternative mode, should we discover a need of them ... "normal" , ... could accept the "slave" ... and is just a suggestion to give more flexibility in case other potential "modes" are discovered later.

Yeah, I was thinking about using one flag instead of multiple `KIVY_WHATEVER_MODE` and use just a `slave` value for that, however if you consider that we might have multiple modes and at multiple places, especially if you consider some Cython parts, well, we can't go and parse the value on each time and it's cheaper to just check `'KIVY_SLAVE_MODE' in os.environ` instead of `if os.environ.get('KIVY_BASE_MODE', '') == 'slave'` and implement other logic. That's why we have [`KIVY_NO_*`](https://kivy.org/doc/stable/guide/environment.html#configuration) flags, because in the end we don't care about the value, we care about the flag's presence unlike with `KIVY_WINDOW` or `KIVY_GL_BACKEND` or metrics.

tshirtman:
> You can't solve problems for someone else especially if that problem is not related to you. It's similar to children - if you move the obstacles, it won't serve the purpose and the child won't learn from it.

Please think about the implication of this sentence in this context, seriously, this is messed up, i don't want this to be a fight, but it really thinks like you want it to be.

> This is not an error this is a technological term.

These two things are not mutually exclusive, the naming of things in science and technology evolves for a lot of reasons, taxonomy for example, changes a *lot* over time, but it's far from the only example.

> Neither history, nor proper description is an error.

I'm arguing that it's not the most proper description. That we can do better.

> It's not a mistake it's the first commit in this repository and has been present in PyMT for 10 years.

Again, these two things are not mutually exclusive, if we went by this criteria, there would be literally no bug in Kivy, everything would be a feature, sometime doing the opposite of what documentation states or what common sense would assume, but still a feature.

> It's not a bad word, it's a word with a specific meaning in this context and with other specific meanings in different contexts. Here it's a technological term. Do you want to go word after a word in the Oxford dictionary because someone might not like it? Would you like to fly in a plane with safety instructions written in pretty words or in strict and direct ones everyone can understand in a split of a second?

Again, i'm merely arguing that it might not be the best word to convey the concept we want to, not only because it's likely to cause distractions, but also because it's not a very clear concept, as you see in you argue yourself, it covers multiple different realities.

> The best effort is responding to a problem if there actually is one and if someone whose problem it is actually calls out for a change and considering the significance of the impact to that target group instead of trying to do stuff for others even though no one cared for 10 years and the term did no harm because it's not meant as a shaming word

Well, *i* think it's a problem, you can't make that go away by saying it's not my problem, i care what signal we send by carelessly using that word these days where it's pretty much common knowledge that it's making some people feel less welcome, the mere fact that we don't care for it should make anybody uneasy joining the project, because there is a clear lack of sensitivity in doing so. I'm saddened that you don't agree on that. 

> That particular sentence is just a piece of crap without any numbers, sources, articles, etc

[CNN 2016 global slavery estimates](https://edition.cnn.com/2017/09/19/world/global-slavery-estimates-ilo/index.html)
[global slavery index](https://www.globalslaveryindex.org/2018/findings/highlights/)

Your reasoning didn't take into account the fact that world population exploded in the XXth century, and is still growing amazingly fast, the world was a smaller place before. Also, i didn't mean the total sum over history, but the amount at any point in history, still i might be off the mark, i didn't find a graph of total amount over time, this is certainly a complicated number, but for example [wikipedia](https://en.wikipedia.org/wiki/Slavery#Americas_2)

(and these numbers don't count amazon warehouse workers, or foxxcon's china's worker, as you feared)

> Brazil was the last country in the Western world to abolish slavery. By the time it was abolished, in 1888, an estimated four million slaves had been imported from Africa to Brazil, 40% of the total number of slaves brought to the Americas. For reference, the United States received 10 percent. 

it means the americas imported around 10millions people as slaves, total, sure, some of these had children, and after a few decades, the amount could have been sensibly bigger, but it's not certain that would have hit the 40Millions mark.

> Also, not every slave had a bad life. This doesn't mean I'm pro-slavery, only responding to the "but caring about its internal state" part.

Seriously, i didn't even imagine you would go there, this not only doesn't answer the question (and also, what you link to is a different — so naive it hurts — question). I'm glad you are not criminally insane (by being pro-slavery), but it's not really a high bar to pass, demeaning the evil of slavery because, yeah, some had it better than others, and certainly even some had it better than a random free peasant, is just… bad, that's not making the whole concept any better, so don't argue in that direction, and double check what you write, to make sure it's not what you are inadvertently doing.

Now, we have that out of the way, i hope. 

> Yes, it fits perfectly as you said because - widget behaviors, input management, providers inserted automatically, basically everything not under the KIVY_SLAVE_MODE flag is considered to have its own will, its own implementation

well, good thing there is still a point in importing kivy in that mode, there wouldn't be much point of having it if it made kivy completly empty. There is basically no difference between "managing" a process and "giving it orders", one word just has less harmful cultural baggage than the other.

> and the only thing the slave mode does is taking some parts of the environment out - e.g. a slave without boots if you want to go there... It's still controlled by the master, still the actions are owned by the master and the master is responsible for them i.e. master == developer/application wrapping Kivy with slave mode. If I run Kivy in the slave mode, obviously the "retribution" part is that the SDL2 window won't be created and everything using OpenGL will bork. Same with the event loop. On the other hand, Clock will work at least partially, EventDispatcher might too, Widget behaviors and some other parts of the ecosystem will function normally too same as an imaginary slave being able to live, think, etc and to exit a building (s)he's kept in from a non-ground floor.

That's not retribution, there is no point in punishing a process because you misused it. This mode just asks for a more involved usage of kivy if you want to use it this way, it gives you more control and responsibilities. That's what a manager is, and in a sense, a slave owner is just a manager with life or death power over its subject, you could argue that it's true there, but that's not a meaningful distinction, because you always had that, as you could always exit the program and kivy altogether.

If you say that the framework is crippled to the point it's a lot of work to make it behave properly, i feel like another term is mandated, i'm not sure which yet, but `barebone`, `no_side_effect`, maybe, `marionette` or `puppet` could be ok too, nobody has use for a crippled slave you have to catter the needs of, such people were typically disposed of.

> In fact, Kivy itself as a framework can be perfectly considered to be a master and its providers slaves

yes, you can apply this metaphor to about everything in computer science, or even engineering, because everything here serves a purpose to something else, that's why it's a lazy and empty word to use, on top of being a sensitive term, it's just not a very useful one. It's logical to want some way to denominate which part control which in all these relationships, but there is almost always a more fit term that master/slave, it just take some work to find it, because you have to think about semantics. 

> Yeah, I was thinking about using one flag instead of multiple KIVY_WHATEVER_MODE and use just a slave value for that, however if you consider that we might have multiple modes and at multiple places, especially if you consider some Cython parts, well, we can't go and parse the value on each time and it's cheaper to just check 'KIVY_SLAVE_MODE' in os.environ instead of if os.environ.get('KIVY_BASE_MODE', '') == 'slave' and implement other logic. That's why we have KIVY_NO_* flags, because in the end we don't care about the value, we care about the flag's presence unlike with KIVY_WINDOW or KIVY_GL_BACKEND or metrics.c 

that's an implementation problem, there are ways to work around that, that shouldn't determine the user interface. You'll want to save that state as soon as the kivy module is imported, and always check the value from there instead of in the environment.

KeyWeeUsr:
Ah, there is no point for me to even lead the discussion further. Slave is a term you don't like whilst you aren't a slave nor related to them and you can't disassociate from that nor from the different associations of the term in the technology and science.

Re the indexes, go read the metrics being used, apply them, retrospectively, to the past civilisations and go figure the result. Get percentages, etc. Even building such a metric is inaccurate because in the end you get to the point when you don't have data since the certain period, so you don't include them and the whole result is distorted. Debt, forced work, etc isn't something new. While slavery has always been an issue and is even in the modern world, just by providing incorrect results by excluding periods you don't have data for means building on a mistake that slavery didn't exist in those periods and *that* is a disrespect and incorrect, not naming someone or something according to some historically valid behavior.

Anyway, I stated my point to using the term.

> Please think about the implication of this sentence in this context, seriously, this is messed up, i don't want this to be a fight, but it really thinks like you want it to be.

No one is beating you or threaten you or anything. This is not a fight, mistreatement, or anything like that. This is a discussion, so use the correct term. Anyway, I don't have time to write novels just to defend usage of a technological term and the time I spent on this PR already I could spend more efficiently. You are actively against it even though you appeared like you might consider using the word. There is a term for that too.

pythonic64:
So we are losing this feature over a word?

tshirtman:
Seems so, because apparently it's this word or nothing, because no other word could possibly describe this feature.

PR #6179, opened 2019-03-05
window: fix multiple resize sent, and always sent the GL size, never …

### LINKED ISSUE #6082: Window on_resize event fires twice with different widths
### Versions

* Python: 3.7
* OS: Mac OS Mojave
* Kivy: 1.10.1

### Description

I'm subscribed to the window.on_resize event. I expect to get 1 callback per resize. Instead I get 2, one for width, one for width * 2.

width 1680
width 3360
width 785
width 1570
width 366
width 732

### Code and Logs

```python
class Example(Widget):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        Window.maximize()
        Window.bind(on_resize=self.on_window_resize)

    def on_window_resize(self, window, width, height):
        print("width", width)

class MainApp(App):
    def build(self):
        return Example()

if __name__ == '__main__':
    MainApp().run()
```

tito:
…the system size, otherwise you'll end up with the size sent twice.

Now, because we send the correct size, there is still a code path that make the resize event sent twice. I introduced a pre_resize internal event to debounce it.

Fixes #6082

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -399,7 +399,7 @@ class WindowBase(EventDispatcher):
             else:
                 self._size = size[1], size[0]
 
-            self.dispatch('on_resize', *size)
+            self.dispatch('on_pre_resize', *size)
             return True
         else:
             return False
@@ -531,7 +531,7 @@ class WindowBase(EventDispatcher):
         self._rotation = x
         if not self.initialized:
             return
-        self.dispatch('on_resize', *self.size)
+        self.dispatch('on_pre_resize', *self.size)
         self.dispatch('on_rotate', x)
 
     rotation = AliasProperty(_get_rotation, _set_rotation,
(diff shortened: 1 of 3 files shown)


tito:
How would you rephrase it? The issue explained everything with a runnable example,and I described the way I handled it. I don't think that the message was missing anything, except GH that splitted the title but that's it?

AndreMiras:
Yes I'm talking about the title split, it's actually a feature that comes fro that 50/72 rule. So I just suggested to be more concise in the title and then put the details you put but in the body (using two new lines).
So basically the commit message could be something like:
```
window: fix multiple resize sent, closes #6082

fix multiple resize sent and always sent the GL size,
never the system size, otherwise you'll end up with the size sent twice
Now, because we send the correct size, there is still a code path that
make the resize event sent twice. I introduced a pre_resize internal
event to debounce it.
````
That's the idea, note the empty line after the title.
The PR is approved anyway, it was just some side feedback

tito:
So it's just because the first line of the commit is too long right?
I'll keep that in mind next time.

AndreMiras:
Yes it's exactly about that too long first line. I should have stated that more clearly :sweat_smile:  https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting

PR #6186, opened 2019-03-07
Supports live resizing (sdl2 only)

### LINKED ISSUE #2844: Live window resizing
Currently, a Kivy window supports resizing by allowing the user to click and drag the window border, similar to most other applications.
However, the user can not see how the new size affects the application until they release the mouse button, so Kivy applications do not feel to be as responsive or fluid as other applications are.
I would like to have the application resize event to occur while the user is dragging the mouse border, rather than only when the user releases the mouse pointer.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7667215-live-window-reszing?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F42681&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tito:
Today, when the window is resized content is black or just partially drawed. This is because poll() is blocked until we release the mouse.
But we can catch the resize event in the event filter. So the idea is to catch and force the window resize, and force drawing. Works like a charm.
Tested on OSX 10.14 with SDL 2.0.5 and retina screen.

Credits goes to @arafangion for the initial code.

Fixes #2844

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -47,7 +47,12 @@ cdef class _WindowSDL2Storage:
         cdef str name = None
         if not self.event_filter:
             return 1
-        if event.type == SDL_APP_TERMINATING:
+        if event.type == SDL_WINDOWEVENT:
+            if event.window.event == SDL_WINDOWEVENT_RESIZED:
+                action = ('windowresized',
+                          event.window.data1, event.window.data2)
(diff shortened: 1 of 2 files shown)


akshayaurora:
Works like a charm :+1:

PR #6192, opened 2019-03-11
Config: map KCFG_ environ variables to Config

### LINKED ISSUE #6190: Simply loading kivy has side-effects, writes logs to console and filesystem.
### Versions

* Python: any
* OS: any
* Kivy: any

### Description

Simply importing `kivy` and nothing else emits 250 bytes of logging to `stdout` and creates a dozen files or directories in the file system.  

This breaks my system for some invocations that emit JSON, and is undesirable in other cases.

I can suppress all logging with the [`KIVY_NO_CONSOLELOG`](https://kivy.org/doc/stable/guide/environment.html#configuration) environment variable, but that's worse, because all ERRORs and WARNINGs will be eliminated.

Writing entries in the users `~/.kivy/config.ini` seems like a really bad idea, because "if everyone else did it..." - even though I'm pretty sure that most of our users have no other `kivy` applications, yet.

My current plan is this.  Before `kivy` is loaded anywhere in my code, I change `os.environ['KIVY_HOME']` to refer to my application's folder, and then write my own `config.ini` to turn up the log level.

This is fragile as the correct functioning of my code will depend on the order that modules happen to be loaded, and I have multiple main programs.

----

Just _loading_ a library should not have any side effects like disk or console output.  

By comparison, the other 20+ Python libraries I use don't log at all or create any files on the file system just by being loaded - here's a subset of those libraries which covers all sorts of areas in Python and many of which you likely already have on your system.

    import (bibliopixel, cv2, flake8, flask, jedi, mido, numpy, pytest,
            requests, six, tornado, websocket, zmq)

### Code and Logs

```python

import kivy
```

tito:
This allow the user to preconfigure kivy before the import, and/or let the final user to configure kivy itself.

Fixes #6190

### DIFF
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -49,6 +49,31 @@ For information on configuring your :class:`~kivy.app.App`, please see the
     converted from ascii to unicode only when needed. The method get() returns
     utf-8 strings.
 
+Changing configuration with environment variables
+-------------------------------------------------
+
+Since 1.11.0, it is now possible to change the configuration using
+environment variables. They take precedence on the loaded config.ini.
+The format is::
+
+    KCFG_<section>_<key> = <value>
+
+For example:
+
+    KCFG_GRAPHICS_FULLSCREEN=auto ...
+    KCFG_KIVY_LOG_LEVEL=warning ...
+
+Or in your file before any kivy import:
+
+    import os
+    os.environ["KCFG_KIVY_LOG_LEVEL"] = "warning"
+
+If you don't want to map any environment variables, you can disable
+the behavior::
+
+    os.environ["KIVY_NO_ENV_CONFIG"] = "1"
+
+
 .. _configuration-tokens:
 
 Available configuration tokens
@@ -874,3 +899,16 @@ if not environ.get('KIVY_DOC_INCLUDE'):
             Config.write()
         except Exception as e:
             Logger.exception('Core: Error while saving default config file')
+
+    # Load configuration from env
+    if 'KIVY_NO_ENV_CONFIG' not in environ:
+        for key, value in environ.items():
+            if not key.startswith("KCFG_"):
+                continue
+            try:
+                _, section, name = key.split("_", 2)
+            except ValueError:
+                raise
+            section = section.lower()
+            name = name.lower()
+            Config.set(section, name, value)
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -27,6 +27,8 @@ Alternatively, you can save these settings permanently using
 restart the app for the changes to take effect. Note that this approach will
 effect all Kivy apps system wide.
 
+Please note that no underscore (`_`) is allowed in the section name.
+
 Usage of the Config object
 --------------------------
 
@@ -539,6 +541,7 @@ class ConfigParser(PythonConfigParser, object):
     def adddefaultsection(self, section):
         '''Add a section if the section is missing.
         '''
+        assert("_" not in section)
         if self.has_section(section):
             return
         self.add_section(section)
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -926,8 +926,8 @@ if not environ.get('KIVY_DOC_INCLUDE'):
 
             # extract and check the option name
             name = name.lower()
-            sections_to_check = (
-                "kivy", "graphics", "widgets", "postproc", "network")
+            sections_to_check = {
+                "kivy", "graphics", "widgets", "postproc", "network"}
             if (section in sections_to_check and
                     not Config.has_option(section, name)):
                 Logger.warning((
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -911,7 +911,30 @@ if not environ.get('KIVY_DOC_INCLUDE'):
             try:
                 _, section, name = key.split("_", 2)
             except ValueError:
-                raise
+                Logger.warning((
+                    "Environ `{}` invalid format, "
+                    "must be KCFG_section_name").format(key))
+                continue
+
+            # extract and check section
             section = section.lower()
+            if not Config.has_section(section):
+                Logger.warning(
+                    "Environ `{}`: unknown section `{}`".format(
+                        key, section))
+                continue
+
+            # extract and check the option name
             name = name.lower()
+            sections_to_check = (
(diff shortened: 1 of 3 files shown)


tito:
@matham Documentation added, and assert about the underscore in the `Config.adddefaultsection`.
It doesn't cover the case of having a manual section with underscore, created by the user. Because i would like to avoid additionnal processing at every import just to check a minor validity.

kivy/config.py
```diff
@@ -539,6 +541,7 @@ def getdefaultint(self, section, option, defaultvalue):
     def adddefaultsection(self, section):
         '''Add a section if the section is missing.
         '''
+        assert("_" not in section)
```
rec: You don't need parens around an `assert` statement!

Overall, I don't think this should be an `assert` statement at all, because you shouldn't be using `assert`s to validate user input data.

The reason is simple - asserts don't go off if you pass the `-O` flag to Python, and in this case, it means you won't check this condition at all.

So if you were going to keep this test, perhaps consider making it throw a `ValueError`.

---

However, can't you do better than that even?  Adding another arbitrary condition that someone needs to remember is not just inelegant, it adds a little extra overhead for each developer.  Yes, it's tiny here, but these things mount up.

Reading this fragment of code makes it seem as if you can add arbitrary sections just by naming them, and you get no warning at all.

This means I can make a mistake like mistyping `KCFG_GRAPHICS_FULLSCREEN` and get no visible error:  for example

    KCFG_GRAPHIC_FULLSCREEN=1 my_program

Isn't there some way to detect the list of valid sections, and conceivably even valid names within that section, perhaps in kivy/config.py?  

Then you could peel off the section name whether or not it had an `_` in it, and as a bonus could also catch all sorts of niggling errors.

So if you did that, you could get rid of the `assert` entirely, _and_ catch more errors, _and_ present a clearer error message when you did.

Just a thought.  I realize I'm proposing work for someone else. :-)  

Although, I have some spare time this week if you needed it knocked out...

Thanks again for doing this!

kivy/config.py
```diff
@@ -874,3 +899,16 @@ def name(self, value):
...
+        for key, value in environ.items():
+            if not key.startswith("KCFG_"):
+                continue
+            try:
+                _, section, name = key.split("_", 2)
+            except ValueError:
+                raise
+            section = section.lower()
+            name = name.lower()
+            Config.set(section, name, value)
```
rec: Ah, sorry, one more quibble - the lack of a newline here is causing Travis to fail!

(I agree with Travis here.  Every line should end in a carriage return, because then all lines are uniform and your code is easier.)

Thanks again!

kivy/config.py
tito: In my mind, assert _ was just here because it's mostly a reminder for us (dev) not for the user. The user don't have to touch the kivy configuration file, as it has its own application configuration. This is why i did the assert compared to raising an exception.

I could check if the section is valid, but i can't check if the name is valid. Because there is some sections like "input" where the key is up to the user. So there is no validation for that. I'll update the implementation.

tito:
@rec It should be better now.
A warning is displayed for invalid section, and nothing is set.
A warning is displayed for unknown key, but only for certain sections, and it set the key anyway (think about some garden widgets or external widget that could save their own configuration inside the kivy configuration).
Nothing is done for modules and input section.

tito:
And the assert() is here just for us when coding as a reminder, and it has no impact for deployement.

rec:
Wow, that was fast, I have friction burns on my inbox!  :-D

Right, the `assert` is fantastic now, because it's just a backstop to make sure your logic is fine, and "should never fire".

Looks Good To Me!

kivy/config.py
```diff
@@ -911,7 +911,30 @@ def name(self, value):
...
             section = section.lower()
+            if not Config.has_section(section):
+                Logger.warning(
+                    "Environ `{}`: unknown section `{}`".format(
+                        key, section))
+                continue
+
+            # extract and check the option name
             name = name.lower()
+            sections_to_check = (
```
rec: OK, two  more comments since I looked this over again.  :-D

Consider bringing that list of sections to the top of the file as a named constant!  Or, not.

kivy/config.py
```diff
@@ -911,7 +911,30 @@ def name(self, value):
...
-                raise
+                Logger.warning((
+                    "Environ `{}` invalid format, "
+                    "must be KCFG_section_name").format(key))
+                continue
+
+            # extract and check section
             section = section.lower()
+            if not Config.has_section(section):
+                Logger.warning(
```
rec: Also, since you're using Python logging under the hood, this is slightly an antipattern, as it evaluates the format even if warnings are suppressed.

Traditionally, logging uses :-/ old style % formatting, so what one is supposed to do is this:

    Logger.warning("Environ `%s`: unknown section `%s`", key, section)

If you hated % formatting enough not to want to do this, I'd totally understand.  :-D

tito: I understand (i didn't knew about it), but this is how we do logging in the whole Kivy library. We could however (not in this PR) make a pass to use the old style, dunno about it.

kivy/config.py
tito: I don't think so, that's not something i wanted to export/expose, it's just a matter of style to prevent having a big "if" but that's it

kivy/config.py
AndreMiras: I second some of the concerns of @rec and would emphasis on not using parenthesis on `assert` because it's a keyword rather than a function. Using parenthesis may lead to unexpected behavior in certain conditions

kivy/config.py
AndreMiras: Also have a super minor remark, this `sections_to_check` should probably be a set rather than a list

rec: I know this is in, but can't resist.  :-D 

It's a tuple actually, and if you replaced it (given you had infinite free time :-D), you might want to replace it with a `frozenset` as it's immutable, _but_ for small collections, `set`/`frozenset` is a little less efficient than a `tuple` and somewhat larger.

rec:
Thanks for doing this - you people rock! 

Let me know if you need any help for maintenance changes in the future, just to keep my hand in.

PR #6199, opened 2019-03-14
Replace "it's" with "its" in several places

rec:
Sorry for these low priority pull requests, I'm reading through your stuff and these jump out at me!  :-)  (Also, I can do this pretty fast and it gives me a good feeling for the code.)

"It's" appears 181 times in the code, and I changed 19 of them to "its" - the possessive.  I've seen worse.  :-D

Rule: if you can't replace "it's" with "it is" or "it has" then it should be "its".  Example sentence: "It's a book with its cover removed."

Again, sorry for the noise here and thanks for this excellent package that I'm making very good use of!


akshayaurora:
@rec thanks a ton for the fixes!

rec:
The pleasure is all mine!  :-)

PR #6204, opened 2019-03-15
Update OSX Install instructions for MakeSymlinks

### LINKED ISSUE #6202: Kivy OSX install instructions fail at first step. DOC issue + install code issue
From https://kivy.org/doc/stable/installation/installation-osx.html:
```
2. Open the dmg
3. Copy the Kivy.app to /Applications.
4. Create a symlink by running the makesymlinks in the window that opens when you open the dmg
```
Point 4) fails on OSX 10.11.6. It fails w/ a console error message in 1.9.1 (access denied), silently in 1.10.x.

So this doesn't work unless your script has no dependencies, which is the **succeeding** topic:
```
You can just drag and drop your main.py to run your app too.
Installing modules
```
But this will fail (because makesymlinks failed):
```
$ kivy -m pip install <modulename>
```
Which means that "drag and drop your main.py to run your app" will fail. Catch-22.

Complete frustration before having made it to the end of the installation page.

The OSX doc issue is really two issues. 1) Makesymlinks is foobar. 2) Onward instructions are therefore useless. Fix makesymlink and it's possible the rest will be right. 

Oddly makesymlinks is sytlistically a lot closer to bash than to Python and that's a pity.

akshayaurora:
fixes #6202

### DIFF
--- a/doc/sources/installation/installation-osx.rst
+++ b/doc/sources/installation/installation-osx.rst
@@ -54,7 +54,13 @@ To install Kivy, you must:
        https://kivy.org/downloads/ and download `Kivy-*-osx-python*.dmg`.
     2. Open the dmg
     3. Copy the Kivy.app to /Applications.
-    4. Create a symlink by running the `makesymlinks` in the window that opens when you open the dmg
+    4. Create a symlink by running the `makesymlinks` in the window that opens when you open the dmg.
+    
+    If you have trouble running this script, you can try right-click->Open or just add the link manually
+    by running the following command::
+        `ln -s /Applications/Kivy<version>/Contents/Resources/script /usr/local/bin/kivy<version>`
+    version is either 2/3 based on which version of the app did you download.
+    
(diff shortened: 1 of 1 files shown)


doc/sources/installation/installation-osx.rst
```diff
@@ -54,7 +54,13 @@ To install Kivy, you must:
        https://kivy.org/downloads/ and download `Kivy-*-osx-python*.dmg`.
     2. Open the dmg
     3. Copy the Kivy.app to /Applications.
-    4. Create a symlink by running the `makesymlinks` in the window that opens when you open the dmg
+    4. Create a symlink by running the `makesymlinks` in the window that opens when you open the dmg.
+    
+    If you have trouble running this script, you can try right-click->Open or just add the link manually
+    by running the following command::
```
tshirtman: Doesn't that require sudo?

tito:
Unfortunately no!
If your /usr/local/bin is managed by brew, you have user access to it, there is no need to use sudo!
If you use it, then just run brew doctor to fix the permissions issue but indicating here in the documentation that sudo is required is not good imo

PR #6217, opened 2019-03-22
add a few kv examples to widget docs

tshirtman:


tito:
Careful about Kv KV, and careful about :: required for sphinx for block code

kivy/uix/scrollview.py
```diff
@@ -86,6 +86,21 @@
...
+        do_scroll_x: False
+        do_scroll_y: True
+
+        Label:
+            size_hint_y: None
+            height: self.texture_size[1]
+            text_size: self.width
+            padding: 10, 10
+            text:
+                'really some amazing text\n' * 100
```
akshayaurora: \\n

Issue #2479, opened 2014-09-19
embedded custom classes don't have ids


cltrudeau:
If you define your own class extending from a widget and use it in the kv-lang inside another class it does not get an id.  There isn't anything in the documentation that says you can do this, but it seems like it fits with how the kv-lang is constructed.

```
<InsideBox>:
    size_hint: (1, 1)
    Label:                    
        id: in_left            
        text: "Inside Left"              
    Label:                                    
        id: in_right              
        text: "Inside Right"              

<OutsideBox>:                              
    size_hint: (1, 1)                    
    orientation: 'vertical'              
    Label:                                
        id: out_over            
        text: "Over"                      
    InsideBox:                    
        id: out_insidebox  
    Label:                                                                      
        id: out_under
        text: "Under"
```

```
class InsideBox(BoxLayout):
    def __init__(self, *args, **kwargs):
        super(InsideBox, self).__init__(*args, **kwargs)                        
        print 'inside ids:',  self.ids                                          

class OutsideBox(BoxLayout):                                                    
    def __init__(self, *args, **kwargs):                                        
        super(OutsideBox, self).__init__(*args, **kwargs)                      
        print 'outside ids:',  self.ids                                        

class MyApp(App):                                                              
    def build(self):                                                            
        #return InsideBox()                                                    
        return OutsideBox()
```

In the code above, "inside ids" will be empty even though it should have two Labels.  The "outside ids" shows correctly.  If you change the comment in MyApp and just have the InsideBox returned, it works as expected.

PR #6221, opened 2019-03-24
[WIP] give the user the opportunity to initialize widget before/after applying kv rules

gottadiveintopython:
(EDIT)

# Problem

If you want to do some stuff **after** kv rules are applied, you **can't** write like this:

```python3
Builder.load_string('''
<MyWidget>:
    Label:
        id: label
''')

class MyWidget(Widget):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        label = self.ids.label  # A
        # do some stuff for label
```

because kv rules ain't always applied inside `__init__()`.

```python3
# This code causes "KeyError: 'label'" at line A
widget = Builder.load_string('''
Widget:
    MyWidget:
''')
```

# Workaround?

One workaround for the problem might be this:

```python3
class MyWidget(Widget):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        def after_rules_are_applied(__):
            label = self.ids.label
            # do some stuff for label
        Clock.schedule_once(after_rules_are_applied)
```

but in this case, you have another problem. You can't use `MyWidget` right after the instantiation because the instance is not fully initialized until the `Clock` ticks.

```python3
widget = MyWidget()  # the instance is not fully initialized!!
```

# so this PR

This PR gives the user a guaranteed opportunity to initialize widget after applying kv rules by implementing special method `__post_kv__()`.

```python3
    def __post_kv__(self):
        label = self.ids.label
        # do some stuff for label
```

Do you think this kind of feature is necessary?

### DIFF
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -397,9 +397,15 @@ class BuilderBase(object):
 
             if parser.root:
                 widget = Factory.get(parser.root.name)(__no_builder=True)
-                self.apply(widget)
-                self._apply_rule(widget, parser.root, parser.root)
-                widget._dispatch_on_kv_post_recursively()
+                rule_children = []
+                self.apply(widget, rule_children=rule_children)
+                self._apply_rule(
+                    widget, parser.root, parser.root,
+                    rule_children=rule_children)
+
+                for child in rule_children:
+                    child.dispatch('on_kv_post', widget)
+                widget.dispatch('on_kv_post', widget)
                 return widget
         finally:
             self._current_filename = None
@@ -435,7 +441,8 @@ class BuilderBase(object):
         self._apply_rule(widget, rule, rule, template_ctx=proxy_ctx)
         return widget
 
-    def apply_rules(self, widget, rule_name, ignored_consts=set()):
+    def apply_rules(
+            self, widget, rule_name, ignored_consts=set(), rule_children=None):
         '''Search all the rules that match `rule_name` widget
         and apply them to `widget`.
 
@@ -452,9 +459,11 @@ class BuilderBase(object):
         if not rules:
             return
         for rule in rules:
-            self._apply_rule(widget, rule, rule, ignored_consts=ignored_consts)
+            self._apply_rule(
+                widget, rule, rule, ignored_consts=ignored_consts,
+                rule_children=rule_children)
 
-    def apply(self, widget, ignored_consts=set()):
+    def apply(self, widget, ignored_consts=set(), rule_children=None):
         '''Search all the rules that match the widget and apply them.
 
         `ignored_consts` is a set or list type whose elements are property
@@ -468,14 +477,16 @@ class BuilderBase(object):
         if not rules:
             return
         for rule in rules:
-            self._apply_rule(widget, rule, rule, ignored_consts=ignored_consts)
+            self._apply_rule(
+                widget, rule, rule, ignored_consts=ignored_consts,
+                rule_children=rule_children)
 
     def _clear_matchcache(self):
         BuilderBase._match_cache = {}
         BuilderBase._match_name_cache = {}
 
     def _apply_rule(self, widget, rule, rootrule, template_ctx=None,
-                    ignored_consts=set()):
+                    ignored_consts=set(), rule_children=None):
         # widget: the current instantiated widget
         # rule: the current rule
         # rootrule: the current root rule (for children of a rule)
@@ -583,8 +594,12 @@ class BuilderBase(object):
                 # apply(), and so, we could use "self.parent".
                 child = cls(__no_builder=True)
                 widget.add_widget(child)
-                self.apply(child)
-                self._apply_rule(child, crule, rootrule)
+                self.apply(child, rule_children=rule_children)
+                self._apply_rule(
+                    child, crule, rootrule, rule_children=rule_children)
+
+                if rule_children is not None:
+                    rule_children.append(child)
 
         # append the properties and handlers to our final resolution task
         if rule.properties:
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -402,7 +402,7 @@ class BuilderBase(object):
                 self._apply_rule(
                     widget, parser.root, parser.root,
                     rule_children=rule_children)
-
+                widget.dispatch('on_kv_applied')
                 for child in rule_children:
                     child.dispatch('on_kv_post', widget)
                 widget.dispatch('on_kv_post', widget)
@@ -595,6 +595,7 @@ class BuilderBase(object):
                 child = cls(__no_builder=True)
                 widget.add_widget(child)
                 self.apply(child, rule_children=rule_children)
+                child.dispatch('on_kv_applied')
                 self._apply_rule(
                     child, crule, rootrule, rule_children=rule_children)
 
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -396,8 +396,9 @@ class BuilderBase(object):
                 self.files.append(fn)
 
             if parser.root:
-                widget = Factory.get(parser.root.name)()
+                widget = Factory.get(parser.root.name)(__rootrule=True)
                 self._apply_rule(widget, parser.root, parser.root)
+                widget._dispatch_on_kv_post_recursively()
                 return widget
         finally:
             self._current_filename = None
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -396,7 +396,8 @@ class BuilderBase(object):
                 self.files.append(fn)
 
             if parser.root:
-                widget = Factory.get(parser.root.name)(__rootrule=True)
+                widget = Factory.get(parser.root.name)(__no_builder=True)
+                self.apply(widget)
                 self._apply_rule(widget, parser.root, parser.root)
                 widget._dispatch_on_kv_post_recursively()
                 return widget
--- a/kivy/tests/test_lang_pre_process_and_post_process.py
+++ b/kivy/tests/test_lang_pre_process_and_post_process.py
@@ -50,7 +50,7 @@ class LangTestCase(unittest.TestCase):
                 if not self.there_is_a_rootrule:
                     testcase.fail(
                         "Strange. The handler was called even though "
-                        "the widget doesn't have root rule.")
+                        "there is no root rule.")
                 self._n_post_from_r += 1
                 ae(self._n_pre_from_d, 1)
                 ae(self._n_post_from_r, 1)
--- a/kivy/tests/test_lang_pre_process_and_post_process.py
+++ b/kivy/tests/test_lang_pre_process_and_post_process.py
@@ -5,49 +5,6 @@ from kivy.factory import Factory
 from kivy.properties import ObjectProperty, NumericProperty, BooleanProperty
 
 
-KV_CODE = '''
-<EventCounter>:
-    on_kv_pre: self.on_kv_pre_from_c()  # This line won't be excuted
-    on_kv_post: self.on_kv_post_from_c()
-
-<TestBoxLayout>:
-    Label:
-        id: label
-        text: textinput.text
-    TextInput:
-        id: textinput
-    Button:
-        id: button
-        on_press: self.text = 'pressed'
-
-<TestLabel>:
-    text: 'A'
-    on_kv_post: self.text += 'B'
-    Label:
-        text: root.assert_the_text_hasnt_changed_yet() or 'hello'
-<OtherTestLabel@TestLabel>:
-    on_kv_post: self.text += 'C'
-    height: self.assert_the_text_hasnt_changed_yet() or 200
-    Label:
-        text: root.assert_the_text_hasnt_changed_yet() or 'hello2'
-'''
-KV_FILENAME = 'test_kv_filename'
-
-
-def setUpModule():
-    Builder.load_string(KV_CODE, filename=KV_FILENAME)
-
-
-def tearDownModule():
-    Builder.unload_file(KV_FILENAME)
-    u = Factory.unregister
-    u('EventCounter')
-    u('TestWidget')
-    u('TestBoxLayout')
-    u('TestLabel')
-    u('OtherTestLabel')
-
-
 class LangTestCase(unittest.TestCase):
 
     def test_how_may_times_handlers_are_called(self):
@@ -105,6 +62,21 @@ class LangTestCase(unittest.TestCase):
                 ae(self._n_post_from_r, 1 if self.does_have_root_rule else 0)
                 ae(self._n_post_from_c, 1)
                 ae(self._n_post_from_d, 1)
+
+        Builder.load_string(textwrap.dedent('''
+        <EventCounter>:
+            on_kv_pre: self.on_kv_pre_from_c()  # This line won't be excuted
+            on_kv_post: self.on_kv_post_from_c()
+        <TestBoxLayout>:
+            Label:
+                id: label
+                text: textinput.text
+            TextInput:
+                id: textinput
+            Button:
+                id: button
+                on_press: self.text = 'pressed'
+        '''))
 
         # case #1: Without root rule
         root = EventCounter(does_have_root_rule=False)
@@ -193,6 +165,19 @@ class LangTestCase(unittest.TestCase):
 
     def test_property_is_evaluated_before_on_kv_post_is_fired(self):
         ae = self.assertEqual
+
+        Builder.load_string(textwrap.dedent('''
+        <TestLabel>:
+            text: 'A'
+            on_kv_post: self.text += 'B'
+            Label:
+                text: root.assert_the_text_hasnt_changed_yet() or 'hello'
+        <OtherTestLabel@TestLabel>:
+            on_kv_post: self.text += 'C'
+            height: self.assert_the_text_hasnt_changed_yet() or 200
+            Label:
+                text: root.assert_the_text_hasnt_changed_yet() or 'hello2'
+        '''))
 
         class TestLabel(Factory.Label):
             def assert_the_text_hasnt_changed_yet(self):
--- a/kivy/tests/test_lang_pre_process_and_post_process.py
+++ b/kivy/tests/test_lang_pre_process_and_post_process.py
@@ -166,31 +166,3 @@ class LangTestCase(unittest.TestCase):
         '''))
         tc.assertTrue(hasattr(root.children[0],
                               '_the_handler_was_actually_called'))
-
-    def test_all_properties_are_evaluated_before_on_kv_post_is_fired(self):
-        ae = self.assertEqual
-
-        Builder.load_string(textwrap.dedent('''
-        <TestLabel>:
-            text: 'A'
-            on_kv_post: self.text += 'B'
-            Label:
-                text: root.assert_the_text_hasnt_changed_yet() or 'hello'
-        <OtherTestLabel@TestLabel>:
-            on_kv_post: self.text += 'C'
-            height: self.assert_the_text_hasnt_changed_yet() or 200
-            Label:
-                text: root.assert_the_text_hasnt_changed_yet() or 'hello2'
-        '''))
-
-        class TestLabel(Factory.Label):
-            def assert_the_text_hasnt_changed_yet(self):
-                ae(self.text, 'A')
-
-        root = Builder.load_string(textwrap.dedent('''
-        BoxLayout:
-            OtherTestLabel:
-                id: label
-                on_kv_post: self.text += 'D'
-        '''))
-        ae(root.ids.label.text, 'ADCB')
--- a/kivy/tests/test_lang_pre_process_and_post_process.py
+++ b/kivy/tests/test_lang_pre_process_and_post_process.py
@@ -184,7 +184,7 @@ class LangTestCase(unittest.TestCase):
                 TestBoxLayout:
         '''))
 
-    def test_order(self):  # TODO: need a good method name
+    def test_property_is_evaluated_before_on_kv_post_is_fired(self):
         ae = self.assertEqual
 
         class TestLabel(Factory.Label):
--- a/kivy/tests/test_lang_pre_process_and_post_process.py
+++ b/kivy/tests/test_lang_pre_process_and_post_process.py
@@ -1,0 +1,176 @@
+import unittest
+import textwrap
+from kivy.lang import Builder
+from kivy.factory import Factory
+from kivy.properties import ObjectProperty, NumericProperty, BooleanProperty
+
+
+KV_CODE = '''
+<EventCounter>:
+    on_kv_pre: self.on_kv_pre_from_c()  # This line won't be excuted
+    on_kv_post: self.on_kv_post_from_c()
+
+<TestBoxLayout>:
+    Label:
+        id: label
+        text: textinput.text
+    TextInput:
+        id: textinput
+    Button:
+        id: button
+        on_press: self.text = 'pressed'
+
+<TestLabel>:
+    text: 'A'
+    on_kv_post: self.text += 'B'
+    Label:
+        text: root.assert_the_text_hasnt_changed_yet() or 'hello'
+<OtherTestLabel@TestLabel>:
+    on_kv_post: self.text += 'C'
+    height: self.assert_the_text_hasnt_changed_yet() or 200
+    Label:
+        text: root.assert_the_text_hasnt_changed_yet() or 'hello2'
+'''
+KV_FILENAME = 'test_kv_filename'
+
+
+def setUpModule():
+    Builder.load_string(KV_CODE, filename=KV_FILENAME)
+
+
+def tearDownModule():
+    Builder.unload_file(KV_FILENAME)
+    u = Factory.unregister
+    u('EventCounter')
+    u('TestWidget')
+    u('TestBoxLayout')
+    u('TestLabel')
+    u('OtherTestLabel')
+
+
+class LangTestCase(unittest.TestCase):
+
+    def test_how_may_times_handlers_are_called(self):
+        '''NOTE: Event handlers are supposed to be called in the order below:
+                 'root rule' -> 'class rule' -> 'default handler'
+        '''
+        testcase = self
+        ae = self.assertEqual
+        NP = NumericProperty
+
+        class EventCounter(Factory.Widget):
+            does_have_root_rule = BooleanProperty(True)
+            _n_pre_from_d = NP(0)  # 'd' stands for 'default handler'
+            _n_post_from_r = NP(0)  # 'r' stands for 'root rule'
+            _n_post_from_c = NP(0)  # 'c' stands for 'class rule'
+            _n_post_from_d = NP(0)
+
+            def on_kv_pre(self):
+                self._n_pre_from_d += 1
+                ae(self._n_pre_from_d, 1)
+                ae(self._n_post_from_r, 0)
+                ae(self._n_post_from_c, 0)
+                ae(self._n_post_from_d, 0)
+
+            def on_kv_pre_from_c(self):
+                testcase.fail('This method is not supposed to be called')
+
+            def on_kv_pre_from_r(self):
+                testcase.fail('This method is not supposed to be called')
+
+            def on_kv_post(self):
+                self._n_post_from_d += 1
+                self.assert_all_handlers_were_called_correctly()
+
+            def on_kv_post_from_c(self):
+                self._n_post_from_c += 1
+                ae(self._n_pre_from_d, 1)
+                ae(self._n_post_from_r, 1 if self.does_have_root_rule else 0)
+                ae(self._n_post_from_c, 1)
+                ae(self._n_post_from_d, 0)
+
+            def on_kv_post_from_r(self):
+                if not self.does_have_root_rule:
+                    testcase.fail(
+                        "Strange. The handler was called even though "
+                        "the widget doesn't have root rule.")
+                self._n_post_from_r += 1
+                ae(self._n_pre_from_d, 1)
+                ae(self._n_post_from_r, 1)
+                ae(self._n_post_from_c, 0)
+                ae(self._n_post_from_d, 0)
+
+            def assert_all_handlers_were_called_correctly(self):
+                ae(self._n_pre_from_d, 1)
+                ae(self._n_post_from_r, 1 if self.does_have_root_rule else 0)
+                ae(self._n_post_from_c, 1)
+                ae(self._n_post_from_d, 1)
+
+        # Test #1: Without root rule
+        root = EventCounter(does_have_root_rule=False)
+        root.assert_all_handlers_were_called_correctly()
+
+        # Test #2: With root rule
+        root = Builder.load_string(textwrap.dedent('''
+        EventCounter:
+            on_kv_pre: self.on_kv_pre_from_r()  # won't be excuted
+            on_kv_post: self.on_kv_post_from_r()
+            EventCounter:
+                id: child
+                on_kv_pre: self.on_kv_pre_from_r()  # won't be excuted
+                on_kv_post: self.on_kv_post_from_r()
+        '''))
+        root.assert_all_handlers_were_called_correctly()
+        root.ids.child.assert_all_handlers_were_called_correctly()
+
+        # Test #3: If the user add a widget during 'on_kv_pre()',
+        #          is 'on_kv_post' still dispatched once on that widget?
+        class TestWidget(Factory.Widget):
+            def on_kv_pre(self):
+                self._ec = EventCounter(does_have_root_rule=False)
+                self.add_widget(self._ec)
+        root = TestWidget()
+        root._ec.assert_all_handlers_were_called_correctly()
+
+    def test_rule_is_applied_on_kv_post(self):
+        tc = self
+
+        class TestBoxLayout(Factory.BoxLayout):
+            def on_kv_post(self):
+                ids = self.ids
+                tc.assertIn('textinput', ids)
+                tc.assertIn('label', ids)
+                tc.assertIn('button', ids)
+
+                textinput = ids.textinput
+                label = ids.label
+                label.text = 'A'
+                textinput.text = 'B'
+                tc.assertEqual(label.text, 'B')
+
+                button = ids.button
+                button.text = ''
+                button.dispatch('on_press')
+                tc.assertEqual(button.text, 'pressed')
+
+                tc.assertTrue(self.parent is not None)
+
+        root = Builder.load_string(textwrap.dedent('''
+        Widget:
+            TestBoxLayout:
+        '''))
+
+    def test_order(self):  # TODO: need a good method name
+        ae = self.assertEqual
+
+        class TestLabel(Factory.Label):
+            def assert_the_text_hasnt_changed_yet(self):
+                ae(self.text, 'A')
+
+        root = Builder.load_string(textwrap.dedent('''
+        BoxLayout:
+            OtherTestLabel:
+                id: label
+                on_kv_post: self.text += 'D'
+        '''))
+        ae(root.ids.label.text, 'ADCB')
--- a/kivy/tests/test_lang_pre_process_and_post_process.py
+++ b/kivy/tests/test_lang_pre_process_and_post_process.py
@@ -102,16 +102,11 @@ class LangTestCase(unittest.TestCase):
         root._ec1.assert_all_handlers_were_called_correctly()
         root._ec2.assert_all_handlers_were_called_correctly()
 
-    def test_all_rules_are_applied_before_on_kv_post_is_fired(self):
+    def test_each_rule_is_applied_at_proper_timing(self):
         tc = self
 
         class TestBoxLayout(Factory.BoxLayout):
-            def on_kv_post(self, root_widget):
-                self._the_handler_was_actually_called = True
-
-                # ---------------
-                # test children
-                # ---------------
+            def assert_my_own_rule_is_applied(self):
                 ids = self.ids
                 tc.assertIn('textinput', ids)
                 tc.assertIn('label', ids)
@@ -130,24 +125,49 @@ class LangTestCase(unittest.TestCase):
                 button.dispatch('on_press')
                 tc.assertEqual(button.text, 'pressed')
 
-                # ---------------
-                # test parent
-                # ---------------
+            def assert_the_rule_i_participate_in_is_applied(
+                    self, reverse=False):
                 parent = self.parent
+                assertTrue = tc.assertFalse if reverse else tc.assertTrue
+                assertEqual = tc.assertNotEqual if reverse else tc.assertEqual
 
                 # check 'parent' property
-                tc.assertTrue(parent is not None)
+                assertTrue(parent is not None)
+
+                # check property binding
+                parent.x = 0
+                parent.y = 50
+                assertEqual(parent.x, 150)
+
+                # check event handler
+                parent.x = 0
+                parent.dispatch('on_press')
+                assertEqual(parent.x, 200)
+
+            def assert_the_rule_i_dont_participate_in_is_applied(
+                    self, reverse=False):
+                parent = self.parent
+                assertTrue = tc.assertFalse if reverse else tc.assertTrue
+                assertEqual = tc.assertNotEqual if reverse else tc.assertEqual
+
+                # check 'parent' property
+                assertTrue(parent is not None)
 
                 # check property binding
                 parent.height = 1
                 parent.width = 50
-                tc.assertEqual(parent.height, 100)
+                assertEqual(parent.height, 100)
 
                 # check event handler
                 parent.height = 1
                 parent.dispatch('on_press')
-                tc.assertEqual(parent.height, 123)
+                assertEqual(parent.height, 123)
 
+            def on_kv_post(self, root_widget):
+                self._the_handler_was_actually_called = True
+                self.assert_my_own_rule_is_applied()
+                self.assert_the_rule_i_participate_in_is_applied()
+                self.assert_the_rule_i_dont_participate_in_is_applied()
         root = Builder.load_string(textwrap.dedent('''
         <TestBoxLayout>:
             Label:
@@ -159,6 +179,8 @@ class LangTestCase(unittest.TestCase):
                 id: button
                 on_press: self.text = 'pressed'
         <TestButton@Button>:
+            x: self.y + 100
+            on_press: self.x = 200
             TestBoxLayout:
         TestButton:
             height: self.width * 2
--- a/kivy/tests/test_lang_pre_process_and_post_process.py
+++ b/kivy/tests/test_lang_pre_process_and_post_process.py
@@ -35,7 +35,7 @@ class LangTestCase(unittest.TestCase):
             def on_kv_pre_from_r(self):
                 testcase.fail('This method is not supposed to be called')
 
-            def on_kv_post(self):
+            def on_kv_post(self, root_widget):
                 self._n_post_from_d += 1
                 self.assert_all_handlers_were_called_correctly()
 
@@ -106,7 +106,7 @@ class LangTestCase(unittest.TestCase):
         tc = self
 
         class TestBoxLayout(Factory.BoxLayout):
-            def on_kv_post(self):
+            def on_kv_post(self, root_widget):
                 self._the_handler_was_actually_called = True
 
                 # ---------------
--- a/kivy/tests/test_lang_pre_process_and_post_process.py
+++ b/kivy/tests/test_lang_pre_process_and_post_process.py
@@ -111,7 +111,7 @@ class LangTestCase(unittest.TestCase):
         root._ec1.assert_all_handlers_were_called_correctly()
         root._ec2.assert_all_handlers_were_called_correctly()
 
-    def test_rule_is_applied_on_kv_post(self):
+    def test_all_rules_are_applied_before_on_kv_post_is_fired(self):
         tc = self
 
         class TestBoxLayout(Factory.BoxLayout):
@@ -163,7 +163,7 @@ class LangTestCase(unittest.TestCase):
             on_press: self.height = 123
         '''))
 
-    def test_property_is_evaluated_before_on_kv_post_is_fired(self):
+    def test_all_properties_are_evaluated_before_on_kv_post_is_fired(self):
         ae = self.assertEqual
 
         Builder.load_string(textwrap.dedent('''
--- a/kivy/tests/test_lang_pre_process_and_post_process.py
+++ b/kivy/tests/test_lang_pre_process_and_post_process.py
@@ -116,6 +116,7 @@ class LangTestCase(unittest.TestCase):
                 textinput = ids.textinput
                 label = ids.label
                 label.text = 'A'
+                textinput.text = ''
                 textinput.text = 'B'
                 tc.assertEqual(label.text, 'B')
 
@@ -132,10 +133,11 @@ class LangTestCase(unittest.TestCase):
                 assertEqual = tc.assertNotEqual if reverse else tc.assertEqual
 
                 # check 'parent' property
-                assertTrue(parent is not None)
+                tc.assertTrue(parent is not None)
 
                 # check property binding
                 parent.x = 0
+                parent.y = 0
                 parent.y = 50
                 assertEqual(parent.x, 150)
 
@@ -151,10 +153,11 @@ class LangTestCase(unittest.TestCase):
                 assertEqual = tc.assertNotEqual if reverse else tc.assertEqual
 
                 # check 'parent' property
-                assertTrue(parent is not None)
+                tc.assertTrue(parent is not None)
 
                 # check property binding
                 parent.height = 1
+                parent.width = 0
                 parent.width = 50
                 assertEqual(parent.height, 100)
 
@@ -163,8 +166,15 @@ class LangTestCase(unittest.TestCase):
                 parent.dispatch('on_press')
                 assertEqual(parent.height, 123)
 
+            def on_kv_applied(self):
+                self._on_kv_applied_was_actually_fired = True
+                self.assert_my_own_rule_is_applied()
+                self.assert_the_rule_i_participate_in_is_applied(reverse=True)
+                self.assert_the_rule_i_dont_participate_in_is_applied(
+                    reverse=True)
+
             def on_kv_post(self, root_widget):
-                self._the_handler_was_actually_called = True
+                self._on_kv_post_was_actually_fired = True
                 self.assert_my_own_rule_is_applied()
                 self.assert_the_rule_i_participate_in_is_applied()
                 self.assert_the_rule_i_dont_participate_in_is_applied()
@@ -187,4 +197,6 @@ class LangTestCase(unittest.TestCase):
             on_press: self.height = 123
         '''))
         tc.assertTrue(hasattr(root.children[0],
-                              '_the_handler_was_actually_called'))
+                              '_on_kv_applied_was_actually_fired'))
+        tc.assertTrue(hasattr(root.children[0],
+                              '_on_kv_post_was_actually_fired'))
--- a/kivy/tests/test_lang_pre_process_and_post_process.py
+++ b/kivy/tests/test_lang_pre_process_and_post_process.py
@@ -16,7 +16,7 @@ class LangTestCase(unittest.TestCase):
         NP = NumericProperty
 
         class EventCounter(Factory.Widget):
-            does_have_root_rule = BooleanProperty(True)
+            there_is_a_rootrule = BooleanProperty(True)
             _n_pre_from_d = NP(0)  # 'd' stands for 'default handler'
             _n_post_from_r = NP(0)  # 'r' stands for 'root rule'
             _n_post_from_c = NP(0)  # 'c' stands for 'class rule'
@@ -42,12 +42,12 @@ class LangTestCase(unittest.TestCase):
             def on_kv_post_from_c(self):
                 self._n_post_from_c += 1
                 ae(self._n_pre_from_d, 1)
-                ae(self._n_post_from_r, 1 if self.does_have_root_rule else 0)
+                ae(self._n_post_from_r, 1 if self.there_is_a_rootrule else 0)
                 ae(self._n_post_from_c, 1)
                 ae(self._n_post_from_d, 0)
 
             def on_kv_post_from_r(self):
-                if not self.does_have_root_rule:
+                if not self.there_is_a_rootrule:
                     testcase.fail(
                         "Strange. The handler was called even though "
                         "the widget doesn't have root rule.")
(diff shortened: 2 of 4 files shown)


tito:
I would like to say, why not:
- Using Clock have more advantage (layout/size may be stabilized, it would be added to parent too, etc).
- This is for a very specific case to prevent subclassing `__init__` and do your stuff. I usually do the same but name it "late_init". Sometime trigerred by Clock, or direct after `__init__`

What i don't like:
- it add a method call for all widget created
- the naming is bad, using 2 underscore indicate something private, really unsure about that
- why not having the same with Clock too? (conditionnally of course) (EDIT: actually, using @mainthread around the late init would work.)

I would suggest it to just rename it to "late_init" ?

gottadiveintopython:
Thanks for the review.

> I would suggest it to just rename it to "late_init" ?

Yup, I always want someone to correct my English including commit message, comment, etc...

 > Using Clock have more advantage (layout/size may be stabilized, it would be added to parent too, etc).

Well even if you use Clock there is no guarantee that the layout is finished (as you stated "may be stabilized") so there is no much difference compared to the PR. And if you have to do some stuff after the widget is added to parent, I think it's usually not that widget's job but its parent's job (or parent's parent's job...).

> it add a method call for all widget created

Yeah I don't like it too. Overall I think the PR is not worth merging. I'll be happy if someone tell me an awesome implementation.

pythonic64:
> Overall I think the PR is not worth merging. 

I second that and I agree with @tito. There can be cases where you want to use a trigger, `Clock.create_trigger(__post_kv__)`, or you want to schedule callback before the next frame (which is better than on the next frame, in my opinion).

matham:
This is a great idea and thanks for working on this. Something like this has been on the top of my todo/wishlist for a very long time. However, I have some slightly different ideas about this and let me explain why. If others are amenable to this, I hope you'd be ok with continuing to work on what I outline below?

I think every `Widget` should have an `on_kv_pre` and `on_kv_post` event, which gets dispatched before and after, respectively, any and all kv rules of this widget are applied.

This may seem simple, but may not actually be. Consider:

```
<MyWidget@Widget>:
    name: 'matt'
    on_kv_post: self.height = 40

<MyOtherWidget@MyWidget>:
    name: 'other matt'
    on_kv_post: self.height = 41

BoxLayout:
    MyOtherWidget:
        on_kv_post: self.height = 42
```

So the first thing to ensure is that `self.height` gets assigned first `40`, then `41`, and finally `42`. 

But the more important question is, does the line `on_kv_post: self.height = 40` get executed when the `MyWidget` rule is applied (e.g. by manually executing that code at that stage), or after the `BoxLayout` rule is applied. Because `on_kv_post` can mean either dispatch after the current rule is applied (so there are 3 dispatches!?), or after *all* the rules for this widget is applied (one dispatch). To me, it should definitely be the latter, especially if you consider that if I were to write

```py
class MyWidget(Widget):

    def on_kv_post(self, *largs):
        print('rule done')
```

the meaning of `on_kv_post` is not tied to any rule, but rather means "dispatch me when you're done with any kv stuff and everything is fully initialized (inclduing id's)". In my view, this should happen by the builder after it applies all the rules and not be clock scheduled. Because if you jump to the next clock cycle, things get messier because there's some violation of the expectation that it is applied right after the rules are done, and not at a later time.

For `on_kv_pre`, it should be dispatched before anything kv happens related to this widget and this gives the chance to init stuff that may be needed in kv rules (i.e. stuff that traditionally happens on `on_parent`).

Together, these two events give a better kv/python interaction, like what I think @gottadiveintopython wants.

This is something I had planned and added in the first version of the compiler (that @tito had started working on), but that faltered after I ran into issues with the compiler so I never followed this up. But this has been at the top of my to list and I think would be immensely helpful to clean up some of the kv/python issues that new users encounter as it would provide a way to do stuff from python before/after the kv rules.

gottadiveintopython:
Thanks for the review.

@pythonic64 
This PR doesn't prevent you from using `Clock`. You can use it if you want to. But `Clock` is not a good workaround in some cases. I'll post a example of that if I come up with.

@matham 

Yup, if event can be used it would be great. But a couple of thing bothering me.

### 1. The order

> So the first thing to ensure is that self.height gets assigned first 40, then 41, and finally 42.

This code:

```python3
layout = Builder.load_string('''
<MyButton@Button>:
    on_press: print('A')

<MyOtherButton@MyButton>:
    on_press: print('B')

BoxLayout:
    MyOtherButton:
        id: button
        on_press: print('C')
''')
layout.ids.button.dispatch('on_press')
```

outputs this:

```
C
B
A
```

So I think your example doesn't work as you expect.
I mean `self.height` gets assigned first 42, then 41, and finally 40.

### 2. `on_kv_pre` might confuse the user

If the user write a code like this:

```python3
widget = Builder.load_string('''
Widget:
    on_kv_pre: print('A')
''')
```

`print('A')` isn't excuted because `print('A')` is binded **after** the kv rule is applied.

kivy/lang/builder.py
```diff
@@ -583,6 +583,7 @@ def _apply_rule(self, widget, rule, rootrule, template_ctx=None,
                 widget.add_widget(child)
                 self.apply(child)
                 self._apply_rule(child, crule, rootrule)
+                child.late_init()
```
gottadiveintopython: I feel this is not a right place to call `late_init()`

matham:
> So I think your example doesn't work as you expect.
I mean self.height gets assigned first 42, then 41, and finally 40.

This is unfortunate but comes from the fact that on_xxx events dispatch in the reverse order. The reason why I thought the order was the opposite is because when you run:

```py
from kivy.lang import Builder

layout = Builder.load_string('''
<MyButton@Button>:
    a: print('A')

<MyOtherButton@MyButton>:
    b: print('B')

BoxLayout:
    MyOtherButton:
        id: button
        c: print('C')
''')
layout.ids.button.dispatch('on_press')
```

It prints:

```
A
B
C
```


> print('A') won't be excuted because print('A') will be binded during the kv rule is applied, which means after on_kv_pre event was triggered.

`on_kv_pre` is meant to be dispatched before any kv happens. This means that `on_kv_pre` will have to be handled differently and that it's rules will have to be bound to the `on_kv_pre` event before any rules are applied. In other words the builder will have to bind the `on_kv_pre` code before anything else and then dispatch it. It won't be bound again when the rest of the kv rules are applied.

gottadiveintopython:
So in order to implement your ideas we have to do two things, which is

1. give `EventDispatcher` the ability to dispatch in forward order
2. make `Builder` handle `on_kv_pre` differently (as you stated)

I feel it's not good design, especially 2.
Do other core developers second it?

matham:
I don't think we should do 1. I had forgotten that the event dispatching
order was flipped when I mentioned that and the order was not my real main
point. Although it did seem that way from my writing. I was just trying to
illustrate that it should be event like where you can bind multiple things
to the event across different rules and it'll be dispatched to all of them.

The order is not really important and it should be consistent with the
other events.

Number 2 I do think should be done.

matham:
I can see why `#2` would seem like bad design, because it's a kv rule that is executed, but it's executed before kv is supposed to have been applied, which is kind of a contradiction. 

I don't feel too strongly about this. I think it's ok if `on_kv_post` is implemented as I mentioned, and if `on_kv_pre` is implemented as an event that is dispatched by builder before rules are applied, but it has no kv support. I.e. that `on_kv_pre: print(42)` would never do anything.

It can definitely start that way, and support for `#2` can be added later if it was really wanted. The main benefit is that this would let you init stuff in python before anything kv happens.

KV rule application
----------------------

There was something related that I didn't plan to bring up here before. But I have wanted `Builder.apply(self, ignored_consts=self._kwargs_applied_init)` in Widget's constructor to be moved out into its own function instead. The function would similarly be called in the constructor, but users could overwrite it in subclasses if they wanted to apply the kv rules in a different order.

Or even better(?), builder would call this function on the widget after the widget is constructed to apply the rules. This would have a similar outcome where the user could inherit and overwrite this function if they wanted to init stuff before the rules are applied. Right now this is much harder to do and your only option is to init stuff before calling super in a subclass init. But that is not great and I think it would be better if users could overwrite that behavior.

I'm just bringing this up because to me, this is all related. In fact over the last couple of weeks I have multiple times thought about coding this up, but didn't because I didn't have the time and energy to push this kind of thing through.

gottadiveintopython:
> The order is not really important and it should be consistent with the other events.

I got it.

One of my concern is how to implement. `Builder` has to apply `on_kv_pre` first, then trigger it, then applies rest of the rule, then trigger `on_kv_post`, which means `Builder` has to apply rules partially. I think that's hard to implement and makes `kivy/lang/*.py` complex. I might try it though.

More easy way that I can come up with is:

1. trigger `on_kv_pre` before the rule is applied
1. apply the rule
1. trigger `on_kv_post`

In this case, `on_kv_pre`rule  in kv won't be excuted, so overriding the default handler is the only way to do stuff before the rule applied.

```python3
class MyWidget(Widget):
    def on_kv_pre(self):
        pass
```

This might be not what you want but it won't make `kivy/lang/*.py` complex.

> Builder.apply(self, ignored_consts=self._kwargs_applied_init) in Widget's constructor to be moved out into its own function instead. 

But what if `no_builder` is `True`.

https://github.com/kivy/kivy/blob/6e1fb402188ace0186f90417806db242e16a25e3/kivy/uix/widget.py#L346-L348

It's conditional, so I think It can't be moved out easily.

> Or even better(?), builder would call this function on the widget after the widget is constructed to apply the rules.

Sorry I don't quite get what this means. If `Builder` call that function on the widget **after** the widget is constructed, who call that function on the widget in the code below?

```python3
widget = MyWidget()  # <= who calls 'that' function on this widget?
```

I probably misunderstand something, but anyways thanks for review.

matham:
> This might be not what you want but it won't make kivy/lang/*.py complex.

I'm actually fine with this and I think now that this is how it should look initially. We can always add support in kv code for `on_kv_pre` later. This is what I was trying to say in the second paragraph of the last post.

> Sorry I don't quite get what this means.

Let me clarify as I wasn't very clear. We can add the following function to `Widget`:
```py
def apply_kv_rules(self, builder, local_rule=None):
    builder.apply(self, ignored_consts=self._kwargs_applied_init)
    if local_rule is not None:
        builder._apply_rule(self, *local_rule)
```

Then https://github.com/kivy/kivy/blob/master/kivy/uix/widget.py#L347 becomes
```py
if not no_builder:
    self.apply_kv_rules(Builder)
```
This deals with the case where the user just instantiates a widget. For the case where the widget is instantiated by Builder, https://github.com/kivy/kivy/blob/master/kivy/lang/builder.py#L582 will become
```py
child = cls(__no_builder=True)
widget.add_widget(child)
child.apply_kv_rules(self, (crule, rootrule))
```

It can be made a little friendly, e.g. `builder._apply_rule(self, *local_rule)` could a more official API, but the main point is that rule application would happen through an explicit widget function, `apply_kv_rules` that the user could inherit and overwrite and call later if they wanted. Or they could execute some code just before they call `super(...).apply_kv_rules(..)` etc.

matham:
> doesn't call Builder.apply() there whereas your code does.

But it does! The line `self.apply(child)`, isn't `self` in that case `Builder`?

> But since I don't fully understand kivy/lang/*.py, if you think "that's not a problem" I'll follow you.

Come on, don't sell yourself short :) If you think it's an issue it's important to speak up about it. Just because I'm maybe more familiar with the code doesn't mean I'm right or have not forgotten things...

In any case, such a `apply_kv_rules` function would not be part of this PR and can be added later if there's interest (by you and other core devs).

kivy/tests/test_lang_pre_process_and_post_process.py
```diff
@@ -0,0 +1,108 @@
...
+                button.text = ''
+                button.dispatch('on_press')
+                tc.assertEqual(button.text, 'pressed')
+
+                tc.assertTrue(self.parent is not None)
+
+        root = Builder.load_string(textwrap.dedent('''
+        Widget:
+            TestBoxLayout:
+        '''))
```
matham: There should be tests for more complex situations. E.g. the code above:

```
<MyWidget@Widget>:
    name: 'matt'
    on_kv_post: self.height = 40
    Label:
        text: print('other') or 'hello'

<MyOtherWidget@MyWidget>:
    name: 'other matt'
    on_kv_post: self.height = 41
    height: print('something') or 90
    Label:
        text: print('other2') or 'hello2'

BoxLayout:
    MyOtherWidget:
        on_kv_post: self.height = 42
```

should set `height` three times, and all should happen *after* all the print statements.

gottadiveintopython:
> But it does!

Yes... you are right. For some reason I thought "it doesn't" for a long time. I don't know why.

>  such a apply_kv_rules function would not be part of this PR

Yeah, I think so too.

> There should be tests for more complex situations

ok I'll add that.

gottadiveintopython:
Hmm, in order to implement `on_kv_post` event, `Widget` may need to have some flag that indicates whether `on_kv_post` was triggered or not, which I think not good design.

kivy/tests/test_lang_pre_process_and_post_process.py
gottadiveintopython: Hi, I think your code above will set `height` **four** times, which are:

- `on_kv_post: self.height = 40`
- `on_kv_post: self.height = 41`
- `height: print('something') or 90`
- `on_kv_post: self.height = 42`

But the more important question is: Are you sure that "all should happen **after** all the print statements"? I think there is a possibility of `height: print('something') or 90` will be excuted **before** other print statements.

matham: I meant the latter - the three  `on_kv_post` height assignments should happen after all the other print statements.

gottadiveintopython: ah ok got it

kivy/tests/test_lang_pre_process_and_post_process.py
```diff
@@ -0,0 +1,176 @@
...
+        root.ids.child.assert_all_handlers_were_called_correctly()
+
+        # Test #3: If the user add a widget during 'on_kv_pre()',
+        #          is 'on_kv_post' still dispatched once on that widget?
+        class TestWidget(Factory.Widget):
+            def on_kv_pre(self):
+                self._ec = EventCounter(does_have_root_rule=False)
+                self.add_widget(self._ec)
+        root = TestWidget()
+        root._ec.assert_all_handlers_were_called_correctly()
```
gottadiveintopython: `Test #3` is the reason why I added a new property `_on_kv_post_was_done`. Without it, `on_kv_post` is fired twice on `self._ec`.

kivy/tests/test_lang_pre_process_and_post_process.py
```diff
@@ -0,0 +1,176 @@
...
+            EventCounter:
+                id: child
+                on_kv_pre: self.on_kv_pre_from_r()  # won't be excuted
+                on_kv_post: self.on_kv_post_from_r()
+        '''))
+        root.assert_all_handlers_were_called_correctly()
+        root.ids.child.assert_all_handlers_were_called_correctly()
+
+        # Test #3: If the user add a widget during 'on_kv_pre()',
+        #          is 'on_kv_post' still dispatched once on that widget?
```
gottadiveintopython: `Test #3` is the reason why I added a new property `_on_kv_post_was_done`. Without the property, `on_kv_post` is fired twice on `self._ec`. So yeah it's not a good design.

matham:
Why did you close it?

gottadiveintopython:
Introducing a flag (`_on_kv_post_was_fired`) is probably a bad thing.

matham:
Okay. I'll take a look in a couple of days to see if I can find another way to implement it without the flag and will let you know. Unless you figure it out first.

gottadiveintopython:
Thank you!  Really.
If you and other core developers don't mind the flag, I'm ok with that though.

matham:
The following diff implements the desired functionality without a flag. Additionally, it dispatches the root widget that instantiated the widget and therefore caused the `on_kv_post` dispatch:

```diff
diff --git a/kivy/lang/builder.py b/kivy/lang/builder.py
index 20292cf59..9108db008 100755
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -396,8 +396,10 @@ class BuilderBase(object):
                 self.files.append(fn)

             if parser.root:
-                widget = Factory.get(parser.root.name)()
+                widget = Factory.get(parser.root.name)(__no_builder=True)
+                self.apply(widget)
                 self._apply_rule(widget, parser.root, parser.root)
+                widget.dispatch('on_kv_post', widget)
                 return widget
         finally:
             self._current_filename = None
@@ -480,9 +482,9 @@ class BuilderBase(object):

         # will collect reference to all the id in children
         assert(rule not in self.rulectx)
-        self.rulectx[rule] = rctx = {
+        self.rulectx[rule] = {
             'ids': {'root': widget.proxy_ref},
-            'set': [], 'hdl': []}
+            'set': [], 'hdl': [], 'widgets': []}

         # extract the context of the rootrule (not rule!)
         assert(rootrule in self.rulectx)
@@ -583,6 +585,7 @@ class BuilderBase(object):
                 widget.add_widget(child)
                 self.apply(child)
                 self._apply_rule(child, crule, rootrule)
+                rctx['widgets'].append(child)

         # append the properties and handlers to our final resolution task
         if rule.properties:
@@ -657,6 +660,10 @@ class BuilderBase(object):
                     '{}: {}'.format(e.__class__.__name__, e), cause=tb)
             raise e

+        root = rctx['ids']['root']
+        for child in rctx['widgets']:
+            child.dispatch('on_kv_post', root)
+
         # rule finished, forget it
         del self.rulectx[rootrule]

diff --git a/kivy/uix/widget.py b/kivy/uix/widget.py
index 5bf822ff7..71a7d80e1 100644
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -316,7 +316,7 @@ class Widget(WidgetBase):
     '''

     __metaclass__ = WidgetMetaclass
-    __events__ = ('on_touch_down', 'on_touch_move', 'on_touch_up')
+    __events__ = ('on_touch_down', 'on_touch_move', 'on_touch_up', 'on_kv_post')
     _proxy_ref = None

     def __init__(self, **kwargs):
@@ -350,6 +350,7 @@ class Widget(WidgetBase):
         # Apply all the styles.
         if not no_builder:
             Builder.apply(self, ignored_consts=self._kwargs_applied_init)
+            self.dispatch('on_kv_post', self)

         # Bind all the events.
         if on_args:
@@ -492,6 +493,9 @@ class Widget(WidgetBase):
             if child.dispatch('on_touch_up', touch):
                 return True

+    def on_kv_post(self, root_widget):
+        pass
+
     #
     # Tree management
     #

```

gottadiveintopython:
Thanks! I get the idea of your patch.

I didn't realize that `__rootrule` isn't necessary. I'm going to fix that part.

But, your patch doesn't pass the test below:

```python3
from kivy.lang import Builder

<TestWidget@Widget>:
    on_kv_post:
        self.parent.height = 30
        assert self.parent.width == 60  # => AssertionError
<TestWidget2@Widget>:
    TestWidget:
TestWidget2:
    width: self.height * 2
''')
```

so

```python
        root = rctx['ids']['root']
        for child in rctx['widgets']:
            child.dispatch('on_kv_post', root)
```

is not the right place to `dispatch()`, I think.

gottadiveintopython:
and in order to pass the test above, I think it needs [this kind of fix](https://github.com/gottadiveintopython/kivy/commit/f6b944cedca2773e2d3e4790ba14313b79eb5cb3)(rough code), which is not good code I guess. So I cannot decide the way to implement this PR. I mean **A) using the flag** or **B) without the flag(your code)**. Which one do you think the better?

matham:
>But, your patch doesn't pass the test below

I don't think this test is valid. `TestWidget` should dispatch when the kv rule it participates in directly is finished being applied. It does not have to wait to dispatch until all the rules that its parent participates in is done.

So its valid for it to fail that assert.

matham:
That's why in my diff it includes the root rule as a parameter when dispatching the event. It's because it dispatches all the widgets created **in** the rule when the rule is done.

gottadiveintopython:
But when I changed the code like this:

```python3
from kivy.lang import Builder

<TestWidget@Widget>:
    on_kv_post:
        self.parent.height = 30
        assert self.parent.width == 60  # => AssertionError
<TestWidget2@Widget>:
    width: self.height * 2  # changed
    TestWidget:
TestWidget2:
''')
```

`AssertionError` didn't occured, which means `on_kv_post` is fired before the root rule is applied but after the class rule is applied. Is that really the behavior you want?

gottadiveintopython:
I thought `on_kv_post` should be fired after applying all rules (including parent's), but you think it should be fired **before** applying parent's rules. I don't know which one is better :smile: .

matham:
`on_kv_post` should fire on a widget after 

* all the rules that this widget is the root of has been applied and the rule, and
* after the rule that created that widget has been completely applied.

This is the minimum amount of time before `on_kv_post` can fire, and that's the precise moment it should fire. It cannot fire before that because the rules it participates in is not applied. But, there's no reason to wait further before firing. We *shouldn't* wait to fire until the widget's parent rules, completely remote from any rule the widgets is participating in, is fired.

>AssertionError didn't occured, which means on_kv_post is fired before the root rule is applied

Which root rule?

tito:
Just a question: why do you guys consider that calling the `on_kv_post` should be the same for:
- the creation of a widget and its own rules applied
- the creation of a widget as a child in kv, that may or may not have additionnal rules and delay the `on_kv_post`

IMO, the second should not be part of the `on_kv_post`. As soon as a widget is created, after the rules are applied, it will be called. Aka after a call of `Builder.apply`.

If you think about the second, it will behave completely differently if the "tree" is created in KV or in the widget itself, and confuse even more about "when" the on_kv_post will be used.

gottadiveintopython:
@matham 
> Which root rule?

`root rule` is

```
WidgetClass:
    something...
```

and `class rule` is

```
<WidgetClass>:
    something...
```

isn't it? (I might misunderstand the terms.)
So there is only one `root rule` in my previous reply, which is:

```
TestWidget2:
```

and there are two `class rule`s  in my previous reply, which is:

```
<TestWidget@Widget>:
    on_kv_post:
        self.parent.height = 30
        assert self.parent.width == 60  # => AssertionError
```

and

```
<TestWidget2@Widget>:
    TestWidget:
```

So by `root rule` I meant `TestWidget2:`.

@tito 
If the second is not part of the `on_kv_post`, the code of this PR would become really clean. But just a bit worrying about the convenience.

@matham 
> after the rule that created that widget has been completely applied

So you think the code below:

```python3
<TestWidget@Widget>:
    on_kv_post:
        self.parent.height = 30
        print(self.parent.width)
<TestWidget2@Widget>:                  # B
    width: self.height * 2             # B
    TestWidget:                        # B
TestWidget2:                           # C
    width: self.height * 3             # C
''')
```

should print `60` not `90` because `B` is the rule that created the `TestWidget` but `C` isn't?

tito:
That's my point, it became a mess trying to figure out "when" it will be actually dispatched. I don't see immediate value of the second, and also introduce a break between how kv or python would be dispatched. Consider:

```
KV = '''
<AWidget@Widget>
<MyWidget>:
    AWidget
'''

class MyWidget(Widget):
    pass
```

Versus:

```
KV = '''
<AWidget@Widget>
'''

class MyWidget(Widget):
    def __init__(self, **kwargs):
        super(MyWidget, self).__init__(**kwargs)
        self.add_widget(AWidget())
```

The first would dispatch AWidget `on_kv_post` event in the KV when the MyWidget is created, and so AWidget is aware of his parent.
But not in the Python version where the event is dispatched before MyWidget
And even seeing the code, the event may not be dispatched if it's embedded and used in another rules.

IMO it's just a mess because then the event depend of the "upper construction", and not dispatched at "widget" level.

The first behavior i described should have its own event.
If you want the second, it should be another event.

matham:
For me, the meaning of the event is that all the kv rules that apply to this widget and used by the widget has been done applying. 

What this specifically means for me is that

* the `ids` are fully initialized. I.e. the following should work

  ```
  <MyWidget@Widget>:
      Label:
          id: a
          on_kv_post: print(root.ids.b.text)
      Label:
          id: b
  ```
  
  If we fire it after apply like you suggests, this will fail because `ids` would not be yet initialized.
  The main usage of this is from python, not in kv, where we want to use the ids in `on_kv_post`.
* From within kv, the following should work:

  ```
  <MyWidget@Widget>:
      value: self.width
      Label:
          value: root.something
          on_kv_post: self.height = root.value * self.value + a.value
      Widget:
          Label:
              id: a
              value: root.other
  ```
  
  Specifically, when `on_kv_post` is fired, all the other `values` should have been initialized because all the rules should have been applied.
  Right now, you have to schedule it for the next clock cycle or use `on_parent` if you want to do this. Although I'm not sure `on_parent` works
  for these example TBH.

In other words, `on_kv_post` means that all the rules that widget participates in has been finished applying and that's when it's fired. That's
how we'd explain it to users, which I think is intuitive enough.

I don't really see that `on_kv_post` behaves differently when manually instantiated vs when kv instantiates it. In but cases `on_kv_post` is only
triggered when all the rules the widget participates in is applied.

matham:
>The first would dispatch AWidget on_kv_post event in the KV when the MyWidget is created, and so AWidget is aware of his parent.
But not in the Python version where the event is dispatched before MyWidget

I don't think this is a weakness in how we dispatch `on_kv_post`. The reason why in the first it's aware of its parent is because kv handles adding the widget to the parent, while if you instantiate in python you're responsible for adding to the parent. I'm not sure why you expect in the python case for `on_kv_post` to know anything about the parent?

I'm fine with adding a third event for the case you wish to have where it's dispatched right after `apply`. But for most cases I have seen, the behavior of the described `on_kv_post` is the most useful.

matham:
@gottadiveintopython 

>should print 60 not 90 because B is the rule that created the TestWidget but C isn't?

yes

tito:
I was using parent as an excuse for my argument, but it doesn't work.
My point was that the first event directly after `apply` is what was initially intented in the PR and works. And you can easily make it run "later" with `@mainthread` decorator. It just prevent you do actually write the `__init__`, even if it will slow down every widget creation because of the dispatching.

Now you introduce another idea that wasn't targetted by the initial PR, which is having an event dispatched only once after direct or indirect instantiation. That's something else no?

matham:
You're right that this is a different idea than the original PR. I was hoping that I convinced @gottadiveintopython to change his idea to what I presented becaud I thought mine was more suited for the problem.

If you want to keep these ideas separate, that is fine and I could open a different PR with what I presented and leave this with the original idea.

I still think that what I presented is a better solution because it takes care of the issues I have commnly encountered before, but I'm OK if we include both ideas.

tito:
@matham i'm all for understanding the issues. Can you give me few real-life examples?

matham:
For examples, 

* the `ids` always keeps coming up by new users. A typical example is https://github.com/kivy/kivy/issues/2479. After hitting it you learn to never access `ids` from `__init__` and instead schedule its usage with Clock. Personally, I avoid `ids` altogether because I don't want to use the Clock as it's too much of an hassle. 

  With my proposed change, I'd always do any python init in `on_kv_post` and not do anything in `__init__`. That will be a easy rule for me to remember so I don't have to think whether I can use `ids`.
* The other examples are those that are solved with `on_parent` in current code. I have many such cases in my code and I'm sure you have also so I'm not going to list these examples. Using `on_parent` is a hack IMHO because if the parent changes it will be triggered again.
  In some cases you can avoid this by doing something like `on_parent: if self.parent and self.parent == some_id: do_something`, but this doesn't work for something more complex like `on_parent: if self.parent and self.parent == some_id: self.x = y` or for multi-line rules.

  The proposed dispatching of `on_kv_post` is very similar to how `on_parent` is currently dispatched and I'm seeing `on_kv_post` as replacing `on_parent` and as something specific for this that we can explicitly teach our users to use. I don't think the original proposal in the PR could replace `on_parent`.
* This is similar to the `ids` problem, but even if `ids` never existed, if you need to use some properties set in kv, you would need to schedule with the clock because you couldn't read them from `__init__` as they may or may not have been set after `super` depending if the widget is created in kv or python. 

  This is something I found very annoying in the first few years of kivy and is something new users always complain about that they cannot really use python with widgets created in kv. If you look at the mailing list survey thread, this always comes up. I just learned to avoid this so today I don't even think about this anymore, but if `on_kv_post` was available it would essentially resolve that problem and would be more much more convenient.

  For a specific example, I'm so used to not being able to do this and work around it that it's hard to think of one anymore, and the examples I do have are dealt with using `on_parent`. But this would be useful any time that you need to access a property that is set in kv from python code, e.g. to initialize something that you'd want to do in `__init__` and now have to schedule with `Clock` instead.


We didn't talk about `on_kv_pre`, but to me, that's probably even more useful.

* One of the most common issues I run into is the following:

  ```
  <MyWidget>:
      prop: self.some_obj.value if self.some_obj is not None else 55
  ```
  
  If we had `on_kv_pre`, I'd in python make sure that `some_obj` is not None before any kv is applied so I could then just write `prop: self.some_obj.value`.

  This can be done today by initializing `some_obj` before calling `super` but sometimes this doesn't work if you need to first call `super` e.g. to handle `kwargs` passed to `super`. With `on_kv_pre`, it would fire just before the rule is applied and after `super` has been called, letting you init `some_obj` just before it's required without having to think about `__init__`.
* A very related issue is about `rebind`. If we want to be able to write `prop: self.some_obj.other_obj.value`, then in addition to having to check if `some_obj` is None like above, in the case where initialization happens on the next clock cycle, you now have to make `some_obj` `rebind` be `True` because when otherwise initialized on the next cycle or after `super` is called, it would need to rebind in order to track `value`. For me this is a common usage of `rebind`. If `on_kv_pre` existed, I can just make sure to init everything there.

These are most of the issues that I can think of and that have been on my mind for a long time and was partially what motivated me to work on the compiler. Especially the `on_kv_pre` parts.

gottadiveintopython:
What I was trying to do at first is, calling `__post_kv__()` on a widget after all rules (including parent's rule) are applied. So it's close to the current state of the PR except it doesn't use event.

But I'm not sticking to that idea, and I want you core developers to decide the design.
Personally, I prefer @matham 's idea even though it's confusing in some cases (e.g. the example in my last comment), because of its convenience.

gottadiveintopython:
I mean I want you core developers to decide the place where `on_kv_post` is fired on a widget.

1. when its own rules are applied (@tito 's idea)
1. when its own rules and the rules the widget participates in are applied (@matham 's idea)
1. when all rules are applied (the current state of the PR)
1. another place

matham:
To expand why I think 2, rather than 3 is correct, consider your example:

```py
<TestWidget@Widget>:
    on_kv_post:
        self.parent.height = 30
        print(self.parent.width)
<TestWidget2@Widget>:                  # B
    width: self.height * 2             # B
    TestWidget:                        # B
TestWidget2:                           # C
    width: self.height * 3             # C
''')
```

IMO it should print `60`, rather than `90`. 

To be clear, the difference is that option 3 would fire after all kv rules are applied for the highest parent in the widget tree, in this case after `C` is done. For option 2, it would fire after the rules the widget is participating in is done applying, i.e. after `B` is done.

So the question is, how do rules the widget is not directly participating in affect the the widget and what is the benefit in delaying firing until they are done? TBH I'm not completely sure what the answer is and I'm open to option 3.

A semi-realistic example pro option 3 is 

```py
<MyWidget@Widget>:
    Widget:
        on_kv_post: print(self.parent.ids.my_name)

MyWidget:
    Label:
        id: my_name
        text: 'Frankenstein'
''')
```

This would still fail under option 2 because the parent's `ids` is not done when it's fired. It's not  specific to `ids` but rather anything requiring initialization.

However, I'm not sure that option 3 would be correct for this. IMO, if someone wants to do something like this they should move `on_kv_post: print(self.parent.ids.my_name)` into the `Label` or the latter `MyWidget` rule because the current example should be considered buggy and not a common usage. That's because in a inheritance structure, a parent class should **not** refer to a property of  child class, rather a child should refer to the property of the parent. In this case, the latter `MyWidget` rule is the child and `<MyWidget@Widget>` should not be expected to know anything about it. 

So I don't think we should delay firing according to option 3 to accommodate this example.

Also, the difference between option 2 and 3 is not coming at a non-zero cost. The longer you delay firing the event, the less likely it is to be useful because the longer you wait to do post initialization stuff, the more likely things are to have changed between widget creation and the event firing.

But, I'd be okay with option 3 if that's decided on. Perhaps I'll even try to do a very clean implementation  when I get a chance.

matham:
Here's a diff that implements option 3 without flags:

```diff
diff --git a/kivy/lang/builder.py b/kivy/lang/builder.py
index 20292cf59..7e945bfa4 100755
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -396,8 +396,16 @@ class BuilderBase(object):
                 self.files.append(fn)

             if parser.root:
-                widget = Factory.get(parser.root.name)()
-                self._apply_rule(widget, parser.root, parser.root)
+                widget = Factory.get(parser.root.name)(__no_builder=True)
+                rule_children = []
+                self.apply(widget, rule_children=rule_children)
+                self._apply_rule(
+                    widget, parser.root, parser.root,
+                    rule_children=rule_children)
+
+                for child in rule_children:
+                    child.dispatch('on_kv_post', widget)
+                widget.dispatch('on_kv_post', widget)
                 return widget
         finally:
             self._current_filename = None
@@ -433,7 +441,8 @@ class BuilderBase(object):
         self._apply_rule(widget, rule, rule, template_ctx=proxy_ctx)
         return widget

-    def apply_rules(self, widget, rule_name, ignored_consts=set()):
+    def apply_rules(
+            self, widget, rule_name, ignored_consts=set(), rule_children=None):
         '''Search all the rules that match `rule_name` widget
         and apply them to `widget`.

@@ -450,9 +459,11 @@ class BuilderBase(object):
         if not rules:
             return
         for rule in rules:
-            self._apply_rule(widget, rule, rule, ignored_consts=ignored_consts)
+            self._apply_rule(
+                widget, rule, rule, ignored_consts=ignored_consts,
+                rule_children=rule_children)

-    def apply(self, widget, ignored_consts=set()):
+    def apply(self, widget, ignored_consts=set(), rule_children=None):
         '''Search all the rules that match the widget and apply them.

         `ignored_consts` is a set or list type whose elements are property
@@ -466,14 +477,16 @@ class BuilderBase(object):
         if not rules:
             return
         for rule in rules:
-            self._apply_rule(widget, rule, rule, ignored_consts=ignored_consts)
+            self._apply_rule(
+                widget, rule, rule, ignored_consts=ignored_consts,
+                rule_children=rule_children)

     def _clear_matchcache(self):
         BuilderBase._match_cache = {}
         BuilderBase._match_name_cache = {}

     def _apply_rule(self, widget, rule, rootrule, template_ctx=None,
-                    ignored_consts=set()):
+                    ignored_consts=set(), rule_children=None):
         # widget: the current instantiated widget
         # rule: the current rule
         # rootrule: the current root rule (for children of a rule)
@@ -581,8 +594,12 @@ class BuilderBase(object):
                 # apply(), and so, we could use "self.parent".
                 child = cls(__no_builder=True)
                 widget.add_widget(child)
-                self.apply(child)
-                self._apply_rule(child, crule, rootrule)
+                self.apply(child, rule_children=rule_children)
+                self._apply_rule(
+                    child, crule, rootrule, rule_children=rule_children)
+
+                if rule_children is not None:
+                    rule_children.append(child)

         # append the properties and handlers to our final resolution task
         if rule.properties:
diff --git a/kivy/uix/widget.py b/kivy/uix/widget.py
index 5bf822ff7..4279ae0d5 100644
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -316,7 +316,7 @@ class Widget(WidgetBase):
     '''

     __metaclass__ = WidgetMetaclass
-    __events__ = ('on_touch_down', 'on_touch_move', 'on_touch_up')
+    __events__ = ('on_touch_down', 'on_touch_move', 'on_touch_up', 'on_kv_post')
     _proxy_ref = None

     def __init__(self, **kwargs):
@@ -349,7 +349,14 @@ class Widget(WidgetBase):

         # Apply all the styles.
         if not no_builder:
-            Builder.apply(self, ignored_consts=self._kwargs_applied_init)
+            rule_children = []
+            Builder.apply(
+                self, ignored_consts=self._kwargs_applied_init,
+                rule_children=rule_children)
+
+            for widget in rule_children:
+                widget.dispatch('on_kv_post', self)
+            self.dispatch('on_kv_post', self)

         # Bind all the events.
         if on_args:
@@ -492,6 +499,9 @@ class Widget(WidgetBase):
             if child.dispatch('on_touch_up', touch):
                 return True

+    def on_kv_post(self, root_widget):
+        pass
+
     #
     # Tree management
     #

```

gottadiveintopython:
> what is the benefit in delaying firing until they are done?

I wasn't thinking much about the benefit. I just thought option 3 is easy to understand/read, but maybe not.

>  I'd be okay with option 3 if that's decided on

I'm not sticking to option 3, and I can't make up my mind which one to choose. In fact, I'm starting to feel option 1 is the best. Let me explain why.

The only weakness of option 1 compared to others that I can come up with is:

```yaml
BoxLayout:
    on_kv_post: do_something_A()  # A
    Label:
        id: label
        on_kv_post: do_something_B()  # B
```

B won't be excuted because B will be bound after `on_kv_post` is fired on `label`. But instead, we can write like this:

```yaml
BoxLayout:
    on_kv_post:
        do_something_A()  # A
        do_something_B()  # B
    Label:
        id: label
```

If the former code always can be converted into the latter form, then the weakness is not an actual weakness I think.

>  if someone wants to do something like this they should move on_kv_post: print(self.parent.ids.my_name) into the Label or the latter MyWidget rule because the current example should be considered buggy and not a common usage. 

I think the same is true of comparision between option1 and the others. For example, the code you posted:

```yaml
<MyWidget@Widget>:
    Label:
        id: a
        on_kv_post: print(root.ids.b.text)  # A
    Label:
        id: b
```

`A` should be moved into top-level, I mean:

```yaml
<MyWidget@Widget>:
    on_kv_post: print(b.text)  # A
    Label:
        id: a
    Label:
        id: b
```

because `A` is accessing out side of `a`. As a result, the code becomes valid under option 1.

> Here's a diff that implements option 3 without flags

Looks great :) I'll test it out later

matham:
I think the comparison between 2 and 3 vs 1 and the others is not at all similar. I didn't make the point that code written in the form of 3 can be converted to 2, but rather it is buggy to write code in the form of 3. That's because a parent class code is referring to objects from an inherited class and this should be rare enough that we probably don't have to worry about it. That's why I tried to list the difference in rules (benefits) between 2 and 3 to see where the difference between them appears.

Regarding option 1, yes we can move the event code into the root rule, but why should we? There's nothing buggy about code that would make use of the event in a instantiated widget in a rule. And in particular from python where this event would be most useful there's no concept of moving this to the root rule. And accessing `ids` and other post kv processing is common from any of the widgets in a rule, not just the root rule.

Consider 

```py
class MyBoxLayout(BoxLayout):
    def on_kv_post(self, widget):
        self.add_widget(Label(text=self.parent.ids.other_widget.text))

SomeWidget:
    MyBoxLayout
    Label:
        id: other_widget
        text: 'apples and fruits'
```

Under 1 this would fail, but under 2 this will work. 

This kind of example is the whole point behind why I wanted to add this, to make sure it works as I listed in the previous post, as well stuff that currently happens in `on_parent`. Moving to the root rule is not an option in many such cases because when you're working from python there's no access to the root rule. 

If all the initialization has to happen in the root rule, then you may as well schedule it for the next clock like current kv, because otherwise, all the initialization that happens in individual widgets, like in `MyBoxLayout` will suddenly have to be moved to `SomeWidget`. That is not ok as that will break current code organization because all of sudden you'll need to re-organize your python code to initialize things in a different class (`SomeWidget`) that is unrelated to `MyBoxLayout`. Furthermore, all initialization relevant for the children of `SomeWidget` (like `MyBoxLayout`) will be done in `SomeWidget`, which is the root of this rule, but the initialization for `SomeWidget` itself, will have to be moved to its parent rule? 

I don't think this would be at all obvious, I certainly would not remember these implementation details 5 months from now if I want to use it, and that option 1 would add very little benefit for the examples I showed here and earlier compared to the current state. In fact for most of them you'd be better off continuing to schedule post initialization on the next clock cycle. To me `on_kv_post` represent a place where you know things are done and you don't have to worry about what you access as it's all ready, option 1 does not provide that guarantee as you still need to keep track of what is done, even within a single rule and only reference widgets that are done.

gottadiveintopython:
> I think the comparison between 2 and 3 vs 1 and the others is not at all similar.

Sorry, I can't explain well because of my English skill but I'll try to explain from another point of view.

The reason I'm starting to feel option1 is the best is, option1 is enough to cover most situations if every event handler does a proper job. For example, the code in your last comment:

```python3
class MyBoxLayout(BoxLayout):
    def on_kv_post(self, widget):
        self.add_widget(Label(text=self.parent.ids.other_widget.text))  # A

somewidget = Builder.load_string('''
SomeWidget:                           # B
    MyBoxLayout:                      # B
    Label:                            # B
        id: other_widget              # B
        text: 'apples and fruits'     # B
''')
```

This is a bad example regardless of which option the code runs under, because you put `A` into wrong place, the default handler. Default handler(`on_kv_post`) is called on every creation of widgets, that means it has to be designed for every way of creation. So the following code must be allowed:

```python
myboxlayout = MyBoxLayout()
somewidget2 = Builder.load_string('''
SomeWidget:
    MyBoxLayout:
''')
```

But it's not allowed, because `A` is depending on rule`B`. The right place where `A` should be is that obviously inside rule`B`, which is:

```yaml
SomeWidget:
    on_kv_post:  # a
    MyBoxLayout:
        on_kv_post:  # b
    Label:
        id: other_widget
        text: 'apples and fruits'
        on_kv_post:  # c
```

- `a`,`b` and `c` if the code runs under option 2 or 3
- `a` if the code runs under option 1

So I can say that option1 can handle your example. I'm trying to find the case that option 1 can't handle, but couldn't find it so far, and if option 1 can handle every case, then it's the best because it's the simplest and the fastest.

matham:
>This is a bad example regardless of which option the code runs under, because you put A into wrong place

I understood what you were trying to say but I very much disagree with this. You're thinking that `MyBoxLayout` needs to be able to handle any positioning. In practical usage, these classes are often written for specific usages so you don't have to worry about `MyBoxLayout` appearing in other circumstances beyond your control and so this pattern is perfectly ok to use.

I have many classes which are written with assumptions about other widgets that are available. And `ids` are not the only thing you need to reference.

A typical realistic example may be:

```yaml
<MyGraph>:
    DeviceWidget:
        device = root.device
        IODevice:
            dev: self.parent.device
            on_kv_post: self.init_device(root)
        BoxLayout:
            Axis
                id: y
            Plot
                id: plot
        Axis:
            id: x

```


```py
class MyGraph(GridLayout):
    device = ObjectProperty(None)
    
    def export_graph(...):
        ...

class DeviceWidget(BoxLayout):
    device = ObjectProperty(None)
    
    def do_something(...):
        ...

class IODevice(Widget):
    dev = ObjectProperty(None)

    def init_device(self, graph):
        graph.ids.x = 'Time'
        graph.ids.y = self.dev.name
        self.dev.set_plot(graph.ids.plot)

class Controller(BoxLayout):
    devices = []
    
    def __init__(self, **kwargs):
        super(Controller, self).__init__(**kwargs)
        for device in self.devices:
            self.add_widget(MyGraph(device=device))
```

I can see at least 2 classes of bugs with this example under option 1 Can you find them? 

Obviously `on_kv_post: self.init_device(root)` can be moved elsewhere, but that is not an expectation we should be having and we not should not be expecting people to be making these subtle decisions where to put code that depends on being able to remember exact implementation details. The code above is perfectly fine, although maybe not the cleanest, but I could easily see myself writing such code - I definitely have (without the bugs of course). The point of option 2 and 3 was to solve these kinds of initialization issues so you don't have to worry what has been executed before what, but rather in `on_kv_post` you could assume that everything is done and you could use anything from the kv rule and assume it has been initialized.

Specifically to what you said that that in the example above the `on_kv_post` should be moved to the root. As I said that is not something I see as a realistic expectation, especially if the `on_kv_post: self.init_device(root)` code was moved into python because all of a sudden you have to move initialization to a different class than what you want (i.e. from `IODevice` to `MyGraph`). But, I also think moving that kv line to `MyGraph` would be wrong anyway.

Consider how I would write code under option 2:

```yaml
<MyGraph>:
    on_kv_post: self.check_weather()
    DeviceWidget:
        IODevice:
            on_kv_post: self.check_devices()
        BoxLayout:
            Axis
                on_kv_post: self.set_something()
            Plot
                on_kv_post: self.set_something_else()
        Axis:
            on_kv_post: self.call_home()
```

To me this looks perfectly fine. But under option 1 this (probably) will not work and instead, to be sure, the code will have to look like this:

```yaml
<MyGraph>:
    on_kv_post:
        self.check_weather()
        dev.check_devices()
        y.self.set_something()
        plot.set_something_else()
        x.call_home()
    DeviceWidget:
        IODevice:
            id: dev
        BoxLayout:
            Axis
                id: y
            Plot
                id: plot
        Axis:
            id: x
```

I don't think this is okay because this violates the principle that stuff should happen locally and not far from where it's used and frankly is a unintuitive thing to do to pile up a bunch of stuff in the root rule.

As I said a few times now, to me the `on_kv_post` should **just** work without having to remember where it is or is not legal and option 1 does not do that at all IMO. In fact I cannot see myself using option 1 in my code. But, if this is what you want to do in this PR and others agree, that is fine, although I still would not be for it.

As to the bugs in the example code above:

* `ids.x`, `ids.y`, `ids.plot` are invalid attributes at this point so it would crash.
* `self.dev` is still None, so it would crash.

Fak3:
Perhaps we can have both options? 
`<MyWidget>` triggers `on_kv_post` (own rules applied) and
`MyWidget: ...` triggers `on_kv_mount` (id is assigned)

matham:
>You can see that the code working fine under option 1.

Yes, after being re-written. The point was people would normally write the code the way I wrote it and it should still work without having to be re-written. But, I think you understand what I meant so I'll leave it.

@tito
I'm fine with two events (or actually three) - `on_kv_pre`, `on_kv_apply`, and `on_kv_post`. `on_kv_apply` would implement option 1 and `on_kv_post` either option 2 or 3. Does the names sound ok? Any prefs between option 2 and 3?

gottadiveintopython:
If there are two seperate events, I feel like option3 is better.

matham:
I can see arguments for both 2 and 3, and lean a little towards 2, but I'll be fine with either.

I guess @tito will decide.

gottadiveintopython:
I'm not arguing against the idea decided yesterday. I just want to explain why I think option 1 is the best (or even with that idea) because my previous explanation may doesn't make sense. 

# CONS

I understand that the necessity of transforming the following code:

```yaml
Widget:
    on_kv_post: do_something_a()
    Widget:
        on_kv_post: do_something_b()
```

into the following form:

```yaml
Widget:
    on_kv_post:
        do_something_a()
        do_something_b()
    Widget:
```

is an option1's weakness, and for me, that's the only weakness.

# CONS(not for me)

The following code:

```python
# Similar to one of matham's examples
from kivy.lang import Builder


class WidgetA(Widget):
    def on_kv_post(self):
        pass


class WidgetB(Widget):
    def on_kv_post(self):
        self.parent.ids.label.text = 'Hello'  # A  (KeyError if under option 1)


Builder.load_string('''
<WidgetA>:  # B
    on_kv_post:  # C
    WidgetB:
        on_kv_post:  # D
        Label:
            id: label
            on_kv_post:  # E
''')

runTouchApp(WidgetA())
```

works fine under option 2 & 3, but not 1. Does that mean option 1 is inconvenient compared to the others? I think not, because putting `A` in `WidgetB.on_kv_post()` is probably a bad thing regardless of which option. The fact that `A` access to `self.parent.ids.label` means `A` always has to be excuted after `B` is applied. And that's actually not guaranteed because the following code doesn't work:

```python3
widget = WidgetB()
```

The proper place where `A` should be has to be:

- `WidgetA.on_kv_post()`, `C`, `D` and `E` if under option 2 or 3
- `WidgetA.on_kv_post()` and `C` if under option 1

because those places are guaranteed to be excuted after `B` is applied. And as a result of moving 'A' to proper place, the code get to work under option 1. So I can say "Option 1 works fine only if `A` is put in the proper place" and "Option 2 & 3 work fine without even doing that". Does that mean option 1 is inconvenient compared to the others? No for me. That means option 1 can let the user realize the fact that `A` was put in the wrong place, earlier.

>  you don't have to worry about `MyBoxLayout` appearing in other circumstances beyond your control and so this pattern is perfectly ok to use

As long as the user put a code in the proper place, the widget always should be able to handle any positioning, so the user doesn't worry about that. But the term 'proper' being used here is subjective. Most pepole probably doesn't feel it as 'proper', and doesn't want to worry about where to put a code. Therefore, this section is considered as CONS but not for me.

# PROS

Needless to say, the implementation is simpler than the others.

# Summary

As long as the user put a code in 'proper' place, option 1 would shine. Otherwise the other one would be better, I think. Sorry for the long sentence :)

gottadiveintopython:
Um sorry for being noisy.
I'll leave this PR to others.

matham:
I'm not sure what you're apologizing for?

kivy/lang/builder.py
```diff
@@ -396,8 +396,17 @@ def load_string(self, string, **kwargs):
                 self.files.append(fn)
 
             if parser.root:
-                widget = Factory.get(parser.root.name)()
-                self._apply_rule(widget, parser.root, parser.root)
+                widget = Factory.get(parser.root.name)(__no_builder=True)
+                rule_children = []
+                append_rule_child = rule_children.append
```
matham: The builder functions should accept a list, like in my example code, to which it appends the children and not a function that is called.

The former is a very clean pattern, as it's just appending to a list, and it may or may not be useful for other things besides our desired usage. The latter, is much more general as it accepts any function so it can work for other needs, but is also therefore more complex. Right now there's no reason to think accepting any function will be ever needed so there's no reason to have this extra complexity. So please change it accepting a list rather than an append function.

kivy/uix/widget.py
```diff
@@ -343,9 +348,28 @@ def __init__(self, **kwargs):
...
+        # Because we currently don't support 'on_kv_pre' in kv, we might be
+        # able to cheese this part like this:
+        #
+        # self.on_kv_pre()
+        #
+        # unless the user does this:
+        #
+        # def __init__(self, **kwargs):
+        #     super().__init__(**kwargs)
+        #     self.bind(on_kv_pre=...)
```
matham: This comment is not really needed because it doesn't add info on the current code, but rather is targeted for us reading the PR so please remove it.

kivy/uix/widget.py
```diff
@@ -316,7 +320,8 @@ class Widget(WidgetBase):
     '''
 
     __metaclass__ = WidgetMetaclass
-    __events__ = ('on_touch_down', 'on_touch_move', 'on_touch_up')
+    __events__ = ('on_touch_down', 'on_touch_move', 'on_touch_up',
+                  'on_kv_pre', 'on_kv_post')
```
matham: Are you planning to also implement `on_kv_applied` like we mentioned in the PR? I.e. option 1.

matham:
If you don't want to work on it anymore, that's okay. You have done a lot of work on it so far. If that's the case, one of us can finish it off.

gottadiveintopython:
@matham 

> Are you planning to also implement `on_kv_applied` like we mentioned in the PR? I.e. option 1.

I'll do it if it's decided. Is the name `on_kv_applied` actually decided?

> If you don't want to work on it anymore, that's okay. You have done a lot of work on it so far. If that's the case, one of us can finish it off.

I want to leave the specification of this PR to you and @tito （and the others) at least.
I'll write a code if needed.

Also, can I make all commits in this PR into 2 or 3 commits? If you don't mind.

matham:
Perhaps @tito is not actively monitoring the thread or he's busy, but he seemed to agree with doing both events.

I'd say that if you agree that the name `on_kv_applied` seems ok to you - and I think it's ok, then you should just go ahead and code it up. Then you could make a new branch, squash the commits into a couple of commits and open a new PR referencing this PR.

This will give it a clean start and make it easier for people to review. We can then give it a week or so waiting for feedback before we merge, if none are forthcoming.

gottadiveintopython:
ah ok, you mean `on_kv_applied` should be implemented by another branch and another PR. And the commits in this PR shouldn't be squashed.

Also I wanna ask, which option does `on_kv_post` implement? option 2 or 3? I'll leave it to you.

matham:
>(EDIT: No, you mean open a new PR after implement on_kv_applied on this PR. Am I correct?)

Yes

>Also I wanna ask, which option does on_kv_post implement? option 2 or 3? I'll leave it to you.

You already implemented option 3 in the PR, so we can stay with that.

kivy/lang/builder.py
```diff
@@ -581,8 +594,13 @@ def _apply_rule(self, widget, rule, rootrule, template_ctx=None,
                 # apply(), and so, we could use "self.parent".
                 child = cls(__no_builder=True)
                 widget.add_widget(child)
-                self.apply(child)
-                self._apply_rule(child, crule, rootrule)
+                self.apply(child, rule_children=rule_children)
+                child.dispatch('on_kv_applied')
+                self._apply_rule(
+                    child, crule, rootrule, rule_children=rule_children)
```
matham: Do you think it should perhaps be dispatched after `_apply_rule ` rather than before?

gottadiveintopython: I might misunderstand what the 'own rule' is. In the code below:

```python3
Builder.load_string('''
<WidgetA@Widget>:  # A
    ...            # A
<WidgetB@Widget>:
    WidgetA:       # B
        ...        # B
''')
root = Factory.WidgetB()
```

`WidgetA`'s own rule is `A` only, am I correct? If so, the current code should be fine. Because the following part applies `B`.

```python3
                self._apply_rule(
                    child, crule, rootrule, rule_children=rule_children)
```

Regardless of that, I probably should leave this PR to you. Because the PR really affects to the usability of Kivy, and misunderstanding your words is critical. So I want to leave this PR to you, if you don't mind.

matham: I only said `own rule` because that's what @tito said. But I doubt that he meant it like you understood it - English is not his first language, because to me it'll be confusing if it's fired before all the elements being applied to the widget is triggered and that includes `B` for me. 


I'm sorry that this took so long and that you had to re-write it so much. I think it was really helpful to arrive at a much better PR than either for us initially proposed. I didn't want to initially just take over the PR and do it myself as a competing PR, because I was hoping to illustrate, through comments, the thought and process that goes into API design so that future contributors can learn from it. So I'm really thankful for you patience.


Can I ask you to do one more thing? Can you please create a new branch locally, squash all the commits into it and open a new PR. That way you'll get the credit for you work on this because it'll be your commit, and then I can just polish it up from there by working on the PR. You can do it with something like this:

```
git checkout master
git checkout -b kv_events
git merge --squash post_kv
git commit -m "..."
git push origin kv_events
```

gottadiveintopython:
I opend a new PR. Really thank you for helping :) @matham @tito

PR #6232, opened 2019-04-04
Respect the alpha value when setting rgb.

matham:
Currently, the alpha channel gets overwritten by 1. when `rgb` is set. This makes it respect the current alpha value.

I also added the initialization so that `self.a` will not crash if only rgb was ever set.


tito:
@matham, you broke Color with self referencing self.a i guess, check the tests

matham:
Fixed!

PR #6237, opened 2019-04-05
Changes to ensure that documentation produced is for the current repo…

DaytonaJohn:
…sitory

Without this change, "make html" will produce documentation of a system
installed kivy rather than of the current repository. These changes modify
the sys.path in sources/conf.py to position the current repository at the
start of the search path.

### DIFF
--- a/doc/sources/conf.py
+++ b/doc/sources/conf.py
@@ -23,6 +23,11 @@ sys.path.insert(0, os.path.abspath('sphinxext'))
 base_dir = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, os.path.dirname(base_dir))
 
+# ensure that the modules in this kivy clone get imported rather than
+# a system installed kivy
+sys.path.insert(0, os.path.abspath(os.path.join(base_dir, '..')))
+sys.path.insert(0, os.path.abspath(os.path.join(base_dir, '..', '..')))
+
 # General configuration
 # ---------------------
 
@@ -61,7 +66,7 @@ copyright = '2010, The Kivy Authors'
 #
 os.environ['KIVY_DOC_INCLUDE'] = '1'
 import kivy
-print(kivy.__file__)
+print('producing documentation for', os.path.dirname(kivy.__file__))
 
 version = kivy.__version__
 release = kivy.__version__


tshirtman:
This change shouldn't be needed, if you use a virtualenv with kivy installed in it in develop mode, either with
`python setup.py develop`
or
`pip install --editable .`
you'll be able to edit and rebuild your the doc for the kivy version you use in this virtualenv.

matham:
Agree. Messing with path seems risky and shouldn't be done unless it truly was needed.

DaytonaJohn:
This is not "messing" with path. This is "truly needed" to get the path for the kivy documentation build correct. I have had extensive experience with software packages and this is the only one I have ever seen where "make <product>" does not produce the desired product from the downloaded source! This simple fix corrects that error. If you insist on continuing to do it wrong, please at least document that "make <any_doc_product>" will not work except under a specific list of requirements.

matham:
Can you give an example standard environment, where this is needed and docs building doesn't work?

DaytonaJohn:
Do a "git clone https://github.com/kivy/kivy". Make any change to a doc string in any class in the downloaded kivy. Go to the `doc` directory of the downloaded kivy and run `Make html`. Then view the resulting html documentation and see that your change will not appear. That is because the path in `doc/sources/conf.py` is incorrect.

matham:
I have done it many times and it always worked for me in the past. 

As @tshirtman was alluding to, it won't work if the cloned repo was not properly installed to your current python version. Hence after cloning you should do `pip install -e .`, which will fix the issue. If you don't do that after cloning, your environment is in a non-standard install situation, and we shouldn't add path manipulation if it can be resolved by properly installing kivy as explained.

DaytonaJohn:
I could not disagree more. You should not have to install a cloned repo 
in order to build its documentation. And the only path change is in the 
configuration for building that documentation.

tshirtman:
Using develop / editable install to work in a project is a very common python pattern, supported by setup.py and pip, it's the proper way to do it, sure, there are other, hacky ways to do it, but this is the proper one, and i'd rather have to educate users than encouraging a bad workflow that will break in a lot of cases.

matham:
If you truly don't want to install the cloned kivy, you should at least use `PYTHONPATH` to make that kivy available. But yeah, modifying the path is not the answer as that will break other, more typical installs.

DaytonaJohn:
What you are describing to me are hacks to get around an improper doc build system. If you can demonstrate just one case where this change will break anything,  I will shut up.

matham:
Ok, let me take back and change my words from 

>But yeah, modifying the path is not the answer as that will break other, more typical installs.

to 

>But yeah, modifying the path is not the answer as that may break other installs.

One obvious way it may break is if the docs directory is moved so this will be adding the wrong version of kivy, or perhaps some random folder to the path. I'll admit that in typical installs that I have used before this is not the case because I have certainly never moved the docs directory in kivy. But it is a possibility that someone else may do it for whatever reason and it'll break. But sure, this is not common either.

A more realistic issue is that moving kivy to the top of the PATH would change the order of imports (i.e. kivy is imported before anything else). I have definitely in the past ordered my imports in response to other bugs. E.g. when I had to import ffpyplayer, (or some other library with binary conflicts with kivy) before kivy, otherwise they both broke. But I'll admit that this is also not something usual.

At the end of day there are potential pitfalls when modifying the path, many of which I have personally experienced, and that should not be done without careful consideration. It definitely makes me nervous to change the path like this, and it seems @tshirtman also. The example you gave does not seem to justify IMO changing the path like this, especially when we gave you at least simple two solutions. 

`PYTHONPATH` is meant to be used for this kind of situation so unless there's a reason you cannot use it, I cannot see us merging this.

DaytonaJohn:
If the `docs` directory is moved, then even `make pdf` in the top level `Makefile` will fail.

The argument about the path is not valid. The path change I am suggesting is a temporary change for the documentation build process only. It does not affect any path for any other kivy or python execution. Modifying PYTHONPATH, however, will affect your every python execution until you revert those changes (definitely a bad hack).

Issue #6210, opened 2019-03-17
video and videoplayer stop one and not to play again


zsheasy:
### Versions

* Python:3.7.1
* OS:ubuntu 18.04
* Kivy:1.11
* Kivy installation method:

### Description

I try to play a m3u8 video by video or videoplayer.it is good on my ubuntu.but,the first time is good on my phone(android6.0),The second time app exit.

### Code and Logs

```python
class VideoScreen(Screen):

    def __init__(self, **kw):
        super(VideoScreen, self).__init__(**kw)
        self.video = None

    def on_enter(self, *args):
        app = App.get_running_app()
        if app.runtime_data.video_data.source: # this is m3u8 url
            self.video = Video(source=app.runtime_data.video_data.source, state='play')
            # self.ids.player.source = app.runtime_data.video_data.source
            # self.ids.player.state = 'play'
            self.ids.sl.add_widget(self.video)
            print('on_enter...sdf.')
            print('on_enter..state', self.video.state)

    def close_video(self):
        self.video.unload()
        self.video = None
        self.ids.sl.clear_widgets()
        # self.ids.player.state = 'stop'
        # self.ids.player.source = ''
        print('close..state')
        Clock.schedule_once(self.to_screen, 0.5)

    def to_screen(self, *args):
        app = App.get_running_app()
        # app.root.current = 'detail_screen'
        app.root.current = 'main_screen'


```
### adb and Logs
```python
03-17 10:31:22.735  2953  3084 I python  : Initializing Python for Android
03-17 10:31:22.735  2953  3084 I python  : Setting additional env vars from p4a_env_vars.txt
03-17 10:31:22.735  2953  3084 I python  : Changing directory to the one provided by ANDROID_ARGUMENT
03-17 10:31:22.735  2953  3084 I python  : /data/user/0/org.zhangsh.w51mt/files/app
03-17 10:31:22.735  2953  3084 I python  : Preparing to initialize python
03-17 10:31:22.735  2953  3084 I python  : _python_bundle dir exists
03-17 10:31:22.735  2953  3084 I python  : calculated paths to be...
03-17 10:31:22.735  2953  3084 I python  : /data/user/0/org.zhangsh.w51mt/files/app/_python_bundle/stdlib.zip:/data/user/0/org.zhangsh.w51mt/files/app/_python_bundle/modules
03-17 10:31:22.736  2953  3084 I python  : set wchar paths...
03-17 10:31:22.810  2953  3084 I python  : Initialized python
[...]

PR #6246, opened 2019-04-10
Disabled set_volume() in core.video.ffpyplayer play() function. Fix for #6210

misl6:
As mentioned in issue #6210 , on Android the video and videoplayer works only once.

Actual implementation on the ffpyplayer provider in core.video.ffpyplayer sets the volume just after the creation of the MediaPlayer.

After some investigation, seems that, the second usage of the Video/VideoPlayer w/ ffpyplayer takes more time to instantiate the AudioTrack that it looks like to be async.

In order to get my app running correctly I've commented out (as proposed in this PR) the set_volume call in the play() function.

I think that is not to be considered as the final fix ( we need to investigate further on the ffpyplayer implementation ), but should be merged ASAP, in order to temporarily solve this issue for all the affected users.

### DIFF
--- a/kivy/core/video/video_ffpyplayer.py
+++ b/kivy/core/video/video_ffpyplayer.py
@@ -346,7 +346,9 @@ class VideoFFPy(VideoBase):
                 self._filename, callback=self._player_callback,
                 thread_lib='SDL',
                 loglevel='info', ff_opts=ff_opts)
-        self._ffplayer.set_volume(self._volume)
+
+        # Disabled as an attempt to fix kivy issue #6210
+        # self._ffplayer.set_volume(self._volume)
 
         self._thread = Thread(target=self._next_frame_run, name='Next frame')
         self._thread.daemon = True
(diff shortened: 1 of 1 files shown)


matham:
I'm not sure how disabling setting the volume solves the crash? I don't see any mention of that line in the error message. In any case, you can pass the `volume` parameter to `ff_opts` instead of calling `set_volume`, rather than just commenting it out.

See the option here: https://github.com/matham/ffpyplayer/blob/master/ffpyplayer/player/player.pyx#L245

tito:
Yep, don't merge this, because even if it solve a possible crash, it also break the wanted volume at the start.

Danapessach:
I'm not sure the volume matters when the video doesn't work at all... 
This is why it was suggested as a temporary solution.

matham:
It does work generally, just not when playing a second time, or a perhaps when playing a particular file. This will break it for everyone even for cases that do currently work because the volume will stop working.

Danapessach:
The problem happened for me with all files - different videos, different formats...

For me the volume worked OK after implementing the fix with the original volume of the video.

For you, when you tried this fix the volume didn't work at all?

misl6:
@tito @matham I've added `volume` in`ff_opts`.
Consider that as a temp fix.

Issue #5617, opened 2018-02-12
Nested scrollview fix attempt


glins97:
### Description

When nesting scrollviews, the inner scrollview always "jumps" to scroll position, rather than scroll smoothly. An easy way to see this behavior is to look at [kived example code](https://gist.github.com/kived/2aa1b6814979340598a7).
Nesting scrollviews has always been an issue, even thought we are able to use scrollview + carousel combo. However, It's a functionality well present in Apps such as Netflix and Instagram. It's expected to work in way that was not possible with current ScrollView code.
 
This is a silly attempt to fix the issue (which is why I haven't submited a pull request, besides never having done it before). I do have it working, as far as I could test it. I also do not know exactly how kivy's touch grabbing works, and so I decided to use the issue tracker, rather than a straight pull request. If this is wrong, I'll be happy to move the discussion to another area.

New scrollview code can be found [here](https://github.com/glins97/ScrollView) with a test App. Below I'll post only example code, since ScrollView code is a thousand lines long. Hope this is the right thing to do.

Thanks, guys. Let's make this work.

Example code:
`# -*- coding: utf-8 -*-
import kivy
from kivy.app import App
from scrollview import ScrollView
from kivy.uix.boxlayout import BoxLayout
from kivy.uix.gridlayout import GridLayout
from kivy.uix.carousel import Carousel
from kivy.uix.button import Button
from kivy.metrics import cm
        
class NotificationDemoApp(App):
    def build(self):
        main_scroll = ScrollView(do_scroll_x=False)
        main_grid = GridLayout(cols=1, spacing=100, padding=[0, 50, 0, 50], size_hint_y=None)
        main_grid.bind(minimum_height=main_grid.setter('height'))
        main_scroll.add_widget(main_grid)
        scroll_count = 4
        for _ in range(scroll_count):
            scroll = ScrollView(size_hint_y=None, height=300, do_scroll_y=False)
            grid = GridLayout(rows=1, spacing=1, size_hint=(None, 1))
            grid.bind(minimum_width=grid.setter('width'))
            scroll.add_widget(grid)
            for r in range(0,50):
                bt = Button(text='Button '+str(r), size_hint_x=None, width=cm(2))
                bt.bind(on_press=self.print_btn)
                grid.add_widget(bt)
            main_grid.add_widget(scroll)
        return main_scroll

    def print_btn(self, instance):
        print 'Pressed', instance.text

if __name__ == '__main__':
    NotificationDemoApp().run()`,

```

[...]

PR #6252, opened 2019-04-14
Attempt to fix nested scrollviews

misl6:
I think this could be an easy fix to get nested scrollviews scroll flawless.
I tested it on a real-world app.

**Code to reproduce:**
```
                ScrollView:
                    size_hint: 1, 1
                    effect_cls: 'ScrollEffect'
                    do_scroll_x: False
                    scroll_type: ['content']
                    GridLayout:
                        cols: 2
                        size_hint_y: None
                        height: self.minimum_height
                        ........
                        ScrollView:
                            effect_cls: 'ScrollEffect'
                            size_hint: .706, 1
                            do_scroll_y: False
                            scroll_timeout: 55
                            scroll_type: ['content']
                            GridLayout:
                                size_hint: None, 1
                                padding: 0
                                spacing: 0
                                cols: 1
                                width: self.minimum_width
                                orientation: 'vertical'
                              .......
```

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -764,7 +764,7 @@ class ScrollView(StencilView):
         if touch.grab_current is not self:
             return True
 
-        if touch.ud.get(self._get_uid()) is None:
+        if not any(key.startswith('sv.') for key in touch.ud):
             # don't pass on touch to children if outside the sv
             if self.collide_point(*touch.pos):
                 return super(ScrollView, self).on_touch_move(touch)


matham:
The fix is great. I tested it and it worked in multiple situations without introducing new issues. 

Bug analysis
=============

This kind of bug has been plaguing `ScrollView` for a very very long time, so following is an **analysis** of the bug relating to the code in `on_touch_move`: https://github.com/kivy/kivy/blob/master/kivy/uix/scrollview.py#L754.

Given a parent and child who are both `ScrollView`, if you drag the scroll bar of the child, in on_touch_down, the parent passes the touch to the child who uses it to start the drag - as appropriate.

Then, upon a move, the parent checks `if self._touch is not touch`, which is `False` so it next checks `if touch.grab_current is not self` which is `True`. Because it's not `self` and scrollview prefers to only handle stuff during a `grab` callback, it ignores the touch and returns and does not try the child. This is the bad part IMO.

Next, the system will start processing the `grab` callbacks so it calls the parent again. This time for the parent `if self._touch is not touch` and `if touch.grab_current is not self` are `False` because it previously got the touch, so then it checks `if touch.ud.get(self._get_uid()) is None` which is `True`, that tells it that the touch is not used by the parent so it passes it on to the children by calling `super(ScrollView, self).on_touch_move(touch)`.

Finally, now the child gets the `move` event. For the child, `if self._touch is not touch` is `False`, but `if touch.grab_current is not self` is True. Now, because scrollviews don't handle touches except only during the `grab` callback, the child returns `True`. But, because of this, this is the end of the touch and the child never gets called with the touch where `if touch.grab_current is not self` is `False`.


Current fix
=========

The implemented fix in the PR, replaces `if touch.ud.get(self._get_uid()) is None` with ``if not any(key.startswith('sv.') for key in touch.ud)). This means that if there's a child who subscribed to the touch, the parent immediately passes it on to the children by calling `self.dispatch_children('on_scroll_move', touch)`. This is not a real fix, as it's more of a bandaid - e.g. you can see a potential issue if a sibling scrollview subscribed to the touch, this may missfire, but I couldn't cause this issue.

Ultimate cause
==========

For me, the ultimate cause, is what I have mentioned previously elsewhere, is that we frequently use the following pattern everywhere, which I think should not be allowed, except if really needed:

```py
if touch.grab_current is not self:
    return True
```

IMO, we should never delay handling an event until `touch.grab_current is self`, because if we delay handling the event, then the `grab` callbacks are handled outside and after the whole widget tree has been processed. To me that is asking for trouble and leads to these kid of bugs because it's hard to account for everything and in the past led to similar bugs. 

Events should **always** be handled during the dispatching of the tree, and `grab` callbacks should only be used when we get a `move` or `up` event outside the widget.

matham:
I discovered an issue with https://github.com/kivy/kivy/issues/5617. Running that example code and dragging the buttons will periodically cause a crash due to this PR. But, this PR did improve that issue, although it's not fully fixed (buttons are still sometimes pressed and vertical scrolling doesn't work).

matham:
So I fixed the crash I mentioned by changing it to `if not any(isinstance(key, str) and key.startswith('sv.') for key in touch.ud):`. Along the way  I also tried fixing #5617. However, it seems impossible to fix because the code will need to be re-written. Specifically, the problem is not just the `grab`, but rather that the status of the touch is a state machine, but with like 10 variables across which the state is distributed (i.e. whether `_touch` is `None`, the values in `ud` etc.), which is impossible to track correctly.

That's why we leak touches. Someone will have to rewrite this making sure all possible states are accounted for. Surprisingly, the example code in #5617 is an amazingly good test to find the issues in the scrollview.

misl6:
> So I fixed the crash I mentioned by changing it to `if not any(isinstance(key, str) and key.startswith('sv.') for key in touch.ud):`. Along the way I also tried fixing #5617. However, it seems impossible to fix because the code will need to be re-written. Specifically, the problem is not just the `grab`, but rather that the status of the touch is a state machine, but with like 10 variables across which the state is distributed (i.e. whether `_touch` is `None`, the values in `ud` etc.), which is impossible to track correctly.
> 
> That's why we leak touches. Someone will have to rewrite this making sure all possible states are accounted for. Surprisingly, the example code in #5617 is an amazingly good test to find the issues in the scrollview.

@matham 
I've also encountered a similar issue during runtime in my app.
As you proposed could be solved it via PR #6255

And yes, I also think that the code have to be re-writtten.

PR #6255, opened 2019-04-16
Fix " object has no attribute 'startswith' "

misl6:
As @matham proposed here: https://github.com/kivy/kivy/pull/6252#issuecomment-483562572 my previous PR added a potentially crash condition.

This PR will fix that problem.

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -764,7 +764,8 @@ class ScrollView(StencilView):
         if touch.grab_current is not self:
             return True
 
-        if not any(key.startswith('sv.') for key in touch.ud):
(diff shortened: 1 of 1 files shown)


kivy/uix/scrollview.py
```diff
@@ -764,7 +764,8 @@ def on_touch_move(self, touch):
         if touch.grab_current is not self:
             return True
 
-        if not any(key.startswith('sv.') for key in touch.ud):
+        if not any(isinstance(key, str) and key.startswith('sv.')
+                   for key in touch.ud):
```
matham: In python, `not` binds before `and`. So you need parentheses around the statement `if not (x and y)`.

PR #6256, opened 2019-04-16
Set cookie header workaround

cruor99:
This is a small workaround for the Set-Cookie header issue, should a resource provide multiple headers of Set-Cookie.

It feels a little bit hacky, but I'm not entirely sure what the best way to improve this specific use case would be.

I have done some testing with this particular implementation, and there doesn't appear to be any interference with anything else operating as usual.

The base issue here seems to be that UrlRequest is based on urllib2 - as Requests doesn't seem to have this particular issue (seems to be based on urllib3)

### DIFF
--- a/kivy/network/urlrequest.py
+++ b/kivy/network/urlrequest.py
@@ -460,6 +460,17 @@ class UrlRequest(Thread):
             except IndexError:
                 return
             if resp:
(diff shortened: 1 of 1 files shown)


tito:
Feels hacky but LGTM.

PR #6257, opened 2019-04-16
Give the user the opportunity to initialize widget before/after applying kv rules

gottadiveintopython:
This is  a continuation of #6221.

EDIT by @matham:

Kv initialization events
====================

This PR introduces **three** new events to help users manage the boundary of python and KV code.

This has been something that has come up a lot before. Looking at the survey of how we can improve python, people often asked for ways to be able to interact with kv code from Python.

* Introducing ``ids`` was one such improvement, however,
this led to new issues, [where accessing](https://github.com/kivy/kivy/issues/2479) ``ids`` from Python failed if accessed too soon.
* Another common workaround is the usage of ``on_parent`` as a chance to initialize stuff from kv after most everything else has been initialized. This has issues because ``on_parent`` may be triggered whenever the widget is removed from the parent, and this doesn't help us integrate it with Python.

This PR proposes the following three events to solve this issue permanently. All these events are fired whether there are kv rules to be applied, but if there are kv rules it is fired at a specific time during the kv rule application.

on_kv_pre
-------------

This event is fired for each widget before any kv rules are applied. This event is fired in the instant before any kv rules are potentially applied to the widget during the ``__init__`` of the widget. This event is meant to give the user an opportunity to initialize stuff from Python before anything kv is run for this widget. Typical usage is e.g.::

```python
class MyWidget(Widget):

    def on_kv_pre(self):
        self.variable_used_in_kv = some_value
```

In the following code ``do_something`` is never called because the widget is instantiated before the kv rule could be bound::

```yaml
<SomeWidget>:
    on_kv_pre: do_something
```

I propose that any kv ``on_kv_pre`` rule should raise a error to help users find such silent bugs which are never executed.

on_kv_applied
---------------

This event is fired after the rules of the widget class is applied (excluding any rules associated with its instantiation). The dispatched parameter to the event is the widget in the root rule that instantiated this widget (or ``None``). Specifically, it is fired after the class rules, but before the instance rules in the example below is applied::

```yaml
<SomeWidget>:
    class_rules

<SomeOtherWidget@SomeWidget>:
    more_class_rules

Widget:
    SomeOtherWidget:
        instance_rules
```

The use of this would be to be able to do something within python after the widget rules has been applied such is in this example::

```python
class MyWidget(Widget):

    def on_kv_applied(self, root_widget):
        assert isinstance(root_widget, BoxLayout)

Builder.load_string("""
BoxLayout:
    Widget:
        MyWidget
""")
```

Something to keep in mind, when it's fired ``ids`` would still not be initialized and is also not a good replacement for ``on_parent`` as it happens before kv is fully initialized.

Issues:
********

I forsee some issues with this event though. Because it is triggered before the rules of the instantiated widget is applied I think it'd lead to confusion. Consider the following basic example, similar to above::

```python
class MyWidget(Widget):

    def on_kv_applied(self, root_widget):
        assert self.name == 'Kivy'

Builder.load_string("""
BoxLayout:
    Widget:
        MyWidget:
            name: 'Kivy'
""")
```

The ``assert`` would fail because ``name`` would not have been yet initialized when the event is fired, because it's fired after the class rules, but before instance rules are applied.

An similar example IMO is the following::

```yaml
<MyWidget>:
    on_kv_applied: do_a

<SomWidget>:
    on_kv_applied: do_c
    MyWidget:
        on_kv_applied: do_b
```

In the above example only ``do_b`` is never executed for the reason mentioned above. To me this is a problem because I don't think people will be able to keep track of when it is and isn't legal to use ``on_kv_applied`` from kv and would lead to complaints and silent bugs when the callback is not executed.

I'm not sure of the utility of the event, but @tito was interested in this event so we included it. @tito, can you confirm that you want to keep this event?

A potential fix for the second example that I want to suggest is to either disallow using ``on_kv_applied`` from kv at all, or raise an exception if the event is used in the ``do_b`` case. This however does not solve the issues with the first example. 

on_kv_post
------------

This event is fired after all the kv rules associated with the widget is applied. This is intended to replace initialization from kv in ``on_parent`` and when it's dispatched ``ids`` would be fully initialized. The dispatched parameter in the event is the ``base_widget`` that was instantiated from python and which caused the application of all the rules. 

The use of this would be to do anything post initialization after all te kv rules has been applied. E.g.::

```python
class MyWidget(Widget):

    def on_kv_applied(self, base_widget):
        assert isinstance(base_widget, BoxLayout)

Builder.load_string("""
<SomeWidget@Widget>:
    MyWidget

BoxLayout:
    on_kv_post: assert self.ids.root == self
    SomeWidget
""")
```

The exact parameter it takes is not as important as the time when it is fired. It is fired after **all** the kv rules above has been fully applied.

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -291,18 +291,25 @@ class Widget(WidgetBase):
     '''Widget class. See module documentation for more information.
 
     :Events:
-        `on_touch_down`:
-            Fired when a new touch event occurs
-        `on_touch_move`:
-            Fired when an existing touch moves
-        `on_touch_up`:
-            Fired when an existing touch disappears
-        `on_kv_pre`:
-            Fired before kv rules are applied
-        `on_kv_applied`:
-            Fired after its own kv rules are applied
-        `on_kv_post`:
-            Fired after all kv rules are applied
+        `on_touch_down`: `(touch, )`
+            Fired when a new touch event occurs. `touch` is the touch object.
+        `on_touch_move`: `(touch, )`
+            Fired when an existing touch moves. `touch` is the touch object.
+        `on_touch_up`: `(touch, )`
+            Fired when an existing touch disappears. `touch` is the touch
+            object.
+        `on_kv_pre`: `()`
+            Fired before kv rules are applied. It doesn't pass any parameters.
+        `on_kv_applied`: `(root_widget, )`
+            Fired after all the kv rules of the widget is applied.
+            `root_widget` is the widget at the root of the kv rule that
+            instantiated this widget, or `None` if it wasn't part of a rule
+            because it is instantiated in python.
+        `on_kv_post`: `(base_widget, )`
+            Fired after all the kv rules associated with the widget
+            and all other widgets that are in any of those rules have had
+            all their kv rules applied. `base_widget` is the base-most widget
+            whose instantiation triggered the kv rules.
 
     .. warning::
         Adding a `__del__` method to a class derived from Widget with Python
@@ -358,7 +365,7 @@ class Widget(WidgetBase):
             Builder.apply(
                 self, ignored_consts=self._kwargs_applied_init,
                 rule_children=rule_children)
-            self.dispatch('on_kv_applied')
+            self.dispatch('on_kv_applied', None)
             for widget in rule_children:
                 widget.dispatch('on_kv_post', self)
             self.dispatch('on_kv_post', self)
@@ -501,10 +508,10 @@ class Widget(WidgetBase):
     def on_kv_pre(self):
         pass
 
-    def on_kv_applied(self):
+    def on_kv_applied(self, root_widget):
         pass
 
-    def on_kv_post(self, root_widget):
+    def on_kv_post(self, base_widget):
         pass
 
     #
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -298,18 +298,12 @@ class Widget(WidgetBase):
         `on_touch_up`: `(touch, )`
             Fired when an existing touch disappears. `touch` is the touch
             object.
-        `on_kv_pre`: `()`
-            Fired before kv rules are applied. It doesn't pass any parameters.
-        `on_kv_applied`: `(root_widget, )`
-            Fired after all the kv rules of the widget is applied.
-            `root_widget` is the widget at the root of the kv rule that
-            instantiated this widget, or `None` if it wasn't part of a rule
-            because it is instantiated in python.
         `on_kv_post`: `(base_widget, )`
             Fired after all the kv rules associated with the widget
             and all other widgets that are in any of those rules have had
             all their kv rules applied. `base_widget` is the base-most widget
-            whose instantiation triggered the kv rules.
+            whose instantiation triggered the kv rules (i.e. the widget
+            instantiated from Python, e.g. ``MyWidget()``).
 
     .. warning::
         Adding a `__del__` method to a class derived from Widget with Python
@@ -329,8 +323,8 @@ class Widget(WidgetBase):
     '''
 
     __metaclass__ = WidgetMetaclass
-    __events__ = ('on_touch_down', 'on_touch_move', 'on_touch_up',
-                  'on_kv_pre', 'on_kv_applied', 'on_kv_post')
+    __events__ = (
+        'on_touch_down', 'on_touch_move', 'on_touch_up', 'on_kv_post')
     _proxy_ref = None
 
     def __init__(self, **kwargs):
@@ -357,15 +351,13 @@ class Widget(WidgetBase):
         if self.canvas is None:
             self.canvas = Canvas(opacity=self.opacity)
 
-        self.dispatch('on_kv_pre')
-
         # Apply all the styles.
         if not no_builder:
             rule_children = []
-            Builder.apply(
-                self, ignored_consts=self._kwargs_applied_init,
+            self.apply_class_lang_rules(
+                ignored_consts=self._kwargs_applied_init,
                 rule_children=rule_children)
-            self.dispatch('on_kv_applied', None)
+
             for widget in rule_children:
                 widget.dispatch('on_kv_post', self)
             self.dispatch('on_kv_post', self)
@@ -401,6 +393,12 @@ class Widget(WidgetBase):
     @property
     def __self__(self):
         return self
+
+    def apply_class_lang_rules(
+            self, root=None, ignored_consts=set(), rule_children=None):
+        Builder.apply(
+            self, ignored_consts=ignored_consts,
+            rule_children=rule_children)
 
     #
     # Collision
@@ -504,12 +502,6 @@ class Widget(WidgetBase):
         for child in self.children[:]:
             if child.dispatch('on_touch_up', touch):
                 return True
-
-    def on_kv_pre(self):
-        pass
-
-    def on_kv_applied(self, root_widget):
-        pass
 
     def on_kv_post(self, base_widget):
         pass
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -297,6 +297,12 @@ class Widget(WidgetBase):
             Fired when an existing touch moves
         `on_touch_up`:
             Fired when an existing touch disappears
+        `on_kv_pre`:
+            Fired before kv rules are applied
+        `on_kv_applied`:
+            Fired after its own kv rules are applied
+        `on_kv_post`:
+            Fired after all kv rules are applied
 
     .. warning::
         Adding a `__del__` method to a class derived from Widget with Python
@@ -316,7 +322,8 @@ class Widget(WidgetBase):
     '''
 
     __metaclass__ = WidgetMetaclass
-    __events__ = ('on_touch_down', 'on_touch_move', 'on_touch_up')
+    __events__ = ('on_touch_down', 'on_touch_move', 'on_touch_up',
+                  'on_kv_pre', 'on_kv_applied', 'on_kv_post')
     _proxy_ref = None
 
     def __init__(self, **kwargs):
@@ -343,9 +350,18 @@ class Widget(WidgetBase):
         if self.canvas is None:
             self.canvas = Canvas(opacity=self.opacity)
 
+        self.dispatch('on_kv_pre')
+
         # Apply all the styles.
         if not no_builder:
-            Builder.apply(self, ignored_consts=self._kwargs_applied_init)
+            rule_children = []
+            Builder.apply(
+                self, ignored_consts=self._kwargs_applied_init,
+                rule_children=rule_children)
+            self.dispatch('on_kv_applied')
+            for widget in rule_children:
+                widget.dispatch('on_kv_post', self)
+            self.dispatch('on_kv_post', self)
 
         # Bind all the events.
         if on_args:
@@ -481,6 +497,15 @@ class Widget(WidgetBase):
         for child in self.children[:]:
             if child.dispatch('on_touch_up', touch):
                 return True
+
+    def on_kv_pre(self):
+        pass
+
+    def on_kv_applied(self):
+        pass
+
+    def on_kv_post(self, root_widget):
+        pass
 
     #
     # Tree management
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -305,6 +305,8 @@ class Widget(WidgetBase):
             whose instantiation triggered the kv rules (i.e. the widget
             instantiated from Python, e.g. ``MyWidget()``).
 
+            .. versionchanged:: 1.11.0
+
     .. warning::
         Adding a `__del__` method to a class derived from Widget with Python
         prior to 3.4 will disable automatic garbage collection for instances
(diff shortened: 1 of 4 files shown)


matham:
>So I want to leave this PR to you, if you don't mind.

I made a couple of small changes and I'm very satisfied with the state of the PR. 

Next I'll try to add a summary to the PR of what it's doing to help reviewers, and then some docs and maybe more tests. 

Given that @tito seemed to be [ok ](https://github.com/kivy/kivy/pull/6221#issuecomment-481301996) with adding multiple versions of these methods - like in this PR, I would like to get this merged in the next week or two.

gottadiveintopython:
That's awesome. I'm sure leaving the PR to others was a right decision, otherwise it will never end, because e.g. I've been thinking option 2 is better than 3 since yesterday.

Anyways, I'm looking forward to where the PR goes.

matham:
I started writing tests and I realized I misunderstood something about `on_kv_applied` (option 1 from the last PR proposed by @tito).

I thought that one would be able to bind to  `on_kv_applied: do_something` in kv in all cases. However, that is not possible. This brings me back to my original objection to having `on_kv_applied`.

Specifically, consider::

```
<SomeWidget>:
    on_kv_applied: do_a

<SomeOtherWidget>:
    SomeWidget:
        on_kv_applied: do_b
```

Under option 1, `do_a` will be executed and but not `do_b`. I hadn't previously realized this was the case. I'm pretty sure this would be very confusing to user to keep track where you're allowed to use this event in kv. Why wouldn't you just use `on_kv_post`, which always works? Is there an example where this would be useful?

@tito can you please clarify that this is what you meant when you said we should offer both options? If you think this is what you want, I'll go ahead and keep it even though I can't see myself using this, but I'd like to move forward with this PR.

gottadiveintopython:
Hm, for me `on_kv_applied` is useful, because of the following example:

```python3
c = Builder.load_string('''
<WidgetA>: # A
    on_kv_applied: do_something(widget)  # AA
    Widget:
        id: widget
<WidgetB>:  # B
    WidgetA:
        id: a
        on_kv_post: do_something(label)  # BB
    Label:
        id: label
BoxLayout: # C
    WidgetA:
        id: a
        on_kv_post: do_something(ti)  # CC
    TextInput:
        id: ti
''')

a = WidgetA()
b = WidgetB()
```

There are three instances of `WidgetA`, which are `a`, `b.ids.a`, `c.ids.a`. And the rule`A` is shared between all instances while rule`B` and rule`C` aren't. In this situation, we should put the code that are common between all instances into `AA`, I think.

And this is related to what I [said](https://github.com/kivy/kivy/pull/6221#discussion_r275757607) before. Given that `on_kv_applied` is for the example above, it should be dispatched before:

```python3
                self._apply_rule(
                    child, crule, rootrule, rule_children=rule_children)
```

matham:
@gottadiveintopython I think you possibly misunderstood what I was asking for. I wasn't asking how you could write code to use this, I know that part. I was asking for real world examples of someone making use of this, where there wouldn't also be possible substantial confusion by end users due to issue I mentioned.

But you're right, that given this is the behavior of `on_kv_applied`, it should be dispatched before `_apply_rule` as that would cause the least inconsistency and confusion. But still, I find this event confusing enough as it could only be used in a base rule, but not in a rule where the widget is actually instantiated.

Nonetheless, I added tests that assume the described behavior, because it seemed that that's what @tito wanted. If he changes his mind we can remove it, otherwise I'll assume `on_kv_applied` stays.

gottadiveintopython:
> I was asking for real world examples of someone making use of this

Sorry for misunderstanding your words. I can't immediately come up with a real world example, but I'll try to find it.

> I find this event confusing enough

Option 1 may be inconvenient, but not confusing for me. Because the restriction that will be brought by option 1 is simple.

```yaml
<WidgetA>:
    on_kv_applied:  # A
    Widget:
        on_kv_applied:  # B
<WidgetB>:
    on_kv_applied:  # A
    WidgetA:
        on_kv_applied:  # B
Widget:
    on_kv_applied:  # A
    WidgetB:
        on_kv_applied:  # B
        Widget:
            on_kv_applied:  # B
```

Only `A`(i.e. `on_kv_applied` at the top-level of each rule) will be excuted, and `B`(not at top-level) won't be excuted. Isn't it quite simple?

I also want to mention about which behavior is better IMO.

1. **option 1 only**
1. **option 2 only** or **option 1 + 2**
1. rest

For me **option 1 only** still is the best because of its simplicity, but it has an disadvantage like you mentioned, I don't push it anymore.
The second best are **option 2 only** or **option 1 + 2**. I belieave **option 1 + 2** can handle a certain situation that **option 2 only** can't, but since having two separated events itself is a bit confusing for the users, I can't certainly say "**option 1 + 2** is better than **option 2 only**".

gottadiveintopython:
Ah I just realized that you moved that line(`widget.dispatch('on_kv_applied', widget)`) to diffirent place compared to the original branch(`post_kv`). In that case

> Only A(i.e. on_kv_applied at the top-level of each rule) will be excuted, and B(not at top-level) won't be excuted.

 doesn't stand.

matham:
>Isn't it quite simple?

No! That is what I was trying to convey in the last PR when you clarified multiple times about how the syntax works. I may agree that the rule is fairly simple to state here in the PR between ourselves, but that is **not** what I mean when I say it's confusing and when I questioned it before.

What I mean is that we need to think of users who don't spend every day writing kv code. First, you'll need to convey to them the meaning of the event. How are you going to explain it clearly? For `on_kv_pre` it's that it fires before any kv is applied to the widget. For `on_kv_post`, it's fired after all the kv rules the widget participates in is applied (i.e. everything is initialized). 

How about for `on_kv_applied`? We would say that it is fired after all the kv rules that define the widget has been applied, except for rules that applied to the widget as it's being instantiated in a kv rule. This wouldn't make sense to most people, so we'd need to give an example showing the subtleties and where you can or cannot use this event. This is already a problem if you're designing an API, because it's too complex to communicate and it's not a thing we should do freely. But ok, we can use an example to illustrate how it work and we can get around this if we really needed.

The biggest problem I have is with the fact that for 

```
<MyWidget>:
    on_kv_applied: do_a

SomWidget:
    MyWidget:
        on_kv_applied: do_b
```

`do_a` will be executed, while `do_b` will not. This adds the complexity and visual incosistency, where a year from now, if I were to try to use this event, I will not remember where it is and is not allowed. This is asking for bugs and confusion by end users (including myself in a year from now).

You have said before that the solution is to move the `do_b` callback to the place where `do_a` is. But that is not a thing we get to do - introduce an inconsistency and tell our users, instead of doing what seems natural, remember to do something else. It's just not how API design works and what we should aim for. We have to make an API that avoids such visual inconsistencies, where if things looks similar, they should have a similar behavior. `on_kv_applied` would violate this. If it works in one place, it should work everywhere unless it's super clear from the context that this is different, which is not the case here at all. 

I would bet many users would not be able to distinguish naively between the context of `do_a` and `do_b`, until they get bitten by what they think is a bug when `do_b` is never executed and then they learn the difference. But that's not something we should ask our users to have to go through.

My solution is to raise an exception if `on_kv_applied` is used in kv, because either it should be legal everywhere or legal nowhere except in python directly.

I'm not sure @tito knew about this behavior when it was originally proposed, so I'm waiting for him to comment with his thoughts.

matham:
I updated the first post with a complete summary to the PR.

gottadiveintopython:
Yes, users who don't spend every day writing kv code will confuse, I agree with that. I valued the simpleness of the implementation rather than the usability. That's why I said "option 1 is enough to handle most situations" in the previous PR.

And I'm probably wrong, so I'll follow you.

tito:
Ok for `on_kv_post`, this is the only new feature i see. `on_kv_pre` and `on_kv_applied` is the same as writing code in `__init__` of the widget before/after the super(), which is what we usually do no?
You made a point on `on_kv_applied`, because you don't see the value, and rather prefer the _post. However, it's not about value, it's more about consistency and separation:
- if you have an event before init/super(), why not an event after init/super() (ok, bad argument?)
- the case you showed it for me perfectly valid: there is a difference between a widget you design, and how it will be used later on by the user; And so there is 2 things about it:
  - you may gain performance in some place because it will reduce the number of event: it may dispatch the final value directly, instead of 2 or more rules execution
  - you may have issue with ordering: before if you instanciate a widget, the rules applied to it would be done before the instance, therefore you are guarantee of some property with the "final" value ready to use. By potentially delaying it, it will also delay the instance that want to use it

Overall, it does have a big impact on the initialization of a Widget itself: 3 new events dispatched by default for any instanciation, it will impact directly the time to create a widget.

After consideration, i would be just happy with just `on_kv_post`, as the 2 others can be done with `__init__` and can be controlled by the user. If that's really really needed, why not considering class decorator for a widget that will create on_kv_pre/on_kv_applied events, and dispatch it?

matham:
@tito, ok I understand now better I think your desire for `on_kv_applied`. I understood your second point, but did not understand what you meant by:

>you may gain performance in some place because it will reduce the number of event: it may dispatch the final value directly, instead of 2 or more rules execution

But you're (almost) right, you can do `on_kv_applied` and `on_kv_pre` by doing something before/after `super`. The difference for me was that `on_kv_pre` is actually executed **after** widget called its `super` but before the rule is applied. This means when `on_kv_pre` is fired, all the widgets in the inheritance chain did call `super` already **and** `EventDispatcher` initialized any kivy properties passed as `kwargs`.

I'm ok with dropping `on_kv_pre` and `on_kv_applied` because you're right about the performence. I do also like the idea to offer `on_kv_pre` and/or `on_kv_applied` as a decorator. I'm not sure how you would implement it? Would you use something like 

```py
def deco(func):
    def inner(*largs, **kw):
        func.__self__.dispatch('on_kv_pre', func.__self__)
        func(*largs **kw)
        func.__self__.dispatch('on_kv_applied', func.__self__)

class MyWidget(Widget):
    @deco
    def __init__(self, **kw):
        super().__init__(**kw)
        ...
```

If so, `deco` is not really that interesting because we could just do something in `__init__` before calling `super`.

matham:
I'll re-post what I posted on discord:

*******

@tito, here's a thought. What if we drop `on_kv_pre` and `on_kv_applied`, keep only `on_kv_post` like you mentioned, but instead of using a decorator, which wouldn't completely resolve the use case I mentioned in my response (`on_kv_pre` being dispatched after EventDispatcher super has been called), we add a method called `apply_kivy_lang_rules`:

```
def apply_kivy_lang_rules(self, ...):
    Builder.apply(self)
```

Then we change the code in widget to something like:

```
if not no_builder:
    self.apply_kivy_lang_rules(...)

```

And in Builder we change the method `_apply_rule` from:

```
widget = Factory.get(parser.root.name)(__no_builder=True)
self.apply(widget)
widget.dispatch('on_kv_applied', widget)
self._apply_rule(widget, parser.root, parser.root)
```

to:

```
widget = Factory.get(parser.root.name)(__no_builder=True)
widget.apply_kivy_lang_rules(...)
self._apply_rule(widget, parser.root, parser.root)
```

This solves all use cases of `on_kv_pre` and `on_kv_applied` because a user could inherit and overwrite `apply_kivy_lang_rules` if they needed to do similar pre/applied initialization?

And it has no additional cost of dispatching two more rules (but someone could then add a decorator to `apply_kivy_lang_rules` if they really wanted those events). I think this is very similar to the original PR, but with the addition of the `on_kv_post` event. Maybe this is what you wanted initially?

****
I implemented the idea from above and it works really nicely. In the tests I added the pre/applied events which are dispatched in the `apply_class_lang_rules` method of the inherited class of the tests and everything works perfectly.

gottadiveintopython:
@tito Thanks for the review.

> on_kv_pre and on_kv_applied is the same as writing code in `__init__` of the widget before/after the super(), which is what we usually do no?

For me, those are not the same. `on_kv_applied` in particular, because the stuff defined in kv is not guaranteed to be availabe inside `__init__()`, and delaying stuff by `Clock.schedule_once()` is not a perfect workaround. (But I feel like this is not your point)

@matham 

The current code looks much better than before, but what is `root` parameter of `apply_class_lang_rules()` for? Is that useful? Considering that `apply_class_lang_rules()` only applies widget's own rule, should it not depend on root widget?

tito:
I wanted to have a look at a benchmark about the potential perfs loss with just the creation.
It doesn't reflect the framework usage obviously, but here is for reference:

Before (https://hastebin.com/jojuhaqepa):
```
System informations
-------------------
OS platform     : darwin
Python EXE      : /Users/tito/miniconda3/bin/python
Python Version  : 3.6.6 |Anaconda, Inc.| (default, Jun 28 2018, 11:07:29) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
Python API      : 1013
Kivy Version    : 1.11.0.dev0
Install path    : /Users/tito/code/kivy/kivy
Install date    : Wed Apr 10 15:26:16 2019

OpenGL informations
-------------------
GL Vendor: b'Intel Inc.'
GL Renderer: b'Intel(R) Iris(TM) Graphics 550'
GL Version: b'2.1 INTEL-12.8.38'

Benchmark
---------
 1/8  Core: button creation (10000 * 10 a-z)                       7.150592
 2/8  Core: button creation (10000 * 10 a-z), with Clock.tick      9.045682
 3/8  Core: label creation (10000 * 10 a-z)                        5.266693
 4/8  Core: label creation (10000 * 10 a-z), with Clock.tick       7.082703
 5/8  Widget: creation (10000 Widget)                              1.790820
 6/8  Widget: creation (10000 Widget + 1 root)                     2.621336
 7/8  Widget: event dispatch (1000 on_update in 10*1000 Widget)    0.062472
 8/8  Widget: empty drawing (10000 Widget + 1 root)                0.001407

Result: 33.021705
```

After (https://hastebin.com/awixomorob):
```
System informations
-------------------
OS platform     : darwin
Python EXE      : /Users/tito/miniconda3/bin/python
Python Version  : 3.6.6 |Anaconda, Inc.| (default, Jun 28 2018, 11:07:29) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
Python API      : 1013
Kivy Version    : 1.11.0.dev0
Install path    : /Users/tito/code/kivy/kivy
Install date    : Wed Apr 10 15:26:16 2019

OpenGL informations
-------------------
GL Vendor: b'Intel Inc.'
GL Renderer: b'Intel(R) Iris(TM) Graphics 550'
GL Version: b'2.1 INTEL-12.8.38'

Benchmark
---------
 1/8  Core: button creation (10000 * 10 a-z)                       7.089378
 2/8  Core: button creation (10000 * 10 a-z), with Clock.tick      8.967459
 3/8  Core: label creation (10000 * 10 a-z)                        5.832057
 4/8  Core: label creation (10000 * 10 a-z), with Clock.tick       7.295211
 5/8  Widget: creation (10000 Widget)                              2.059494
 6/8  Widget: creation (10000 Widget + 1 root)                     2.250796
 7/8  Widget: event dispatch (1000 on_update in 10*1000 Widget)    0.047028
 8/8  Widget: empty drawing (10000 Widget + 1 root)                0.001594

Result: 33.543017
```

tito:
I'm ok to remove `on_kv_applied` then.

matham:
@tito I removed `on_kv_pre` and `on_kv_applied`, left `on_kv_post` and added the `apply_kivy_lang_rules` method as described here https://github.com/kivy/kivy/pull/6257#issuecomment-484997620. Is that OK? I think that's the best of both worlds!?

kivy/uix/widget.py
```diff
@@ -379,6 +396,78 @@ def __hash__(self):
...
+            ``<MyWidget>`` and ``<OtherWidget>`` rules to it - it does not
+            apply the ``width: 124`` rule. The ``width: 124`` rule is part of
+            the ``MyBox`` rule and is applied by the ``MyBox``'s instance's
+            :meth:`apply_class_lang_rules`.
+
+        .. versionchanged:: 1.11.0
+        '''
+        Builder.apply(
+            self, ignored_consts=ignored_consts,
+            rule_children=rule_children)
```
Fak3: I feel uneasy about passing the rule_children in all these methods. Could it be returned instead, and thus accumulated in the calls up the call stack? *This is just a guess, I have not analyzed the code thoroughly.

matham: It could be returned as you suggest, but I find it worse to return that from apply, rather than appending to a rule_children list passed. 

The reason is that `apply` applies kv rules and if it were to return something it should be the root widget, or maybe the list of rules applied or something similar. I.e. it should return something kv rule related that was just applied - and in the future that's how we might change it so that it returns something like that. 

If we were to return a list of children, that would preclude being able to change it to return something more relevant in the future. And specifically, returning a list of children doesn't seem relevant, except for this specific need to dispatch `on_kv_post`. If we didn't have this need, we would never return the list of children as this doesn't seem useful elsewhere in how that method would be used. That's why it seemed more appropriate to pass the list of children to be filled.

Can you explain why you feel uneasy? I get that it's not particularity pretty, but I don't see issues with it. Function parameters we have many - we have but one return statement.

Issue #4680, opened 2016-10-19
Raspberry pi touchring (mouse pointer) can move endless behind screen boarders


wolkstein:
hi,
i will control my kivy application with an normal mouse. no touch screen. if i develop on my desktop the mouse pointer change from touchring to normal mouse pointer if i cross the kivy application window boarder. on an raspberry pi the application is fullscreen against .kivy/config.ini settings witch are default on "fullscreen = 0". this is not the problem because the app must run in fullscreen mode on raspberry pi. the problem is the touchring. on raspberry pi you can move the mouse(touchring) endless behind the screen boarder. in this case, if you not remember in which direction you loose the mouse pointer behind the boarder you have no chance to get the pionter back to visible screen area.

what goes wrong here? or how can i configure config.ini that this will not happens?

thx
wolke

iconnectviacalculator:
Hello,
I've got the same issue. Currently working on it but no couldn't figure it out yet. Did you find a solution?
Thanks.

wolfmanjm:
I think this is an issue with hidinput, and can be fixed by limiting the rel mouse moves to stay within the screen limits. I'll try to add it to my rework of hidinput. PR #6264

Julian-O:
This looks like it was fixed.

Closing.

If the problem still occurs, please let us know.

Issue #4007, opened 2016-02-17
kivy Keyboard issues on Raspberry : "ctrl" key crashes kivy


public-hons:
kivy 1.9.1 from the latest KivyPie Version 0.7: 

There seems to be a "ctrl" entry missing in Keyboard.keycodes - whenever i hit the "ctrl" key i get an exception:
 Exception in thread Thread-2:
 Traceback (most recent call last):
   File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
     self.run()
   File "/usr/lib/python2.7/threading.py", line 763, in run
     self.__target(_self.__args, *_self.__kwargs)
   File "/usr/local/lib/python2.7/dist-packages/kivy/input/providers/hidinput.py", line 701, in _thread_run
     process_as_mouse_or_keyboard(*infos)
   File "/usr/local/lib/python2.7/dist-packages/kivy/input/providers/hidinput.py", line 551, in process_as_mouse_or_keyboard
     Keyboard.keycodes[l.lower()],
 KeyError: 'ctrl'

I have a logitech Keyboard with german keyboard-layout. 
Can be tested with kivy-examples/widgets/keyboardlistener.py or any other example that binds to the keyboard.
I can work around this by manually setting 
Window._system_keyboard.keycodes['ctrl'] = 305
in on_start(..), but this seems to be rather ugly.

Andrewnetwork:
Just got this error.

KeyWeeUsr:
@public-hons @Andrewnetwork only on rpi, or does it happen on other platforms with the same keyboard?

public-hons:
only on rpi, have not had it on other platforms yet

Andrewnetwork:
Only on RPI

KeyWeeUsr:
@public-hons is there any difference in the error when you press right or left `ctrl`? If not, I think it may be the provider's fault of not being able to differentiate between those keys (or basically using wrong key in dictionary), because there actually is `ctrl` value but in two separate keys as [`lctrl == 305` and `rctrl == 306`](https://github.com/kivy/kivy/blob/master/kivy/core/window/__init__.py#L63)

public-hons:
no, there is no difference for left and right control key and yes, that seems to be the problem (that is why i added Window._system_keyboard.keycodes['ctrl'] = 305 in the startup routine to prevent this crash)

[...]

PR #6264, opened 2019-04-20
Fix the ctrl bug in hidinput (Issue #4007)

wolfmanjm:
Add mouse button handling to hidinput (was partially implemented before), enabled button profile.
Add scroll wheel events to hidinput, converts to correct button event.
Add ctrl key handling to hidinput, previously ctrl key was not handled at all.
Tighten up protections against crash if unexpected keys are used.
Also fixes Issue #4680

### DIFF
--- a/kivy/input/providers/hidinput.py
+++ b/kivy/input/providers/hidinput.py
@@ -595,9 +595,9 @@ else:
                             dispatch_queue.append(('key_up', (
                                 keycode, ev_code,
                                 keys_str.get(z, z), Window._modifiers)))
-                            if z == 'shift' or z == 'alt':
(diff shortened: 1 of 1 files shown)


kivy/input/providers/hidinput.py
```diff
@@ -548,22 +568,40 @@ def process_as_mouse_or_keyboard(
                                 point['id'] += 1
                                 point['_avoid'] = True
                     else:
+                        if ev_value < 0 or ev_value > 1:
```
tshirtman: i would have rather written it this way
```suggestion
                        if not 0  <= ev_value <= 1:
```
but seems that produces more opcodes (26 instead of 21), so your version might be faster, not a biggie anyway.

PR #6269, opened 2019-04-23
Add ability to specify dash offsets for Line

matham:
Continuation from https://github.com/kivy/kivy/pull/3524.

I tested it, cleaned it up and added an example to the example code.

During testing, I discovered that when switching the mode of a line, settings leak from the previous modes. I fixed as much a I could, but it's a unrelated issue. E.g. when switching between `close` true/false, things are not properly cleaned between them. But as I said this predates the PR and is unrelated to it.


tito:
Fantastic! Good job.

PR #6273, opened 2019-04-24
Fix "Object no attribute active" (Bug introduced via PR #4898)

misl6:
When using the `ToggleButtonBehavior` in a object without the `.active` variable (ex. tabbedbar) the PR #4898 introduces a "object has no attribute `active`" error.

This fix moves the edited `on_group` function, in the checkbox implementation.

### DIFF
--- a/kivy/uix/checkbox.py
+++ b/kivy/uix/checkbox.py
@@ -33,6 +33,7 @@ __all__ = ('CheckBox', )
 from kivy.uix.widget import Widget
 from kivy.properties import BooleanProperty, StringProperty, ListProperty
 from kivy.uix.behaviors import ToggleButtonBehavior
+from weakref import ref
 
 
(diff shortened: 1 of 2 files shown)


kivy/uix/checkbox.py
```diff
@@ -172,6 +173,22 @@ def on_active(self, instance, value):
...
+                if item() is self:
+                    group.remove(item)
+                    break
+        group = self._previous_group = self.group
+        if group not in groups:
+            groups[group] = []
+        r = ref(self, ToggleButtonBehavior._clear_groups)
+        groups[group].append(r) 
+        if self.active:
+            self._release_group(self)
```
matham: Instead of copying the code from button's `on_group`, it's better to just call `super`. and then do stuff after it.

misl6: You're perfectly right. Fixed it.

PR #6280, opened 2019-05-01
Add helpful troubleshooting tips

matham:
This adds some clearer error messages and tools to help users struggling with kivy installation issues.

Among the bigger things is it tries to detect is if the user forgot to compile kivy and warns about. It also adds the opportunity to use the dependency walker tool to figure out which of the dependencies is missing in the **Unable to find any valuable Window provider** case. They can then share the generated file, which anyone could open with the dependency walker tool to figure out which dlls are missing etc.

I'm hopeful that this will significantly cut down on the number of users who get stuck on installation issues and open requests on github about it. And in case it does happen, people on discord/mailing list will have the tools to help them using the generated log.

### DIFF
--- a/kivy/clock.py
+++ b/kivy/clock.py
@@ -359,8 +359,18 @@ from kivy.config import Config
 from kivy.logger import Logger
 from kivy.compat import clock as _default_time, PY2
 import time
-from kivy._clock import CyClockBase, ClockEvent, FreeClockEvent, \
-    CyClockBaseFree
+try:
+    from kivy._clock import CyClockBase, ClockEvent, FreeClockEvent, \
+        CyClockBaseFree
+except ImportError:
+    Logger.error(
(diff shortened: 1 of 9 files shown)


kivy/clock.py
```diff
@@ -359,8 +359,18 @@ def my_callback(dt):
...
 from kivy.compat import clock as _default_time, PY2
 import time
-from kivy._clock import CyClockBase, ClockEvent, FreeClockEvent, \
-    CyClockBaseFree
+try:
+    from kivy._clock import CyClockBase, ClockEvent, FreeClockEvent, \
+        CyClockBaseFree
+except ImportError:
+    Logger.error(
+        'Clock: Unable to import kivy._clock. Have you perhaps forgotten to '
```
inclement: Amazing!

PR #6281, opened 2019-05-01
SetWindowLongPtrW ctypes prototype bug

CristiFati:
*SetWindowLongPtrW*'s prototype (*ctypes*) is incorrectly defined (type sizes are correct though, and in *C*  it would work with some *cast*s, but in *Python* things are a bit trickier).

Also, when used in conjunction with other modules (e.g. *Pywinauto*) that correctly set the prototype, (there are cases when) program abnormally exits.

Encountered on v*1.10.1*.

Investigation details are on **[[SO]: ctypes.ArgumentError when using kivy with pywinauto (@CristiFati's answer)](https://stackoverflow.com/questions/55928463/ctypes-argumenterror-when-using-kivy-with-pywinauto/55931295#55931295)**.

The proposed fix:
- Corrects the function prototype
- Creates a wrapper function that does all the conversions required to work with current scenario
- Removed *GWL\_WNDPROC* (2<sup>nd</sup> argument) from all client places, and only specified it once
- Since only the *GWL\_WNDPROC* use-case is required, changed wrapper name from *SetWindowLong\_wrapper* to *SetWindowLong\_WndProc\_wrapper*, as it better reflects the behavior

I didn't test it on *32bit* (lol, I didn't test a lot of things, good thing *CI* is in place).

### DIFF
--- a/kivy/input/providers/wm_common.py
+++ b/kivy/input/providers/wm_common.py
@@ -62,8 +62,6 @@ if 'KIVY_DOC' not in os.environ:
                         c_int, c_longlong, c_void_p, Structure,
                         sizeof, byref, cast)
 
-    LONG_PTR = c_longlong
-
     class RECT(Structure):
         _fields_ = [
             ('left', LONG),
@@ -127,6 +125,7 @@ if 'KIVY_DOC' not in os.environ:
         return _closure
 
     try:
+        LONG_PTR = c_longlong
         windll.user32.SetWindowLongPtrW.restype = LONG_PTR
         windll.user32.SetWindowLongPtrW.argtypes = [HWND, c_int, LONG_PTR]
         SetWindowLong_WndProc_wrapper = \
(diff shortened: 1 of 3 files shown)


kivy/input/providers/wm_common.py
```diff
@@ -116,14 +117,26 @@ def _event_type(self):
                 return 'end'
         event_type = property(_event_type)
 
+    def SetWindowLong_WndProc_wrapper_generator(func):
+        def _closure(hWnd, wndProc):
+            oldAddr = func(hWnd, GWL_WNDPROC, cast(wndProc, c_void_p).value)
```
matham: This function is used both for `SetWindowLongPtrW` and `SetWindowLongW`. But for the latter case, why are you casting to void pointer? I get that on 32 bit a void pointer and a long probably are the same, but still.

CristiFati: As I specified in the description, I did not test on *32bit* (I don't want to install all the modules on any of my *32bit* *VEnv*s).

Nevertheless, this is consistent behavior, and more:
- `LONGLONG` and `void*` are the same on *64bit* (8)
- `LONG` and `void*` are the same on *32bit* (4)

Even if their size matches (and from *C* one could get away with a compiler warning - depending on flags), they are still different (from *ctypes* perspective). That's why all those conversions are required.

Bottom line: there's absolutely no difference between *64bit* and *32bit* cases.

matham: Yeah, I guess it's not worth replicating the function.

PR #6283, opened 2019-05-01
Limit live resize to desktop

### LINKED ISSUE #6277: Events bound to Window properties may crash if updating opengl when Window doesn't exist
Apparently this can happen on Android when the Window changes due to e.g. app rotation, as the Window context actually gets destroyed and created again (I think).

This is described at https://github.com/kivy/python-for-android/issues/1797, issue opened here because I think this needs a fix within Kivy.

akshayaurora:
Closes #6277

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -11,7 +11,8 @@ from kivy.graphics.cgl cimport *
 
 from cpython.mem cimport PyMem_Malloc, PyMem_Realloc, PyMem_Free
 
-is_desktop = Config.get('kivy', 'desktop') == '1'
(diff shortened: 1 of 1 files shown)


kivy/core/window/_window_sdl2.pyx
```diff
@@ -11,6 +11,8 @@ from kivy.graphics.cgl cimport *
 
 from cpython.mem cimport PyMem_Malloc, PyMem_Realloc, PyMem_Free
 
+is_desktop = Config.get('kivy', 'desktop') == '1'
```
matham: I think you need to guard it with `if not environ.get('KIVY_DOC_INCLUDE'):`

Issue #6285, opened 2019-05-02
Caps Lock not detected


Inspirateur:
### Versions

* Python: 3.7
* OS: Windows 10
* Kivy: 1.10.1
* Kivy installation method: pip

### Description

I'm trying to add a little warning message on my passworld field when Caps Lock is on, but I can't seem to get this info from kivy.

### Code and Logs
I just tried to print keycode, text and modifiers
```python
class TextInputCapsWarn(TextInput):
	def keyboard_on_key_down(self, window, keycode, text, modifiers):
		print(keycode, text, modifiers)
```
And this is what I get when typing `a, z d, Caps Lock ON, a, a, z` :
```
(97, 'a') a []
(122, 'z') z []
(100, 'd') d []
(301, 'capslock') ĭ []
(97, 'a') a []
(97, 'a') a []
(122, 'z') z []
``` 
So basically the modifiers list remain empty even with caps locks ON, and while I do detect the caps lock press I can't tell if it's ON or OFF.

This is a really minor issue but I thought documenting on it wouldn't hurt.
Thanks for your time.

matham:
Here's a fix https://github.com/kivy/kivy/pull/6286.

P.S. take a look at the `keyboard_on_textinput` method. That provides the properly capitalized text.

Inspirateur:
Thanks for the effort :D
The `keyboard_on_textinput` doesn't solve my issue on its own because I can't tell if the text has been capitalized with SHIFT (which would be intentionnal) or Caps Lock (which could be unintentionnal), that being said I'll keep an eye on your PR and I hope it gets accepted !

PR #6286, opened 2019-05-03
Add caps and numlock to the modifiers

matham:
Fixes https://github.com/kivy/kivy/issues/6285.

I looked at https://wiki.libsdl.org/SDL_Keymod and SDL_keycode.h and updated the code based on this. 

Specifically,
* I renamed the `KMOD_xMETA` key to `KMOD_xGUI`, because that's what sdl2 calls it (it's the key with the windows icon typically).
* I added `KMOD_MODE`, this is basically the same as the right alt key.
* I added the `KMOD_NUM` and `KMOD_CAPS` keys, which are the numslock and capslock keys.

### DIFF
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -37,14 +37,20 @@ from kivy.utils import platform, deprecated
 from kivy.compat import unichr
 from collections import deque
 
-KMOD_LCTRL = 64
-KMOD_RCTRL = 128
-KMOD_RSHIFT = 2
-KMOD_LSHIFT = 1
-KMOD_RALT = 512
-KMOD_LALT = 256
-KMOD_LMETA = 1024
-KMOD_RMETA = 2048
+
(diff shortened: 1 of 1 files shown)


pythonic64:
@matham Should `Scroll Lock` also be added to modifiers?

matham:
I didn't see it in the sdl2 list. I added most everything from their site. 

But also, probably not as it is probably not used much? We can always add it later if someone does need it.

pythonic64:
I agree that it is not probably used much, but it would make modifiers list complete imho. 
This issue https://github.com/kivy/kivy/issues/5062 is still present in current master ca3a68d891f9dd6233fbe0f520cb304825208c4b.
Can you take a look at its pull request? Just asking because you knowledge of  key handling.

matham:
I'm not really very familiar with the keyboard logic. Also, it'a not super clear to me how ready that PR is to be merged (for starters it has merge conflicts).

PR #6288, opened 2019-05-03
Cache `text` property in TextInput

pythonic64:
This request aims to speedup calls of `text` property by caching it. 

Attribute `_lines_flags` isn't converted to `ListProperty` as it is always set before `_lines` property and they work together to produce text. Converting it to property can introduce bugs because `text` value will get rebuild first on `_lines_flags` and then on `_lines` changes, making a space for possible bug between two rebuilds. 
Internal calls `self._get_text(encode=False)` are replaced with `self.text` because `self.text` will now return cached value which is built with `encode=False`.

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -1054,8 +1054,8 @@ class TextInput(FocusBehavior, Widget):
             self._lines_flags = list(reversed(
                 flags[:m1srow] + flags[m2srow:m2erow] + flags[m1srow:m1erow] +
                 flags[m2erow:]))
-            self._lines = (lines[:m1srow] + lines[m2srow:m2erow] +
-                           lines[m1srow:m1erow] + lines[m2erow:])
+            self._lines[:] = (lines[:m1srow] + lines[m2srow:m2erow] +
+                              lines[m1srow:m1erow] + lines[m2erow:])
             self._lines_labels = (labels[:m1srow] + labels[m2srow:m2erow] +
                                   labels[m1srow:m1erow] + labels[m2erow:])
             self._lines_rects = (rects[:m1srow] + rects[m2srow:m2erow] +
@@ -1896,7 +1896,7 @@ class TextInput(FocusBehavior, Widget):
         if mode == 'all':
             self._lines_labels = _lines_labels
             self._lines_rects = _line_rects
-            self._lines = _lines
+            self._lines[:] = _lines
         elif mode == 'del':
             if finish > start:
                 self._insert_lines(start,
@@ -1961,7 +1961,7 @@ class TextInput(FocusBehavior, Widget):
         if len_lines:
             _lins.extend(_lines)
         _lins.extend(self._lines[finish:])
-        self._lines = _lins
+        self._lines[:] = _lins
 
     def _trigger_update_graphics(self, *largs):
         self._update_graphics_ev.cancel()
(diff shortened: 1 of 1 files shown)


pythonic64:
@tito @matham Can you review this pull request?

matham:
I'm not familiar with the textinput code. @akshayaurora is probably the person to review this.

kivy/uix/textinput.py
```diff
@@ -3083,31 +3083,27 @@ def _get_text(self, encode=False):
         lf = self._lines_flags
         l = self._lines
         len_l = len(l)
-
         if len(lf) < len_l:
+            lf = lf[:]
```
tshirtman, line 3087: not sure about the intent of this?

pythonic64: Next line is `lf.append(1)` which changes copied `lf` list and we don't want to change original `lf` as it references `self._lines_flags`. 

On other places `sefl._lines[:]` is used to update `_lines` property as `_lines` is a `ListProperty` type and that type creates internally new `list` of set `list` so updating property is faster. 
```python
_lines = ListProperty()

new_value = [...]
self._lines = new_value  # slower as it creates new copy of `new_value`
self._lines[:] = new_value # faster as it updates `self._lines` as `ListProperty` does not creates new copy by calling `list(new_value)`
```

Same goes for `_hint_text_lines` and other `ListProperty` types.

tshirtman: but isn't updating `_line_flags` desired?

pythonic64: Better solution would be to append(1) and call pop() after text gets built, that way we don't have to copy `lf` list. I don't think that changing `self._lines_flags` inside of getter method is a good idea as getter should only get the value and that change to `self._lines_flags` should go somewhere else.

tshirtman: did you test the shift lines feature after this patch? i wonder if it impacts, as this feature reverse the lines_flags before working and then reversing it again, using index lookup in the middle, an off-by-one error could cause issues, but damn does this code needs some comments.

tshirtman: as for a better solution to avoid changing the value, i think i would just update the test to return '\n' if `i > len(lf)` so no copy, and no change. there are probably more sophisticated functions with itertools.izip and repeat, but not sure that's warranted here.

pythonic64: Just test shifting of line(s) and it works.
In method `_shift_lines` attribute `_lines_flags` is set before property `_lines` so getter of `text` will be called with updated `_lines_flags` as `text` is only updated on `_lines` change. The same is true as long as `_lines_flags` are set before `_lines` as currently is the state in `TextInput`.

pythonic64: Can you post your version of `_get_text` method?

pythonic64: And what do you think about `encode` argument and `encode` part of `_get_text` method?

tshirtman:
```python
    def _get_text(self, encode=False):
        lf = self._lines_flags
        l = self._lines
        len_l = len(l)
        len_lf = len(lf)

        text = u''.join([
            (u'\n' if (i >= len_lf or lf[i] & FL_IS_LINEBREAK) else u'') + l[i]
            for i in range(len_l)
        ])

        if encode and not isinstance(text, bytes):
            text = text.encode('utf8')
        return text
```
is what i would try

tshirtman: about encode, i'm not sure, i don't see where it's even used, i see it was True in the paste and then changed to false, i think it could maybe be dropped completly, but that's certainly a change for another PR.

pythonic64: I made an update but decided to use two `if` statement instead of using `if` in text join as that will make `i >= len_lf ` test for every line. If we have more than two lines then current code would be better and acceptable trade-off for case with two or less lines.

tshirtman: ok, i didn't think that int comparison in the loop was a big deal, but your solution is fine as well, i can imagine problems with multithreading, but it's not supposed to happen anyway.

pythonic64: Yeah, multithreading is no-go with widgets and can open a whole set of problems.
Merge this request if you think that current changes are enough.

Issue #6144, opened 2019-02-08
resizing the window halves image sizes on retina displays


toejough:
### Versions

* Python: 3.7.2
* OS: macOS 10.14.3
* Kivy: 1.11.0.dev0 (and before that, 1.10.1)
* Kivy installation method: git clone (and before that, pip install)

### Description

*What are you trying to get done?*
I'm trying to display an image correctly when resizing the window on a retina display.

*what has happened?*
The image is halved in height and width when resizing the window.

*what went wrong, and what did you expect?*
The image being half-sized (well, really, quarter-sized, by area), is wrong.  I expect it to scale correctly for the display.

*other*
The image scales just fine on a non-retina display.  With the code sample below, it scales fine when the window is *moved* on a retina display.  

Without the image reload code, the image remains 1/4 size when moving or resizing the window.

With the image reload code, the image remains 1/4 size when resizing the window, but scales up as I expect/want when moving the window.

I'm confused what's going on - the reload from disk every 0.1s is gross, but I expect it at least to work consistently.  I don't understand why reloading the image makes it scale correctly when the window moves, but not when it's resized.

I've tried putting the image in a layout and calling `do_layout`, I've tried explicitly setting the image size with & without `kivy.metrics.dp` in the `_resize` func, and I've tried taking action on window resize, to no avail.

### Code and Logs

```python
class App(app.App):
    def build(self):
        clock.Clock.schedule_interval(self._resize, 0.1)
        self._image = image.Image(source=IMAGE_PATH, allow_stretch=False)
        return self._image

    def _resize(self, dt):
        self._image.reload()


if __name__ == "__main__":
    App().run()
```

```
[INFO   ] Kivy: v1.10.1
[INFO   ] Python: v3.7.2 (default, Jan 10 2019, 12:43:58) 
[Clang 10.0.0 (clang-1000.10.44.4)]
[INFO   ] Factory: 194 symbols loaded
[INFO   ] Image: Providers: img_tex, img_imageio, img_dds, img_sdl2, img_pil, img_gif (img_ffpyplayer ignored)
[INFO   ] Window: Provider: sdl2
[INFO   ] GL: Using the "OpenGL ES 2" graphics system
[INFO   ] GL: Backend used <gl>
[INFO   ] GL: OpenGL version <b'2.1 ATI-2.4.10'>
[INFO   ] GL: OpenGL vendor <b'ATI Technologies Inc.'>
[...]

PR #6289, opened 2019-05-04
- calculate dpi on window update

maho:
this is fix of issue #6144.

It's not ideal solution, as fonts are 2x larger after moving window from retina to non-retina, or 2x smaller if moved window from non-retina to retina. 

However, it's way better than it was before, so I think it's worth to apply.

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -404,6 +404,14 @@ class WindowBase(EventDispatcher):
         else:
             return False
 
+    def _recalc_dpi(self):
+        # calculate density
+        w, h = self.system_size
+        sw, sh = self._win._get_gl_size()
+        self._density = density = sw / w
+        if self._is_desktop and self.size[0] != w:
+            self.dpi = density * 96.
+
     minimum_width = NumericProperty(0)
     '''The minimum width to restrict the window to.
 
@@ -1440,6 +1448,9 @@ class WindowBase(EventDispatcher):
         from math import radians
 
         w, h = self.system_size
+
+        self._recalc_dpi()
+
         if self._density != 1:
             w, h = self.size
 
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -81,6 +81,7 @@ from functools import partial
 from copy import copy
 from kivy import kivy_data_dir
 from kivy.config import Config
+from kivy.logger import Logger
 from kivy.utils import platform
 from kivy.graphics.texture import Texture
 from kivy.core import core_select_lib
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -277,15 +277,9 @@ class WindowSDL(WindowBase):
             resizable = Config.getboolean('graphics', 'resizable')
             state = (Config.get('graphics', 'window_state')
                      if self._is_desktop else None)
-            self.system_size = _size = self._win.setup_window(
+            self.system_size = self._win.setup_window(
                 pos[0], pos[1], w, h, self.borderless,
                 self.fullscreen, resizable, state)
-
-            # calculate density
-            sz = self._win._get_gl_size()[0]
-            self._density = density = sz / _size[0]
-            if self._is_desktop and self.size[0] != _size[0]:
-                self.dpi = density * 96.
 
             # never stay with a None pos, application using w.center
             # will be fired.


kivy/core/window/__init__.py
```diff
@@ -404,6 +404,14 @@ def _set_size(self, size):
         else:
             return False
 
+    def _recalc_dpi(self):
+        # calculate density
+        w, h = self.system_size
+        sw, sh = self._win._get_gl_size()
```
akshayaurora: This function is specific to window_sdl2, won't work in anything else as that does not exist else where. 
This is why this function was only in window_sdl2

akshayaurora: This code runs on various platforms too so there needs to be a check `platform == osx` too.

akshayaurora:
There is also the issue of one having to manually resize the window to display the updated window. 
Where as this should happen automatically, we should be listening to the selector for BackingScaleFactor on osx to update the window automatically, ref: https://stackoverflow.com/questions/11067066/mac-os-x-best-way-to-do-runtime-check-for-retina-display .

kivy/core/window/__init__.py
matham: I don't think this would be specific to osx. Windows (and probably ubuntu) also support screens with different `dpi`. E.g. I have a 4k screen on my laptop, but my external monitor is a normal HD screen. 

Although I'm not sure sdl2 currently listens to these dpi changes on windows. See https://discourse.libsdl.org/t/wip-windows-high-dpi-support/22446 (`SDL_WINDOW_ALLOW_HIGHDPI`).

akshayaurora: @mahomahomaho search for def _get_gl_size it is only defined here https://github.com/kivy/kivy/blob/master/kivy/core/window/_window_sdl2.pyx#L291

@matham I forgot that you could manually set DPI in windows, so you are right.

In light of this probably a better idea is to implement `get_content_size` abstract function in window.__init__ and have it be overridden in specific implementations like window_sdl2 etc.

Even if SDL2 does not directly support DPI on windows yet, we could always call direct WINAPI's and OSX API with `if platform ==` checks. We do not have to depend on sdl2 to implement this considering we also have other window providers too.

akshayaurora:
@matham @tito comment here https://github.com/kivy/kivy/issues/6144#issuecomment-475140897 `it would make send to provide a `density_dynamic` kind of a new prop that user could listen to and update their own UI accordingly`

let's keep this pr atomic, i.e not changing stuff like reify etc that would effect a lot of stuff, not just dpi etc.

We also should test things further in SDL2 as it currently manages moving from high density to low density without any issues. Only dpi-Fonts are wrong/double in that case. let's forget fonts and only focus on density we can talk about fonts/reify issue in another pr.
Need to check maybe this can be fixed in sdl2 itself? Considering `High to Low` works automatically... so should `Low to High`

kivy/core/window/__init__.py
akshayaurora: search for def _get_gl_size it is only defined here https://github.com/kivy/kivy/blob/master/kivy/core/window/_window_sdl2.pyx#L291

akshayaurora:
This is related too; the Screen module would not  work cause of these issue on a retina screen as we try to change the density * dpi after the window has already been initialized.

matham:
I hadn't seen @tito 's comment. We can make this PR only add code to recompute window's `dpi` and leave Metrics for another PR. But I'm not really sure this PR would then solve anything, given that the `dpi` is cached in the two places I mentioned in my original comment.

But it is ok to be modular, and focus this PR only on Window. However, @tito did mention that this PR should make dpi more dynamic, which this PR doesn't solve as `dpi` is still not a kivy property.

kivy/core/window/__init__.py
mahomahomaho: Hmm, if _get_gl_size is defined only in SDL2 provider, but is used outside it (in _get_size), then it's strange. 

Does it mean that other providers are not used at all, so nobody reported error about undefined _get_gl_size?

mahomahomaho:
> There is also the issue of one having to manually resize the window to display the updated window.
> Where as this should happen automatically, we should be listening to the selector for BackingScaleFactor on osx to update the window automatically, ref: https://stackoverflow.com/questions/11067066/mac-os-x-best-way-to-do-runtime-check-for-retina-display .

Dragging window from monitor to monitor is enough to update content (in my case ofc)

mahomahomaho:
> @matham @tito comment here [#6144 (comment)](https://github.com/kivy/kivy/issues/6144#issuecomment-475140897) `it would make send to provide a `density_dynamic` kind of a new prop that user could listen to and update their own UI accordingly`
> 
> let's keep this pr atomic, i.e not changing stuff like reify etc that would effect a lot of stuff, not just dpi etc.
> 
> We also should test things further in SDL2 as it currently manages moving from high density to low density without any issues. Only dpi-Fonts are wrong/double in that case. let's forget fonts and only focus on density we can talk about fonts/reify issue in another pr.

I second that. I think that this change should make kivy "not worse than it was before" :). Thus I think that resolving problem with scope of _get_gl_size should be enough for this ticket. 

Unfortunately I have no monitors in nearest days, so I need to wait until I will be able to work on it again (Monday?)

kivy/core/window/__init__.py
matham: >Does it mean that other providers are not used at all, so nobody reported error about undefined _get_gl_size

Probably. I don't think anyone uses pygame, and in fact we plan to remove it after this release.

akshayaurora: The Rpi and the X11 window provider on linux are the ones that I can think of people using... maybe people switch to sdl2 just cause these other providers were broken?

Anyways,  we had identified a problem, makes sense to get that fixed rather than keep ignoring.

matham:
As @akshayaurora said, this will likely break on other platforms that don't use sdl2 because `self._win._get_gl_size` is specific to sdl2 and doesn't exist on other platforms.

kivy/core/window/__init__.py
```diff
@@ -1440,6 +1448,9 @@ def update_viewport(self):
         from math import radians
 
         w, h = self.system_size
+
+        self._recalc_dpi()
```
tshirtman, line 1452: Indentation seems weird here

kivy/core/window/__init__.py
tshirtman: So maybe just move this implementation to sdl2_window.py and put a placeholder here so we can merge?

akshayaurora:
@tshirtman this does not really work right now,  it needs a architectural change.
 Recalculation of DPI on every frame is also not a wise thing to do.

Currently with his change you only get the window.dpi and window.density value change.   The UI get’s changed to a invalid mix,  with the metrics.dpi still in the old value and thus the fonts and widgets still use old values.

tshirtman:
> @tshirtman this does not really work right now, it needs a architectural change.
> Recalculation of DPI on every frame is also not a wise thing to do.

I don't think this recomputes on every frame, i only see it being recomputed on window resize (which would be triggered by the dpi change of the system.

> Currently with his change you only get the window.dpi and window.density value change. The UI get’s changed to a invalid mix, 
with the metrics.dpi still in the old value and thus the fonts and widgets still use old values.

Is it better or worse than the current result when dpi changes? Is it an acceptable stepping stone to a more complete solution later?

akshayaurora:
I can’t honestly count this as a stepping stone or a stop gap solution,  you see currently.  This is what happens. 

If you start on a window with high res dpi you get this 

Screenshot 2020-06-25 at 10 31 13 PM

Moving  to lowers dpi gives you this

Screenshot 2020-06-25 at 10 27 10 PM

This is with master. Issue in the second sshot is the metrics don’t change so ui is not adjusted.

Now going reverse from low dpi to high dpi 

Starting on a low res display, you get this
Screenshot 2020-06-25 at 10 33 45 PM
 
Moving to a high res display you get this
Screenshot 2020-06-25 at 10 34 52 PM

With the change you should get something like
Screenshot 2020-06-25 at 10 37 19 PM
^ This result from my local branch with some similar testing

I guess you could call the last case a improvement. 

Is it acceptable stepping stone? From what I understand of what needs to be done with change in `reify` and windows and linux support. I would say no.

I wanted to check when the update_viewport get’s called.  It’s quite often https://youtu.be/LD2lG_ZGNHE

matham:
https://github.com/kivy/kivy/pull/7293 adds support for dynamic dpi/density updating. However, that PR does not add the changes to Window to reload the DPI.

So something like this is needed to recalculate the dpi. However, it needs to be done properly, like @akshayaurora suggested above:
>In light of this probably a better idea is to implement get_content_size abstract function in window.init and have it be overridden in specific implementations like window_sdl2 etc.
>
>Even if SDL2 does not directly support DPI on windows yet, we could always call direct WINAPI's and OSX API with if platform == checks. We do not have to depend on sdl2 to implement this considering we also have other window providers too.

AccelQuasarDragon:
Just want to drop a workaround on Apple MacBook Air M1:
```
def dpupdate(self, *args):
    arg1 = args[0]
    return dp(arg1)*(1/kivy.metrics.Metrics.density)
```
The idea behind dpupdate is that you know dp returns a value that is offset by a fixed ratio (and you can optionally check if you're on a mac using from sys import platform / platform == "darwin" )... With a higher pixel density, you draw a smaller text font, with less pixel density, you draw a larger text font. I was going to find a "sweet spot" fixed number but I got lucky with what I initially wrote.

However, I don't think that's the case with what's happening. kivy.metrics.Metrics.density for the M1 is 2.0, whereas on my PC it is 1.0. I think I just got lucky that dividing by 2 gets me a consistently nice font size. I suspect that sdl2 is already adjusting for the retina display, so kivy is trying to adjust an already fixed issue and accidentally makes text twice as large. I tried using using screen module with -m screen:one, -m :screen:onex, -m screen:note2 and they all look fine BUT I'm not sure that's a good substitute for using the actual hardware.

Hope this helps out someone with Mac M1 font issues, cheers

with regular dp (too large):

https://github.com/kivy/kivy/assets/138998466/3c18268a-563d-4449-a081-17de2d1f4876

with dpupdate (just right):

https://github.com/kivy/kivy/assets/138998466/c6d122c3-fddc-4beb-97cf-551fa11407dc

misl6:
Closing as dpi calculation on window update (e.g. moving a window from a screen(highdpi) to another(non-high-dpi)) is already supported since a while.

Additionally, we now migrated to SDL3 which introduces a lot of enhancements regarding high-dpi screens and windows.

Feel free to re-open if the issue persists on latest `master`

PR #5748, opened 2018-05-12
Add canvas argument to FloatLayout.add_widget

LogicalDash:
It's another PR just like #5715, this time for FloatLayout. It lets you pass the "canvas" argument to the add_widget method.

Not sure how I missed that. Sorry!

PR #6290, opened 2019-05-04
Widget: Fix signature of add/remove/clear_widget  to be consistent with base class

pythonic64:
Proposed changes are to bring consistency for arguments of methods:
- `add_widget`
- `remove_widget`
- `clear_widgets`

for all subclasses of `Widget` class. Classes affected with this change are:
- AccordionItem
- Accordion
- ActionGroup
- ActionOverflow
- ActionView
- ActionBar
- BoxLayout
- Bubble
- Carousel
- Dropdown
- EffectWidget
- FileChooser
- FloatLayout
- Layout
- Popup
- RecycleView
- ScatterLayout
- ScreenManager
- ScrollView
- ContentPanel
- SettingsItem
- Settings
- Splitter
- TabbedPanelItem
- TabbedPanel

### DIFF
--- a/kivy/uix/accordion.py
+++ b/kivy/uix/accordion.py
@@ -269,10 +269,10 @@ class AccordionItem(FloatLayout):
             return super(AccordionItem, self).add_widget(*args, **kwargs)
         return self.container.add_widget(*args, **kwargs)
 
-    def remove_widget(self, widget):
+    def remove_widget(self, *args, **kwargs):
         if self.container:
-            self.container.remove_widget(widget)
-        super(AccordionItem, self).remove_widget(widget)
+            self.container.remove_widget(*args, **kwargs)
+        super(AccordionItem, self).remove_widget(*args, **kwargs)
 
     def on_collapse(self, instance, value):
         accordion = self.accordion
--- a/kivy/uix/accordion.py
+++ b/kivy/uix/accordion.py
@@ -369,12 +369,11 @@ class Accordion(Widget):
         fbind('pos', update)
         fbind('min_space', update)
 
-    def add_widget(self, widget, index=0, canvas=None):
+    def add_widget(self, widget, *args, **kwargs):
         if not isinstance(widget, AccordionItem):
             raise AccordionException('Accordion accept only AccordionItem')
-
         widget.accordion = self
-        ret = super(Accordion, self).add_widget(widget, index, canvas)
+        ret = super(Accordion, self).add_widget(widget, *args, **kwargs)
         return ret
 
     def select(self, instance):
--- a/kivy/uix/accordion.py
+++ b/kivy/uix/accordion.py
@@ -266,12 +266,14 @@ class AccordionItem(FloatLayout):
 
     def add_widget(self, *args, **kwargs):
         if self.container is None:
-            return super(AccordionItem, self).add_widget(*args, **kwargs)
-        return self.container.add_widget(*args, **kwargs)
+            super(AccordionItem, self).add_widget(*args, **kwargs)
+            return
+        self.container.add_widget(*args, **kwargs)
 
     def remove_widget(self, *args, **kwargs):
         if self.container:
             self.container.remove_widget(*args, **kwargs)
+            return
         super(AccordionItem, self).remove_widget(*args, **kwargs)
 
     def on_collapse(self, instance, value):
@@ -373,8 +375,7 @@ class Accordion(Widget):
         if not isinstance(widget, AccordionItem):
             raise AccordionException('Accordion accept only AccordionItem')
         widget.accordion = self
-        ret = super(Accordion, self).add_widget(widget, *args, **kwargs)
-        return ret
+        super(Accordion, self).add_widget(widget, *args, **kwargs)
 
     def select(self, instance):
         if instance not in self.children:
--- a/kivy/uix/accordion.py
+++ b/kivy/uix/accordion.py
@@ -264,10 +264,10 @@ class AccordionItem(FloatLayout):
         fbind('title_args', trigger_title)
         trigger_title()
 
-    def add_widget(self, widget):
+    def add_widget(self, widget, index=0, canvas=None):
         if self.container is None:
             return super(AccordionItem, self).add_widget(widget)
-        return self.container.add_widget(widget)
+        return self.container.add_widget(widget, index, canvas)
 
     def remove_widget(self, widget):
         if self.container:
@@ -369,12 +369,12 @@ class Accordion(Widget):
         fbind('pos', update)
         fbind('min_space', update)
 
-    def add_widget(self, widget, *largs):
+    def add_widget(self, widget, index=0, canvas=None):
         if not isinstance(widget, AccordionItem):
             raise AccordionException('Accordion accept only AccordionItem')
 
         widget.accordion = self
-        ret = super(Accordion, self).add_widget(widget, *largs)
+        ret = super(Accordion, self).add_widget(widget, index, canvas)
         return ret
 
     def select(self, instance):
--- a/kivy/uix/accordion.py
+++ b/kivy/uix/accordion.py
@@ -264,10 +264,10 @@ class AccordionItem(FloatLayout):
         fbind('title_args', trigger_title)
         trigger_title()
 
-    def add_widget(self, widget, index=0, canvas=None):
+    def add_widget(self, *args, **kwargs):
         if self.container is None:
-            return super(AccordionItem, self).add_widget(widget)
-        return self.container.add_widget(widget, index, canvas)
+            return super(AccordionItem, self).add_widget(*args, **kwargs)
+        return self.container.add_widget(*args, **kwargs)
 
     def remove_widget(self, widget):
         if self.container:
--- a/kivy/uix/actionbar.py
+++ b/kivy/uix/actionbar.py
@@ -452,17 +452,17 @@ class ActionGroup(ActionItem, Button):
             # auto_dismiss applies to touching outside of the DropDown
             item.bind(on_release=ddn.dismiss)
 
-    def add_widget(self, item):
+    def add_widget(self, widget, index=0, canvas=None):
         # if adding ActionSeparator ('normal' mode,
         # everything visible), add it to the parent
-        if isinstance(item, ActionSeparator):
-            super(ActionGroup, self).add_widget(item)
+        if isinstance(widget, ActionSeparator):
+            super(ActionGroup, self).add_widget(widget, index, canvas)
             return
 
-        if not isinstance(item, ActionItem):
+        if not isinstance(widget, ActionItem):
             raise ActionBarException('ActionGroup only accepts ActionItem')
 
-        self.list_action_item.append(item)
+        self.list_action_item.append(widget)
 
     def show_group(self):
         # 'normal' mode, items can fit to the view
@@ -489,21 +489,21 @@ class ActionOverflow(ActionGroup):
     and defaults to 'atlas://data/images/defaulttheme/overflow'.
     '''
 
-    def add_widget(self, action_item, index=0):
-        if action_item is None:
+    def add_widget(self, widget, index=0, canvas=None):
+        if widget is None:
             return
 
-        if isinstance(action_item, ActionSeparator):
+        if isinstance(widget, ActionSeparator):
             return
 
-        if not isinstance(action_item, ActionItem):
+        if not isinstance(widget, ActionItem):
             raise ActionBarException('ActionView only accepts ActionItem'
-                                     ' (got {!r}'.format(action_item))
+                                     ' (got {!r}'.format(widget))
 
         else:
             if index == 0:
                 index = len(self._list_overflow_items)
-            self._list_overflow_items.insert(index, action_item)
+            self._list_overflow_items.insert(index, widget)
 
     def show_default_items(self, parent):
         # display overflow and its items if widget's directly added to it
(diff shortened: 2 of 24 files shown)


kivy/uix/accordion.py
```diff
@@ -264,10 +264,10 @@ def __init__(self, **kwargs):
         fbind('title_args', trigger_title)
         trigger_title()
 
-    def add_widget(self, widget):
+    def add_widget(self, widget, index=0, canvas=None):
```
matham: This should be `def add_widget(self, widget, *largs, **kwargs):` and `return self.container.add_widget(widget, *largs, **kwargs)`. Rather than explicitly listing all the args. 

The same for everywhere else in the PR.

pythonic64: What about `clear_widgets` and `remove_widget` method? 
What do you think about cases like `actionbar.py`, `bubble.py` and `splitter.py`?

Solution you're suggesting seems like a good half-way solution, but doesn't make headers consistent for all `Widget` subclasses and user still has to unpack arguments from `*args`. 
I'll update pr with your suggestion if you and other developers think that this is a better way to go.

matham: >Solution you're suggesting seems like a good half-way solution, but doesn't make headers consistent for all Widget subclasses and user still has to unpack arguments from *args.

I don't understand what you mean by this. Can you give more details or an example as to what you think the problem is here? The reason for not explicitly listing the arguments, is that as soon as something changes in the argument, like we gain a new argument, all these methods will break again. That's why simply catching unneeded positional and keyword argument and forwarding them to the base class is much preferable.

`remove_widget` takes no arguments, so it doesn't need `*largs, **kwargs`. But it also doesn't break anything and is future proof so it's better to not remove them if it already has it. Unless something is broken, it's best to not change it. `clear_widget` does take parameters, so in cases that overwrite it, it needs to be able to accept these additional arguments, because otherwise it is broken. And in those cases `*largs, **kwargs` is preferable than explicitly listing the arguments for the same reasons. Unless you need the actual argument of course.

pythonic64: Just the case of unpacking the arguments from `*largs`. When `index` and `canvas` are in `largs` then you do: 
```
index = largs[0]
canvas = largs[1]
```
So this adds two lines in method or one line 
```index, canvas = largs[0], largs[1]```.
and makes arguments position dependent. Also, it makes user to go up to base classes or to read the documentation on kivy.org to find which arguments are in the list when overriding method in more down-the-tree classes (like Splitter). 

1. What do you suggest for `add_widget` case in `splitter.py`? There `index` is declared in `add_widget` method. 
2. In `bubble.py` we have `*l` for `add_widget`. Should we go with `(widget, *largs, **kwargs)`? 
3. What about `actionbar.py` where `item` is used instead of `widget` and `screen.py` where `screen` is used?

> `remove_widget` takes no arguments, so it doesn't need `*largs, **kwargs`
Method `remove_widget` currently takes one argument `widget` and in some cases `*l` is used or `widget, *largs` or `widget, *largs, **kwargs`. 

Method `clear_widget` is defined is some cases with `clear_widgets(self, **kwargs)` or without any arguments, but its argument is `Widget` class is `children=None`. 

For all `remove_widget` occurrences we can do: 
`def remove_widget(self, widget, *largs, **kwargs)`
and for `clear_widgets` 
`def clear_widgets(self, children=None, *largs, **kwargs)`? 

Another thing, through repository `*args` is used more instead of `*largs`, by `*args` is used in 54 cases more then `*largs`. Should we use `*largs` or `*args`? I would go with `*args`.

**Review by tshirtman — DISMISSED** [review]
much needed cleanup indeed, but as discussed, i think it's better to do it more implicitly/automatically than this, leveraging argument unpacking.

File: kivy/uix/accordion.py (hunk 1 above)
Comment (tshirtman) [reply kivy/uix/accordion.py:267]: in case you are concerned about having to explicitly unpack arguments, this example shows that it's not required.

```py
class Something:
    def some_method(self, a, b, c=1, d=2):
        print(a, b, c, d)


class SomethingSpecific(Something):
    def some_method(self, a, *args, **kwargs):
        print("woo: {}".format(a))
        super().some_method(a, *args, **kwargs)


Something().some_method(1, 2, 3, 4)

SomethingSpecific().some_method(1, 2, 3, 4)
```
```
gabriel@gryphon:/tmp> python3 arguments.py
1 2 3 4
woo: 1
1 2 3 4
```

matham: @tshirtman wrote out pretty much exactly what I was thinking. The only additional point is that if the signature is `def func(a, b, c=True, d=False)` and the subclass uses only keys `a` and `c`. Then I'd write the subclass signature as `def func(a, b, c=True, **kwargs): super().func(a, b, c=c, **kwargs)` as opposed to `def func(a, *args, c=True, **kwargs): super().func(a, *args, c=c, **kwargs)`. Because the latter would fail if you call it like `func(1, 2, 3, 4)`.

kivy/uix/actionbar.py
```diff
@@ -452,17 +452,17 @@ def _toggle_dropdown(self, *largs):
             # auto_dismiss applies to touching outside of the DropDown
             item.bind(on_release=ddn.dismiss)
 
-    def add_widget(self, item):
+    def add_widget(self, widget, index=0, canvas=None):
```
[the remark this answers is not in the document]
matham: Yeah, agree it's better to simply fix the API and indicate in the docs that it changed. This is 2.0 after all anyway, and `item` was not the correct API anyway.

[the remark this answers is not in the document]
pythonic64: So this would be defined as:
`def add_widget(self, widget, *args, **kwargs):` 
with note in doc stating that `item` is renamed to `widget` to match `Widget.add_widget` method.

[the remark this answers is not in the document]
matham: Yeah!

kivy/uix/accordion.py
pythonic64: The case I had in mind with unpacking is if user wants to access `canvas` argument (or `index`) to do some checking and pass new values to super call or to use them in its own sub-class, but method signature of some `Widget` subclass is:
`def add_widget(self, widget, *args, **kwargs):` 

Then user have to do:
`canvas = args[1] if len(args) > 1 else None`
or
`index = args[0] if args else 0`
or change method signature to:
`def add_widget(self, widget, index=0, canvas=None, *args, **kwargs):`
All cases require of user to know position of these arguments in `args` and their default values. When overriding methods in sub-class editors like PyCharm will autocomplete signature from super-class, making user go back to add `index` or `canvas` if they are needed by sub-class. 

If in the future `Widget` or `Layout` or some other class adds a new argument, then user have to unpack it from `args`. On current changes we can add `**kwargs` making those three methods into:
- `def add_widget(self, widget, index=0, canvas=None, **kwargs):`
- `def remove_widget(self, widget, **kwargs):`
- `def clear_widgets(self, children=None, **kwargs):`

which enables us to use `kwargs` to pass/get all future arguments, enables user to pass its own arguments for its own subclasses and have same signature for all classes for which user can read in `Widget` class. To avoid collision of Kivy and user keys in `kwargs` user would have to use a prefix, like `user_`, for its keys.

By reading your suggestions, approach without `index` and `canvas` would make methods look like this:
- `def add_widget(self, widget, *args, **kwargs):`
- `def remove_widget(self, widget, *args, **kwargs):`
- `def clear_widgets(self, children=None, *args, **kwargs):`

for all classes. This includes classes with `index` or `canvas` already in signature but those arguments are not used in override method. This classes are: `BoxLayout`, `Carousel`, `FloatLayout`, `Layout`, `ScrollView`, `Splitter`, `TabbedPanel`.
For `Widget` class we would have:
- `def add_widget(self, widget, index=0, canvas=None, *args, **kwargs):`
- `def remove_widget(self, widget, *args, **kwargs):`
- `def clear_widgets(self, children=None, *args, **kwargs):`

For classes which use only `index`, like `ActionOverflow`, `ActionView`, we would have:
- `def add_widget(self, widget, index=0, *args, **kwargs):`
- or `index` gets unpacked in method's body with `index = args[0] if args else 0`

Whichever solution we settle on, errors like described in https://github.com/kivy/kivy/issues/6956, https://github.com/kivy/kivy/pull/5715, https://github.com/kivy/kivy/pull/5748 won't happen in the future.

tshirtman: > or change method signature to:
`def add_widget(self, widget, index=0, canvas=None, *args, **kwargs):`
All cases require of user to know position of these arguments in args and their default values. When overriding methods in sub-class editors like PyCharm will autocomplete signature from super-class, making user go back to add index or canvas if they are needed by sub-class.

Right, the user can indeed spell out these arguments in order if they want to access them in their overriding of the method, and yes, they have to check the signatures of parent methods for that. I don't think the fact that an IDE tries to be helpful and doesn't come up with the exact argument list the user wants should drive our decision, if we go this way, then we'll just spell out everything everywhere, so such generated code won't ever have the `*args` and `**kwargs` argument (unless we eat them up in `Widget.add_widget`, but that's bad practice), and then if we ever add a new argument to the signature of `Widget.add_widget` then the user's generated code will simply break (unless they manually added `*args` and `**kwargs` defensively, to the code generated by pycharm).

I don't think arguments should get unpacked in the body, when needed, unpacking them in the signature is the proper approach. When they are currently unpacked in the signature, but not used, i would advise removing them from the signature to be handled by *args/**kwargs, and passed to the parent method if this one is called, if the parent method is not called, *args and **kwargs shouldn't be accepted, as the extra arguments are not going to be used by anything.

pythonic64:
@matham I made the changes to `add_widget` methods. Can you advise on what to do with `remove_widget` and `clear_widgets` methods? 

Possible options for `remove_widget`:
1) (*args, **kwargs)
2) (widget, *args, **kwargs)
3) (widget)

If argument `widget` is used in the body then it stays in the header. Same goes for `clear_widgets` with `children=None` instead of `widget` argument. 

My opinion is to use 1) for all cases expect for cases where first argument is used in the body and for that we use 2) option.

kivy/uix/bubble.py
```diff
@@ -303,7 +301,7 @@ def on_arrow_pos(self, *l):
         self_arrow_layout.clear_widgets()
         self_arrow_img = self._arrow_img
         self._sctr = self._arrow_img
-        self.clear_widgets(do_super=True)
+        self._super_clear_widgets()
```
matham, line 304:
```suggestion
        super(Bubble, self).clear_widgets()
```
Calling super directly should work.

kivy/uix/carousel.py
```diff
@@ -664,12 +664,16 @@ def remove_widget(self, widget, *args, **kwargs):
...
-
-    def clear_widgets(self):
-        for slide in self.slides[:]:
-            self.remove_widget(slide)
+            container.remove_widget(widget)
+        super(Carousel, self).remove_widget(widget)
+
+    def clear_widgets(self, children=None):
+        # `children` must be a list of slides or None
+        if not children:
```
matham, line 672:
```suggestion
        if children is None:
```

kivy/uix/screenmanager.py
```diff
@@ -978,53 +978,60 @@ def _screen_name_changed(self, screen, name):
...
-        # iterate over a copy of screens, as self.remove_widget
-        # modifies self.screens in place
-        for screen in screens[:]:
-            self.remove_widget(screen)
+    def clear_widgets(self, children=None):
+        '''
+        .. versionchanged:: 2.1.0
+            Renamed argument `screens` to `children`.
+        '''
+        if not children:
```
matham, line 1028:
```suggestion
        if children is None:
```

kivy/uix/tabbedpanel.py
```diff
@@ -576,14 +576,12 @@ def remove_widget(self, widget):
...
-        if kwargs.get('do_super', False):
-            super(TabbedPanel, self).clear_widgets()
-        else:
-            content.clear_widgets()
+    def clear_widgets(self, children=None):
+        if self.content:
+            self.content.clear_widgets(children)
+
+    def _super_clear_widgets(self):
+        super(TabbedPanel, self).clear_widgets()
```
matham, line 584:
```suggestion
```

kivy/uix/tabbedpanel.py
```diff
@@ -706,7 +704,7 @@ def _update_tabs(self, *l):
         tabs.bind(width=self._partial_update_scrollview)
 
         # remove all widgets from the tab_strip
-        self.clear_widgets(do_super=True)
+        self._super_clear_widgets()
```
matham, line 707:
```suggestion
        super(TabbedPanel, self).clear_widgets()
```

kivy/uix/bubble.py
```diff
@@ -210,34 +210,32 @@ def __init__(self, **kwargs):
...
-        if kwargs.get('do_super', False):
-            super(Bubble, self).clear_widgets()
-        else:
-            content.clear_widgets()
+    def clear_widgets(self, children=None):
+        if self.content:
+            self.content.clear_widgets(children)
+
+    def _super_clear_widgets(self):
+        super(Bubble, self).clear_widgets()
```
matham, line 238:
```suggestion
```

matham (changes requested):

matham:
About `remove_widget` and `clear_widgets`, I agree with what you said (use 1 or 2 as needed).

However, it's up to you if you want to do that because currently it should be fine as is since there's no additional args and it's probably annoying to do these fixes. But ideally these methods should have `*args`, `*kwargs` because any inherited method should probably do that to be safe so go ahead if you want to fix them.

pythonic64:
Can you suggest what to do about `add_widget` and `remove_widget` in `WindowBase` class?

pythonic64:
And should we add `*args, **kwargs` to Widget class?

matham:
>Can you suggest what to do about add_widget and remove_widget in WindowBase class?

It would be nice if it's `add_widget` also accepted a index parameter to be consistent with Widget. And the same for a `clear_widgets` method. However, adding that would require adding tests and should be in another PR.

>And should we add *args, **kwargs to Widget class?

I don't think so. The reason for doing it in the child class is to not have to worry if the super class signature changes and to prevent the subclass signature from missing args or being incorrect. The Widget signature though is part of the API and there's no reason to add the ambiguity with `args`/`kwargs`. In fact, it's better to be concrete so the accepted args is clear in the API. And any future signature change would be considered a API change/break.

matham (approved):

matham:
Can you also please edit the PR comment to list all the classes whose signature changed (e.g. `action_item` was renamed to `widget`). And perhaps also add a doc comment on that class (`versionchanged`) that it was changed. To help people understand what they may need to fix.

pythonic64:
Do you mean adding:
```
.. versionchanged:: 2.10
    Signature changed from <old> to <new>.
```

for all `add_widget`, `remove_widget`, `clear_widgets` methods?

matham:
Yeah, but not for all of them, only the few for whom you renamed the variable from something other than widget to widget.

And you can also put it in the class or module docs if the method doesn't have docs.

pythonic64:
That's only in `actionbar.py` and `screenmanager.py` modules and `versionchanged` tag is added there.

There are changes from `(widget)` to `(*args, **kwargs)` but those won't be codebreaking.

matham:
Perfect, thanks for all your work!

pythonic64:
You're welcome!

PR #6316, opened 2019-05-18
Update doc of AliasProperty

pythonic64:
Improved doc of `AliasProperty` better explains `setter` method and `bind` argument.

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -1417,14 +1417,18 @@ cdef class AliasProperty(Property):
             self.x = value - self.width
         right = AliasProperty(get_right, set_right, bind=['x', 'width'])
 
-    If `x` were to be an instance level attribute and not Kivy property then
-    you have to return `True` from setter to dispatch value of `right`::
+    If `x` were a non Kivy property then you have to return `True` from setter
+    to dispatch new value of `right`::
 
         def set_right(self, value):
             self.x = value - self.width
             return True
 
-    If your want to cache the value returned by getter then pass `cache=True`.
+    Usually `bind` list should contain all Kivy properties used in getter
+    method. If you're setting a property from `bind` list then you shouldn't
+    `return True` in setter as that will dispatch new value twice.
+
+    If you want to cache the value returned by getter then pass `cache=True`.
     This way getter will only be called if new value is set or one of the
     binded properties changes. In both cases new value of alias property will
     be cached again.
@@ -1438,15 +1442,11 @@ cdef class AliasProperty(Property):
         `getter`: function
             Function to use as a property getter.
         `setter`: function
-            Function to use as a property setter. Callbacks bound to the
-            alias property won't be called when the property is set (e.g.
-            `right = 10`), unless the setter returns `True`.
+            Function to use as a property setter.
         `bind`: list/tuple
-            Properties to observe for changes, as property name strings.
-            Changing values of this properties will dispatch value of the
-            alias property.
+            Properties to observe for changes as property name strings.
         `cache`: boolean
-            If `True`, the value will be cached, until one of the binded
+            If `True`, the value will be cached until one of the binded
             elements changes or if setter returns `True`.
         `rebind`: bool, defaults to `False`
             See :class:`ObjectProperty` for details.
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -1425,8 +1425,11 @@ cdef class AliasProperty(Property):
             return True
 
     Usually `bind` list should contain all Kivy properties used in getter
-    method. If you're setting a property from `bind` list then you shouldn't
-    `return True` in setter as that will dispatch new value twice.
+    method. If you return `True` it will cause a dispatch which one should do
+    when the property value has changed, but keep in mind that the property
+    could already have dispatched the changed value if a kivy property the
+    alias property is bound was set in the setter, causing a second dispatch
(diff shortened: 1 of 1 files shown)


kivy/properties.pyx
```diff
@@ -1438,15 +1442,11 @@ cdef class AliasProperty(Property):
         `getter`: function
             Function to use as a property getter.
         `setter`: function
-            Function to use as a property setter. Callbacks bound to the
-            alias property won't be called when the property is set (e.g.
-            `right = 10`), unless the setter returns `True`.
+            Function to use as a property setter.
```
matham: I'm not sure why you removed this?

kivy/properties.pyx
```diff
@@ -1438,15 +1442,11 @@ cdef class AliasProperty(Property):
...
         `bind`: list/tuple
-            Properties to observe for changes, as property name strings.
-            Changing values of this properties will dispatch value of the
-            alias property.
+            Properties to observe for changes as property name strings.
```
matham: I'm not sure why you removed this either?

kivy/properties.pyx
```diff
@@ -1417,14 +1417,18 @@ cdef class AliasProperty(Property):
...
+    to dispatch new value of `right`::
 
         def set_right(self, value):
             self.x = value - self.width
             return True
 
-    If your want to cache the value returned by getter then pass `cache=True`.
+    Usually `bind` list should contain all Kivy properties used in getter
+    method. If you're setting a property from `bind` list then you shouldn't
+    `return True` in setter as that will dispatch new value twice.
```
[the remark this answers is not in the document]
pythonic64: There are so many cases with `AliasProperty`. 
If you swap `x` and `_value` lines in setter: 
```
def setter(self, value):
    self._value = value
    self.x = value * 25
    return True
```

and still `return True` then `prop` value will be dispatched twice with same value, first when `x` is set and again when `True` is returned. 

Do you have idea on how to better phrase that paragraph so that is clear when user should `return True`? I'm right now out of ideas :)

kivy/properties.pyx
pythonic64: Removed sentence is partially true, because callbacks will be also dispatched when some of binded properties changes and getter returns new value. 
I thought that it would be better to explain in text above how setter should work and don't repeat same thing in `Parameters` section.

kivy/properties.pyx
pythonic64: Same as the answer above. That sentence is partially true as new value will be dispatched only when binded property changes and getter returns new value. 
If getter is doing something like clamping (`min(max(x, min_value), max_value)`) and returns the same value then property won't get dispatched.

kivy/properties.pyx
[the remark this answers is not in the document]
pythonic64: Maybe we can update `set` method of `AliasProperty` from this: 
```
cpdef set(self, EventDispatcher obj, value):
    cdef PropertyStorage ps = obj.__storage[self._name]
    if ps.setter(obj, value):
        if self.use_cache:
            if ps.alias_initial:
                ps.alias_initial = 0
            ps.value = ps.getter(obj)
        self.dispatch(obj)
    elif self.force_dispatch:
        self.dispatch(obj)
```
to this: 
```python
cpdef set(self, EventDispatcher obj, value):
    cdef PropertyStorage ps = obj.__storage[self._name]
    if ps.setter(obj, value):
        cdef int changed = 1
        if self.use_cache:
            changed = 0
            dvalue = ps.getter(obj)
            if dvalue != ps.value:
                if ps.alias_initial:
                    ps.alias_initial = 0
                ps.value = dvalue
                changed = 1
        if changed:
            self.dispatch(obj)
    elif self.force_dispatch:
        self.dispatch(obj)
```

Then we can tell the user to `return True` in setter to **accept** value, but value will be dispatched only if it has changed.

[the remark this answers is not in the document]
matham: You are right that it would be dispatched twice, but I think it's ok. It's better to dispatch multiple times (after all that's how it would generally work), then not dispatch when needed. Kivy doesn't really make a guarantee that properties won't dispatch even if the value is the same (although we do try to prevent that if possible). We just promise that properties will definitely dispatch if they are changed.

As you say, it's hard to say something meaningful here in the docs because we could have many different situations. That's why I think saying something like, `if you return True it will cause a dispatch which one should do to when the property value has changed, but keep in mind that the property could already have dispatched the changed value if a kivy property the alias property is bound was set in the setter, causing a second dispatch if the setter returns True.` is better. Because this makes no recommendation, but just describes the behavior.

We can definitely change `set` to what you suggest, but that would change the interpretation of returning True from meaning we should dispatch to mean that the value may have changed so maybe we should dispatch. This change may be a compatibility break and I'm not sure it is worth to do unless this comes up as an actual issue.

kivy/properties.pyx
matham: I think it is better to give a very short summary here, even if it was already explained earlier because sometimes I don't get the full implication from the example and I want a short summary to make sure I understood the implications. Repeating it twice gives me chance to make sure I understood by comparing what I understand the two places to say.

So I'd say it's good to fix it to give a very short and concise summary here and for `bind`.

kivy/properties.pyx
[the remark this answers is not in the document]
pythonic64: Ok, I'll update pr with your suggestions.

PR #6339, opened 2019-05-29
Fix PermissionError when reconnecting MTDEV input device

Cheaterman:


tshirtman:
Seems like this will create a busy loop for a corner case, to avoid an error in the log, is there a reason to try again when there was a permission error? is there a relevant chance that next try will work?

Cheaterman:
Definitely fixes the issue, but I did notice the potential infinite loop
there - I think there should be an easy fix, either by attempting a fixed
amount of times ("attempts = 0" above the "try:" block, and "attempts += 1;
if attempts >= max_attempts: raise" in the "except:" block).

If you want to reproduce, it should be as easy as disconnecting and
reconnecting the USB cable of a MTDEV device - udev will make the input
device ("/dev/input/event*") owned by "root:root" for a brief instant
(before it changes its group to "input"), causing the "PermissionError" -
this also explains why retrying fixes the issue in my case.

For what it's worth, it looks to me like MTDEV is supposed to be able to
manage disconnects/reconnects already, and the only obstacle is this
"PermissionError" - I get "device disconnected" and "device reconnected" in
the Kivy logs after fixing the "PermissionError".

Thanks for your attention :-)

Le dim. 2 juin 2019 à 17:22, Gabriel Pettier <notifications@github.com> a
écrit :

akshayaurora:
closed in #7587

Issue #6338, opened 2019-05-28
Switching a screen with ShaderTransition removes previous screen


gottadiveintopython:
### Versions

* Python: 3.7.1
* OS: LinuxMint 18.2
* Kivy: master branch(v1.11.0rc1, git-e8a4202, 20190528)
* Kivy installation method: `pip install -e .`

### Description

Switching a screen with `ShaderTransition`(and its subclass) removes previous screen from `ScreenManager`. But this issue doesn't occure if revert #6279.

### Code and Logs (EDIT: May 30)

The following code:

```python
import kivy.uix.screenmanager as sm

TRANSITION_CLASSES = tuple(
    getattr(sm, name) for name in sm.__all__ if name.endswith('Transition'))


for TRANSITION_CLASS in TRANSITION_CLASSES:
    scrmgr = sm.ScreenManager()
    for i in range(3):
        scrmgr.add_widget(sm.Screen(name=str(i)))

    print(TRANSITION_CLASS.__name__)
    print('Is ShaderTransition?:',
          issubclass(TRANSITION_CLASS, sm.ShaderTransition))
    print(scrmgr.screen_names)

    scrmgr.transition = TRANSITION_CLASS()
    scrmgr.current = '1'

    print(scrmgr.screen_names)

    scrmgr.transition = TRANSITION_CLASS()
    scrmgr.current = '2'

    print(scrmgr.screen_names)
    print('\n')
```

prints:

```
ShaderTransition
Is ShaderTransition?: True
['0', '1', '2']
['1', '2']
['2']


SlideTransition
Is ShaderTransition?: False
['0', '1', '2']
['0', '1', '2']
['0', '1', '2']


SwapTransition
Is ShaderTransition?: False
['0', '1', '2']
['0', '1', '2']
['0', '1', '2']


FadeTransition
Is ShaderTransition?: True
['0', '1', '2']
['1', '2']
['2']


WipeTransition
Is ShaderTransition?: True
['0', '1', '2']
['1', '2']
['2']


FallOutTransition
Is ShaderTransition?: True
['0', '1', '2']
['1', '2']
['2']


RiseInTransition
Is ShaderTransition?: True
['0', '1', '2']
['1', '2']
['2']


NoTransition
Is ShaderTransition?: False
['0', '1', '2']
['0', '1', '2']
['0', '1', '2']


CardTransition
Is ShaderTransition?: False
['0', '1', '2']
['0', '1', '2']
['0', '1', '2']
```

[...]

PR #6347, opened 2019-06-01
add tests for #6338

gottadiveintopython:
#6338


matham:
The tests seem to be failing on python 2 though.

gottadiveintopython:
Thanks for the advice! I'll fix that part (`TRANSITION_CLASS_NAMES`) if I become more familiar with `pytest` and find a better way.

matham:
Thanks!

Issue #5560, opened 2018-01-04
Backwards-incompatible changes - SUBSCRIBE TO THIS THREAD if you use Kivy!


matham:
Stability is great for users! It lets them focus on solving their problem without worrying about their platform shifting under their feet. But stability is also bad for users! It means that anywhere an API is error-prone or hard to use or missing important functionality, they're stuck with it. Making the optimal trade-off here is tricky and context-dependent.

Kivy is an active project. So as we build things with it we'll probably discover places where the API is less awesome than desired. Hopefully we won't discover any major issues, but you never know.

This means that if you're a Kivy user, it'd be good to have some strategy to make this as painless as possible. Our suggestions:

* Pin your version. For example, in your install_requires= or requirements.txt, do not write: kivy >= 1.8.0. Instead, write: kivy ~= 1.8.0.
  * You can also use == if you prefer. The difference is that while both == 1.8.0 and ~= 1.8.0 will disallow upgrading to 1.9.0, ~= allows upgrading to 1.8.1 but == 1.8.0 does not. Our intention is that 0.x.y and 0.x.(y+1) will be backwards compatible.
* **Subscribe to this issue** (for example, by pressing the little "Subscribe" button in the right column →). We'll link to backwards-incompatible changes here before we make them, so this will give you fair warning and a chance to give feedback on the original issue.

If you would like to propose a backwards incompatible change to Kivy, please add the `api-break` label to your issue (or request that we add it), and a link to the issue will automatically be posted here.

Zen-CODE:
This PR would probably also classify as a breaking change?

https://github.com/kivy/kivy/pull/6095

matham:
We are moving our kivy dependencies on windows from `kivy.deps.xxx` stored in `kivy/deps/xxx` to `kivy_deps.xxx`, stored under `kivy_deps/xxx`. To prevent any issues, please always pin your dependency versions when you install them from pip to the version on pypi at the time of the release you're using.

matham:
Python 2 support has been dropped from master https://github.com/kivy/kivy/pull/6351.

matham:
I'm closing this since we don't really use it. Instead you should look under the `Breaking changes` section of the release notes for each release. Or, look at the [API-break](https://github.com/kivy/kivy/pulls?q=is%3Apr+label%3A%22Notes%3A+API-break%22+) label to see PRs that were an API break.

PR #6351, opened 2019-06-01
Drop Python 2 support.

matham:
Drops Python 2 support from the codebase.

I didn't remove every reference to the `PY2` variable, only those used in packaging and a few core areas. The remaining areas where it is used can be removed later because I didn't want to create new bugs by mass changing code that checks `PY2`.


Cheaterman:
I didn't realize this was a thing. I was reading code last week, and noticed several places where there's still Py2-compat code, and I was thinking of excising it myself (because 2020 you know...). It's not explicitly using PY2 variable, rather some more subtle things about unicode strings vs bytestrings (the usual deal...). I'll try to find it again, and make a PR.

On the same topic, what are the current target versions for Kivy? Is there a plan to make it 3.6+-only so we can use f-strings and the likes?

I think your warning to users in https://github.com/kivy/kivy/issues/5560 is great @matham , forcing them to pin target version is indeed beneficial to everyone - we can be more bold in our changes, and they can be mostly certain their apps won't break.
In the same mindset, wouldn't it be fantastic if `kivy.require()` function (you know, the one we all SHOULD use, and never do?) could specify a maximum/specific target version rather than specifying only the minimum?

matham:
I believe we already only support 3.6+ in master. I'm not sure if there's a single place we say this though? I suppose we need to update [setup.py](https://github.com/kivy/kivy/blob/master/setup.py#L1135).

About `kivy.require()` I'm not actually into it. Why duplicate python's built-in package requirements (in setup.py) into kivy? We should aim to re-use python eco-system standards rather than creating our own when possible.

Cheaterman:
Yeah, I guess that's a fair point! As long as we clarify which versions we
support and stick to them, that will be fine :)

PR #6352, opened 2019-06-02
force to use sphinx 1.7.9 to restore search

tshirtman:
maybe there is a better fix, but seems our search has been broken by updates to sphinx, tried 2.1, 2.0 then 1.8.5 localy, none of them worked, falling back to 1.7.9 did, so falling back to that for now.


matham:
I'm not sure the requirements file is used on Travis?

tshirtman:
ah, indeed https://github.com/kivy/kivy/blob/c9157d3f898469cdf6d3414ea387d3c3c84e0950/.travis.yml#L89

PR #6357, opened 2019-06-03
Merge some fixes into stable

matham:
It also bumps stable to 1.11.0.post0 so we can upload the fixes to pypi.

### DIFF
--- a/kivy/tools/packaging/pyinstaller_hooks/pyi_rth_kivy.py
+++ b/kivy/tools/packaging/pyinstaller_hooks/pyi_rth_kivy.py
@@ -5,7 +5,8 @@ root = os.path.join(sys._MEIPASS, 'kivy_install')
 
 os.environ['KIVY_DATA_DIR'] = os.path.join(root, 'data')
 os.environ['KIVY_MODULES_DIR'] = os.path.join(root, 'modules')
-os.environ['GST_PLUGIN_PATH'] = os.path.join(sys._MEIPASS, 'gst-plugins')
+os.environ['GST_PLUGIN_PATH'] = '{}{}{}'.format(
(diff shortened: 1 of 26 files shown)


matham:
Don't merge yet. We need to include a fix for https://github.com/kivy/kivy/issues/6358.

matham:
Also https://github.com/kivy/kivy/pull/6362.

kivy/tools/packaging/pyinstaller_hooks/pyi_rth_kivy.py
```diff
@@ -5,7 +5,8 @@
 
 os.environ['KIVY_DATA_DIR'] = os.path.join(root, 'data')
 os.environ['KIVY_MODULES_DIR'] = os.path.join(root, 'modules')
-os.environ['GST_PLUGIN_PATH'] = os.path.join(sys._MEIPASS, 'gst-plugins')
+os.environ['GST_PLUGIN_PATH'] = '{}{}{}'.format(
+    sys._MEIPASS, os.pathsep, os.path.join(sys._MEIPASS, 'gst-plugins'))
```
tito: Why X/(X/Y) ? (X=_MEIPASS, Y=gst-plugins) ?

matham: `os.pathsep` is not the directory join (`/`), this is `X;X/Y`. because both directories are on the path. On windows, the plugins are in the root.

matham: I.e. on windows the plugins are in the same directory as the main binaries. We don't separate them out. So pyinstaller puts all the deps binaries in `X`, so it should be in the plugin path.

PR #6372, opened 2019-06-07
make 'Carousel.remove_widget()' not cause 'IndexError'

### LINKED ISSUE #6370: 'Carousel.remove_widget()' causes IndexError when the last slide is active, and the error is not delivered to a user
### Versions

* Python: 3.7.1
* OS: Linux Mint 18.2
* Kivy:  v2.0.0.dev0, git-e828f22
* Kivy installation method: `git clone` then `pip -e`

### Description

When a `Carousel` meets the following conditions:

- having 2 or more slides (`len(carousel.slides) >= 2`)
- the last slide is active  (`carousel.index == len(carousel.slides) - 1`)

removing any slide causes `IndexError`. The error is shown in the strerr, but can't be catched.

### Code and Logs

```python
from kivy.app import App
from kivy.clock import Clock
import random


N_SLIDES = random.choice(range(2, 4))  # Has to be 2 or greater

class SampleApp(App):
    def build(self):
        c = Factory.Carousel()
        for i in range(N_SLIDES):
            c.add_widget(Factory.Label(text=str(i), font_size=100))
        c.index = N_SLIDES - 1  # switch to the last slide
        return c
    
    def on_start(self):
        c = self.root
        Clock.schedule_once(
            lambda __: c.remove_widget(random.choice(c.slides)),  # remove any slide
            c.anim_move_duration + 1)


if __name__ == '__main__':
    SampleApp().run()
```

The following errors are printed, but the app still is running and I can switch to other slide.

```
   File "kivy/properties.pyx", line 599, in kivy.properties.Property.dispatch
     ps.observers.dispatch(obj, ps.value, None, None, 0)
     result = self._dispatch(
     return f(obj, value)
   File "kivy/properties.pyx", line 1503, in kivy.properties.AliasProperty.trigger_change
     cpdef trigger_change(self, EventDispatcher obj, value):
   File "kivy/properties.pyx", line 1505, in kivy.properties.AliasProperty.trigger_change
     dvalue = ps.getter(obj)
   File "/tmp/firefox/kivy/kivy/uix/carousel.py", line 194, in _curr_slide
     return self.slides[self.index or 0]
 IndexError: list index out of range
 Exception ignored in: 'kivy.properties.observable_list_dispatch'
   File "kivy/properties.pyx", line 599, in kivy.properties.Property.dispatch
     ps.observers.dispatch(obj, ps.value, None, None, 0)
     result = self._dispatch(
     return f(obj, value)
   File "kivy/properties.pyx", line 1503, in kivy.properties.AliasProperty.trigger_change
     cpdef trigger_change(self, EventDispatcher obj, value):
   File "kivy/properties.pyx", line 1505, in kivy.properties.AliasProperty.trigger_change
     dvalue = ps.getter(obj)
   File "/tmp/firefox/kivy/kivy/uix/carousel.py", line 194, in _curr_slide
     return self.slides[self.index or 0]
 IndexError: list index out of range
```

#### pytest

```python
import pytest


@pytest.fixture(
    scope='session', params=(True, False), ids=lambda v: 'loop=' + str(v))
def loop(request):
    return request.param


@pytest.fixture(
    scope='session', params=tuple(range(5)),
    ids=lambda v: 'n_slides=' + str(v))
def n_slides(request):
    return request.param


@pytest.fixture
def carousel(loop, n_slides):
    from kivy.uix.carousel import Carousel
    c = Carousel(loop=loop)
    for i in range(n_slides):
        c.add_widget(Widget())
    return c


def test_remove_a_slide_when_the_last_slide_is_active(capsys, carousel):
    import random
    if len(carousel.slides) == 0:
        pytest.skip()
    carousel.index = len(carousel.slides) - 1
    carousel.remove_widget(random.choice(carousel.slides))
    stderr_text = capsys.readouterr().err
    for line in stderr_text.split('\n'):
        assert not line.startswith('IndexError:')


if __name__ == "__main__":
    pytest.main(args=[
        __file__,
    ])
```

gottadiveintopython:
fix #6370

### DIFF
--- a/kivy/tests/test_uix_carousel.py
+++ b/kivy/tests/test_uix_carousel.py
@@ -1,15 +1,60 @@
 import pytest
 
 
-def test_remove_widget():
+@pytest.fixture(
+    scope='session', params=(True, False), ids=lambda v: 'loop=' + str(v))
+def loop(request):
+    return request.param
+
+
+def test_remove_widget(loop):
     from kivy.uix.carousel import Carousel
     from kivy.uix.widget import Widget
 
-    c = Carousel()
-    w = Widget()
-    c.add_widget(w)
-    assert len(c.children) == 1
-    assert len(c.slides) == 1
-    c.remove_widget(w)
+    c = Carousel(loop=loop)
+    assert c.index is None
+    assert c.current_slide is None
     assert len(c.children) == 0
     assert len(c.slides) == 0
+
+    N_SLIDES = 4
+    for i in range(N_SLIDES):
+        c.add_widget(Widget())
+    assert c.index == 0
+    assert c.current_slide == c.slides[0]
+    assert len(c.children) == 3 if loop else 2
+    assert len(c.slides) == N_SLIDES
+
+    # test issue #6370
+    c.index = len(c.slides) - 1
+    c.remove_widget(c.slides[0])
+
+    # remove a slide(smaller index than the current_slide's).
+    c.index = 1
+    old_current_slide = c.current_slide
+    c.remove_widget(c.slides[c.index - 1])
+    assert c.index == 0
+    assert c.current_slide is old_current_slide
+    assert len(c.children) == 2
+    assert len(c.slides) == 2
+
+    # remove a slide(bigger index than the current_slide's).
+    old_current_slide = c.current_slide
+    c.remove_widget(c.slides[c.index + 1])
+    assert c.index == 0
+    assert c.current_slide is old_current_slide
+    assert len(c.children) == 1
+    assert len(c.slides) == 1
+
+    # remove the current_slide(the last one left).
+    c.remove_widget(c.current_slide)
+    assert c.index is None
+    assert c.current_slide is None
+    assert len(c.children) == 0
+    assert len(c.slides) == 0
+
+
+if __name__ == "__main__":
+    pytest.main(args=[
+        __file__,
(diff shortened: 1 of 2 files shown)


kivy/tests/test_uix_carousel.py
```diff
@@ -1,15 +1,73 @@
...
+    c.remove_widget(c.current_slide)
+    assert c.index is None
+    assert c.current_slide is None
     assert len(c.children) == 0
     assert len(c.slides) == 0
+
+    # check if there is no `IndexError` in stderr. See issue #6370.
+    stderr_text = capsys.readouterr().err
+    for line in stderr_text.split('\n'):
+        assert not line.startswith('IndexError')
```
matham: https://github.com/kivy/kivy/pull/6373 will fix so that it'll actually raise an exception and so that you won't need to hunt in the sdterr. Can you rebase on that once I merge it (when the my PR's test pass)?

Additionally, for the future, instead of `startswith` which is a little unreliable in case there are beginning spaces etc., using a regex match which is tolerant to that would have been better.

gottadiveintopython: Thank you. I updated the PR.

> Additionally, for the future, instead of `startswith`...

Yes I agree with that. I was writing `stderr_text.find('IndexError')` at first, but it was too slow. That's why.

Issue #1763, opened 2014-01-14
Carousel does't change index


glebpushkov:
Hello. I faced with problem, when Carousel does't go to the next slide because index wasn't changed. I looked at on__offset method and found next:

``` python
if direction[0] == 'r':
    if _offset <= -width:
        index += 1
    if _offset >= width:
        index -= 1
```

in my case values was:
(Pdb) _offset
-85.93333333333334
(Pdb) -width
-728.7500000000001
##

gottadiveintopython:
On my desktop, this issue happens only when `loop==True` and `len(slides) == 2`. For example a code like this:

```python
from kivy.factory import Factory
from kivy.app import runTouchApp

carousel = Factory.Carousel(direction='left', loop=True)
carousel.bind(
    index=lambda __, value: print('index:', value),
    current_slide=lambda __, value: print('current_slide:', value),
)
for i in range(2):
    l = Factory.Label(text=str(i), font_size=100)
    carousel.add_widget(l)

runTouchApp(carousel)
```

You can confirm:

1. Sometimes `index` and `current_slide` aren't updated even though the next slide is actually displayed.
1. The loop doesn't work. You can't swipe infinity.

And if you look at [carousel.py#L167-L171](https://github.com/kivy/kivy/blob/master/kivy/uix/carousel.py#L167-L171) and [carousel.py#L211-L215](https://github.com/kivy/kivy/blob/master/kivy/uix/carousel.py#L211-L215), you can confirm both are ignoring `loop` when `len(self.slides) == 2`.

If those 2 parts are removed, the issue doesn't occur. But another issue(the next slide is not visible until the switching is done) will be occur instead. So I think it's not easy to fix this issue.

### Versions

- Python 3.7.1
- Kivy v2.0.0.dev0, git-e828f22

gottadiveintopython:
I kinda gave up fixing this issue, and just leave a code that might be helpful to test `Carousel`.

```python
from kivy.app import runTouchApp
from kivy.lang import Builder


KV_CODE = '''
<DebugCarousel@Carousel>:
    canvas.before:
        Clear:  # clear instructions inherited from StencilView
    canvas.after:
        Clear:  # clear instructions inherited from StencilView
        Color:
            rgba: 1, 1, 1, 1
        Line:  # draw border
            rectangle: self.x + 1, self.y + 1, self.width - 1, self.height - 1
            dash_offset: 5
            dash_length: 3

FloatLayout:
    DebugCarousel:
[...]

PR #6374, opened 2019-06-09
Make 'Carousel' able to handle the case where 'loop == True' and 'len(slides) == 2'

gottadiveintopython:
When a `Carousel` meets the following conditions:

- `carousel.loop == True`
- `len(carousel.slides) == 2`

it doesn't work properly, e.g. like [this](https://github.com/kivy/kivy/issues/1763#issuecomment-498751560). This is probably because the next slide and the previous slide are the same under that conditions. 

The tests for this PR aren't fully written because I don't have enough knowledge to automate GUI. Instead I'm using [this](https://gist.github.com/gottadiveintopython/15e4565269c5268e739f9697cbfaefe9), and testing manually.


kivy/uix/carousel.py
```diff
@@ -284,6 +274,8 @@ def _next_slide(self):
     _next = ObjectProperty(None, allownone=True)
     _offset = NumericProperty(0)
     _touch = ObjectProperty(None, allownone=True)
+    _prev_equals_next = BooleanProperty(False)
+    _prioritize_next = BooleanProperty(False)
```
matham: Unless the properties are needed for the users to be able to bind to, or it really simplifies the code, it is better to use a `@property` or normal python property rather than a KivyProperty.

That's because with Kivy properties it is harder to trace the code logic (due to action at a distance), and they add loading performance cost to the widgets.

So for example, `_prev_equals_next` doesn't seem to benefit at all from being a KivyProperty, so it should just be a normal property. But even `_prioritize_next`, I would make it a normal property and inline the code in `on__prioritize_next` at the one place you set it. Or you can make it a python `@property` with a setter that does the stuff currently in `on__prioritize_next` when set.

kivy/uix/carousel.py
```diff
@@ -446,12 +450,19 @@ def on_index(self, *args):
         self._trigger_position_visible_slides()
         self._offset = 0
 
+    def on_loop(self, *args):
```
matham: It is generally better to directly `fbind` `_insert_visible_slides` to `loop` in the `__init__` method, than to use the `on_loop` method. I know this principle is frequently violated in the Kivy codebase (even in `on_slides` below), but I try to fix it when I happen to work on code that does this.

The problem with this pattern is that if a user inherits and overwrites the `on_xxx` method for their own reason, we lose all the stuff that we wanted to happen as a result to `xxx` changing. So it's better to bind directly. For this case you'd need to bind a local function that eats the `args` passed from the callback, but it's still better.

kivy/uix/carousel.py
```diff
@@ -361,6 +353,18 @@ def _insert_visible_slides(self, _next_slide=None, _prev_slide=None):
         else:
             self._next = None
 
+        expected_value = self.loop and (len(self.slides) == 2)
+        if self._prev_equals_next != expected_value:
+            from kivy.logger import Logger
+            Logger.error(
+                "Carousel: I expected '_prev_equals_next' to be already "
```
matham: This smells to me like it should be an `assert`. Logging errors is generally discouraged because it trains people to ignore them, and if this path is hit, users would see an error without much for them to do. This seems like if this path is hit it would be a bug in kivy if I read this correctly, hence an `assert` is more appropriate.

kivy/data/style.kv
```diff
@@ -538,6 +538,10 @@
         BoxLayout:
             id: container
 
+
+<Carousel>:
+    _prev_equals_next: self.loop and (len(self.slides) == 2)
```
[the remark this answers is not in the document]
gottadiveintopython: I like using KV like this because it's much simpler compared to doing the same thing in python. But if you don't think so, I don't do this anymore on this project.

Regardless of that, I'll remove that part because I'll make `_prev_equals_next` into non KivyProperty as you suggested.

kivy/uix/carousel.py
gottadiveintopython: I agree with you. I'll make those into non KivyProperty.

kivy/uix/carousel.py
gottadiveintopython: Yes, I understand what you mean. I'll fix that.

kivy/uix/carousel.py
gottadiveintopython: This may be related to the order of evaluation. I mean things like this:

```python
from kivy.lang import Builder
from kivy.properties import StringProperty


class TestEvaluationOrder(Factory.Widget):
    prop1 = StringProperty()
    prop2 = StringProperty()

    def apply_class_lang_rules(self, *args, **kwargs):
        def handler(c, self, *args):
            print(c, ':', self.prop2 == self.prop1 + ' prop2')

        self.fbind('prop1', handler, 'A')
        r = super().apply_class_lang_rules(*args, **kwargs)
        self.fbind('prop1', handler, 'B')
        return r

KV_CODE = '''
<TestEvaluationOrder>:
    prop2: self.prop1 + ' prop2'

TestEvaluationOrder:
'''
w =  Builder.load_string(KV_CODE)
w.prop1 = 'prop1'
```

```
A : False
B : True
```

> Logging errors is generally discouraged because it trains people to ignore them

got it

gottadiveintopython:
Thanks for your reviews as always

matham:
I didn't test, but the changes look good. Let me know when you think it's ready to merge by removing WIP from the title.

gottadiveintopython:
Well didn't pass Travis CI.
I'm OK with either of 1d2ce89c12b991331746ee395f95272341129913(passed) or 950677861273694effd66301d673f07b57719933(didn't pass), because 950677861273694effd66301d673f07b57719933 is not related to the PR.

PR #6408, opened 2019-07-06
Create a transparent "overlay" GUI on raspberry PI (Dispmanx)

deganii:
# Summary:
## Transparent/Overlay GUI's in Raspberry PI

This feature allows the Kivy GUI to act as a transparent overlay on top of a terminal window or another application.  It is implemented within the Raspberry Pi EGL Display provider. An illustration with sample code is given below. Normally, the sample would load on-screen like this:

snapshot

However, a user can now set KIVY_BCM_DISPMANX_OVERLAY as follows:
`os.environ['KIVY_BCM_DISPMANX_OVERLAY'] = "1"`

They will then see this:

snapshot1

The blending is done at the Dispmanx layer.  Below is an example of a C++/OpenVG application showing a live camera feed at 30FPS and drawing native vguRect paths underneath the Kivy interface. 

snapshot-vid

# Working Test Example 
## (Raspian Stretch Lite / Kernel version 4.19.50 / Raspberry Pi 3B a02082)

Below is a working "hello world" app with two buttons. One is transparent, and the other is opaque, as described above. No special code is needed, other than setting the KIVY_BCM_DISPMANX_OVERLAY environment variable.

The patch was tested against the Kivy stable-1.11.0 branch. It works seamlessly in conjunction with the KIVY_BCM_DISPMANX_LAYER variable for easy stacking/mixing of native and python apps. Hopefully others will find it useful!

```
import os
os.environ['KIVY_GL_BACKEND'] = 'gl'
os.environ['KIVY_BCM_DISPMANX_OVERLAY'] = "1"
import kivy
kivy.require('1.11.0')

from kivy.app import App

class MyApp(App):
    def build(self):
        layout = FloatLayout(size=(800, 480), pos=(0,0))
        buttonO = Button(text='Hello Opaque', size_hint=(0.2,0.25),
                         pos_hint={'pos':(0.5,0.4)},
                         background_color=(1.,0,0,1.))
        buttonT = Button(text='Hello Transparent', size_hint=(0.2,0.25),
                         pos_hint={'pos':(0.25,0.4)},
                         background_color=(1.,0,0,0.8))
        layout.add_widget(buttonO)
        layout.add_widget(buttonT)

        return layout

if __name__ == '__main__':
    MyApp().run()
```


kivy/core/window/window_egl_rpi.py
```diff
@@ -26,6 +26,9 @@ class WindowEglRpi(WindowBase):
 
     _rpi_dispmanx_id = int(environ.get("KIVY_BCM_DISPMANX_ID", "0"))
     _rpi_dispmanx_layer = int(environ.get("KIVY_BCM_DISPMANX_LAYER", "0"))
+    _rpi_dispmanx_overlay = int(environ.get("KIVY_BCM_DISPMANX_OVERLAY", "0"))
```
matham: This needs to be documented at https://github.com/kivy/kivy/blob/715417e389b3d13115811cfccaf8b408963739e0/doc/sources/guide/environment.rst.

kivy/core/window/window_egl_rpi.py
```diff
@@ -44,7 +47,10 @@ def _create_window(self, w, h):
         display = egl.bcm_display_open(self._rpi_dispmanx_id)
         update = egl.bcm_update_start(0)
         element = egl.bcm_element_add(
-            update, display, self._rpi_dispmanx_layer, dst, src)
+            update, display, self._rpi_dispmanx_layer, dst, src,
+            self._rpi_dispmanx_overlay)
+        if self._rpi_dispmanx_overlay:
+            self.clearcolor = (0., 0., 0., 0.)
```
matham: I don't think we should do this by default as this goes against the docs https://github.com/kivy/kivy/blob/master/kivy/core/window/__init__.py#L480. Although this flag is new, it's still better to have the user set the desired opacity and update based on that.

Specifically, can you add `clearcolor` to the properties watched [here](https://github.com/kivy/kivy/blob/master/kivy/core/window/__init__.py#L994). Then, you won't need the `KIVY_BCM_DISPMANX_OVERLAY` variable, but instead, if the clear color opacity value is `>=1`, use `opaque` window mode, otherwise, use the `blend` mode and set `alpha.opacity` to the appropriate value based on `clearcolor[3]`.

If this can work, I think what I described would be preferable.

misl6:
Closing as DISPMANX support is not available anymore on recent Raspberry Pi (and not supported anymore)

We're still including support for DISPMANX on Legacy Raspberry Pi OS Buster, but related code will be removed as soon we remove support for Buster. (or Python 3.7, which is the default on Buster and reached EOL)

PR #6415, opened 2019-07-11
Add tests for coordinates translation

gottadiveintopython:


kivy/tests/test_uix_translate_coordinates.py
```diff
@@ -0,0 +1,74 @@
+import pytest
+import functools
+
+NON_RELATIVE_TYPE_WIDGET_CLS_NAMES = ('Widget', )
+RELATIVE_TYPE_WIDGET_CLS_NAMES = (
```
matham: Can you please make them lowercase? That fits in better with our and Python's style.

gottadiveintopython: ok done

gottadiveintopython:
Looks like I don't have the permission to do that. 
(I don't need it though.)

PR #6422, opened 2019-07-17
Fixes multiline label w/ line_height < 1

misl6:
This PR fixes a "rendering" issue that prevents the user from take advantage of the `Label` `line_height` property when`line_height` is less than `1`

I think that needs some tests from devs and users in order to check unexpected behaviors on different devices and configurations.

Testcase example:

```python
BoxLayout:
    size_hint: 1, .5
    Label:
        size_hint: 1, 1
        color: 0, 0, 0, 1
        text_size: self.size
        font_size: dp(13)
        halign: 'left'
        valign: 'top'
        max_lines: 3
        shorten_from: 'right'
        line_height: .7
        padding: 0, 0
        text: "Lorem ipsum dolor sit amet demo demo demo demo"
```

### DIFF
--- a/kivy/core/text/_text_sdl2.pyx
+++ b/kivy/core/text/_text_sdl2.pyx
@@ -127,7 +127,10 @@ cdef class _SurfaceContainer:
         r.w = st.w
         r.h = st.h
         SDL_SetSurfaceAlphaMod(st, <int>(color[3] * 255))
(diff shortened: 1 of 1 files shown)


matham:
I don't understand the relationship between `font_blended` and line height being less than 1?

tshirtman:
when lines where less than 1, lines would cut each others instead of sharing the overlapping space, this looked bad.

Issue #6392, opened 2019-06-21
Bubble widget does not dismiss itself


EntPyle:
### Versions

* Python: 3.6.7
* OS: Windows 10
* Kivy: 1.11.0
* Kivy installation method: conda

### Description

I think the bubble widget should have the same self dismissal behavior as a spinner menu or a popup, such that it is destroyed when a touch occurs outside of it's boundaries. As it stands, this behavior has to be created and controlled by each user. Is there a reason for the current behavior? If not, I'd be happy to submit a PR.

matham:
A PR would be welcome, But the new behavior should be optional, like `auto_dismiss` for modal view.

EntPyle:
I'll get to work on the PR, using the modal view behavior as a baseline.

PR #6432, opened 2019-07-22
Added auto dismiss behavior to Bubble, as an option. 

EntPyle:
#6392 
Used the modal view dismiss() and the TextInputCutCopyPaste hide() functions as templates to add the option to have bubbles automatically dismiss when a touch is observed outside of the widget.

### DIFF
--- a/kivy/uix/bubble.py
+++ b/kivy/uix/bubble.py
@@ -59,17 +59,19 @@ To change the appearance of the bubble::
 
 __all__ = ('Bubble', 'BubbleButton', 'BubbleContent')
 
-from kivy.uix.image import Image
-from kivy.uix.widget import Widget
-from kivy.uix.scatter import Scatter
-from kivy.uix.gridlayout import GridLayout
-from kivy.uix.boxlayout import BoxLayout
-from kivy.uix.button import Button
 from kivy.properties import ObjectProperty, StringProperty, OptionProperty, \
     ListProperty, BooleanProperty
-from kivy.clock import Clock
+
+from kivy.animation import Animation
 from kivy.base import EventLoop
+from kivy.clock import Clock
 from kivy.metrics import dp
+from kivy.uix.boxlayout import BoxLayout
+from kivy.uix.button import Button
+from kivy.uix.gridlayout import GridLayout
+from kivy.uix.image import Image
+from kivy.uix.scatter import Scatter
+from kivy.uix.widget import Widget
 
 
 class BubbleButton(Button):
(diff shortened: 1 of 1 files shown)


kivy/uix/bubble.py
```diff
@@ -235,6 +245,33 @@ def clear_widgets(self, **kwargs):
         else:
             content.clear_widgets()
 
+    def hide(self, *largs, **kwargs):
+        '''Close the Bubble if it is open.
+        When the view is dismissed, it will be faded out before being
+        removed from the parent. If you don't want animation, use::
+
+            view.dismiss(animation=False)
+
```
tshirtman: this is a nice change, but i think instead of passing animation=False here, i would prefer the app to have an animation ObjectProperty, defaulting to the current animation you put, and with allownone=True, so one can disable it easily, or replace it with another animation.
also add the versionadded info.

kivy/uix/bubble.py
```diff
@@ -91,6 +93,14 @@ class Bubble(GridLayout):
     '''Bubble class. See module documentation for more information.
     '''
 
+    auto_dismiss = BooleanProperty(True)
+    '''This property determines if the widget is automatically
+    removed when the user clicks outside it.
+
+    :attr:`auto_dismiss` is a :class:`~kivy.properties.BooleanProperty` and
+    defaults to True.
+    '''
```
tshirtman: please add indication of the version it was added like in other docstrings.

kivy/uix/bubble.py
```diff
@@ -235,6 +245,33 @@ def clear_widgets(self, **kwargs):
...
+            anim.start(self)
+        else:
+            parent.remove_widget(self)
+
+    def on_touch_down(self, touch):
+        if not self.collide_point(*touch.pos):
+            if self.auto_dismiss:
+                self.hide()
+                return True
+        super(Bubble, self).on_touch_down(touch)
```
tshirtman: I think you should return the result of the call to super, here, instead of always True.

chavarinbonbyn:
Hello, 
Has this been fixed ?, if so in which version ?. I' currently using 1.10.1 and the issue persist.
Thanks.

matham:
To add to what @tshirtman asked, please make the changes similar to dropdown and modal widgets. You can see some explanations at https://github.com/kivy/kivy/pull/6650.

Zen-CODE:
@EntPyle It seems you can't make these changes, so I'm going to close this ticket. Pity, seems like a few people wanted these changes. If you do get around to making these changes, please re-open the PR. Thanks.

PR #6433, opened 2019-07-22
Center cache problem on MacOS

ElliotGarbus:
On MacOS, The position of PopUps was wrong prior to a resize.  After a resize, popups were in the correct position.  Removing the cache=true setting from the center alias property solved the problem.

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -523,7 +523,7 @@ class WindowBase(EventDispatcher):
     def _get_center(self):
         return self.width / 2., self.height / 2.
 
-    center = AliasProperty(_get_center, bind=('width', 'height', '_density'),cache=True)
+    center = AliasProperty(_get_center, bind=('width', 'height', '_density'), cache=True)
     '''Center of the rotated window.
 
     .. versionadded:: 1.0.9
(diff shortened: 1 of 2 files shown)


kivy/core/window/__init__.py
```diff
@@ -523,7 +523,7 @@ def _get_height(self):
     def _get_center(self):
         return self.width / 2., self.height / 2.
 
-    center = AliasProperty(_get_center, bind=('width', 'height'), cache=True)
+    center = AliasProperty(_get_center, bind=('width', 'height'))  # Do not cache center
```
matham: The problem seems to have been that density is changed, but the property being cached was not updated. Is that correct?

If so, rather than not caching which has performance implications, a better solution would be to make density a `NumericProperty`, and then add it to the bind list.

matham:
Also, you don't need to make a new PR. You can just update this PR with your changes.

ElliotGarbus:
Done.  Thanks for your help and all you do on Kivy!

kivy/core/window/__init__.py
```diff
@@ -523,7 +523,7 @@ def _get_height(self):
     def _get_center(self):
         return self.width / 2., self.height / 2.
 
-    center = AliasProperty(_get_center, bind=('width', 'height'), cache=True)
+    center = AliasProperty(_get_center, bind=('width', 'height', '_density'),cache=True)
```
matham: The PEP8 test on CI is failing. I'm guessing the line is too long?

ElliotGarbus:
Needed a space… done!

matham:
*Now* it is complaining about the line being too long :)

PR #6443, opened 2019-07-28
Don't init property created in KV with given value

matham:
In the case where a widget doesn't have a property (e.g. property `my_prop`), if you set this property in kv its value is stored in the property instance for the widget *class* as opposed to in the widget *instance*.

E.g.:

```
<SomeWidget>:
    my_prop: app.large_obj
    OtherWidget:
        other_prop: app.other_obj
```

Currently, both `my_prop` and `other_prop` will have a property instance created for the `SomeWidget` and `OtherWidget`, respectively because the didn't yet exist. Similarly, they will each hold a ref to `app.large_obj` and `app.other_obj`, respectively.

The problem is that this object is held indefinitely and cannot be garbage collected. This is also undesired because it's not clear from kv, that the value is saved forever as the default. Instead, this PR changes it such that the property will be initialized to the default value of that property. The property will still be set to the appropriate value, it just will be set for each widget instance, instead of as the default stored in the class.

Backward compatibility
---------------------------

If you do:

```
<SomeWidget>:
    my_prop: app.large_obj
```
Then it doesn't make a difference whether it's the class default or instance default because every instance will be set to the value. The only difference is that if it's the class default the value will of course be whatever default value is immediately, while with the changes the property will not be set to the desired value until the rule is executed (it's value until the rule is set is the default value for that property class). But this is not different than other kv rules where the property already exists so I don't think it's an issue.

For the other case, where the rule is not a class, but rather a instance rule, e.g.:

```
<SomeWidget>:
    OtherWidget:
        other_prop: app.other_obj
```

This is very similar to the previous case, except for `OtherWidget` instances created before the rule is added to `Builder` and executed the very first time. Previously, all of them would default to the given value as it's stored in the class. But with these changes, it is not stored in the class, so existing widgets would default to the default value for the property class.

However, this is not an issue because you shouldn't really access this property for widgets created before the rule is added to `Builder` because they are not properly initialized.

So I think this can be merged with little backward compat issues.

I'm planning to add tests for this stuff testing memory leaks once the other PRs I created today are merged as they are all required together to prevent memory leaks.

### DIFF
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -800,7 +800,7 @@ cdef class EventDispatcher(ObjectWithUid):
             ret[x] = p[x]
         return ret
 
-    def create_property(self, str name, value=None, *largs, **kwargs):
+    def create_property(self, str name, value=None, default_value=True, *largs, **kwargs):
         '''Create a new property at runtime.
 
         .. versionadded:: 1.0.9
@@ -817,6 +817,10 @@ cdef class EventDispatcher(ObjectWithUid):
             Also, now and positional and keyword arguments are passed to the
             property when created.
 
+        .. versionchanged:: 2.0.0
+
+            default_value has been added.
+
         .. warning::
 
             This function is designed for the Kivy language, don't use it in
@@ -829,7 +833,10 @@ cdef class EventDispatcher(ObjectWithUid):
             `value`: object, optional
                 Default value of the property. Type is also used for creating
                 more appropriate property types. Defaults to None.
-
+            `default_value`: bool, True by default
+                If True, `value` will be the default for the property. Otherwise,
+                the property will be initialized with the the property type's
+                normal default value, and subsequently set to ``value``.
 
         ::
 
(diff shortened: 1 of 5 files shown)


kivy/_event.pyx
```diff
@@ -840,25 +847,46 @@ cdef class EventDispatcher(ObjectWithUid):
...
-        elif isinstance(value, (int, float)):
-            prop = NumericProperty(value, *largs, **kwargs)
-        elif isinstance(value, string_types):
-            prop = StringProperty(value, *largs, **kwargs)
-        elif isinstance(value, (list, tuple)):
-            prop = ListProperty(value, *largs, **kwargs)
-        elif isinstance(value, dict):
-            prop = DictProperty(value, *largs, **kwargs)
+
+        if default_value:
```
AndiEcker: DRY - replacing the code within this if-block with the following code line:
```suggestion
        if default_value:
            largs = (value, ) + largs
```
... you could indent out the code in the `else:` code block (using it for both cases: default_value == True and == False).

kivy/_event.pyx
```diff
@@ -840,25 +847,46 @@ cdef class EventDispatcher(ObjectWithUid):
...
+                prop = ListProperty(value, *largs, **kwargs)
+            elif isinstance(value, dict):
+                prop = DictProperty(value, *largs, **kwargs)
+            else:
+                prop = ObjectProperty(value, *largs, **kwargs)
         else:
-            prop = ObjectProperty(value, *largs, **kwargs)
+            if value is None:  # shortcut
+                prop = ObjectProperty(*largs, **kwargs)
+            if isinstance(value, bool):
```
AndiEcker:
```suggestion
            **el**if isinstance(value, bool):
```

kivy/_event.pyx
matham: I've simplified it in an alternative way.

AndiEcker: great job - your simplification is more clean :)

PR #6448, opened 2019-08-01
Move `__self__` from widget to EventDispatcher and fix tests.

matham:
Given that `proxy_ref` is now in `EventDispatcher`, it's helpful to also move `__self__` to it so we can get the actual object.

Also adds memory leaks tests making sure the app is gc'd after each test is done, and fixes the often failing network test to not fail due to lack of internet.

### DIFF
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -925,6 +925,10 @@ cdef class EventDispatcher(ObjectWithUid):
         self._proxy_ref = _proxy_ref = WeakProxy(self)
         return _proxy_ref
 
+    @property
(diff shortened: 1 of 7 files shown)


jabdoa2:
This change should probably be listed in https://github.com/kivy/kivy/issues/5560 as it breaks our current code around adding a ScreenManager to any other widget.

matham:
But how is it an API break? Moving to a superclass should not be an API break?

jabdoa2:
I will investigate this some more. It came up in git bisect. Maybe it is coincidence.

Issue #2198, opened 2014-05-17
ModalView background_color is ignored


PhoenixWright:
Platform: OS X
Kivy Version: 1.8.1-dev pulled on 05/16/2014
Test Code: https://gist.github.com/PhoenixWright/2476f3194d39e635c57e
My log output is in a comment on that gist.

I would have expected to not see the background color of the ModalView, and I see this instead, which looks like the normal background has been unaffected by background_color:

![screen shot 2014-05-17 at 11 53 03 am](https://cloud.githubusercontent.com/assets/382245/3005814/87059d3c-ddf4-11e3-9e4c-5fbb08b4e7f7.png)
##

darkopevec:
Test code is unavailable, though I believe this is not an issue.
background-color affects the overlay around the ModalView, whereas you would like to change background:

    background = StringProperty(
        'atlas://data/images/defaulttheme/modalview-background')
    '''Background image of the view used for the view background.

KeyWeeUsr:
@darkopevec exactly, but maybe the naming is kind of weird. It's not in this widget only, so maybe it was intended. And maybe `shadow_color` or something would be better.

PhoenixWright:
@darkopevec, you're correct. Sorry about deleting the test code, I figured a year or so was fine, should have pasted it in here instead.

darkopevec:
@KeyWeeUsr I agree the naming is misleading and I would second your suggestion to rename it kivy wide.

Julian-O:
Believed fixed by #6469.

If the problem persists, please let us know.

PR #6469, opened 2019-08-14
Updating ModalView to improve theming

snuq:
Some simple proposed changes to the ModalView class to make its background_color behave more closely to the same variable in other classes in kivy.  Also, added overlay_color variable to allow independent setting of the 'fade out' color when a popup is activated.

Note that the doc string for background_color is copied from the Button class, hopefully they parse correctly, I do not know anything about the documentation format or how to test it.

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -490,7 +490,7 @@
 
 # ModalView widget
 <ModalView>:
-    canvas.before:
+    canvas:
         Color:
             rgba: root.overlay_color[:3] + [root.overlay_color[-1] * self._anim_alpha]
         Rectangle:
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -490,14 +490,14 @@
 
 # ModalView widget
 <ModalView>:
-    canvas:
+    canvas.before:
         Color:
-            rgba: root.background_color[:3] + [root.background_color[-1] * self._anim_alpha]
+            rgba: root.overlay_color[:3] + [root.overlay_color[-1] * self._anim_alpha]
         Rectangle:
             size: self._window.size if self._window else (0, 0)
 
         Color:
-            rgb: 1, 1, 1
+            rgba: root.background_color
         BorderImage:
             source: root.background
             border: root.border
--- a/kivy/uix/modalview.py
+++ b/kivy/uix/modalview.py
@@ -118,11 +118,16 @@ class ModalView(AnchorLayout):
     defaults to None.
     '''
 
-    background_color = ListProperty([0, 0, 0, .7])
-    '''Background color in the format (r, g, b, a).
+    background_color = ListProperty([1, 1, 1, 1])
+    '''Background color, in the format (r, g, b, a).
 
-    :attr:`background_color` is a :class:`~kivy.properties.ListProperty` and
-    defaults to [0, 0, 0, .7].
+    This acts as a *multiplier* to the texture colour. The default
+    texture is grey, so just setting the background color will give
+    a darker result. To set a plain color, set the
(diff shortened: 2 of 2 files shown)


snuq:
i believe this will fix 
https://github.com/kivy/kivy/issues/2198
as well, tho it is difficult to tell without the user's code available.

kivy/data/style.kv
```diff
@@ -490,14 +490,14 @@
 
 # ModalView widget
 <ModalView>:
-    canvas:
+    canvas.before:
```
inclement: Is there a particular reason to use `canvas.before`? I think `canvas` is probably a better choice.

snuq: I was having trouble with it since the usual way to change the background image via canvas is using canvas.before, and when you try to use a canvas.before on a Popup (which is a child class of ModalView), it is completely overlayed by this, and the only way to effectively nullify this is to use a transparent png as the background image.

inclement: As @matham mentioned on discord, the canvas is itself ordered background to foreground, the point of canvas.before and canvas.after is to bypass that ordering when adding instructions in different places (in which case managing the single flattened canvas list is inconvenient). When you make a widget like this, you have full control over the ordering, so it's usually best to do everything on the normal canvas.

This is especially important for Kivy's own widgets, because these are building blocks that many apps will use. They generally want to avoid using the canvas.before and canvas.after specifically because that makes it harder for users to do their own drawing before and after the Kivy widget. Also note that 'drawing' encompasses more than displaying shapes, but also context instructions, see below.

Your change here is well motivated, but I think it's still the worse choice - if anything, the fact that the ModalView draws an opaque background image by default may be the bad part of the api. As I would see it, this change breaks the normal convention: you've specifically made canvas.before draw *after* the widget, which is the opposite of its normal behaviour. I understand why it may be initially surprising that drawing a background before everything else means it's behind the image that the widget displays...but isn't that what you should expect? That's why it's called canvas.before!

There may be other practical issues. For instance, users would normally expect that using a Rotate or Translate instruction on canvas.before would affect the widget, but this would not work here.

For these reasons, I think using canvas is the most consistent and preferable option.

inclement: Also, setting the `background_color` to have 0 opacity should be a more convenient alternative to using a transparent png.

snuq: ok, you are right about the background_color, i was thinking about adding this with the way it used to behave (i had to do the hack of making the background transparent on my app to get it to work properly).
What do you think about making the opacity fadeout thing a canvas.before, and the background image itself canvas?  Not asking for any usability reasons, just does that seem to make more sense?

inclement: My initial reaction is that even the fadeout should probably be on the canvas rather than canvas.before, simply because I don't see a major reason not to do it, but if you have a good reason for it then that could be fine.

snuq: Honestly, I dont have a good reason for putting it on canvas.before, just thought it made sense so I would suggest it, either way is fine by me.

snuq: Changed it back to just use canvas.

misl6:
I had the oppurtunity to check this PR on a project I'm working on, on my side it's working fine.

tshirtman:
> Nice work @snuq, thanks for your patience in working through the issues.
> 
> I'm mildly concerned about this being a breaking change, but have no major objection to merging it if we're happy with that. @tshirtman how does the break look to you?

i think this change is indeed for the better, the current way certainly prevent some reusability of the widget, and the api is not clear, since this is a 2.0 i think i'm ok with such breaks. But i don't have a lot of kivy apps to maintain currently, so i'm not very impacted.

PR #6479, opened 2019-08-17
doc: Fix typos, grammar in install instructions

Zen-CODE:
The only significant change here is removing the ```--no-binary``` flag. This is not required and missing the format parameter anyway.

https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-no-binary

### DIFF
--- a/doc/sources/installation/installation-linux-venvs.rst
+++ b/doc/sources/installation/installation-linux-venvs.rst
@@ -130,7 +130,7 @@ Replace `python` with `python3` for Python 3.
     pip install |cython_install|
 
     # Install stable version of Kivy into the virtualenv
-    pip install --no-binary kivy .
+    pip install --no-binary kivy kivy
     # For the development version of Kivy, use the following command instead
     pip install git+https://github.com/kivy/kivy.git@master
 
--- a/doc/sources/installation/installation-linux-venvs.rst
+++ b/doc/sources/installation/installation-linux-venvs.rst
@@ -109,7 +109,7 @@ After you ensure that a 3rd-party repository containing any packages that dnf is
 
 Installation
 ------------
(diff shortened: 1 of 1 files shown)


doc/sources/installation/installation-linux-venvs.rst
```diff
@@ -123,14 +123,14 @@ Replace `python` with `python3` for Python 3.
...
 
-    # if you'd like to be able to use the x11 winodw backend do:
+    # if you'd like to be able to use the x11 window backend do:
     export USE_X11=1
 
     # install the correct Cython version
     pip install |cython_install|
 
     # Install stable version of Kivy into the virtualenv
-    pip install --no-binary kivy
```
matham: `--no-binary` *is* needed. This is instructions for how to install from source. Without `--no-binary` it'll install the wheel, which we're trying to avoid here because we want to compile it ourselves. 

For some systems the apparently doesn't seem to work, so they have to do this. The fix should be to list `kivy` as the `--no-binary` argument. I think I made the same mistake elsewhere in the docs (of missing the `kivy` argument).

Zen-CODE: Ha. Yes, it's still wrong though. It's missing the fot.
```
$ pip install --no-binary kivy
You must give at least one requirement to install (see "pip help install")
```
If should be ```$ pip install --no-binary kivy .```, I think? Change made accordingly.

matham: It shouldn't be a dot because they haven't cloned so it's not local and they are getting it from pypi. So it should be `pip install kivy --no-binary kivy`

Zen-CODE: That works. Done.

Zen-CODE:
@matham Thanks :-)

PR #6481, opened 2019-08-17
Made ScreenManager clear_widgets correctly iterate over screens

inclement:
Currently ScreenManager's clear_widgets iterates over self.screens while modifying it, so it misses entries doesn't actually remove everything. The fix is straightforward, this is just an oversight.

### DIFF
--- a/kivy/uix/screenmanager.py
+++ b/kivy/uix/screenmanager.py
@@ -1013,7 +1013,7 @@ class ScreenManager(FloatLayout):
         self.screens.remove(screen)
 
     def clear_widgets(self, screens=None):
-        if not screens:
+        if screens is None:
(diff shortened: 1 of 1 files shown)


kivy/uix/screenmanager.py
```diff
@@ -1013,11 +1013,13 @@ def remove_widget(self, *l):
         self.screens.remove(screen)
 
     def clear_widgets(self, screens=None):
-        if not screens:
+        if screens is None:
             screens = self.screens
-        remove_widget = self.remove_widget
```
matham: Just FYI, the reason for `remove_widget = self.remove_widget` is performance, but it probably doesn't really make a difference here either.

Issue #6484, opened 2019-08-22
Explicitly setting KIVY_EVENTLOOP to 'sync' causes an error


gottadiveintopython:
### Versions

* Python: 3.7.1
* OS: Linux Mint 18.2 Mate Edition (derived from Ubuntu 16.04 LTS)
* Kivy: v2.0.0.dev0, git-3d87584, 20190822
* Kivy installation method:

### Description

When `KIVY_EVENTLOOP` is set to `"sync"` explicitly, `ValueError: async library sync not recognized` is occured.

### Code and Logs

```python
import os
os.environ['KIVY_EVENTLOOP'] = 'sync'
from kivy.app import App
from kivy.uix.widget import Widget

class TestApp(App):
    def build(self):
        return Widget()

TestApp().run()
```

```python
import os
os.environ['KIVY_EVENTLOOP'] = 'sync'
from kivy.app import runTouchApp
from kivy.uix.widget import Widget

runTouchApp(Widget())
```

Both code produce  `ValueError: async library sync not recognized`

gottadiveintopython:
Setting `KIVY_EVENTLOOP` to `'sync'`is not allowed in the first place, so this issue is invalid.

PR #6485, opened 2019-08-22
fix KIVY_EVENTLOOP doc

gottadiveintopython:
#6484

### DIFF
--- a/doc/sources/guide/environment.rst
+++ b/doc/sources/guide/environment.rst
@@ -223,13 +223,11 @@ Event Loop
 ----------
 
 KIVY_EVENTLOOP
-    Whether to use an async or synchronous event loop. See :mod:`kivy.app`
-    for example usage.
+    Which async library should be used when the app is run in an asynchronous
+    manner. See :mod:`kivy.app` for example usage.
 
-    ``'sync'``: When the app is run normally in a synchronous manner.
-        The default if not set.
     ``'asyncio'``: When the app is run in an asynchronous manner and the standard
-        library asyncio package should be used.
+        library asyncio package should be used. The default if not set.
     ``'trio'``: When the app is run in an asynchronous manner and the `trio`
         package should be used.
 


matham:
Sync is not an accepted backend for asynio so there's no reason to support it. Especially to mean asyncio.

We did support it before, but then unrealized there was no reason for it so I removed it. Why do you want to accept sync as a backend name to mean asyncio?

gottadiveintopython:
If the user doesn't set `KIVY_EVENTLOOP`

```python

runTouchApp(Widget())
```

the following part

https://github.com/kivy/kivy/blob/3d87584b963a79465ed52e09f4de17395c6aae60/kivy/clock.py#L1075-L1077

will be evaluated to

```python
    Clock = register_context(
        'Clock', _classes[_clk],
        async_lib='asyncio')
```

which doesn't cause the error. But if the user set  `KIVY_EVENTLOOP` to `sync`

```python
import os
os.environ['KIVY_EVENTLOOP'] = 'sync'

runTouchApp(Widget())
```

that part will be evaluated to

```python
    Clock = register_context(
        'Clock', _classes[_clk],
        async_lib='sync')
```

which causes the error. This PR makes the latter behave the same as the former.

gottadiveintopython:
The error comes from

https://github.com/kivy/kivy/blob/3d87584b963a79465ed52e09f4de17395c6aae60/kivy/clock.py#L532-L559

For me, it looks like we have to set `async_lib` to `asyncio` even though the event loop is run in 'sync' manner.

gottadiveintopython:
@matham 
> Sync is not an accepted backend for asynio so there's no reason to support it. Especially to mean asyncio.

This PR is not trying to do something like that.

matham:
I don't understand why you're setting `KIVY_EVENTLOOP` to `sync`? `KIVY_EVENTLOOP` is supposed to be used to set the default async library to use when you're doing async. If you're not doing async, there's no reason to set it, especially not to `sync`.

Perhaps the variable needs to be renamed to e.g. `KIVY_ASYNC_LIB` and that is the confusion?

gottadiveintopython:
> KIVY_EVENTLOOP is supposed to be used to set the default async library ...

The reason why I set `KIVY_EVENTLOOP` to `'sync'` is the [doc](https://github.com/kivy/kivy/blob/master/doc/sources/guide/environment.rst#event-loop). Shoud I fix the doc?

> Perhaps the variable needs to be renamed ...

Either is fine. For me both sound natural.

matham:
Yeah, that doc needs to be fixed. I do think `KIVY_ASYNC_LIB` as a little better as the name is clearer in relation to what the variable does, but I'm not sure if it's a good idea to change it at this point.

gottadiveintopython:
Done. But since I'm not fluent in English, I probably shouldn't do this kind of stuffs.

> but I'm not sure if it's a good idea to change it at this point.

I'm ok with changing it.

gottadiveintopython:
thanks @matham

PR #6496, opened 2019-08-29
[WIP] add 'tmp_builder' fixture for more isolated unit tests

gottadiveintopython:
When writing unit tests, I have to be careful of the widget-class-name. More specifically, I have to make sure the widget-class-name I'm going to use is not used by other unit tests, which is pain. This `tmp_builder` makes the all the changes to `Builder` and  `Factory` temporary. So now we can write tests like this:

```python
def test_1(tmp_builder):
    assert 'MyWidget' not in Factory.classes
    tmp_builder.load_string('<MyWidget>:\n\tButton')
        pass
    w = MyWidget()
    assert len(w.children) == 1

def test_2(tmp_builder):  # same as test_1
    assert 'MyWidget' not in Factory.classes
    tmp_builder.load_string('<MyWidget>:\n\tButton')
        pass
    w = MyWidget()
    assert len(w.children) == 1
```

### note

- Initially, I was trying to add two fixtures, which are `tmp_builder` and `tmp_factory`, but failed. I'd be glad if someone finds the cause out. ([first attempt](https://github.com/gottadiveintopython/kivy/tree/tmp_builder))

- I feel like `@pytest.fixture(scope='function')` is over kill for this fixture. `@pytest.fixture(scope='module')` might be better.

- Your tests still affected by other tests that don't use this fixture, even if you use it.

### DIFF
--- a/kivy/tests/conftest.py
+++ b/kivy/tests/conftest.py
@@ -10,6 +10,8 @@ except SyntaxError:
     # it's ok to fail here as it won't be used anyway
     pass
 
+from .fixtures import tmp_builder
+
 if kivy_eventloop != 'trio':
     @pytest.fixture()
     def nursery():
--- a/kivy/tests/fixtures.py
+++ b/kivy/tests/fixtures.py
@@ -4,7 +4,7 @@ import weakref
 import time
 import os.path
 
-__all__ = ('kivy_app', )
+__all__ = ('kivy_app', 'tmp_builder')
 
 # keep track of all the kivy app fixtures so that we can check that it
 # properly dies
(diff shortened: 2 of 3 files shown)


matham:
This is a good idea, but I'm not sure this can work in practice given how kivy works.

Specifically, as I'm sure you saw https://github.com/kivy/kivy/blob/master/kivy/tests/fixtures.py#L63, but I'd probably actually remove that part from it. Specifically, the problem is as follows:

You create `kivy/uix/foo.py` containing

```py
from kivy.lang import Builder

class FooWidget(Widget):
    pass

Builder.load_string("""
<FooWidget>:
    ...
""")
```

Now you add the test:

```py
def test_foo(tmp_builder):
    from kivy.uix.foo import FooWidget
    assert x
```

But then in the next test file you similarly have:

```py
def test_app(tmp_builder):
    from kivy.uix.foo import FooWidget
    assert y
```

The problem is that the kv rules would not be applied to `FooWidget` in `test_app` because `Builder.load_string` was called while the `tmp_builder` of `test_foo` active (I'm assuming it's the first time `from kivy.uix.foo import FooWidget` is called) so any rules loaded at that time are local to it.

The only solution is to import all kivy packages that may declare kv rules before any tests, but that is not a great solution. And what makes it harder, is that unless you import everything, you have no way of knowing when a kivy module with kv code will be imported the first time.

This is a fundamental flaw in kivy because kivy imports stuff so aggressively and makes use of singletons/global variables everywhere. And while that is generally convenient, it is not compatible with what you are trying to do or what I was trying to do in the `kivy_app` fixture by making things more local. 

Unfortunately, I'm not sure how to fix this without changing everything in kivy, which would require many breaking changes and the end result may not be better.

gottadiveintopython:
True, I understand. Thanks for the explanation.

PR #6498, opened 2019-09-01
There is no way to get event.edit.start and event.edit.length

tamo:
Firstly, Kivy should not ignore event.edit.start and event.edit.length.
**edit: I misunderstood them**
~~They have important meanings for Japanese IME.
They indicate the place where the user is editing.
I'll try to explain it metaphorically.
IME is like a translator.
You say "I love you," and a French translator says, "Je vous aime."
You'd think, "No, use 'te' instead of 'vous.'"
In this case, event.edit.text=="Je vous aime", event.edit.start==3, and event.edit.length==4.~~

You press space key, and the translator (IME) will give you a number of candidates (including "te" in the above case).
So secondly, this space key (and other keys while editing) should be ignored by Kivy. It's consumed by IME.

Lastly, honestly speaking, I haven't compiled Kivy myself yet.
The only thing I have really used is here:
https://osdn.net/users/tamomo/pf/MekikuViewer/scm/blobs/master/IMETextInput.py
So, I'm violating Kivy's contribution guideline.
I will install Visual C++ and other dependencies next week or later. But I'd like to give you a pull request for now.

Thanks!

### DIFF
--- a/examples/miscellaneous/on_textedit_event.py
+++ b/examples/miscellaneous/on_textedit_event.py
@@ -19,7 +19,7 @@ class TextInputIME(TextInput):
         super(TextInputIME, self).__init__(**kwargs)
         EventLoop.window.bind(on_textedit=self._on_textedit)
 
-    def _on_textedit(self, window, text):
+    def _on_textedit(self, window, text, start, length):
         self.testtext = text
 
 
--- a/examples/miscellaneous/on_textedit_event.py
+++ b/examples/miscellaneous/on_textedit_event.py
@@ -18,9 +18,29 @@ class TextInputIME(TextInput):
     def __init__(self, **kwargs):
         super(TextInputIME, self).__init__(**kwargs)
         EventLoop.window.bind(on_textedit=self._on_textedit)
+        self._editing = None
 
     def _on_textedit(self, window, text, start, length):
         self.testtext = text
+        if len(text):
+            self._editing = True
+        else:
+            self._editing = False
+
+        if length:
+            end = start + length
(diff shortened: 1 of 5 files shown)


tamo:
Hey, I managed to build Kivy with Visual Studio on my machine, sacrificing my sleep.
I've just tested it, and it (almost) works.

Bugs:
1. SDL always returns length==0: https://bugzilla.libsdl.org/show_bug.cgi?id=4782
2. Kivy's TextInput doesn't allow suggestion_text when len(self.text)==0: I have a workaround in https://osdn.net/users/tamomo/pf/MekikuViewer/scm/blobs/master/IMETextInput.py

So there's nothing wrong in my code so far.  ;)

matham:
Won't this break backwards compatibility for `on_textedit`? I suppose if there's no other way to do this it's ok, but is this the only way?

tamo:
Hmm, I should have thought of it more deeply.
Sorry to have bothered you.
I'll try it out some day again.

tamo:
Closing it, I'd tell you that on_suggestion_text has a bug I fixed in this fork:
  suggestion_text should appear after cursor_col, but the current code ignores cursor_col.
You can cherry-pick the fix.

matham:
I hope you didn't close it because of my comment. I didn't mean to reject the changes, but rather to see if you could find a way to do it without new parameters to `on_textedit`. But if there's no other way, this is fine if it fixes a bug.

tamo:
Thanks for your kind words, Matham.
I haven't found another way so far, but I will try someday.

PR #6515, opened 2019-09-13
Fixes positioning (valign) issue when using max_lines

misl6:
When using `max_lines`, I noticed a inconsistency with the selection made in the `valign` property.

Demo `kv` code:
```
Label:
    size_hint: 1, .190
    text_size: self.size
    text: "Text w/ max 2 lines, Text w/ max 2 lines, Text w/ max 2 lines, Text w/ max 2 lines, Text w/ max 2 lines, Text w/ max 2 lines, "
    color: 0, 0, 0, 1
    halign: 'left'
    valign: 'middle'
    font_size: dp(15)
    max_lines: 2
```

Before this PR:

Screenshot from 2019-09-13 15-48-55

After this PR (and how is supposed o be):

Screenshot from 2019-09-13 15-47-26

### DIFF
--- a/kivy/core/text/text_layout.pyx
+++ b/kivy/core/text/text_layout.pyx
@@ -566,8 +566,10 @@ def layout_text(object text, list lines, tuple size, tuple text_size,
     if max_lines > 0 and len(lines) > max_lines:
         val = True
         if dwn:
+            h-= sum(l.h for l in lines[max_lines:])
             del lines[max_lines:]
         else:
+            h-= sum(l.h for l in lines[:max(0, <int>len(lines) - max_lines)])
             del lines[:max(0, <int>len(lines) - max_lines)]
 
     # now make sure we don't have lines outside specified height


kivy/core/text/text_layout.pyx
```diff
@@ -565,6 +565,7 @@ def layout_text(object text, list lines, tuple size, tuple text_size,
     # above, we might have gone a few lines over
     if max_lines > 0 and len(lines) > max_lines:
         val = True
+        h -= lines[0].h * (len(lines) - max_lines)
```
matham: Shouldn't this be `if ...: h-= sum(l.h for l in lines[max_lines:])` as opposed to just using the first line's height?

misl6: Each line have a different height?
If yes, I missed something, so your advice is correct.

misl6: Yeah, You were right.
If the user is using for example a markup, We need to handle this kind of situation.
Screenshot from 2019-09-13 21-12-40
I've just committed the fix you proposed.

kivy/core/text/text_layout.pyx
```diff
@@ -565,6 +565,7 @@ def layout_text(object text, list lines, tuple size, tuple text_size,
     # above, we might have gone a few lines over
     if max_lines > 0 and len(lines) > max_lines:
         val = True
+        h-= sum(l.h for l in lines[max_lines:])
```
matham: Sorry, I was just giving you a starting point - this is still not correct. Depending on the condition value below `dwn`, different lines are removed. See below which lines are deleted and they need to be subtracted.

misl6: Did it. Thank you.

PR #6525, opened 2019-09-17
Add a "scroll_to_index" method to the ScrollView class

clickworkorange:
I had the need to scroll to a given item in a RecycleView and could not find a simple way to do this. The \<Row\> objects didn't seem to work with the existing `scroll_to` method, so I ended up rolling my own method, taking some inspiration from what's already in the ScrollView class. While it's very likely that I've simply overlooked a better way of doing this, it seemed simple and clean enough that I thought I'd share it with you in a PR - if nothing else I might end up learning something new.

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -1010,7 +1010,8 @@ class ScrollView(StencilView):
 
         vp = self._viewport
         pos = (vp.default_size[1] + vp.spacing) * index
-        dsx,dsy = self.convert_distance_to_scroll(0, pos - (self.height * 0.5))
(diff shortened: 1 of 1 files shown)


matham:
Thanks for your contribution, but this cannot be added to scrollview, because you added a method that would only work on a very specific use case of scollview - recycleview, to the generaic scrollview class. This method would fails for all other scrollview uses.

In addition, even for recycleview, it needs to be implemented more generally as explained here https://github.com/kivy/kivy/issues/5014.

clickworkorange:
Not only that - I completely forgot to calculate the distance for horizontal lists... :) A case of coding while tired! I'll set up some proper test cases for the different types and see if I can get somewhere with it. I still think it's a useful function.

Issue #6524, opened 2019-09-16
Selected files in the top directory are not highlighted.


MatthewTienWells:
### Versions

* Python: 3.7.2
* OS: Windows 10
* Kivy: 1.11.1
* Kivy installation method: pip

### Description

Expected behavior: When a file is selected in FileChooserIconView, the icon corresponding to it is highlighted.

Observed behavior: When the FileChooserIconView is initialized, files in the starting directory are not highlighted when selected. Changing directories results in expected behavior.

### Code and Logs

```python
import kivy
from kivy.app import App
from kivy.lang import Builder

kvfile = Builder.load_string("""
FileChooserIconView:
    id: Example
    path: '.'
""")

class Path_Bug(App):
    def build(self):
        return kvfile

if __name__ == '__main__':
    main = Path_Bug()
    main.run()


```

PR #6527, opened 2019-09-20
Fixed Issue #6524

MatthewTienWells:
Converts path to absolute path when first generating files for FileChooserController, preventing files from being unrecognized in the initial directory when attempting to select them.

### DIFF
--- a/kivy/uix/filechooser.py
+++ b/kivy/uix/filechooser.py
@@ -811,8 +811,16 @@ class FileChooserController(RelativeLayout):
         else:
             if platform == 'win':
                 is_root = splitdrive(path)[1] in (sep, altsep)
+                if not path:
+                    path = abspath(sep)
+                else:
+                    path = abspath(self.path)
             elif platform in ('macosx', 'linux', 'android', 'ios'):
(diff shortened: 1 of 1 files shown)


kivy/uix/filechooser.py
```diff
@@ -811,8 +811,16 @@ def _generate_file_entries(self, *args, **kwargs):
...
+                if not path:
+                    path = abspath(sep)
+                else:
+                    path = abspath(self.path)
             elif platform in ('macosx', 'linux', 'android', 'ios'):
                 is_root = normpath(expanduser(path)) == sep
+                if not path:
+                    path = abspath(sep)
+                else:
+                    path = abspath(self.path)
```
tshirtman: Since that code is identical, can't it be moved to after the else and thus deduplicated? Is it really ok to do that on windows?

Zen-CODE:
@MatthewTienWells  Thanks, but this issue has already been fixed via this PR: https://github.com/kivy/kivy/pull/6624

None the less, your effort is appreciated. :+1:  Thanks

PR #6533, opened 2019-09-22
fixes fbo/renderbuffer freeze on iOS

### LINKED ISSUE #3428: Fbo with stencilbuffer freeze on iOS

### LINKED ISSUE #5591: Creation of Fbo with_depthbuffer=True hangs UI thread on iOS
This issue occurs on iOS specifically - it does not reproduce on other platforms. 

When running a 3D OpenGL example, the creation of the FBO with_depthbuffer=True hangs the UI thread.  If this setting is set to false, then the UI thread does not hang, but the OpenGL rendering is broken (rendered flat), as one would expect.

The simple example, attached, uses the kivy3 wrapper for rendering 3D objects.  The UI thread hangs exactly at the point where the FBO is created, inside kivy3/renderer.py:

image

Here is the self-contained example code:

[ios_3d_test.zip](https://github.com/kivy/kivy/files/1676398/ios_3d_test.zip)

Other notes:
* The app does not crash, the UI thread merely hangs.
* There are no exceptions thrown in the log when the issue occurs
* Other background threads continue to run.

tito:
On iOS, a Fbo with depthbuffer was freezing the rendering.

The core issue was that GL_RENDERBUFFER was set back to 0, but iOS platform have already a RENDERBUFFER that we need to restore. There is not a single GL error during the creation or usage, but when the view is presented to iOS, it has a GL_INVALID_OPERATION, and XCode explain about the incorrect RENDERBUFFER.

It also remove the experimental on depth with stencil Fbo: by comparing our code with [openFrameworks](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworks/gl/ofFbo.cpp#L572), it appear that depthbuffer cannot work without stencil on GLES. So this patch also fixes the behavior, allowing to transparently make Fbo(with_depthbuffer=True) works on iOS.

Fixes #3428
Fixes #5591
Fixes kivy/kivy-ios#386

### DIFF
--- a/kivy/graphics/fbo.pyx
+++ b/kivy/graphics/fbo.pyx
@@ -73,6 +73,7 @@ include "opcodes.pxi"
 from os import environ
 from kivy.compat import PY2
 from kivy.logger import Logger
+from kivy.utils import platform
 from kivy.weakmethod import WeakMethod
 from kivy.graphics.texture cimport Texture
 from kivy.graphics.transformation cimport Matrix
@@ -87,6 +88,11 @@ from kivy.graphics.stencil_instructions cimport (
 
 cdef list fbo_stack = []
 cdef list fbo_release_list = []
+
+# must be running GLES2 not just the headers
+cdef bool is_gles_platform = USE_OPENGL_ES2
+if platform == "darwin":
+    is_gles_platform = False
 
 
 cdef class Fbo(RenderContext):
@@ -176,8 +182,11 @@ cdef class Fbo(RenderContext):
         self._texture = kwargs['texture']
         self.observers = []
 
-        if self._depthbuffer_attached and self._stencilbuffer_attached:
-            Logger.warning('Fbo: depth+stencil buffer support is experimental')
+        if is_gles_platform:
+            # on GLES/iOS, depthbuffer works with stencil enabled
+            # that's how openFrameworks did for all GLES platform in ofFbo.cpp
+            if self._depthbuffer_attached:
+                self._stencilbuffer_attached = True
 
         self.create_fbo()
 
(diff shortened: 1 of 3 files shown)


tito:
@matham can you review again?

jabdoa2:
Could this become a 1.11.1 release? This is one of the reasons why we still pin to 1.10.1.

matham:
You probably mean `1.11.2`. But it's very unlikely we'll be doing such a release, unless a core-dev volunteers to do that, but that's not likely. We really don't have the resources.

Why don't you simply maintain your own 1.11.1 branch in your kivy fork and cherry pick fixes to that and use that version of kivy in your app?

PR #6540, opened 2019-09-27
fix erroneous check of KIVY_NO_ENV_CONFIG

fluxrider:
The check to see if KIVY_NO_ENV_CONFIG is set was incorrect.

Without this fix, one has to set os.environ["KIVY_NO_ENV_CONFIG"] = "1" to be able to use KCFG_KIVY_LOG_LEVEL=warning and the likes. Which is the opposite of what config.py's doc says.

### DIFF
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -906,7 +906,7 @@ if not environ.get('KIVY_DOC_INCLUDE'):
             Logger.exception('Core: Error while saving default config file')
 
     # Load configuration from env
-    if environ.get('KIVY_NO_ENV_CONFIG', '0') != '0':
+    if environ.get('KIVY_NO_ENV_CONFIG', '0') != '1':
         for key, value in environ.items():
             if not key.startswith("KCFG_"):
                 continue


kivy/config.py
```diff
@@ -906,7 +906,7 @@ def name(self, value):
             Logger.exception('Core: Error while saving default config file')
 
     # Load configuration from env
-    if environ.get('KIVY_NO_ENV_CONFIG', '0') != '0':
+    if environ.get('KIVY_NO_ENV_CONFIG', '0') == '0':
```
matham: It should actually be `if environ.get('KIVY_NO_ENV_CONFIG', '0') != '1':`, because if it's empty it should disable it. Only if it's explicitly `1` should it be disabled.

fluxrider: Yes, that's alright. I went the C style if not zero then it's set, but it would allow people to be sloppy (a bit like how it was before by just checking if the variable existing in env instead of checking the value). Cheers.

fluxrider:
Thanks, it's always fun to contribute even if it's very small patch like this one.

PR #6541, opened 2019-09-28
remove the 'App'-less classname hack, there is no rationale

fluxrider:
There is hard to justify code that strips the 'App' of the classname when using it for things like the default title bar text. This stripping causes issues when the classname is just App, and looks silly in cases like MyApp. I'm sure the original author meant well, maybe for something like HelloWorldApp, but the stripping is really not needed.

The following code gives me an empty titlebar, whereas any other classname will show:
```
import kivy
import kivy.app
import kivy.uix.button

class App(kivy.app.App):
  def build(self):
    # self.title = 'App' # workaround
    return kivy.uix.button.Button(text='Hello World')

App().run()
```

Now I did not test this fix because I'm not setup nor ready at this time to invest the time having a dev setup, not knowing if Kivy is for me. I'm hoping someone else can do some sanity test. The string might be used for other things than just the title bar. Though in all honesty, right now it's pretty broken if the name of the class is just App.

### DIFF
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -698,10 +698,7 @@ class App(EventDispatcher):
         '''
         if self.title is not None:
             return self.title
-        clsname = self.__class__.__name__
-        if clsname.endswith('App'):
-            clsname = clsname[:-3]
-        return clsname
+        return self.__class__.__name__
 
     def get_application_icon(self):
         '''Return the icon of the application.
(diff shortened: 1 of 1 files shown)


matham:
We can't really do this because this will break all backward compatibility, because this is widely used in kivy for a default kv filename etc. We generally assume the app class name will be prefaced by something and not be just `App`.

If the title bar is an issue, you can manually set the title, which you probably should do anyway for most apps?

fluxrider:
I saw at line 681 that it checked both for the full non-stripped name and then stripped version when dealing with the filename.

But no worries, I understand it's a low reward, higher risk change. I'll delete the branch.

Issue #6473, opened 2019-08-14
TextInput Bubble disappears immediately after it appears


brianddk:
### Versions

* Python: Python 3.6.8
* OS: Ubuntu 18.04 u18-build 4.15.0-55-generic x86_64
* Kivy:  master ce9b5719 
* Kivy installation method: Python 3.6 Nightly wheel
* Buildozer: 40 ( master 7a0ef4f ) from pip tarball 
* Android: 8.0.0
* Android NDK: 17.2.4988734
* Android Build Tools: 29.0.2
* Android API target: 27 

### Description

***Observation:*** On Android, when selecting text from a TextInput, the cut/copy/paste bubble only shows up for a few milliseconds then disappears.  Far to quick to be of any use.

***Expectation:*** The cut/copy/paste bubble should remain visible for multiple seconds or until focus on TextInput is lost.

### Code and Logs

```python
from kivy.app import App
from kivy.uix.textinput import TextInput

class Sample(App):
    def build(self):
        return TextInput(text="Test Text")

_app = Sample()
_app.run()
```

shakna-israel:
Being able to control dismiss in [#6432](https://github.com/kivy/kivy/pull/6432) should go a ways to working around this, but the ms timing of the bubble vanishing is definitely "unexpected" behaviour.

---

I hit this as well, under:

Python: 3.7.4
Host OS: Arch Linux 5.2.8-arch1-1-ARCH
Buildozer: 40 (master 7a0ef4f10c)
Android: 5.1.1

btonasse:
I'm having the same issue on android 10. Is there anything I need to do on the code to prevent it from disappearing?

rachiteagles:
i solved this problem.

i created a CustomTextInput class that inherits from TextInput.
After that I overrided the function _hide_cut_copy_paste to comment out bubble.hide().
PFB code snippet

class CustomTextInput(TextInput):
    def _hide_cut_copy_paste(self, win=None):
        bubble = self._bubble
        if not bubble:
            return

        **#bubble.hide()**

Also, replaced TextInput with CustomTextInput in .kv file

Zen-CODE:
I think this could be solved by this merge:

    https://github.com/kivy/kivy-ios/issues/499

Would it be possible to test using a checkout of master that included this fix?

tshirtman:
This has been confirmed to be fixed in 2.0.0, closing :).

duarteleonardosales:
The problem continues in the new version 2.
[...]

PR #6542, opened 2019-09-28
Fixes TextInput Bubble from diseappering immediately after it appears

misl6:
Fixes issue #6473


btonasse:
My textinput.py does not reflect this change. Is this normal?

matham:
You're probably using the last stable release, try using master.

btonasse:
I've just tried using master, but the bubble is still disappearing immediately. :/

PR #6547, opened 2019-10-06
Image Zoom

christiandleonr:
A simple zoom on an image and the scale of the image on a label

### DIFF
--- a/examples/widgets/zoom.py
+++ b/examples/widgets/zoom.py
@@ -1,0 +1,163 @@
+from kivy.app import App
+from kivy.uix.boxlayout import BoxLayout
+from kivy.uix.scatterlayout import ScatterLayout
+from kivy.uix.scatter import Scatter
+from kivy.core.window import Window
+from kivy.uix.image import Image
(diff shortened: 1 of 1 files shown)


Zen-CODE:
So, if this is to go into the examples, it should ideally be a well documented, clear and concise piece of code. Keep in mind, people are going to be looking at it trying to understand what it does and how.

To that end, I would suggest the following:

1. Clear doc strings indicating what the purpose of the examples is, what is does and how to use it.
2. Pep 8 fixes - there are many pep 8 violations in the code.
3. More careful naming e.g. Line 145 starts a `MainApp` class, when it's not an app but a `BoxLayout`.
4. You might consider using kv to create to widget tree, as it would be more concise and readable?
5. The zoom currently only seems to support mouse scrolling as zoom, not pinching or multi-touch. It might be nice to add this for mobile use cases?

Those are some suggestions. You are welcome to question some of them if you feel otherwise?

Thanks for the contribution +1. But we do need to ensure the example are of the highest quality. Hoping to hear back from you.

Regards

Issue #6155, opened 2019-02-11
Meaningfull error message on lower case in kv file 


kjeldflarup:
It may just be me being a newbie, but it is a little up hill when a keyword in the .kv file starts with a lowercase, then all that is reported is just an attribute error.

Putting a meaningful message here would be good.
https://github.com/kivy/kivy/blob/4905319cd397f76df7d6683707426c92cca0a491/kivy/factory.py#L130

MetalDent:
Hi, anyone working on this issue?

7040934:
เปิด

pythonic64:
Solved by https://github.com/kivy/kivy/pull/6548.

PR #6548, opened 2019-10-09
Meaningful Error Message

PB020:
Addresses Issue #6155, adds a meaningful message regarding the cause of an AttributeError.


Franccisco:
As this pull request is already merged, the issue #6155 should be
closed.

PR #6549, opened 2019-10-12
Lazy loader for zip images

jabdoa2:
Loading animations from zips can take a lot of time when there are a lot of frames inside a zip. In my tests around 1ms per frame. However, we only need the first frame initially.

This change provides a lazy loader for images from within a zip file. Currently, it will load images + textures on demand. Initially, I though about preloading the next one or two images but it turned out that this is fast enough for all our cases. I might still add it later.


jabdoa2:
@tito @akshayaurora are you still planning to review this or should i close it?

Issue #6474, opened 2019-08-15
Buster RPi4B unable to find valuable Window provider from Desktop-less console


frankgould:
After logging on the Raspberry Pi 4B to the command line (using raspi-config boot to CLI), cannot execute kivy because of critical error when attempting to use any kivy Window provider: egl_rpi, sdl2, pygame, or x11. This issue lists the attempts and results of each of these Window providers. These providers are listed on the following page:

https://kivy.org/doc/stable/guide/environment.html

> KIVY_WINDOW
Implementation to use for creating the Window
Values: sdl2, pygame, x11, egl_rpi

### Versions
* Python: 3.7.3
* OS: Raspbian Buster 4.19
* Kivy: 1.11.1
* Kivy installation method: https://github.com/kivy/kivy/issues/6413

### Description
The same test.py program listed below was used to test all of the Window providers listed above. The second line is changed for each test results.
```
import os
#os.environ['KIVY_WINDOW'] = 'egl_rpi'
from kivy.app import App
from kivy.uix.button import Button

class TestApp(App):
	def build(self):
		return Button(text='Hello World')

TestApp().run()
```
When no (default) KIVY_WINDOW defined (test.py above), gets the following results:
```
[DEBUG  ] [Window      ] Provider <egl_rpi> ignored by config
[DEBUG  ] [Window      ] Provider <sdl2> ignored by config
[DEBUG  ] [Window      ] Provider <x11> ignored by config
[CRITICAL] [Window      ] Unable to find any valuable Window provider. 
[CRITICAL] [App         ] Unable to get a Window, abort.
```
After changing line 2 by removing the # sign, got the following results:
```
[DEBUG  ] [Window      ] Ignored <egl_rpi> (import error)
[DEBUG  ] [Window      ] Provider <sdl2> ignored by config
[DEBUG  ] [Window      ] Provider <x11> ignored by config
[CRITICAL] [Window      ] Unable to find any valuable Window provider. 
egl_rpi - ImportError: cannot import name 'bcm' from 'kivy.lib.vidcore_lite' (/usr/local/lib/python3.7/dist-packages/kivy/lib/vidcore_lite/__init__.py)
  File "/usr/local/lib/python3.7/dist-packages/kivy/core/__init__.py", line 63, in core_select_lib
    fromlist=[modulename], level=0)
  File "/usr/local/lib/python3.7/dist-packages/kivy/core/window/window_egl_rpi.py", line 12, in <module>
[...]

PR #6553, opened 2019-10-16
Added 'RPI' environmental variable

Lauszus:
This is used when compiling Kivy for the Raspberry Pi 4.

See: https://github.com/kivy/kivy/issues/6474#issuecomment-542679712

Please let me know if you want me to modify setup.py to autodetect the Raspberry Pi 4 instead of setting a environmental variable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -132,7 +132,7 @@ kivy_ios_root = environ.get('KIVYIOSROOT', None)
 if kivy_ios_root is not None:
     platform = 'ios'
 # proprietary broadcom video core drivers
-if exists('/opt/vc/include/bcm_host.h'):
+if environ.get('RPI', 1) and exists('/opt/vc/include/bcm_host.h'):
     platform = 'rpi'
 # use mesa video core drivers
 if environ.get('VIDEOCOREMESA', None):


matham:
Oh, I overlooked that you set rpi variable to `0` in the example. But still, since you set `VIDEOCOREMESA`, why do you need the new variable? It's anyway overwritten.

Lauszus:
@matham I see your point. 

The reason why I added it was so it still worked when ```VIDEOCOREMESA=0```, but as I wrote in my comment I can write some code that check specifically for the RPi4, so the user does not have to set any variable manually. That might be the better solution?

matham:
Does it still work when `VIDEOCOREMESA=0`?

But yeah, you're right, auto detection would be better either way.

Lauszus:
@matham yes it does. 

I've added auto detection in: https://github.com/kivy/kivy/pull/6562. Will close this one, as it's better to do auto detection anyway.

OutsourcedGuru:
`Lauszus deleted the Lauszus:rpi4 branch yesterday...`

Argghghghghghghgffhfhghghfhhg...

Lauszus:
@OutsourcedGuru please use the newest PR i.e. use the following command:

```bash
export VIDEOCOREMESA=1; pip3 install --verbose git+https://github.com/Lauszus/kivy.git@rpi4_auto#egg=kivy
```

Issue #3691, opened 2015-10-12
Pictures example should probably scale pics by screen size


stuaxo:
When you run it on a modern phone the pics are tiny

screenshot

http://imgur.com/j6wXrSm

they should probably have an initial scale determined by the available area

ivlevdenis:
Check this http://kivy.org/docs/api-kivy.metrics.html#manual-control-of-metrics

stuaxo:
I've noticed weird scaling on quite a few kivy apps... on my HTC X - here is a bug I reported to particle panda with a screenshot

https://github.com/chaosbuffalolabs/ParticlePanda/issues/10

PR #6559, opened 2019-10-19
Relative initial size for images

blu-base:
Addressing issue #3691 regarding the `pictures` example in `example/demo/pictures`.

Replacing the fixed initial size of the images of 400px by a relative size given by `Window.width`

### DIFF
--- a/examples/demo/pictures/pictures.kv
+++ b/examples/demo/pictures/pictures.kv
@@ -40,8 +40,9 @@ FloatLayout:
         id: image
         source: root.source
 
-        # create initial image to be 400 pixels width
-        size: 400, 400 / self.image_ratio
+        # create initial image with relative size
+        size: win.Window.width * 0.6, win.Window.width * 0.6 / self.image_ratio
+        allow_stretch: True
 
         # add shadow background
         canvas.before:


examples/demo/pictures/pictures.kv
```diff
@@ -40,8 +40,9 @@ FloatLayout:
         id: image
         source: root.source
 
-        # create initial image to be 400 pixels width
-        size: 400, 400 / self.image_ratio
+        # create initial image with relative size
+        size: win.Window.width * 0.6, win.Window.width * 0.6 / self.image_ratio
```
matham: I don't think you need to divide by `self.image_ratio`, because `win.Window.width` should already have then that.

blu-base: I see. Currently, i am trying to get familiar with kivy, so i surfed through the open issues, and found the 'easy' tagged ones. Lacking experience, I intended to change the least possible amount, so i only replaced the 400px of the loaded image by a runtime value, the root's window size.

But i don't understand what `win.Window.width` could do to determine the ratio of the loaded image. As far as i understood the code, `self.image_ratio` is the aspect ratio of the individually loaded image file. The original script, and my proposed change yield such a result:

pictures_w_ratio


Using `size: win.Window.width * 0.6, win.Window.width * 0.6` will yield unnecessary white space in the picture's frame. See below: 

pictures_wo_ratio


Therefore, i disagree. The aspect ratio of the image `self.image_ratio` should be in the computations determining the individual frame size.  Whether, the ratio divides or multiplies the base size is arguable. Of course it is also arguable which base size (`height` or `width`) is to be used as well its ratio (0.6). The proposed choice seemed to give the best impression in landscape and portrait mode.

Julian-O:
Sorry - wrong issue. Ignore my changes.

PR #6568, opened 2019-10-24
Cross compile wheel for armv7l (Raspberry Pi 4) using Github Actions CI

### LINKED ISSUE #6413: Raspbian 4.19 Buster RPi4B Install Instructions
This is a request to share my successful kivy installation instructions for Raspbian 4.19 Buster Desktop OS running on Raspberry Pi 3B+ and RPi4B:4GB. 

For issues relative to the Buster Lite (without Desktop), please refer to Issue https://github.com/kivy/kivy/issues/6474.

### Versions
* Raspberry Pi 3B+ and 4B (4GB RAM)
* Python: 3.7.3
* OS: 4.19 Buster Desktop (2019-07-10)
* Kivy: 1.11.1
* Kivy installation method:
> sudo apt update
> sudo apt upgrade
> sudo apt install libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev libsdl2-ttf-dev pkg-config libgl1-mesa-dev libgles2-mesa-dev python3-setuptools libgstreamer1.0-dev git-core gstreamer1.0-plugins-{bad,base,good,ugly} gstreamer1.0-{omx,alsa} python3-dev libmtdev-dev xclip xsel libjpeg-dev
> python3 -m pip install --upgrade --user pip setuptools
> python3 -m pip install --upgrade --user Cython==0.29.10 pillow
> sudo pip3 install kivy

### Description

These install instructions so far have been successful in initial tests using the screenmanagertest.py, code listing below. These same tests were used to prove the previous Raspbian OS exhibited memory leaks within minutes of execution. Initial tests with the latest Raspbian Buster release (2019-07-10) running on a Raspberry Pi 3B+ have exhibited no memory leaks within this same amount of time. More tests will be run over the next few days to confirm that smaller leaks don't show up after longer periods of time.

Note: On RPi3B, this test only executes successfully with max Kivy screen resolution of 800x600; any larger size and fullscreen results in system lockup after a few images are loaded. This may be a problem due to 1GB system RAM. 

On 08/09/19, this was tested successfully with a RPi4B with 4GB system RAM Desktop that confirmed to run full screen at 1366x768. More tests will be tried in the near future to confirm 1920x1080 resolution works over long periods of time.
 
### Code and Logs
[screenmanagertest.py code here](https://gist.github.com/frankgould/2464c0a3bf4387165d4968f9e849524b)

### LINKED ISSUE #6474: Buster RPi4B unable to find valuable Window provider from Desktop-less console

### LINKED ISSUE #6487: Arch Linux ARM 4.19 RPi4B Install Instructions
This issue is to report the success and failure of the kivy install on Arch Linux ARM 4.19 Desktop. The non-desktop (aka desktopless) installation does not work with these instructions due to limitations of compiling to include an X window provider.

### Versions
* Raspberry Pi: Model 4B, 4GB System RAM
* Python: 3.7.4
* OS: Arch Linux ARM 4.19, Exton RaspArch Build 190809
* Kivy: v2.0.0.dev0
* Kivy installation method: 
```
sudo pacman -S sdl2 sdl2_gfx sdl2_image sdl2_net sdl2_ttf sdl2_mixer python-setuptools cython python2-pillow
git clone https://github.com/kivy/kivy.git
cd kivy
make
```
### Description
Although the desktop version of kivy works, there has not been extensive tests to confirm other modules needed are available.

### Code and Logs
Here was what I added to /etc/environment to try the same install without the desktop (from the console):
```
DISPLAY=:0.0
USE_X11=1
VIDEOCOREMESA=1
KIVY_GRAPHICS="gles"
PYTHONPATH="/home/archpi/kivy"
```
NOTE: In a previous post, I used pip to install kivy and other packages. This resulted in a corrupted package management issue when attempting to install other Arch Linux ARM apps. This version has been confirmed to install and run the settings example app.

### LINKED ISSUE #6567: Cross compile wheel for armv7l (Raspberry Pi 4) during the Travis build
Would you be willing to accept a PR for cross compiling Kivy for the Raspberry Pi 4 using Travis, Docker and QEMU?

I believe this is possible on the Raspberry Pi 4, as it no longer use the proprietary Broadcom video core drivers.

I have done something similar for another project of mine: <https://github.com/Lauszus/socketsocketcan/blob/pybind11/Dockerfile.armv7l> and <https://github.com/Lauszus/socketsocketcan/blob/pybind11/.travis.yml>.

I think it will save a lot of users a lot of trouble and time if Kivy shipped precompiled wheels for the Raspberry Pi 4.

Lauszus:
Fixes #6567, fixes #6413, fixes #6474, fixes #6487

### DIFF
--- a/.ci/Dockerfile.armv7l
+++ b/.ci/Dockerfile.armv7l
@@ -1,0 +1,19 @@
+ARG image=balenalib/armv7hf-debian:buster
+FROM $image
+
+COPY . /kivy
+WORKDIR /kivy
+
+RUN [ "cross-build-start" ]
+
+# Install dependencies.
+RUN /bin/bash -c 'source .ci/ubuntu_ci.sh && \
+    export PIP_EXTRA_INDEX_URL="https://www.piwheels.org/simple" && \
+    install_kivy_test_run_apt_deps && \
+    install_python && \
+    install_kivy_test_run_pip_deps'
+
+# Build the wheel.
+RUN VIDEOCOREMESA=1 python3 -m pip -v wheel --extra-index-url https://www.piwheels.org/simple . -w /kivy-wheel
+
+RUN [ "cross-build-end" ]
--- a/.ci/Dockerfile.armv7l
+++ b/.ci/Dockerfile.armv7l
@@ -14,6 +14,6 @@ RUN /bin/bash -c 'source .ci/ubuntu_ci.sh && \
     install_kivy_test_run_pip_deps'
 
 # Build the wheel.
-RUN VIDEOCOREMESA=1 python3 -m pip -v wheel --extra-index-url https://www.piwheels.org/simple . -w /kivy-wheel
+RUN KIVY_SPLIT_EXAMPLES=1 VIDEOCOREMESA=1 python3 -m pip -v wheel --extra-index-url https://www.piwheels.org/simple . -w /kivy-wheel
 
 RUN [ "cross-build-end" ]
--- a/.github/workflows/armv7l_wheels.yml
+++ b/.github/workflows/armv7l_wheels.yml
@@ -8,7 +8,6 @@ on:
     - cron:  '23 1 * * *'
 
 env:
-  KIVY_SPLIT_EXAMPLES: 1
   SERVER_IP: '159.203.106.198'
   DOCKER_IMAGE: 'balenalib/armv7hf-debian:buster'
 
--- a/.github/workflows/armv7l_wheels.yml
+++ b/.github/workflows/armv7l_wheels.yml
@@ -9,18 +9,20 @@ on:
 
 env:
   SERVER_IP: '159.203.106.198'
-  DOCKER_IMAGE: 'balenalib/armv7hf-debian:buster'
 
 jobs:
   armv7l:
     runs-on: ubuntu-18.04
     if: github.event_name == 'schedule' || (github.event_name == 'create' && github.event.ref_type == 'tag') || contains(github.event.head_commit.message, '[build wheel]') || contains(github.event.head_commit.message, '[build wheel armv7l]') || contains(github.event.pull_request.title, '[build wheel]') || contains(github.event.pull_request.title, '[build wheel armv7l]')
+    strategy:
+      matrix:
+        docker_images: ['balenalib/armv7hf-debian:stretch', 'balenalib/armv7hf-debian:buster']
     steps:
     - uses: actions/checkout@v1
-    - name: Make armv7l wheels
+    - name: Make ${{ matrix.docker_images }} wheel
       run: |
         source .ci/ubuntu_ci.sh
-        generate_armv7l_wheels $DOCKER_IMAGE
+        generate_armv7l_wheels ${{ matrix.docker_images }}
     - name: Rename wheels
       if: github.event.ref_type != 'tag'
       run: |
--- a/.github/workflows/armv7l_wheels.yml
+++ b/.github/workflows/armv7l_wheels.yml
@@ -39,12 +39,6 @@ jobs:
       with:
         name: armv7l_wheels
         path: dist
-    - name: Publish to PyPI
-      if: github.event_name == 'create' && github.event.ref_type == 'tag'
-      uses: pypa/gh-action-pypi-publish@master
-      with:
-        user: __token__
-        password: ${{ secrets.pypi_password }}
     - name: Upload to GitHub Release
       uses: softprops/action-gh-release@78c309ef59fdb9557cd6574f2e0be552936ed728
       if: startsWith(github.ref, 'refs/tags/')
--- a/.github/workflows/armv7l_wheels.yml
+++ b/.github/workflows/armv7l_wheels.yml
@@ -1,0 +1,65 @@
+name: armv7l wheels
+
+on:
+  push:
+  pull_request:
+  create:
+  schedule:
+    - cron:  '23 1 * * *'
+
+env:
+  KIVY_SPLIT_EXAMPLES: 1
+  SERVER_IP: '159.203.106.198'
+  DOCKER_IMAGE: 'balenalib/armv7hf-debian:buster'
+
+jobs:
+  armv7l:
+    runs-on: ubuntu-18.04
+    if: github.event_name == 'schedule' || (github.event_name == 'create' && github.event.ref_type == 'tag') || contains(github.event.head_commit.message, '[build wheel]') || contains(github.event.head_commit.message, '[build wheel armv7l]') || contains(github.event.pull_request.title, '[build wheel]') || contains(github.event.pull_request.title, '[build wheel armv7l]')
+    steps:
+    - uses: actions/checkout@v1
+    - name: Make armv7l wheels
+      run: |
+        source .ci/ubuntu_ci.sh
+        generate_armv7l_wheels $DOCKER_IMAGE
+    - name: Rename wheels
+      if: github.event.ref_type != 'tag'
+      run: |
+        source .ci/ubuntu_ci.sh
+        rename_wheels
+    - name: Upload wheels to server
+      if: github.event_name != 'pull_request'
+      env:
+        UBUNTU_UPLOAD_KEY: ${{ secrets.UBUNTU_UPLOAD_KEY }}
+      run: |
+        source .ci/ubuntu_ci.sh
+        upload_file_to_server "$SERVER_IP" "armv7l/kivy/"
+    - name: Upload wheels as artifact
+      uses: actions/upload-artifact@master
+      if: github.event_name == 'pull_request'
+      with:
+        name: armv7l_wheels
+        path: dist
+    - name: Publish to PyPI
+      if: github.event_name == 'create' && github.event.ref_type == 'tag'
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.pypi_password }}
+    - name: Upload to GitHub Release
+      uses: softprops/action-gh-release@78c309ef59fdb9557cd6574f2e0be552936ed728
+      if: startsWith(github.ref, 'refs/tags/')
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        files: dist/*
+        draft: true
+
+  always_job:
+    name: Always run job
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Always run
+        run: |
+          echo "This is run to prevent the workflow from showing an error if the wheels job is not run and no jobs run is an error."
+          echo "See https://github.community/t5/GitHub-Actions/Workflow-is-failing-if-no-job-can-be-ran-due-to-condition/m-p/38085"
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ by many projects managed by the [Kivy Organization](https://kivy.org/#organizati
 [![OSX Unittests Status](https://github.com/kivy/kivy/workflows/OSX%20Unittests/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22OSX+Unittests%22)
 [![Windows wheels Status](https://github.com/kivy/kivy/workflows/Windows%20wheels/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22Windows+wheels%22)
 [![Manylinux wheels Status](https://github.com/kivy/kivy/workflows/Manylinux%20wheels/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22Manylinux+wheels%22)
+[![armv7l wheels Status](https://github.com/kivy/kivy/workflows/armv7l%20wheels/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22armv7l+wheels%22)
 [![OSX wheels Status](https://github.com/kivy/kivy/workflows/OSX%20wheels%2Fapp/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22OSX+wheels%2Fapp%22)
 
 Installation, Documentation and Examples
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ by many projects managed by the [Kivy Organization](https://kivy.org/#organizati
 [![OSX Unittests Status](https://github.com/kivy/kivy/workflows/OSX%20Unittests/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22OSX+Unittests%22)
 [![Windows wheels Status](https://github.com/kivy/kivy/workflows/Windows%20wheels/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22Windows+wheels%22)
 [![Manylinux wheels Status](https://github.com/kivy/kivy/workflows/Manylinux%20wheels/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22Manylinux+wheels%22)
-[![armv7l wheels Status](https://github.com/kivy/kivy/workflows/armv7l%20wheels/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22armv7l+wheels%22)
+[![Raspberry Pi wheels Status](https://github.com/kivy/kivy/workflows/RPi%20wheels/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22RPi+wheels%22)
 [![OSX wheels Status](https://github.com/kivy/kivy/workflows/OSX%20wheels%2Fapp/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22OSX+wheels%2Fapp%22)
 
 Installation, Documentation and Examples
--- a/doc/sources/installation/installation-rpi.rst
+++ b/doc/sources/installation/installation-rpi.rst
@@ -125,7 +125,7 @@ _`Raspberry Pi 1-4 installation` on Raspbian Jessie/Stretch/Buster
     # or to recompile all files
     make force
 
-   Or on a Raspberry Pi 4 it is possible to use a precompiled wheel. The precompiled wheel can be downloaded from the latest `release <https://github.com/kivy/kivy/releases>`_. A wheel is also automatically build daily and can be downloaded here: `<https://kivy.org/downloads/ci/armv7l/kivy/>`_.
+   Or on a Raspberry Pi 4 it is possible to use a precompiled wheel. The precompiled wheel can be downloaded from the latest `release <https://github.com/kivy/kivy/releases>`_. A wheel is also automatically build daily and can be downloaded here: `<https://kivy.org/downloads/ci/raspberrypi/kivy>`_.
 
    The wheel can be installed as follows::
 
--- a/doc/sources/installation/installation-rpi.rst
+++ b/doc/sources/installation/installation-rpi.rst
@@ -3,9 +3,6 @@
 Installation on Raspberry Pi
 ============================
 
-You can install Kivy manually (recommended), or you can download and boot KivyPie on the
-Raspberry Pi. Both options are described below.
-
 Raspberry Pi 4 manual installation (On Raspbian Buster)
 ------------------------------------------------
 
@@ -234,16 +231,6 @@ Images to use::
     in "Unable to find any valuable Window provider" errors. If you experience
     this issue, please upgrade or consult `ticket #5360.
     <https://github.com/kivy/kivy/issues/5360>`_
-
-KivyPie distribution
---------------------
-
-KivyPie is a compact and lightweight Raspbian based distribution that comes
-with Kivy installed and ready to run. It is the result of applying the manual
-installation steps described above, with a few more extra tools. You can
-download the image from http://kivypie.mitako.eu/kivy-download.html and boot
-it on a Raspberry PI.
-
 
 Running the demo
 ----------------
--- a/doc/sources/installation/installation-rpi.rst
+++ b/doc/sources/installation/installation-rpi.rst
@@ -6,50 +6,134 @@ Installation on Raspberry Pi
 You can install Kivy manually (recommended), or you can download and boot KivyPie on the
 Raspberry Pi. Both options are described below.
 
-
-Manual installation (On Raspbian Jessie/Stretch)
+Raspberry Pi 4 manual installation (On Raspbian Buster)
 ------------------------------------------------
 
-In the following instructions, for Python 3, replace `python` with `python3`.
+#. If you have installed Raspbian with a desktop, then you can skip to `Kivy dependencies`_.
+
+   If you are using a the Raspbian Lite, then you first need to install X::
+
+    sudo apt-get install xserver-xorg
+
+#. Next install nodm from source, so it includes the following fix: `spanezz/nodm#1 <https://github.com/spanezz/nodm/pull/10>`_::
+
+    sudo apt-get install libpam0g-dev help2man libx11-dev debhelper
+    git clone https://github.com/slashblog/nodm.git
+    pushd nodm
+        git checkout d48a8f6266d3f464138e0e95b65896917c35c89f
+        wget http://deb.debian.org/debian/pool/main/n/nodm/nodm_0.13-5.debian.tar.xz
+        tar xf nodm_0.13-5.debian.tar.xz
+        sudo dpkg-buildpackage -us -uc -b
+    popd
+    sudo dpkg -i nodm_0.13-*_armhf.deb
+    sudo rm -rf nodm*
+
+#. Now enable graphical login::
+
+    sudo systemctl set-default graphical.target
+
+#. Create a small test app::
+
+    cat <<EOF > ~/app.py
+    from kivy.app import App
+    from kivy.uix.button import Button
+
+
+    class TestApp(App):
+
+        def build(self):
+            return Button(text='hello world')
+
+
+    if __name__ == '__main__':
+        TestApp().run()
+    EOF
+
+#. It is recommend to use the the logger functionality instead of print statements: `<https://kivy.org/doc/stable/api-kivy.logger.html>`_, so they show up in the log.
+
+   A small script can then be used to follow the latest log and restart the application by using the ``--restart`` argument::
+
+    #!/bin/bash -e
+
+    if [[ $* == *--restart* ]]; then
+        sudo service nodm restart
+        inotifywait -q ~/.kivy/logs -e create --format %w%f | xargs tail -f
+    else
+        ls -t -d ~/.kivy/logs/* | head -n1 | xargs tail -f
+    fi
+
+   Note that you need to install `inotify-tools` in order to use the restart functionality.
+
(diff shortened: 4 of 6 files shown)


matham:
Looks great, but I have two questions.

* How reliable are these cross builds?
* How do you designate these wheels to a specific pi version? E.g. I'm guessing this would only be usable on the Pi4? But then the name wouldn't indicate this to pip I'm guessing as unlike manylinux there's no standard!?

Lauszus:
> How reliable are these cross builds?

I have only tested it quickly on a Raspberry Pi 4 at work. I'll do some more tests the coming days.

> How do you designate these wheels to a specific pi version? E.g. I'm guessing this would only be usable on the Pi4? But then the name wouldn't indicate this to pip I'm guessing as unlike manylinux there's no standard!?

Hmm I am actually not sure if they will work with the older Raspberry Pi. It could be the case. I'll test that tomorrow. I believe it should work with any armv7l based board if they have compatible OpenGL drivers. Yes it's not a standard. Worst case I'll just create a separate repo and publish them there for every Kivy release.

Btw any idea why the Travis build is not triggering?

.travis.yml
```diff
@@ -159,7 +159,20 @@ script:
...
+        CONTAINER_ID=$(docker create kivy/kivy-armv7l);
+        docker cp "${CONTAINER_ID}":/kivy-wheel .;
+        cp kivy-wheel/Kivy-* wheelhouse/;
+        for name in wheelhouse/*armv7l*.whl; do
+          new_name="${name/$tag_name-/$wheel_name}";
+          if [ ! -f "$new_name" ]; then
+            cp -n $name $new_name;
+          fi;
+        done;
+        ls wheelhouse/;
```
Lauszus: Btw I intend to move this code up into the code above, but I just put it here, so it would get run by Travis.

matham:
Not sure why travis is not triggered. Feel free to fix the code to run with the other linux wheel building and I'll merge it and we'll see how it works.

I'm assuming you tested it locally?

Lauszus:
@matham I got the Travis build working, but it's exceeding the maximum time limit of 50 minutes: <https://travis-ci.org/kivy/kivy/jobs/603837010> :/ I'm considering using Github Actions CI instead, as there the time limit is 6 hours?

Yes I used an instance on AWS to build it and then installed and tested the wheel on a Raspberry Pi 4 locally.

Lauszus:
@matham I got it working by switching to Github Actions CI, as the cross compiling took 1 hr and 15 min and the maximum time for a Travis build is 50 min in the free version.

I managed to speed up the build a lot by creating a base image with all the apt-get dependencies: https://cloud.docker.com/repository/docker/lauszus/kivy-armv7l-base (source: https://github.com/Lauszus/kivy-armv7l-base) and adding piwheels as an extra index, so it does not have to compile Cython as well: https://github.com/kivy/kivy/pull/6568/commits/038b94f37b085f7f8c48f1535330872a342d5d37

Since I am not a contributor I can not enable Github Actions for this repo: https://github.com/kivy/kivy/actions, but you should be able to trigger the build once this PR is merged.

You can see a successful build at my fork: https://github.com/Lauszus/kivy/commit/5c79f3425a218ec9238a61488e2d4716b71c08a4/checks?check_suite_id=284901365 and here is the link to the build wheel: https://github.com/Lauszus/kivy/suites/284901365/artifacts/196353 for anyone that wants to test it. The wheel is automatically uploaded as an artifact every time the CI job is run, so this also allows people to quickly test the master branch without having to recompile Kivy.

I also enabled so the wheel will automatically get uploaded to the release page when a new tag is created.

Lauszus:
@matham anything blocking this from being merged? :)

matham:
Sorry for not getting back to you earlier - I have very little time these days for Kivy, and wasn't sure exactly how to handle this because of the github action switch.

Github actions 
-----------------

First gh actions was only publicly enabled recently so we couldn't have merged this before that anyway. But, also this is entirely different from our current workflow and the team would need to discuss many of these changes before we can merge it. 

First, we'd need to agree with switching to gh actions. I expect that will happen eventually, but cannot be sure until the team actually agrees. But even if/when we do switch, the specific workflow you have chosen is very different than the current one. E.g. we only generate wheels once a day rather than for each commit - because it takes so much longer to generate them. We don't upload directly to the release page on GH, but rather to our server. 

I except we'll take the time to re-write the full workflow then and am wary of changing it up until then.

wheels
--------

Regarding the actual wheels, I didn't have the time to test the wheel and would want some confirmation from multiple people that it works on a clean Pi and on which py versions it works. Because otherwise, we'll get a bunch of bug reports and we don't have the resources really to fix it atm. 

That's not to say I think we shouldn't generate these wheels somehow, just that I want more confirmation that it works. After that, and when we figure out the CI thing, we could add a link on the rpi page to the wheel on the server.

If you get it to work on travis, I would probably merge it because then we can just add it to the server and people can test it etc without major changes to the workflow. Otherwise, I'm not sure we can merge it yet.

matham:
So, we recently switched to using github actions. Please take a look at the workflows structure and update this PR similarly, if you have time.

There's one request I'd mention, which I didn't bring up before - to not use a docker image which is not under kivy's control. If we need to use a pre-compiled image beyond the standard OSs, it would need to be published by Kivy.

matham:
In addition, support for armv7l will be in the new manylinux2014 (https://www.python.org/dev/peps/pep-0599/#the-manylinux2014-policy, https://github.com/pypa/manylinux/issues/338), but that won't be ready for a bit.

Lauszus:
@matham okay, I'll update the PR when I have time. 

Yes I understand you point about the Docker image. I will simply remove it causing the build to take longer time, but since we are now using Github CI we will not be restricted by the 50 min build time of Travis.

Lauszus:
@matham I have now updated the PR, so now the yaml file is basically identical to the one for manylinux :)

I also addded the installation instructions for the Raspberry Pi 4, which should then close a bunch of issues: #6413, #6474 and #6487.

For anyone that is interrested, then the cross compiled wheel can be downloaded here: https://github.com/kivy/kivy/suites/364981942/artifacts/751103.

Lauszus:
Btw I changed `python` to `python3` several places in the documentation, as Python 2 will be removed in the next version of Kivy.

Lauszus:
Also shouldn't the `rename_wheels` move the wheel and not just copy it, so we do not duplicated artifacts? Or is that on purpose?

Lauszus:
I have now tested the wheel and everything seems to work fine :)

.ci/Dockerfile.armv7l
```diff
@@ -0,0 +1,19 @@
+ARG image=balenalib/armv7hf-debian:buster
+FROM $image
+
+COPY . /kivy
+WORKDIR /kivy
+
+RUN [ "cross-build-start" ]
+
+# Install dependencies.
+RUN /bin/bash -c 'source .ci/ubuntu_ci.sh && \
```
matham: Rather than generating a wheel for the default Python version, could you maybe generate wheels for multiple python versions like at https://www.piwheels.org/project/Kivy/ and like we do for manylinux? It is ok to not do that if it's too complicated, it's just a thought.

.github/workflows/armv7l_wheels.yml
```diff
@@ -0,0 +1,65 @@
...
+
+on:
+  push:
+  pull_request:
+  create:
+  schedule:
+    - cron:  '23 1 * * *'
+
+env:
+  KIVY_SPLIT_EXAMPLES: 1
```
matham: `KIVY_SPLIT_EXAMPLES=1` needs to be set in the environment of the docker container because setting it here doesn't get passed into the docker container. See https://github.com/kivy/kivy/pull/6656.

.github/workflows/armv7l_wheels.yml
```diff
@@ -0,0 +1,65 @@
...
+      run: |
+        source .ci/ubuntu_ci.sh
+        upload_file_to_server "$SERVER_IP" "armv7l/kivy/"
+    - name: Upload wheels as artifact
+      uses: actions/upload-artifact@master
+      if: github.event_name == 'pull_request'
+      with:
+        name: armv7l_wheels
+        path: dist
+    - name: Publish to PyPI
```
matham: I don't think we can upload armv71 wheels to pypi yet so please remove this section. Also, even if pypi does accept it, it's too early to do that yet. Maybe uploading to piwheels would be ok, if that is something we can do.

doc/sources/installation/installation-rpi.rst
```diff
@@ -6,50 +6,134 @@ Installation on Raspberry Pi
 You can install Kivy manually (recommended), or you can download and boot KivyPie on the
```
matham: Can you please remove any reference to KivyPie? I don't think we support this anymore so it's better to simply have the direct installation instructions.

doc/sources/installation/installation-rpi.rst
```diff
@@ -6,50 +6,134 @@ Installation on Raspberry Pi
...
 Raspberry Pi. Both options are described below.
 
+Raspberry Pi 4 manual installation (On Raspbian Buster)
```
matham: The word "manual" can be removed.

doc/sources/installation/installation-rpi.rst
```diff
@@ -6,50 +6,134 @@ Installation on Raspberry Pi
...
+------------------------------------------------
+
+#. If you have installed Raspbian with a desktop, then you can skip to `Kivy dependencies`_.
```
matham: Can you please expand a little more here on what is meant by "with a desktop"? That is likely to be confusing to people who haven't thought about running in a terminal vs. desktop so they wouldn't know if this applies to them.

doc/sources/installation/installation-rpi.rst
```diff
@@ -6,50 +6,134 @@ Installation on Raspberry Pi
...
+
+   If you are using a the Raspbian Lite, then you first need to install X::
```
matham: "a" is extra.

doc/sources/installation/installation-rpi.rst
```diff
@@ -6,50 +6,134 @@ Installation on Raspberry Pi
...
+    sudo sh -c 'echo "NODM_X_OPTIONS='\''-nolisten tcp'\''" >> /etc/default/nodm'
+    sudo sh -c 'echo "NODM_MIN_SESSION_TIME=60" >> /etc/default/nodm'
+    sudo sh -c 'echo "NODM_X_TIMEOUT=300" >> /etc/default/nodm'
+
+    # Start the app using nodm
+    echo '#!/bin/bash' > ~/.xsession
+    echo 'export DISPLAY=:0.0' >> ~/.xsession
+    echo "python3 ~/app.py" >> ~/.xsession
+
+#. For installing the _`Kivy dependencies` refer to step 1-2 in the `Raspberry Pi 1-3 manual installation`_ instructions below.
```
matham: Maybe break here and start a new sub section called "Installing Kivy"or something? Then this section can be named installing Kivy and the previous sub section can be named something like "Installing kivy headless prerequisites" and it flow better together rather than being one long instruction list.

matham: Or maybe this whole section should just apply on the headless install and the next section, `Raspberry Pi 1-3 manual installation` should also include Pi4, but instead of the instructions to install from master you say that for Pi4 you can use a wheel?

doc/sources/installation/installation-rpi.rst
```diff
@@ -6,50 +6,134 @@ Installation on Raspberry Pi
...
+
+#. Kivy can be installed either from source or from a precompiled wheel. Refer to step 3 in the `Raspberry Pi 1-3 manual installation`_ instructions below for installing Kivy from source.
+
+   The precompiled wheel can be downloaded from the latest `release <https://github.com/kivy/kivy/releases>`_ or downloaded from the latest `CI build <https://github.com/kivy/kivy/actions?query=workflow%3A%22armv7l+wheels%22>`_ by clicking on a completed "armv7l wheels/armv7l" build and then click on "Artifacts" in the top right corner.
```
matham: Instead of linking the CI build, you should link to https://kivy.org/downloads/ci/armv7l/kivy/ - that's where the wheels would be published.

doc/sources/installation/installation-rpi.rst
```diff
@@ -6,50 +6,134 @@ Installation on Raspberry Pi
...
+   The wheel can be installed as follows::
+
+    python3 -m pip install --upgrade --user wheel
+    python3 -m pip install --user *armv7l.whl
+
+_`Raspberry Pi 1-3 manual installation` (On Raspbian Jessie/Stretch/Buster)
 ------------------------------------------------
 
-In the following instructions, for Python 3, replace `python` with `python3`.
+In the following instructions, for Python 2, replace `python3` with `python`.
```
matham: You can maybe remove "In the following instructions, for Python 2, replace `python3` with `python`." and instead say that we don't support python 2 anymore? Because master doesn't support python 2 anymore.

matham:
>One thing we could improve would be to compile it for several version of Python. Currently I only compile it for Python 3.7.

Right, didn't see this when I reviewed. But there's no reason to hold this PR off for that.

>Btw I changed python to python3 several places in the documentation, as Python 2 will be removed in the next version of Kivy.

Python 2 support is gone from master, we can remove any Python2 reference altogether.

>Also shouldn't the rename_wheels move the wheel and not just copy it, so we do not get duplicated artifacts? Or is that on purpose?

No, because we need to have the wheel with the original name so we can link to it directly with a name that doesn't change every night. In addition to the dated wheels so we can keep around the wheels for the last few days. See e.g. https://kivy.org/downloads/ci/win/kivy/.

.github/workflows/armv7l_wheels.yml
Lauszus: See: 6d2ceb51de88ac5830b67cee667f03c28eda7e14

.github/workflows/armv7l_wheels.yml
Lauszus: Fixed in: 1678473d684941d4924c33cb7f8b478b9bf5de7c

doc/sources/installation/installation-rpi.rst
Lauszus: Fixed in: 6d2ceb51de88ac5830b67cee667f03c28eda7e14

Lauszus:
All your comments should be fixed now except for building the wheel using multiple Python versions. My plan is to build it for Python 3.5, 3.6 and 3.7, so it matches the Manylinux wheels.

matham:
Just FYI, for a PR, [build wheel XXX] must be in the PR title, not commit message. That's why it's not building wheels. The reason is that there's no easy way to get the commit message of a PR from the actions.

Also, from piwheels:

>Why are the wheel files are tagged with Armv6 and Armv7? What about Armv8?

>While the hardware in Raspberry Pi 1 and Zero is Armv6, Pi 2 is Armv7, Pi 3 and 4 are Armv8, the Raspbian operating system images provided on raspberrypi.org are made to support all three architectures (and are compatible with all Raspberry Pi products to date) by operating in Armv6 userland.

>However, wheels built on a Raspberry Pi 2 or 3 are tagged armv7l. Since wheels built on a Raspberry Pi 3 will work on a Pi 4, 3, 2, 1 or Zero, we simply provide Pi 3-built wheels renamed armv6l, with a few exceptions (some packages, like opencv and tensorflow, are built with optimisations available on Pi 3).

https://www.piwheels.org/faq.html

I now wonder if we should also make named copies like that for v6? Have you tried these wheels on a pi3 e.g.?

Lauszus:
> Just FYI, for a PR, [build wheel XXX] must be in the PR title, not commit message. That's why it's not building wheels. The reason is that there's no easy way to get the commit message of a PR from the actions.

Yes it does trigger, but it builds at my fork instead: https://github.com/Lauszus/kivy/commit/14c1783d1cb9f70ee90a3b7ccf1892f76e6df030/checks?check_suite_id=369873129 :)

> https://www.piwheels.org/faq.html
> I now wonder if we should also make named copies like that for v6? Have you tried these wheels on a pi3 e.g.?

Yes, I'll update the script, so it creates a copy with the `armv6l` suffix. That is what piwheels seems to do:

```bash
$ md5sum Kivy-1.11.1-cp37-cp37m-linux_armv*
5a38955577867e1ccfa3451f9537bce9  Kivy-1.11.1-cp37-cp37m-linux_armv6l.whl
5a38955577867e1ccfa3451f9537bce9  Kivy-1.11.1-cp37-cp37m-linux_armv7l.whl
```

Note that I downloaded the wheels from: https://www.piwheels.org/simple/kivy/

Lauszus:
Hmm Jessie fails, as the Cython version we currently use does not support Python 3.4: https://github.com/Lauszus/kivy/commit/ee83dfbfa602743d02bb3ceb4d23995037f44d86/checks?check_suite_id=369862780.

I've added Stretch, so that should cover most of the Raspberry Pi users.

.ci/Dockerfile.armv7l
Lauszus: Fixed in: https://github.com/kivy/kivy/pull/6568/commits/e512e1f66a283ffd20536f81fe30204b4a9a5c17

matham:
OK, one not thing, rather than using `depends`, you can specify a build matrix e.g. https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategy. Then you can just list all the images in the matrix and reference it in the build step as in the example in the link.

Lauszus:
@matham I have now modified the script to use a build matrix instead.

Lauszus:
Btw we could also compile wheels specifically for the Raspberry Pi 1-3, as you can get the Broadcom drivers here: https://github.com/raspberrypi/firmware.

.github/workflows/armv7l_wheels.yml
```diff
@@ -0,0 +1,60 @@
...
+      run: |
+        source .ci/ubuntu_ci.sh
+        rename_wheels
+    - name: Upload wheels to server
+      if: github.event_name != 'pull_request'
+      env:
+        UBUNTU_UPLOAD_KEY: ${{ secrets.UBUNTU_UPLOAD_KEY }}
+      run: |
+        source .ci/ubuntu_ci.sh
+        upload_file_to_server "$SERVER_IP" "armv7l/kivy/"
```
matham: Now that we also do armv61 wheels, I changed the name to be more reflective of the platform, so it should be `"raspberrypi/kivy/"`. See https://kivy.org/downloads/ci/raspberrypi/kivy/.

.github/workflows/armv7l_wheels.yml
```diff
@@ -0,0 +1,60 @@
+name: armv7l wheels
```
matham: Please also rename this to `RPi wheels.`

README.md
```diff
@@ -29,6 +29,7 @@ by many projects managed by the [Kivy Organization](https://kivy.org/#organizati
 [![OSX Unittests Status](https://github.com/kivy/kivy/workflows/OSX%20Unittests/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22OSX+Unittests%22)
 [![Windows wheels Status](https://github.com/kivy/kivy/workflows/Windows%20wheels/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22Windows+wheels%22)
 [![Manylinux wheels Status](https://github.com/kivy/kivy/workflows/Manylinux%20wheels/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22Manylinux+wheels%22)
+[![armv7l wheels Status](https://github.com/kivy/kivy/workflows/armv7l%20wheels/badge.svg)](https://github.com/kivy/kivy/actions?query=workflow%3A%22armv7l+wheels%22)
```
matham: Also here to RPi.

matham:
>Btw we could also compile wheels specifically for the Raspberry Pi 1-3, as you can get the Broadcom drivers here: https://github.com/raspberrypi/firmware.

If I read the piwheels FAQ correctly, shouldn't these wheels we generate automatically work on the other RPis already without needing to do anything more?

.github/workflows/armv7l_wheels.yml
Lauszus: But this should work on any armv6l platform, as far as I understand? The reason why the Raspberry Pi needed a special wheel before was because it used the proprietary Broadcom drivers.

Lauszus:
> If I read the piwheels FAQ correctly, shouldn't these wheels we generate automatically work on the other RPis already without needing to do anything more?

Currently they do not: https://github.com/piwheels/packages/issues/5.

.github/workflows/armv7l_wheels.yml
matham: I don't think that's correct. If you read their FAQ they explicitly say that it may not work on other armv6 platforms.

The reason is that unlike manylinux, where before publishing a wheel we use auditwheel to gather all the OS level deps and include them in the wheel - except for a PEP specified list of .so files that are considered part of the OS and must be available by default on any OS that supports manylinux. Here, we simply assume any OS level deps are available because the user explicitly installs them with apt. So this wheel makes the specific assumption that it has been done, and it's specific to the OS/OS family on which this has been built.

So we cannot assume it will work on other armv6 platforms, e.g. android, except by coincidence, until there's support for these platforms on auditwheel. That's why they are/won't be available on pypi: https://github.com/pypa/warehouse/issues/3668.

Lauszus: Ahh I see. I'll change the files.

Lauszus: See: b31d6baeaa18b1ed39a92f455f872a38299ad00c

Lauszus:
Hmm would it be better to create a symlink for the `armv6l` wheels or are you happy with the copies?

matham:
Thank you for all your work on this!

>Hmm would it be better to create a symlink for the armv6l wheels or are you happy with the copies?

I'm not sure we can upload symlinks to the server, so copies are better for now.

Lauszus:
You're welcome. Can't wait for the wheels to start popping up here: https://kivy.org/downloads/ci/raspberrypi/kivy/ :)

Lauszus:
@matham FYI I just tested the wheel on a Raspberry Pi 3 and it works fine when X is installed. I do not have physical access to it, so I can not check the performance. Anyway I'll look into cross compiling for Raspberry 1-3 using the Broadcom drivers next.

matham:
Nice. Could one wheel support all the versions?

Lauszus:
@matham no, as the Raspberry Pi 4 do not have them. FYI I have opened up a PR with for compiling wheels for Raspberry Pi 1-3 as well: https://github.com/kivy/kivy/pull/6662.

Issue #2002, opened 2014-03-22
Error in docs for Properties


sirpercival:
http://snag.gy/SrlfR.jpg

PR #6578, opened 2019-11-01
Revert  #4219 as it causes performance issue

quitegreensky:
https://github.com/kivy/kivy/pull/4219

These commits actually cause fps regression. I found this by doing some version comparisons. looking at getting this revert will solve the issue.

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -652,10 +652,7 @@ cdef class _WindowSDL2Storage:
             pass
 
     def flip(self):
-        win = self.win
-        with nogil:
-            SDL_GL_SwapWindow(win)
-            cgl.glFinish()
+        SDL_GL_SwapWindow(self.win)
 
     def save_bytes_in_png(self, filename, data, int width, int height):
         cdef SDL_Surface *surface = SDL_CreateRGBSurfaceFrom(
--- a/kivy/lib/sdl2.pxi
+++ b/kivy/lib/sdl2.pxi
@@ -604,7 +604,7 @@ cdef extern from "SDL.h":
     cdef SDL_GLContext SDL_GL_GetCurrentContext()
     cdef int SDL_GL_SetSwapInterval(int interval)
     cdef int SDL_GL_GetSwapInterval()
-    cdef void SDL_GL_SwapWindow(SDL_Window * window) nogil
+    cdef void SDL_GL_SwapWindow(SDL_Window * window)
     cdef void SDL_GL_DeleteContext(SDL_GLContext context)
 
     cdef int SDL_NumJoysticks()


kivy/uix/textinput.py
```diff
@@ -2252,6 +2252,13 @@ def _get_line_options(self):
 
     def _create_line_label(self, text, hint=False):
         # Create a label from a text, using line options
+
```
akshayaurora: what this have to do with the actual issue?

akshayaurora: what this have to do with the actual issue?

akshayaurora:
@quitegreensky how much of a regression? Can you add some data?

kivy/uix/textinput.py
quitegreensky: Oops! the last two commits are not related to issue. My mistake!

quitegreensky:
> 
> 
> @quitegreensky how much of a regression? Can you add some data?

The whole story began from [#2002](https://github.com/kivy/python-for-android/issues/2002). There are all informations here.

kivy/uix/textinput.py
AndreMiras: Then could you remove this commit if it's not related. Also try to squash your changes so it's only one commit.

kivy/core/window/_window_sdl2.pyx
```diff
@@ -652,10 +652,7 @@ cdef class _WindowSDL2Storage:
             pass
 
     def flip(self):
-        win = self.win
-        with nogil:
-            SDL_GL_SwapWindow(win)
-            cgl.glFinish()
+        SDL_GL_SwapWindow(self.win)
```
AndreMiras: I'm actually curious to know which change/line specifically made the regression and why

matham:
I'm worried about merging this, given that we didn't merge the last request to revert for the reasons stated there. We could maybe disable it on mobile, but that feels hacky.

Either way, it would be good to get confirmation that specifically reverting this commit fixes the fps issues.

akshayaurora:
It seems the original fix https://github.com/kivy/kivy/pull/4219 was for a specific situation that caused a blocking of the main app when using multiple background threads, and getting a gl block due to swap being deferred by default by the drivers.

Calling glFinish() ensured the swap on previous line happened immediately, however this also leads to a general slowdown as mentioned by a few people.

@matham @tito how about this be made a compile time option?

For people that know what they are doing they could enable this,  we could add information about it in the doc.

matham:
If we can confirm that disabling glfinish fixes the issue, then unless someone has better ideas why having glfinish would lead to a unnecessary slowdown, I think a compile, or runtime option is the only solution I can think of.

It's not great, hence why I thought it was hacky. Because when someone new to kivy runs into this issue, since it'll use glfinish by default, they probably won't know to lookup this issue and workaround. And they'd just assume it's a kivy problem.

Perhaps disabling glfinish by default is a better option? And have a config option to turn it on?

akshayaurora:
> Perhaps disabling glfinish by default is a better option? And have a config option to turn it on?

I would assume that this makes much more sense. However Let's add some real `tests` to the argument. Right now it's just user reports that too not with reproducible apps. Having tests we can refer to would make it easier to judge the issue. 

Is this issue only on android? Does it effect iOS? Desktops?

We probably should have these tests in kivy test suit even if they might not run every time?

inclement:
I'm strongly in favour of reverting this, it looks to me like we're taking a huge performance hit to improve a relatively rare use case. We can still improve it later.
I note that #4219 sat for over 3 years before eventually being merged, and people started complaining very shortly afterwards. I think that's strong evidence that #4219 fixes much less than it breaks.

AndreMiras:
Thanks for your feedback on this @inclement 
I'm actually already patching it on p4a for one project even though I didn't notice a big difference on my devices https://github.com/AndreMiras/EtherollApp/commit/c064fd0416dcad281c0f8ecaddc8bca010884f4e

matham:
Ok, I merged it. I guess if someone wants to implement a config option that optionally enables the original patch they are welcome.

PR #6579, opened 2019-11-01
Horizontal value track is offset from the center of Slider

DanTheMan2000:
Horizontal value track is offset from the center of the Slider. The value track is shifted down. The issue is most noticeable whenever the background_width/value_track_width are large.


DanTheMan2000:
@matham This is ready for review. Thanks

PR #6584, opened 2019-11-03
Multiple splitters size

### LINKED ISSUE #5714: Issue with multiple splitters spreading out of the parent widget
* Python: 3.5.2
* OS: linux
* Kivy: 1.10.0

### Description

I have been using the splitter widget to create some nice table, however very often if the splitters are resized to too large values, the other splitters go out of the parent widget without taking into account the min_size of other widget and the keep_within_parent.

This behavior also happens in the example splitter.py in the official examples.

### Code and Logs
Here is a simplified code which shows this behavior:
```
from kivy.lang import Builder

bl = Builder.load_string('''

BoxLayout:
    orientation: 'horizontal'
    Splitter:
        min_size: '40sp'
        sizable_from: 'right'
        keep_within_parent: True
        rescale_with_parent: True
        Button:
            text: "Btn1"
    Splitter:
        min_size: '40sp'
        sizable_from: 'right'
        keep_within_parent: True
        rescale_with_parent: True
        Button:
            text: "Btn2"
    Splitter:
        min_size: '40sp'
        sizable_from: 'right'
        keep_within_parent: True
        rescale_with_parent: True
        Button:
            text: "Btn3"
    Splitter:
        min_size: '40sp'
        sizable_from: 'right'
        keep_within_parent: True
        rescale_with_parent: True
        Button:
            text: "Btn4"
''')

runTouchApp(bl)
```
This is a simple app with 4 buttons within splitters.

To reproduce the behavior:

1. Increase the size of, let's say, button1. So far so good, the maximum size is limited by the min_size of other widgets, no problem here.
2. Now increase the size of button 2: the other buttons are pushed out of the window

Usually the first change of size is correct, but after the first change everything seems to be messed up.

Somehow there is still a maximum size for each splitter(I can't resize them as large as the window for instance although I didn't set max_size) but it seems to be calculated not properly.

Anyway, thanks for the work :)

crigaut:
Fixes #5714 
The problem is that once the size of a splitter is modified, its size_hint is set to None.
As a consequence, the size of the splitters cannot be controlled externally after that.

### DIFF
--- a/kivy/uix/splitter.py
+++ b/kivy/uix/splitter.py
@@ -368,7 +368,9 @@ class Splitter(BoxLayout):
         self.dispatch('on_release')
 
     def on_release(self):
-        pass
+        for slider in self.parent.children:
+            slider.size_hint_y = slider.height / self.parent.height
+            slider.size_hint_x = slider.width / self.parent.width
 
 
 if __name__ == '__main__':


gottadiveintopython:
The meaning of `size_hint` depends on the parent. This PR works only when a Splitter belongs to `FloatLayout` / `RelativeLayout` / `Screen`.

Issue #6538, opened 2019-09-27
When GridLayout(rows=3) has 4 children, the layout becomes 2x2


gottadiveintopython:
### Versions

* Python: 3.7.1
* OS: Linux Mint 18.2 Mate Edition (Ubuntu 16.04LTS)
* Kivy: v2.0.0.dev0, git-0d1156a, 20190927
* Kivy installation method:

### Description

```python
from kivy.app import runTouchApp
from kivy.factory import Factory

root = Factory.GridLayout(rows=3)
for i in range(4):
    root.add_widget(Factory.Label(text=str(i)))
runTouchApp(root)
```

![Screenshot at 2019-09-27 16-48-25](https://user-images.githubusercontent.com/26050135/65751976-c88aa300-e146-11e9-8b0a-d959d408dfb6.png)

I also tested with kv, but it was the same thing.

```python
from kivy.app import runTouchApp
from kivy.lang import Builder

root = Builder.load_string('''
GridLayout:
    rows: 3
    Label:
        text: '0'
    Label:
        text: '1'
    Label:
        text: '2'
    Label:
        text: '3'
''')
runTouchApp(root)
```

ZurMaD:
You can fix adding one row 

```
from kivy.app import runTouchApp
from kivy.factory import Factory

root = Factory.GridLayout(rows=4)
for i in range(4):
    root.add_widget(Factory.Label(text=str(i)))
runTouchApp(root)
```

gottadiveintopython:
Well, that solves nothing...

me2beats:
The problem is seem to apear from here
`_init_rows_cols_sizes`
```
if current_cols is None:
    current_cols = int(ceil(count / float(current_rows)))
elif current_rows is None:
     current_rows = int(ceil(count / float(current_cols)))
```
so one of quick solutions is add/patch this behavior
(assuming that without specifying rows you want to have 1 row by default)
```
class MyGrid(GridLayout):
    def __init__(self, **kw):
        self.cols = self.cols or 1
        self.rows = self.rows or 1

        super().__init__(**kw)
```
https://github.com/kivy/kivy/blob/master/kivy/uix/gridlayout.py

gottadiveintopython:
Thanks. When I watched attributes at [this point](https://github.com/kivy/kivy/blob/master/kivy/uix/gridlayout.py#L327), the values were

```
self._rows == [0, 0, 0]
self._cols == [0, 0]
self._rows == [None, None, None]
[...]

PR #6593, opened 2019-11-09
make GridLayout fill a column first when 'cols' is None

gottadiveintopython:
Current implementation of the `GridLayout` always fills a row first regardless `cols` is None or not. And this behavior causes #6538, the GridLayout puts `3` into the 2rd row instead of the 3rd row.

This PR changes that behavior. So the following code:

```python

root = Factory.GridLayout(rows=3)
for i in range(4):
    root.add_widget(Factory.Label(text=str(i)))
runTouchApp(root)
```

which produces the following result under the master branch:

before_6538
(fills a row first)

will produce the following result under this PR.

under_6538
(fills a column first)

### DIFF
--- a/kivy/tests/test_uix_gridlayout.py
+++ b/kivy/tests/test_uix_gridlayout.py
@@ -3,6 +3,7 @@ uix.gridlayout tests
 ========================
 '''
 
+import pytest
 import unittest
 from kivy.tests.common import GraphicUnitTest
 
@@ -50,6 +51,21 @@ class UixGridLayoutTest(GraphicUnitTest):
         gl.cols_minimum = {i: 10 for i in range(10)}
         gl.add_widget(GridLayout())
         self.render(gl)
+
+
+def test_fills_a_column_first_when_cols_is_None():
+    # github issue #6538
+    from kivy.uix.gridlayout import GridLayout
+    from kivy.uix.widget import Widget
+    gl = GridLayout(rows=3, size=(300, 300), pos=(0, 0))
+    for i in range(4):
+        gl.add_widget(Widget())
+    gl.do_layout()
+    children = tuple(reversed(gl.children))
+    assert children[0].pos == [0, 200]
+    assert children[1].pos == [0, 100]
+    assert children[2].pos == [0, 0]
+    assert children[3].pos == [150, 200]
 
 
 if __name__ == '__main__':
--- a/kivy/uix/gridlayout.py
+++ b/kivy/uix/gridlayout.py
@@ -95,6 +95,7 @@ from kivy.properties import NumericProperty, BooleanProperty, DictProperty, \
     BoundedNumericProperty, ReferenceListProperty, VariableListProperty, \
     ObjectProperty, StringProperty
 from math import ceil
+import itertools
 
 
 def nmax(*args):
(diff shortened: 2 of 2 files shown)


tshirtman:
I don't think that's a good api change, what happens if you still want to define a number of columns, but you want to fill col-first? you could compute the target number of row depending of the number of items, but that's a hack. Instead, i think it's better to add an `orientation` property, behaving like the one of StackLayout. So you could independently set the number of rows and columns, and the direction of filling, this would add the often-missing feature of reverse filling and avoid another number of hacks in applications that require it.

gottadiveintopython:
I updated the example in the first comment because it wasn't a good example of showing why I think `GridLayout` needs to be fixed.

> what happens if you still want to define a number of columns, but you want to fill col-first?

If `cols` is not None, `GridLayout` behaves same as before. This PR affects the behavior only when `cols` is None and `rows` is not None.

> Instead, i think it's better to add an orientation property

I agree with that. ok I'll try to implement it.

matham:
I don't have much to add, other than to say that I agree with tshirtman. Also, please make sure that recycleview's version of gridlayout is also working appropriately with the changes.

tshirtman:
> If cols is not None, GridLayout behaves same as before. This PR affects the behavior only when cols is None and rows is not None.
yes, that's what i meant, the feature you added is not available in this situation, that was my critique, sorry i wasn't clear enough. Glad you agree about the proposed api :). Thanks.

gottadiveintopython:
@tshirtman I did misread your statement. I interpreted `what happens ... col-first?` as `what happens ... row-first?`, which is the default behavior on both master branch and this PR. That's where the confusion coming from. Sorry. I understand it now.

@matham I got it.

Also, I wonder what is the layout should be if the user doesn't define a number of columns, but want to fill row-first, for example:
```python

root = Factory.GridLayout(rows=3, orientation='lr-tb')
for i in range(4):
    root.add_widget(Factory.Label(text=str(i)))
runTouchApp(root)
```

My guess could be:

Screenshot at 2019-12-17 16-10-58

Computing this kind of layout is bit complex. And I believe the user usually wants to fill col-first when defines a number of rows,  and usually wants to fill row-first when defines a number of columns. So I'm not sure the example above is worth implementing.

tshirtman:
Right, this is a weird case indeed, i think in this case it's ok to solve the contradiction by not having the asked number of rows, (here 2 instead of 3) and either leaving the missing row space unoccupied, or to squash it entirely, (probably better to keep the space unoccupied), it would be less surprising than the possible ordering you presented, i think i'd be ok with whatever maintainable code for the sane cases produces for the weird one. :)

gottadiveintopython:
Thank for your advice. I understand. I'm going to close this PR because adding `orientation` is totally different from original purpose of the PR.

gottadiveintopython:
@tshirtman , @matham 

Hi, again. I want to ask a question.

`GridLayout` has `cols_minimum` and `row_minimum`, which are `DictProperty`, and its key is the index of columns/rows. When `orientation` is set to `rl-bt`, index zero should still mean leftmost/topmost? or it should mean rightmost/bottommost? thx.

matham:
I think the former. The meaning of indices in one property should not be dependant on what value another property takes.

Even though this would mean a user would need to set the dict to different values depending on the ordering chosen, but it would at least be always consistent in its meaning.

With children, we don't have a choice and the order in children means different things depending on the ordering property. But we shouldn't make the meaning of more props dependent on other props than strickly necessary.

gottadiveintopython:
ok thanks! I got it.

PR #6622, opened 2019-11-25
Switch from Travis/Appveyor to GitHub Actions

matham:
Changes
---------

There are two changes in this PR:

* This switches from our current travis/appveyor CI to github actions.
* In addition, rather than always manually installing all kivy deps I added the following `extras_require`: `base, full, dev, base_src, full_src`. This means, that on you just need to do `pip install kivy[base]` without having to install any dep manually.

The meaning of **base** is that it installs all the deps we think is needed for a basic install on that OS (e.g. glew, sdl2, and angle on windows). **full** means all deps for the OS (e.g. glew, sdl2, angle, and gstreamer on windows). **dev** means all the deps if you develop kivy (e.g. pytest etc.). Finally, ***_src** is used to install the kivy_deps.xxx_dev needed to install kivy from source.

I switched the CI to use this and it makes life much simpler. I'll also update the install docs on windows at some point to reflect the simplified instructions.

**This only works for master, naturally**. Take a look at setup.py for the details.

With https://github.com/kivy/kivy-server/pull/15, it should simplify installing from master by a lot.

Github actions
-----------------

Github actions have much longer time limits (6 hours) and have a consistent interface to the CI. Among other features, this seems to make it a really good CI provider and worth switching over.

However, to make sure we are not **tied** to Github, I re-wrote all the CI using shell and powershell files, and the CI simply makes function calls to these files. So no actual code happens in the github yaml. If we need to switch in the future, it should be pretty easy.

Issues
-------

There are two issues :

* Windows + py3.5 **does not work** - kivy cannot be compiled on it. I ran into [this](https://stackoverflow.com/questions/40018405/cannot-open-include-file-io-h-no-such-file-or-directory) issue. I sent a bug report to github about it. Hopefully they'll fix it. But anyway by the next kivy release, we likely won't need to support 3.5.
* When kivy is in a folder where the parent and parent's parent are also named kivy (like on github/travis CI), cython cannot compile when you do `pip install -e folder_name`. I was able to replicate this locally, but only for kivy - kivy must be doing something strange with the include directories. As a **workaround** on the CI, I did some symlinking to break the naming. We didn't observer the issue before because we didn't install kivy like this on the CI but only set the python path.

Workflows
------------

There are 6 workflows: osx, win, and ubuntu, and for each we have unit tests and wheel generation.

* Unittests happen for each push, tag, and PR.
  * pep8 check happens only on linux platform.
  * Only on windows do we test all py3.6-3.8 and x86/x64, on all others we test using the latest python 3.x.
* Docs are generated on linux as before, but on each push, tag, and PR. In all cases, the generated docs are available as **artifacts** so you can see them. If it's master/stable-xxx it's also pushed to the server.
* Wheels, they are generated using a daily cron job for each platform and for a tag. In addition, they are generated on push if the head commit contains `[build wheel]` or `[build wheel platform_name]` where `platform_name` is one of `osx, win, linux`. For a PR, it needs to be in the PR title. If it's a PR, the wheels are available as artifact, otherwise, they are also pushed to the server.
* OSX App is generated like for wheels, except the commit name would contain `[build app osx]`. Also, it is never generated for a PR, only for tags or pushes.

Side notes
------------

* We may consider not building wheels for PRs ever, but it could be helpful.
* On osx, where the build downloads large files (>500MB, gst, sdl2), I made use of github's cache, so it would only need to be downloaded again, when we change the dependency versions, specified in the versions file under `.ci`.


matham:
So far, building the osx app seems to take a while - maybe it's hung. That's the only component that I'm unsure if it succeeded. But I don't think we should wait to merge based on that.

matham:
I disabled building osx app for now because we get `Finder got an error: AppleEvent timed out` at https://github.com/kivy/kivy-sdk-packager/blob/master/osx/create-osx-dmg.sh#L51.

PR #6629, opened 2019-11-30
Fix Exception when device name cant be decoded (maybe because of Chinese Caracters?)

D4SK:
This was the traceback I got...

stderr: Traceback (most recent call last):
stderr:   File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
stderr:     self.run()
stderr:   File "/usr/lib/python2.7/threading.py", line 754, in run
stderr:     self.__target(*self.__args, **self.__kwargs)
stderr:   File "/home/pi/klippy-env/local/lib/python2.7/site-packages/kivy/input/providers/hidinput.py", line 660, in _thread_run
stderr:     " " * 256).decode().strip()
stderr: UnicodeDecodeError: 'ascii' codec can't decode byte 0xe6 in position 0: ordinal not in range(128)

### DIFF
--- a/kivy/input/providers/hidinput.py
+++ b/kivy/input/providers/hidinput.py
@@ -656,8 +656,11 @@ else:
             fd = open(input_fn, 'rb')
 
             # get the controler name (EVIOCGNAME)
-            device_name = fcntl.ioctl(fd, EVIOCGNAME + (256 << 16),
-                                      " " * 256).decode().strip()
+            try:
+                device_name = fcntl.ioctl(fd, EVIOCGNAME + (256 << 16),
+                                        " " * 256).decode().strip()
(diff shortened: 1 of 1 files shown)


Zen-CODE:
So, a few thoughts.

* Does this error with Python3? The error message indicates that you are using Python2, and as python 2 is no longer supported, I don't think we should be adding try/except clauses that are relevant to Python 2 only.
* It is generally considered bad practise to have very board exception clauses. In this case, it would be be better to explicitly test for the "UnicodeDecodeError" error.

D4SK:
Just tested and indeed it works with Python3, so im closing this. 
Thanks for looking at it so quickly!

PR #6650, opened 2019-12-18
Make modal and dropdown consistent

matham:
This makes the behavior of modal and dropdown consistent with respect to auto-dismissing.

Currently they behave differently, and I think buggy. One auto-dismisses on touch down if clicked outside, while the other on touch up. Also, one eats all the touch move, while the other ignores them, unlike up/down touches.

In the past, with one touch a dropdown would be closed, but the same touch could also open something else. I don't think this is right - one touch should close it and do nothing else. Another touch should be required to do something new. This is the fix required in action bar.

Is there a reason for these differences? What this does is:

* Auto-dismiss happens when the touch down start outside the modal/dropdown - even if it's released inside. Otherwise, it stays up, even if the touch starts inside but ends outside. Consequently, it auto-dismisses in touch up.
* All touches that are outside the modal/dropdown are just eaten - modal already behaved this way so I just fixed dropdown to also do this. There's no reason for a touch outside to be passed on when the popup is visible.
* Once the touch down is going to cause dismissal - the touch is eaten, even if it's inside during move/up.
* Adds gui tests to test auto-dismiss.

### DIFF
--- a/kivy/tests/test_uix_actionbar.py
+++ b/kivy/tests/test_uix_actionbar.py
@@ -218,6 +218,11 @@ class ActionBarTestCase(GraphicUnitTest):
         self.assertTrue(group2.is_open)
         self.assertFalse(group1.is_open)
 
+        # click away from ActionBar and wait for it to disappear
+        TouchPoint(0, 0)
+        sleep(g2dd.min_state_time)
+        self.move_frames(1)
+
         # click on Group 1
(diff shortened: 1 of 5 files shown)


matham:
I updated the behavior after taking a closer look at how gmail behaves. The only difference is that gmail closes the dropdown immediately during a touch down when it's outside the dropdown, while I wait and close it in touch up.

PR #6658, opened 2019-12-21
Fixes incorrect path to kivy garden libs on iOS

Dirk-Sandberg:
When working on iOS, the path to the garden module is fixed. Assumes garden flowers have been added to the project by using `garden install --app flower_name`, which adds the flower to a folder named `lib`, which should be in the same directory as main.py

### DIFF
--- a/kivy/garden/__init__.py
+++ b/kivy/garden/__init__.py
@@ -140,6 +140,7 @@ import sys
 import imp
 from os.path import dirname, join, realpath, exists, abspath
 from kivy import kivy_home_dir
+from kivy.utils import platform
 import kivy
 
 #: system path where garden modules can be installed
@@ -151,6 +152,12 @@ if getattr(sys, 'frozen', False) and getattr(sys, '_MEIPASS', False):
     garden_app_dir = join(realpath(sys._MEIPASS), 'libs', 'garden')
 else:
     garden_app_dir = join(realpath(dirname(sys.argv[0])), 'libs', 'garden')
+#: Fixes issue #4030 in kivy where garden path is incorrect on iOS
+if platform == "ios":
+    from os.path import join, dirname
+    import __main__
+    main_py_file = __main__.__file__
+    garden_app_dir = join(dirname(main_py_file), 'libs', 'garden')
 
 
 class GardenImporter(object):


kivy/garden/__init__.py
```diff
@@ -151,6 +152,12 @@
     garden_app_dir = join(realpath(sys._MEIPASS), 'libs', 'garden')
 else:
     garden_app_dir = join(realpath(dirname(sys.argv[0])), 'libs', 'garden')
+#: Fixes issue #4030 in kivy where garden path is incorrect on iOS
+if platform == "ios":
+    from os.path import join, dirname
+    import __main__
+    main_py_file = __main__.__file__
```
matham: Why can't we use `import kivy.garden`? I'm weary about `import __main__`  because I'm not sure if this will always work.

Dirk-Sandberg: Not sure exactly what you mean, could you clarify with an example maybe? The fix I suggest is from Tito's answer here: https://github.com/kivy-garden/garden.mapview/issues/7#issuecomment-188471962

matham: Yes, but you changed tito's code by adding `import __main__` and I'm not sure that's correct. Could you not use the identical code that tito posted or was there a reason you added you're changing it?

Dirk-Sandberg: Ah. The goal is to reference the folder where `main.py` lives, because that's where the `lib` folder should be installed. (Assuming `main.py` is directly inside your project folder, and the user runs `garden install --app flower_name` from inside that project folder)

In Tito's example, that code is to be put in your `main.py` file, so `dirname(__file__)` refers to the directory that `main.py` is in. 
To make this a general fix, I needed to replace `__file__` with something that referred to the `main.py` file being run.

Perhaps there's a better way to refer to the folder where `main.py` lives?

matham: Oh, ok, I understand. For some reason I thought this was an issue when it's installed with `garden install --kivy flower_name`. Should've read the text more carefully.

I'm assuming you tested this on ios and it works?

Dirk-Sandberg: Yep, it worked for me on iOS.

PR #6677, opened 2019-12-31
Remove `id` from Widget.

matham:


tito:
IMO, you could also remove the ID selector in KV lang as well

matham:
Is the above change what you had in mind? I didn't even know that existed...

tito:
Yep

Issue #6665, opened 2019-12-25
ScrollView does not have an option for enabling overscroll effects when content fits in the viewport. It should.


elver:
How can I enable overscroll in scrollview for cases where the list is shorter than the viewport? Right now overscroll effects (and therefore overscroll-based refreshing as is common on touch interfaces) work great when the list does not fit on the screen, but when it does, there is no visual overscroll.

This problem can be seen most easily with this example: https://github.com/HeaTTheatR/KivyMD/wiki/Components-Refresh-Layout

Just limit the number of lines to about 5 so they can fit on the screen at once:

```python
    def set_list(self):
        names_icons_list = list(md_icons.keys())[self.x : self.y]
        for name_icon in names_icons_list[:5]:
            self.root.ids.box.add_widget(ItemForList(icon=name_icon, text=name_icon))
```

When creating the scrollview content by hand, there's a workaround as provided on discord by tshirtman:

```
ScrollView:
    id: scrollview
    BoxLayout:
        orientation: 'vertical'
        size_hint_y: None,
        height: self.minimum_height
        BoxLayout:
            id: content
            size_hint_y: None
            height: self.minimum_height
                # your current content
        Widget:
            size_hint_y: None
            height: max(0, scrollview.height - content.height + 1)
```

But in my case I'm using a RecycleView-derived class, so things get a lot more complicated and the fix is not as trivial.

* Python: 3.8
* OS: Linux Mint
* Kivy: dev
* Kivy installation method: pip install git+https://github.com/kivy/kivy.git@master

PR #6678, opened 2020-01-02
Add always_enable_overscroll property on scrollview

tshirtman:
fix: #6665

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -262,7 +262,7 @@ class ScrollView(StencilView):
         return (self.do_scroll_x, self.do_scroll_y)
 
     def _set_do_scroll(self, value):
-        if type(value) in (list, tuple):
+        if isinstance(value, (list, tuple)):
             self.do_scroll_x, self.do_scroll_y = value
         else:
             self.do_scroll_x = self.do_scroll_y = bool(value)
@@ -274,6 +274,21 @@ class ScrollView(StencilView):
 
     :attr:`do_scroll` is a :class:`~kivy.properties.AliasProperty` of
     (:attr:`do_scroll_x` + :attr:`do_scroll_y`)
+    '''
+
+    always_overscroll = BooleanProperty(True)
+    '''Make sure user can overscroll even if there is not enough content
+    to require scrolling.
+
+    This is useful if you want to trigger some action on overscroll, but
+    there is not always enough content to trigger it.
+
+    :attr:`always_overscroll` is a
+    :class:`~kivy.properties.BooleanProperty` and defaults to `True`.
+    versionadded:: 2.0.0
+    The option was added and enabled by default, set to False to get the
+    previous behavior of only allowing to overscroll when there is
+    enough content to allow scrolling.
     '''
 
     def _get_vbar(self):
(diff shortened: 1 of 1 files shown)


matham:
Shouldn't this be True by default? Also, I think `always_overscroll` is probably better as it's more concise. Although it would be nice to be more specific than `always`, as that's quite vague as to what it means. Maybe even `overscroll_smaller_viewport`, although it's a bit wordy.

tshirtman:
I figured making it True by default could be an unwanted change in behavior, and i tend to be conservative about these things, even if i think it's indeed a better default behavior, if you think it's good enough to allow people who don't want this behavior to disable it, then i'm fine with switching the default value.

the always seems important to me as currently we have overscroll only in the specific situation where the content is big enough to allow scrolling, this just makes the behavior permanent.

Also, it did remind me that it's impossible to overscroll with the mousewheel currently, (and with this patch it's now possible… but only if the content is small enough to prevent normal scrolling :grimacing: :laughing:) i think this should be changed and mouse scroll should work the same as touch scroll in that regard, any opinion?

tshirtman:
Renamed the property, and enabled it by default, documented the fact that you should disable it to get the previous behavior back.

matham:
Looks ok if you're happy with it!?

kivy/uix/scrollview.py
```diff
@@ -276,6 +276,21 @@ def _set_do_scroll(self, value):
...
+    '''Make sure user can overscroll even if there is not enough content
+    to require scrolling.
+
+    This is useful if you want to trigger some action on overscroll, but
+    there is not always enough content to trigger it.
+
+    :attr:`always_overscroll` is a
+    :class:`~kivy.properties.BooleanProperty` and defaults to `True`.
+    versionadded:: 2.0.0
+    The option was added and enabled by default, set to False to get the
```
pythonic64, line 289: @tshirtman Shouldn't there be a new line before `versionadded` tag?

tshirtman: hm, maybe, i'll add it along the unittest fix, good call.

pythonic64: Maybe docs won't be generated property without new line.
Also, better readability :)

tshirtman: yes, and the `..` before was missing as well, good thing you mentioned it.

idevo89:
Scrollview and recycleview do not work properly when content size is smaller than parent size.

**always_overscroll=True & content size bigger**
ok

**always_overscroll=True & content size smaller**
Not ok, scrolling down scrolls the scrollview (recycleview) off the screen

**always_overscroll=False & content size bigger**
ok

**always_overscroll=False & content size smaller**
ok untill you resize down so content size is bigger than parent, then you scroll won't work

**To Reproduce**
```python
if __name__ == '__main__':
    from kivy.app import App

    class ScrollViewApp(App):

        def build(self):
            layout1 = GridLayout(cols=1, spacing=10, size_hint=(None, None))
            layout1.bind(minimum_height=layout1.setter('height'),
                         minimum_width=layout1.setter('width'))
            for i in range(40):
                btn = Button(text=str(i), size_hint=(None, None),
                             size=(200, 100))
                layout1.add_widget(btn)
            scrollview1 = ScrollView(bar_width='2dp', smooth_scroll_end=10)
            scrollview1.add_widget(layout1)

            layout2 = GridLayout(cols=1, spacing=10, size_hint=(None, None))
            layout2.bind(minimum_height=layout2.setter('height'),
                         minimum_width=layout2.setter('width'))
            for i in range(4):
                btn = Button(text=str(i), size_hint=(None, None),
                             size=(200, 100))
                layout2.add_widget(btn)
            scrollview2 = ScrollView(scroll_type=['bars'],
                                     bar_width='9dp',
                                     scroll_wheel_distance=100)
            scrollview2.add_widget(layout2)

            layout3 = GridLayout(cols=1, spacing=10, size_hint=(None, None))
            layout3.bind(minimum_height=layout3.setter('height'),
                         minimum_width=layout3.setter('width'))
            for i in range(40):
                btn = Button(text=str(i), size_hint=(None, None),
                             size=(200, 100))
                layout3.add_widget(btn)
            scrollview3 = ScrollView(bar_width='2dp', smooth_scroll_end=10,
                                     always_overscroll=False)
            scrollview3.add_widget(layout3)

            layout4 = GridLayout(cols=1, spacing=10, size_hint=(None, None))
            layout4.bind(minimum_height=layout4.setter('height'),
                         minimum_width=layout4.setter('width'))
            for i in range(4):
                btn = Button(text=str(i), size_hint=(None, None),
                             size=(200, 100))
                layout4.add_widget(btn)
            scrollview4 = ScrollView(scroll_type=['bars'],
                                     bar_width='9dp',
                                     scroll_wheel_distance=100,
                                     always_overscroll=False)
            scrollview4.add_widget(layout4)

            root = GridLayout(cols=4)
            root.add_widget(scrollview1)
            root.add_widget(scrollview2)
            root.add_widget(scrollview3)
            root.add_widget(scrollview4)
            return root

    ScrollViewApp().run()
```

PR #6681, opened 2020-01-03
Switch to flake8 and fix PEP8 issues

matham:


examples/widgets/effectwidget.py
```diff
@@ -51,8 +46,6 @@ def update_effectwidget(self, *args):
                 effects.append(InvertEffect())
             if text == 'mix':
                 effects.append(ChannelMixEffect())
-            if text == 'flash':
-                effects.append(FlashEffect())
```
AndreMiras: this is not really a PEP8 change, is it?

kivy/clock.py
```diff
@@ -408,7 +408,7 @@ def _get_sleep_obj():
 
         _kernel32 = ctypes.windll.kernel32
 
-        def _get_sleep_obj():
+        def _get_sleep_obj():  # noqa
```
AndreMiras: I would suggest to be more specific on the error/warning you want to ignore, flake8 supports that

kivy/clock.py
```diff
@@ -460,7 +460,7 @@ def _time():
 
                 return _time
 
-            _default_time = _libc_clock_gettime_wrapper()
+            _default_time = _libc_clock_gettime_wrapper()  # noqa
```
AndreMiras: same here about being more specific

Makefile
```diff
@@ -1,5 +1,5 @@
 PYTHON = python
-CHECKSCRIPT = kivy/tools/pep8checker/pep8kivy.py
+CHECKSCRIPT = -m flake8 --count --ignore=E125,E126,E127,E128,E402,E741,E731,W503,F401,W504,F841,E722 --show-source --statistics --max-line-length=80 --exclude=__pycache__,.tox,.git/,doc/,build/,.eggs/
```
AndreMiras: can't the ignore list be moved in `setup.cfg` instead?

examples/widgets/effectwidget.py
matham: More of flake style issue. `FlashEffect` doesn't seem to exist anywhere, so it complained and I removed it.

matham:
Yeah, about the py2 stuff I removed, flake complained about the objects referred to in the py2 code, e.g. `raw_input`, saying these don't exist in py3. So I removed them also.

Makefile
matham: Yeah, good idea. Will investigate.

examples/widgets/effectwidget.py
AndreMiras: Ahah true that! I thought it was the `ComparisonWidget` deleted above, but I misread, it's definitely a different class

matham:
Ok, fixed the issues you mentioned.

PR #6702, opened 2020-01-21
Adding support for non-file SVGs

ludwigkraatz:
source.setter now uses Svg.set_tree() to set the tree and execute .reload()
Svg() can now reloaded/initiated using set_tree(tree) when source is not a file.
Svg() Constructor therefore does not require source parameter - which now defaults to None.
source.setter is only executed during __init__ when parameter is given.

### DIFF
--- a/kivy/graphics/svg.pyx
+++ b/kivy/graphics/svg.pyx
@@ -366,7 +366,7 @@ cdef class Svg(RenderContext):
     """Svg class. See module for more informations about the usage.
     """
 
-    def __init__(self, source, anchor_x=0, anchor_y=0,
+    def __init__(self, source=None, anchor_x=0, anchor_y=0,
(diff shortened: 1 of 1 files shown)


kivy/graphics/svg.pyx
```diff
@@ -510,14 +511,17 @@ cdef class Svg(RenderContext):
...
-            self.tree = parse(fd)
-            self.reload()
+            self.set_tree(parse(fd))
             end = time()
             Logger.debug("Svg: Loaded {} in {:.2f}s".format(filename, end - start))
         finally:
             self._source = filename
             fd.close()
 
+    def set_tree(self, tree):
```
tshirtman: A docstring to describe usage and add the :versionadded:: tag would be cool. :)

PR #6703, opened 2020-01-22
Fix https in python3.x

chris-hamberg:
Added code to set the operating system environmental variable SSL_CERT_FILE to cacert.pem for Android.

Solving an extremely nasty bug for HTTPS on Android deployments effecting UrlRequest and AsyncImage.

COMMENTS:

Non-invasive global solution for a nasty HTTPS request bug on Android
physical deployments using python 3.x.

This solution is required to enable Android functionality for both:
   kivy.network.urlrequest     
   kivy.uix.image.AsyncImage
                                                                                                   
                                                         
Notes about testing:
                                                                                                   
Solution implementation was manually verified for correctness in Python 3.7
on Xubuntu (Linux-Lite) 18.04, and for deployment on Android Slim 6 (ROM.)
                                                                                                   
Solution implementation, and integration was manually verified for                 
correctness in Python 3.8 on Xubuntu (Linux-Lite) 18.04 and deployment on
Android Slim 6 (ROM.)
                                                                                                   
(Solution was not verified for Python 2.x, and does not suppose to support
2.x, nor has it been verified as correct for iOS deployment (hence, the                   
conditional prefix.))

### DIFF
--- a/kivy/loader.py
+++ b/kivy/loader.py
@@ -42,6 +42,7 @@ from kivy.cache import Cache
 from kivy.core.image import ImageLoader, Image
 from kivy.compat import PY2, string_types
 from kivy.config import Config
+from kivy.utils import platform
 
 from collections import deque
 from time import sleep
@@ -49,6 +50,7 @@ from os.path import join
 from os import write, close, unlink, environ
 import threading
 import mimetypes
+import certifi
 
 # Register a cache for loader
 Cache.register('kv.loader', limit=500, timeout=60)
@@ -121,6 +123,12 @@ class LoaderBase(object):
         self._running = False
         self._start_wanted = False
         self._trigger_update = Clock.create_trigger(self._update)
+
+        try:
+            assert platform == 'android'
+            environ['SSL_CERT_FILE'] = certifi.where()
+        except AssertionError as ok:
+            pass
 
     def __del__(self):
         if self._trigger_update is not None:
--- a/kivy/loader.py
+++ b/kivy/loader.py
@@ -124,12 +124,9 @@ class LoaderBase(object):
         self._start_wanted = False
         self._trigger_update = Clock.create_trigger(self._update)
 
-        try:
-            assert platform == 'android'
+        if platform == 'android':
             import certifi
             environ['SSL_CERT_FILE'] = certifi.where()
-        except AssertionError as ok:
-            pass
 
     def __del__(self):
         if self._trigger_update is not None:
--- a/kivy/loader.py
+++ b/kivy/loader.py
@@ -50,7 +50,7 @@ from os.path import join
 from os import write, close, unlink, environ
 import threading
 import mimetypes
-import certifi
+
 
 # Register a cache for loader
 Cache.register('kv.loader', limit=500, timeout=60)
@@ -126,6 +126,7 @@ class LoaderBase(object):
 
         try:
             assert platform == 'android'
+            import certifi
             environ['SSL_CERT_FILE'] = certifi.where()
         except AssertionError as ok:
             pass
--- a/kivy/loader.py
+++ b/kivy/loader.py
@@ -126,7 +126,7 @@ class LoaderBase(object):
 
         if platform == 'android':
             import certifi
-            environ['SSL_CERT_FILE'] = certifi.where()
+            environ.setdefault('SSL_CERT_FILE', certifi.where())
 
     def __del__(self):
         if self._trigger_update is not None:
--- a/kivy/network/urlrequest.py
+++ b/kivy/network/urlrequest.py
@@ -88,7 +88,6 @@ from kivy.clock import Clock
 from kivy.weakmethod import WeakMethod
 from kivy.logger import Logger
 from kivy.utils import platform
-import certifi
 
 
 # list to save UrlRequest and prevent GC on un-referenced objects
@@ -211,13 +210,18 @@ class UrlRequest(Thread):
         self._chunk_size = chunk_size
         self._timeout = timeout
         self._method = method
-        self.ca_file = ca_file if (platform != 'android') else (
-                ca_file or certifi.where())
         self.verify = verify
         self._proxy_host = proxy_host
         self._proxy_port = proxy_port
         self._proxy_headers = proxy_headers
         self._cancel_event = Event()
+
+        try:
+            assert platform == 'android'
+            import certifi
+            self.ca_file = ca_file or certifi.where()
+        except AssertionError as ok:
+            self.ca_file = ca_file
 
         #: Url of the request
         self.url = url
--- a/kivy/network/urlrequest.py
+++ b/kivy/network/urlrequest.py
@@ -216,11 +216,10 @@ class UrlRequest(Thread):
         self._proxy_headers = proxy_headers
         self._cancel_event = Event()
 
-        try:
-            assert platform == 'android'
+        if platform == 'android':
             import certifi
             self.ca_file = ca_file or certifi.where()
-        except AssertionError as ok:
+        else:
             self.ca_file = ca_file
 
         #: Url of the request
--- a/kivy/network/urlrequest.py
+++ b/kivy/network/urlrequest.py
@@ -87,6 +87,8 @@ except ImportError:
 from kivy.clock import Clock
 from kivy.weakmethod import WeakMethod
 from kivy.logger import Logger
+from kivy.utils import platform
+import certifi
 
 
 # list to save UrlRequest and prevent GC on un-referenced objects
@@ -209,7 +211,8 @@ class UrlRequest(Thread):
         self._chunk_size = chunk_size
         self._timeout = timeout
         self._method = method
-        self.ca_file = ca_file
+        self.ca_file = ca_file if (platform != 'android') else (
+                ca_file or certifi.where())
         self.verify = verify
         self._proxy_host = proxy_host
         self._proxy_port = proxy_port


tshirtman:
Not really a fan of this solution, isn't it easier to depend on certify, and use the environ fix automatically if we are on android/ios? (or maybe all the time, if there is no issue doing that).

chris-hamberg:
> Not really a fan of this solution, isn't it easier to depend on certify, and use the environ fix automatically if we are on android/ios? (or maybe all the time, if there is no issue doing that).

import kivy

is a lot more elegant than

import certify, os
os.environ['SSL_CERT_FILE'] = certifi.where()

...in every application.

Also, this is the only way that I could get the solution to work. I tried several architectures, and it took a tremendous amount of effort to get this to work.

chris-hamberg:
> Not really a fan of this solution, isn't it easier to depend on certify, and use the environ fix automatically if we are on android/ios? (or maybe all the time, if there is no issue doing that).

As an aside I also have implemented solutions for GPS, and programmatically requesting permissions (as required by the most current Android API.) But I still have yet to integrate those.

Also, one should note that this HTTPS bug was extremely difficult to debug in my experience, and was actually the worst problem that I have ever seen in programming. There is absolutely no debugging feedback without the solution. Logcat reports nothing, and documentation is sparse. I would consider that for new comers to Kivy such a solution is absolutely required. I considered giving up several times.

tshirtman:
i agree it's more elegant, but this could be done in kivy itself, when detecting we are on android, it would just require to depend on certify (at least on android), i'd prefer if we can avoid vendoring a file from certify that is likely to evolve in the future, requiring us to update it.

chris-hamberg:
> Not really a fan of this solution, isn't it easier to depend on certify, and use the environ fix automatically if we are on android/ios? (or maybe all the time, if there is no issue doing that).

Finally, note that certifi is doing exactly what this solution is doing, but in more lines of code. Look at the certifi implementation.

tshirtman:
Also i think you meant to add the cacert.pem file to the repository and you didn't? am i missing something?

chris-hamberg:
> i agree it's more elegant, but this could be done in kivy itself, when detecting we are on android, it would just require to depend on certify (at least on android), i'd prefer if we can avoid vendoring a file from certify that is likely to evolve in the future, requiring us to update it.

Initially I attempted to import certify into kivy, but kivy won't allow it to happen.

tshirtman:
https://github.com/certifi/python-certifi/commits/master/certifi/cacert.pem

as you can see, this file evolves pretty often, and i would prefer if we could avoid copying it in kivy.

> Initially I attempted to import certifi into kivy, but kivy won't allow it to happen.
interesting, do you have any info about what prevented it?

chris-hamberg:
> https://github.com/certifi/python-certifi/commits/master/certifi/cacert.pem
> 
> as you can see, this file evolves pretty often, and i would prefer if we could avoid copying it in kivy.
> 
> > Initially I attempted to import certifi into kivy, but kivy won't allow it to happen.
> > interesting, do you have any info about what prevented it?

I would have to reproduce the error, but it was similar to missing buildozer requirements

No you aren't missing anything regarding the missing cacert.pem. It is in my local repo, and was in my local repo upon committing, merging, and pushing. But I can't even find the remote branch now to see where it is missing!

chris-hamberg:
> https://github.com/certifi/python-certifi/commits/master/certifi/cacert.pem
> 
> as you can see, this file evolves pretty often, and i would prefer if we could avoid copying it in kivy.
> 
> > Initially I attempted to import certifi into kivy, but kivy won't allow it to happen.
> > interesting, do you have any info about what prevented it?

It was a module not found error.

I was trying to place the application level fix:
environ['SSL_CERT_FILE'] = certifi.where()

As described by:
https://github.com/kivy/python-for-android/issues/1827

But after recompiling, and rebuilding the application several times, and even attempting to include certifi directly in the kivy package, buildozer always gave an import error.

chris-hamberg:
> https://github.com/certifi/python-certifi/commits/master/certifi/cacert.pem
> 
> as you can see, this file evolves pretty often, and i would prefer if we could avoid copying it in kivy.
> 
> > Initially I attempted to import certifi into kivy, but kivy won't allow it to happen.
> > interesting, do you have any info about what prevented it?

It did work when certifi was included in the buildozer.spec requirements field.

chris-hamberg:
> https://github.com/certifi/python-certifi/commits/master/certifi/cacert.pem
> 
> as you can see, this file evolves pretty often, and i would prefer if we could avoid copying it in kivy.
> 
> > Initially I attempted to import certifi into kivy, but kivy won't allow it to happen.
> > interesting, do you have any info about what prevented it?

Is there a way to import certifi into kivy at the kivy framework level? And if so, would refactoring the solution under those conditions be of interest? How can I get the compiler to accept certifi?

tshirtman:
> But after recompiling, and rebuilding the application several times, and even attempting to include certifi directly in the kivy package, buildozer always gave an import error.

You mean kivy, not buildozer right? I'm not surprised if it's buildozer, at it means you tried in the wrong place, but if it's in kivy, at the place of your current change, i'm actually surprised, and will likely try on my side to try to understand how that's failing.

chris-hamberg:
> > But after recompiling, and rebuilding the application several times, and even attempting to include certifi directly in the kivy package, buildozer always gave an import error.
> 
> You mean kivy, not buildozer right? I'm not surprised if it's buildozer, at it means you tried in the wrong place, but if it's in kivy, at the place of your current change, i'm actually surprised, and will likely try on my side to try to understand how that's failing.
> 
> > Is there a way to import certifi into kivy at the kivy framework level? And if so, would refactoring the solution under those conditions be of interest? How can I get the compiler to accept certifi?
> 
> yes that would be my prefered solution.

I imported certifi into kivy/__init__.py

buildozer -v android debug

failed with certifi module not found when attempting to deploy to a physical device.

tshirtman:
> failed with certifi module not found when attempting to deploy to a physical device.

ok, thanks.

chris-hamberg:
> > failed with certifi module not found when attempting to deploy to a physical device.
> 
> ok, thanks.

I'll try it again, and post the result, if I can copy and paste it fast enough.

kivy/__init__.py
```diff
@@ -78,6 +79,49 @@
         '1.11.0 - the last Kivy release that still supports Python 2.')
 
 
+###############################################################################
```
matham: Can you please remove the following comments and instead post them to the pull request. Then, you can simply link to the PR in a single line comment.

As it is, the comments are too in depth and to be included in kivy because having them in the issue should be enough.

chris-hamberg:
> > failed with certifi module not found when attempting to deploy to a physical device.
> 
> ok, thanks.

This is what buildozer says:

[INFO]:    STDOUT:                                                                                    
  File "setup.py", line 12, in <module>	
    from kivy.utils import pi_version	
  File "/home/chris/Documents/Python/tester/.buildozer/android/platform/build-armeabi-v7a/build/other_builds/kivy-python3/armeabi-v7a__ndk_target_21/kivy/kivy/__init__.py", line 31, in <module>	
    import certifi	
ModuleNotFoundError: No module named 'certifi'

But certifi is in my venv.

tshirtman:
in your venv, sure, but is it in the apk? did you add it to your requirements in buildozer.spec or to the kivy deps in the recipe?

chris-hamberg:
> in your venv, sure, but is it in the apk? did you add it to your requirements in buildozer.spec or to the kivy deps in the recipe?



> in your venv, sure, but is it in the apk? did you add it to your requirements in buildozer.spec or to the kivy deps in the recipe?

I don't think it should be added to buildozer.spec. Although this is how I had to do it in an application that I had recently build.

How would it be added to kivy deps? The kivy/deps/__init__.py doesn't have any fields. Sorry, but I don't understand how deps is supposed to work. Do I just create an import statement in the __init__.py or do I need to add the certifi package to the deps directory?

tshirtman:
my idea was to add it here 
https://github.com/kivy/python-for-android/blob/master/pythonforandroid/recipes/kivy/__init__.py#L13

chris-hamberg:
> my idea was to add it here
> https://github.com/kivy/python-for-android/blob/master/pythonforandroid/recipes/kivy/__init__.py#L13

Yes. That is great. I'd be happy to test that out!

tshirtman:
I'm currently trying using a local recipe (copied the whole `kivy` parent dir of the file i linked to in a `local_recipes` dir, and told buildozer to use it), i'll have to add a patch to kivy with a modified version of your patch, importing certifi instead, it's currently building the distribution, we'll see how that goes.

chris-hamberg:
> I'm currently trying using a local recipe (copied the whole `kivy` parent dir of the file i linked to in a `local_recipes` dir, and told buildozer to use it), i'll have to add a patch to kivy with a modified version of your patch, importing certifi instead, it's currently building the distribution, we'll see how that goes.

Awesome! Please keep me posted and let me know the results. If everything works as desired, or if I need to do more to help with this issue, otherwise I can move forward to integrating the solution for programmatic permissions, thereby trivially enabling GPS for kivy users, and I assume other Android API access points that are currently pretty difficult to access since Google refactored its specification.

I really appreciate all of your help!

tshirtman:
https://github.com/kivy/python-for-android/pull/2055/files

this seems to work well for me, if you want to try it.

chris-hamberg:
> https://github.com/kivy/python-for-android/pull/2055/files
> 
> this seems to work well for me, if you want to try it.

I am getting ready to try that; although, with deps in python-for-android supporting certifi, I have devised the correct integration. Using certifi.where() in network/urlrequest.py initialization and setting SSL_CERT_FILE to certifi.where in the kivy/loader.py initialization.

I say that it is correct because it totally conceals the SSL issue from kivy user; adhering to the original kivy API specification. The kivy user does not need to write any additional code whatsoever for urlrequest, and asyncimage deployed on Android devices with this integration.

Would you be interested in taking a look at this? If so, would I need to submit a new pull request? The changes require the deps modification to python-for-android, and a small change to each of the classes in kivy.

Also, sorry, but git patches are a new concept for me, and I am having a hard time understanding it. Although I believe that I understand what it accomplishes. So I just modified my local forks of kivy and python-for-android with your changes, and built from that. That does produce the same end result, correct?

tshirtman:
the easiest way to test that PR of mine is to put  `fix_1827` in the `p4a.branch` option of your buildozer.spec, as it'll build the distribution using this version of p4a, and so apply the patch file (a patch is just a diff that can be applied to the original file), and make sure the distribution includes certifi.

I made the PR with the patch in P4A, but it could certainly be a better idea to directly put that change in kivy itself. About where to put it, i agree maybe putting it in the `kivy.network.urlrequest` module would make more sense, i think it would work just as well, but i didn't try.

chris-hamberg:
> the easiest way to test that PR of mine is to put `fix_1827` in the `p4a.branch` option of your buildozer.spec, as it'll build the distribution using this version of p4a, and so apply the patch file (a patch is just a diff that can be applied to the original file), and make sure the distribution includes certifi.
> 
> I made the PR with the patch in P4A, but it could certainly be a better idea to directly put that change in kivy itself. About where to put it, i agree maybe putting it in the `kivy.network.urlrequest` module would make more sense, i think it would work just as well, but i didn't try.

But how can I show you the alternative integration? Can I just push an update to the current PR 
 branch, or do I need to create a new PR altogether?

chris-hamberg:
> the easiest way to test that PR of mine is to put `fix_1827` in the `p4a.branch` option of your buildozer.spec, as it'll build the distribution using this version of p4a, and so apply the patch file (a patch is just a diff that can be applied to the original file), and make sure the distribution includes certifi.
> 
> I made the PR with the patch in P4A, but it could certainly be a better idea to directly put that change in kivy itself. About where to put it, i agree maybe putting it in the `kivy.network.urlrequest` module would make more sense, i think it would work just as well, but i didn't try.

It looks like kivy.network.urlrequest can be fixed with self.cafile = cafile or certifi.where() in the UrlRequest initialization.

It looks like AsyncImage has to use environ['SSL_CERT_FILE'] = certifi.where() since it uses a prepared request, which by the Python3 documentation doesn't have a cafile access point. So I placed that in the initialization of kivy.loader.LoaderBase.

This, with certifi in deps, worked as expected on deployment to my device.

tshirtman:
> But how can I show you the alternative integration? Can I just push an update to the current PR
branch, or do I need to create a new PR altogether?

I think you can push on top of this PR (or even squash and force push if we agree the alternative implementation is preferable).

chris-hamberg:
> > But how can I show you the alternative integration? Can I just push an update to the current PR
> > branch, or do I need to create a new PR altogether?
> 
> I think you can push on top of this PR (or even squash and force push if we agree the alternative implementation is preferable).

Great! I am currently building your patch. Once it is finished I will report the results, and then attempt pushing the new integration on top of this PR.

chris-hamberg:
> > But how can I show you the alternative integration? Can I just push an update to the current PR
> > branch, or do I need to create a new PR altogether?
> 
> I think you can push on top of this PR (or even squash and force push if we agree the alternative implementation is preferable).

I built with a fresh kivy fork using p4a.branch = fix_1827

This deployed successfully to my physical device, and the urlrequest and AsyncImage methods worked as expected.

It appears that the alternative integration, which assumes certifi is in the deps list in p4a, has pushed successfully. Please let me know what you think, or what else we need to do with this.

tshirtman:
> I built with a fresh kivy fork using p4a.branch = fix_1827
> 
> This deployed successfully to my physical device, and the urlrequest and AsyncImage methods worked as expected.

:tada: great, thanks for double checking.

> 
> It appears that the alternative integration, which assumes certifi is in the deps list in p4a, has pushed successfully. Please let me know what you think, or what else we need to do with this.

That seems good, though i think the import and usage should be conditional on being on android, or at least in try/except to avoid failing if it's missing on desktop.

Both solutions would be fine imho as we could keep just the the part of my PR that add certifi to dependencies, and so we would know it's going to be there on android.

chris-hamberg:
> > I built with a fresh kivy fork using p4a.branch = fix_1827
> > This deployed successfully to my physical device, and the urlrequest and AsyncImage methods worked as expected.
> 
>  great, thanks for double checking.
> 
> > It appears that the alternative integration, which assumes certifi is in the deps list in p4a, has pushed successfully. Please let me know what you think, or what else we need to do with this.
> 
> That seems good, though i think the import and usage should be conditional on being on android, or at least in try/except to avoid failing if it's missing on desktop.
> 
> Both solutions would be fine imho as we could keep just the the part of my PR that add certifi to dependencies, and so we would know it's going to be there on android.

I will make the recommended changes to check for android with try/except blocks. 

But the push is showing failed unittests; is that correct?

I strongly prefer this fix. Primarily because it yields consistent behavior across platforms. I created a huge application that consisted of 6 layers of architecture, and kivy gui. This was my first kivy project, and did all of the preliminary testing and debugging in Linux. Although the application is mobile. Everything just worked on Linux, using the API that the most recent fix enforces. But after nearly completing the application, and doing a first push to a physical device, I ran into this problem. As you may be well aware, just identifying the problem was a painful ordeal, and it took about 1 week of persistent effort to figure out what was wrong.

And that is why my bias leans towards a solution more like this; it will work the same on Linux or Android. So I do strongly prefer it, and personally recommend this version.

chris-hamberg:
> > I built with a fresh kivy fork using p4a.branch = fix_1827
> > This deployed successfully to my physical device, and the urlrequest and AsyncImage methods worked as expected.
> 
>  great, thanks for double checking.
> 
> > It appears that the alternative integration, which assumes certifi is in the deps list in p4a, has pushed successfully. Please let me know what you think, or what else we need to do with this.
> 
> That seems good, though i think the import and usage should be conditional on being on android, or at least in try/except to avoid failing if it's missing on desktop.
> 
> Both solutions would be fine imho as we could keep just the the part of my PR that add certifi to dependencies, and so we would know it's going to be there on android.

I tried to keep the code as elegant as possible. So I used a ternary conditional in UrlRequest instead of try/catch.

Please let me know your thoughts about this iteration!

chris-hamberg:
> > I built with a fresh kivy fork using p4a.branch = fix_1827
> > This deployed successfully to my physical device, and the urlrequest and AsyncImage methods worked as expected.
> 
>  great, thanks for double checking.
> 
> > It appears that the alternative integration, which assumes certifi is in the deps list in p4a, has pushed successfully. Please let me know what you think, or what else we need to do with this.
> 
> That seems good, though i think the import and usage should be conditional on being on android, or at least in try/except to avoid failing if it's missing on desktop.
> 
> Both solutions would be fine imho as we could keep just the the part of my PR that add certifi to dependencies, and so we would know it's going to be there on android.

The unit tests passed on this one too. That is so weird that they did not, on the previous push.

chris-hamberg:
> > I built with a fresh kivy fork using p4a.branch = fix_1827
> > This deployed successfully to my physical device, and the urlrequest and AsyncImage methods worked as expected.
> 
>  great, thanks for double checking.
> 
> > It appears that the alternative integration, which assumes certifi is in the deps list in p4a, has pushed successfully. Please let me know what you think, or what else we need to do with this.
> 
> That seems good, though i think the import and usage should be conditional on being on android, or at least in try/except to avoid failing if it's missing on desktop.
> 
> Both solutions would be fine imho as we could keep just the the part of my PR that add certifi to dependencies, and so we would know it's going to be there on android.

Ahh, I missed encapsulating the imports. Really sorry. Let me try pushing one more revision.

kivy/loader.py
```diff
@@ -122,6 +124,13 @@ def __init__(self):
         self._start_wanted = False
         self._trigger_update = Clock.create_trigger(self._update)
 
+        try:
+            assert platform == 'android'
```
tshirtman: I'm really curious about the reason of this construction, that seems exactly equivalent to the more straightforward `if platform == 'android'`, am i missing something?

chris-hamberg: try/except is generally considered more "pythonic" and should be used instead of conditional statements whenever possible. I read it in a book, but don't recall which book. It has something to do with the fundamental philosophy of Python; along the lines of "it's better to try and fail, then to ask permission" - to paraphrase

tshirtman: I think it's a common myth in the python community, and it's been addressed elsewhere 
https://www.reddit.com/r/Python/comments/3o36h8/would_it_be_better_to_use_tryexcept_rather_than/ points to 
https://mail.python.org/pipermail/python-dev/2014-March/133118.html
as reference, while not of biblical importance, Guido weighted on that, and i think we can consider a good reference on the philosophy of Python :).
> and I disagree with the position that EAFP is better than
LBYL, or "generally recommended" by Python.

there are situations where it's the best tool, and it's there to be used (especially when we don't have the easy possiblity to check for error conditions before calling the code). But here you use assert to do the test right after, so i don't think it's pythonic at all, but thanks for trying, i appreciate the intention, even if it comes from a misconception.

After this is switched to a simple test, I think i'd be fine with merging it, i'll update my own PR to p4a to make sure we always have certifi in kivy on android. Thanks for putting up with my feedback :).

chris-hamberg: > I think it's a common myth in the python community, and it's been addressed elsewhere
> https://www.reddit.com/r/Python/comments/3o36h8/would_it_be_better_to_use_tryexcept_rather_than/ points to
> https://mail.python.org/pipermail/python-dev/2014-March/133118.html
> as reference, while not of biblical importance, Guido weighted on that, and i think we can consider a good reference on the philosophy of Python :).
> 
> > and I disagree with the position that EAFP is better than
> > LBYL, or "generally recommended" by Python.
> 
> there are situations where it's the best tool, and it's there to be used (especially when we don't have the easy possiblity to check for error conditions before calling the code). But here you use assert to do the test right after, so i don't think it's pythonic at all, but thanks for trying, i appreciate the intention, even if it comes from a misconception.
> 
> After this is switched to a simple test, I think i'd be fine with merging it, i'll update my own PR to p4a to make sure we always have certifi in kivy on android. Thanks for putting up with my feedback :).

I agree it is mostly aesthetic preference. But what about handling the hypothetical edge case of a missing certifi when it is running on Android? Wouldn't the try/except (with an additional except block to handle missing a certifi) be necessary, anyways?

What does a missing certifi error look like? Would it be a missing .pem or module not found?

Or should there be no concern for a potential missing certifi on Android, and so we should just use conditional statements?

tshirtman: it would be a `ModuleNotFoundError`, in such case yeah we could totally try except for that in all cases, that's the alternative solution i was thinking about earlier, to me either/or, check platform and import, or try import and if it just silently ignore it, maybe with a warning in log, but that feels less clean than checking.

The current way (commit ab63037) seems clean to me, thanks.

chris-hamberg: So I suppose this is finished. Thank you, you have been super helpful throughout this process, and I am thankful that the Kivy developers have provided such a powerful and comprehensive framework supporting portable gui applications with Python.

I will now attempt to integrate GPS and permissions fixes to plyer (which are very much needed!) (This might take awhile!)

chris-hamberg: > it would be a `ModuleNotFoundError`, in such case yeah we could totally try except for that in all cases, that's the alternative solution i was thinking about earlier, to me either/or, check platform and import, or try import and if it just silently ignore it, maybe with a warning in log, but that feels less clean than checking.
> 
> The current way (commit [ab63037](https://github.com/kivy/kivy/commit/ab63037d0d35f2b16abaef016e0bb86970349351)) seems clean to me, thanks.

Do I need to re-request review for matham?

tshirtman:
https://github.com/kivy/python-for-android/pull/2055 updated

kivy/loader.py
```diff
@@ -122,6 +124,10 @@ def __init__(self):
         self._start_wanted = False
         self._trigger_update = Clock.create_trigger(self._update)
 
+        if platform == 'android':
+            import certifi
+            environ['SSL_CERT_FILE'] = certifi.where()
```
matham: I didn't follow the conversation, but can you summarize why this variable is set?

matham: Ah, I see https://github.com/kivy/python-for-android/issues/1827#issuecomment-500028459.

I'll trust @tshirtman judgement here, but I am a little weary of overriding this variable from within kivy. I.e. how do we know someone may not have set it to some specific value? Is this a kivy specific issue that we need to set it like this only from kivy and not elsewhere? What about non-kivy apps on android? Maybe py4a should set this?

And if this is a global issue, maybe this needs to be set in kivy's `__init__`?

chris-hamberg: Without setting this variable: a kivy application works out of the box on Linux. However, on physical deployment a black screen is rendered, and no error messages are reported in buildozer nor in logcat. Leading to an extremely difficult problem to debug for kivy users (literally the hardest problem I have ever seen; harder than any problem I have seen even in mathematics.)

This setting this variable causes consistent kivy behavior across Linux, and Android. Also it maintains the original intent of the kivy API on these access points.

chris-hamberg: > I didn't follow the conversation, but can you summarize why this variable is set?

Sorry, I misspoke. The AsyncImage issue only caused AsyncImages to fail to load, but did not prevent the application from rendering. UrlRequest caused the black screen.

chris-hamberg: > I didn't follow the conversation, but can you summarize why this variable is set?

Addressing this issue: https://github.com/kivy/python-for-android/issues/1827

tshirtman: Hm, if it's already set, should we override it? That's a fair point, i don't think it's likely to be an issue in practice, but we could check and only define it if it doesn't exist already, os.environ acting like a dict certainly has a setdefault method that would make this convenient.

matham: Yeah, and in the long-term, @andnovar [seems to think](https://github.com/kivy/python-for-android/issues/1827#issuecomment-500078114) he'd do it in py4a. But until then it should be ok to use `setdefault`.

chris-hamberg: > Ah, I see [kivy/python-for-android#1827 (comment)](https://github.com/kivy/python-for-android/issues/1827#issuecomment-500028459).
> 
> I'll trust @tshirtman judgement here, but I am a little weary of overriding this variable from within kivy. I.e. how do we know someone may not have set it to some specific value? Is this a kivy specific issue that we need to set it like this only from kivy and not elsewhere? What about non-kivy apps on android? Maybe py4a should set this?
> 
> And if this is a global issue, maybe this needs to be set in kivy's `__init__`?

On my Android

~$ echo SSL_CERT_FILE

Is empty.

Environment variables set in Python are not persistent, and are isolated to their process. A kivy application could set this value, thus overwriting the solution here, but doing so would break the contract given by this fix. So existing applications that have this statement in their main.py are unaffected by the fix. And new applications do not need to include the statement, although if they do, it won't cause any issues.

Specifically the issue solved occurs at urllib.request.Request in the kivy.loader.LoaderBase class, within the kivy framework.

I didn't look at p4a. But p4a should set this if it uses the urllib.request.Request to make https requests and fails. I believe that the causes of the issue occurs around the transition from Python 2, to Python 3, with porting urllib to Python 3, along with (at around the same time) the web adopting the HTTPS standard over HTTP.

I don't think this issue is a global issue, because it depends on how the request is implemented. The statement in __init__ does fix UrlRequest, for example. Or UrlRequest can just directly ask for the cacert.pem location since it uses the urllib.request.urlrequest interface. AsyncImage cannot directly ask for cacert.pem, and must use the environment variable because it uses the urllib.request.Request interface. Changing this, by refactoring the request code in LoaderBase, could solve the issue without having to set the environment variable, although I am leery of refactoring code that otherwise totally works in every other respect.

Putting it in __init__ does work, and was the originally submitted solution; because I couldn't get the build to include certifi in the APK (which was what I was actually trying to do at first.) Placing it in __init__ is better for kivy maintainers because it fully abstracts the problem into a single statement, but is worst for the API because a behavioral inconsistency still exists in that case: Linux applications would not require import kivy to make requests successfully, but Android applications would require import kivy. If the user doesn't know this they will still run into a similar problem that I faced, although that problem might be easier to deal with for new users with the kivy/__init__.py solution since a lot of existing tutorials on kivy have the user import kivy anyways. However, the current fix implements a consistent application behavior across all platforms, without having to write platform specific code, and I feel that this fully solves the problems that I, and many others have faced. What I mean is that, in all cases, on Linux you would not need to import kivy, but in the case of placing the fix in kivy/__init__.py Android HTTPS applications would have to have the extra import kivy statement; thus, platform specific code for kivy users.

TL;DR: this version completely solves the problem of needing platform specific code for kivy users in making https requests; the __init__ version attempts to lessen the severity of the problem, but doesn't really fully solve the problem, and still results in platform specific code. It was my original submission, because I thought it was the best we could do. But with tshirtmans help and knowledge, we were able to arrive at maintaining a 100% platform independent behavioral consistency. So I am super biased in favor of this version.

chris-hamberg: > Ah, I see [kivy/python-for-android#1827 (comment)](https://github.com/kivy/python-for-android/issues/1827#issuecomment-500028459).
> 
> I'll trust @tshirtman judgement here, but I am a little weary of overriding this variable from within kivy. I.e. how do we know someone may not have set it to some specific value? Is this a kivy specific issue that we need to set it like this only from kivy and not elsewhere? What about non-kivy apps on android? Maybe py4a should set this?
> 
> And if this is a global issue, maybe this needs to be set in kivy's `__init__`?

Sorry the comments are out of order. The TL;DR pertains to your most recent comment.

The comments above pertain to your second most recent comment. For some reason I could not see your most recent comment until, basically, just now.

chris-hamberg: > Ah, I see [kivy/python-for-android#1827 (comment)](https://github.com/kivy/python-for-android/issues/1827#issuecomment-500028459).
> 
> I'll trust @tshirtman judgement here, but I am a little weary of overriding this variable from within kivy. I.e. how do we know someone may not have set it to some specific value? Is this a kivy specific issue that we need to set it like this only from kivy and not elsewhere? What about non-kivy apps on android? Maybe py4a should set this?
> 
> And if this is a global issue, maybe this needs to be set in kivy's `__init__`?

We could check whether SSL_CERT_FILE is not empty before setting, as suggested by AndreMiras in kivy/python-for-android#1827. But it is empty by default according to Sirfanas.

Writing without checking shouldn't break anything, unless some other part of the framework is a determinate and setting it to a different value.

tshirtman: The only way i would see it not empty, would be that the user somehow manipulated the env themselves before importing kivy, and in such case we would be overriding their choice.

using setdefault instead of `=` avoids this corner case.

```
>>> import os
>>> os.environ.get('TEST')
>>> os.environ.setdefault('TEST', 'test')
'test'
>>> os.environ.get('TEST')
'test'
>>> os.environ.setdefault('TEST', 'test1')
'test'
>>> os.environ.get('TEST')
'test'
```

chris-hamberg: Hey guys. I've pushed the code using setdefault in place of the assignment statement, in the loader module.

tshirtman:
Ok, squashing and merging, need to merge the mentionned PR on P4A to be approved/merged too.

Thanks!

PR #6713, opened 2020-01-27
Fix memory leak for screen manager shader transitions

audryste:
Python garbage does not automatically free GPU ram.

Linux pc
[INFO ] [Kivy ] v1.11.1
[INFO ] [Python ] v2.7.12
Using shader transtion I get RAM memory leak in GPU. I used radeontop program to see memory leak.

There is a lot of created issues in Kivy repo about screenmanager transitions errors.  This should fix issues on RPI where GPU memory is even smaller. (That is why more related issues are created for rpi)

On platform where memory card uses shared memory leak might be more visible.

Note: gb.collect fixes issue alone but keeps a lot of memory usage, When setting to None a lot of memory is freed.


kivy/uix/screenmanager.py
```diff
@@ -530,6 +531,10 @@ def remove_screen(self, screen):
         self.manager.canvas.remove(self.render_ctx)
         self._remove_out_canvas()
         self.manager.real_add_widget(self.screen_in)
+        self.fbo_in = None
+        self.fbo_out = None
+        self.render_ctx = None
+        gc.collect()
```
matham: We should not explicitly call `gc.collect`, but rather leave it to python to do the garbage collection as appropriate. But setting things to None after they are not used anymore should enable python to collect when it needs to.

tito: The issue that he's trying to solve is to minimize the GPU memory. Python itself have no idea how much GPU memory is used; therefore we must able to recollect asap the GPU memory for memory-constrained platform as the RPi.

My point would be to have a "dealloc" method available in the FBO, then call it before setting to None, that would ensure that the GPU memory is recollected directly, and leave the general memory management to Python.

tito: But yes, the PR cannot be accepted with a gc.collect() in the middle.

matham: Ah, I assumed that `dealloc` is automatically called on GPU memory when the cython objects are garbage collected with `__dealloc__`? Is that the case? But I guess you're saying that it still doesn't happen quickly enough since python gc doesn't track GPU memory.

I guess adding a `destroy` method to these classes, which can be called manually but is also called in `__dealloc__` should take care of this?

audryste: As tito told setting to None before gc.collect helps to save gpu memory. But without gc.collect rams just keep growing endlessly until 4 GB pc with shared RAM  for gpu and PC crashes after a lot of time. :)
But now I see that you know the right way solution :) That is super nice! :)

rnixx: The FBO provides a ``delete_fbo()`` function https://github.com/kivy/kivy/blob/master/kivy/graphics/fbo.pyx#L201 which calls ``get_context().dealloc_fbo(self)``

@tito i guess this can be used to free GPU memory?

@matham @audryste

rnixx: also the ``Context`` https://github.com/kivy/kivy/blob/master/kivy/graphics/context.pyx has all sorts of ``dealloc_*`` functions. Does ``render_ctx`` hold such a context instance?

PR #6720, opened 2020-02-05
remove --enable-pulseaudio-shared=no from ci manylinux build

tshirtman:


matham:
This doesn't change the audio issues so...

PR #6721, opened 2020-02-06
remove gpl gif implementation

tshirtman:
to avoid depending on pillow just for our loading animation, created a
zip alternative, and replaced it for usage in all places.


matham:
Is this the only implementation for gif (other than ffpyplayer)? If so, don't we need to go through a deprecation cycle?

tshirtman:
We also have the Pillow implementation, this is the only one we have without third party dependencies though.
- It's quite buggy
- It's GPL, which is a violation to distribute in an MIT package, nobody complained yet, but this is bad.

If we think we need to support gif out of the box, we should either make a dependency like Pillow or ffpyplayer mandatory, or shop/build a better implementation to include in kivy.

Issue #6717, opened 2020-01-29
[Feature Request] Hover motion event 


pythonic64:
Based on discussion at https://github.com/kivy/kivy/pull/5291 and https://paste.ubuntu.com/25661582/ I'm writing on what is needed to develop generic hover event. First implementation will be a mouse hover event, but later other events (like eye tracker or kinect) can be added.

To be supported: 
1) Must be cursor independent
2) Support multi-hovering
3) Must be aware of widget tree (child-parent relationship)
4) Updated `profile` list for hover events
5) Event `on_motion` added to `WindowBase` and `Widget` classes
6) Property `motion_filter` in `Widget` class to filter out events

While I work on the pull request, any suggestions are welcome.

pythonic64:
Solved by https://github.com/kivy/kivy/pull/7658 adds an event manager which is layer between window and widgets. Feature will be available in 2.1.0 version of Kivy.

For an example on how to write a hover event manager and behavior take a look at https://github.com/pythonic64/hover .

PR #6724, opened 2020-02-09
[WIP] Hover event

pythonic64:
This pull request aims to add generic hover event which can be used by hover capable providers. First implementation is for mouse provider to test out event dispatching and handling to find acceptable architecture. After that gets done for single hover, then support for multi-hover event will be added.

What's changed on Widget class:
1. Added `on_motion` event to be used as generic event handler
2. Added methods `add_motion_event` and `remove_motion_event` to `motion_filter` property
3. Added `motion_filter` dict which keeps a `set` of child widgets (and `self` if needed) for each event (in this case hover)

What needs to be done for single hover event:
1. Try to replace HoverEvent `handled_by` with `grab` method and `handled_widgets` with `grab_list`.
2. Decide on how to notify widgets when indicator moves from lower to upper widget (overlapping widgets). Seems better to use argument for `on_motion`, currently 'end', instead of attribute to `HoverEvent`. See `post_dispatch_input` method in kivy/base.py. 
3. Decide if `HoverEvent.name` is a good choice to be used as key for registering in `Widget` and for inspecting events.
4. Make `add_motion_event` and `remove_motion_event` methods accept `event_name` as first argument and `widget` as second. If `widget` is `None` then use `self` to add event to `motion_filter`.

Widgets to override `on_motion` method and handle/transform hover event:
1. RelativeLayout
2. ScrollView
3. ModalView
4. Scatter

@tito @matham and other developers, guidance for this feature would be most welcome. Ref. issue https://github.com/kivy/kivy/issues/6717

- Update (1): Posted test app at https://gist.github.com/pythonic64/b9cc5c854738f536f2c691c8ff2796c0 . Meant for commit 42235e005971ee1bb99d3acd9f01aae96a10f1fb.
- Update (2): Added example app to showcase `HoverBehavior` at examples/widgets/hover_behavior.py


kivy/base.py
```diff
@@ -229,11 +230,32 @@ def post_dispatch_input(self, etype, me):
             if me in self.me_list:
                 self.me_list.remove(me)
 
+        if hasattr(me, 'name') and me.name == 'hover':
```
tito: Can you make a single variable for hasattr(me, 'name') and me.name == 'hover' ? As it is used in 2 if.

kivy/base.py
```diff
@@ -229,11 +230,32 @@ def post_dispatch_input(self, etype, me):
...
         if not me.grab_exclusive_class:
             for listener in self.event_listeners:
                 listener.dispatch('on_motion', etype, me)
 
+        if hasattr(me, 'name') \
+                and me.name == 'hover' \
+                and len(self.hover_events) > 1:
+            current = self.hover_events[-1]
+            prev = self.hover_events[-2]
+            for w in set(prev.handled_widgets) - set(current.handled_widgets):
```
tito: I'm a little bit worried here, is it executed every time there is a hover? If it's possible not to create 2 set to compare every call, that would be perfect!

kivy/input/motionevent.py
```diff
@@ -484,3 +484,21 @@ def is_mouse_scrolling(self, *args):
         .. versionadded:: 1.6.0
         '''
         return 'button' in self.profile and 'scroll' in self.button
+
+
+class HoverEvent(MotionEvent):
```
tito: Don't forget the documentation here, with the ..versionadded tag (look around the code to find out)

kivy/input/providers/mouse.py
```diff
@@ -127,6 +127,15 @@ def clear_graphics(self, win):
             win.canvas.after.remove(de[1])
 
 
+class MouseHoverEvent(HoverEvent):
```
tito: Documentation missing too

kivy/input/providers/mouse.py
```diff
@@ -293,6 +304,12 @@ def on_mouse_release(self, win, x, y, button, modifiers):
             self.alt_touch = None
         return True
 
+    def on_mouse_pos(self, win, mouse_pos):
+        width, height = EventLoop.window.system_size
+        args = (mouse_pos[0] / width, mouse_pos[1] / height)
+        event = MouseHoverEvent(self.device, None, args)
```
tito: Would if be possible to have only one event at all for the mouse ? If you look at MouseMotionEvent, it is created everytime the user touch down, and deleted when the user touch up. In your case, we could have a single hover event for the whole time of the app (for the mouse event provider).

pythonic64:
https://github.com/kivy/kivy/pull/6724#pullrequestreview-355892677 Property `motion_filter` uses `dict` + `set`, but that's only for registering child widgets and not for dispatch. Dispatching goes through `children` list as hover event can be handled by an overlapping widget. If you run app from https://gist.github.com/pythonic64/b9cc5c854738f536f2c691c8ff2796c0 you can see case where button is overlapping hover widget.

https://github.com/kivy/kivy/pull/6724/files#diff-bdc3bfa9adff8fbe222e520bbe7abc63R246 is meant to handle overlapping widgets in two cases:
1) Two hover aware widgets overlapping
2) Non hover widget is overlapping the hover aware one

To handle those cases set difference is made between previous widgets which handled the event and the current ones. If there is a difference then current event is dispatched directly to those widgets with 'end' argument so that they can go to non-hover state. Two consecutive hover events are kept in `EventLoopBase.hover_events` and checked to detect when indicator moves from lower widget to widget above it. 

I'll add a method to compute that difference instead of creating those sets, as that is only there temporary as changes to `post_dispatch_input` method are very wip.

kivy/base.py
pythonic64: I can use `if getattr(me, 'name', None) == 'hover':` instead of that.

kivy/base.py
pythonic64: Will be replaced with generator method which will yield members of the difference set.

kivy/input/motionevent.py
pythonic64: I'll add documentation at the end development, once we iron-out this feature.

kivy/input/providers/mouse.py
pythonic64: Will be added that the end of development.

kivy/input/providers/mouse.py
pythonic64: We will need at least two events so that we can send 'end' argument to handle two widgets overlapping case. I think we have cases where touch event is kept in widgets instance like a last touch, so maybe there is a case where hover event is also needed to be kept. And we still need to handle multi-hover which also needs to work correctly for overlapping case.

pythonic64: I would have to think more about single event per provider or cache of events per provider.

matham: Yeah, I kinda agree that reusing a hover event, while maybe performant would lead to usability issues because people will expect they can save the event, like last touch event and maybe manipulate it after the event is dispatched.

matham:
I kind of want to talk about this at a high level before considering the implementation: The goal is to notify interested widgets of motion events. But also implementing a specific (mouse) hover event.

In this design, the touch provider creates the motion event, e.g. on each position change and sends it to the window, which sends it to all the widget children. This is then dispatched all the way down the widget tree, and widgets who registered for a particular motion event will be notified, everything else is skipped. But, once a widget "handles" it, `kivy.base` keeps track of who currently gets the touch and notifies them etc.

I like this design overall, however, I have some potential improvements:

* I don't like that we add these hover specific things to base. Base should know nothing about any specific event type and `handled_widgets` should not be there. Imagine if we add future event specific code to base - it'd get more complex and coupled to these events. Base is a high level manager who shouldn't care about lowly employees 😁

  A better approach is to slightly model `FocusBehavior`. Where the hover specific code should be in a HoverEventManager. Then, this manager is instantiated and registered with base, who calls it as needed, without knowing what the manager does. This manager then tracks which widget needs to be notified etc. The point is that we can then separate the hover implementation, which could even be added to garden and it would still work once we add the basic machinery to kivy.

  And in particular, most of the performance issues goes away, because only people who register and want to use hover events suffer it. Although, perhaps a manager like hover could be auto-registered. Or perhaps not because it's not that common to need it.

  Basically, all the current base code would go into this manager.

* Another change is not to add the motion specific stuff in Widget, but somehow use mixins, like `FocusBehavior`. There are multiple options here.
  We can add a on_motion event to widget like you have now. The issue is that users would have a giant `on_motion` method if they use different motion events (hover, gaze, etc.). Also, I don't fully feel comfortable with adding filters using dicts that need to track all the widgets and their children to what to dispatch. Every time a new widget is added, this executes more code and we need to keep track of the full widget tree essentially.

  Basically, the issue is that you're deciding right now how to make the trade-off between dispatching to all the children (like `on_touch_xxx`) which is appropriate when the events don't come very often. And for a hover type event where we can't do that because the event happens continuously. The extra complexity you added is optimized for the latter. I think we shouldn't make this decision at all in kivy but leave it up to the specific event.

  Some other possibilities that we can mix and match are:
  * Basically the same, but in addition also add a Mixin for each manager, so that widgets who want to e.g. participate  also get a `on_hover` event added by the mixin. Then Widget does whatever it does, but the mixin auto registers the `hover` event and dispatches a `on_hover` from `on_motion`.
  * Instead of using dict filters, rely fully on the mixin and use `isinstance` and only dispatch to those who inherit from the mixin that implement the specific behavior. This makes it a little more complex to participate in a event (you need to inherit from a mixin), but separates it from Widget so each event can be fully independent with its own mixin. Widget's `on_motion` would simply forward all motion events to all its children, but specific mixins would handle specific motion events.

    The problem is that all widgets would now dispatch all events which can be prohibitive for e.g. hover.
  * A hybrid: Widget has `on_motion` that forwards all events and we use mixins, like the last option. This works well for low frequency events. But, for high frequency events, the Manager would not dispatch as normal. Instead, whenever a mixin widget is added to the widget tree, we tell the manager. Then, this manager is responsible for dispatching to whatever widgets it wants. So a high frequency event manager may keep a flat list of all the mixins of its type and dispatch to that list! That's what `FocusBehavior` does. And whenever we use `add_widget` with a mixin, upon the next event, it recomputes the flattened tree and dispatches them.

I really like this last option because basically kivy requires minimal changes. We just add the ability for base to accept event managers, which it calls as appropriate. And we just add `on_motion` to Widget which simply forwards all events.

Then, in garden, or maybe in kivy for e.g. hover we add a manager and mixin which depending on the frequency of the event, the manager either simply dispatches to all children of Window etc., or keeps a flat list of the participating widgets and dispatches to them. Then people who don't use hover don't pay a penalty cost!

kivy/base.py
tito: It won't help, and instead of just creating 2 sets and do a diff, it will also create a generator. My question was if it would be possible not to do that every time there is a dispatch.

pythonic64:
Hey @matham ! Thanks for your suggestions. 

I updated `Widget` class and now `motion_filter` is only interested in keys and only when key is added/removed then widget's parent updates its own `motion_filter`. This way if key is already present in parent's `motion_filter` then dispatching doesn't propagate up. Since for one key we can have many widgets adding/removing trigger dispatch as that's how `DictProperty` works. 

Current code in `base.py` is temporary, that was just for me to test out dispatching through widgets. I was thinking about a class to move that logic and `HoverEventManager` seems good to me too. That class would handle logic for all hover events and for now detection when indicator is no longer over a widget. 

I plan to add `HoverBehavior` to `kivy.uix.behaviors` which will have events: 
- `on_hover_enter` 
- `on_hover_move`
- `on_hover_leave`

and `hovered` attribute which will indicate that at least one event is hovering. User only needs to inherit from this class to have `on_hover_xxx` events to work as behavior will register to `hover` event and override `on_motion`. 

I don't think that something like `FocusBehavior` wont't work for this because that behavior cycles through instances of itself and sets `focused` to `True`/`False`. Hover events needs to be more aware of widget tree. 
Example, if we have a non-hover `DropDown` which is on top of widget tree and hoverable widget which 5+ levels down the tree and dropdown is completely or partially over it than for case where indicator is over dropdown, then dropdown will handle that event and not let it to go down the tree. If indicator is moved from hoverable widget to dropdown, then dropdown handles that widget and `HoverEventManager` will send `end` event to hoverable widget to revert to non-hover state. 

Also, we need to transform position of hover as it goes through the tree. I don't see a way how can hover widget inside several `RelativeLayout` or `Screen` or inside some container inside of `Scatter` know its local position. When dispatched from window it position is in window coords. 

Problem I see with `on_motion` in `Widget` class is its `else` part. One solution would be to have attribute in `MotionEvent`, like `handle_on_collision` which if `True` and `event.pos` collides with widget then stop the dispatching. This way we are not limited to hover events and this attribute can be set by managers. And I have to add check when widget is disabled and for that case handle the event if position collides. 

This hover event should enabled/disabled by config, but maybe we should enable it by default for desktop platforms.

pythonic64:
Highlights of the latest update:

1. Added `EventManagerBase` and `HoverEventManager` to handle dispatching of events after post-processing
2. Added `type_name` to `MotionEvent` to be used as general type event name. Multiple hover event providers can have same `type_name` and "hover" is used as `type_name` for `HoverEvent`
3. Added `HoverBehavior` mixin
4. Transformed motion event in `RelativeLayout`, `Scatter` and `ScrollView` widgets.
5. Always accepting event in `on_motion` for `ModalView` and `DropDown`


Updated `HoverEventApp` https://gist.github.com/pythonic64/b9cc5c854738f536f2c691c8ff2796c0

pythonic64:
Ok, the feature is almost complete. Example app which showcase how `HoverBehavior` and `HoverEventManager` work is at `examples/widgets/hover_behavior.py`. Behavior has `hover_mode` property which enables user to set how dispatching of hover event will be handled. It has three modes:
- `default` - disptach to children first, but handle by widget if none of the children handles it
- `self` - only current widget can handle the event. Useful for cases where children should not get the event, even if they inherit from `HoverBehavior` (like changing a background color of a widget)
- `all` - dispatch to children first and handle by widget. Useful for cases where children and widget want to handle the event (like hover inside hover) 

Class `StencilViewHoverMixin` is added which when used with `Widget`, or subclass of it, will make hover events pass down the widget tree only if they collide with widget first. 

Attribute `min_wait_time` is added to `HoverEventManager` and it specifies time in seconds after which manager will dispatch copy of last hover event (which is not `'end'` type) to event listeners. This makes event continuous and gives option to handle hover when widgets are moving even if indicator does not move.

Things still left to do and need assistance from core devs to decide: 
1. Is `exclusive` grab needed for hover event?
Event type `'update'` is always dispatched with new instance of hover event, so grabbed widgets will only get 'end' events, so maybe `exclusive` is not needed for hover events.
2. Should event type `'end'` be added to `MouseMotionEventProvider ` just for completeness? 
Event type `'start'` can be dispatched for first event and also when mouse indicator enters the window. Currently only `'update'` and `'end'` are dispatched. 
4. Should `device_id` be attribute instead of Python property in `MotionEvent` class? 
This adds new value in `MotionEvent.__attrs__` list and duplicates value which is now in `id` and `device` attributes, but it does not create new tuple as property `device_id` now does and `device_id` is often accessed by `HoverBehavior` and `HoverEventManager`.
5. How to setup config changes? 
Do we need new section for event managers? Like:
```
[eventmanagers]
hover = 0.03
touch = 
keyboard = 
laser = 
```
as configuration for manager would be needed. Like:
`hover = 0.03`, which will register `HoverEventManager` in `EventLoop` and set `min_wait_time` to 0.03 seconds. Right now, manager is has to be register manually. 
6. Are current names for class, methods etc good or someone can suggest better, more descriptive ones?

pythonic64:
I rewrote `HoverBehavior` so that event handlers behave the same way as handlers for touch events (calling `super().on_touch_xxx` would dispatch event to children), so behavior now fits more to Kivy way of handling events. Also, what I noticed is that now we can introduce `TouchBehavior` class which would look like this:
```python
class TouchBehavior(object):

    def __init__(self, **kwargs):
        self.add_motion_event('touch')
        super().__init__(**kwargs)

    def on_motion(self, etype, me):
        if me.type_name != 'touch':
            return super().on_motion(etype, me)
        if etype == 'start':
            return self.dispatch('on_touch_down', me)
        elif etype == 'update':
            return self.dispatch('on_touch_move', me)
        elif etype == 'end':
            return self.dispatch('on_touch_up', me)

    def on_touch_down(self, me):
        return super().on_motion('start', me)

    def on_touch_move(self, me):
        return super().on_motion('update', me)

    def on_touch_up(self, me):
        return super().on_motion('end', me)

```
Widgets or other behaviors which need touch events would inherit from this class, but `Widget` class would no longer have `on_touch_xxx` methods so this would be a breaking change.

pythonic64:
Last commit fixed how `min_wait_time` is used in `HoverEventManager` when ensuring a copy of last event. Now: 
- negative value means that only events received from providers will be dispatched to listeners and copied events won't be dispatch
- value `0` means that copy of last dispatched event will dispatched but only if event from providers is not present
- value greater than `0` means that manager will wait `min_wait_time` to get the event from providers and if event is not received in that time copy of last dispatched non-end event will be dispatched

pythonic64:
I updated mouse event provider and `HoverEventManager` to work with single instance of hover event(s). If user want to keep instance of hover event in their code, then they would have to:
1. Keep instance of event which will get updated on mouse motion (and use it as read-only)
2. Make and keep a copy of instance like this:
```python
temp = type(me)(me.device, me.id, me.args)
me.copy_to(temp)
```

Mouse event provider is also updated to dispatch 'start' event type `on_cursor_enter` event. This checks question (2) from https://github.com/kivy/kivy/pull/6724#issuecomment-609976064 .

While working on latest changes I noticed that `Window.mouse_pos` property is not updated when cursor is static and window gets created, and for events `on_cursor_enter`, `on_cursor_leave` and probably for other cases like when window is maximized with keyboard or non-mouse move event. To cover these cases would be better to make another pull request. 

@matham @tito I can contact you on Discord so we can talk about this feature there if you think that can speed up the review.

matham:
Apologies for the length of this post, but to me this a design doc for kivy input and how it should be in the future to accommodate arbitrary input events in the widget tree, rather than this PR and hover only. Consider this a proposal for the managers mentioned and overall widget event processing, like a PEP. Hopefully it's not overwhelming.

Also, sorry for taking so long to get back, but I needed to find some time to get a very clear understanding of the intentions behind kivy's input abstraction. Kivy has a very rich input abstraction, however, there are some long standing issues (see below) and I wanted to be sure I understood the cause and whether it's a problem with the implementation or with the abstractions. Because what we propose here would determine how events are processed and I want to be certain we are not making a decision that prevents future fixes. My conclusions are that the abstraction is fine and that the issues arise from quick hacks around the abstractions and perhaps lack of additional abstraction in some areas.

Following is a overview of kivy input. It's not completely tied to this PR, but before we can discuss the PR we need to have a good context of what we need to solve so I'm going over the basic elements, both for my own benefit and to contextualize the PR and the issues addressed by the PR in my understanding. As well to contextualize many of the issues with touch in Kivy and how it relates to this PR and Kivy's future. I may add this to kivy docs to function as a design doc if it seems right.

# Overview

There are three parts to Kivy input: input providers that generate motion events, the kivy window that translates motion events into something that has meaning to the widgets, and the `EventLoop` that is the bridge between input providers and the window and offers some intermediate processing before passing it on to the `Window`.

Most of the existing issues arise when these abstractions are violated. E.g. When `EventLoop` dispatches `on_touch_xxx` events that are grabbed, directly to widgets instead of going through `Window`. Or people having access directly to `on_mouse_xxx` events in Window before they become proper motions events.

Similarly, once window has a motion event, it only does `on_touch_xxx` events for touch events. This PR and e.g. the need for (lack of) accessibly API, and some hacks to make `FocusBehavior` work indicate there's some additional missing abstraction to help translate motion events to widgets in the `Window`.

The manager idea from this PR should address this specific abstraction IMO.

# Components

Following are the input components and their place in the abstraction.

## Input providers

They are called periodically and process input from hardware or software that has occurred since the last call. They know nothing about Kivy widgets or the Kivy window (for the most part), all they know is how to create motion events. Specifically, they create a motion event and an action. The action can be

1. `'begin'` when we encounter this motion event the first time.
2. `'update'` when an existing motion event changes in any way.
3. `'end'` when an existing motion event ends.

Examples are:
* Windows touch events that are created from the Windows system and are associated with the Kivy window.
* TUIO events that are streamed in from a server indicating the position and id of some fiducials.
* Mouse events. These are added by the `Window` as `on_mouse_xxx` events that are sent to the mouse input provider, which creates motion events for them (that is then sent back to the Window through `EventLoop`).
* Motion events from a 3d controller that gives the position of each each finger in 3d (or 2d) space.
* Motion events from a controller that gives the configuration of the hand and the symbol indicated by the hand.
* The locations gazed upon by each eye.

## Window motion events

The `Window` receives motion events when `EventLoop` dispatches Window's `on_motion` event. It includes the event and the action. The role of `Window` is to translate the event into something meaningful to `Window` and the widget tree.

Currently, it only handles touch events and it translates it to the widget tree using `on_touch_xxx` of `Window` and `Widget`. The works by passing the event to each widget in turn, until something handles it. But the touch is translated into the widget's parent coordinate system before it's passed to each widget.

Here is where we'd need to add managers that can be registered with `Window` (see subsequent sections). Each manager would pass the motion event on according to its logic. A default manager would perform the actions currently done by `Window`, scaling the touch events and dispatching (or we could leave the logic for touch as is).

Managers would decide how to pass the event to the widget tree. E.g. they could pass it to a list of added widgets, or to widgets of a specific class, or to all widgets using a `on_motion` event.

It is expected that users could create managers to handle motion events for custom widgets without any need to internal Kivy changes.

## EventLoop bridge

The `EventLoop` allows input providers to be registered with it. Then, for each clock iteration it gets all the new or changed motion events, process them using any registered post-processors and then dispatches the event to all listeners, including `Window`.

`Window` doesn't know anything about the motion events origins, and the input providers know nothing about the window or widgets (except e.g. for mouse events that come from the window).

# MotionEvent Widget abstraction

Looking closely, we can see that there's one abstraction missing between Window and widgets. Currently, the only abstraction is that touch (pos) events get translated from a motion event into `on_touch_xxx` events dispatched to the widget tree. However, we can think of many ways that the `Window` can translate and forward motion events to individual widgets.

E.g. we may want to support sending the same event to multiple widgets (e.g. a hand in 3D space pointing to multiple widgets), or an accessibility API that takes certain motion events and handles them in a special way. Or the motivation for the PR - a manager that takes the window motion events and activates hovering for a widget when the mouse hovers over it.

This suggests that `Window` needs a way to register managers that take motion events and pass them on to widgets in their own way. Then users can do this without needing to add stuff to kivy itself, but simply register a manager.

# Issues

Here I go over some long standing issues or problem areas that has come up before and the reasons for them and possible fixes in the context of this PR.

## Hovering

Previously to get the mouse pos to detect e.g. hovering, we'd bind to `on_mouse_xxx`. However, this event should have been internal to Window IMO because as with all input providers, this event should be considered the low-level mouse position that generates a motion event that is passed to `on_motion` of `Window` after optional post processing in `EventLoop`.

So ideally, users would not have gotten access to these events, but instead if they needed a mouse event, they'd have listened to `on_motion` and handled any events with a mouse profile.

For this PR, that's exactly where a hover manager needs to be registered - it should take motion events and decide which widget is being hovered over and activate the hovering of that widget. As discussed at the start of the PR, there are multiple ways to do this and different hover managers can do it differently. I'll get to the specifics of this at the end.

## FocusBehavior

Currently, `FocusBehavior` needs to know when a `on_touch_up` event occurred outside any `FocusBehavior` widget to clear focus. We do this by keeping track of touch events within any `FocusBehavior` and then in `Window`'s `on_motion`, we do `FocusBehavior._handle_post_on_touch_up(me)` after `on_touch_up`.

This is hacky as `Window` ideally wouldn't need to know about `FocusBehavior`'s specific needs. This would be resolved by making and registering a `FocusBehavior` manager that processes all touch events and clears focus when a touch outside any `FocusBehavior` is released.

## Accessibility

Similar to `FocusBehavior`, an Accessibility API would benefit from being able to register a manager that handles any accessibility needs by processing touch events as needed. E.g. it may take a touch or hover and a keyboard input (we'd have to figure out the accessibility API can get keyboard inputs if that's used) and read aloud the text under the mouse.

How the manager decides which touch to process or which widget to "focus" on is upto the specific manager. Being able to register managers would allow an Accessibility API to be written e.g. in garden and we experiment with different APIs before possibly including it in Kivy.

## Touch grabbing

Currently, the `EventLoop` has special code to handle grabbed touch events. I.e. if a widget grabs an event, the `EventLoop` will dispatch the `on_touch_xxx` event to the widget.

This has multiple issues:

* The `EventLoop` should not know anything about `on_touch_xxx`. These touch events are part of the `Window` abstraction and not `EventLoop` or the input providers. This leads to issues where e.g. the coordinate system of the event when dispatched when grabbed is not the widget's, unlike the events when normally dispatched. This means that sometimes it is and sometimes it isn't in thw widget's coordinate system. While this can be fixed, it likely happened because this was a hack to enable some use cases.
* The grabbing API is quite confusing to newcomers. Storing metadata in a user dict and also running into issues when subclassing widgets that make use of grabbing is quite common. Because it's unclear if and when the the touch is dispatched because it's grabbed. So users wonder why they got a touch multiple times.
* When a touch may be used by multiple widgets, again the user data dict can be confusing and knowing whether the touch was used frequently is the cause of issues when (ab)using complex widgets.

This is not meant to criticize the grabbing API (although improving these pain points should be part of the long-term goals) as it serves the need. The point is that if we had managers, then we could have managers handle grabbing in different ways and we can experiment in the best ways. Furthermore, users could add their own grabbing managers if they don't like the default API.

## Sharing event between multiple widgets

Sometimes a touch event may be shared between multiple widgets. Currently this is hard to do as both have to be written with each other in mind. Perhaps that will always be required. But with managers, we could have user-written smarter managers that can specifically handle this use-case if one arises in user code. This known use-case is somewhat similar to the `FocusBehavior` use-case of needing to handle a touch in multiple widgets.

This issue is included for completeness sa it's not clear if it's a actual issue.

## Sharing event between widget and its superclass

This is a cooperative touch issue, and it's not clear how to solve it, if at all. But the issue is when a super class handles touch in a certain way but a subclass needs access to the touch. It's not clear if this can be solved or if it needs to be solved and the best way is to make motion events more robust. But having managers would make experimenting possible.

## Multi-touch simulation

Currently, multi-touch simulation happens as a input provider. Including drawing the red circles. However, it would be more appropriate if this drawing, which requires knowledge about the window and graphics were to happen in Window, not in a input provider that is supposed to be `Window` agnostic. And to allow managers to decide how to display the multi-touch simulation, rather than being forced to use the rec circle.

Again, using a manager that visualizes this for multi-touch simulated events in `Window` would be better as a user would have more control over this. Currently it requires setting hard coded config properties of the input provider.

# Managers and how to handle shared events

One consideration, is what to do when one managers uses a motion event. Do we still pass it on to the other managers? Consider two situations:

* A `FocusBehavior` widget is focused. Then, the Accessibility manager uses the next touch for its own purpose. The `FocusBehavior` manager still needs to be notified of the touch's release so it can release focus.
* A touch is used by the Accessibility manager to read aloud the text of a button. The touch should not be further used by window to activate the button.

This indicates that each manager as it processes the event should return whether it used the event. Then, this bool should be passed to subsequent manager, which should still be called. And each manager would decide whether to still use the event.

matham:
This post is specifically about this PR.

I think the PR generally goes in the same direction as the above design doc. However, to move forward I'd ask to split the PR into two parts. The abstract manager part and the hover part.

Once the manager part is in, the hover part can be designed independently of kivy, and specifically can be first designed in garden and then added to kivy if everyone is happy. Kinda like what we did with `RecycleView`. If the manager is designed well, as laid out above, hover should need no changes in kivy itself to work. My main concern here initially is to make sure the manager API works and can accommodate all expected use cases. And designing hover first in garden will confirm this.

So, can you make a PR with just the manager? Then I can give specific feedback about how you implemented it. Some initial thoughts:

* These event managers should be added to `Window`, not `base.py`. Perhaps under `core/window/motion`?
* In `on_motion` in `Window`, the managers get all the motion events, not just those that are not touch events.
* The `EventManagerBase` should have a `begin`, `update`, and `end` method!?
* I'm not sure the base manager should have a `transform_grabbed_event` method. We shouldn't assume that the motion event will have a singular position. We can talk details.
* Anything specific to Hover should be separate, so the manager API is clearer to evaluate.
* I don't think hover will require its own `HoverEvent` `MotionEvent` subclass!? Or am I overlooking something that requires it? Especially since I'm assuming this will all be done in the manager with existing motion events. Same for `MouseHoverEvent`. All this should be available from existing motion events passed to `on_motion`!? If not, we need to update the events so the required info is in the event.
* The last and largest part of the PR/API that makes me uncomfortable is the implementation of `on_motion` in `Widget`. I don't like that we have filters and code executed in `Widget`'s `on_motion`. IMO, `Widget`'s `on_motion` should simply be `pass`. And in particular, each manager and each widget that accepts motion events from managers would decide how to handle motion events.

  E.g. for hover, all the code to call `on_motion` of its children widgets will be in the `HoverBehavior`. E.g. one way would be for `HoverBehavior`'s `on_motion` iterate all its children recursively, but only children who are `HoverBehavior`. Then only `HoverBehavior` will pay this cost and only for motion events used by the hover (the manager could inject something into the existing event so the `HoverBehavior` knows this motion event is meant for it) and not other motion events.

  Another, and perhaps better approach would be for the manager to iterate all the widgets in Window who are `HoverBehavior` and call `on_motion` until someone accepts it, but the widget themselves don't call any children. This would allow us to optimize by computing ahead of time the widget tree and keep track as it changes and only dispatch to more limited number of widgets. This would accommodate the situation you described with drop down I believe.

  The point is that unlike `on_touch_xxx` whose reason to exist is to call it on to all children widgets through the whole tree. For `on_motion`, the manager and the widget decide how to handle it and potentially pass it on to the children, if at all. The alleviates `Widget` from having to do any work of filtering etc and lets us pick managers that balance performance as it sees fit. This is required to handle arbitrary motion events.

  In fact we could have two competing hover managers, one who does as suggested above, and the other who goes through the a list of `HoverBehavior` that registered themselves with the manager. And depending on use cases, the user decides which to use. Point being, none of this logic should be in `widget.py`.

  This part I expect is to be the hardest to design given all the tradeoffs involved. But I'm hopeful we can design `on_motion` like I just described and still be able to do everything.

pythonic64:
I'll start working on a new pull request which will include the following:

1) Define abstract `EventManagerBase` class.
2) Both specific manager implementations and behaviors (like `HoverBehavior`) should (for now) exist in Kivy Garden. I'll add manager and behavior for hover events on my GitHub account first and later we can move it to Garden.
3) `WindowBase` must have methods to register/unregister event managers.
4) Method `on_motion` of `WindowBase` will pass events to event managers.
5) Method `transform_grabbed_event` can stay in `EventManagerBase` for now, but it can be moved somewhere else later.
6) Add a default event manager which will do current logic from `WindowBase.on_motion` and `EventLoopBase.post_dispatch_input` methods if event is not handled by registered managers.


Thoughts other things:

- I would suggest `core/window/managers` but we can change this to `core/window/input` or to other appropriate location

- My opinion is that only managers which specify event type (`touch`, `hover`, `keyboard` etc.) should get those events. If a manager is only handling `hover` than it doesn't need any other events.

- HoverEvent is overriding `grab` and `depack` method, MouseHoverEvent is overriding only `depack` method, so we see how to move that logic to MotionEvent. MouseMotionEvent sets `is_touch` attribute to `True` and important thing is that touch is not the same as hover event. I think we can use `type_name` attribute to differentiate between touch, hover and other types.

- If `disabled` is not handled in `on_motion` then we would break the existing behavior. If we are not passing event down the widget tree then we cannot do transform of events for widgets like `RelativeLayout` or `ScrollView` etc or for user defined widgets. And for those widgets inheriting from any behavior is not mandatory. Also we lose control which parent has over events which gets passed to children.

- Idea of subscribe/unsubscribe to motion events was from matham's and tito's discussion https://paste.ubuntu.com/25661582/ . I think I implemented that functionality with `motion_filter` property and `add/remove` methods. This enable events to flow down the widget tree to subscribed widgets only and we get some performance increase and parent handling of events (like in `on_motion` in `Dropdown`, `ScrollView` etc) where parent itself isn't subsribed to event but can accept the event even if children do not. 

- Another case to consider is where parent handling of event is important is `StencilView` and classes which inherit from it (like `ScrollView` and `RecycleView`). It this case StencilView has to test if event position collides before passing it to children, otherwise child widget can detect hovered state when child widget is clipped and cursor is outside of `StencilView`. In the example app this is handled by `StencilViewHoverMixin` which is used by `HoverRecycleView` class and `HoverRecycleView` does not subsribe to hover events

- Another case is where both parent and child are handling/using the same hover event. If we are iterating through a list of `HoverBehavior` and disregading parent-children relationship the child will get the event first, accept it by `return True` and parent (or ancestor) will never get that event. Example for this case in example app is `ChatItem` in which icon and text are changing color and item is changing its background color on hover event.


I don't think we can have `Widget.on_motion` with just `pass` in the body and satisfy cases from above. `FocusBehavior` works by iterating to next widget and focusing it, but any event with position would have to go through widget tree to transform itself or collide with parent before it gets (or not) passed children.

Maybe we can have `flags` attribute in `MotionEvent` which `Widget.on_motion` can check to decide whether to pass the event to children or not, or to test for collision if widget is disabled or anything else. Managers can set this attribute and through it influence handling of event through widget tree and attribute value can be stored/restored using `push/pop` methods.

pythonic64:
@matham Can you take a look at https://github.com/kivy/kivy/pull/7387 ? I'll fix the failing tests after we get the code in order.

matham:
I will take a look at what you wrote and the linked PR, but it'd be a bit of time before I get to it as I don't have a lot of time for Kivy currently.

pythonic64:
Ok, if you're short on time, then only take a look at https://github.com/kivy/kivy/pull/7387. Two modules are changed, changes are straightforward and they enable mouse provider to dispatch hover events.

pythonic64:
This feature is split and redesigned in the following requests:
- https://github.com/kivy/kivy/pull/7387
- https://github.com/kivy/kivy/pull/7658
- https://github.com/kivy/kivy/pull/7785

PR #6732, opened 2020-02-14
img_dds.py fix for upside down dds files.

wezu:
All DDS files I tested where loaded upside down, but to be honest I'm not sure if some tools save the data inside dds file in some other arrangement - OpenGL and Direct3D use different conventions for where up and down is in a texture.


Zen-CODE:
It probably stems from the fact that Window uses the top left of the screen as (0, 0), and OpenGL the bottom left. The option is already there to flip, so I don't really see the point of changing the default? It's so easy to change and so obvious. Do you really think it warrants the change?

PR #6741, opened 2020-02-19
add 'orientation'property to GridLayout

gottadiveintopython:
Continuation of #6593 .

- ~~`RecycleGridLayout._update_rows_cols_sizes()` is not tested at all.~~
- [This](https://gist.github.com/gottadiveintopython/513629555a6eb83d8fec45a93725b958) is what I'm using on the test. (last update 2020 May 31)

### DIFF
--- a/kivy/uix/gridlayout.py
+++ b/kivy/uix/gridlayout.py
@@ -93,8 +93,9 @@ from kivy.logger import Logger
 from kivy.uix.layout import Layout
 from kivy.properties import NumericProperty, BooleanProperty, DictProperty, \
     BoundedNumericProperty, ReferenceListProperty, VariableListProperty, \
-    ObjectProperty, StringProperty
+    ObjectProperty, StringProperty, OptionProperty
 from math import ceil
+from itertools import accumulate, product
 
 
 def nmax(*args):
@@ -251,6 +252,27 @@ class GridLayout(Layout):
     only.
     '''
 
+    orientation = OptionProperty('lr-tb', options=(
+        'lr-tb', 'tb-lr', 'rl-tb', 'tb-rl', 'lr-bt', 'bt-lr', 'rl-bt',
+        'bt-rl'))
+    '''Orientation of the layout.
+
+    :attr:`orientation` is an :class:`~kivy.properties.OptionProperty` and
+    defaults to 'lr-tb'.
+
+    Valid orientations are 'lr-tb', 'tb-lr', 'rl-tb', 'tb-rl', 'lr-bt',
+    'bt-lr', 'rl-bt' and 'bt-rl'.
+
+    .. versionadded:: 2.0.0
+
+    .. note::
+
+        'lr' means Left to Right.
+        'rl' means Right to Left.
+        'tb' means Top to Bottom.
+        'bt' means Bottom to Top.
+    '''
+
     def __init__(self, **kwargs):
         self._cols = self._rows = None
         super(GridLayout, self).__init__(**kwargs)
@@ -268,6 +290,7 @@ class GridLayout(Layout):
         fbind('children', update)
         fbind('size', update)
         fbind('pos', update)
+        fbind('orientation', update)
 
     def get_max_widgets(self):
         if self.cols and self.rows:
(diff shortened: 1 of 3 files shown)


kivy/uix/recyclegridlayout.py
```diff
@@ -205,26 +206,35 @@ def get_view_index_at(self, pos):
...
+            at_idx((x, top)),
+            at_idx((right, top)),
+        ))
 
+        n = len(data)
         indices = []
-        row = len(self._cols)
-        if row:
+        stride = len(self._cols) if self._fills_row_first else len(self._rows)
+        if stride:
```
gottadiveintopython: Not sure why this truthy check is necessary. Both `len(self._cols)` and `len(self._rows)` look always greater than zero because of:

https://github.com/kivy/kivy/blob/53290db64b95a655f09e7c737b8e97fda7264451/kivy/uix/gridlayout.py#L303-L304

tshirtman (approved):
LGTM, especially the tests, sorry for taking so much time to review, it's a really useful contribution, and the code looks good. It's great that we have it for RecycleGridLayout as well, that was certainly not trivial.
I fear the way to compute the indices might be a bit heavy if we have a lot of items in the recycleview, as it'll have to create a really big list that maybe we could do without if we kept various divmod cases, but it might not be such a big deal, compared to the memory requirements of such an usage anyway, so i think it's fine to merge as is, and optimize later if we need and can.

gottadiveintopython:
Yes, there are couple of places that I should have avoided list/tuple creation.

kivy/uix/recyclegridlayout.py
```diff
@@ -39,16 +39,17 @@ def _fill_rows_cols_sizes(self):
         cols_sh_min, rows_sh_min = self._cols_sh_min, self._rows_sh_min
         cols_sh_max, rows_sh_max = self._cols_sh_max, self._rows_sh_max
         self._cols_count = cols_count = [defaultdict(int) for _ in cols]
+        # !! bottom-to-top, the opposite of the other attributes.
```
gottadiveintopython, line 42: This might be a misunderstanding

kivy/uix/gridlayout.py
```diff
@@ -295,9 +322,18 @@ def _init_rows_cols_sizes(self, count):
             Logger.warning('%r have no cols or rows set, '
                            'layout is not triggered.' % self)
             return
+
         if current_cols is None:
+            if self._fills_row_first:
+                Logger.warning(
```
matham, line 328: I'm just taking a quick skim, and that should be a error, not a warning. We should almost never have warnings in code. This is an error in the configuration so a error should be raised.

gottadiveintopython: ok, got it.

gottadiveintopython:
https://kivy.org/doc/master/api-kivy.uix.bubble.html#kivy.uix.bubble.Bubble.orientation

`Bubble` is a subclass of `GridLayout` and already has `orientation`property, which may be the cause of errors on unit tests. I just noticed after I made GridLayout raise an exception when the configuration is wrong. (@matham 's suggestion above).

You can confirm the errors by using [this branch](https://github.com/gottadiveintopython/kivy/tree/optimize_gridlayout2)

tshirtman:
Err, that would be an api break, i'm not sure how many people use that feature, i guess we could make "vertical" and horizontal" each aliases of one of the existing option in GridLayout?

gottadiveintopython:
That still breaks the API, because of the following code:

```python
if bubble.orientation != 'horizontal':
    # assuming the orientation is 'vertical', but there are 'lr-tb', 'rl-tb' etc...
```

So I believe we need another name if we don't want to break existing code.

tshirtman:
Ah, i just read more in detail, and it doesn't do what i had assumed, it sets the cols/rows for the nested gridlayout (it's a grid layout with another grid layout inside it for content…) so just relying on GridLayout's implementation wouldn't work.

Replacing an OptionProperty with another with different options used to work though? why does the change in GridLayout affects the Bubble?

gottadiveintopython:
ok I'll try in that way.

PR #6769, opened 2020-03-12
Kivy now works on the Raspberry Pi 4 without X11

Lauszus:
Thanks to @ddimensia for figuring out the missing piece in getting SDL2 to work without X11 i.e. setting `SDL_GL_ALPHA_SIZE` to 0.

The SDL2 instructions are more or less copied from the following site: https://github.com/midwan/amiberry/wiki/Compile-SDL2-from-source.

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -135,7 +135,7 @@ cdef class _WindowSDL2Storage:
         SDL_GL_SetAttribute(SDL_GL_RED_SIZE, 8)
         SDL_GL_SetAttribute(SDL_GL_BLUE_SIZE, 8)
         SDL_GL_SetAttribute(SDL_GL_GREEN_SIZE, 8)
-        SDL_GL_SetAttribute(SDL_GL_ALPHA_SIZE, int(environ.get('KIVY_SDL_GL_ALPHA_SIZE', '8')))
+        SDL_GL_SetAttribute(SDL_GL_ALPHA_SIZE, KIVY_SDL_GL_ALPHA_SIZE)
         SDL_GL_SetAttribute(SDL_GL_RETAINED_BACKING, 0)
         SDL_GL_SetAttribute(SDL_GL_ACCELERATED_VISUAL, 1)
 
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -135,7 +135,7 @@ cdef class _WindowSDL2Storage:
         SDL_GL_SetAttribute(SDL_GL_RED_SIZE, 8)
         SDL_GL_SetAttribute(SDL_GL_BLUE_SIZE, 8)
         SDL_GL_SetAttribute(SDL_GL_GREEN_SIZE, 8)
-        SDL_GL_SetAttribute(SDL_GL_ALPHA_SIZE, 8)
+        SDL_GL_SetAttribute(SDL_GL_ALPHA_SIZE, int(environ.get('KIVY_SDL_GL_ALPHA_SIZE', '8')))
         SDL_GL_SetAttribute(SDL_GL_RETAINED_BACKING, 0)
         SDL_GL_SetAttribute(SDL_GL_ACCELERATED_VISUAL, 1)
 
--- a/.ci/Dockerfile.armv7l
+++ b/.ci/Dockerfile.armv7l
@@ -24,6 +24,6 @@ RUN if [ "$KIVY_CROSS_PLATFORM" = "rpi" ]; then \
     fi
 
 # Build the wheel.
-RUN KIVY_SPLIT_EXAMPLES=1 USE_X11=1 USE_SDL2=1 USE_PANGOFT2=0 USE_GSTREAMER=0 KIVY_CROSS_PLATFORM="$KIVY_CROSS_PLATFORM" KIVY_CROSS_SYSROOT="$KIVY_CROSS_SYSROOT" python3 -m pip -v wheel --extra-index-url https://www.piwheels.org/simple . -w /kivy-wheel
+RUN KIVY_SPLIT_EXAMPLES=1 USE_X11=1 USE_SDL2=1 USE_PANGOFT2=0 USE_GSTREAMER=0 KIVY_SDL_GL_ALPHA_SIZE=0 KIVY_CROSS_PLATFORM="$KIVY_CROSS_PLATFORM" KIVY_CROSS_SYSROOT="$KIVY_CROSS_SYSROOT" python3 -m pip -v wheel --extra-index-url https://www.piwheels.org/simple . -w /kivy-wheel
 
 RUN [ "cross-build-end" ]
--- a/doc/sources/installation/installation-rpi.rst
+++ b/doc/sources/installation/installation-rpi.rst
@@ -6,32 +6,65 @@ Installation on Raspberry Pi
 Raspberry Pi 4 headless installation on Raspbian Buster
 -------------------------------------------------------
 
-#. If you have installed Raspbian with a desktop i.e. if you Raspberry Pi boot into a desktop environment, then you can skip to `Raspberry Pi 1-4 installation`_.
+#. If you have installed Raspbian with a desktop i.e. if you Raspberry Pi boots into a desktop environment, then you can skip to `Raspberry Pi 1-4 installation`_.
 
-   If you are using Raspbian Lite, then you first need to install X::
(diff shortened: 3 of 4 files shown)


Lauszus:
I have confirmed that the wheel build using the CI now works on the Raspberry Pi 4 without X11.

kivy/core/window/_window_sdl2.pyx
```diff
@@ -135,7 +135,7 @@ cdef class _WindowSDL2Storage:
         SDL_GL_SetAttribute(SDL_GL_RED_SIZE, 8)
         SDL_GL_SetAttribute(SDL_GL_BLUE_SIZE, 8)
         SDL_GL_SetAttribute(SDL_GL_GREEN_SIZE, 8)
-        SDL_GL_SetAttribute(SDL_GL_ALPHA_SIZE, 8)
+        SDL_GL_SetAttribute(SDL_GL_ALPHA_SIZE, int(environ.get('KIVY_SDL_GL_ALPHA_SIZE', '8')))
```
matham: I don't like the variable name `KIVY_SDL_GL_ALPHA_SIZE` because when someone wants to use console mode in the pi, they should be setting a variable e.g. `KIVY_WINDOW_PI_CONSOLE=1` rather than `KIVY_SDL_GL_ALPHA_SIZE`, which is something internal.

However, can't we always set this to zero on the Pi? Why is this user configurable? Could we not set it to zero when the platform is pi as detected in `setup.py`?

Lauszus: @matham yes I would prefer to always set that to 0 on the Pi 4 and forward. I just didn't want to break anything, that's why I added that variable.

Lauszus: See: https://github.com/kivy/kivy/pull/6769/commits/a0ff5212d5e2a60deb13cc9e7165df141d2df99b. Let me know if you want me to do it differently or use another name for the option.

matham: Yeah, this seems ok to me.

Lauszus: Okay super. Let me clean it up and test the wheels on RPi 3 and RPi 4, so I can make sure the change does not break anything. I hope to do that tomorrow.

Lauszus:
@matham I have now tested the wheels on both RPi 3 and RPi 4 and it works just fine. So this PR is ready to merge :)

Lauszus:
Argh it seems like it's not hardware accelerated:

```
[INFO   ] Window: Provider: sdl2
[INFO   ] GL: Using the "OpenGL ES 2" graphics system
[INFO   ] GL: Backend used <sdl2>
[INFO   ] GL: OpenGL version <b'OpenGL ES 3.1 Mesa 19.3.2'>
[INFO   ] GL: OpenGL vendor <b'VMware, Inc.'>
[INFO   ] GL: OpenGL renderer <b'llvmpipe (LLVM 9.0.1, 128 bits)'>
[INFO   ] GL: OpenGL parsed version: 3, 1
[INFO   ] GL: Shading version <b'OpenGL ES GLSL ES 3.10'>
```

rnixx:
are you compiling with VIDEOCOREMESA=1?

rnixx:
when i did experiments with mesa, vc4 and kivy, i also compiled mesa on my own.

the compile flags were

    ./autogen.sh \
            --prefix=/usr \
            --libdir=/usr/lib/arm-linux-gnueabihf \
            --with-gallium-drivers=vc4 \
            --with-dri-drivers= \
            --with-egl-platforms=x11,drm

my sdl2 compile flags were

    ./configure \
            --disable-pulseaudio \
            --disable-esd \
            --disable-video-mir \
            --disable-video-wayland \
            --disable-video-opengl \
            --disable-video-rpi \
            --host=arm-raspberry-linux-gnueabihf

So maybe you should check if vc4 is in mesa and, add --enable-video-kmsdrm to the flags above and compile kivy with VIDEOCOREMESA=1 for RPI4

Lauszus:
It was simply a permission issue. It works when I run my app using `sudo`.

For anyone else getting this issue, then you can enable debugging like so:

```
export LIBGL_DEBUG=verbose
```

Then you will see errors like this `drmOpenDevice: open result is -1, (Permission denied)` when running your app.

rnixx:
@Lauszus cool, great. would be good to find out what's needed to run program as user. running as root not sounds like a very good idea

Lauszus:
@rnixx I already compiled mesa manually, as I was running an old version of Buster. This is how I did it:

```bash
cd
git clone --depth=1 https://gitlab.freedesktop.org/mesa/drm
cd drm
mkdir build
cd build
sudo meson --prefix="/usr" ..
sudo ninja install

cd
git clone --depth=1 --branch mesa-19.3.5 https://gitlab.freedesktop.org/mesa/mesa
cd mesa
mkdir build
cd build
sudo meson --prefix="/usr" ..
sudo ninja install
```

Note to get all the needed dependencies you can enable sources in `/etc/apt/sources.list` and then running:

```bash
sudo apt-get update
sudo apt-get -y build-dep mesa
```

Lauszus:
@rnixx yes. I'm working on that right now.

Lauszus:
Found it! Had to add my user to the `render` group:

```bash
sudo adduser "$USER" render
```

ddimensia:
Hmm, I didn't have this issue when I was setting this up to verify my initial steps to get this working.  One thing I think I might have done, although it shouldn't be necessary, was to enable the fkms driver using raspi-config.  When you do this it does issue some apt commands to download a few packages, I wonder if one of those was changing permissions.

Lauszus:
@ddimensia I looked through the raspi-config script: https://github.com/RPi-Distro/raspi-config/blob/master/raspi-config, but it does not seem like they are doing anything special. FYI fkms was enabled by default on my system.

Lauszus:
@ddimensia I did some digging and it seems like a fairly new change to systemd: https://salsa.debian.org/systemd-team/systemd/commit/e3772a013721083a740ab9dedbf060cf5b3c3709. What is the group of your render (run `ls -l /dev/dri/renderD*`)?

rnixx:
@Lauszus I just gave it a try and at a first glance it works like charm!

pietervandermeer:
Finally! Thank you guys! The page out building for the Pi1..4 is really useful. I have to say that you still need to have a solid grasp of the SDL build configuration and where the shared libs will be located. Once that's out of the way it works fine. There's something strange going on with the frame rate, though.

Lauszus:
@pietervandermeer what issue do you see with the frame rate?

pietervandermeer:
Clock object reports high fps .. like 82 or so.. but it seems more like 30 + some tearing issues.

rnixx:
@Lauszus is there a reason you used libsdl 2.0.10 infavor of 2.0.9 which gets shipped with buster? I'm asking because i want do build custom deb packages from source.

Lauszus:
@rnixx nope! Would be great if you do that, that will speed up the installation a lot!

pietervandermeer:
Update on the tearing issues on pi4. I tested the 3Drendering example and setting config.ini maxfps=0 .. fps is 12000 which seems very healthy! but tearing still happens.. only after a second or two. coming from a computer graphics background i tested some of my own code and can verify that only every 3rd frame is actually visualized. i suspect that internally kivy/sdl uses a triple buffered screen and the buffer swap is turned off at some point. you're now using a single buffered screen (tearing artifacts) at one third of the desired frame rate. This is with Kivy 2.0.0.. built from git source.

matham:
@pietervandermeer Given what you said, maybe look at https://github.com/kivy/kivy/pull/4854. Maybe you'll have insight about that!?

pietervandermeer:
@matham thanks, i see this has a long history. but having a hard time linking this issue to what i'm experiencing. it looks like vsync is done.. but only the first 2 seconds. no vsync always means tearing if the graphics can be rendered faster than physical refresh rate. the patch was removed, so i guess there should actually be no vsync whatsoever?

pietervandermeer:
After some more testing it seems the issue occurs after a specific number of frames rendered. Wall clock time seems to play no role. I've now also tried the latest kivy dev :

[INFO   ] [Kivy        ] v2.0.0rc2, git-f044806, 20200604

pietervandermeer:
Ok guys. Sorry for all the chatter! I found out the reason for the frame rate troubles. I built SDL from mercurial repo. Apparently there was a bug in it or so. Pulled the latest and all is fine. Once again @Lauszus many thanks. You made my day! Kivy on Pi4 rocks now!

PR #6774, opened 2020-03-13
Build the extensions in parallel if the options has not been set

Lauszus:
This roughly halved my compilation times when using 8 threads instead of just 1.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -265,7 +265,8 @@ class KivyBuildExt(build_ext, object):
             # build will be disabled
             self.parallel = min(4, os.cpu_count() or 0)
             if self.parallel:
-                print('Building extensions in parallel')
+                print('Building extensions in parallel using {} cores'.format(
+                    self.parallel))
 
         global build_path
         if (self.build_lib is not None and exists(self.build_lib) and
--- a/setup.py
+++ b/setup.py
@@ -258,11 +258,18 @@ class KivyBuildExt(build_ext, object):
 
     def finalize_options(self):
         retval = super(KivyBuildExt, self).finalize_options()
+
+        # Build the extensions in parallel if the options has not been set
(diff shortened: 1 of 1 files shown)


matham:
Definitely a good idea, but I'm worried about making this decision for everyone. E.g. for make you can always specify how many threads.

I think we first should look at what other projects do (e.g. numpy), if they make it optional and what the default is!?

Lauszus:
Numpy has parallel build enabled by default, but limit to use maximum of 8 cores: https://github.com/numpy/numpy/blob/8087e1de04d011f7ce66bf80dbef4a0599b14f7a/numpy/distutils/misc_util.py#L78-L112.

This is where the actual code is compiled: https://github.com/numpy/numpy/blob/36a2bf0ce8b072d557af5b58a9003a6bc72caaef/numpy/distutils/ccompiler.py#L352-L361.

What I have done is similar however I do not limit the number of cores, but I can do that if you want to?

FYI here's the relevant code in distutils that checks the `parallel` attribute: https://github.com/python/cpython/blob/master/Lib/distutils/command/build_ext.py#L443-L469, as you can see if will get the number of cores on the CPU and use all of them, but if we set `parallel` to an integer, then that number of cores will be used.

matham:
Ok, makes sense. I'd say that by default we should also limit the number of parallel jobs to like 4 for similar reasons as numpy. Numpy is a much larger project with more code so it makes sense for them to benefit of using upto 8 cores. 

For us, there probably isn't much benefit above like 3 or 4 cores so lets limit it to the min of the number of cores and like 4.

Lauszus:
@matham fixed in: https://github.com/kivy/kivy/pull/6774/commits/0f57d47eac7de44372dcbdeb27416190358a4531.

PR #6787, opened 2020-03-19
Add ImageButton to style.kv

Cheaterman:
I essentially use this in ALL my projects. It would be a shame not to include it out of the box.
The only drawback in making it a dynamic class, is the lack of documentation/discoverability... If you think it should be a full-blown class then so be it!

More to the point of the "why" - I remember my early Kivy days (when behaviors weren't even a thing), I used the normal Button, set border to `(0, 0, 0, 0)` and changed `background_normal` and `background_down` to the SAME image. On all my buttons (until I discovered dynamic classes). You can see how it "feels" like excessive boilerplate for a framework like Kivy.
How hard can it be to make "an image as button" for a beginner who doesn't know about behaviors or dynamic classes?
How do you think they feel when their image is totally distorted because they didn't realize the Button background is a BorderImage and they need to set all borders to zero?
How do you think they feel when they find out the "complex" case (having a border image) takes as many lines to address as the simple one (just an image as button)?
When giving support to such users, wouldn't it be easier and faster to say "just use ImageButton, set the source, and do what you want in on_press"? :-)

Thanks for reading! Let's improve our QoL as Kivy coders, line by line :-)

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -27,6 +27,8 @@
             size: self.texture_size
             pos: int(self.center_x - self.texture_size[0] / 2.), int(self.center_y - self.texture_size[1] / 2.)
 
+<ImageButton@ButtonBehavior+Image>:
+
 <BubbleContent>
     opacity: .7 if self.disabled else 1
     rows: 1


Cheaterman:
(like, literally, if I make a project that's "big enough" to have a ScreenManager, you can be pretty sure there'll be an `ImageButton` rule in its main KVlang file...)

rnixx:
I'd prefer a "real" class. As you said for documentation reasons.

What i also have all over the place in my projects are ``IconButton`` and ``IconImage`` classes. They use icon fonts (https://github.com/kivy-garden/garden.iconfonts) to display icons as vector graphics on buttons and as images. If i find some time i'll open a PR with this classes at garden.iconfonts. IMHO garden.iconfonts even belongs to kivy core and should be integrated.

Cheaterman:
I honestly didn't know about iconfonts! I agree with you, if you use them all the time, you should PR these classes to their upstream, and if they're as good and useful as you make them look, they probably could be integrated upstream!

I'll try to make a self-documenting class version later today, thanks for your comment :-)

matham:
I'm not actually sure this belongs in kivy directly as opposed to simply added to the docs so people can copy this (as a recipe). I also have icon buttons etc. in my code, but there's both a normal button and a toggle button. And given that all you're doing is defining a class, being in the docs for people to copy should be good enough?

This is also part of my bias to not add any more widgets to kivy-core. We barely have a handle maintaining core kivy code. Maintaining widgets is harder by orders of magnitudes because it's not generally written with a very abstract API, which leads to many bugs as people try to use it in ways that is not what the original implementer was thinking of.

In the long term I even think we should have the kivy repo and a separate kivy repo or section in garden for kivy widgets that kivy depends on (although we may not be able to pull this off practically). Kinda like python's standard-library. This would let us separate our focus and maintaining core-kivy very well, but let the widgets be implemented in different ways in the side repo, with maybe a higher API turnover.

Because in truth, I and I think a few other core-devs also think of the built-in kivy widgets as demo widgets, as opposed to really well fleshed out widgets, with some obviously better than others.

Obviously doing something like that would require lots of thinking, but in the meantime my bias is to not include new widgets into kivy itself if we can. That's why I re-wrote the garden repos to support normal python packages to make it easier for our users to use. We just need to make the garden widgets more discoverable!

Now, you may wonder how simply adding a class definition would make the maintenance burden worse, but it's because once something goes in, it doesn't come out and it's part of the API. So maybe we may want to have similarly named icon button classes which has more features or whatever in the future but now it would be taken by this definiton. So since it's simply a class definition, on the balance I think this would be better off as a doc recipe than in kivy code.

Cheaterman:
That's a fair point matham, I guess we should have more snippets in the
documentation, and this should probably be part of them. Thanks for your
feedback!

matham:
I'll close this since we are not planning to merge, as discussed.

PR #6798, opened 2020-03-26
corrected platform detection on Android

### LINKED ISSUE #6797: incorrect platform detection on Android
The platform in the module kiv.utils returns linux on Android platform. It should return  Android.

Tested on a M10 Lenovo tablet running Android 8.1 with pydroid3.

olivier-boesch:
the function _get_platform was not detecting android. Android was detected as conventionnal linux.
Fixes #6797

### DIFF
--- a/kivy/utils.py
+++ b/kivy/utils.py
@@ -426,9 +426,11 @@ def _get_platform():
     # On Android sys.platform returns 'linux2', so prefer to check the
     # presence of python-for-android environment variables (ANDROID_ARGUMENT
     # or ANDROID_PRIVATE).
-    if 'ANDROID_ARGUMENT' in environ:
-        return 'android'
-    elif environ.get('KIVY_BUILD', '') == 'ios':
+    # it could be better to check for any ANDROID_* key in os.environ and return 'Android' when we find the first one
+    for key in environ.keys():
+        if match('ANDROID_', key):
+            return 'android'
+    if environ.get('KIVY_BUILD', '') == 'ios':
         return 'ios'
     elif _sys_platform in ('win32', 'cygwin'):
         return 'win'
--- a/kivy/utils.py
+++ b/kivy/utils.py
@@ -424,14 +424,10 @@ def format_bytes_to_human(size, precision=2):
 
 def _get_platform():
     # On Android sys.platform returns 'linux2', so prefer to check the
-    # presence of python-for-android environment variables (ANDROID_ARGUMENT
-    # or ANDROID_PRIVATE).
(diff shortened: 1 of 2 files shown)


olivier-boesch:
I modified the code to walk through os.environ keys and search for ANDROID_* keys. If I find one then we are on Android.

tshirtman:
Hm, this method wasn't really great to start with, and this increases the risk of false positive, if we don't find a better way to discriminate, i'd be fine with merging that, but i do wonder if we can't use something more specific like checking the presence of /default.prop or the existence of the logcat command, or the /system/xbin existing and being in the PATH, all of these i think being very rare things for a linux system that is not android.

matham:
I agree with @tshirtman - this seems pretty risky because I may have a variable named android in my env for unrelated reasons.

olivier-boesch:
I integrated your propositions. the three conditions that @tshirtman told are checked. 'android' is returned as platform if they are all met. works for my phone (s6 running android 7) and my tablet (lenovo M10 running android 8.1). Still returns 'linux' on my computer.

Can you test ?

kivy/utils.py
```diff
@@ -424,9 +424,8 @@ def format_bytes_to_human(size, precision=2):
 
 def _get_platform():
     # On Android sys.platform returns 'linux2', so prefer to check the
-    # presence of python-for-android environment variables (ANDROID_ARGUMENT
-    # or ANDROID_PRIVATE).
-    if 'ANDROID_ARGUMENT' in environ:
+    # existence of system specific commands and files
+    if path.exists('/default.prop') and path.exists('/system/bin/logcat') and path.exists('/system/xbin'):
```
tshirtman:
```suggestion
    if (
        path.exists('/default.prop')
        and path.exists('/system/bin/logcat')
        and path.exists('/system/xbin')
    ):
```
to stay above the 80 columns limit for pep8 :)
aside that, looks good to me, thanks :).

tshirtman:
oh and also the unittest for it must be adapted!

kivy/utils.py
olivier-boesch: ok. done.

olivier-boesch:
> oh and also the unittest for it must be adapted!
> https://github.com/kivy/kivy/pull/6798/checks?check_run_id=542406623#step:8:70

huu.. I'll work on this... mocking path and file existence doesn't seem easy...

olivier-boesch:
PEP8 Ok.
code seems ok.
Unit test seems ok.

olivier-boesch:
> Looks good to me, Thanks!

I'm happy to help. Kivy is a wonderful tool. :)

PR #6821, opened 2020-04-11
Expand on the current logger docs

Zen-CODE:
The current logger documentation does not really explain clearly what the Kivy logger offers and why you should use it. Because really you should when writing a kivy app....:-)

### DIFF
--- a/kivy/logger.py
+++ b/kivy/logger.py
@@ -55,8 +55,8 @@ some solid benefits and useful features. These include:
 * access to log history even if logging is disabled
 * built-in handling of various cross-platform considerations
 
-For these reasons, we advise that you use the Kivy Logger unless you have good
-reasons to choose otherwise.
+Kivys' logger was designed to be used with kivy apps and makes logging from
+Kivy apps more convenient.
 
 Logger Configuration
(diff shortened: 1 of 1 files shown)


kivy/logger.py
```diff
@@ -17,17 +25,40 @@
...
+some solid benefits and useful features. These include:
+
+* simplied usage (single instance, simple configuration, works by default)
+* color-coded output
+* output to `stdout` by default
+* message categorization via colon separation
+* access to log history even if logging is disabled
+* built-in handling of various cross-platform considerations
+
+For these reasons, we advise that you use the Kivy Logger unless you have good
```
matham, line 58: I'm hesitant to push for it that strongly. I would rather state it indirectly - "that kivy's logger was designed to be used with kivy apps and it makes logging from such app more convenient", and leave it at that.

Zen-CODE: Revised accordingly. Thanks

Zen-CODE: @matham Bump :-)

Zen-CODE:
Thanks :-)

PR #6844, opened 2020-04-27
Use GLES context when ES2 is forced

jakogut:
On the desktop, multiple context types are generally available,
including full OpenGL with core and compatibility contexts, in addition
to ES contexts.

Certain extensions, such as GL_OES_EGL_image_external require a GLES
context, but SDL2 defaults to a core or compatibility context when
available.

This patch forces a GLES context when ES2 is forced using KIVY_GRAPHICS.


matham:
I had to revert it: https://github.com/kivy/kivy/pull/6846.

PR #6852, opened 2020-05-02
:robot: Fix android platform detection when using p4a

opacam:
close #6851 

**Note:**still untested

PR #6854, opened 2020-05-02
:ok_hand: Reuse `kivy_build` var (complements #6852)

opacam:
See also: https://github.com/kivy/kivy/pull/6852#pullrequestreview-404528740


opacam:
Aaa...one test got stucked and when I restarted it fails...I think that we face this with p4a some weeks ago. I think it's github action checkout

Working on a fix...

See also: kivy/python-for-android#2070

PR #6855, opened 2020-05-02
:arrow_up: Upgrade to actions/checkout@v2 & actions/setup-python@v2

opacam:
To fix errors when restarting github actions tests like at #6854

See also: https://github.com/kivy/kivy/pull/6854/checks?check_run_id=638888562#step:2:388


opacam:
Ok, now restart github actions works ok...but it seems that `Install kivy` test for windows is failing...I will try to also upgrade actions/setup-python to v2 to see if solves this issue..

opacam:
Now we have the green tick mark, but I'm not sure that we actually solved the `Install kivy` test...anyway...I suppose we will see...

PR #6876, opened 2020-05-13
resources.py

salt-die:
Use explicit `str.startswith` for checking matches with start of string.
Rename iteration variable `dir_` as it reads better.
Should note that walrus fits in here:
```
if exists(path := abspath(filename)):
    return path
```

### DIFF
--- a/kivy/resources.py
+++ b/kivy/resources.py
@@ -59,6 +59,7 @@ def resource_find(filename):
         if exists(path):
             return path
 
+
 def resource_add_path(path):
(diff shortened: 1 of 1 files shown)


matham:
Thank you for your contribution, but this is more of a cosmetic change as it doesn't fix any bug or performance issue. We don't normally accept such changes to the codebase because at best it makes the code look nicer but also leads to code churn and possibly bugs. 

So functioning code is generally better left alone, unless there's a actual bug being fixed and the code is re-worked.

PR #6877, opened 2020-05-13
No need to import Properties when sub-classing Widget!

salt-die:
We add all properties to a Widgets dict in `__prepare__` with a ChainMap.  This allows, in the context of a Widget class def, to avoid importing any properties at all -- as property imports can get rather lengthy.  We also added a proof-of-concept auto-loader of kv lang which could be provided either in a special class attribute, `__KV__`, or a file in the same directory with the same name (in lowercase) as the widget.

Alternatively, auto-loading could be passed as a kwarg in class def, e.g.:
```py
class MyWidget(Widget, autoload_kv=True):
```

The end result is simpler widget creation, e.g.:
```py


class TestMeta(Label):
    __KV__ = """
    <TestMeta>:
        text: self.name
    """
    name = StringProperty()


if __name__ == '__main__':
    from kivy.app import App

    class MyApp(App):
        def build(self):
            return TestMeta(name='Hello world!')

    MyApp().run()
```

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -289,8 +289,9 @@ class WidgetMetaclass(type):
         return ChainMap({}, all_properties)
 
     def __new__(metas, name, bases, methods):
-        # Remove Properties from __dict__
-        methods = methods.maps[0]
+        if hasattr(methods, 'maps'):
+            # Remove Properties from __dict__
+            methods = methods.maps[0]
         return super(WidgetMetaclass, metas).__new__(metas, name, bases, methods)
 
     def __init__(cls, name, bases, attrs):
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -249,12 +249,12 @@ from kivy.base import EventLoop
 from kivy.lang import Builder
 from kivy.context import get_current_context
 from kivy.weakproxy import WeakProxy
-from collections import ChainMap           
+from collections import ChainMap
 from functools import partial
-from inspect import getfile                
+from inspect import getfile           
 from itertools import islice
-from os.path import dirname, join, exists  
-from textwrap import dedent                
+from os.path import dirname, join, exists
+from textwrap import dedent  
 
 all_properties = {name: getattr(properties, name) for name in properties.__all__}
 
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -280,6 +280,7 @@ class WidgetException(Exception):
 class WidgetMetaclass(type):
     '''Metaclass to automatically register new widgets for the
     :class:`~kivy.factory.Factory`.
+
     .. warning::
         This metaclass is used by the Widget. Do not use it directly!
     '''
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -238,6 +238,7 @@ __all__ = ('Widget', 'WidgetException')
 
 from kivy.event import EventDispatcher
 from kivy.factory import Factory
+import kivy.properties as properties
 from kivy.properties import (
     NumericProperty, StringProperty, AliasProperty, ReferenceListProperty,
     ObjectProperty, ListProperty, DictProperty, BooleanProperty)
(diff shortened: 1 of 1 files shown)


tshirtman:
I'm not sure about this, sure, it makes things slightly easier, but i'm not a fan of magically defining variables in the code, one thing i like a bout python is that even "self" is coming from somewhere obvious, you can always follow up the definition of something (unless you imported `*` but that's widely considered bad practice). And i don't think having to import properties is such an hassle, at worst you can decide to `from kivy import properties` and then use `properties.NumericProperty` (or even alias `properties` to something shorter).

The automatic loading of kv sounds cool though, that's more in line with the magic kivy currently does of loading the app's kv file, but allows for a better code organisation. I've used similar things (but slightly less automatic) in my projects and it was useful.

One reason not to encourage splitting kv in multiple files is in regard to performances, as IO is slower on lower end mobile devices, more files to load means worse performances, so one kv file for eachh widget class can be an expensive design, that could be solved by packaging tools, but that does add complexity.

The feature is cool, still, maybe the consequences can be documented.

matham:
So I just want to say first that I really appreciate you contributing to kivy, but I do have some issues with the PR, which I hope you take purely as feedback to how I think about Kivy and frameworks/libraries in general and what is good for the long-term benefit for kivy.

As @tshirtman said, I'm also not a fan at all of magic, expect where it's truly needed and makes things significantly easier. One of the guiding principles of Python is explicit is better than implicit and some of the long-standing issues in kivy is because we already do too much magic without it being clear to the user when or how we do it until they run into inconsistencies.

I see 3 changes here:

1. **Automatically making properties available as a widget attribute**: I think this is too magical and having clear imports at the top is desirable. Both for new users so the program logic can be followed more clearly and to be able to have your own property classes and being able to track clearly where these class are defined and whether they clash with user defined property classes. But overall, this just does too much magic without a big enough benefit to justify this IMO. If typing imports is too much, you can probably configure your code editor to auto-import them for you.

   I'm a pretty slow typer, but this has never quite been an issue for me.
2. **kv files are automatically loaded based on the the class name**: I also think at a framework level we shouldn't make this kind of decision on how users structure their kv files as that should be up to the app developer. Specifically, making this decision on our end will leave it in our hands as to when kv files are loaded for a user app, but normally the developer knows best when and if a kv file should be loaded.

   Yes, this is optional as users don't have to name files as such, but (a) this could lead to issues if they accidentally name files like a class name and they wonder why it's auto loading - which is part of the issue of this being too magical and (b) adding support for this will make kivy more opinionated with respect to how users may structure kv files. Granted, if being opinionated helps our users have better code structure we should be opinionated, but the problem here is that I don't see a overall reason why this would be a good structure to encourage. And perhaps this would be a bad structure for performance or other reasons as @tshirtman mentioned.

   To accept such a change you'd have to show that typical kv developers would benefit from this change and then we could consider it. But in my experience this is not something that would be commonly helpful.
3. **Defining class level kv-code in with `__kv__`**: I quite like this actually, however, there are two issues I see with this. 

   1. This will auto-load kv code into the default kivy factory at the first time the module is imported. This may seemingly be only a good thing, and it can be, but this leads to the same issues of not being able to overwrite whatever default kv style you define for a class. 

      With your current code, sub-classes will simply add, rather than overwrite kv rules - like the normal kv file behavior. But the difference is, these `__kv__` rules will be loaded whenever the python code is imported. So unlike kv files where we can make some mechanisms to selectively load them (e.g. discussions about breaking up kivy's builtin style.kv and make components optional), these kv rules all always be loaded and are not optional once someone defines them for a class. 

      And even though we can clear rules, it's still initially loaded and you have to explicitly clear them in your own rule, which may not be desirable.

      Maybe we can add a `_kv_default_` flag that when set will overwrite, instead of adding to the class. Or maybe a class level `load_kv` that loads the kv and is called by the metaclass, but that gives the user a chance to customize loading the kv for the subclass.
   2. These rules are registered with the default factory, as I mentioned. The issue there is that sometimes, especially when testing, you want to use multiple kv factories to encapsulate the app. So we will need to outline a strategy about how to auto reload them into new factories. It may not be hard, but we do need to add support for it once we manage loading kv code as in the PR.
   3. Purely cosmetic, but kv files tend to be long, have long lines, and look ugly relative to python code  :( So I'm not sure it'd be good fit to include in python code as an official pattern, rather than now where it's sometimes included with `load_string` but is not super common. And this is considering that `__kv__` already is a few levels indented so you already lose some line width.
      Also, we think of kv files as styling the python classes; so it maybe conceptually weird to include styling directly in python.

So in summary, I'm partially in favor of number three, but only if it's part of an overall strategy of handling multiple factories, making the kv code loading optional (including the related problem of making kivy's style.kv optional), and theming considerations. And only if other kivy-devs think this would make real-world kv code handling easier for users (I'm on the fence).

matham:
As mentioned, I don't think this is the correct approach for Kivy itself, but it can be a user lib if people find this useful. So I'm closing.

PR #6879, opened 2020-05-15
Fix key_background_color property is not used

matthiasharrer:
When trying to style the VKeyboard Widget I came across the key_background_color property that is also described in the [docs](https://kivy.org/doc/stable/api-kivy.uix.vkeyboard.html#kivy.uix.vkeyboard.VKeyboard.key_background_color).

It however wasn't used when rendering the keys. This PR tries to fix this and applies the color when drawing the keys.

### DIFF
--- a/kivy/uix/vkeyboard.py
+++ b/kivy/uix/vkeyboard.py
@@ -562,10 +562,9 @@ class VKeyboard(Scatter):
         texture = Image(background, mipmap=True).texture
 
         with self.active_keys_layer:
-            Color(1, 1, 1)
+            Color(*self.key_background_color)
             for line_nb, index in active_keys.values():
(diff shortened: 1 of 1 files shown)


tshirtman:
indeed looks like it was unduly removed in 7b3b209f4332e31a0cdad18c2fab5e678b5bc01b, i guess removing it was "speeding up the loop" as it doesn't need to be applied every time, but it should be at least applied once, before the loop.

matthiasharrer:
Sure :)

matthiasharrer:
It seems it did never exist for the active keys though .. not sure if this is intentional or not.

tshirtman:
yes, in any case it seems that it has been wrong for a long time, thanks for the fix!

PR #6881, opened 2020-05-16
Add variable frame delay for GIF animated images

yobagram:


kivy/core/image/__init__.py
```diff
@@ -644,10 +685,35 @@ def _set_anim_delay(self, x):
...
+
+    def _set_durations(self, x):
+        # Waiting for list of ints or floats in microseconds like in GIF format, not seconds
+        if x is None:
+            self._durations = None
+            return
+        if not(isinstance(x, list) and all(isinstance(elem, (int, float)) for elem in x)) and len(x) > 0:
+            raise Exception('Property durations must be list of int or float with len > 0')
+
+        if all(elem == x[0] for elem in x[1:]):
```
tshirtman, line 708: many way of doing that (https://twitter.com/raymondh/status/1254126501856870400) i think it's better to compare [0] against itself than duplicating the list, so `all(e = x[0] for e in x)`.

kivy/core/image/__init__.py
```diff
@@ -644,10 +685,35 @@ def _set_anim_delay(self, x):
...
+
+    def _get_durations(self):
+        return self._durations
+
+    def _set_durations(self, x):
+        # Waiting for list of ints or floats in microseconds like in GIF format, not seconds
+        if x is None:
+            self._durations = None
+            return
+        if not(isinstance(x, list) and all(isinstance(elem, (int, float)) for elem in x)) and len(x) > 0:
```
tshirtman, line 705: better split the test over multiple lines for readability
```
        if not (
            isinstance(x, list)
            and all(isinstance(elem, (int, float)) for elem in x)
            and len(x) > 0
        ):
```
(also i think the closing parenthesis of the not mas misplaced).

kivy/core/image/__init__.py
```diff
@@ -644,10 +685,35 @@ def _set_anim_delay(self, x):
...
+            raise Exception('Property durations must be list of int or float with len > 0')
+
+        if all(elem == x[0] for elem in x[1:]):
+            # durations are constant, so we have standart anim_delay case
+            self._durations = None
+        else:
+            self._durations = [elem / 1000 for elem in x]
```
tshirtman, line 712: i think we shouldn't change the values here, if i put a value x in a property, i expect to find x in the property later, not x / 1000.

kivy/core/image/__init__.py
```diff
@@ -677,6 +743,12 @@ def _img_iterate(self, *largs):
         imgcount = len(self.image.textures)
         if imgcount > 1:
             self._anim_available = True
+
+            if (self._auto_anim_delay
+                    and imgcount == len(self.image.durations)
+                    and all(isinstance(elem, int) for elem in self.image.durations)):
```
tshirtman, line 749: fix indentation here.
```python
            if (
                self._auto_anim_delay
                and imgcount == len(self.image.durations)
                and all(isinstance(elem, int) for elem in self.image.durations)
            ):
```

kivy/core/image/img_pil.py
```diff
@@ -92,7 +92,8 @@ def _img_read(self, im):
                     img_tmp = img_ol
                 img_ol = img_tmp
                 yield ImageData(img_tmp.size[0], img_tmp.size[1],
-                                img_tmp.mode.lower(), img_tmp.tobytes())
+                                img_tmp.mode.lower(), img_tmp.tobytes(),
+                                duration=img_tmp.info.get('duration', None))
```
tshirtman, line 96: maybe do the division logic here?

kivy/uix/image.py
```diff
@@ -169,7 +169,18 @@ def get_image_ratio(self):
     defaults to False.
     '''
 
-    anim_delay = NumericProperty(.25)
+    durations = ListProperty(None, allownone=True)
+    '''Delay the animation if the image is sequenced (like an animated gif).
+    Set if frame delay is variable, not constant.
+    Ignored if anim_delay is set to -1 or auto_anim_delay set to True.
+
+        .. versionadded:: 1.0.8
```
tshirtman, line 177: 2.0.0

kivy/uix/image.py
```diff
@@ -188,6 +199,16 @@ def get_image_ratio(self):
     defaults to 0.
     '''
 
+    auto_anim_delay = BooleanProperty(False)
+    '''If this property is set True and image is GIF animation,
+    delay of each frame would be get from GIF-image file.
+    
+    .. versionadded:: 1.6.0
```
tshirtman, line 206: 2.0.0

matham: Why do we need to define both `auto_anim_delay` and `durations`? Basically, if `durations` is truthy, that's used, otherwise it uses `anim_delay`.

yobagram: > Why do we need to define both `auto_anim_delay` and `durations`? Basically, if `durations` is truthy, that's used, otherwise it uses `anim_delay`.

Compatibility. It must be auto_anim_delay == False and durations is None.

matham: But why? By default `durations` is None so the old behavior is used. It's only if `durations` is set that the new behavior is used. Unless I'm reading the code wrong.

kivy/core/image/__init__.py
```diff
@@ -278,6 +283,8 @@ def populate(self):
             # set as our current texture
             self._textures.append(texture)
 
+            self._durations.append(self._data[count].duration)
```
tshirtman, line 286: @matham the data is yielded in im_pil and is stored here, so duration is not empty by default. I'm not sure about the code that tests that durations is not None, because i don't see how the users could have a case of durations==None.

yobagram: > @matham the data is yielded in im_pil and is stored here, so duration is not empty by default. I'm not sure about the code that tests that durations is not None, because i don't see how the users could have a case of durations==None.

We use this in kivy.core.image.Image._img_iterate() if auto_anim_delay is True. So? steps are
create Image with parameters
Image.__init__()
1) if auto_anim_delay == False and durations == None then like before
2) if auto_anim_delay == False and durations then Image.durations = durations from creating parameters
3) if auto_anim_delay == True then when in _img_iterate() we set self.durations = self.image.durations - thiese durations are from GIF file

In case 3) durations from creating parameters are not used.

kivy/uix/image.py
yobagram: > But why? By default `durations` is None so the old behavior is used. It's only if `durations` is set that the new behavior is used. Unless I'm reading the code wrong.

I tried to answer below

kivy/core/image/__init__.py
```diff
@@ -626,28 +658,68 @@ def anim_reset(self, allow_anim):
...
+            if self._anim_delay >= 0 and not self._durations:
                 self._anim_ev = Clock.schedule_interval(self._anim,
                                                         self._anim_delay)
 
     anim_delay = property(_get_anim_delay, _set_anim_delay)
+
+    def _get_durations(self):
+        return self._durations
+
+    def _set_durations(self, x):
```
tshirtman, line 698: i'd rename `x` to `durations` or maybe "value" if you don't mind (i think i prefere `durations`), short variables names are only ok in very small contexts or when their name is really explicit.

yobagram: Yes, of course.

matham:
Ok, makes sense. Long term though, wouldn't you say that by default we should be using animation delays provided by the gif loader? Isn't that something you'd almost always want?

If so, maybe we should break backward compat (this is 2.0.0 after all) and have it default to True? Alternatively, we should make it clear in the docs that maybe in a couple of releases we'll change it to default to True?

yobagram:
> Ok, makes sense. Long term though, wouldn't you say that by default we should be using animation delays provided by the gif loader? Isn't that something you'd almost always want?
> 
> If so, maybe we should break backward compat (this is 2.0.0 after all) and have it default to True? Alternatively, we should make it clear in the docs that maybe in a couple of releases we'll change it to default to True?

@matham @tshirtman 
I agree, but I don’t know how things like this are being solved, maybe you will somehow agree among themselves, core developers? Default sets here
kivy/uix/image.py:202
`    auto_anim_delay = BooleanProperty(False)
    '''If this property is set True and image is GIF animation,
    delay of each frame would be get from GIF-image file.`

kivy/core/image/__init__.py
```diff
@@ -626,28 +658,68 @@ def anim_reset(self, allow_anim):
...
+            self._durations = None
+            # if durations set None stop until anim_delay set again
+            self.anim_reset(False)
+            return
+        if x and not(isinstance(x, list)
+                     and all(isinstance(elem, (int, float)) for elem in x)):
+            raise Exception('Property durations must be list of int or '
+                            'float with len > 0')
+
+        if all(elem == x[0] for elem in x):
```
matham, line 711: There's really no reason to do this optimization, is there? It's simpler to trust the durations passed in. And if they are all the same the user would just have used `anim_delay` directly. And either way I don't think this would make it that much faster to be worth the added complexity of checking all the elements.

kivy/core/image/__init__.py
```diff
@@ -626,28 +658,68 @@ def anim_reset(self, allow_anim):
...
+        # Waiting for list of ints or floats in microseconds like in GIF
+        # format, not seconds
+        if x is None:
+            self._durations = None
+            # if durations set None stop until anim_delay set again
+            self.anim_reset(False)
+            return
+        if x and not(isinstance(x, list)
+                     and all(isinstance(elem, (int, float)) for elem in x)):
+            raise Exception('Property durations must be list of int or '
```
matham, line 708: What is the reason for checking the types of the elements? The standard way is to do `durations = list(x)`, this creates a copy of the durations in case the user modifies it after passing in. And if the user passed in a non-numeric item, it'll crash soon enough when it's used (and see about duck typing). So I'd suggest we don't do this error checking.

kivy/core/image/__init__.py
yobagram: **Durations from the GIF file get here too.** GIF-file with constant delay is more common than with variable. We check it once and use many times with common cases with constant delay. So, the real standart case is Image('image.gif'):
1) load GIF-file
2) GIF-file has constant delay(common case)
3) set this constant delay as anim_delay
4) create once Clock.schedule_interval()

matham: Additionally, if we can't simply overwrite `anim_delay` that may have been set by the user.

kivy/core/image/__init__.py
yobagram: Ok I will remove this check

kivy/core/image/__init__.py
matham: But what is wrong with simply using the list of durations when they are all the same? I.e. even if it is all the same (i.e. `all(elem == x[0] for elem in x)` is True) why can't we do `self._durations = x`?

Is it possible that the length of x is not the same as the length of the frames? If that's what you're trying to check, then it's better to check for this directly!?

yobagram: If we want to consider anim_delay by default, we should set auto_anim_delay False. We must consider only one thing as dominant. We can't consoder durations as dominant and anim_delay as dominant together. I can rewrite code for one dominant or another but not for both. e.g. user set anim_delay and durations both what should we do? Combine anim_delay and durations in one property? I tried, but could not do it normally without errors.

yobagram: > But what is wrong with simply using the list of durations when they are all the same? I.e. even if it is all the same (i.e. `all(elem == x[0] for elem in x)` is True) why can't we do `self._durations = x`?
> 
> Is it possible that the length of x is not the same as the length of the frames? If that's what you're trying to check, then it's better to check for this directly!?

There is one reason. I think Clock.schedule_interval() once is faster then creating Clock.schedule_once() every frame.

matham: I may be misunderstanding something, but why can't we simply replace

```py
        if all(elem == x[0] for elem in x):
            # durations are constant, so we have standart anim_delay case
            self._durations = None
            self.anim_delay = x[0]  # anim resets inside _set_anim_delay()
                                    # if not durations
        else:
            # reset and stop for error case when durations count < frames count
            self.anim_reset(False)
            self._durations = x
            self.anim_reset(True)
```
with 
```py
# reset and stop for error case when durations count < frames count
self.anim_reset(False)
self._durations = x
self.anim_reset(True)
```

You mentioned about dominance, but isn't durations used first and then `anim_delay` if durations is empty? That's how it should be in my POV. But either way, if can be converted to a list successfully, we can just keep it as a list of durations, even if the durations are all the same values in the list!?

yobagram: I think there is reason not to change durations, ok. The user should expect the durations to remain in place. But I think we should still call once schedule_interval() for constant durations case just for performance.

matham: >There is one reason. I think Clock.schedule_interval() once is faster then creating Clock.schedule_once() every frame.

That's a good point, but as you mentioned, checking floats for equality is not exactly correct, so I think from a correctness POV it's better to be simply keep the list of durations rather than rounding and comparing values.

Also, I'd consider it a premature optimization. You're probably right it's a little faster, but in the grand scheme of things, drawing and copying the image probably dominates any speed issues so an additional `Clock.schedule_once()` once per frame is not likely to be noticeable. Especially since this should be quite small. Unless you have measured it.

Additionally, if you cared about performance and measured and found it is causing some lag, the better way is to create a single event with `event = Clock.create_event(...)` and then for each image manually set the timeout for the event and then cancel and trigger it with `event.cancel(); event()`. But all this is not really needed unless we have actual performance profiling data.

yobagram: Trust you, I don’t know as well what's going on under the hood as you. So I will do as you said: schedule_once every frame.

yobagram:
@matham could you see my updated code

kivy/core/image/__init__.py
```diff
@@ -626,28 +658,57 @@ def anim_reset(self, allow_anim):
...
+        # Waiting for list of ints or floats in microseconds like in GIF
+        # format, not seconds
+        if durations is None:
+            self._durations = None
+            # if durations set None stop until anim_delay set again
+            self.anim_reset(False)
+            return
+
+        self.anim_reset(False)
+        self._durations = durations
```
matham, line 708: I might do `self._durations = list(durations)`, but that's up to you.

kivy/core/image/__init__.py
```diff
@@ -677,7 +738,14 @@ def _img_iterate(self, *largs):
         imgcount = len(self.image.textures)
         if imgcount > 1:
             self._anim_available = True
-            self.anim_reset(True)
+
+            if (self._auto_anim_delay
```
matham, line 742: I think this should be more like:

```py
if self._auto_anim_delay:
    if imgcount != len(self.image.durations):
        raise ValueError(...)
    self.durations = self.image.durations
else:
...
```
Because we shouldn't quietly move on if there's an error and not matching lengths is an error (I believe)!? Similarly, do we need to check whether it's a float/int? If it isn't an error will be raised when you try to use it.

If you really want to check for an error, maybe above instead of doing `self._durations = list(durations)` like I suggested, simply do `self._durations = list(map(float, durations))`. This will convert everything floats and raise an error immediately if it can't. Then you don't have to do any checking here.

yobagram:
@matham Could you please review my code again?
1) I deleted this check because taking these durations from GIF file and then everything goes as if you explicitly set the durations.
2) Now if we specify the duration explicitly, we consider the automatic setting to be false

There is little merge conflict. Should I resolve it in my branch locally and push or is it better not to touch?

kivy/core/image/__init__.py
```diff
@@ -521,7 +539,21 @@ def __init__(self, arg, **kwargs):
...
         self._anim_index = 0
         self._anim_delay = 0
+        # if duration every frame is variable
+        self._durations = None
         self.anim_delay = kwargs.get('anim_delay', .25)
+
+        # if True: get real frame duration from file info if it possible
+        # otherwise like False
+        # if False: get anim_delay only from kwargs if it possible or default
+        self._auto_anim_delay = kwargs.get('auto_anim_delay', False)
```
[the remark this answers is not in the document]
matham: Also, shouldn't this default to True?

kivy/uix/image.py
```diff
@@ -168,7 +168,18 @@ def get_image_ratio(self):
     defaults to False.
     '''
 
-    anim_delay = NumericProperty(.25)
+    durations = ListProperty(None, allownone=True)
+    '''Delay the animation if the image is sequenced (like an animated gif).
+    Set if frame delay is variable, not constant.
+    Ignored if anim_delay is set to -1 or auto_anim_delay set to True.
```
matham, line 174: Isn't it the opposite that `auto_anim_delay` is ignored if `durations` is specified?

So if I understand it correctly, this should say: 
```
"Ignored if :attr:`anim_delay` is set to ``-1``. If :attr:`durations` is specified, :attr:`auto_anim_delay` will be ignored and treated as if it's False."
```

Unless you meant like you wrote it, and `auto_anim_delay` is dominant? But then above `            self._auto_anim_delay = False` should probably be deleted!? Given that we have `auto_anim_delay` it makes sense to me that it should be dominant, even when `durations` is specified. Because otherwise what's the point of having it? 

From your implementation I'm thinking that maybe you'd prefer `durations` to be dominant if specified. But if not specified we look at `auto_anim_delay` to decide whether to use the constant delay or the file provided delays. Is that right? 

A third alternative is for `auto_anim_delay` to be dominant, but fallback on `durations` or `anim_delay` when it's True but the file loader doesn't provide delays for the images. I think this version makes the most sense to me, but we should decide which you like and make sure the docs are consistent with it :)

kivy/uix/image.py
```diff
@@ -187,6 +198,16 @@ def get_image_ratio(self):
     defaults to 0.
     '''
 
+    auto_anim_delay = BooleanProperty(True)
+    '''If this property is set True and image is GIF animation,
+    delay of each frame would be get from GIF-image file.
```
matham, line 203: Similarly, shouldn't this also additionally say, 

```
If :attr:`durations` is specified, :attr:`auto_anim_delay` will be ignored and treated as if it's False and :attr:`durations` will be used to compute the delays.
```
?

kivy/uix/image.py
```diff
@@ -168,7 +168,18 @@ def get_image_ratio(self):
...
+
+        .. versionadded:: 2.0.0
+
+        :attr:`durations` is a :class:`~kivy.properties.ListProperty` and
+        defaults to None.
+        '''
+
+    anim_delay = NumericProperty(0.25)
     '''Delay the animation if the image is sequenced (like an animated gif).
     If anim_delay is set to -1, the animation will be stopped.
```
matham, line 184: You should probably also add here,
```
Please also see :attr:`auto_anim_delay`.
```

yobagram:
@matham Could you please see my code again?
I changed the use of parameters and described them in code doc-strings. Fix code for using with AsyncImage and zipsequence.
Thank you for your patience with parsing my fixes.

kivy/uix/image.py
```diff
@@ -250,42 +290,70 @@ def texture_update(self, *largs):
         if not self.source:
             self._clear_core_image()
             return
+
+        self._loops = 0
```
pythonic64, line 294: This line should be moved to after line 303 because we reset `_loops` if `source` exist and it can be found, otherwise `_clear_core_image` will do the job.

kivy/uix/image.py
```diff
@@ -250,42 +290,70 @@ def texture_update(self, *largs):
...
     def on_anim_delay(self, instance, value):
         if self._coreimage is None:
             return
         self._coreimage.anim_delay = value
-        if value < 0:
+        if value < 0 and not(self.auto_anim_delay or self.durations):
             self._coreimage.anim_reset(False)
 
+    def on_auto_anim_delay(self, instance, value):
+        if self._coreimage is None:
```
pythonic64, line 332: Can be replace with:
```
if self._coreimage:
    self._coreimage.auto_anim_delay = value
```

No need to test specifically for `None` and it saves one line.

kivy/uix/image.py
```diff
@@ -250,42 +290,70 @@ def texture_update(self, *largs):
...
+            return
+        self._coreimage.auto_anim_delay = value
+
+    def on_durations(self, instance, value):
+        if self._coreimage is None:
```
pythonic64, line 337: Same case as for `def on_auto_anim_delay`.

yobagram:
@pythonic64 Thanks, I fixed.

kivy/core/image/__init__.py
```diff
@@ -626,28 +679,67 @@ def anim_reset(self, allow_anim):
...
             self._anim_ev = None
 
-        if allow_anim and self._anim_available and self._anim_delay >= 0:
-            self._anim_ev = Clock.schedule_interval(self._anim,
-                                                    self.anim_delay)
-            self._anim()
+        if allow_anim and self._anim_available:
+            if self._durations:
+                if not len(self._durations) == len(self.image.durations):
+                    raise Exception(
```
tshirtman, line 685: Maybe make that a ValueError?

tshirtman (approved):

tshirtman (dismissed):

matham:
I'd feel much more comfortable if there were tests testing these new options/properties. Because I'm not sure it's working correctly.  I see I have some requested changes that I started working on back then but didn't complete, so I'll post it now, but since I didn't spend time on it since, you can just think of it as potential concerns rather than for sure issues.

PR #6887, opened 2020-05-20
Wait for ffpyplayer to fully load a video before playing to prevent a/v desync

snuq:
Videos (especially obvious with uhd videos) played with ffpyplayer would start the audio playback before the video is ready causing a desync and severe stuttering in the video.  This commit helps prevent that by waiting for ffpyplayer to load the first frame.

Note that the sleep is at the end of the last while loop is for a good reason too - if the video starts playing immediately after the frame is gotten, it will still stutter a decent amount, delaying by .1 second helps that significantly.

### DIFF
--- a/kivy/core/video/video_ffpyplayer.py
+++ b/kivy/core/video/video_ffpyplayer.py
@@ -218,6 +218,12 @@ class VideoFFPy(VideoBase):
         did_dispatch_eof = False
         seek_queue = self._seek_queue
 
+        # Wait for first frame to load before playing so audio doesnt get out of sync too badly
+        frame = None
+        while frame is None:
+            frame, value = ffplayer.get_frame(force_refresh=True)
+            sleep(0.1)
+
         # fast path, if the source video is yuv420p, we'll use a glsl shader
         # for buffer conversion to rgba
         while not self._ffplayer_need_quit:


kivy/core/video/video_ffpyplayer.py
```diff
@@ -218,6 +218,12 @@ def _next_frame_run(self):
         did_dispatch_eof = False
         seek_queue = self._seek_queue
 
+        # Wait for first frame to load before playing so audio doesnt get out of sync too badly
+        frame = None
+        while frame is None:
+            frame, value = ffplayer.get_frame(force_refresh=True)
+            sleep(0.1)
+
```
matham, line 226: I'm not sure this is ok, because this can lead to never timing out and waiting here forever, e.g. if this file happens to not actually have video, just audio. Also, this leads to discarding the first video frame.

But I'm not sure what this solves? Does the audio start playing before the video? This shouldn't happen because the player starts out paused so it shouldn't start playing audio until we unpause it, which should happen when the first frame is about ready when the metadata becomes available.

snuq:
i didnt think about the issue about loading audio-only files... is there a way to detect that?  i can definitely implement a limit on the loops, that would help with possible video loading problems, but that would be unacceptable for the audio only files...

to illustrate why i made this change, here's what happens when i try to play a 4k 10bit video:
[http://www.snuq.com/snuphotomanager/testvideos/original.mp4](url)
is what happens with the original play function,
and:
[http://www.snuq.com/snuphotomanager/testvideos/withdelay.mp4](url)
is what happens when its delayed a bit.  still has a bit of a skip, but the first is REALLY bad.

matham:
I'm going to close since as I mentioned I don't think this is correct.

snuq:
Is there another way to improve this problem then? the audio desync issue is pretty bad for UHD videos.  Additionally, this fix does seem to slightly improve the issue where it will loop the last second of audio several times when the video finishes playing (see issue #7416 ), tho it does not fix that problem entirely.

PR #6897, opened 2020-05-26
Add class that sets RV data using kv ids.

### LINKED ISSUE #6878: Allow using kv ids in RecycleView data keys
**Software Versions**
* Python: 3.6.9 
* OS: Ubuntu 18.04.4 LTS
* Kivy: 1.11.0
* Kivy installation method: PIP

**Describe the bug**
The official information regarding recycleview shows the following: 
```yaml
<RV>:
    viewclass: 'Label' ### Here
    RecycleBoxLayout:
        default_size: None, dp(56)
        default_size_hint: 1, None
        size_hint_y: None
        height: self.minimum_height
        orientation: 'vertical'
```

Where viewclass is a simple Label widget with the attribute "text".  Created Label widgets within RV are populated as follows:

```python
class RV(RecycleView):
    def __init__(self, **kwargs):
        super(RV, self).__init__(**kwargs)
        self.data = [{'text': str(x)} for x in range(100)] ###### here one can see the class attribute 'text'
```

If one creates a more complicated widget for the viewclass within RecycleView as follows: 

```yaml
<MyRecycleView>:
    viewclass: 'ThisInfo' ### Mix of widgets
    RecycleBoxLayout:
        orientation: 'vertical'
        default_size: None, dp(120)
        default_size_hint: 1, None
        size_hint_y: None
        height: self.minimum_height
        spacing: 3

<ThisInfo@BoxLayout>:
    BoxLayout:
        size_hint_x: 1.3
        orientation: 'vertical'
        Label:
            id: name                 ### I should be able to modify this 'text' attribute from self.data by its id
        GridLayout:
            cols: 2
            Label:
                text: 'Position'
            Label:
                text: 'Salary'
            Label:
                id: position         ## Same here
            Label:
                id: salary           ### And here
```

And tries to populate in any of this ways, it does not populate any info, all the id fields appear blank.

```python
############### Method 1, placing the ID and attribute of the pertinent widget ###############
class MyRecycleView(RecycleView):
    def __init__(self,  **kwargs):
        super(MyRecycleView, self).__init__(**kwargs)

        self.data = [{'name.text': employee.name, 
                            'position.text': employee.position, 
                            'salary.text': employee.salary} 
                                 for employee in all_employees] 

######## Method 2, placing Id as the key of the widget and a dictionary for all of its attributes #####
class MyRecycleView(RecycleView):
    def __init__(self,  **kwargs):
        super(MyRecycleView, self).__init__(**kwargs)

        self.data = [{'name': {'text': employee.name}, 
                           'position': {'text': employee.position},
                           'salary': {'text':  employee.salary}} 
                               for employee in all_employees]
```

**Expected behavior**
The behavior is not consistent, when using a base class as Label, one may populate its attributes directly in the self.data list with 'text'. When using a mixture of widgets with separate ids one should be able to populate its attributes by its ID just as in all the rest of the kivy framework.

matham:
Adds the `RecycleKVIDsDataViewBehavior` class. Fixes #6878.

It's similar to the existing `RecycleDataViewBehavior` class, except that the data can signify properties of objects named with an id in KV. E.g. given a KV rule::

```
<MyRule@RecycleKVIDsDataViewBehavior+BoxLayout>:
    Label:
        id: name
    Label:
        id: value
```

Then setting the data list with `rv.data = [{'name.text': 'Kivy user', 'value.text': '12'}]` would automatically set the corresponding labels.


kivy/uix/recycleview/views.py
```diff
@@ -108,6 +111,34 @@ def apply_selection(self, rv, index, is_selected):
...
+
+        <MyRule@RecycleKVIDsDataViewBehavior+BoxLayout>:
+            Label:
+                id: name
+            Label:
+                id: value
+
+    Then setting the data list with
+    ``rv.data = [{'name.text': 'Kivy user', 'value.text': '12'}]`` would
+    automatically set the corresponding labels.
```
tshirtman, line 128: a versionadded flag for the class would be nice.

kivy/uix/recycleview/views.py
```diff
@@ -108,6 +111,34 @@ def apply_selection(self, rv, index, is_selected):
...
+    """
+
+    def refresh_view_attrs(self, rv, index, data):
+        sizing_attrs = RecycleDataAdapter._sizing_attrs
+        for key, value in data.items():
+            if key not in sizing_attrs:
+                name, *ids = key.split('.', maxsplit=1)
```
tshirtman, line 135: why the star? the maxsplit makes it a 2-tuple, so we should be able to directly use the value without indexing on the next line.

Also, i think like `ids` is not a great name here, as `name` actually refer to a kv id, and `ids` here would be a chain of one or more attributes.
And why the maxsplit actually? Do we really want to set an attribute with dots in it? wouldn't it be better to iterate on the names, getting the next value until the penultimate one and use setattr on this one?

```python
wid, *attrs = key.split('.')
obj = self
for attr in attrs[:-1]:
    obj = getattr(obj, attr)
setattr(obj, attrs[-1], value)
```
or would that cause issues i can't think of right now?

kivy/uix/recycleview/views.py
matham: Done

kivy/uix/recycleview/views.py
matham: >why the star? the maxsplit makes it a 2-tuple, so we should be able to directly use the value without indexing on the next line.

Because the `maxsplit` is 1 we can have **at most** 2 items. But it could also be only 1. In the latter case, `ids` will be empty.

matham: >Also, i think like ids is not a great name here, as name actually refer to a kv id, and ids here would be a chain of one or more attributes.

I feel like that's nitpicking because in the normal case it is the property name, except when it's interpreted as an id :) But also, I think you may have misunderstood what this is doing. Specifically, with the following KV examples:

```
MyRule:
    street: ''
    Label:
        id: address

<RootRule@RecycleKVIDsDataViewBehavior>:
    country: ''
    Label:
        id: phone
    MyRule:
        id: my_rule
```

Setting `data` to `[{'country': 'US', 'phone.text': '123456', 'my_rule.street': 'Home ave', 'my_rule.address.text': '555'}]`. Under the current code, `my_rule` will have a attribute named `address.text` created and set to `555`. That's because we don't recursively go through rules like your example code. We only allow accessing the `ids` at the first level. Hence the only 1 split by period. Under your code, it would set the `text` property of the `address` widget to `555`.

The reason is that I don't think it would be common in kivy to set properties this way for sub rules and sub-sub rules etc. So I don't think we should complicate the API by adding this. Because once we add that it's hard to remove such functionality and I'd prefer to wait until this specifically is requested. And even then I think we can just add such recursive `ids` access a doc recipe. For now, this API only allows setting the property of a widget named in the root rule of the class.

tshirtman: ok, i did think a recursive definition could be useful, but i agree it might be a strong commitment that we might not need to do.

matham: I suppose we can raise an exception if there is more than one dot, in case users think that it does work recursively.

matham: Ok, I added that.

matham:
Yeah, I agree the message can be improved. But I couldn't think of a way to state it without writing a essay so I figured explaining it in the docs will have to be sufficient :)

chavarinbonbyn:
Which kivy version will contain this change ?

tshirtman:
next RC or release of 2.0.

PR #6904, opened 2020-05-30
Remove deprecation warning

sanderland:
As long as the kivy source code contains templates itself, and this hardcoded exception list is needed, there should be no exceptions logged.
In particular, making your own FileListEntry appears the only way to get around font issues in internationalization.

### DIFF
--- a/kivy/lang/parser.py
+++ b/kivy/lang/parser.py
@@ -348,14 +348,6 @@ class ParserRule(object):
 
     def _build_template(self):
         name = self.name
-        exception = ParserException(
-            self.ctx, self.line,
-            'Deprecated Kivy lang template syntax used "{}". Templates will '
-            'be removed in a future version'.format(name))
-        if name not in ('[FileListEntry@FloatLayout+TreeViewNode]',
-                        '[FileIconEntry@Widget]',
-                        '[AccordionItemTitle@Label]'):
-            Logger.warning(exception)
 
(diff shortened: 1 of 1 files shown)


inclement:
I disagree, the code is correctly warning you that your modifications will stop working at some point, even though they are the only direct way to achieve what you want via the file chooser api.

sanderland:
> I disagree, the code is correctly warning you that your modifications will stop working at some point, even though they are the only direct way to achieve what you want via the file chooser api.

I think as long as you need `if name not in <hardcoded list>` there should not be a warning.  The first step to deprecation should be for the core code itself to not rely on it, not to hack around it like this.

matham:
I agree with @inclement Kivy's core and widget base are two separate independent projects in a way. The code quality in core is much better than the widgets and in many ways we think of our widgets (except the core layouts etc) as demos. IMO we'd even separate kivy's widgets into its own sub-project if we could.

I'm saying this to point out that we made the decision for kivy's core to drop support for templates and that no one should be using it, hence the warning. Our filechooser widget needs to be completely re-written because to be frank, it's fairly crappy. So just because we haven't done that yet doesn't mean we should keep supporting templates, because that is not the correct decision for kivy's core.

The best workaround for you is to [suppress these warnings](https://docs.python.org/3/library/warnings.html#warnings.filterwarnings).

PR #6905, opened 2020-05-30
file chooser font property

sanderland:
This adds a `font_name` property to `FileChooserController`, which helps avoid characters in file names not being in the default font.
Probably there are other places in the files code this could be used as well.

### DIFF
--- a/kivy/uix/filechooser.py
+++ b/kivy/uix/filechooser.py
@@ -87,6 +87,8 @@ __all__ = ('FileChooserListView', 'FileChooserIconView',
 
 from weakref import ref
 from time import time
+
+from kivy.core.text import DEFAULT_FONT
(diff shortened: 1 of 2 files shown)


kivy/uix/filechooser.py
```diff
@@ -512,6 +514,16 @@ class FileChooserController(RelativeLayout):
     .. versionadded:: 1.8.0
     '''
 
+    font_name = StringProperty(DEFAULT_FONT)
```
matham, line 517: I think `font` is a better name to use as that's what we use elsewhere in kivy, I believe. @tshirtman ?

sanderland: Label uses font_name at least, but I wouldn't be surprised if there were inconsistencies in this.

Issue #6909, opened 2020-06-01
Hard crash in logger due to codec issues


sanderland:
**Software Versions**
* Python: 3.7
* OS: Windows 10
* Kivy: 2.0.0rc2
* Kivy installation method: pypi

**Describe the bug**
When setting logging level to trace, _write_message can crash with a UnicodeError

**Expected behavior**
Never hard crash while logging.

Probably there is some issue with picking up the codec, but logging binary strings with a catch Exception: something went wrong would be better than the current behaviour.

**To Reproduce**
Have a Korean version of windows 10 (which has the default cp949 codec) and a function call which returns a utf-8 string with special characters in a kv file.

**Code and Logs and screenshots**

```
 kivy.lang.builder.BuilderException: Parser: File "C:\Users\sande\Desktop\katrain\katrain\popups.kv", line 56:
 ...
      54:<I18NPopup>:
      55:    background_color: [*[c*255/66 for c in BACKGROUND_COLOR[:3]], 1]
 >>   56:    title: i18n._(self.title_key)
      57:    title_size: DESC_FONT_SIZE
      58:    title_font: self.font_name
 ...

 UnicodeEncodeError: 'cp949' codec can't encode character '\xe8' in position 29: illegal multibyte sequence
   File "C:\Users\sande\Anaconda3\lib\site-packages\kivy\lang\builder.py", line 699, in _apply_rule
     setattr(widget_set, key, value)
   File "kivy\weakproxy.pyx", line 35, in kivy.weakproxy.WeakProxy.__setattr__
   File "kivy\properties.pyx", line 497, in kivy.properties.Property.__set__
   File "kivy\properties.pyx", line 544, in kivy.properties.Property.set
   File "kivy\properties.pyx", line 599, in kivy.properties.Property.dispatch
   File "kivy\_event.pyx", line 1248, in kivy._event.EventObservers.dispatch
   File "kivy\_event.pyx", line 1130, in kivy._event.EventObservers._dispatch
   File "C:\Users\sande\Anaconda3\lib\site-packages\kivy\lang\builder.py", line 68, in call_fn
     trace('Lang: call_fn => value=%r' % (e_value, ))
   File "C:\Users\sande\Anaconda3\lib\logging\__init__.py", line 1449, in log
     self._log(level, msg, args, **kwargs)
   File "C:\Users\sande\Anaconda3\lib\logging\__init__.py", line 1519, in _log
     self.handle(record)
[...]

PR #6910, opened 2020-06-01
logger fix

sanderland:
Fix for #6909

Not sure if this can lead to infinite loops.
```
        except Exception as e:
            Logger.exception("Unexpected exception in logging", e)
```

### DIFF
--- a/kivy/logger.py
+++ b/kivy/logger.py
@@ -239,7 +239,6 @@ class FileHandler(logging.Handler):
         except Exception as e:
             Logger.exception("Unexpected exception in logging", e)
 
-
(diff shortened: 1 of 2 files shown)


tshirtman:
to be sure, could you write a simple test showing the issue on your machine, and check that the fixed code works with it? bonus points if that's a unittest we can run in CI, that would be most comforting :).

AndreMiras:
To me, the whole thing should actually be unit tested to show the point.
It shouldn't be that hard to add a test calling the _write_message() method showing that `Logger.exception()` is then getting called

PR #6918, opened 2020-06-03
Use ColorProperty instead of ListProperty for color property

pythonic64:
For all color properties except:
- `GestureSurface.color`
- `Label.outline_color`
- `Label.disabled_outline_color`

as for those properties format `rgb` is used and additional changes are needed to enable using of `ColorProperty`

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -1208,8 +1208,12 @@ class WindowBase(EventDispatcher):
 
     .. versionadded:: 1.10.1
 
-    :attr:`shape_color_key` is a :class:`~kivy.properties.ColorProperty`
+    :attr:`shape_color_key` is a :class:`~kivy.properties.ListProperty`
     instance and defaults to [1, 1, 1, 1].
+
+    .. versionchanged:: 2.0.0
+        Changed from :class:`~kivy.properties.ListProperty` to
+        :class:`~kivy.properties.ColorProperty`.
     '''
     def on_shape_color_key(self, instane, value):
         self._set_shape(
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -21,7 +21,8 @@ from kivy.base import EventLoop, stopTouchApp
 from kivy.modules import Modules
 from kivy.event import EventDispatcher
 from kivy.properties import ListProperty, ObjectProperty, AliasProperty, \
-    NumericProperty, OptionProperty, StringProperty, BooleanProperty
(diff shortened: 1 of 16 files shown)


pythonic64:
Should be merged after https://github.com/kivy/kivy/pull/6917 .

pythonic64:
Added `versionchanged` tag. 

Regression problem I can see is setting of individual components for color (with `__setitem__`) or setting of slice, but those are rare cases in my opinion.

matham:
>Regression problem I can see is setting of individual components for color (with __setitem__) or setting of slice,

How do you mean? `ColorProperty` still uses a list to represent the color, doesn't it?

pythonic64:
Setting like
```
obj.color[0] = 0.5
```
 won't dispatch event that `color` changed.

matham:
>won't dispatch event that color changed.

Oh, that's definitely bad. I wouldn't assume it's not a common usage, I'm pretty sure I used that with slicing. But even if it's not common, we can break it like that.

The solution is to change `ColorProperty` to use a `ObservableList` instead of a normal `list` https://github.com/kivy/kivy/blob/master/kivy/properties.pyx#L710.

pythonic64:
How can we ensure that iterable of 3 or 4 components are used when setting a slice? 
Example:
```python
obj.color[:] = [0.1, 0.2, 0.3, 1.0, 1.0]
```

`ColorProperty` should ensure that 4 components or 3 with 1.0 alpha by default.

matham:
>How can we ensure that iterable of 3 or 4 components are used when setting a slice?

The current implementation has the same problem though!? That would be user error and we don't have to protect against it. Trying to protect against that would make things much more complicated.

pythonic64:
Ok. I'll add a note to `ColorProperty` about slicing and that should cover it.

matham:
Cool. Although it's not just slicing. Presumably doing `obj.color.extend([1, 2, 3, 4, 5, 6])` would also be a problem.

pythonic64:
Added note under `versionchanged` tag because indexing and slicing are now supported. See https://github.com/kivy/kivy/pull/6930.

pythonic64:
@matham Are we waiting for another developer to approve or I can squash and merge?

matham:
Yeah, because it's such a large change I'd like it if someone like @tshirtman would also approve.

tshirtman:
looks good to me, as long the change to ColorProperty now means we can use it exactly like any ListProperty would be in this context, yes it changes quite a few pieces of the code, but the change every time is pretty trivial.

PR #6922, opened 2020-06-07
Cleanup WeakMethod

matham:
This cleans up WeakMethod a bit. Specifically

* Removes py2 code.
* `is_dead` called `dir`, which is super slow.
* Doing a comparison needlessly caught all exceptions.
* Added tests
* Replaced all the code that first checked whether the callback is dead and then called it. That pays the de-referencing cost twice. Instead, I changed it to simply call it once by de-referencing it and inspecting the result to see whether it is dead.
* Also caught a bug in context that did `for callback in self.observers:` instead of `for callback in self.observers[:]:`.

### DIFF
--- a/kivy/weakmethod.py
+++ b/kivy/weakmethod.py
@@ -1,4 +1,4 @@
-'''
+"""
 Weak Method
 ===========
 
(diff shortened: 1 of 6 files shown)


kivy/weakmethod.py
```diff
@@ -1,4 +1,4 @@
-'''
+"""
```
pythonic64, line 1: @matham Is Kivy standard to use single-quote strings instead of double-quote for docstring?

matham: It's not exactly a standard. That's just what some of the core-devs used. Personally, I used that too in my personal projects after seeing that in kivy. But pycharm complains about this, and looking at other projects double quotes are much more common.

That's why I consciously used double quotes here because I'd like to use that going forward. Although maybe that's not a good idea mixing different quoting styles.

tshirtman: black is a tool that pushes double quotes as the default everywhere it can, and the pep 257 also advises to use them for docstrings https://www.python.org/dev/peps/pep-0257/#specification

pythonic64: I would go with pep 257 too, but I think single-quote is dominant for docstrings in Kivy, so I would stick with single-quote until Kivy starts using black or someone does single to double change for all docstrings.

PR #6924, opened 2020-06-07
Buildozer is now in Beta.

Julian-O:
Documentation says Buildozer is in alpha.

It is currently in beta. ([Ref](https://pypi.org/project/buildozer/))

### DIFF
--- a/doc/sources/guide/packaging-android.rst
+++ b/doc/sources/guide/packaging-android.rst
@@ -36,7 +36,7 @@ downloads and sets up all the prequisites for python-for-android,
 including the android SDK and NDK, then builds an apk that can be
 automatically pushed to the device. 
 
-Buildozer currently works only in Linux, and is an alpha
+Buildozer currently works only in Linux, and is a beta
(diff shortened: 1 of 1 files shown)


matham:
That version tag may just be wrong because it started out in beta: https://github.com/kivy/buildozer/commit/d053ef783159e6d862c101f78f420896ed2bdb37.

Julian-O:
I understand your point.

So, now I don't know whether Buildozer or Kivy documentation is right, and where the issue should be raised. It seems to me the Buildozer's setup.py is a more authoritative source, but I don't know the project's history.

What do you suggest I do next?

matham:
I would ask on the dev channel on discord because there are buildozer devs there would know the true state of it, whether it should be considered as alpha or beta. Personally, I'm not familiar with it enough to know either way.

Julian-O:
Buildozer already has merged a batch updating their status to Production/Stable that should become visible in their next release. Should revisit this at that point.

PR #6926, opened 2020-06-08
Add clock lifecycle, better exception handling and other cleanup

matham:
This adds some features to clock as well as some docs cleanup.

The overall goal is to make Clock more reliable when interfacing with other threads. Specifically, to make sure that we don't end up in deadlock when the thread is waiting for something to be executed in the kivy thread, but the kivy thread has stopped because the app stopped and clock events are not executed anymore so the second thread gets stuck.

What it adds:

* A better defined clock lifecycle with a clock start and stop function that us called by the event loop.
  * If a event created with `create_lifecycle_aware_trigger` is triggered after the clock is stopped, a exception is raised because it would never be executed.
  * `create_lifecycle_aware_trigger` has an additional `clock_ended_callback` parameter. If the app exited normally, then either the `callback` or `clock_ended_callback` will have been called. The latter is called if the clock is stopped while the event is scheduled. 

    This allows e.g. a the other thread to be notified that the clock was stopped and the callback won't happen so that it can exit. That is, if triggering the event didn't raise a error, then we guarantee that either one of `callback` or `clock_ended_callback` will have been called by the time the clock stops.
 * A `release_ref` parameter that defaults to True was similarly added to `create_trigger` and `create_lifecycle_aware_trigger`. This allows more control, by preventing the callback from being garbage collected, if that's not desired.

    These ware only added to `create_trigger`/`create_lifecycle_aware_trigger` because it seemed better to leave the schedule_xxx methods simpler.

* Previously, an exception raised in a event's callback would cause the clock to crash and subsequent events may or may not be executed. Even if the exception was handled by the `ExceptionManager`, there was no guarantee that some scheduled events would not be skipped. This fixes it so that if a event's exception is handled by the `ExceptionManager`, other events will be shielded from the exception and will execute normally.
  We do it by adding `handle_exception` to clock that calls into the `ExceptionManager`.
* Fixed clock tests to use pytest and added more tests. Also fixed other tests that use the clock to use the new clock lifecycle.
* Adds better docs for the clock and for the `ExceptionManager`.

### DIFF
--- a/kivy/_clock.pxd
+++ b/kivy/_clock.pxd
@@ -96,11 +96,15 @@ cdef class CyClockBase(object):
     cdef public object _lock_release
 
     cdef public int has_ended
+    cdef public object _del_safe_lock
+    cdef public int _del_safe_done
 
     cpdef get_resolution(self)
-    cpdef create_trigger(
-        self, callback, timeout=*, interval=*, clock_ended_callback=*, release_ref=*)
-    cpdef schedule_del_safe(self, callback, clock_ended_callback=*)
+    cpdef create_lifecycle_aware_trigger(
+        self, callback, clock_ended_callback, timeout=*, interval=*, release_ref=*)
+    cpdef create_trigger(self, callback, timeout=*, interval=*, release_ref=*)
+    cpdef schedule_lifecycle_aware_del_safe(self, callback, clock_ended_callback)
+    cpdef schedule_del_safe(self, callback)
     cpdef schedule_once(self, callback, timeout=*)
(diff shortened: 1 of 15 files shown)


kivy/_clock.pxd
```diff
@@ -24,7 +24,19 @@ cdef class ClockEvent(object):
     cdef public double _dt
     cdef public list _del_queue
 
+    cdef public object clock_ended_callback
+    """A Optional callback for this event, which if provided is called by the clock
```
pythonic64, line 28: @matham More double-quote instead of single-quote docstring inconsistency.

PR #6931, opened 2020-06-09
Fix and improve vim syntax highlighting for kv lang

Gobbel2000:
Due to some errors and inconsistencies I've experienced with the vim syntax highlighting plugin I looked into that file and made the following changes. I tested it with vim 8.2.

* Fix back reference error (E65) in lines 21, 23, 24
* Reliably close kivyCanvas region when a widget block ends with a canvas block
* Be more generous to trailing whitespaces
* Highlight PreProc commands (`#:set ...`) below top-level

However there are still a few problems left that I can see where I'm not really sure what the best behavior would be or what the original intention was:

* Line 21: This assumes that attributes can have multiline expressions, which is not currently possible as far as I understand.
* Multiline expressions in kivyBind are currrently really wanky (you have to insert and delete a whitespace directly after `on_event:` to trigger it) as it clashes with line 25 for single line expressions. 
* Does line 23 add anything that's not covered by line 24?
* Also regarding kivyBind: the Function highlighting style doesn't work well with the python highlighting in longer expressions. 
* A _very_ minor issue: comments in a line after `#:set...` commands are still highlighted as kivyPreProc. Comments in the same line as `#:import...` fail the parser though, so maybe this should stay the way it is.

### DIFF
--- a/kivy/tools/highlight/kivy.vim
+++ b/kivy/tools/highlight/kivy.vim
@@ -11,29 +11,30 @@ syn include @pyth $VIMRUNTIME/syntax/python.vim
 
 syn match kivyComment       /#.*\n/ display contains=pythonTodo,Spell
 syn match kivyPreProc       /^\s*#:.*/
-syn match kivyAttribute     /\I\i*/ nextgroup=kivyValue
-syn match kivyBind          /on_\I\i*:/ nextgroup=kivyValue
-syn match kivyRule          /<-*\I\i*\%([,@]s*\I\i*\)*>:/
+syn match kivyAttribute     /\I\i*/ nextgroup=kivyValue skipwhite
+syn match kivyBind          /on_\I\i*:/
+syn match kivyRule          /<-*\I\i*\%([,@+]\I\i*\)*>:/
+syn match kivyRule          /\[-*\I\i*\%([,@+]\I\i*\)*]:/
 syn match kivyRootRule      /^\I\i*:\s*$/
-syn match kivyInstruction   /^\s\+\u\i*:/ nextgroup=kivyValue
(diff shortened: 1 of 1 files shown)


tshirtman:
> Reliably close kivyCanvas region when a widget block ends with a canvas block
omg, i though for *hours* against that, every solution i tried had problems, i'm going to test right away.

Gobbel2000:
I've added a few more changes, mainly regarding the issues mentioned about kivyBind. The body of kivyBind now only uses python highlighting and properly functions in oneline and multiline settings. 

The value/id differentiation used `\@<!` and `\@<=` for lookback which is slow so I changed that, and at least on a raspberry Pi the difference is very noticable. Otherwise this should behave like it did before. I also removed some lines that didn't add anything anymore as far as I could tell.

tshirtman:
image
it seems multiline expressions for properties got broken though. I think that's what you were refering to in
 
> Line 21: This assumes that attributes can have multiline expressions, which is not currently possible as far as I understand.

but I assure you multiline expressions like this are supported, so it would be nice to keep support for them :).

Aside that it looks really good.

Gobbel2000:
> it seems multiline expressions for properties got broken though

This should now work as well. I finally found the matchgroup option to highlight the start and end matches of a region differently. 

On a side note, I just now realized that vim has a very old kivy syntax file by default: https://github.com/vim/vim/blob/master/runtime/syntax/kivy.vim
It would be nice to update the file over there as well.

tshirtman:
Yes, i did try to propose this version in the past, the author of the previous version agreed, but i had to submit by mail to vim author, not PR, for some reason, i kind of forgot about it after that, it would be cool to do so, though.

PR #6942, opened 2020-06-14
don't crash when scrollview's content is the same size

tshirtman:
also remove test that prevents scrolling when size_hint is not None,
that's not necessary.

There is still something that prevent scrolling when the size is exactly
the size of content, but that's rare enough if you don't set size_hint
to 1 in the direction you are interested in.

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -286,7 +286,9 @@ class ScrollView(StencilView):
 
     :attr:`always_overscroll` is a
     :class:`~kivy.properties.BooleanProperty` and defaults to `True`.
-    versionadded:: 2.0.0
+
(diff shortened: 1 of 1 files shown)


kivy/uix/scrollview.py
```diff
@@ -286,7 +286,8 @@ def _set_do_scroll(self, value):
 
     :attr:`always_overscroll` is a
     :class:`~kivy.properties.BooleanProperty` and defaults to `True`.
-    versionadded:: 2.0.0
+
+    .. versionadded:: 2.0.0
```
matham, line 290: Don't you need a newline after this?

tshirtman: fixed

Issue #6946, opened 2020-06-16
Regular http requests via UrlRequest fail on Android


idevo89:
**Software Versions**
* Python: v3.8.1
* OS: Android 9 API29
* Kivy: v2.0.0rc2
* Kivy installation method: python-for-android develop

**Describe the bug**
Since #6703, regular http requests via urlrequest no longer work on android. The on_error function of urlrequest is called, while on desktop it is successful. Going back to kivy v2.0.0rc1 fixes it, but this version has broken https.

**Expected behavior**
Successful regular http requests.

**To Reproduce**
```python
from kivy.app import runTouchApp
from kivy.uix.widget import Widget
from kivy.network.urlrequest import UrlRequest


class Test(Widget):
    def __init__(self, **kwargs):
        super(Test, self).__init__(**kwargs)
        UrlRequest(
            url='http://lipsum.com/',
            on_error=lambda *args: print('error:', args),
            on_failure=lambda *args: print('fail:', args),
            on_redirect=lambda *args: print('redir:', args),
            on_success=lambda *args: print('success:', args)
        )
        UrlRequest(
            url='https://lipsum.com/',
            on_error=lambda *args: print('error:', args),
            on_failure=lambda *args: print('fail:', args),
            on_redirect=lambda *args: print('redir:', args),
            on_success=lambda *args: print('success:', args)
        )


runTouchApp(Test())

```

**Code and Logs and screenshots**
```python
TypeError("__init__() got an unexpected keyword argument 'context'")
```

AndreMiras:
Thanks for the report. Please share the full stacktrace as it often helps developers to spot the exact issue more quickly.
Also could you try on the  2.0.0rc3 that was released recently?

idevo89:
It is the same for v2.0.0rc3. I omitted the try-exception around
```python
result, resp = self._fetch_url(url, req_body, req_headers, q)
```
and added 'win' to line 219 in urlrequest to simulate the error on desktop and get a traceback.
https://github.com/kivy/kivy/blob/ab58fbecea6ae1ee4f8f465d9eaad2b41bdd6f65/kivy/network/urlrequest.py#L219


```python
Exception in thread Thread-1:
 Traceback (most recent call last):
   File "C:\Program Files (x86)\Python38-32\lib\threading.py", line 932, in _bootstrap_inner
     self.run()
   File "C:\Program Files (x86)\Python38-32\lib\site-packages\kivy\network\urlrequest.py", line 253, in run
[...]

PR #6951, opened 2020-06-17
:recycle: Refactors test_urlrequest.py

AndreMiras:
- uses @pytest.mark.skipif() rather than return
- couple of DRY fixes using helper functions

### DIFF
--- a/kivy/tests/test_urlrequest.py
+++ b/kivy/tests/test_urlrequest.py
@@ -5,17 +5,46 @@ UrlRequest tests
 import pytest
 import threading
 
+from datetime import datetime
 from time import sleep
 from base64 import b64encode
 import os
 
 
-class UrlRequestQueue(object):
+def wait_request_is_finished(kivy_clock, request, timeout=10):
+    start_time = datetime.now()
+    timed_out = False
+    while not request.is_finished and not timed_out:
+        kivy_clock.tick()
+        sleep(.1)
+        timed_out = (datetime.now() - start_time).total_seconds() > timeout
+    assert request.is_finished
 
-    queue = []
+
+def ensure_called_from_thread(queue):
+    """Ensures the callback is called from this thread (main)."""
+    tid = threading.get_ident()
+    assert queue[0][0] == tid
+    assert queue[-2][0] == tid
+    assert queue[-1][0] == tid
+
+
+def check_queue_values(queue):
+    """Helper function verifying the queue contains the expected values."""
+    # we should have 2 progress minimum and one success
+    assert len(queue) >= 3
+
+    assert queue[0][1] == 'progress'
+    assert queue[-2][1] == 'progress'
+    assert queue[-1][1] in ('success', 'redirect')
+
+    assert queue[0][2][0] == 0
+    assert queue[-2][2][0] == queue[-2][2][1]
+
+
+class UrlRequestQueue:
 
     def __init__(self, queue):
-        super(UrlRequestQueue, self).__init__()
         self.queue = queue
 
     def _on_success(self, req, *args):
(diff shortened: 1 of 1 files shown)


kivy/tests/test_urlrequest.py
```diff
@@ -5,17 +5,50 @@
...
+    assert queue[-1][1] in ('success', 'redirect')
+
+    assert queue[0][2][0] == 0
+    assert queue[-2][2][0] == queue[-2][2][1]
+
+
+class UrlRequestQueue:
 
     def __init__(self, queue):
-        super(UrlRequestQueue, self).__init__()
```
AndreMiras, line 18: Removed because we don't inherit from anything

kivy/tests/test_urlrequest.py
```diff
@@ -5,17 +5,50 @@
...
+def ensure_called_from_thread(queue):
+    """Ensures the callback is called from this thread (main)."""
+    tid = threading.get_ident()
+    assert queue[0][0] == tid
+    assert queue[-2][0] == tid
+    assert queue[-1][0] == tid
 
-class UrlRequestQueue(object):
 
-    queue = []
```
AndreMiras, line 15: The `queue` is already defined in the init, less is more. Having it defined here can lead to the misleading assumption we want it to be static attribute

kivy/tests/test_urlrequest.py
```diff
@@ -5,17 +5,50 @@
...
+    from kivy.network.urlrequest import UrlRequest
+    from kivy.clock import Clock
+
+
+def wait_request_is_finished(request, timeout=10):
+    start_time = datetime.now()
+    timed_out = False
+    while not request.is_finished and not timed_out:
+        Clock.tick()
+        sleep(.1)
```
AndreMiras, line 23: reduced the sleep time from .5 to .1 because nobody likes to wait :)

AndreMiras:
A small test refactoring before addressing #6946, also refs https://discord.com/channels/423249981340778496/490493814281338890/722883388889104446

kivy/tests/test_urlrequest.py
```diff
@@ -5,17 +5,50 @@
...
 import threading
 
+from datetime import datetime
 from time import sleep
 from base64 import b64encode
 import os
 
+if not os.environ.get('NONETWORK'):
+    from kivy.network.urlrequest import UrlRequest
+    from kivy.clock import Clock
```
matham, line 15: Can you use the `kivy_clock` fixture and pass it in, instead of importing it? 

Also, I don't like when we import kivy things globally in tests because then the kivy stuff is initialized before the tests start because kivy imports are so aggressive. It'd be better if you just do `from kivy.network.urlrequest import UrlRequest` in each test.

AndreMiras: I've updated the code as per your suggestion.
Out of curiosity, regarding kivy imports being aggressive. Dont't we have a deeper problem we want to fix one day? Or is it a feature that can't be fixed or that we don't want to fix?

AndreMiras:
Addressed all the comments, including:
> What's up with the :recycle: in the title?

It's commit emoji, https://gitmoji.carloscuesta.me/ I assume you didn't like it, so I removed it :smile:

kivy/tests/test_urlrequest.py
matham: Yeah, we do have a very deep problem there that I hope will be fixed one day (the sooner the better). So this is just in preparation for that bright sunny morning :D

matham:
Oh, it didn't render as a ♻️ in the title so I was confused by what it was supposed to be.

AndreMiras:
Yes, BTW, I just see the merge message picked the GitHub title rather than the actual commit title. So now we have a merge commit with the icon :man_facepalming:

PR #6965, opened 2020-06-24
Allow assignment of color names as values for ColorProperty

pythonic64:


kivy/properties.pyx
```diff
@@ -1983,11 +1983,18 @@ cdef class ConfigParserProperty(Property):
 cdef class ColorProperty(Property):
     '''Property that represents a color. The assignment can take either:
 
-    - a list or tuple of 3 to 4 float values between 0-1 (kivy default)
+    - a collection of 3 or 4 float values between 0-1 (kivy default)
     - a string in the format #rrggbb or #rrggbbaa
+    - a string representing color name (eg. 'red', 'yellow', 'green')
+
+    Color names definitions can be found at this
+    `link <https://www.w3.org/TR/SVG11/types.html#ColorKeywords>`_. Color can
```
matham, line 1991: It'd be nice to also link to `kivy.utils.colormap` from here.

kivy/properties.pyx
```diff
@@ -1996,34 +2003,38 @@ cdef class ColorProperty(Property):
...
         with float values between 0-1.
+        Assingning color name as value is now supported.
+        Value `None` is allowed as default value for property.
     '''
 
-    def __init__(self, defaultvalue=None, **kw):
-        defaultvalue = defaultvalue or [1, 1, 1, 1]
+    def __init__(self, defaultvalue=0, **kw):
+        defaultvalue = \
+            [1.0, 1.0, 1.0, 1.0] if defaultvalue == 0 else defaultvalue
```
matham, line 2012: Why support None? Why not just `if defaultvalue`?

kivy/properties.pyx
```diff
@@ -1996,34 +2003,38 @@ cdef class ColorProperty(Property):
...
-        if isinstance(x, string_types):
-            return ObservableList(self, obj, self.parse_str(obj, x))
-
+        cdef object color = x
         try:
-            count = len(x)
-        except TypeError as e:
+            if isinstance(x, string_types):
+                color = self.parse_list(obj, self.parse_str(obj, x))
+            color = self.parse_list(obj, color)
```
matham, line 2022: You're parsing list twice if it's a string.

kivy/properties.pyx
```diff
@@ -1996,34 +2003,38 @@ cdef class ColorProperty(Property):
...
+        except Exception as e:
```
matham, line 2023: I don't think we should be catching so broadly as `Exception`. Specifically, `parse_str`, and more deeply `get_color_from_hex` don't raise any explicit exceptions. So we shouldn't try handling any implicit exceptions it may raise, even if we immediately re-raise, because generally if you don't know the reason for the exception it's best to not catch it and change the exception class.

So I would specifically only put the try/except around `parse_list` and only catch `ValueError`.

kivy/properties.pyx
pythonic64: Right, for sting case it should be ```color = self.parse_str(obj, x)``` and `color` will be a `list` and `parse_list` then can return `ObservableList`.

kivy/properties.pyx
pythonic64: To skip empty collections like it is in `ListProperty` and to support `None` as default value (initial value) in case that user wants to use `allownone=True`.

kivy/properties.pyx
pythonic64: I'll add:

"...found at his `link` and at :obj:`kivy.utils.colormap`."

kivy/properties.pyx
pythonic64: If we have `some_color = ColorProperty('matham')`, then with `Exception` we get this traceback:
```
   File "kivy\properties.pyx", line 2021, in kivy.properties.ColorProperty.convert
   File "kivy\properties.pyx", line 2032, in kivy.properties.ColorProperty.parse_str
   File "D:\Work\Kivy Framework\kivy\kivy\utils.py", line 121, in get_color_from_hex
     for x in split('([0-9a-f]{2})', s.lower()) if x != '']
   File "D:\Work\Kivy Framework\kivy\kivy\utils.py", line 121, in <listcomp>
     for x in split('([0-9a-f]{2})', s.lower()) if x != '']
 ValueError: invalid literal for int() with base 16: 'matham'
 
 The above exception was the direct cause of the following exception:
 
   File "D:\Work\Kivy Framework\kivy\dev\color_app.py", line 43, in <module>
     ColorApp().run()
   File "kivy\_event.pyx", line 190, in kivy._event.EventDispatcher.__cinit__
   File "kivy\properties.pyx", line 448, in kivy.properties.Property.link
   File "kivy\properties.pyx", line 419, in kivy.properties.Property.init_storage
   File "kivy\properties.pyx", line 2024, in kivy.properties.ColorProperty.convert
 ValueError: ColorApp.some_color has an invalid format (got 'matham')
```

and with try/except only around `parse_list` with `ValueError` exception we get this:
```
   File "D:\Work\Kivy Framework\kivy\dev\color_app.py", line 43, in <module>
     ColorApp().run()
   File "kivy\_event.pyx", line 190, in kivy._event.EventDispatcher.__cinit__
   File "kivy\properties.pyx", line 448, in kivy.properties.Property.link
   File "kivy\properties.pyx", line 419, in kivy.properties.Property.init_storage
   File "kivy\properties.pyx", line 2020, in kivy.properties.ColorProperty.convert
   File "kivy\properties.pyx", line 2032, in kivy.properties.ColorProperty.parse_str
   File "D:\Work\Kivy Framework\kivy\kivy\utils.py", line 121, in get_color_from_hex
     for x in split('([0-9a-f]{2})', s.lower()) if x != '']
   File "D:\Work\Kivy Framework\kivy\kivy\utils.py", line 121, in <listcomp>
     for x in split('([0-9a-f]{2})', s.lower()) if x != '']
 ValueError: invalid literal for int() with base 16: 'matham'
```

The former gives more clear information to user at the end of traceback, such as class name, property name and that format of assigned color is invalid which helps with debug.

I suggest just to replace `Exception` with `ValueError` and keep sting `parse_str` inside of try/except block.

pythonic64: Actually, we should use `(ValueError, TypeError)` to cover case when new value is not `Sized`.

PR #6969, opened 2020-06-28
fix docs for on_dropfile

sanderland:
particularly this being bytes was surprising


kivy/core/window/__init__.py
```diff
@@ -307,7 +307,7 @@ class WindowBase(EventDispatcher):
                 The *unicode* parameter has be deprecated in favor of
                 codepoint, and will be removed completely in future versions.
 
-        `on_dropfile`: str
+        `on_dropfile`: filename (bytes):
```
matham, line 310: I think this could potentially be either a string or bytes, depending on the OS and window backend. So maybe say it could be either bytes or string?

PR #6976, opened 2020-07-04
Prevent modalview dismissal without on_touch_down

Zen-CODE:
There is currently a bug is master due to recent changes in the ModalView code.
Full details with a runnable examples are here: https://github.com/kivy/kivy-ios/issues/499

### DIFF
--- a/kivy/uix/modalview.py
+++ b/kivy/uix/modalview.py
@@ -277,7 +277,8 @@ class ModalView(AnchorLayout):
         return True
 
     def on_touch_up(self, touch):
-        if self.auto_dismiss and not self._touch_started_inside:
+        # Explicitly test for False as None occurs when shown by on_touch_down
+        if self.auto_dismiss and self._touch_started_inside is False:
             self.dismiss()
             return True
         super(ModalView, self).on_touch_up(touch)
--- a/kivy/uix/modalview.py
+++ b/kivy/uix/modalview.py
@@ -280,8 +280,9 @@ class ModalView(AnchorLayout):
         # Explicitly test for False as None occurs when shown by on_touch_down
         if self.auto_dismiss and self._touch_started_inside is False:
             self.dismiss()
-            return True
-        super(ModalView, self).on_touch_up(touch)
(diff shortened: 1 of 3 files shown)


kivy/uix/modalview.py
```diff
@@ -277,7 +277,8 @@ def on_touch_move(self, touch):
         return True
 
     def on_touch_up(self, touch):
-        if self.auto_dismiss and not self._touch_started_inside:
+        # Explicitly test for False as None occurs when shown by on_touch_down
+        if self.auto_dismiss and self._touch_started_inside is False:
             self.dismiss()
             return True
         super(ModalView, self).on_touch_up(touch)
```
matham, line 284:
```suggestion
        super(ModalView, self).on_touch_up(touch)
        self._touch_started_inside = None
```

kivy/uix/modalview.py
```diff
@@ -277,7 +277,8 @@ def on_touch_move(self, touch):
         return True
 
     def on_touch_up(self, touch):
-        if self.auto_dismiss and not self._touch_started_inside:
+        # Explicitly test for False as None occurs when shown by on_touch_down
+        if self.auto_dismiss and self._touch_started_inside is False:
             self.dismiss()
```
matham, line 282:
```suggestion
            self.dismiss()
            self._touch_started_inside = None
```

kivy/uix/modalview.py
Zen-CODE: Funnily enough, that occurred to me, but then I saw this comment in the code: that a modal view can only be opened once? This is does not seem to be the case, as this example proves?

https://github.com/kivy/kivy/blob/0df47fc686c66397931a97d839a929c0338b4fbf/kivy/uix/modalview.py#L214

Zen-CODE: Ah! It has to be dismissed before it can be shown again :-) My bad...

Zen-CODE:
@matham That should do it? Thanks

PR #6984, opened 2020-07-10
Accept numeric instances in NumericProperty

### LINKED ISSUE #6982: Setting a NumericProperty to a numpy float gives an error
**Software Versions**
* Python: 3.7
* OS: Windows
* Kivy: 2.0.0rc3
* Kivy installation method: pypi

**Describe the bug**
When setting a NumericProperty to (e.g.) a numpy.float64, the result is a ValueError:
```
ERROR: Exception ValueError: 'somevar has an invalid format (got -0.555499255657196)'
```
This is a very confusing error, as the format seems ok, and numpy float types are typically compatible with python floats.

**Expected behavior**
Either accept such floating point types, or give a TypeError

**To Reproduce**
```python
import numpy as np


class A(Widget):
    p = NumericProperty(None)

arr = np.arange(1,10,0.1)
a = A()
a.p = arr[1]
```

Franccisco:
NumericProperty class only accepts int, float, long, tuple and list
types. Because of that, numeric instances like numpy's float64 raise a
ValueError exception. For example, this code will crash (it is
simplified to focus on the error only):

```py
import numpy as np


class CrashWidget(Widget):
    array_of_numbers = np.array([1.0])
    numeric_value = NumericProperty(array_of_numbers[0])


if __name__ == "__main__":
    CrashWidget()
```

For the NumericProperty class, array_of_numbers[0] is not a valid type
because it is not one of int, float, long, tuple or list. However, it is
an instance of float (`isinstance(array_of_numbers[0], float)` is True).
That said, the proposed solution is to check for the instance of the
accepted NumericProperty types instead of checking for specific types
only.

Fixes #6982

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -635,18 +635,16 @@ cdef class NumericProperty(Property):
     cdef check(self, EventDispatcher obj, value):
         if Property.check(self, obj, value):
             return True
-        if type(value) not in (int, float, long):
-            raise ValueError('%s.%s accept only int/float/long (got %r)' % (
+        if not isinstance(value, (int, float, long)):
+            raise ValueError('%s.%s accept only instances of int/float/long'
+                             '(got %r)' % (
                 obj.__class__.__name__,
                 self.name, value))
 
     cdef convert(self, EventDispatcher obj, x):
-        if x is None:
+        if x is None or isinstance(x, (int, float, long)):
             return x
-        tp = type(x)
-        if tp is int or tp is float or tp is long:
-            return x
-        if tp is tuple or tp is list:
+        if isinstance(x, (tuple, list)):
             if len(x) != 2:
                 raise ValueError('%s.%s must have 2 components (got %r)' % (
                     obj.__class__.__name__,


kivy/properties.pyx
```diff
@@ -635,18 +635,16 @@ cdef class NumericProperty(Property):
     cdef check(self, EventDispatcher obj, value):
         if Property.check(self, obj, value):
             return True
-        if type(value) not in (int, float, long):
-            raise ValueError('%s.%s accept only int/float/long (got %r)' % (
+        if not isinstance(value, (int, float, long)):
```
pythonic64, line 638: You should use `isinstance(value, Number)` for number check (https://docs.python.org/3.7/library/numbers.html).

pythonic64: Just seeing that `isinstance(value, Number)` would include support for `complex` numbers, but that should be supported because it's a `NumericProperty`?

kivy/properties.pyx
```diff
@@ -635,18 +635,16 @@ cdef class NumericProperty(Property):
...
 
     cdef convert(self, EventDispatcher obj, x):
-        if x is None:
+        if x is None or isinstance(x, (int, float, long)):
             return x
-        tp = type(x)
-        if tp is int or tp is float or tp is long:
-            return x
-        if tp is tuple or tp is list:
+        if isinstance(x, (tuple, list)):
```
pythonic64, line 647: I would not test for `tuple`/`list` as it can be any collection of size two, but write to test for string value first with single `raise` statement like in `ColorProperty` https://github.com/kivy/kivy/blob/master/kivy/properties.pyx#L2020 .

kivy/properties.pyx
Franccisco: > You should use `isinstance(value, Number)` for number check (https://docs.python.org/3.7/library/numbers.html).

I agree with this one. Thanks!

> Just seeing that `isinstance(value, Number)` would include support for `complex` numbers, but that should be supported because it's a `NumericProperty`?

Exactly, if we also include support for complex numbers via the `Number`
class, it will also accept complex numbers included in a numpy's array.

kivy/properties.pyx
Franccisco: > I would not test for `tuple`/`list` as it can be any collection of size two, but write to test for string value first with single `raise` statement like in `ColorProperty` https://github.com/kivy/kivy/blob/master/kivy/properties.pyx#L2020 .

I get your point, but I would not check for `string_types`, because we
already dropped the Python 2 support.

These changes are not directly related to the pull request title and
objective, so I will rename it and make some changes.

Thanks for the review, I will let you know when it is ready.

kivy/properties.pyx
matham: I don't completely agree about using `isinstance(value, Number)`. 

The goal of `NumericProperty` is to accept numbers specifically to be used by widgets as part of their positioning and sizing. It's not to accept arbitrary numbers, because arbitrary numbers cannot be used to set a graphics objects position in opengl. You can see it by the fact that it accepts strings e.g. `'5dp'` and that this kind of issue rarely comes up for users or they find good workarounds. 

So we should just accept numbers that can be used for that kind of purpose. This e.g. precludes complex numbers. But, it is reasonable to accept e.g. numpy floats or ints. If someone does want to use complex numbers for their own purpose, they should just use `ObjectProperty`.

Ultimately, if we accept numpy numbers, they'll have to be converted to normal ints and floats in the convert method, because that's the only kind of number `NumericProperty` users typically know how to use. I'd say it's probably best to keep the current logic of `if type(value) not in (int, float, long)` as a quick path for most numbers and same for strings or lists/tuples (maybe with a try/except for iterables as mentioned below). And then if it all fails try to check explicitly `isinstance(x, int)` and `float` and then convert them to the appropriate type.

pythonic64: Test `isinstance(value, Number)` won't break `NumericProperty` for current use cases but expand it. I don't think that @Franccisco or @sanderland want to use numpy floats to assign values for pos/size/color (which are used by graphics) but to have property which check/accept all number values and dispatch event when value gets changed through assign or in-place operations (not supported by `ObjectProperty`). 

Documentation of `NumericProperty` should state that `float` type must be used if property is used to update attributes of graphics instructions, but otherwise all number types are accepted. 

If I remember correctly type `long` is removed from Python3 and `int` should be used in its place. Don't know if the same is valid for Cython.

sanderland: The usecase was something like, there is a widget displaying a numeric value (say, win rate). The .kv handles setting the .text given the NumericProperty value. The value comes from some machine learning model.
It would have been fine to raise a TypeError and tell me to cast it to float (although accepting the numpy float would be nicer, of course). The thing that took quite a bit of time was the error about an 'invalid format'.

matham: >Test isinstance(value, Number) won't break NumericProperty for current use cases but expand it.

True, but I don't think we should expand it. `NumericProperty` should be limited to things that can be and should be converted to either a `float` or `int`.

>and dispatch event when value gets changed through assign or in-place operations (not supported by ObjectProperty).

Right, object property doesn't support in-place assign and neither does numeric property?

>Documentation of NumericProperty should state that float type must be used if property is used to update attributes of graphics instructions, but otherwise all number types are accepted.

I very much disagree. Currently, we specifically only accept floats or ints (and longs). Because `NumericProperty` is intended specifically for widgets where we need a float/int for layouts or graphics. If you give it anything else it'll crash currently, and that's how it should be. There's really no reason to have specific string and numeric properties, beyond a basic object property, except for the reason that this is needed for setting widget layouts or similar properties.

We need to keep our API as simple as minimal as possible, until there's demonstarted need to accept more number types. And so far I haven't seen a reason given for expanding numeric property beyond floats/int or things that can be converted to floats/ints.

But the docs should clearly say this that it'll be converted to a float or int.

>long is removed from Python3

Yeah, we should remove it.

matham: >The usecase was something like, there is a widget displaying a numeric value (say, win rate). The .kv handles setting the .text given the NumericProperty value. The value comes from some machine learning model.

Why can't you use just a object property? Or maybe convert it to a float before setting it. Or even a string property since you'll be converting it to a string. Either way, there's no benefit in storing it as a numeric property. I think the `NumericProperty` is a tempting distraction in this case, because it's intended for places where you need float/ints as mentioned, and it's mainly useful for internal kivy stuff to convert from e.g. `'5dp'` to a number. For normal stuff I'd highly recommend to use object property instead of string/numeric property.

Perhaps we need to say this in the docs.

Franccisco: > I don't completely agree about using isinstance(value, Number).
>
> The goal of NumericProperty is to accept numbers specifically to be used by widgets as part of their positioning and sizing. It's not to accept arbitrary numbers, because arbitrary numbers cannot be used to set a graphics objects position in opengl. You can see it by the fact that it accepts strings e.g. '5dp' and that this kind of issue rarely comes up for users or they find good workarounds.
>
> So we should just accept numbers that can be used for that kind of purpose. This e.g. precludes complex numbers. But, it is reasonable to accept e.g. numpy floats or ints. If someone does want to use complex numbers for their own purpose, they should just use ObjectProperty.

The reason of this Pull Request is because it is more reasonable to use
something like a numpy's float/int/complex in a NumericProperty class,
rather than ObjectProperty. To be honest, I did not know that the
purpose of NumericProperty is related to be used by widgets, and nothing
else.

> Ultimately, if we accept numpy numbers, they'll have to be converted to normal ints and floats in the convert method, because that's the only kind of number NumericProperty users typically know how to use. I'd say it's probably best to keep the current logic of if type(value) not in (int, float, long) as a quick path for most numbers and same for strings or lists/tuples (maybe with a try/except for iterables as mentioned below). And then if it all fails try to check explicitly isinstance(x, int) and float and then convert them to the appropriate type.

I don't agree with converting numpy numbers to normal ints or floats,
because it is an additional task that would go against the purpose of
using numpy numbers (performance and speed).

Franccisco: > Test isinstance(value, Number) won't break NumericProperty for current use cases but expand it. I don't think that @Franccisco or @sanderland want to use numpy floats to assign values for pos/size/color (which are used by graphics) but to have property which check/accept all number values and dispatch event when value gets changed through assign or in-place operations (not supported by ObjectProperty).

Exactly. As I said, the idea is to expand NumericProperty. After all,
when I want to use something like a numpy number, the ObjectProperty
class is not the first thing that comes to my mind.

> Documentation of NumericProperty should state that float type must be used if property is used to update attributes of graphics instructions, but otherwise all number types are accepted.

I agree. After all, if you use something that is not a
float/int/list/tuple, you get an error that is not so clear.

> If I remember correctly type long is removed from Python3 and int should be used in its place. Don't know if the same is valid for Cython.

I did not remove it because I thought it was there for a good reason.
As it is deprecated, it should be removed.

kivy/properties.pyx
Franccisco: > > The usecase was something like, there is a widget displaying a numeric value (say, win rate). The .kv handles setting the .text given the NumericProperty value. The value comes from some machine learning model.
> 
> Why can't you use just a object property? Or maybe convert it to a float before setting it. Or even a string property since you'll be converting it to a string. Either way, there's no benefit in storing it as a numeric property. I think the `NumericProperty` is a tempting distraction in this case, because it's intended for places where you need float/ints as mentioned, and it's mainly useful for internal kivy stuff to convert from e.g. `'5dp'` to a number. For normal stuff I'd highly recommend to use object property instead of string/numeric property.
> 
> Perhaps we need to say this in the docs.
> 
> > It would have been fine to raise a TypeError and tell me to cast it to float (although accepting the numpy float would be nicer, of course). The thing that took quite a bit of time was the error about an 'invalid format'.
> 
> Yes, the error should be clearer. And we should ideally convert anything convertible to a float/int to float/int and only then raise an error if it doesn't work.

I get your point about using an ObjectProperty for this use case
@matham, but it is not clear at first sight. When you are using a
numeric type, the first thing that comes to mind is to use a
NumericProperty instead of an ObjectProperty.

So, to summarize (and solve this):

- We should add these use cases in the docs to let people know when they
  should use ObjectProperty instead of NumericProperty.

- The TypeError must be more explicit. Ideally, it should suggest the
  use of ObjectProperty.

- `long` must be removed as it is deprecated.

As I mentioned before, I do not think we should cast the numpy numbers
to float/ints because it is not the idea behind using these. They might
be useful for mathematical computation without losing performance and
speed.

matham: I'll try to make the case for converting to float/int. As I mentioned, and you seem to kinda agree now(?) that `NumericProperty` is not a general numeric property but intended for widget sizing/pos etc. 

I kinda said this before, but it's even stronger than that. Its primary purpose seemingly is meant to take the input and convert it to something that can be used by the layout/graphics system. Hence why it converts strings to numbers using `dpi`. This is why we do all the type checking to make sure we only end up with float/int currently. Hence why I really think that if we were to broaden the input, they still will need to output either a float or int, otherwise it may end up breaking the layout/graphics system because you'd pass it a numpy number that is not understood by gl.

It is unfortunate that we named it generically `NumericProperty`, as it does make it seem like it's accepting a generic number and I'll admit to using it wherever I use numbers in the past. But that's not intended really the case I think. Actually, even the performance point is contrary to your argument. That's because `NumericProperty` does all this additional type checking, which `ObjectProperty` does not. So if you were worried about performance of setting the property, I'd definitely go with a `ObjectProperty` which doesn't do any checking (except for equality).

That said, I'm open to changing my mind about this, but would need a stronger argument of actual use cases for not converting. And would like to hear if e.g. @tshirtman thinks differently as he has much more experience with this. Either way, I agree about the docs having to be much clearer about intended usage and your other bullet points.

kivy/properties.pyx
pythonic64: I checked and both `NumericProperty` and `ObjectProperty` will dispatch events for in-place operations. 

Kivy doesn't completely safeguard graphics attributes when assigning from properties as user can have a `ListProperty` of `numpy.float64` (or values any other type) and try to assign it to:
- pos/size of `Rectangle` instruction
- vertices of `Mesh` instruction
- rgba of `Color` instruction

and those values will get pass to graphics without type check and invalid types will raise errors on graphics side.

Also, element-wise changes are possible like:
`label.color[0] = <instance not float or int>`
as `ColorProperty` nor `ObservableList` do not check for types of each element and Kivy still expects from user to know which types are allowed for graphics. Currently graphics instructions will cast `numpy.float64` to float, so cases like this:
`Window.clearcolor[0] = np.float64(0.5)`
do work. I didn't test but I assume that pos/size of `Rectangle` will also cast to `float` if `numpy.float64` values are assigned. 

Because of its naming and its documentation stating: "Property that represents a numeric value.", user will expect that this property can be used for any number for bind/dispatch feature. Expanding it doesn't break its current usage and it give user full support for numbers which are stored without cast except for strings which are cast to `float`. 

From my view it's up to graphics instructions to state in documentation (some classes already do) which types are compatible/required, like any value which can be cast to `float` is valid and not up to `NumericProperty` to ensure `float/int`.

pythonic64: Difference in `NumericProperty` vs `ObjectProperty` is the `convert` method which for `ObjectProperty` returns value unchanged and for expanded `NumericProperty` we would have: 
```
if isinstance(x, Number) or x is None:
    return x
# Next lines would check if `x` is a str first and then collection and return `float`
```
as first two lines in `convert` method. Method `check` is similar as `NumericProperty` checks for `Number` or `str` and `ObjectProperty` checks for `baseclass`, both on top of `Property.check` method.

Since for most common use case, assigning a number value, `NumericProperty` does what it name implies, but `ObjectProperty` is slightly faster. 

Someone can argue that users can use:
`number = ObjectProperty(0, baseclass=numpy.float64)`
or
`number = ObjectProperty(0, baseclass=Number)`
but how many users know about `baseclass` argument? @Franccisco @sanderland

Even with `ObjectProperty` used to store number, user can still pass that number to graphics instruction and if it can be cast to float (or any other type which instruction accepts) it still works as extended `NumericProperty` and makes pointless to accept only `float/int`.

matham: I've responded to this in the other thread below, but to summarize the goal of `NumericProperty` is not to accept any number, but to accept float/int or things that can be converted to float/int. I don't think we should expand the API beyond that, except to try to convert to int/float if we can, otherwise raise an error.

>Kivy doesn't completely safeguard graphics attributes when assigning from properties as user can have a ListProperty of numpy.float64 (or values any other type) and try to assign it to

True that not all of the properties guard for this, but still that's unrelated to the design of `NumericProperty` whose goal in kivy is currently float/int to be used for pos etc.

>Even with ObjectProperty used to store number, user can still pass that number to graphics instruction and if it can be cast to float (or any other type which instruction accepts) it still works as extended NumericProperty and makes pointless to accept only float/int

We don't do explicit casting in cython because it is expected only ints or floats will be passed to graphics. It is the user's responsibility to ensure that. In fact we have special code in `blit_buffer` to be able accept numpy arrays because it is not normally accepted. And I don't think forcing conversions to float/int in each graphics instruction is a good idea. So this is the job of `NumericProperty`, or the user if they are using a different property.

matham:
I haven't seen any use cases for accepting anything beyond what can be converted to ints/floats and I don't think that's part of common Kivy patterns, considering this issue rarely comes up for numpy types let alone other numeric types. And as I explained `NumericProperty`, while named generally is actually used with for int/float, the basic numeric type in python and understood by our graphics and layout system.

@Franccisco please update the PR as I suggested to state in the docs that it accepts numbers that can be converted to int/float, and in the code check if it's int/float/str and otherwise try to convert to float and re-raise an error if that fails.

Franccisco:
> @Franccisco please update the PR as I suggested to state in the docs that it accepts numbers that can be converted to int/float, and in the code check if it's int/float/str and otherwise try to convert to float and re-raise an error if that fails.

Ok, I'll do it tomorrow.

kivy/properties.pyx
tshirtman: Sorry for not answering earlier.
The consideration here is wide, do we want the kivy property system to be useful outside of kivy's internal management? It seems we use it for other things when we do kivy apps, when the technical constraints it imposes allow it, so there is a case for it. But do we want to relax some of these constraints to allow more use cases? It seems like a positive thing at first, but on the other hand, it will add more complexity to the codebase, as we will have to care about cases that didn't exist before, and we must check the failure modes when these new usages are wrongly applied to the main kivy usage, e.g, what happens if someone decides to pass a complex number as the width of a widget? Surely there is an error, but is it an acceptable one? Another consideration is performance, the added complexity might slow down the processing, and considering how important the performance of NumericProperty is to Kivy, that might be a cost we don't want to bear, despite the convenience of (ab)using kivy property system outside of its originaly intended use.

Zen-CODE:
@Franccisco Any movement on that change request? Thanks

pythonic64:
Can we add `NumberProperty` to Kivy to avoid misuse of `NumericProperty`?
`NumberProperty` can inherit from `ObjectProperty`, set default value to `0` and default for `baseclass` to `numbers.Number`. Then users can do:
```
a = NumberProperty() # all numbers
b = NumberProperty(baseclass=int) # only integers
c = NumberProperty(baseclass=numbers.Complex) # only complex numbers
# or any other combination
```

The documentation can state the difference between `NumericProperty` and `NumberProperty`. I think that if user finds both classes they will read the docs and learn the difference, so we can avoid this kind of issues in the future.

matham:
>Can we add NumberProperty to Kivy to avoid misuse of NumericProperty?

I really don't think this is a good idea. This would cause huge confusion because you'd have two similarly named properties and no one would be able to keep straight which is which.

But as I mentioned above I don't think this is an issue because I don't think this comes up often (it came up a few times  onlyover the life of kivy?) and that we shouldn't add this just because of consistency.

The best solution I think is to make it clear in the docs, both in `ObjectProperty` and `NumericProperty` about what they are meant to be used for, as well as a overall explanation when we introduce properties to help people understand the context. Then they'll naturally just use `ObjectProperty` as needed.

matham:
https://github.com/kivy/kivy/pull/7439 added the docs and small fixes we discussed.

SomberNight:
https://github.com/kivy/kivy/pull/7439 changed `StringProperty` so that it no longer accepts values with types that are subclasses of `str`.

> I haven't seen any use cases for accepting anything beyond what can be converted to ints/floats and I don't think that's part of common Kivy patterns, considering this issue rarely comes up for numpy types let alone other numeric types. And as I explained NumericProperty, while named generally is actually used with for int/float, the basic numeric type in python and understood by our graphics and layout system.

I have a use case: we use a subclass of `str` for i18n translated strings with `StringProperty` in Electrum.
We subclass `str` to be able to change the language at runtime without restarting the whole application.
This no longer works with https://github.com/kivy/kivy/pull/7439. (ref: https://github.com/spesmilo/electrum/issues/7156)

Is there a problem with this kind of use cases? There might be other projects affected downstream.
Would it be better to use `ObjectProperty` instead? That might have its own issues though...

matham:
Hmm, I guess this is a good reason to revert back StringProperty to use `isinstance`.

I think this is different than for numbers, because we never accepted these additional number types. But we did accept subclasses of str before though. The change to StringProperty to only check `is str` is purely for consistency (because in other places we checked it directly, rather than using `isinstance` and performance (`isinstance` is considerably slower)).

So, I'll revert it. See https://github.com/kivy/kivy/pull/7459.

PR #6985, opened 2020-07-11
Fix scrollview scroll/effect recursion

### LINKED ISSUE #1078: Updating ScrollView from outside and scrolling with mouse after.

### LINKED ISSUE #5638: Window resize recursion error while using ScrollView
### Versions

* Python: 3.6
* OS: Linux (Debian Jessie 8.2)
* Kivy: v1.10.1.dev0, git-3f01ba0, 20180228

* Kivy installation method: 
### Description

I'm using a large custom widget that draws a grid/spots and this is put inside a ScrollView, everything works as intended until the window is resized and a recursion error occurs.

### Code and Logs

```python


[INFO   ] [Logger      ] Record log in /home/xxxx/.kivy/logs/kivy_18-03-01_17.txt
[INFO   ] [Kivy        ] v1.10.1.dev0, git-3f01ba0, 20180228
[INFO   ] [Python      ] v3.6.3 (default, Nov  9 2017, 06:46:20) 
[GCC 4.9.2]
[INFO   ] [Factory     ] 194 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_sdl2, img_gif (img_pil, img_ffpyplayer ignored)
[INFO   ] [GL          ] Backend used <gl>
[INFO   ] [GL          ] OpenGL version <b'4.5.0 NVIDIA 375.82'>
[INFO   ] [GL          ] OpenGL vendor <b'NVIDIA Corporation'>
[INFO   ] [GL          ] OpenGL renderer <b'GeForce GTX 1050 Ti/PCIe/SSE2'>
[INFO   ] [GL          ] OpenGL parsed version: 4, 5
[INFO   ] [GL          ] Shading version <b'4.50 NVIDIA'>
[INFO   ] [GL          ] Texture max size <32768>
[INFO   ] [Window      ] auto add sdl2 input provider



   File "kivy/properties.pyx", line 478, in kivy.properties.Property.__set__
   File "kivy/properties.pyx", line 516, in kivy.properties.Property.set
   File "kivy/properties.pyx", line 571, in kivy.properties.Property.dispatch
   File "/home/xxxx/.virtualenvs/kivytest36/lib/python3.6/site-packages/kivy/effects/scroll.py", line 108, in on_value
     self.scroll = self.value
   File "kivy/properties.pyx", line 478, in kivy.properties.Property.__set__
   File "kivy/properties.pyx", line 516, in kivy.properties.Property.set
   File "kivy/properties.pyx", line 571, in kivy.properties.Property.dispatch
   File "/home/xxxx/.virtualenvs/kivytest36/lib/python3.6/site-packages/kivy/uix/scrollview.py", line 576, in _update_effect_x
     self.scroll_x = -sx
   File "kivy/properties.pyx", line 478, in kivy.properties.Property.__set__
   File "kivy/properties.pyx", line 516, in kivy.properties.Property.set
   File "kivy/properties.pyx", line 571, in kivy.properties.Property.dispatch
   File "/home/xxxx/.virtualenvs/kivytest36/lib/python3.6/site-packages/kivy/uix/scrollview.py", line 557, in _update_effect_bounds
     self._update_effect_x_bounds()
   File "/home/xxxx/.virtualenvs/kivytest36/lib/python3.6/site-packages/kivy/uix/scrollview.py", line 544, in _update_effect_x_bounds
     self.effect_x.value = self.effect_x.min * self.scroll_x

```

### Quick fix

https://github.com/kivy/kivy/blob/3f01ba0b9983ae60ca69d4d34d6b07ad3777c36a/kivy/uix/scrollview.py#L553

This function causes the recursion error, I've put a pass on it temporarily to see if it would solve the issue, it does but I'm not entirely sure if something else is negatively affected. (It doesn't appear so in my case)

```python

    def _update_effect_bounds(self, *args):
        pass
        # if not self._viewport:
        #     return
        # if self.effect_x:
        #     self._update_effect_x_bounds()
        # if self.effect_y:
        #     self._update_effect_y_bounds()

```

### Example code

```python

import kivy
kivy.require("1.10.0")
from kivy.app import App

Config.set('graphics', 'width', '1920')
Config.set('graphics', 'height', '1080')
from random import randint
from kivy.graphics import Color, Ellipse

# Steps to reproduce error
# Window opens, move scrollbars, resize window, maximize
# ** It might not happen right away.

class MyPaintWidget(Widget):
    def __init__(self, **kwargs):
        super(MyPaintWidget, self).__init__(**kwargs)

        self.size_hint=(None,None)
        self.size=(5000,5000)
        with self.canvas:
            Color(1, 1, 0)
            for x in range(5000):
                Ellipse(pos=(randint(0,self.size[0]),randint(0,self.size[1])), size=(5, 5))


class BaseLayout(GridLayout):
    def __init__(self, **kwargs):
        super(BaseLayout, self).__init__(**kwargs)


class BaseApp(App):
    def __init__(self, **kwargs):
        super(BaseApp, self).__init__(**kwargs)

    def build(self):
        base_layout = BaseLayout()

        step_base = ScrollView(
            size_hint=(1,1),
            do_scroll_y=True,
            do_scroll_x=True)
        step_base.bar_width=20
        step_base.scroll_type = ['bars']

        paint_widget = MyPaintWidget()
        step_base.add_widget(paint_widget)

        return step_base

base_app = BaseApp()
base_app.run()

```

matham:
Fixes #5638 fixes #1078

Starting with https://github.com/kivy/kivy/commit/dce021f9d6d4b20f32e93876302c89e27e1b3f44 (which appears to be a fix for https://github.com/kivy/kivy/issues/1078), whenever `scroll_x` or `scroll_y` changed, the `value` of the effect was also updated. IMO this is a buggy change because it may lead to recursion.

Normally, there's no recursion because the effect's `value` and the scroll value in the scrollview are synchronized so they are identical. The problem is that while the scroll effect is active, if the `viewport` size when the effect is started is different from the current size they go out of sync. That's because while the affect is active we use the original widget size (`sw = vp.width - self._effect_x_start_width`), not the current widget size (https://github.com/kivy/kivy/blob/master/kivy/uix/scrollview.py#L630).

This leads to the recursion issue when resizing while the scroll effect is active. Still, why does the linked issue crash even after the effect should have stopped? Because if dragging in one of the bars, the effect is never stopped, so once you drag, it always executes the bad recursive scroll update path.

This fixes that problem by disabling the effect when scrolling on the bars as originally intended (first commit).

But, this wouldn't fix the larger problem because if you were to resize it while the effect is active you'd likely get the same recursion problem. Instead, this recursion problem is fixed by removing the binding `fbind('scroll_x', self._update_effect_bounds)` and `fbind('scroll_y', self._update_effect_bounds)` that causes recursion. Instead, to fix the original issue demonstrated in the following example, we simply make sure the effect is initialized to the current scroll value when the effect is started (second commit).

Ah, and I also added some typing that was useful while debugging to help my IDE.

(copied from https://github.com/kivy/kivy/commit/dce021f9d6d4b20f32e93876302c89e27e1b3f44)
```py
from kivy.app import App
from kivy.clock import Clock


class Root(ScrollView):
    def __init__(self, **kwargs):
        super(Root, self).__init__(**kwargs)

    def after_init(self):
        grid = GridLayout(cols=1, size_hint_y=None)
        grid.bind(minimum_height=lambda w,v: setattr(grid, 'height', v))
        self.add_widget(grid)
        for i in range(200):
            new = Button(text=str(i), size_hint_y=None, height=100)
            grid.add_widget(new)


class Test(App):
    def build(self):
        root = Root()
        root.after_init()
        Clock.schedule_once(lambda *a: setattr(root, 'scroll_y', 0.5), 0)
        return root


app = Test()
app.run()
```

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -140,6 +140,7 @@ All the effects are located in the :mod:`kivy.effects`.
 '''
 
 __all__ = ('ScrollView', )
+
 from functools import partial
 from kivy.animation import Animation
 from kivy.compat import string_types
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -568,8 +568,6 @@ class ScrollView(StencilView):
         fbind('scroll_y', trigger_update_from_scroll)
         fbind('pos', trigger_update_from_scroll)
         fbind('size', trigger_update_from_scroll)
-        fbind('scroll_y', self._update_effect_bounds)
-        fbind('scroll_x', self._update_effect_bounds)
 
         trigger_update_from_scroll()
         update_effect_widget()
@@ -751,6 +749,8 @@ class ScrollView(StencilView):
                 e = self.effect_y if ud['in_bar_y'] else self.effect_x
 
             if e:
+                # make sure the effect's value is synced to scroll value
+                self._update_effect_bounds()
                 if btn in ('scrolldown', 'scrollleft'):
                     if self.smooth_scroll_end:
                         e.velocity -= m * self.smooth_scroll_end
@@ -803,12 +803,18 @@ class ScrollView(StencilView):
 
         if (self.do_scroll_x and self.effect_x and not ud['in_bar_x']
                 and not ud['in_bar_y']):
+            # make sure the effect's value is synced to scroll value
+            self._update_effect_bounds()
+
             self._effect_x_start_width = self.width
             self.effect_x.start(touch.x)
             self._scroll_x_mouse = self.scroll_x
 
         if (self.do_scroll_y and self.effect_y and not ud['in_bar_x']
                 and not ud['in_bar_y']):
+            # make sure the effect's value is synced to scroll value
+            self._update_effect_bounds()
+
             self._effect_y_start_height = self.height
             self.effect_y.start(touch.y)
             self._scroll_y_mouse = self.scroll_y
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -140,7 +140,7 @@ All the effects are located in the :mod:`kivy.effects`.
 '''
 
 __all__ = ('ScrollView', )
-
+from typing import Optional, Type
 from functools import partial
 from kivy.animation import Animation
 from kivy.compat import string_types
@@ -149,6 +149,7 @@ from kivy.clock import Clock
 from kivy.factory import Factory
 from kivy.uix.stencilview import StencilView
 from kivy.metrics import sp, dp
+from kivy.effects.scroll import ScrollEffect
 from kivy.effects.dampedscroll import DampedScrollEffect
 from kivy.properties import NumericProperty, BooleanProperty, AliasProperty, \
     ObjectProperty, ListProperty, ReferenceListProperty, OptionProperty, \
@@ -428,7 +429,8 @@ class ScrollView(StencilView):
     to 0
     '''
 
-    effect_cls = ObjectProperty(DampedScrollEffect, allownone=True)
+    effect_cls: Optional[Type[ScrollEffect]] = ObjectProperty(
+        DampedScrollEffect, allownone=True)
     '''Class effect to instantiate for X and Y axis.
 
     .. versionadded:: 1.7.0
@@ -442,7 +444,7 @@ class ScrollView(StencilView):
 
     '''
 
-    effect_x = ObjectProperty(None, allownone=True)
+    effect_x: Optional[ScrollEffect] = ObjectProperty(None, allownone=True)
     '''Effect to apply for the X axis. If None is set, an instance of
     :attr:`effect_cls` will be created.
 
@@ -452,7 +454,7 @@ class ScrollView(StencilView):
     defaults to None.
     '''
 
-    effect_y = ObjectProperty(None, allownone=True)
+    effect_y: Optional[ScrollEffect] = ObjectProperty(None, allownone=True)
     '''Effect to apply for the Y axis. If None is set, an instance of
     :attr:`effect_cls` will be created.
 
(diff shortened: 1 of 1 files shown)


pythonic64:
I would remove typing, but otherwise request looks good.

matham:
Why remove typing? To me it's like a doc fix which we tend to push into other commits when we work in that file.

matham:
Unless you think we shouldn't include typing in kivy?

pythonic64:
If type hinting should be added to Kivy then new issue should be open to discuss about it. Merging this request with type hinting would add more inconsistencies to framework on top of existing ones, seems out of request scope and seems redundant as default value is already stated in the docstring. 

My current opinion is not to add type hinting to Kivy, but if decided to add it, then it should be done per entire module\package\framework.

matham:
Ok, I do think we should start adding it when we can , because it's very helpful in a IDE, but we don't have to do it here.

AndiEcker:
thank you very much @matham for this fix - unfortunately there is currently no RC version available that is including this fix (the newest one - 2.0.0rc3 - got released on 15-June-2020).

Are there any plans to to make a new RC or for to release the final V 2.0.0?

PR #6990, opened 2020-07-12
Make setuptools use its local distutils

matham:
Fixes the randomly failing CI as described in https://github.com/pypa/setuptools/issues/2212.


matham:
Currently it's stuck at https://github.com/pypa/setuptools/issues/2257 (https://github.com/pypa/setuptools/issues/2259).

PR #6404, opened 2019-07-01
fix pausing without app instance

Fak3:
On android, pause event can come in early, before the app instance is
created.

fixes #6400

Issue #6400, opened 2019-06-25
android: [WindowSDL] No running App found, exit.


Fak3:
Sometimes android app is being paused  during initialization, before the kivy.app.App instance is created. In such case it exits with error `[WindowSDL] No running App found, exit.`

It happens with some apps, when `--window` arg is passed to p4a.
Essentially what happens is: 
1) SDL2 Window is created
2) App paused
3) kivy tries to find App instance to handle pause, and fails
4) stopTouchApp() called from window_sdl2.py `do_pause()`
5) App instance is created
6) stopTouchApp() forces app to exit.

Log:
```
06-25 19:27:20.316 I/python  (27304): [INFO   ] [Window      ] Provider: sdl2
06-25 19:27:20.346 V/SDL     (27304): setOrientation() orientation=-1 width=800 height=600 resizable=true hint=
06-25 19:27:20.356 I/python  (27304): [INFO   ] [GL          ] Using the "OpenGL ES 2" graphics system
06-25 19:27:20.356 I/python  (27304): [INFO   ] [GL          ] Backend used <sdl2>
06-25 19:27:20.356 I/python  (27304): [INFO   ] [GL          ] OpenGL version <b'OpenGL ES 3.1 V@124.0 (GIT@I041b18199e)'>
06-25 19:27:20.366 V/SDL     (27304): surfaceDestroyed()
06-25 19:27:20.366 V/SDL     (27304): nativePause()
06-25 19:27:20.396 I/python  (27304): [INFO   ] [WindowSDL   ] No running App found, exit.
06-25 19:27:20.406 V/SDL     (27304): surfaceCreated()
06-25 19:27:20.406 V/SDL     (27304): surfaceChanged()
06-25 19:27:20.406 V/SDL     (27304): pixel format RGBA_8888
06-25 19:27:20.406 I/python  (27304): [INFO   ] [Window      ] auto add sdl2 input provider
06-25 19:27:20.406 V/SDL     (27304): Window size: 1080x1845
06-25 19:27:20.406 V/SDL     (27304): Device size: 1080x1920
06-25 19:27:20.406 I/python  (27304): [INFO   ] [Window      ] virtual keyboard not allowed, single mode, not docked
06-25 19:27:20.406 V/SDL     (27304): nativeResume()
06-25 19:27:20.576 I/python  (27304): [INFO   ] [Text        ] Provider: sdl2
[...]

PR #6994, opened 2020-07-15
Don't do event loop stuff when stopped.

matham:
Once we call stop on the app, `get_current_app` returns None because the app is reset. The problem is once this happens, kv rules and graphics instructions that reference the current app may get a None value and crash. 

But ideally, once we stop the app, the clock should not continue ticking and the graphics should not be continued to update.

### DIFF
--- a/kivy/base.py
+++ b/kivy/base.py
@@ -387,21 +387,26 @@ class EventLoopBase(EventDispatcher):
         Clock.tick()
 
         # read and dispatch input from providers
-        self.dispatch_input()
+        if not self.quit:
+            self.dispatch_input()
 
         # flush all the canvas operation
-        Builder.sync()
+        if not self.quit:
+            Builder.sync()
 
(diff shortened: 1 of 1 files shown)


kivy/base.py
```diff
@@ -387,21 +387,26 @@ def idle(self):
         Clock.tick()
 
         # read and dispatch input from providers
-        self.dispatch_input()
+        if not self.quit:
```
pythonic64, line 390: Can we do only one `if not self.quit:`?

pythonic64: Probably not, because each of these calls (`self.dispatch_input()`, `Builder.sync()`,...) can cause `app.stop()` which results in `App.get_current_app()` returning `None` which raises an error. 

So a comment about `if not self.quit:` would be good to add.

Fak3:
May be related: #6400 #6404

PR #7017, opened 2020-07-27
Added VectorProperty to kivy Properties

Kolandiolaka:
Created a kivy VectorProperty based on kivy ListProperty.
 
Vector operations were added mostly based on kivy Vector. Vector property has no set dimension unlike kivy Vector but some operations that are not easily scalable are limited to less than three dimensions.

 Did not add intersection methods from kivy Vector but dded 3D rotation and cross-product methods.

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -495,7 +495,7 @@ cdef class Property:
                 'Deprecated property "{}" of object "{}" has been set, it '
                 'will be removed in a future version'.format(self, obj))
             self.deprecated = 0
-        self.set(obj, val)                                                      # method set is called from __set__
+        self.set(obj, val)
 
     def __get__(self, EventDispatcher obj, objtype):
         if obj is None:
@@ -532,10 +532,10 @@ cdef class Property:
             This method then accesses the PropertyStorage(class defined in properties.pxd) of the curresponding Property 'a'
             from EventDispatcher.__storage. (class EventDispatcher defined in _events.pyx)
 
-              Widget.__storage['a']   --> PropertyStorage of 'a'
+              EventDispatcher.__storage['a']   --> PropertyStorage of 'a'
 
             then it sets
-              Widget.__storage['a'].value  = 5 --> PropertyStorage.value = 5
+              EventDispatcher.__storage['a'].value  = 5 --> PropertyStorage.value = 5
 
             Then it calls the dispatch function of the Property, passing in the EventDispatcher class its an attribute ofself.
             a.dispatch(Widget)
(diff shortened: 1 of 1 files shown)


matham:
Thanks for your contribution, but I think this is a perfect candidate for garden because I don't think this fits in kivy itself. 

I don't think there is a general enough need for this to belong in Kivy and for us to take the burden of maintaining it. That's why we created garden so contributors can create packages that others can use, without it having to go in kivy directly.

In particular, we have a [cython garden flower template](https://github.com/kivy-garden/cython_flower) that you can use and it should work fine for your cython code!

Kolandiolaka:
Thank You, I will check out the cython flower.

Kolandiolaka:
But if I were to make a case, 

I felt the need for such a property white manipulation Widget's positions and size. I think those are natural vectors and often it becomes cumbersome to write down otherwise simple vector manipulations componentwise. This makes working with them a bit tedious, the code unnecessarily long and decreases the code readability, especially in kv language. 

While some of the methods like 3D rotation might be overkill, a property very similar to ListProperty with at least vector additions and multiplication might not be that outlandish.

But as you said, rather than it being general it might have been an issue specific to what I was working on.

matham:
>But as you said, rather than it being general it might have been an issue specific to what I was working on.

I do think that's the case. This is why I think garden is better because if in the future people use this a lot then we could consider again adding it to kivy. But I like to keep kivy as minimal as possible to reduce our maintence burden. I.e. once something is in kivy we are stuck with it and have to maintain it forever. But if it's in garden there's much more flexibility and even the possibility to start from scratch in a new project if needed.

tshirtman:
I understand it's not currently strictly necessary in kivy, but i think it would be useful, just like ColorPropery is a better option for colors than ListProperty, a VectorPropery would allow for more on everything in python that is a 2D (or even 3D) coordinate, it could make sense to convert all kivy coordinates to this VectorProperty, and start reaping the benefits while developing apps.

Zen-CODE:
I tend to agree with @matham here. It adds more complexity which is not necessary and most won't use. Although we could use it for co-ordinates, it would make kivy even harder to learn and more obscure without any clear benefit. And who would do that re-write? And update all the docs? It's a huge change.

I really do like this feature, but I also agree the garden is a better place for this. Keeping Kivy minimal is important, both for maintenance, speed and  "minimizing the learning curve" reasons. I would vote it goes to the garden, if @Kolandiolaka is willing to move it there? Thanks for this by the way, but not everything can (or needs to) go into the core framework. But keep up the good work :-)

matham:
I'm going to close since there's no plan to currently merge this.

PR #7021, opened 2020-07-29
Use ColorProperty for remaining color properites

pythonic64:
Continuation of https://github.com/kivy/kivy/pull/6918

Use ColorProperty for:
- GestureSurface.color
- Label.outline_color
- Label.disabled_outline_color

### DIFF
--- a/kivy/uix/gesturesurface.py
+++ b/kivy/uix/gesturesurface.py
@@ -22,7 +22,7 @@ from kivy.vector import Vector
 from kivy.uix.floatlayout import FloatLayout
 from kivy.graphics import Color, Line, Rectangle
 from kivy.properties import (NumericProperty, BooleanProperty,
-                             DictProperty, ListProperty)
+                             DictProperty, ColorProperty)
 from colorsys import hsv_to_rgb
 
 # Clock undershoot margin, FIXME: this is probably too high?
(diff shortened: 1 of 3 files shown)


kivy/uix/gesturesurface.py
```diff
@@ -238,9 +238,13 @@ class GestureSurface(FloatLayout):
             Color used to draw the gesture, in RGB. This option does not
             have an effect if :attr:`use_random_color` is True.
 
-            :attr:`draw_timeout` is a
+            :attr:`color` is a
             :class:`~kivy.properties.ListProperty` and defaults to
```
tshirtman, line 242:
```suggestion
            :class:`~kivy.properties.ColorProperty` and defaults to
```

kivy/uix/label.py
```diff
@@ -792,10 +792,15 @@ def on_ref_press(self, ref):
     .. versionadded:: 1.10.0
 
     :attr:`outline_color` is a :class:`~kivy.properties.ListProperty` and
```
tshirtman, line 794:
```suggestion
    :attr:`outline_color` is a :class:`~kivy.properties.ColorProperty` and
```

kivy/uix/label.py
```diff
@@ -806,6 +811,11 @@ def on_ref_press(self, ref):
 
     :attr:`disabled_outline_color` is a :class:`~kivy.properties.ListProperty`
```
tshirtman, line 812:
```suggestion
    :attr:`disabled_outline_color` is a :class:`~kivy.properties.ColorProperty`
```

kivy/uix/gesturesurface.py
pythonic64: The existing approach is the same as in previous pull request. Like this for example: https://github.com/kivy/kivy/blob/master/kivy/uix/actionbar.py#L214, and that was approved by matham.

matham: I think I overlooked that by only looking at the diff 😬 sorry

tshirtman: mistakes have been made :P

PR #6971, opened 2020-06-28
Fix an inconsistency between docs and code on Camera

misl6:
This PR fixes an inconsistency between docs and code.

The issue was been introduced by PR #658

PR #7049, opened 2020-08-18
Fix GI camera provider crash when no texture is available after loading

Cheaterman:
Traceback:

```
   File "/home/cheaterman/Dev/Kivy/kivy/core/camera/camera_gi.py", line 137, in _update
     self.dispatch('on_load')
   File "kivy/_event.pyx", line 705, in kivy._event.EventDispatcher.dispatch
   File "kivy/_event.pyx", line 1248, in kivy._event.EventObservers.dispatch
   File "kivy/_event.pyx", line 1172, in kivy._event.EventObservers._dispatch
   File "/home/cheaterman/Dev/Kivy/kivy/uix/camera.py", line 112, in _camera_loaded
     self.texture_size = list(self.texture.size)
 AttributeError: 'NoneType' object has no attribute 'size'
```

Also see #6971.

### DIFF
--- a/kivy/uix/camera.py
+++ b/kivy/uix/camera.py
@@ -90,7 +90,9 @@ class Camera(Image):
         fbind('resolution', on_index)
         on_index()
 
-    def on_tex(self, *l):
+    def on_tex(self, camera):
+        self.texture = texture = camera.texture
+        self.texture_size = list(texture.size)
         self.canvas.ask_update()
 
     def _on_index(self, *largs):
@@ -102,14 +104,9 @@ class Camera(Image):
         else:
             self._camera = CoreCamera(index=self.index,
                                   resolution=self.resolution, stopped=True)
-        self._camera.bind(on_load=self._camera_loaded)
         if self.play:
             self._camera.start()
             self._camera.bind(on_texture=self.on_tex)
-
-    def _camera_loaded(self, *largs):
-        self.texture = self._camera.texture
-        self.texture_size = list(self.texture.size)
 
     def on_play(self, instance, value):
         if not self._camera:


Cheaterman:
Might seem a bit aggressive but since we can't assume the texture is assigned in `on_load`... Regarding texture size, I'm not sure how guaranteed we are it won't change dynamically (say when your phone flips from portrait to landscape?), so that might be a better idea either way ; but for `self.texture`, if we can assume the same texture will be reused across the CoreCamera lifetime (even through `start()`s and `stop()`s), we could make sure to only assign it once?

I'm not sure that's a safe assumption either, so here we are.

Cheaterman:
(FWIW I'm not quite sure how the `Camera` widget was invoked - I am using `garden.zbarcam` (which uses `garden.xcamera`) with default settings)

wolfmanjm (commented):
This indeed fixes the issue for me. My camera works again with master.

Cheaterman:
Happy to help 😊

kivy/uix/camera.py
```diff
@@ -90,7 +90,9 @@ def __init__(self, **kwargs):
         fbind('resolution', on_index)
         on_index()
 
-    def on_tex(self, *l):
+    def on_tex(self, camera):
+        self.texture = texture = camera.texture
+        self.texture_size = list(texture.size)
```
pythonic64, line 95: You don't have to convert to `list` because `texture_size` is a `ListProperty` and assigning to it will convert `texture.size` to `ObservableList`.

kivy/uix/camera.py
```diff
@@ -90,7 +90,9 @@ def __init__(self, **kwargs):
         fbind('resolution', on_index)
         on_index()
 
-    def on_tex(self, *l):
+    def on_tex(self, camera):
+        self.texture = texture = camera.texture
```
akshayaurora, line 94: We can still check and update only when texture is None or when camera.texture.size is not equal to self.texture_size, right ?

akshayaurora (changes requested):

kivy/uix/camera.py
pythonic64: Actually, you don't have to set `texture_size` at all as it will be updated by `on_texture` method from `Image` class.

pythonic64: You can safely remove line `self.texture_size = list(texture.size)` from `on_tex` and `_camera_loaded` methods.

matham:
Seems to have been fixed in https://github.com/kivy/kivy/pull/7071.

Cheaterman:
TFW ^^" anyway at least it's fixed. The line deletions I've made might still be a good idea though - mo' code mo' problems.

wolfmanjm:
FYI #7071 breaks gi based camera in 2.0.0, whereas this PR stops it crashing and allows it to work again.
I would recommend reverting #7071 and using this PR instead.

kivy/uix/camera.py
matham: @Cheaterman I'm perplexed at where this `tex` variable comes from. Is this from I couldn't use pycharm to tell me where `self.tex` is defined.

matham (approved):
This seems correct as a texture may not exist at `on_load`.

Cheaterman:
And there you go 😊 thanks everyone !

PR #7054, opened 2020-08-20
User Agent and Cookies added to UrlRequest

ibrahimcetin:
I added user agent and cookies to UrlRequest


Zen-CODE:
Thanks. Would it be too much to ask to please update the documentation to include example usage? i.e. the module doc string.  Otherwise it might go unnoticed. Thanks

tshirtman:
I guess we could just merge it and add docs later, rather than let it rot.

tshirtman (approved):

Zen-CODE:
@ibrahimcetin Thanks! +1

ibrahimcetin:
Sorry, I couldn't add docs because my computer was in technical service. I will add it after I get my computer.

Zen-CODE:
That would  be appreciated. Thanks

PR #7058, opened 2020-08-25
More correct definition of each android

wrongnull:


Zen-CODE:
The `ANDROID_ARGUMENT` is set from buildozer, so should always be there for a Kivy App on android. Where do you get the information that `ANDROID_DATA` will always be there?

tshirtman (approved):

matham:
Also, shouldn't we be checking the value of the variable?

Zen-CODE:
@matham I don't think so. It's simply to  detect whether it's android or not. The value is irrelevant surely?

matham:
Right, but what if on my desktop I have it set to `ANDROID_ARGUMENT=0` for whatever reason? We can't assume it's not set elsewhere. We can only assume it's set to some known value on android so we should check for that value.

Zen-CODE:
Fair point, but there are no such existing checks. It might be nice to add them, but I think the point here is that adding a check for `ANDROID_DATA` does not really add value. I cannot find any documentation that says it will always be there?

So, thanks @wrongnull , we really do appreciate the effort, but unless you can point us to official documentation that justifies this change, I would suggest we close it?

Zen-CODE:
Closing this due to lack of feedback (and for reasons outline above). Please feel free to re-open if you feel the PR is still valid 👍 Thanks

PR #7070, opened 2020-09-08
Added the Russian translation of README.md

leader2one:


README-RU.md
```diff
@@ -0,0 +1,129 @@
+Kivy
```
leader2one, line 1: Russian version of README

tshirtman:
We don't have localized version of README or any documentation, i appreciate the effort, but why do you feel this is needed? Is it useful enough to justify having to maintain multiple versions of the documentation in the future? Keep it mind that having multiple versions means they need to be updated together, so if someone updates the english version, and we have 50 languages to update, then we need potentially 50 contributors to contribute a corresponding update to their version.

Zen-CODE:
@leader2one We really do appreciate the contribution, and would love to see you contributing again in the future. As @tshirtman  pointed out though, a Russian translation is not a realistic goal at this point. Please don't be discouraged, but perhaps look at other ways to contribute? Fixing reported bugs? Tutorials? Adding tests?

Thanks again. Remember, closing this does not does not belittle your effort, it's just something we cannot use :-)

PR #7071, opened 2020-09-09
Fixes crash during camera configuration

misl6:
This PR fixes a potential runtime crash during camera re-configuration where the `texture` it's still `None`.


tshirtman (approved):
LGTM

matham:
This causes https://github.com/kivy/kivy/issues/7259. I believe the issue is that when `on_load` is dispatched, we may not have the final texture, and the real texture never gets set. So https://github.com/kivy/kivy/pull/7049 is a better fix as it only sets the texture when the texture is updated.

PR #7074, opened 2020-09-11
Fix: Splitter handling offset.

smrg-lm:
Hi

There is an offset problem when moving the splitter bar beyond its limit position in parent. The offset of the pointer isn't calculated and gets visual displacement. The method _is_moving in the patch is a refactored version of the following example (that is only for the fist if branch in Splitter.strip_move):

```python
if diff_y == 0:
    return
elif diff_y > 0:
    if sz_frm == 't' and touch.y <= self.top:
        return
    elif sz_frm == 'b' and touch.y <= self.y:
        return
elif diff_y < 0:
    if sz_frm == 't' and touch.y >= self.top:
        return True
    elif sz_frm == 'b' and touch.y >= self.y:
        return
```
Note this check don't keep pointer's on_touch_down position within the widget because that requires to add instance variables.

### DIFF
--- a/kivy/uix/splitter.py
+++ b/kivy/uix/splitter.py
@@ -291,6 +291,20 @@ class Splitter(BoxLayout):
         else:
             self.height = max(self.min_size, min(self.height, self.max_size))
 
+    @staticmethod
+    def _is_moving(sz_frm, diff, pos, minpos, maxpos):
+        if sz_frm in ('l', 'b'):
+            cmp = minpos
+        else:
+            cmp = maxpos
+        if diff == 0:
+            return False
+        elif diff > 0 and pos <= cmp:
+            return False
+        elif diff < 0 and pos >= cmp:
+            return False
+        return True
+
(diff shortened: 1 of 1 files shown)


akshayaurora:
Can you give more details on what’s the offset issue? I only see a issue with your branch, master seems to work as expected when testing with running `python kivy/uix/splitter.py`

smrg-lm:
It's very specific to the splitter:

Peek 2020-09-27 12-08

As it receives a delta It goes down and up ok because is always moving but when it goes down again the pointer don't visually reach the splitter bar to move it down which is the behavior I see in other frameworks in general:

Peek 2020-09-27 12-16

akshayaurora:
ok, seems you fixed it only on vertical movement.

akshayaurora (approved):

akshayaurora:
@smrg-lm nice work 👍

smrg-lm:
Thank you, I was rebasing and pushing but you solved it.

PR #7080, opened 2020-09-13
Generate Kivy.app on the CI

matham:
This PR includes a re-work of kivy-sdk-packager for osx so that we can generate a kivy bundle and dmg more incrementally and fixes a bunch of issues there that prevented it to work on the CI. You can see the full readme at https://github.com/kivy/kivy-sdk-packager/tree/master/osx.

With these changes we can now generate a Kivy.dmg for use in packaging. Additionally, we're now also testing the generated Kivy.dmg in a separate clean environment. The only test failure is with gstreamer, but that can be fixed later.

Finally, I also updated the osx docs to make it simpler and point directly to kivy-sdk-packager for more complete docs.

An unrelated issue is that osx wheels tests fail, but it doesn't fail for osx unittests that are built in place. It makes me wonder if the osx wheels are safe to use in a clean environment.

### DIFF
--- a/.ci/osx_ci.sh
+++ b/.ci/osx_ci.sh
@@ -159,10 +159,6 @@ mount_osx_app() {
 activate_osx_app_venv() {
   pushd app/Kivy.app/Contents/Resources/venv/bin
   source activate
+  source kivy_activate
   popd
-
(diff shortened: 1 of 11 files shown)


matham:
OSX wheels and Kivy.app pass all tests now!

Issue #7109, opened 2020-09-27
Add a check for python version in `make` and gracefully exit rather than fail on python version 2.


akshayaurora:
**Software Versions**
* Python:
* OS: OSX 10.15.6
* Kivy: master 27 sept 2020
* Kivy installation method: clone master and `make`

**Describe the bug**
Make fails with `python` referring to python2 does not fail gracefully

**Expected behavior**
```
 kivy % make 
python setup.py build_ext —inplace
Python version 2 is not supported, please use python3
```
**To Reproduce**
```
> git clone github.com/kivy/kivy
> cd kivy
> make 
```

**Code and Logs and screenshots**
```
 kivy % make 
python setup.py build_ext --inplace
  File "setup.py", line 1050
    def glob_paths(*patterns, excludes=('.pyc', )):
                                     ^
SyntaxError: invalid syntax
make: *** [build] Error 1
```

**Additional context**

tshirtman:
maybe something like
```python
PYTHON_MINIMUM_VERSION = (3, 6, 0)

if sys.version_info < PYTHON_MINIMUM_VERSION:
     logger.error("Kivy requires at least python %s.%s.%s", *PYTHON_MININUM_VERSION)
```

akshayaurora:
@tshirtman your solution is already implemented in setup.py, issue is it fails due to syntax error in py2 while it works with py3

tshirtman:
Oh right, makes sense.

tshirtman:
but in which situation does the one in makefile helps?

akshayaurora:
If you try to run `make` to compile kivy and your `python` points to a version < 3.6 it would now simply give you this error. 

```
kivy % make
Makefile:12: *** "detected Python 2.7 need Python >= 3.6".  Stop.
```

tshirtman:
right, but few users actually do that, right?

Maybe we should use a stub setup.py that is python2 compatible if we want that, and only conditionally import/run the actual (renamed) setup.py from it if we detect that we have a supported python version, then?

akshayaurora:
@tshirtman I am not really sure of the consequences of that but #7111 does that, feel free to merge which ever method you feel works,

matham:
Eh. I'm not sure we should have non standard setup.py (msetup.py) just for make users, which are pretty rare these days (except for us core-devs). The default cade should be for 3+, not 2.

matham:
[...]

PR #7110, opened 2020-09-27
Detect python verion and gracefully fail on unsupported ver.

akshayaurora:
fixes https://github.com/kivy/kivy/issues/7109

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,16 @@
+ifeq (, $(shell which python ))
+  $(error "PYTHON=$(PYTHON) not found in $(PATH)")
+endif
+
 PYTHON = python
+PYTHON_VERSION_MIN=3.6
+PYTHON_VERSION=$(shell $(PYTHON) -c 'import sys; print("%d.%d"% sys.version_info[0:2])' )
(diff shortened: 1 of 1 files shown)


akshayaurora:
closing in support for #7111

Makefile
```diff
@@ -1,4 +1,16 @@
+ifeq (, $(shell which python ))
+  $(error "PYTHON=$(PYTHON) not found in $(PATH)")
+endif
+
 PYTHON = python
+PYTHON_VERSION_MIN=3.6
```
matham, line 6: I'd only check here for python2. And if it's not python 2 I'd then rely on https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires in the setup.cfg, rather than checking it here. That way all the metadata is in one place (going under the assumption people have a updated pip).

matham (changes requested):

akshayaurora:
@matham made changes to Make here and setup.cfg here https://github.com/kivy/kivy/pull/7112

matham (approved):
LGTM (but I didn't test).

PR #7111, opened 2020-09-27
move setup.py to msetup.py inline with suggestions from #7109, closes…

akshayaurora:
… #7109

more details at https://github.com/kivy/kivy/issues/7109

This is a alternative solution to https://github.com/kivy/kivy/pull/7110 for #7109


setup.py
```diff
@@ -1,1137 +1,9 @@
...
-        author='Kivy Team and other contributors',
-        author_email='kivy-dev@googlegroups.com',
-        url='http://kivy.org',
-        license='MIT',
-        description=('Kivy examples.'),
-        long_description_content_type='text/markdown',
-        long_description=get_description(),
-        data_files=list(examples.items()))
+import subprocess
+subprocess.call(['python', 'msetup.py'] + sys.argv[1:])
```
tshirtman, line 9: no that we can't do, at the very least we should use sys.executable to make sure we run it with the same python version that called us, but i think just importing the module and calling a function from it would be better.

tshirtman (changes requested):

setup.py
```diff
@@ -1,1137 +1,9 @@
...
-    for pat in patterns:
-        for f in base.glob(pat):
-            if f.suffix in excludes:
-                continue
-            files.append(str(f.relative_to(base)))
-    return files
+PYTHON_MINIMUM_VERSION = (3, 6, 0)
 
+if sys.version_info < PYTHON_MINIMUM_VERSION:
+    sys.stderr.write("\nKivy requires Python >= " + str(PYTHON_MINIMUM_VERSION) + ' detected:' + str(sys.version_info[:3]) + ' \n\n')
```
tshirtman, line 5:
```suggestion
    sys.stderr.write("\nKivy requires Python >= %s  detected: %s \n\n", PYTHON_MINIMUM_VERSION, sys.version_info[:3])
```

tshirtman (changes requested):

PR #7113, opened 2020-09-27
Add run on repl.it badge to README

Viniciusgruba:
This pull request adds a  badge to the . This will allow users to easily run this repository in their browser, without having to set up an environment. You can learn more about Repl.it [here](https://repl.it).

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,2 @@
+language = "python3"
+run = ""
--- a/README.md
+++ b/README.md
@@ -138,3 +138,5 @@ Support this project by becoming a sponsor. Your logo will show up here with a l
 <a href="https://opencollective.com/kivy/sponsor/9/website" target="_blank"><img src="https://opencollective.com/kivy/sponsor/9/avatar.svg"></a>
 
 
+
+[![Run on Repl.it](https://repl.it/badge/github/kivy/kivy)](https://repl.it/github/kivy/kivy)


matham:
Thanks for sharing the link, but I don't think it is appropriate for kivy to link to it given that kivy doesn't actually run there.

You'd either need to compile kivy every time you start the repl, which fails anyway since you need to install all the deps on linux. Or you can use an existing wheel with pip, but then linking to github master is misleading. Finally, kivy doesn't actually run headless so I don't think this will be useful much.

tshirtman:
yeah, doesn't seem to work, i guess if there was a way to either force python3.7 or to force kivy version to be 2.0.0rc3, it could work, but i didn't manage to do that.

I'm pretty sure i remember someone linking a similar service on which they were able to run kivy, though, but can't seem to find it again.

matham:
You can install kivy more easily with `pip install kivy --pre`. I didn't see the x Window because of an ad-blocker, but now it runs. 

However, it's not connected to the cloned version of kivy and you need to `cd` to a different directory to work. It would be better to start from a clear shell, rather once that cloned kivy.

It's also somewhat slow and unresponsive. Do you think there's a benefit to linking to it?

tshirtman:
I'm unsure, i feel like our installation procedure has improved a lot compared to a few years ago, and it's rarely an issue for an user to install kivy to start experimenting with it, so while this might give a solution to some users, i'm not sure it's a really interesting one, so, meh? i'm not against it if there is literally 0 instructions to maintain to use it, but i'm not the one going to make a strong case for it.

matham:
OK, closing then because it doesn't seem it would be that helpful for us or our users.

PR #7118, opened 2020-10-01
[wip] optimize GridLayout

gottadiveintopython:
#6741 (a recently merged PR I wrote) may be terrible in terms of efficiency, because it creates list/tuple in the situation where iterator is enough. This PR tries to fix it.

### TODO

- [x] replace temporal lists/tuples with iterators
- [x] avoid holding pre-calculated indices (`_col_and_row_indices`)
- [x] keep the compatibility with [Bubble](https://kivy.org/doc/master/api-kivy.uix.bubble.html#kivy.uix.bubble.Bubble.orientation)
- [x] keep the compatibility with [TabbedPanel](https://kivy.org/doc/master/api-kivy.uix.tabbedpanel.html)
- [ ] ~~raise an exception when the configuration is wrong (https://github.com/kivy/kivy/pull/6741#discussion_r498640070)~~ (not going to do this at least in this PR)

### DIFF
--- a/kivy/uix/gridlayout.py
+++ b/kivy/uix/gridlayout.py
@@ -255,7 +255,7 @@ class GridLayout(Layout):
 
     orientation = OptionProperty('lr-tb', options=(
         'lr-tb', 'tb-lr', 'rl-tb', 'tb-rl', 'lr-bt', 'bt-lr', 'rl-bt',
-        'bt-rl', ))
+        'bt-rl'))
     '''Orientation of the layout.
 
     :attr:`orientation` is an :class:`~kivy.properties.OptionProperty` and
--- a/kivy/uix/gridlayout.py
+++ b/kivy/uix/gridlayout.py
@@ -309,15 +309,15 @@ class GridLayout(Layout):
 
     @property
     def _fills_row_first(self):
-        return self.orientation[0] not in 'tb'
+        return self.orientation[0] in 'lr'
 
     @property
     def _fills_from_left_to_right(self):
-        return 'rl' not in self.orientation
+        return 'lr' in self.orientation
 
     @property
     def _fills_from_top_to_bottom(self):
-        return 'bt' not in self.orientation
+        return 'tb' in self.orientation
 
     def _init_rows_cols_sizes(self, count):
         # the goal here is to calculate the minimum size of every cols/rows
--- a/kivy/uix/gridlayout.py
+++ b/kivy/uix/gridlayout.py
@@ -352,7 +352,7 @@ class GridLayout(Layout):
         self._rows_sh = [None] * current_rows
         self._rows_sh_min = [None] * current_rows
         self._rows_sh_max = [None] * current_rows
-        self._col_and_row_indices = tuple(_create_col_and_row_index_iter(
+        self._col_and_row_indices = tuple(_create_idx_iter(
             current_cols, current_rows, self.orientation))
 
         # update minimum size from the dicts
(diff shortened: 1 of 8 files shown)


gottadiveintopython:
Unit tests still fail, which means [my assumption](https://github.com/kivy/kivy/pull/6741#issuecomment-702656508) might be wrong.

gottadiveintopython:
@matham  If I try to make GridLayout raise an exception like you [suggested](https://github.com/kivy/kivy/pull/6741#discussion_r498640070), it breaks some stuff (ActionBar, `examples/widgets/tabbed_panel_showcase.py`, and potentially more), and it's hard to find the cause of it. Do have any idea why they don't work? If you don't, how about doing it on another PR?

gottadiveintopython:
@matham If I try to change GridLayout as you [suggested](https://github.com/kivy/kivy/pull/6741#discussion_r498640070), it will break some stuff (ActionBar, `examples/widgets/tabbed_panel_showcase.py`, and potentially more), and it's hard to find the cause of it. So I want to do it in a separated PR. Do you agree with it?

If you want to confirm the issue, use 9ef3e94a254d264f71cd37008260d6b1b8be49de, run unit tests, run `examples/widgets/tabbed_panel_showcase.py`.

kivy/uix/gridlayout.py
```diff
@@ -307,8 +308,16 @@ def on_children(self, instance, value):
                 'Too many children in GridLayout. Increase rows/cols!')
 
     @property
-    def _fills_row_first(self):
-        return self.orientation[0] in 'lr'
+    def fills_row_first(self):
+        return self.orientation[0] not in 'tb'
+
+    @property
+    def fills_from_left_to_right(self):
```
matham, line 315: Perhaps these should be private properties (with leading underscore). And the same for the other properties.

matham (changes requested):
It is ok to do the exception thing in a different PR.

I did notice in my app that I suddenly got a bunch of these warnings, when I think I didn't before. So perhaps instead of raising an error or a warning, we could just formalize that if neither is specified, we default it to e.g. cols being one.

kivy/uix/gridlayout.py
gottadiveintopython: > Perhaps these should be private properties

got it.

> we default it to e.g. cols being one.

Yes, that's the behavior before #6741, and thus probably is the safest option I believe.

kivy/uix/gridlayout.py
```diff
@@ -308,7 +309,15 @@ def on_children(self, instance, value):
 
     @property
     def _fills_row_first(self):
-        return self.orientation[0] in 'lr'
+        return self.orientation[0] not in 'tb'
```
gottadiveintopython, line 312: Bubble over-writes `orientation`, makes it only accepts `vertical` and `horizontal`, which is why I changed this property. By this change, Bubble is always `lr-tb`.

gottadiveintopython:
I misunderstood what you said. but anyways, I reverted the behavior of under the wrong configuration

kivy/uix/gridlayout.py
matham: I think this is too sensitive to the issue of action-at-a-distance, where the exact way we do it here effect bubble. Instead, it's better if you override this in bubble by re-implementing the methods there and always return `True/False`

matham (changes requested):
Other than the requested change for bubble, it seems ok from a brief skim.

matham (approved):

PR #7130, opened 2020-10-06
Enable SDL2 ios keyboard selection

GoBig87:
This allows ios developers the ability to change keyboard types

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -468,7 +468,7 @@ cdef class _WindowSDL2Storage:
         SDL_DestroyWindow(self.win)
         SDL_Quit()
 
-    def show_keyboard(self, system_keyboard, softinput_mode):
+    def show_keyboard(self, system_keyboard, softinput_mode, keyboard_type=None):
         if SDL_IsTextInputActive():
             return
         cdef SDL_Rect *rect = <SDL_Rect *>PyMem_Malloc(sizeof(SDL_Rect))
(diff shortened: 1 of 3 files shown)


matham:
https://github.com/kivy/kivy/pull/7231 does something related for android.

misl6:
@matham @GoBig87 that's a nice feature, but before reviewing it, We should make a decision on how to handle this kind of feature that needs patching on a third-party library.

What I'm concerned about it's that We (potentially) need to re-build the patch every time a specific recipe needs to be updated. It's ok for fixing cross-compiling issues but seems a potential headache when We should do that in order to support features that could be merged on the third-party library upstream. 

How about asking for a merge into the SDL upstream? Obviously couldn't be merged as-is and should be improved (in order to make it backward-compatible), but shouldn't be too hard.

matham:
@misl6 I'm a little confused about your comment. Because this PR has no patches. Were you talking about the android PR?

But I'm wondering if this PR is fine?

misl6:
@matham 

This PR itself doesn't work (and shouldn't me merged) until We merge https://github.com/kivy/kivy-ios/pull/559 on kivy-ios.

The kivy-ios related PR provides the needed edits to the `SDL_StartTextInput`.

(Basically the two PRs complete each other, and if one of the two PRs doesn't get merged We have a broken upstream for kivy or kivy-ios)

matham:
Ok, thanks for clarifying. I missed that context. I put it on hold until the kivy-ios side is worked out.

PR #7132, opened 2020-10-06
Add support for Python 3.9.

matham:
Bumps the dependencies versions to include newer deps.


matham:
Pillow is currently blocking support for 3.9: https://github.com/python-pillow/Pillow/issues/4953.

tshirtman:
Are we making it a non-optional dependency? I guess we can wait a bit for them, but we could certainly release without it (and release a 2.0.1 later to add the support and the various bug fixes we'll probably have), if they get stuck for too long.

matham:
On the CI we install all our deps, including all our optional ones so it gets stuck there. We shouldn't rush, I think. We waited this long and there are still tasks to be done anyway before we can practically do a release (release notes etc) so it won't hurt to wait for pillow to make a release, which should be soon.

radarhere:
Fyi, a new release of Pillow should come out on October 15 with Python 3.9 support.

hugovk:
> Pillow is currently blocking support for 3.9: [python-pillow/Pillow#4953](https://github.com/python-pillow/Pillow/issues/4953).

Pillow 8.0.0 supporting Python 3.9 is now out!

* https://pillow.readthedocs.io/en/stable/releasenotes/8.0.0.html
* https://pypi.org/project/Pillow/8.0.0/
* https://twitter.com/PythonPillow/status/1316452445824921602

matham:
> Pillow 8.0.0 supporting Python 3.9 is now out!

Thank you!

We should have 3.9 nightly wheels on our server as soon as the CI builds it in about 4 hours and I'll make a rc release to pypi shortly after.

PR #7152, opened 2020-10-15
Stop building wheels for RPi stretch

matham:
This stops building wheels for RPi stretch and python 3.5 manylinux. 

Previously we already didn't build 3.5 wheels for some platforms because master does not support 3.5. However, now pip has deprecated support for 3.5 and will completely stop supporting it in January. In addition now Kivy definitely doesn't build on 3.5 anymore because of the python version selector metadata so pip refuses to build kivy on 3.5.

Stretch comes with python 3.5, hence we cannot build wheels anymore on it. We can re-enable it if someone thinks it's worth it by installing python 3.6 on stretch. But that will probably not be useful because most people will expect 3.5 if they use stretch.

### DIFF
--- a/.ci/build-wheels-linux.sh
+++ b/.ci/build-wheels-linux.sh
@@ -43,7 +43,7 @@ make distclean;
 
 cd /io;
 for PYBIN in /opt/python/*3*/bin; do
-    if [[ $PYBIN != *"34"* ]]; then
+    if [[ $PYBIN != *"34"* && $PYBIN != *"35"* ]]; then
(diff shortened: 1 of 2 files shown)


matham:
@Lauszus any thoughts?

matham:
Merged so that we could make a rc4 release, but we can figure out what to do about this, if anything, later.

Lauszus:
@matham fine with me. If people can figure out how to compile Python 3.6 they can probably also figure out how to compile Kivy from source.

PR #7170, opened 2020-10-22
Add (and run) pre-commit hooks for enforcing basic coding standards

rambo:
## Pros:
  - Everyone will have one standard pep8 style thanks to [black][black]
  - line endings, tabs vs spaces all of those problems are thing of the past
  - very easy to add mode hooks supported by the [pre-commit][precommit] framework

[black]: https://github.com/psf/black
[precommit]: https://pre-commit.com/

## Cons:

Just one, but it's big:

There's never a "good time" to do something like this, all existing PRs must be rebased after merge and the hooks run for example but I doubt there's ever going to be a time when there are not several of those in process.


rambo:
The line width is set to 80 as is/was the flake8 config but in these days or large monitors I could argue for at least 120 chars  instead of the oldskool default text terminal width.

matham:
Thanks for your efforts on this, but I can't really see us merging this given that we are a long existing project.

The benefits are minor, if at all.
* We don't get that many PRs where this is a large issue. At most it requires one extra commit to fix potential issues, but with any decent editor it should flag PEP8 issues well before you commit. And for core-devs it should be second nature to fix PEP8 issues before committing or creating PRs.
* Besides PEP8 issues, we have rarely had to correct style issues in incoming PRs.
* We are a small team so there's very little strain in regards to style issues.

But the cons are enormous.
* We are a long existing project, to suddenly change to an entirely different formatting scheme would require (much) more work than the status quo and there would need a good reason for it, which I don't see presented and I'm not aware of a strong reason.
* It changes the complete code-base git history. This kind of change is almost never justified, especially if it's because of style changes. Style changes should happen in the background and not get in the way, otherwise, they are rarely the effort.
* We won't be able to accept PRs generated from commits created in github's user interface as the style hook would not have run. To fix this we'd need to run a special bot that fixes all commits before the PR is merged. This adds further complexity and I'm not inclined to trust a bot to do such automatic changes on our code base without good reason for the trust.
* Similarly, what happens if a core-dev makes a commit and pushs, without going through a PR, but also doesn't run the hook. The next time a person commits their code would be re-written in a unrelated PR. There has to be good reason to force us to be disciplined to never accept even a single change without running through black first and I don't see one presented.
* Commit hooks are slow. I don't even run our pep8 hook since I rely on PyCharm to point out PEP8 issues. So this will probably slow down commits.

If this was a new project, we'd should consider whether black is a good fit for us. But given what I mentioned, I don't think black is for us.

tshirtman:
also a good reference about pep8 and code quality
https://www.youtube.com/watch?v=wf-BqAjZb8M

Pep8 doesn't need to be a hard, enforced rule, it's a style guideline, and that's good like this imho.

matham:
>Pep8 doesn't need to be a hard, enforced rule, it's a style guideline, and that's good like this imho.

I'm assuming by `like this` you mean like it is currently in Kivy and that you agree with me :)

rambo:
In general I get where you are coming from, just some specific comments:

> It changes the complete code-base git history

Which is why there's the ignore-revs -file, or is there some other reason than git-blame giving the "wrong commit"
for a change for this being a problem ?

> So this will probably slow down commits.

You already have the flake8 hook there (or is no-one actually using the hook if you say it slows you down?) and black feels faster to me (I haven't done any actual timing comparisons).

Technically yes; any hook will slow down commits from not having a hook at all, whether it actually matters is the important question (having the commit take an extra second or few is not a problem IMO)

> Similarly, what happens if a core-dev makes a commit and pushs, without going through a PR, but also doesn't run the hook. 

This you should simply not do. It took me while to "learn" but now even in my own repos where I'm the only one working on the code I always work in a branch and always go through a PR. Having a pre-commit hook check for this was a major help in teaching myself this bit of discipline.

> We won't be able to accept PRs generated from commits created in github's user interface as the style hook would not have run. 

Do you actually get these ? and what if those happen to not pass your flake8 check in CI ? Do you override the CI or add a manual commit to fix the issue, how is adding a commit to fix the issue different from manually running black and adding that commit to fix the issue ? Sure the black thing *could* be automated (unlike the manual fix for flake8) but there is no *need* to.

> to suddenly change to an entirely different formatting scheme would require (much) more work than the status quo

The beauty of Black is that it's automatic, only extra work is `git add`ing the file again and re-running `git commit` if the hook changed the file. But as Black itself says it's "opinionated", not everyone does not agree with the choices they made. I personally find it very handy having one uniform style everywhere (down to details like single vs double quotes, one of the things I disliked about Black at first).

It's also way easier for new devs to get into the project when there's an autoformatter for the specific flavour
of pep8 the project uses, if they happen to be used to some detail being slightly different they don't need
to learn to do it differently for this particular project and can focus on the actual program logic instead of
formatting details. Your current flake8 config for example disagrees with black on some details, as I'm used to writing
in black style this means I will have to go back and change things that were second nature to me when the flake8 check fails... For those used to the current flavour the autoformatter will fix things automagically and thus no conscious effort required to change their habits.


## non -Black stuff

Now onto the the other hooks configured, how do you feel about them ? I can make another PR where black isn't used but the other checks/fixes (branch checking, whitespace normalization, executables having shebangs) are still done. Maybe that flake8 as well check if it's not "too slow".

There were some whitespace normalizations and a whole bunch of files that should not have been be executable (and some more that maybe could be but then they would have to have shebangs added too), those checks at least are very quick.

Basically all of these checks are like any other automated testing, when they're there the developer can focus on thinking about the code logic instead of possible regressions in small details.

Zen-CODE:
Although I like intention behind this PR, it's WAY too noisy to expect a meaningful code review. Automatic refactors when the changes are entirely cosmetic ("" in place if '') does not really add value. But most has already been said. @rambo  The effort is appreciated and the intention is good, but please consider less extreme approaches. Pragmatism is a virtue...

rambo:
> Pep8 doesn't need to be a hard, enforced rule, it's a style guideline, and that's good like this imho

Yet your CI enforces it (which is good IMHO), just pointing out the contradiction here.

> Automatic refactors when the changes are entirely cosmetic ("" in place if '') does not really add value

Agreed, the value would come from going forward with the autoformatter making sure everyone is always in same style without having to spend conscious effort in flavour details.

anyway no for Black. I would suggest figuring out how to have something (autopep8) that auto-formats to your preferred style anyway. As stated above so that it's way easier for people used to different flavour of pep8 to contribute since a flavour of pep8 is enforced by CI.

Then how about those other things, as seen above: there's a mess of files marked executable that should not be, there's mixed tabs vs spaces usage etc, if I setup pre-commit for those (and flake8) is that going to be wasted effort ? That will again result in "purely cosmetic" changes in a largish commit in the history (though strictly speaking it's not mandatory to do that cleanup at this point unless CI is set up to check for those mistakes as well) but that's why git added the possibility to ignore revisions in `git blame`

meanwhile I'll see what I can do about the side effects of `import kivy` starting with overriding sys.stderr and the root logger.

Zen-CODE:
Hmm.. It's difficult. I really agree in principle, and I like what you are trying to do. Perhaps put your thoughts into a new PR and request me as a reviewer? Thanks.

Zen-CODE:
Or re-open this. There are valid discussions and points here. Up to you.

matham:
Basically, anything that changes existing code on mass for style changes will likely not get accepted. However, optional hooks that fix any PEP8 issues by re-writing, instead of simply complaining about it like our current hook is welcome.

The way to move forward is to have hooks that only re-write code immediately around any changes made in a particular commit. Not to go back in history and make changes to unrelated (of a given PR) parts of the code. 

I frequently edit code and create PRs from GitHub's user interface, so any mandatory hooks that re-write code would just get in the way (especially if a later PR would then have to re-write the changes, even if PEP8 passes, just to adhere to a style). Our current policy of adhering to mostly PEP8 issues, but no major style prefs, are easy to follow without major tooling. And the CI indicates any PEP8 issues in the PR to be fixed manually.

The video linked by @tshirtman is exactly how I view style issues.

matham:
Oh, and about the shebangs, yeah, a PR to fix that should be fine, although is that a problem worth fixing in existing files?

PR #7176, opened 2020-10-23
Add pre-commit.com framework hooks

rambo:
For some basic good practises, see the [framework for more info][precommit] but TLDR is that it handles passing
only files that are in the commit to the hooks and it's very easy to manage hooks.

[precommit]: https://pre-commit.com/

Fixed some executable flags while at it but all the whitespace normalizations and other changes that the hooks will do automatically will wait for whenever someone actually touches code so that we don't have a giant commit doing only those (as discussed in #7170 )


rambo:
ping @Zen-CODE, here's the conservative version.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+default_language_version:
+    python: python3.7
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.4.0
+    hooks:
+    -   id: no-commit-to-branch
```
matham, line 9: I'm not sure about this one. There are sometimes you may commit to master, although it's not frequent I suppose.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+    -   id: check-ast
+    -   id: check-merge-conflict
+    -   id: check-toml
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: check-added-large-files
+    -   id: check-case-conflict
+    -   id: check-json
+    -   id: check-merge-conflict
```
matham, line 20: This is duplicated above.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,35 @@
...
+        args:
+        - --autofix
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.4
+    hooks:
+    -   id: flake8
+-   repo: https://github.com/Lucas-C/pre-commit-hooks
+    rev: v1.1.7
+    hooks:
+    -   id: forbid-crlf
```
matham, line 32: How does this work on Windows where git will auto convert crlf to lf before committing? Would it cause a error?

Makefile
```diff
@@ -96,8 +96,7 @@ style:
 hook:
 	# Install pre-commit git hook to check your changes for styleguide
 	# consistency.
-	cp kivy/tools/pep8checker/pre-commit.githook .git/hooks/pre-commit
-	chmod +x .git/hooks/pre-commit
+	pre-commit install -f
```
matham, line 99:
```suggestion
	pre-commit install -f --config=kivy/tools/pep8checker/.pre-commit-config.yaml
```

Can we please move the config file to `kivy/tools/pep8checker/.pre-commit-config.yaml` so that it's not polluting the root?

matham (changes requested):
You may want to add to the [docs ](https://kivy.org/doc/stable/contribute.html#coding-style), that `--no-verify` or `-n` can be used to temporarily disable the hook.

.pre-commit-config.yaml
rambo: You can always skip a check.Docs should still be updated with pointers to pre-commit.com documentation (because you can also say to skip single check instead of the whole hook).

.pre-commit-config.yaml
rambo: removed duplicate

Makefile
rambo: it's possible sure, but root of repo is the canonical location and the pre-commit framework has an optional post-clone -hook to check for that file and auto-install it.

.pre-commit-config.yaml
rambo: None of my mates at work that are on windows have complained but I'll fire up a VM and doublecheck this personally. That'll probably happen tomorrow though (or maybe much later tonight is I can't sleep or something).

I'm pretty sure the lf<->crlf conversion that git does on windows is done in such way that the hook only sees the actual content going to repo (which would have been converted back to lf)

.pre-commit-config.yaml
matham: Probably both. I'm more likely to remember to use `-n` than the specific syntax to disable a specific hook and having to remember its exact name.

matham: But I guess most of the time commiting to master is not desired so I guess it's fine as long as we doc how to turn off the hook.

Makefile
matham: Yeah, but I still would rather not pollute our project root with more config files than absolutely necessery. They accumulate quickly.

I guess we'll see what the others think.

Zen-CODE: Although root is the standard place,  there are pre-existing standards for kivy. I would 2nd @matham here. In this case,  kivy/tools/pep8checker is a cleaner and more consistent place.

.pre-commit-config.yaml
rambo: the hook name is shown when it fails so remembering it is not a problem :) Also committing/pushing directly to master should be so extraordinary as not to  matter the slightest if it's difficult, way bigger problems have come up if direct pushing to master is *actually needed* to solve an issue.

.pre-commit-config.yaml
rambo: hmm, this is getting way more complicated than just installing git and python for windows, what kind of env are the windows people actually running for development ?? If they're on WSL or cygwin they probably do not have the CRLF conversion even enabled if not how do they even enable the pre-commit hooks (ie where do they get `make` from, for example...)

Anyway gits own documentation says the conversion is done before committing and the hook sees the to-be-committed content so this *shoud* be transparent.

Makefile
rambo: Moved it to a separate directory as is does more than just pep8 and the old checker script should be removed after a deprecation period.

.pre-commit-config.yaml
matham: Personally I use msys and I set `autocrlf=input`.

matham (approved):
Looks good, thanks. It's a pretty innocuous change, but I'll wait until after the upcoming release/tag before merging on the off-chance it causes issues.

PR #7199, opened 2020-11-03
added downloads

Kulothungan16:


matham:
Thank you for the PR, but I have some trepidation about this.

* The downloads are not accurate because it includes all the CI and mirroring. This is a well known issue with PyPI stats and I believe contributed to them removing the stats, which used to be listed on the page IIRC. So it doesn't tell use really how many people downloaded Kivy since pretty much daily there will be many CI services downloading kivy.
* I'm not sure I trust the site. It seems harmless and has been around for a bit, but I don't like the idea someone simply opening our repo (on github or locally) being forced to make a connection to the third party-site, unless it's a much more established website.

But the second issues is a more minor concern, compared to the concern that it doesn't really at much info. But I'm open to other opinions.

matham:
I'm closing as after thinking about this, I still think overall I don't like adding it.

PR #7203, opened 2020-11-06
Tests: Ensure Bubble uses it's superclass's valid private API

gottadiveintopython:
Maintainer merge checklist

## Problem

Bubble currently overwrites some of GridLayout's internal properties: `_fills_row_first`, `_fills_from_left_to_right` and `_fills_from_top_to_bottom`. If any of them are renamed/removed, Bubble's should be renamed/removed as well. The problem is that the unittests don't say anything about this. So if someone only modifies GridLayout's, and forgot to modify Bubble's, Bubble might silently get broken.

 ## What this PR does

This PR lets the unittests check if those properties actually exist in Bubble's super classes.
(I feel like there is a better way to solve the problem, so I put "WIP" in the title.)

### DIFF
--- a/kivy/tests/test_uix_bubble.py
+++ b/kivy/tests/test_uix_bubble.py
@@ -6,7 +6,7 @@ import pytest
     '_fills_from_left_to_right',
     '_fills_from_top_to_bottom',
 ))
-def test_a_certain_properties_are_actually_overwritten(prop_name):
+def test_a_certain_properties_from_the_super_class_are_overwritten(prop_name):
     from kivy.uix.bubble import Bubble
     from kivy.uix.gridlayout import GridLayout
     assert issubclass(Bubble, GridLayout)
--- a/kivy/tests/test_uix_bubble.py
+++ b/kivy/tests/test_uix_bubble.py
@@ -1,6 +1,17 @@
 import pytest
 
 
+@pytest.mark.parametrize('prop_name', (
+    '_fills_row_first',
+    '_fills_from_left_to_right',
+    '_fills_from_top_to_bottom',
+))
+def test_a_certain_properties_exist_in_any_of_the_super_classes(prop_name):
+    from kivy.uix.bubble import Bubble
+    super_classes = Bubble.mro()[1:]  # exclude Bubble itself
+    assert any(hasattr(klass, prop_name) for klass in super_classes)
+
+
 @pytest.mark.parametrize('orientation', ('vertical', 'horizontal'))
 def test_always_lr_tb(orientation):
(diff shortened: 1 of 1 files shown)


kivy/tests/test_uix_bubble.py
```diff
@@ -1,6 +1,17 @@
...
 
+@pytest.mark.parametrize('prop_name', (
+    '_fills_row_first',
+    '_fills_from_left_to_right',
+    '_fills_from_top_to_bottom',
+))
+def test_a_certain_properties_exist_in_any_of_the_super_classes(prop_name):
+    from kivy.uix.bubble import Bubble
+    super_classes = Bubble.mro()[1:]  # exclude Bubble itself
+    assert any(hasattr(klass, prop_name) for klass in super_classes)
```
matham, line 12: I'd probably test directly that GridLayout has these properties, rather than super classes.

matham (changes requested):

kivy/tests/test_uix_bubble.py
gottadiveintopython: Thanks, I fixed. I thought the previous code is more robust against changes.

matham: It is perhaps more robust to changes in super class, but is also too clever :) and perhaps too flexible. So my worry was that there are more ways for it to go wrong than checking grid layout directly. After all the worry is mainly someone changing grid layout and forgetting about bubble?

matham (dismissed):

kivy/tests/test_uix_bubble.py
gottadiveintopython: > After all the worry is mainly someone changing grid layout and forgetting about bubble?

Yes.

> So my worry was that there are more ways for it to go wrong than checking grid layout directly

Yes, there might be. Since I don't come up with any good idea than the current and the previous code, I'm ok with the current code.

gottadiveintopython: Actually putting `issubclass(Bubble, GridLayout)` in the unittests might remove my concern.

matham (approved):

gottadiveintopython:
It may be too late, but I guess you think `issubclass(Bubble, GridLayout)` is unnecessary because we'll never replace the superclass with another, as it breaks the api. If so, you may be right.

Also, I'll fix the error on unittest soon. (`test_animation.py`)

matham:
>we'll never replace the superclass with another, as it breaks the api.

Exactly.

>Also, I'll fix the error on unittest soon. (test_animation.py)

👍

PR #7209, opened 2020-11-09
Animation: Allow canceling all animated widgets

gottadiveintopython:
Maintainer merge checklist

## description

Since Kivy doesn't provide any virtual clock, all `Animation`s are affected by real-time, which makes it harder to test. As a result, `test_animations.py` sometimes succeeds sometimes fails. This PR tries to stabilize it as possible.

### DIFF
--- a/kivy/tests/test_animations.py
+++ b/kivy/tests/test_animations.py
@@ -38,10 +38,7 @@ def ec_cls():
 @pytest.fixture(autouse=True)
 def cleanup():
     from kivy.animation import Animation
-    for anim in Animation._instances:
-        anim._widgets.clear()
-        anim._clock_uninstall()
-    Animation._instances.clear()
+    Animation.cancel_all(None)
 
 
 def no_animations_being_played():
@@ -78,6 +75,38 @@ class TestAnimation:
         w = Widget()
         a.start(w)
         sleep(.5)
+        assert no_animations_being_played()
+
+    def test_cancel_all(self):
+        from kivy.animation import Animation
+        from kivy.uix.widget import Widget
+        a1 = Animation(x=100)
+        a2 = Animation(y=100)
+        w1 = Widget()
+        w2 = Widget()
+        a1.start(w1)
+        a1.start(w2)
+        a2.start(w1)
+        a2.start(w2)
+        assert not no_animations_being_played()
+        Animation.cancel_all(None)
+        assert no_animations_being_played()
+
+    def test_cancel_all_2(self):
+        from kivy.animation import Animation
+        from kivy.uix.widget import Widget
+        a1 = Animation(x=100)
+        a2 = Animation(y=100)
+        w1 = Widget()
+        w2 = Widget()
+        a1.start(w1)
+        a1.start(w2)
+        a2.start(w1)
+        a2.start(w2)
+        assert not no_animations_being_played()
+        Animation.cancel_all(None, 'x')
+        assert not no_animations_being_played()
+        Animation.cancel_all(None, 'y')
         assert no_animations_being_played()
 
     def test_stop_animation(self):
--- a/kivy/tests/test_animations.py
+++ b/kivy/tests/test_animations.py
@@ -38,8 +38,9 @@ def ec_cls():
 @pytest.fixture(autouse=True)
 def cleanup():
     from kivy.animation import Animation
-    for anim in Animation._instances.copy():
-        anim.cancel()
+    for anim in Animation._instances:
+        anim._widgets.clear()
+        anim._clock_uninstall()
     Animation._instances.clear()
 
 
--- a/kivy/tests/test_animations.py
+++ b/kivy/tests/test_animations.py
@@ -59,8 +59,8 @@ def sleep(t):
     from time import time, sleep
     from kivy.clock import Clock
     tick = Clock.tick
-    start = time()
-    while time() < start + t:
+    deadline = time() + t
+    while time() < deadline:
         sleep(.01)
         tick()
 
--- a/kivy/tests/test_animations.py
+++ b/kivy/tests/test_animations.py
@@ -104,7 +104,7 @@ class TestAnimation:
         a2.start(w1)
         a2.start(w2)
         assert not no_animations_being_played()
-        Animation.cancel_all(None, 'x')
+        Animation.cancel_all(None, 'x', 'z')
         assert not no_animations_being_played()
         Animation.cancel_all(None, 'y')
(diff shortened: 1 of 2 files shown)


kivy/tests/test_animations.py
```diff
@@ -35,11 +35,18 @@ def assert_(self, n_start, n_progress_greater_than_zero, n_complete):
     return EventCounter
 
 
+@pytest.fixture(scope='module')
+def approx():
+    from functools import partial
+    return partial(pytest.approx, abs=1)
```
matham, line 41: Hmm, do we need to make this a fixture rather than simply a global function? But also, why 1? Doesn't the tests use `sleep` to make sure we waited long enough, or there a problem with it?

The canonical way would be to mock the clock, but I'm confused why there would be any timing issues, given that we wait for a given time.

kivy/tests/test_animations.py
```diff
@@ -35,11 +35,18 @@ def assert_(self, n_start, n_progress_greater_than_zero, n_complete):
...
+
+
 @pytest.fixture(autouse=True)
 def cleanup():
     from kivy.animation import Animation
-    for anim in Animation._instances.copy():
-        anim.cancel()
+    for anim in Animation._instances:
```
matham, line 47: It is concerning if `cancel` doesn't work. Shouldn't `cancel` work?

kivy/tests/test_animations.py
gottadiveintopython: > do we need to make this a fixture rather than simply a global function?

The reason I made it a fixture is to avoid importing a module on top-level. (`from functools import partial`). I was told to not do it, as it might have side effects to tests.

> But also, why 1?

There is no reason. I'd appreciate if you tell me what the appropriate value is.

> why there would be any timing issues, given that we wait for a given time.

I have no idea. It's really strange. For example the error in [#7203](https://github.com/kivy/kivy/runs/1366986755):

```
        # during the second half of the second round of the animation
        ec.assert_(1, True, 0)
>       ec1.assert_(2, True, 2)

kivy/tests/test_animations.py:306: 

self = <kivy.tests.test_animations.ec_cls.<locals>.EventCounter object at 0x109d527f0>
n_start = 2, n_progress_greater_than_zero = True, n_complete = 2

    def assert_(self, n_start, n_progress_greater_than_zero, n_complete):
        assert self.n_start == n_start
        if n_progress_greater_than_zero:
            assert self.n_progress > 0
        else:
            assert self.n_progress == 0
>       assert self.n_complete == n_complete
E       assert 1 == 2
E        +  where 1 = <kivy.tests.test_animations.ec_cls.<locals>.EventCounter object at 0x109d527f0>.n_complete
```

this error means the progress of an animation was *behind* than the expectation. If it were *ahead*, I wouldn't be surprised. But it were "behind" for some reason.

kivy/tests/test_animations.py
gottadiveintopython: This is a pure bug because `Animation.cancel()` takes one positional argument. The reason I didn't notice it at the time is `anim.cancel()` is excuted only when a test case fails, which didn't happen when I wrote it.

gottadiveintopython:
@matham Do you think the following fixture stabilizes the tests?

```python
@pytest.fixture(autouse=True)
def reset_clock():
    from kivy.clock import Clock
    Clock.tick()
```

The reason I asked this is the following code.

```python
from kivy.clock import Clock
from time import sleep

Clock.tick()
sleep(.5)
Clock.schedule_once(print, 1)  # A
sleep(.7)
Clock.tick()  # B
print('----')
sleep(.5)
Clock.tick()
print('end')
```

```
1.202135275118053
----
end
```

When the line B is excuted, `print()` is called even though only 0.7s passed from the line A, because 1.2s(0.5 + 0.7) passed from the first `Clock.tick()`. So I feel like it's important to call `Clock.tick()` before Animation starts.

kivy/tests/test_animations.py
gottadiveintopython: @matham Don't you think `pytest.approx()` by default is too tight?

```python
assert 100 == pytest.approx(100.5)  # => False
```

This is why I give it `abs=1`

matham: >The reason I made it a fixture is to avoid importing a module on top-level. (from functools import partial). I was told to not do it, as it might have side effects to tests.

I did mention this before, but only for kivy modules that have side effects like import kivy logger or widget etc. But it's not a problem for normal python imports which should happen at the top.

>Don't you think pytest.approx() by default is too tight?

Hmm, perhaps, but that feels like a band-aid without fixing the underlying problem. We should first know why 1 sec is too tight, because presumably whoever wrote it didn't think it should be an issue. And once we know why it may be too tight, then we can perhaps increase the interval.

kivy/tests/test_animations.py
matham: Ok, but I think a better solution is to allow `cancel_all` https://github.com/kivy/kivy/blob/master/kivy/animation.py#L183 to take `None` for the `widget` parameter and then clear all the widgets. Because it's better to not access the private API outside `Animation`.

matham (changes requested):

kivy/tests/test_animations.py
```diff
@@ -52,22 +59,22 @@ def sleep(t):
     from time import time, sleep
     from kivy.clock import Clock
     tick = Clock.tick
-    start = time()
```
matham, line 55: I think we need to put a `tick()` before any call to `time`. This may perhaps be the cause of the bug.

matham (changes requested):

matham:
>Do you think the following fixture stabilizes the tests?

I think my comment on the `sleep` method changes may fix the issue you're thinking about.

kivy/tests/test_animations.py
gottadiveintopython: ok I'll remove the fixture for now. If its tightness becomes a problem, I might look into it.

matham:
Looks good. Just need to add `versionchanged` about `cancel_all` accepting `None`.

gottadiveintopython:
ok did it, but I believe I shouldn't write any doc.

matham (dismissed):

matham:
>ok did it, but I believe I shouldn't write any doc. I mean I can't express what I want since I'm not fluent in English.

I think your English is fine with regards to the docs. It doesn't have to be perfect for users to understand what is being said.

matham (approved):

PR #7213, opened 2020-11-12
ScrollView: Match scroll effect stop condition to start condition.

### LINKED ISSUE #7207: ScrollView Scroll Throws Bug IndexError
**Software Versions**
* Python: 3.7.6
* OS:
* Kivy: 2.0.0rc4
* Kivy installation method:

**Describe the bug**
This was NOT a Bug in Kivy 2.0.0rc3 but appeared in Kivy 2.0.0rc4.  I have created a List of items with ScrollView.  I had the option to scroll my list with either the slider or touching the list with your finger.
self.FoodList.scroll_type = ['bars','content']
But this will throw an error the moment you touch the Slider Bar and say:
IndexError: List Index out of Range.

if you just make the scroll_type
self.FoodList.scroll_type = ['bars']
bars ONLY without the content part, it works and does not throw an error.
This used to work with both types but now doesn't.

The part of the code that throws the error is:
/kivy/uix/scrollview.py   line 903
if self.scroll_type != ['bars']:

The issue tracker is a tool to address **bugs only** (search known bugs: https://git.io/vM1iE).
Please use the #support Discord channel at https://chat.kivy.org/ or Stack Overflow for
support questions, more information at https://git.io/vM1yQ.

**Expected behavior**
I used to be able to scroll this list on the screen either by sliding the slider bar and/or just touching the list anywhere inside the list and scrolling it.

**To Reproduce**
A short, *runnable* example that reproduces the issue with *latest kivy master*.

```python
// REPLACE ME: Paste your code and logs here, a template can be found at https://git.io/vM1id.
```

**Additional context**

matham:
Maintainer merge checklist

Fixes #7207.

Currently the stop condition is such that the effect stop may be executed even if the effect start was not called. Because the start and stop were guarded under different conditions. This makes them equal.

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -973,14 +973,12 @@ class ScrollView(StencilView):
         self._touch = None
         uid = self._get_uid()
         ud = touch.ud[uid]
-        if self.do_scroll_x and self.effect_x:
-            if not touch.ud.get('in_bar_x', False) and\
-                    self.scroll_type != ['bars']:
-                self.effect_x.stop(touch.x)
-        if self.do_scroll_y and self.effect_y and\
-                self.scroll_type != ['bars']:
-            if not touch.ud.get('in_bar_y', False):
-                self.effect_y.stop(touch.y)
+        not_in_bar = not touch.ud.get('in_bar_x', False) and \
+            not touch.ud.get('in_bar_y', False)
+        if self.do_scroll_x and self.effect_x and not_in_bar:
+            self.effect_x.stop(touch.x)
+        if self.do_scroll_y and self.effect_y and not_in_bar:
+            self.effect_y.stop(touch.y)
         if ud['mode'] == 'unknown':
             # we must do the click at least..
             # only send the click if it was not a click to stop
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -895,6 +895,9 @@ class ScrollView(StencilView):
                 ud['mode'] = 'scroll'
 
         if ud['mode'] == 'scroll':
+            not_in_bar = not touch.ud.get('in_bar_x', False) and \
+                not touch.ud.get('in_bar_y', False)
+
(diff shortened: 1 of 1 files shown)


tshirtman:
Doesn't seem to fix #7207 for me, if i drag the bar first, without touching anything before that, it crashes 100% of the time.

tshirtman:
Probably not the prettiest solution, but 
```diff
--- a/kivy/effects/kinetic.py
+++ b/kivy/effects/kinetic.py
@@ -145,7 +145,7 @@ def update(self, val, t=None):
         See :meth:`start` for the arguments.
         '''
         t = t or time()
-        distance = val - self.history[-1][1]
+        distance = val - (self.history[-1][1] if self.history else 0)
         self.apply_distance(distance)
         self.history.append((t, val))
         if len(self.history) > self.max_history:
diff --git a/kivy/effects/scroll.py b/kivy/effects/scroll.py
index 7a722b8d7..f35b5c39b 100644
--- a/kivy/effects/scroll.py
+++ b/kivy/effects/scroll.py
@@ -113,7 +113,7 @@ def start(self, val, t=None):
         return super(ScrollEffect, self).start(val, t)
 
     def update(self, val, t=None):
-        self.displacement += abs(val - self.history[-1][1])
+        self.displacement += abs(val - (self.history[-1][1] if self.history else 0))
         return super(ScrollEffect, self).update(val, t)
 
     def stop(self, val, t=None):
```
seems to solve it for me without side effects.

matham:
I overlooked that crash because when I translated from kivymd I didn't fully match the kivymd rules. But it should now be fixed for scroll move as well.

matham:
The history fix would definitely fix it, but it doesn't fix the underlying issue.

tshirtman:
Right, but unless we have time to properly fix (and i can't promise i will) i would rather have that.

matham:
Oh, yeah, my recent commit should have fixed it completely.

PR #6716, opened 2020-01-28
[WIP] Fix input type on android using SDL2 as window provider

Franccisco:
This pull request allows the use of `input_type` and `keyboard_suggestions` properties to modify the keyboard displayed on Android. In summary, it uses the `changeKeyboard` function in PythonActivity.java (see kivy/python-for-android#2057). Additionally, `keyboard_suggestions` was moved from kivy/uix/textinput.py to kivy/uix/behaviors/focus.py, so it's more accesible to send to the  `request_keyboard` method.

This is ready to review. Note that you may encounter some bugs with these changes:
- Inserting any emoji will crash the app in older Android versions, and a tofu will appear in latest versions.
- Keyboard suggestions have a weird behavior. You won't see anything you type until you select the suggestion.

The details are in (kivy/python-for-android#2015).  Unless those bugs are solved, this PR shouldn't be merged.

PR #7231, opened 2020-11-27
Keyboard: Add keyboard suggestions and fix input type on android

dwmoffatt:
This PR gives kivy the ability to set keyboard input type correctly on android. 

Emojis do not work as the default font doesn't support them. It also creates some oddities when using input filtering and emojis are entered.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/behaviors/focus.py
+++ b/kivy/uix/behaviors/focus.py
@@ -237,6 +237,15 @@ class FocusBehavior(object):
         2.0.0.
     '''
 
+    keyboard_suggestions = BooleanProperty(True)
+    '''If True provides auto suggestions on top of keyboard.
+    This will only work if :attr:`input_type` is set to `text`, `url`, `mail` or
+    `address`.
+    .. versionadded:: 2.0.0
+    :attr:`keyboard_suggestions` is a :class:`~kivy.properties.BooleanProperty`
+    and defaults to True
+    '''
+
     def _set_on_focus_next(self, instance, value):
         ''' If changing code, ensure following code is not infinite loop:
         widget.focus_next = widget
@@ -388,9 +397,12 @@ class FocusBehavior(object):
     def _ensure_keyboard(self):
         if self._keyboard is None:
             self._requested_keyboard = True
-            keyboard = self._keyboard =\
-                EventLoop.window.request_keyboard(
-                    self._keyboard_released, self, input_type=self.input_type)
+            keyboard = self._keyboard = EventLoop.window.request_keyboard(
+                self._keyboard_released,
+                self,
+                input_type=self.input_type,
+                keyboard_suggestions=self.keyboard_suggestions,
+            )
             keyboards = FocusBehavior._keyboards
             if keyboard not in keyboards:
                 keyboards[keyboard] = None
--- a/kivy/uix/behaviors/focus.py
+++ b/kivy/uix/behaviors/focus.py
@@ -241,9 +241,9 @@ class FocusBehavior(object):
     '''If True provides auto suggestions on top of keyboard.
     This will only work if :attr:`input_type` is set to `text`, `url`, `mail` or
     `address`.
-    
+
     .. versionadded:: 2.1.0
-    
+
     :attr:`keyboard_suggestions` is a :class:`~kivy.properties.BooleanProperty`
     and defaults to True
     '''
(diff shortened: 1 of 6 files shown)


gottadiveintopython:
Have you seen #6716 and https://github.com/kivy/python-for-android/pull/2057? I'd recommend you to talk to @Franccisco if you want to work on this kind of feature.

dwmoffatt:
I have seen both of those PRs. I commented on one of them and never heard back from @Franccisco .This implementation is his but finished via patching the SDL Activity. I have tried it and the only issue I have with it relates to emojis, which cause some oddities. 

I am going to look into a solution for that as well.

shashi278:
@dwmoffatt But there's a known issue with `TYPE_CLASS_TEXT` that is, textinput doesn't show text as you type. It only shows when you tap on a suggestion. So just wanted to know if this PR fixes that?

dwmoffatt:
@shashi278 I did not notice that when testing it. What you are speaking of seemed to be related to a missing on_textedit which is now part of the textInput class.

shashi278:
@dwmoffatt My bad. Yes it's already there. But you know, that still doesn't work unless to change `inputType` inside `SDLActivity.java`. Will that be fixed in this one?

dwmoffatt:
@shashi278 Not in this PR, as it requires patching SDLActivity. There is a python-for-android PR for that fix. https://github.com/kivy/python-for-android/pull/2361

JuditSN:
Is already merged? @tshirtman

tshirtman:
> Is already merged? @tshirtman

No, neither PR are merged at the moment, as indicated by the green "open" label at the top of the page.

dwmoffatt:
@tshirtman @matham 

the python-for-android or that is associated with this one is merged. Is there anything needed for this PR?

kivy/uix/behaviors/focus.py
```diff
@@ -237,6 +237,15 @@ def _get_keyboard(self):
         2.0.0.
     '''
 
+    keyboard_suggestions = BooleanProperty(True)
+    '''If True provides auto suggestions on top of keyboard.
+    This will only work if :attr:`input_type` is set to `text`, `url`, `mail` or
+    `address`.
+    .. versionadded:: 2.0.0
```
matham, line 244:
```suggestion

    .. versionadded:: 2.0.0

```

matham (changes requested):
Also, please update `request_keyboard` in `window_pygame.py` (even though we don't really support pygame anymore)

kivy/uix/behaviors/focus.py
dwmoffatt: Sounds good. I will do it today. Thanks @matham for getting back to me quickly.

matham: Oh, and I just noticed, it should be 2.1.0, not 2.0.0.

dwmoffatt: @matham regarding the request_keyboard in window_pygame.py, what do you want to me change regarding it? The android.show_keyboard looks to handle keyboard suggestions and input type already.

matham: Just add the new keyboard to `request_keyboard` here https://github.com/kivy/kivy/blob/master/kivy/core/window/window_pygame.py#L435. It doesn't have to do anything with, just pass it on to super I guess. Because I think right now it would crash with the unexpected keyword.

kivy/uix/behaviors/focus.py
```diff
@@ -237,6 +237,15 @@ def _get_keyboard(self):
         2.0.0.
     '''
 
+    keyboard_suggestions = BooleanProperty(True)
+    '''If True provides auto suggestions on top of keyboard.
+    This will only work if :attr:`input_type` is set to `text`, `url`, `mail` or
+    `address`.
+    .. versionadded:: 2.1.0
```
matham, line 244:
```suggestion

    .. versionadded:: 2.1.0

```

There needs to be blank lines around this, I believe.

matham (changes requested):

kivy/uix/behaviors/focus.py
dwmoffatt: fixed it.

matham (approved):

PR #7245, opened 2020-12-07
Resources: Add a cache for resource_find

sanderland:
Before:
resource_find dominates render time, causing noticable lag in rendering. 

image

Caching in user code, always passing an absolute path, the large amount of abspath calls still take significant time, but don't dominate.

image

After this PR, no file calls to be found and rendering is smooth.
image

Maintainer merge checklist

### DIFF
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -296,7 +296,9 @@ class BuilderBase(object):
 
             `encoding`: File charcter encoding. Defaults to utf-8,
         '''
-        filename = resource_find(filename) or filename
+
+        # Cache here can give problems since Clock sometimes does not exist.
+        filename = resource_find(filename, use_cache=False) or filename
         if __debug__:
             trace('Lang: load file %s, using %s encoding', filename, encoding)
 
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -297,8 +297,7 @@ class BuilderBase(object):
             `encoding`: File charcter encoding. Defaults to utf-8,
         '''
 
-        # Cache here can give problems since Clock sometimes does not exist.
-        filename = resource_find(filename, use_cache=False) or filename
+        filename = resource_find(filename) or filename
         if __debug__:
             trace('Lang: load file %s, using %s encoding', filename, encoding)
 
--- a/kivy/resources.py
+++ b/kivy/resources.py
@@ -32,6 +32,7 @@ __all__ = ('resource_find', 'resource_add_path', 'resource_remove_path')
 
 from os.path import join, dirname, exists, abspath
 from kivy import kivy_data_dir
+from kivy.cache import Cache
 from kivy.utils import platform
 from kivy.logger import Logger
 import sys
@@ -42,23 +43,39 @@ if platform == 'ios':
     resource_paths += [join(dirname(sys.argv[0]), 'YourApp')]
 resource_paths += [dirname(kivy.__file__), join(kivy_data_dir, '..')]
 
+Cache.register('kv.resourcefind', timeout=60)
 
-def resource_find(filename):
+
+def resource_find(filename, use_cache=True):
     '''Search for a resource in the list of paths.
     Use resource_add_path to add a custom path to the search.
+    By default, results are cached for 60 seconds.
+    This can be disabled using use_cache=False.
     '''
     if not filename:
         return
+    found_filename = None
+    if use_cache:
+        found_filename = Cache.get('kv.resourcefind', filename)
+        if found_filename:
+            return found_filename
     if filename[:8] == 'atlas://':
-        return filename
-    if exists(abspath(filename)):
-        return abspath(filename)
-    for path in reversed(resource_paths):
-        output = abspath(join(path, filename))
-        if exists(output):
-            return output
-    if filename[:5] == 'data:':
-        return filename
+        found_filename = filename
+    else:
+        abspath_filename = abspath(filename)
+        if exists(abspath_filename):
+            found_filename = abspath(filename)
+        else:
+            for path in reversed(resource_paths):
+                output = abspath(join(path, filename))
+                if exists(output):
+                    found_filename = output
+                    break
+        if not found_filename and filename.startswith("data:"):
+            found_filename = filename
+    if use_cache:
+        Cache.append('kv.resourcefind', filename, found_filename)
+    return found_filename
 
 
 def resource_add_path(path):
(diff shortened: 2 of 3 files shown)


kivy/resources.py
```diff
@@ -42,23 +43,41 @@
...
+Cache.register('kv.resourcefind', timeout=60)
 
-def resource_find(filename):
+
+def resource_find(filename, use_cache=True):
     '''Search for a resource in the list of paths.
     Use resource_add_path to add a custom path to the search.
+    By default, results are cached for 60 seconds.
+    This can be disabled using use_cache=False.
+    .. versionchanged:: 2.0.0rc5
```
tshirtman, line 54:
```suggestion
    .. versionchanged:: 2.0.0
```

kivy/resources.py
```diff
@@ -42,23 +43,41 @@
...
+        A default cache and the `use_cache` parameter were added.
```
tshirtman, line 55:
```suggestion
        `use_cache` parameter added and made True by default.
```

tshirtman (approved):
LGTM

kivy/resources.py
matham: Just a reminder that this should not be merged in 2.0.0 as we're in a freeze until the release. Especially something with significant changes as this.

sanderland:
Alright, will let you figure out where and when to merge it, let me know if you need anything from me.

For anyone else having this issue in the mean time, I found that using `texture=cached_texture(path)` instead of `source=path` on canvas instructions eliminates this problem completely.
```
# direct cache to texture, bypassing resource_find
def cached_texture(path,_cache={}):
    tex = _cache.get(path)
    if not tex:
        tex = _cache[path] = Image(resource_find(path)).texture
    return tex
```

kivy/resources.py
matham: The current version is now `2.1.0`.

sanderland, line 54:
```suggestion
     .. versionchanged:: 2.1.0
```

sanderland: Fixed

kivy/lang/builder.py
```diff
@@ -296,7 +296,9 @@ def load_file(self, filename, encoding='utf8', **kwargs):
 
             `encoding`: File charcter encoding. Defaults to utf-8,
         '''
-        filename = resource_find(filename) or filename
+
+        # Cache here can give problems since Clock sometimes does not exist.
```
matham, line 300: Why is the presence of Clock an issue? It doesn't seem to be used in `resource_find`.

kivy/resources.py
```diff
@@ -42,23 +43,41 @@
...
+Cache.register('kv.resourcefind', timeout=60)
 
-def resource_find(filename):
+
+def resource_find(filename, use_cache=True):
     '''Search for a resource in the list of paths.
     Use resource_add_path to add a custom path to the search.
+    By default, results are cached for 60 seconds.
+    This can be disabled using use_cache=False.
+    .. versionchanged:: 2.1.0
```
matham, line 54:
```suggestion

    .. versionchanged:: 2.1.0
```

kivy/lang/builder.py
sanderland: Cache uses Clock

matham: Are you worried that the timeout won't happen because the clock is not running? Because the clock should be available whether it is running or not so this shouldn't cause issues beyond the timeout not happening.

kivy/resources.py
```diff
@@ -42,23 +43,42 @@
...
+            found_filename = abspath(filename)
+        else:
+            for path in reversed(resource_paths):
+                abspath_filename = abspath(join(path, filename))
+                if exists(abspath_filename):
+                    found_filename = abspath_filename
+                    break
+        if not found_filename and filename.startswith("data:"):
+            found_filename = filename
+    if use_cache:
```
matham, line 79:
```suggestion
    if use_cache and found_filename:
```

Because this can happen e.g. when `not found_filename and not filename.startswith("data:")`.

matham (changes requested):
Would you be able to add unit tests for this? I won't hold up the PR if you can't, but I'd very much recommend it.

kivy/lang/builder.py
sanderland: https://github.com/kivy/kivy/runs/1513086942

This actually happens in building docs, very strange. I've reverted it so you can investigate, I am unable to reproduce it on windows.
```
'lastaccess': Clock.get_time(),
AttributeError: 'NoneType' object has no attribute 'get_time'
```

kivy/resources.py
sanderland: fixed, and will look into tests

sanderland: tests added

kivy/resources.py
```diff
@@ -32,6 +32,7 @@
 
 from os.path import join, dirname, exists, abspath
```
matham, line 33:
```suggestion
from os.path import join, dirname, exists, abspath, environ
```

kivy/resources.py
```diff
@@ -42,23 +43,42 @@
     resource_paths += [join(dirname(sys.argv[0]), 'YourApp')]
 resource_paths += [dirname(kivy.__file__), join(kivy_data_dir, '..')]
 
+Cache.register('kv.resourcefind', timeout=60)
 
-def resource_find(filename):
+
+def resource_find(filename, use_cache=True):
```
matham, line 49:
```suggestion
def resource_find(filename, use_cache=environ.get('KIVY_DOC_INCLUDE', None) != '1'):
```

kivy/lang/builder.py
matham: It's because of this: https://github.com/kivy/kivy/blob/master/kivy/clock.py#L1155. The `KIVY_DOC_INCLUDE` is required to be able to generate rtd docs because we can't create a window there.

Long term the solution is to remove `KIVY_DOC_INCLUDE` and make kivy imports less aggressive so running on rtd will work. But for now, I guess for now we can do the hacky thing and disable cache in this environment as in my suggested edits.

kivy/resources.py
sanderland: added suggestions in a way to mirror clock and imported os.environ rather than os.path.environ

kivy/lang/builder.py
sanderland: I see. I've had my own tests fail on occasion as well, usually due to sp() or something making a window

kivy/resources.py
matham: Technically, the way clock does it is wrong, because we need to check the actual value, not just that it has been defined. E.g. someone could have `KIVY_DOC_INCLUDE=0` rather than deleting the variable from their env. But it's fine since hopefully we'll remove it anyway in the not so distant future.

matham (approved):

PR #7260, opened 2020-12-11
Use NSString instead of AVCaptureSessionPreset in order to support MacOS < 10.13

misl6:
As discussed with @matham on discord, in order to support MacOS 10.12 (which is the current  `MACOSX_DEPLOYMENT_TARGET` for `conda`), We decided to use `NSString` instead of `AVCaptureSessionPreset`.

The `AVCaptureSessionPreset` constants were only introduced since MacOS SDK 10.13.

Maintainer merge checklist


matham (approved):
Indeed, the patch seems to fix https://github.com/conda-forge/kivy-feedstock/pull/26.

PR #7262, opened 2020-12-12
RecycleGridLayout : Fix layout when number of widgets match number of columns

### LINKED ISSUE #7255: RecycleGridLayout not able to render viewclass when number of columns is equal to the length of data
**Software Versions**
* Python: 3.8.3
* OS: Windows 10 Home
* Kivy: 2.0.0
* Kivy installation method: installed kivy base through pip

Description:
    The recycle grid layout has an issue in rendering the viewclass objects when the number of columns are equal to the number of viewclass objects, to demonstrate this there is a demo below.

**Expected behavior**
It is supposed to display the objects of the view class in a grid, but here they disappear and only the first object is visible on-screen

**To Reproduce**
To reproduce this bug just make a recycle grid layout in a recycle view with the same number of columns as the number of the objects of the view class you want to add.

```python
from kivy.app import App
from kivy.lang import Builder
kv = '''
RecycleView:
    data: [{"text": "Label "+str(x)} for x in range(5)]
    viewclass: 'Button'
    RecycleGridLayout:
        cols: 5
        default_size: None, 200
        default_size_hint: 1, None
'''
class MyApp(App):
    def build(self):
MyApp().run()
```
Screenshot (401)

**Additional context**
This works in kivy 1.11.1, also this behavior is not present in the regular grid layout.

gottadiveintopython:
Maintainer merge checklist

fixes #7255 

I put "WIP" in the title because I feel like there is a better implemetation.

### DIFF
--- a/kivy/uix/recyclegridlayout.py
+++ b/kivy/uix/recyclegridlayout.py
@@ -221,13 +221,20 @@ class RecycleGridLayout(RecycleLayout, GridLayout):
         top = y + h
         at_idx = self.get_view_index_at
         # 'tl' is not actually 'top-left' unless 'orientation' is 'lr-tb'.
-        # But we can pretend it always is. Same for 'bl' and 'br'.
-        tl, __, bl, br = sorted((
-            at_idx((x, y)),
-            at_idx((right, y)),
-            at_idx((x, top)),
-            at_idx((right, top)),
-        ))
+        # But we can pretend it always is. Same for 'tr', 'bl' and 'br'.
+        tl = at_idx((x, top))
+        tr = at_idx((right, top))
+        bl = at_idx((x, y))
+        br = at_idx((right, y))
+        cond1 = not self._fills_from_top_to_bottom
+        cond2 = not self._fills_from_left_to_right
+        if not self._fills_row_first:
+            tr, bl = bl, tr
+            cond1, cond2 = cond2, cond1
+        if cond1:
+            tl, tr, bl, br = bl, br, tl, tr
+        if cond2:
+            tl, tr, bl, br = tr, tl, br, bl
 
         n = len(data)
         indices = []
--- a/kivy/uix/recyclegridlayout.py
+++ b/kivy/uix/recyclegridlayout.py
@@ -220,12 +220,17 @@ class RecycleGridLayout(RecycleLayout, GridLayout):
         right = x + w
         top = y + h
         at_idx = self.get_view_index_at
-        # 'tl' is not actually 'top-left' unless 'orientation' is 'lr-tb'.
-        # But we can pretend it always is. Same for 'tr', 'bl' and 'br'.
         tl = at_idx((x, top))
         tr = at_idx((right, top))
         bl = at_idx((x, y))
         br = at_idx((right, y))
+
+        # What is doing here is that re-ordering the view-indices above so
+        # that the last part of the method doesn't have to care about the
+        # 'orientation' so much.
+        # When this part is done, 'tl' will not actually be 'top-left' unless
+        # the 'orientation' is either of 'lr-tb' or 'tb-lr'. Similar thing can
+        # be said to 'tr', 'bl' and 'br'.
         cond1 = not self._fills_from_top_to_bottom
         cond2 = not self._fills_from_left_to_right
         if not self._fills_row_first:
--- a/kivy/uix/recyclegridlayout.py
+++ b/kivy/uix/recyclegridlayout.py
@@ -220,28 +220,16 @@ class RecycleGridLayout(RecycleLayout, GridLayout):
         right = x + w
         top = y + h
         at_idx = self.get_view_index_at
-        tl = at_idx((x, top))
-        tr = at_idx((right, top))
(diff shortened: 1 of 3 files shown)


tshirtman:
The parametrised test seems to repeut the implementation, i think it's better to be less smart in such case and be explicit about tho expected result, each parameter could be a tuple of (input, exptected_output) instead.

matham:
>The parametrised test seems to repeat the implementation, i think it's better to be less smart in such cases and be explicit about tho expected result, each parameter could be a tuple of (input, exptected_output) instead.

Agree. But also, this is not testing the implementation directly, but rather an analog of the implementation. So if one changes the implementation, the test would not be helpful anymore. So it's better to test the actual implementation and use hard-coded input/output values as @tshirtman suggested.

matham:
I think I wasn't clear enough with what I was trying to say, but could you test an actual gridlayout widget? Beause this basically tests that the current implementation is current.  But if someone changes the gridlayout in the future, the test will not test that.

gottadiveintopython:
@tshirtman Ah sorry for my explanation being not enough. There is a situation where we cannot just sort the view indices in ascending order as the latest commit shows.

@matham No, you were clear enough. I pushed the latest commit *before* I read your comment. 

> it's better to test the actual implementation

OK, so is it ok to create a new function that implements that specific part, and use it in the unittests? or I better to use `compute_visible_views()` itself in the unittests somehow?

gottadiveintopython:
When I implemented the 'orientation', I kept the original code as possible in order to avoid introducing new bugs. But maybe this `compute_visible_views()` should be re-written entirely (if I come up with good implementation).

matham:
Yeah, I'd say to call `compute_visible_views` directly and ensure the it returns them in the correct order.

Or even better, create a actual recyclegridlayout (and normal gridlayout) and set the data/children, run `do_layout` or whatever is required to do the layout, and then check that the children widgets are at the expected positions. But `compute_visible_views` is enough if you don't want to do that.

matham:
>But maybe this compute_visible_views() should be re-written entirely (if I come up with good implementation) to avoid this complicated unittests.

Ultimately, it'd be nice if we can test that the widgets are laid out correctly, irrespective of the implementation. So I'm not sure if that would help.

matham:
Perfect. Just one more thing, it's probably better if you use the [clock fixture](https://github.com/kivy/kivy/blob/master/kivy/tests/fixtures.py#L11) rather than the raw clock.

gottadiveintopython:
To use that fixture I have to do something like this:

```python
class Test_layout:
    def test_case1(self, kivy_clock, ...):
       self.compute_layout(..., clock=kivy_clock)
    def test_case2(self, kivy_clock, ...):
       self.compute_layout(..., clock=kivy_clock)
```

This isn't desireble, right?

matham:
>This is not desirable, isn't it?

I think that's fine!? It's just more verbose, but I don't really see another way that is not hacky.

gottadiveintopython:
@matham OK, how about the following fixture? Do you think it's a good idea?

```python
# kivy/tests/fixtures.py
@pytest.fixture(scope='module')
def module_scope_kivy_clock():
    # same as kivy_clock


# kivy/tests/test_uix_recyclegridlayout.py
global_kivy_clock = None

@pytest.fixture(scope='module', autouse=True)
def acquire_kivy_clock(module_scope_kivy_clock):
    global global_kivy_clock
    global_kivy_clock = module_scope_kivy_clock
    yield module_scope_kivy_clock
    global_kivy_clock = None

class Test_layout:
    def compute_layout(...):
         global_kivy_clock.tick()
```

gottadiveintopython:
`TestLayout_all_the_data_is_visible` might be unnecessary if `TestLayout_only_a_part_of_the_data_is_visible` exists.

matham:
>OK, how about the following fixture?

I think the way it is currently is better. It's simpler even if more verbose, which is not really a big deal.

>TestLayout_all_the_data_is_visible might be unnecessary if TestLayout_only_a_part_of_the_data_is_visible exists.

Extra tests have never hurt anyone 😄

matham (dismissed):

matham:
Feel free to merge if you're happy with it.

gottadiveintopython:
@matham I think the pr is mergable now. I'd appreciate if you approve it again :)

kivy/uix/recyclegridlayout.py
```diff
@@ -220,22 +222,24 @@ def compute_visible_views(self, data, viewport):
...
             at_idx((x, y)),
             at_idx((right, y)),
             at_idx((x, top)),
             at_idx((right, top)),
         ))
 
         n = len(data)
+        if len({tl, tr, bl, br}) < 4:
+            # visible area is one row/column
+            return list(range(min(n, tl), min(n, br + 1)))
```
gottadiveintopython, line 235: This part doesn't have to create a list. I have to fix it.

matham: Hmm I suppose. The interface doesn't say that it needs to return a list so it should be fine.

matham (approved):

PR #7263, opened 2020-12-12
Camera: Added API to change avfoundation camera provider orientation

misl6:
This PR adds the ability to programmatically change the orientation of the video feed provided by the AvFoundation camera provider.

Maintainer merge checklist

### DIFF
--- a/kivy/core/camera/camera_avfoundation_implem.m
+++ b/kivy/core/camera/camera_avfoundation_implem.m
@@ -5,7 +5,6 @@
  *
  * TODO:
  * - add interface for setting some capabilities as focus/exposure/...
- * - add interface for setting orientation
  * I've let the code concerning caps, even if it's not yet used. uncomment
  * WITH_CAMERA_CAPS to compile with it.
  */
@@ -140,6 +139,7 @@ public:
     bool attemptCapturePreset(NSString *preset);
     bool attemptStartMetadataAnalysis();
     bool haveNewMetadata();
+    bool setVideoOrientation(int orientation);
 
 #ifdef WITH_CAMERA_CAPS
     double getProperty(int property_id);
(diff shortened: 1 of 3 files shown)


kivy/core/camera/camera_avfoundation_implem.m
```diff
@@ -626,7 +633,10 @@ -(int) updateImage {
         if (image == NULL)
             image = new CameraFrame((int)width, (int)height);
 
-        if (image->datasize != width * height * sizeof(char) * 4) {
+        bool resolution_change_detected = image->datasize != width * height * sizeof(char) * 4;
+        bool orientation_change_detected = (image->width == height) && (image->height == width);
```
matham, line 637: What if the width and the height are the same?

misl6: Well, in that case, We don't need to update the image properties, cause there's no need to re-create a Texture on the Kivy side.

The user will just see a flipped/rotated image accordingly to his selection.

matham: Hmm what I meant, unless I'm misreading it, when the width and height are the same, `orientation_change_detected` will be true and then the `if` will be executed for each frame. Meaning it will do a malloc and free on a every frame, which is not desired.

misl6: Ahh, no you don't misread that, I was not thinking at that case at all (on iOS I should stick to one of the available presets, so I didn't had the opportunity to test this specific case).

But, on MacOS, the user can specify the width + height, so it's a good point, let me fix this one ASAP.

matham: Looking at this again, I think we're trying to be too clever, when we should always update the width and height, but only update the buffer if the buffer size needs to change.

Setting the width and height on every frame doesn't cost much. And i think the original and current  implementation is still buggy. E.g. what if width and height are 5, 4, respectively. And it changes to 2, 10. The width and height won't be updated. Maybe it never happens, but it's asking fot trouble.

misl6: What a cleaner and nicer alternative. Pushed and tested it.

Yeah, unfortunately all the provider implementation it's asking for trouble, as I can see We have a multitude of issues on the avfoundation camera provider.

Sometimes also crashes for a `BADACCESS`, I'm trying to fix one thing per time, but it's going to be a long journey :)

matham (approved):

matham:
>Yeah, unfortunately all the provider implementation it's asking for trouble, as I can see We have a multitude of issues on the avfoundation camera provider.

The only solution to these and other such issues is to have unittests. We can have a mock camera class that just serves images from a memory or a video file to test core camera or uix/camera. However, I don't see how that can help us test these providers. Perhaps there's a way to mock avfoundation camera and somehow test it? It would be worth investigating.

PR #7264, opened 2020-12-12
Camera: Change play default to False

wolfmanjm:
Also fixes play = False|True not working.

There was a regression in the uix/camera.py for 2.0.0 where the initial start of the camera as specified in a kv file by `play: False` was being ignored, and the camera would always start and subsequently any further changes to play would be ignored.

The changes between 1.11.1 and 2.0.0 that caused this are explained below in a comment.

Also by luck that bug also allowed the binding of on_texture to on_tex which would not have occurred if play was False.
Both issues are now fixed and seem to work as expected.


matham:
I somewhat worry about merging this because the clock scheduling seems like a band-aid since we have not identified the real cause. It's hard to know if this fixes the issue properly without knowing that.

About race conditions, that should not be possible since this all happens in one thread AFAIK. So it would be good to understand what is going on.

wolfmanjm:
The issue is pretty clear to me... The KV file has not been loaded in the __init__ so play has not been set properly. __init__ initially calls the on_index(). If you are allowed to set play: False in a KV file then the KV file must be loaded when on_index is called. Without this patch the camera is running continously which is a real drain on the system (especially if it is an RPI).
I think in older versions the KV file had been loaded when __init__ was called, I shouldn't have said it was a race condition, but I don't know what changes occured between 1.11.1 and 2.0.0.

wolfmanjm:
Ok I figured out the difference between 1.11.1 which works and 2.0.0 which doesn't.

1.11.1 does not try to start the camera when `on_index` is called from` __init__` if the KV has not been loaded yet due to this test: https://github.com/kivy/kivy/blob/stable-1.11.1/kivy/uix/camera.py#L101 whereas in V2.0.0 it always tries to start the camera when called from `__init__`  https://github.com/kivy/kivy/blob/master/kivy/uix/camera.py#L102  which won't work becuase the KV has not been loaded yet. If there is no KV file then resolution will have been set as will play in the ctor.
If we want to keep the code in on_index the same then delaying the call to allow the KV to load fixes the case where there is a KV and still works if there is no KV.

matham:
Ok, thanks for explaining it. 

The reason I worry about using Clock is that we can't be sure what happened between scheduling the callback and it being executed. E.g. What if the user creates the camera, and in the same clock cycle sets play to False and then True. This will end with the camera started. But then, the callback is executed, and the camera is started again. I can see some of the cameras would be confused by this as they don't check if they are already started.

The real issue is that play defaults to True, when it should default to False to prevent the camera starting before you're ready. I wonder if it would make sense to change the default? I'd be fore changing it. Even though it's a backward compat break, I think it's the proper solution.

Alternatively, in the clock callback, if you check whether the underlying camera is `stopped`, and only then start it.

wolfmanjm:
Yes I see the issue now. I thought Clock was the suggested way to wait for the KV to be loaded. (although on_kv_post  is now available but not suitable for this case).
I'll test if simply setting play to False fixes the issue with no other changes, I agree this is a better solution if it works.
Thanks

wolfmanjm:
Ok I tested the fix and it is not sufficient to just set play to False, because in that case the on_tex is not bound. So I moved that stanza out of the `if self.play:` and it now works as it should.
Hopefully this is an acceptible fix.

kivy/uix/camera.py
```diff
@@ -37,7 +37,7 @@ class Camera(Image):
     '''Camera class. See module documentation for more information.
     '''
 
-    play = BooleanProperty(True)
+    play = BooleanProperty(False)
```
matham, line 40: >:attr:`play` is a :class:`~kivy.properties.BooleanProperty` and defaults to True.

Looks good  just need to change the docs below to say that it defaults to False

matham (changes requested):
Otherwise looks good.

kivy/uix/camera.py
wolfmanjm: Done

matham (approved):

PR #7270, opened 2020-12-15
Graphics: Check whether user updated GL instructions from external thread.

matham:
Maintainer merge checklist

New users often don't realize that you cannot set graphics or change properties that then set graphics properties outside the main Kivy thread. But they notice something is wrong, but don't know why.

This adds a option that will error out of graphics are created or updated outside the main kivy thread. By default the option is ON, but it can be turned off in config.

In terms of performance loss, if the option is turned OFF, there's basically no additional cost. But if it's turned ON, it's slower, but only in the tens of microseconds per property, which I think is negligible and worth making Kivy more bullet proof and friendlier to new users.

Here is the performance testing code:

```py
from timeit import timeit
from kivy.graphics import Rectangle, Color, Line

widget = Widget()
with widget.canvas:
    color = Color()
    rect = Rectangle()
    line = Line()

number = 1_000_000
values = []
for val in ('0', '1'):
    Config.set('graphics', 'verify_gl_main_thread', val)
    values.append(sum(timeit(setup='from __main__ import color, rect, line', stmt='color.rgb = .1, .2, .3', number=number // 10) for _ in range(10)))
    values.append(sum(timeit(setup='from __main__ import color, rect, line', stmt='rect.pos = .1, .2', number=number // 10) for _ in range(10)))
    values.append(sum(timeit(setup='from __main__ import color, rect, line', stmt='line.points = [1, 2, 3, 4, 5, 6]', number=number // 10) for _ in range(10)))

for i in range(3):
    print(f'{values[i]:0.4f} -> {values[i + 3]:0.4f}: {(values[i + 3] - values[i]) / values[i] * 100:0.2f}%, {((values[i + 3] - values[i]) / number):0.2E} difference')
```

This prints:
```
0.2593 -> 0.3300: 27.24%, 7.07E-08 difference
0.0428 -> 0.0453: 5.75%, 2.46E-09 difference
0.1718 -> 0.2618: 52.41%, 9.00E-08 difference
```
The percent may seem large, but look at the actual increase, which is in microseconds.

I also ran it with master (where setting the config option does nothing), and it printed:
```
0.2397 -> 0.2262: -5.64%, -1.35E-08 difference
0.0447 -> 0.0471: 5.41%, 2.42E-09 difference
0.1762 -> 0.1735: -1.51%, -2.67E-09 difference
```
which shows that when the option is turned OFF (compare first number to the first number abover), it basically is the same.

### DIFF
--- a/.ci/ubuntu_ci.sh
+++ b/.ci/ubuntu_ci.sh
@@ -102,7 +102,7 @@ install_kivy_sdist() {
 
 test_kivy() {
   rm -rf kivy/tests/build || true
-  KIVY_NO_ARGS=1 python3 -m pytest --timeout=300 --cov=kivy --cov-report term --cov-branch "$(pwd)/kivy/tests"
+  KIVY_NO_ARGS=1 python3 -m pytest --maxfail=10 --timeout=300 --cov=kivy --cov-report term --cov-branch "$(pwd)/kivy/tests"
 }
 
 test_kivy_install() {
@@ -116,7 +116,7 @@ test_kivy_install() {
   plugins = kivy.tools.coverage
 
 EOF
-  KIVY_TEST_AUDIO=0 KIVY_NO_ARGS=1 python3 -m pytest --timeout=300 .
+  KIVY_TEST_AUDIO=0 KIVY_NO_ARGS=1 python3 -m pytest --maxfail=10 --timeout=300 .
 }
 
 upload_coveralls() {
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -369,7 +369,7 @@ from weakref import ref
 _is_rpi = exists('/opt/vc/include/bcm_host.h')
 
 # Version number of current configuration format
-KIVY_CONFIG_VERSION = 22
+KIVY_CONFIG_VERSION = 23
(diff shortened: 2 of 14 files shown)


tshirtman:
This looks like a great idea, didn't test yet, but i believe the numbers you quoted should be fine, especially as it's possible to disable it when releasing your apps without any downside, as any issue of this kind should have been caught during development, code looks fine.

matham:
Looks like it required a 64 bit in to store the thread id that's why it originally failed on linux/osx.

Unit tests, testing both enabling and disabling the config options and verifying it raises/doesn't raise a exception is included in the PR.

xloem:
I reviewed the last commit and found python uses `unsigned long` ( https://github.com/python/cpython/blob/master/Modules/_threadmodule.c#L1208 64-bits on 64-bit systems) for the datatype in question.  Under the hood it seems to make the value in a platform-dependent way.  I'm happy to see automerge.  I imagine software projects growing as big wikis, like wikipedia.

matham:
Thanks for the info about `unsigned long`, I updated the code to use it.

PR #7271, opened 2020-12-16
Inspector: Prevent circular import breaking Window

tshirtman:
This was caused by a circular import
- When the window is created, it registers the module
- The module imports TextInput
- TextInput imports the Window object, to bind to it
- But the window is still being created, so the module tries to create
  it again.
- Python detects the circular import, and returns an incomplete
  inspector module, without a start method
- the module can't be started, and the Window object is fubar.

Solution.
The TextInput import being the cause of the problem, instead of
importing it, we can import the Factory and lookup the TextInput object
from it.
To avoid similar problems potentially happening in the future, the other
widgets and instructions have also been converted to using the Factory.

This might still be brittle, and possibly changing the way the modules
are registered would be more future-proof.

Maintainer merge checklist


matham:
Hmm, that indeed sounds a little brittle. I suppose removing the aggressive importing and window creation on import would also fix this.

Do you think there's a way to unit test this?

My one concern is that having to go through the factory every time we need a widget, which will probably be slower. But I don't think it's a big deal.

tshirtman:
Another way to make this less brittle would be to wait after Window creation to register modules, but it might break their current expectations, i didn't try yet.

matham (approved):

matham:
Yeah, I think we should maybe change when modules become active. I.e. start/stop modules when the app starts instead of when window is created. Or a better fix would be to make Window a proxy type object and create a dummy one when it's imported, but then change to a real window when the app starts. But there are all part of the aggressive import problem.

Issue #6646, opened 2019-12-15
Scrollview not working properly on Lenovo Yoga 920 W10


ElliotGarbus:
### Versions

* Python: 3.73
* OS: Windows10
* Kivy: 1.11.1
* Kivy installation method: pip installed in the venv

### Description
Scrollview not working properly on Lenovo Yoga 920 W10: using the touchpad to scroll down, scrolls up.  Using a mouse wheel works as expected, using touch works as expected.  The touchpad does not scroll properly.

### Code and Logs

```python
# Spinner and Scrollview tests. Scrollup does not work properly with a trackpad.

from kivy.app import App
from kivy.lang import Builder
from kivy.metrics import dp
from kivy.uix.spinner import Spinner

kv = '''
BoxLayout:
    orientation: 'vertical'
    BoxLayout:  # Line 1
        size_hint_y: 1
        BankSpinner:
            values: ['Bank: ' + str(n) for n in range(1,31)]
            text: 'Bank Select'
    ScrollView:
        size_hint_y:3
        do_scroll_x: False
        do_scroll_y: True
        scroll_type: ['bars', 'content']
        bar_width: 20
        Label:
            size_hint_y: None
            height: self.texture_size[1]
            text_size: self.width, None
            padding: 10, 10
            text: 'really some amazing text\\n' * 100    
    
    BoxLayout:
        size_hint_y: 8
'''


class BankSpinner(Spinner):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.dropdown_cls.max_height = 10 * dp(48)

    def on_touch_up(self, touch):
        print(f'touch up:{touch}, touch.device: {touch.device} {touch.profile} ')
        if 'button' in touch.profile:
            print(f'button: {touch.button}')
        return super().on_touch_up(touch)

    def on_touch_move(self, touch):
        print(f'touch move:{touch}, touch.device: {touch.device}')
        return super().on_touch_move(touch)

    def on_touch_down(self, touch):
        print(f'touch down:{touch}, touch.device: {touch.device}')
        return super().on_touch_down(touch)


class SpinnerTestApp(App):

    def build(self):
        return Builder.load_string(kv)


SpinnerTestApp().run()
```
Additionally under Inclement's direction I instrumented window_sdl2.py
[...]

PR #7279, opened 2020-12-18
Window: prevent "empty" mousewheel events from breaking scrollview

tshirtman:
It seems on some machines, (and specificaly on windows) lenovo ones, but maybe others, scrolling with the touchpad can create a lot of "fake" scrolling events, without any associated offset (x or y), the way kivy works currently dispatches these events as "mousewheelup" (which indicates a scroll down). This makes it really hard to scroll up this way, as there is a constant stream of events pushing in the other direction, and the more we try to scroll, the more there are.

This change ignore these empty events.

fixes: #6646

Maintainer merge checklist

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -571,10 +571,16 @@ cdef class _WindowSDL2Storage:
         elif event.type == SDL_MOUSEWHEEL:
             x = event.wheel.x
             y = event.wheel.y
+            # TODO we should probably support events with both an x and y offset
             if x != 0:
                 suffix = 'left' if x > 0 else 'right'
-            else:
+            elif y != 0:
                 suffix = 'down' if y > 0 else 'up'
+            else:
+                # It's possible to get mouse wheel events with no offset in
+                # either x or y direction, we just ignore them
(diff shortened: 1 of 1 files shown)


matham:
Great find! However, I think it may be better if fixed at the [source](https://github.com/kivy/kivy/blob/master/kivy/core/window/_window_sdl2.pyx#L571)? In fact looking at the sdl2 [docs for that event](https://wiki.libsdl.org/SDL_MouseWheelEvent), we should ignore the event if both are zero. So it's probably better if we also just return `None` in `_window_sdl2.pyx` and never process it.

Interestingly enough, it seems that we should also support a situation where there's scrolling in both x and y, right now we just pick scroll in x if there's scrolling in both (e.g. a joystick). Or maybe that's just not a thing.

tshirtman:
Yes, i considered that, and i agree it could be better, however, i couldn't *test* that, as i only have a normal installation, not a development one (and i'm very much lazy to make one, considering my history with visual studio), in windows, but i guess i could try to get wheels from the CI to confirm it works.

matham:
Yeah, you could use wheels. Also, can you add a link to the sdl2 doc in a comment indicating the issue so that in future it'd be clear why we skip zeros?

matham (approved):

PR #7286, opened 2020-12-24
Doc: Add negative size warning

### LINKED ISSUE #7226: CollideWidget does not work for widgets with negative size
**Software Versions**
* Python: 3.7.9
* OS:  Ubuntu 16.04
* Kivy: 1.11.1
* Kivy installation method: `pip install kivy`

**Describe the bug**
If the width and/or height of a widget is negative, `collide_widget` cannot accurately detect collisions. The method assumes `self.right > self.x` and `self.top > self.y` which is not true when `self.width < 0` or `self.height < 0`
```
# kivy.uix.widget
class Widget(WidgetBase):
    ...
    def collide_widget(self, wid):
       """..."""
        if self.right < wid.x:
            return False
        if self.x > wid.right:
            return False
        if self.top < wid.y:
            return False
        if self.y > wid.top:
            return False
        return True
```

**Expected behavior**
`collide_widget` detects collisions no matter what the dimensions of the widget are

**To Reproduce**
see code below

The below widgets are clearly  colliding but `collide_widget` cannot detect it because of the above stated bug. Below is the code that produced this screenshot.
kivy_issue_7226

```python
from kivy.app import App
from kivy.lang import Builder

kv = """
<TextBoundLabel@Label>:
    size_hint: None, None
    size: self.texture_size

<LabeledWidget@Widget>:
    size_hint: None, None
    canvas:
        Color:
            rgb: 1, 0, 0, 0.1
        Rectangle:
            pos: self.pos
            size: self.size
    
    TextBoundLabel:
        text: "Bottom Left"
        pos: self.parent.pos
    
    TextBoundLabel:
        text: "Bottom Right"
        pos: self.parent.right, self.parent.y
    
    TextBoundLabel:
        text: "Top Left"
        pos: self.parent.x, self.parent.top
    
    TextBoundLabel:
        text: "Top Right"
        pos: self.parent.right, self.parent.top

FloatLayout:
    Label:
        pos_hint: {"center_x": 0.5, "center_y": 0.9}
        text: f"Is there a collision?: {positive_wid.collide_widget(negative_wid)}"
    
    LabeledWidget:
        id: positive_wid
        pos_hint: {"center_x": 0.5, "center_y": 0.5}
        
        Label:
            pos: positive_wid.pos
            text: "Positive Widget"
    
    LabeledWidget:
        id: negative_wid
        size: -100, -100
        pos_hint: {"center_x": 0.6, "center_y": 0.6}
        
        Label:
            pos: negative_wid.right, negative_wid.top  # notice how I didn't use self.pos because the position would be wrong
            text: "Negative Widget"
"""


class MyApp(App):
    def build(self):


if __name__ == '__main__':
    MyApp().run()

```
I was able to fix this by overwriting `collide_widget` like so:
```python
    def collide_widget(self, wid: Widget) -> bool:
        """patch for negative dimensions"""
        if max(self.right, self.x) < min(wid.x, wid.right):
            return False
        if min(self.x, self.right) > max(wid.right, wid.x):
            return False
        if max(self.top, self.y) < min(wid.y, wid.top):
            return False
        if min(self.y, self.top) > max(wid.top, wid.y):
            return False
        return True
```
**Additional context**
I would've just submitted a PR patching `collide_widget`, but I suspect the code base assumes `self.right` is always right of `self.x` and `self.top` is always above `self.y`. What are the core devs thoughts on this? I see two options:
1. Modify the definition of `right` and `top` such that they are always the right edge and top edge of a widget. This will potentially fix unknown bugs in the code base that make the above position attribute assumptions.
1. Patch `collide_widget` as describe above and let the programmer deal with the consequences of a negative dimensions.

xayhewalo:
fixes #7226. I'm leaving the maintainer checklist below but I don't think most of it is relevant.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -1055,11 +1055,7 @@ class Widget(WidgetBase):
         method.
 
     .. warning::
-        Kivy allows the `width` property to be negative, but Kivy does not
-        support negative sizes. It is recommended to always have a positive
-        `width` property. If not, the programmer must account for the
-        consequences of a negative `width`. For example, `right` will actually
-        be left of `x` when `width` is negative.
+        A negative width is not supported.
     '''
 
     height = NumericProperty(100)
@@ -1074,11 +1070,7 @@ class Widget(WidgetBase):
         method.
 
     .. warning::
-        Kivy allows the `height` property to be negative, but Kivy does not
-        support negative sizes. It is recommended to always have a positive
-        `height` property. If not, the programmer must account for the
-        consequences of a negative `height`. For example, `top` will actually
-        be below of `y` when `height` is negative.
+        A negative height is not supported.
     '''
 
     pos = ReferenceListProperty(x, y)
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -1053,6 +1053,13 @@ class Widget(WidgetBase):
         Keep in mind that the `width` property is subject to layout logic and
         that this has not yet happened at the time of the widget's `__init__`
         method.
+
+    .. warning::
+        Kivy allows the `width` property to be negative, but Kivy does not
+        support negative sizes. It is recommended to always have a positive
+        `width` property. If not, the programmer must account for the
+        consequences of a negative `width`. For example, `right` will actually
+        be left of `x` when `width` is negative.
     '''
 
     height = NumericProperty(100)
@@ -1065,6 +1072,13 @@ class Widget(WidgetBase):
         Keep in mind that the `height` property is subject to layout logic and
         that this has not yet happened at the time of the widget's `__init__`
         method.
+
+    .. warning::
+        Kivy allows the `height` property to be negative, but Kivy does not
+        support negative sizes. It is recommended to always have a positive
+        `height` property. If not, the programmer must account for the
+        consequences of a negative `height`. For example, `top` will actually
+        be below of `y` when `height` is negative.
     '''
 
     pos = ReferenceListProperty(x, y)


kivy/uix/widget.py
```diff
@@ -1053,6 +1053,13 @@ def get_window_matrix(self, x=0, y=0):
         Keep in mind that the `width` property is subject to layout logic and
         that this has not yet happened at the time of the widget's `__init__`
         method.
+
+    .. warning::
+        Kivy allows the `width` property to be negative, but Kivy does not
+        support negative sizes. It is recommended to always have a positive
+        `width` property. If not, the programmer must account for the
+        consequences of a negative `width`. For example, `right` will actually
+        be left of `x` when `width` is negative.
```
matham, line 1062:
```suggestion

    .. warning::
        A negative size is not supported.
```

I think something very concise is better because we don't need say anything beyond it not being supported. And most people will not care much beyond that.

kivy/uix/widget.py
```diff
@@ -1065,6 +1072,13 @@ def get_window_matrix(self, x=0, y=0):
         Keep in mind that the `height` property is subject to layout logic and
         that this has not yet happened at the time of the widget's `__init__`
         method.
+
+    .. warning::
+        Kivy allows the `height` property to be negative, but Kivy does not
+        support negative sizes. It is recommended to always have a positive
+        `height` property. If not, the programmer must account for the
+        consequences of a negative `height`. For example, `top` will actually
+        be below of `y` when `height` is negative.
```
matham, line 1081:
```suggestion

    .. warning::
        A negative size is not supported.
```

matham (changes requested):

matham (approved):

PR #7290, opened 2020-12-27
Camera: improve avfoundation by using memoryview and re-scheduling the interval on framerate change

misl6:
The purpose of this PR is to enhance the avfoundation camera provider on both iOS and macOS.

By using Cython memoryview I'm avoiding to call `PyBytes_FromStringAndSize` when assigning the buffer via `self._buffer = <bytes>data[:rowsize * height]`.

Now, the overall performance it's slightly better and I'm not incurring anymore random `BADACCESS` errors that were leading the App into a crash.

I've also added an update rescheduling procedure, to have a consistent update timing with the selected framerate.

Maintainer merge checklist

### DIFF
--- a/kivy/core/camera/camera_avfoundation.pyx
+++ b/kivy/core/camera/camera_avfoundation.pyx
@@ -11,7 +11,7 @@ cdef extern from "camera_avfoundation_implem.h":
     ctypedef void *camera_t
     camera_t avf_camera_init(int index, int width, int height)
     void avf_camera_deinit(camera_t camera)
-    void avf_camera_update(camera_t camera)
+    bint avf_camera_update(camera_t camera)
     void avf_camera_start(camera_t camera)
     void avf_camera_stop(camera_t camera)
     void avf_camera_get_image(camera_t camera, int *width, int *height, int *rowsize, char **data)
@@ -28,6 +28,7 @@ from kivy.clock import Clock
 from kivy.graphics.texture import Texture
 from kivy.core.camera import CameraBase
 from kivy.utils import platform
+from cython cimport view as cyview
 
 
 cdef class _AVStorage:
@@ -45,6 +46,7 @@ class CameraAVFoundation(CameraBase):
         self._storage = _AVStorage()
         self._update_ev = None
         self._metadata_callback = None
+        self._framerate = 30
         super(CameraAVFoundation, self).__init__(**kwargs)
 
     def init_camera(self):
@@ -52,22 +54,39 @@ class CameraAVFoundation(CameraBase):
         storage.camera = avf_camera_init(
             self._index, self.resolution[0], self.resolution[1])
 
+    @property
+    def _scheduled_rate(self):
+        # We're going 4 times faster the framerate to avoid frame skipping.
+        return 1 / (self._framerate * 4)
+
     def _update(self, dt):
         cdef _AVStorage storage = <_AVStorage>self._storage
         cdef int width, height, rowsize
         cdef char *data
         cdef char *metadata_type
         cdef char *metadata_data
+        cdef cyview.array cyarr
 
         if self.stopped:
             return
 
-        avf_camera_update(storage.camera)
+        if not avf_camera_update(storage.camera):
+            return
+
         avf_camera_get_image(storage.camera,
             &width, &height, &rowsize, &data)
 
         if data == NULL:
             return
+
+        cyarr = cyview.array(
+            shape=(rowsize * height,),
+            itemsize=sizeof(char),
+            format="B",
+            mode="c",
+            allocate_buffer=False,
+        )
+        cyarr.data = data
 
         self._resolution = (width, height)
         
@@ -79,20 +98,23 @@ class CameraAVFoundation(CameraBase):
             self._texture.flip_vertical()
             self.dispatch('on_load')
 
-        self._buffer = <bytes>data[:rowsize * height]
         self._format = 'bgra'
+        self._texture.blit_buffer(cyarr, colorfmt=self._format)
         self._copy_to_gpu()
         if self._metadata_callback:
             if avf_camera_have_new_metadata(storage.camera):
                 avf_camera_get_metadata(storage.camera, &metadata_type, &metadata_data)
                 self._metadata_callback(metadata_type, metadata_data)
+    
+    def _copy_to_gpu(self):
+        self.dispatch('on_texture')
 
     def start(self):
         cdef _AVStorage storage = <_AVStorage>self._storage
         super(CameraAVFoundation, self).start()
         if self._update_ev is not None:
             self._update_ev.cancel()
-        self._update_ev = Clock.schedule_interval(self._update, 1 / 30.)
+        self._update_ev = Clock.schedule_interval(self._update, self._scheduled_rate)
         avf_camera_start(storage.camera)
 
     def stop(self):
@@ -105,7 +127,11 @@ class CameraAVFoundation(CameraBase):
 
     def set_framerate(self, framerate):
         cdef _AVStorage storage = <_AVStorage>self._storage
-        avf_camera_attempt_framerate_selection(storage.camera, framerate)
+        if avf_camera_attempt_framerate_selection(storage.camera, framerate):
+            if self._update_ev is not None:
+                self._update_ev.cancel()
+            self._framerate = framerate
+            self._update_ev = Clock.schedule_interval(self._update, self._scheduled_rate)
 
     def set_preset(self, preset):
         cdef _AVStorage storage = <_AVStorage>self._storage
--- a/kivy/core/camera/camera_avfoundation_implem.h
+++ b/kivy/core/camera/camera_avfoundation_implem.h
@@ -2,7 +2,7 @@ typedef void *camera_t;
 
 camera_t avf_camera_init(int index, int width, int height);
 void avf_camera_deinit(camera_t camera);
-void avf_camera_update(camera_t camera);
+bool avf_camera_update(camera_t camera);
 void avf_camera_start(camera_t camera);
 void avf_camera_stop(camera_t camera);
 void avf_camera_get_image(camera_t camera, int *width, int *height, int *rowsize, char **data);
--- a/kivy/core/camera/camera_avfoundation_implem.m
+++ b/kivy/core/camera/camera_avfoundation_implem.m
@@ -218,20 +218,23 @@ Camera::~Camera() {
 }
 
 bool Camera::grabFrame(double timeOut) {
-
+    
     NSAutoreleasePool* localpool = [[NSAutoreleasePool alloc] init];
+    bool haveFrame = false;
     double sleepTime = 0.005;
     double total = 0;
     NSDate *loopUntil = [NSDate dateWithTimeIntervalSinceNow:sleepTime];
-    [capture updateImage];
-    while (![capture updateImage] && (total += sleepTime)<=timeOut &&
+    haveFrame = [capture updateImage];
+    while (!haveFrame && (total += sleepTime)<=timeOut &&
             [[NSRunLoop currentRunLoop] runMode: NSDefaultRunLoopMode
-            beforeDate:loopUntil])
+            beforeDate:loopUntil]){
+        haveFrame = [capture updateImage];
         loopUntil = [NSDate dateWithTimeIntervalSinceNow:sleepTime];
+    }
 
     [localpool drain];
 
-    return total <= timeOut;
+    return haveFrame;
 }
 
 CameraFrame* Camera::retrieveFrame() {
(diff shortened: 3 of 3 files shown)


kivy/core/camera/camera_avfoundation.pyx
```diff
@@ -58,14 +59,25 @@ class CameraAVFoundation(CameraBase):
         cdef char *data
         cdef char *metadata_type
         cdef char *metadata_data
+        cdef cyview.array cyarr
 
         if self.stopped:
             return
 
+
```
tshirtman, line 67: was this added empty line intentional? it doesn't seem pep8 compliante

tshirtman (approved):
LGTM, but didn't test.

kivy/core/camera/camera_avfoundation.pyx
misl6: Absolutely not, I was doing some debug in that area, and I missed it during cleanup. Removed.

kivy/core/camera/camera_avfoundation.pyx
```diff
@@ -66,6 +68,15 @@ class CameraAVFoundation(CameraBase):
         avf_camera_get_image(storage.camera,
             &width, &height, &rowsize, &data)
 
+        cyarr = cyview.array(
+            shape=((rowsize * height),),
```
tshirtman, line 72: just saw that now, maybe it's a remainder of debugging too :)
```suggestion
            shape=(rowsize * height,),
```

tshirtman (dismissed):

kivy/core/camera/camera_avfoundation.pyx
misl6: 🥶 Fixed and rebased. Thank you

tshirtman (dismissed):

kivy/core/camera/camera_avfoundation.pyx
```diff
@@ -105,7 +119,10 @@ class CameraAVFoundation(CameraBase):
 
     def set_framerate(self, framerate):
         cdef _AVStorage storage = <_AVStorage>self._storage
-        avf_camera_attempt_framerate_selection(storage.camera, framerate)
+        if avf_camera_attempt_framerate_selection(storage.camera, framerate):
+            if self._update_ev is not None:
+                self._update_ev.cancel()
+            self._update_ev = Clock.schedule_interval(self._update, 1 / float(framerate))
```
matham, line 125: You're better off scheduling it at e.g. 4 times the framerate. Because if a frame is not available, it'll just return early. But you don't want to be waiting for a frame if one is available. I'd also change it for the 1 / 30 case.

matham (changes requested):

kivy/core/camera/camera_avfoundation.pyx
misl6: It makes sense, but We need to make a couple of (simple?) new edits to the provider, in order to stop the update process immediately if it's not needed.

Without any change, multiplying the updates to x4 the frame rate will request to handle x4 times all the update process that is huge.

As an example, now, after these changes I made, on an iPhone 12 Pro, @ 30 FPS, I'm having a 30/35 % of constant CPU usage (When using the default Preset in High Res). I don't imagine doing it x4 times.

P.S: Now, after these changes all the CPU usage seems to be on `glTexImage2D` (I noticed that via XCode Instruments)

matham: Hmm, but I don't expect the CPU would be working 4 times as harder. Rather kivy will just execute the callback 4 times as much, but on average, 3 out of those 4 times it'll return immediately with no new frame? 

I expect the high CPU is coming from reading the actual frames, and although you'd lose less frames, I expect the CPU increase will be minimal, proportional to the increase in frames you're not missing. But it shouldn't increase 4x. And if the CPU is too much, then decreasing the frame rate should be the solution, not missing frames. I.e. ideally no frame is missed.

But I get it, this can be done in a future PR, not need to troubleshoot it all in this PR.

misl6: > Hmm, but I don't expect the CPU would be working 4 times as harder. Rather kivy will just execute the callback 4 times as much, but on average, 3 out of those 4 times it'll return immediately with no new frame?

That's the point, now We don't check if it's a new frame, and then We will process it again by re-updating the texture.
Not necessarily in a future PR, it's quite simple. Let me see if I have some time in the next very few days.

matham: Oh, ok. I assumed that it would return NULL if we already got that frame.

kivy/core/camera/camera_avfoundation.pyx
```diff
@@ -66,6 +68,15 @@ class CameraAVFoundation(CameraBase):
...
             &width, &height, &rowsize, &data)
 
+        cyarr = cyview.array(
+            shape=(rowsize * height,),
+            itemsize=sizeof(char),
+            format="B",
+            mode="c",
+            allocate_buffer=False,
+        )
+        cyarr.data = data
```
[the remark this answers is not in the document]
misl6: Absolutely OK for the `data == NULL` part, will do that ASAP.

By looking at the code, We could have a complete idea of the entire lifecycle of that pointer, cause We create and update it. ?And seems to be safe?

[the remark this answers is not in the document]
matham: I don't know the rules of the language of the .m files (I don't even know what language it is, presumably some c derivative). But I can see that the data pointer `data` is created with `malloc` in `updateImage` and that that buffer dies in the `~CameraFrame()` destructor.

Actually, when does that garbage collected? The image doesn't stay around forever, or does it? Is there maybe only one image and we simply pass around that one buffer? Or does a buffer get created for each image. If the former, then it's probably fine (although we need to be sure no one gets a hold of the buffer as it'll be overwritten by the next frame. If the latter, then we need to know how long the data pointer stays around?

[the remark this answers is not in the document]
matham: I guess the problem is that once we pass a pointer on to Python, we can't be sure how long a user may keep a hold of the pointer implicitly by keeping a ref to the cython array. So if they hold a ref to the array, but the camera has been shut off and the pointer deleted, trying to read the array value will access a dead pointer.

So we need to make sure to warn very clearly next to the `_buffer` variable that it cannot be used outside the camera internals so we keep full control over it. So maybe just add a comment next to the creation of `_buffer` saying it cannot be copied externally so one one makes this mistake?

kivy/core/camera/camera_avfoundation.pyx
misl6: Updated it. Was quite easy, now We schedule an update 4 times faster than the framerate, but the texture is updated only if a new frame is detected. As I can see, now it's smoother and consistent between runs. It wasn't scheduled for this PR, but thank you for the advice.

kivy/core/camera/camera_avfoundation.pyx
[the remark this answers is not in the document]
misl6: So, let's do a little brief:

The data buffer is created with a `malloc`over here:
https://github.com/kivy/kivy/blob/8aca43ed9322e1fe98ca231106691d77e4d684fd/kivy/core/camera/camera_avfoundation_implem.m#L644

The destructor of CameraFrame, where the buffer dies is called over here into the `dealloc` function of the `CaptureDelegate`:
https://github.com/kivy/kivy/blob/8aca43ed9322e1fe98ca231106691d77e4d684fd/kivy/core/camera/camera_avfoundation_implem.m#L577

Now, considering that the `release` function of a NSObject is something like that, the `dealloc` function is called when the object is released and the retainCount is equal to 0:
```
- (void)release
{
    _retainCount--;
    if (_retainCount == 0) {
        [self dealloc]
    }
}
``` 
The release function, for the `CaptureDelegate` is called over here, into the `Camera::stopCaptureDevice()` function:
https://github.com/kivy/kivy/blob/8aca43ed9322e1fe98ca231106691d77e4d684fd/kivy/core/camera/camera_avfoundation_implem.m#L271

The `stopCaptureDevice` function it's called into the `Camera::~Camera()`destructor, or into the `avf_camera_stop`function.

The `Camera::~Camera()`destructor, is called over here, by the `avf_camera_deinit` function, but  it's like a dead-end track, cause We don't use it :
https://github.com/kivy/kivy/blob/8aca43ed9322e1fe98ca231106691d77e4d684fd/kivy/core/camera/camera_avfoundation_implem.m#L689

The `avf_camera_stop` function is called from the `stop` function of the provider.

So, speaking about the `data` buffer, is available during the entire usage of the provider.

**_WARNING!!_ Not-so-fun fact**: While doing this whitepaper, to clarify the `data` buffer lifecycle, I found out, that the provider implementation doesn't implement a `__del__` function to deinit (and stop the provider if the user left that in `play`) when needed. I could also notice that into my notification bar (The green dot, in iOS > 14, it means that the Camera is in use) :
IMG_0177

kivy/core/camera/camera_avfoundation.pyx
```diff
@@ -45,30 +46,48 @@ class CameraAVFoundation(CameraBase):
...
         cdef char *data
         cdef char *metadata_type
         cdef char *metadata_data
+        cdef cyview.array cyarr
 
         if self.stopped:
             return
 
-        avf_camera_update(storage.camera)
+        if avf_camera_update(storage.camera) is False:
```
matham, line 73:
```suggestion
        if not avf_camera_update(storage.camera):
```
This is the recommended pythonic way.

kivy/core/camera/camera_avfoundation.pyx
```diff
@@ -45,30 +46,48 @@ class CameraAVFoundation(CameraBase):
...
+        self._framerate = 30
         super(CameraAVFoundation, self).__init__(**kwargs)
 
     def init_camera(self):
         cdef _AVStorage storage = <_AVStorage>self._storage
         storage.camera = avf_camera_init(
             self._index, self.resolution[0], self.resolution[1])
 
+    @property
+    def update_speed(self):
```
matham, line 58:
```suggestion
    def _scheduled_rate(self):
```
Just making the name a little more concise and also not a public method.

kivy/core/camera/camera_avfoundation.pyx
```diff
@@ -79,20 +98,23 @@ class CameraAVFoundation(CameraBase):
...
+    def _copy_to_gpu(self):
+        self.dispatch('on_texture')
 
     def start(self):
         cdef _AVStorage storage = <_AVStorage>self._storage
         super(CameraAVFoundation, self).start()
         if self._update_ev is not None:
             self._update_ev.cancel()
-        self._update_ev = Clock.schedule_interval(self._update, 1 / 30.)
+        self._update_ev = Clock.schedule_interval(self._update, self.update_speed)
```
matham, line 117:
```suggestion
        self._update_ev = Clock.schedule_interval(self._update, self._scheduled_rate)
```

kivy/core/camera/camera_avfoundation.pyx
```diff
@@ -105,7 +127,11 @@ class CameraAVFoundation(CameraBase):
 
     def set_framerate(self, framerate):
         cdef _AVStorage storage = <_AVStorage>self._storage
-        avf_camera_attempt_framerate_selection(storage.camera, framerate)
+        if avf_camera_attempt_framerate_selection(storage.camera, framerate):
+            if self._update_ev is not None:
+                self._update_ev.cancel()
+            self._framerate = framerate
+            self._update_ev = Clock.schedule_interval(self._update, self.update_speed)
```
matham, line 134:
```suggestion
            self._update_ev = Clock.schedule_interval(self._update, self._scheduled_rate)
```

matham (changes requested):
Otherwise, it looks good.

kivy/core/camera/camera_avfoundation.pyx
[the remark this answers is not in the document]
matham: Thanks for tracing this down.

About the `__del__`, yeah, that should be fixed (but can be in a different PR). But just be careful not do any python stuff from `__del__` because `__del__` can be called from any other part of the code in the middle of doing other stuff or even from other threads.

kivy/core/camera/camera_avfoundation.pyx
misl6: Did it as suggested. Thank you.

kivy/core/camera/camera_avfoundation.pyx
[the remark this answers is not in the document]
misl6: Yes, let's do that into a different PR.
Will look at that during the next few days.

matham (approved):

matham:
Thanks!

BTW, there's no need to squash and force push when you add changes. Because github offers the option to squash the changes into one commit before merging. And seeing the individual commits actually makes it easier to see what exactly has changed so it makes it easier to review. Because otherwise it's harder to know if maybe there has been additional changes that need to be reviewed.

PR #7293, opened 2020-12-28
Properties: Add dynamic screen density/dpi support

matham:
Maintainer merge checklist

This PR adds support for dynamically updating and rescaling all `dpi` and `density` dependent NumericProperty instances. This allows automatically rescaling the UI as the `dpi` changes as in the following demo:

dpi

To make this demo, all I had to do was add the following code to the kivycatalog:
```py
from functools import partial
def set_density(density, *args):
    from kivy.metrics import Metrics
    Metrics.density = density
Clock.schedule_once(partial(set_density, 1.5), 15)
Clock.schedule_once(partial(set_density, 1), 20)
```

Reasons for this change
---------------------------
1. Accessibility, so users could more easily change visual sizing by faking the `density`/`dpi`.
2. To be able to simulate different device configuration more easily.
3. The primary reason I worked on this: to support late importing of Window as the first step to reduce aggressive importing. Previously, as soon as you used `10dp` or directly called `dp(x)`, the Window got created because you needed the window to compute the density/dpi. You couldn't lazily have a default density/dpi value and then later update it to the real value as it became available, because all the properties would already have been initialized to the default value and stuck with it. 

   With this change, we can now initialize to a default value without creating a `Window`, and then when the Window is created, all the properties that used the default `dp` will be updated automatically. Note, this PR does not change the creation of Window when `dp` is computed, this will happen in the future.

The new feature
------------------

Any `NumericProperty` or `VariableListProperty` that is set or initialized to a string using any of the format suffixes, e.g.
```py
widget = Widget()
widget.width = '10dp'
# or
class MyWidget(Widget):
    value = NumericProperty('25dp')
    value = VariableListProperty(['25dp'])
```
when `Metrics.density`, `Metrics.dpi`, or `Metrics.fontscale` is changed, all these properties will have their value re-computed and anything bound to these properties will be dispatched.

If the property is set or initialized to `'10px'`, or just a number (`5`), these properties will NOT dispatch or be re-computed (as there's nothing to recompute). 

However, to help users who can't want to use the string version, I added a property for each scaling format to `Metrics`. I.e. given that all the conversions essentially multiply the input by a constant value, these new properties can be used directly to multiply the input. You can just do `width: self.texture_size[0] + 10 * Metrics.dp` instead of `width: self.texture_size[0] + dp(10)` and the former will update when `dp` changes since it's a kivy property. So in KV you can do either of the following and the result will be the same:

```
Label:
    size_hint_x: None
    offset: '10dp'
    width: self.texture_size[0] + self.offset

Label:
    size_hint_x: None
    width: self.texture_size[0] + 10 * Metrics.dp
```

Implementation strategy
---------------------------

We create a global `EventObservers` in `_metrics.pyx`, like those used by `EventDispatcher` and `Property` for binding, and every time a new `NumericProperty` of a `EventDispatcher` object is created and "`linked`", we bind to the global `EventObservers`. Such that if that is dispatched, these properties will be called and their value re-evaluated and dispatched (if it changed).

To make sure we don't prevent garbage collection, we only store a weakref. Furthermore, if the widget dies, the binding is undone. 

When `Metrics` density/dpi/fontscale changes, then we cause a dispatch to the `EventObservers`. And similarly, when Window's `dpi` changes, it causes a change in `Metrics.dpi`, which causes the dispatch.

I considered instead, when `Window.dpi` changes, to simply dispatch all widgets in the tree and all properties in the widget that use `dp` as this would be much faster for the widget creation phase, but this would leave widgets that are not currently in the tree out (e.g. a pop up that is hidden).

Performance 
--------------

The primary concern is to slow down creation of `NumericProperty` bearing widgets as little as possible, and similarly to slow down garbage collection when these widgets die as little as possible.  This is traded off against making any `dpi` updates as fast as possible, because I consider that to happen infrequently, but widget creation happens a lot.

To accomplish this I added `_metrics.pyx/pxd`, and moved the metrics related code from `properties.pyx` over there. I couldn't convert `metrics.py` to `metrics.pyx` because that leads to circular imports (metrics needs _event/properties, properties needs metrics), so the new stuff was added to `_metrics.pyx` instead.

So, the binding and unbinding happens within cython which makes it super fast. Additionally, I added a cython only way to fbind (`fbind_existing_callback`), this allows a more direct binding from cython and is the fastest way to bind. Next, unbinding was also sped up. Currently, the fastest unbinding happens with `funbind_uid`, however, this is `O(n)` still. So I added a new way to unbind if you have the callback (API is only internal, although perhaps we should change `fbind` so it can be used) that is a very speedy `O(1)`.

Using the benchmark method from https://github.com/kivy/kivy/pull/7292, I compared `EventDispatcher` instantiation with master, if the class a `NumericProperty` vs a `StringProperty`:

```
Name (time in ns)                          Min                     Max                  Mean                StdDev                Median      

NumericProperty:
(master)                               940.0000 (2.60)      12,030.0000 (2.48)     1,258.1731 (2.47)       646.0043 (2.06)     1,040.0000 (2.55)
(dpi)                                1,370.0000 (3.79)      71,840.0000 (14.80)    1,877.1397 (3.69)       919.9020 (2.94)     1,540.0000 (3.78)

StringProperty:
(dpi)                                  880.0000 (2.43)      10,360.0000 (2.13)     1,156.8685 (2.28)       572.0594 (1.83)       980.0000 (2.40)
(master)                               920.0000 (2.54)      10,650.0000 (2.19)     1,195.3162 (2.35)       576.8812 (1.84)     1,020.0000 (2.50)
```
The results show that adding a `NumericProperty` is slower `1.9` vs `1.3` or about 0.6 us and my manual tests show this similarly. I similarly compared how long it takes to create a `Widget`
```
Name (time in us)         Min         Max     Mean   StdDev   Median
--------------------------------------------------------------------
dpi                    32.2000  496.2000    54.6560  26.5227  46.4000
master                 29.4000  2,095.9000  49.4672  29.9499  40.5000
```

and it's about 5 or 6 us slower. I don't think this is a big deal in the scheme of things, but it is slower.

Additional changes
---------------------

To be able to add this feature, I had to do some refactoring as follows:

* Changed instantiation of cython classes used in binding to use `__new__`, this allows cython to skip having to call `__init__`, which makes it faster. I changed this for all the bindings and creating `PropertyStorage`. This is ok, because these classes don't do anything in `__init__`.
* Previously, we had one `PropertyStorage` class, so variables used by every property type was added there, making the class huge. Instead, I made a per property type `PropertyStorage`, e.g. `NumericPropertyStorage` that inherit from `PropertyStorage`. And split up the property specific variables to them making `PropertyStorage` small.
* Changed the global `Metrics` variable to be a proxy, like `Clock`.
* Fixed a bug. We allow a property to be renamed and it's `PropertyStorage` to be reused. E.g.
   ```py
   class A(EventDispatcher):
       name = ObjectProperty()
       other_name = prop
   ```
   But this resulted in 
   ```py
   class A(EventDispatcher):
       name = ObjectProperty()
       name = NumericProperty()
   ```
   also reusing the internal `PropertyStorage`. So I fixed it to test to make sure it's only reused if the property object is the same object.

### DIFF
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -1022,6 +1022,24 @@ cdef class EventObservers:
             self.last_callback = new_callback
         return uid
 
+    cdef inline BoundCallback fbind_callback(
+            self, object observer, tuple largs, dict kwargs, int is_ref):
+        '''Similar to bind, except it accepts largs, kwargs that is forwards.
+        is_ref, if true, will mark the observer that it is a ref so that we
+        can unref it before calling.
+        '''
+        cdef BoundCallback new_callback = BoundCallback(
+            observer, largs if largs else None, kwargs if kwargs else None,
+            is_ref)
+
+        if self.first_callback is None:
+            self.last_callback = self.first_callback = new_callback
+        else:
+            self.last_callback.next = new_callback
+            new_callback.prev = self.last_callback
+            self.last_callback = new_callback
+        return new_callback
+
     cdef inline void unbind(self, object observer, int stop_on_first) except *:
         '''Removes the observer. If is_ref, he observers will be derefed before
         comparing to observer, if they are refed. If stop_on_first, after the
@@ -1088,6 +1106,13 @@ cdef class EventObservers:
             if callback.lock != deleted:
                 self.remove_callback(callback)
             return
+
+    cdef inline object unbind_callback(self, BoundCallback callback):
+        '''Remove the callback identified by the uid. If passed uid is None,
+        a ValueError is raised.
+        '''
+        if callback.lock != deleted:
+            self.remove_callback(callback)
 
     cdef inline void remove_callback(self, BoundCallback callback, int force=0) except *:
         '''Removes the callback from the doubly linked list. If the callback is
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -1025,12 +1025,7 @@ cdef class EventObservers:
             callback = callback.next
 
         new_callback = self.make_callback(observer, None, None, is_ref)
-        if self.first_callback is None:
-            self.last_callback = self.first_callback = new_callback
-        else:
-            self.last_callback.next = new_callback
-            new_callback.prev = self.last_callback
-            self.last_callback = new_callback
+        self.fbind_existing_callback(new_callback)
 
     cdef inline object fbind(self, object observer, tuple largs, dict kwargs,
                                int is_ref):
@@ -1042,13 +1037,7 @@ cdef class EventObservers:
         self.uid += 1
         cdef BoundCallback new_callback = self.make_callback(
             observer, largs, kwargs, is_ref, uid)
-
-        if self.first_callback is None:
-            self.last_callback = self.first_callback = new_callback
-        else:
-            self.last_callback.next = new_callback
-            new_callback.prev = self.last_callback
-            self.last_callback = new_callback
+        self.fbind_existing_callback(new_callback)
         return uid
 
     cdef inline BoundCallback fbind_callback(
@@ -1059,13 +1048,7 @@ cdef class EventObservers:
         '''
         cdef BoundCallback new_callback = self.make_callback(
             observer, largs, kwargs, is_ref)
-
-        if self.first_callback is None:
-            self.last_callback = self.first_callback = new_callback
-        else:
-            self.last_callback.next = new_callback
-            new_callback.prev = self.last_callback
-            self.last_callback = new_callback
+        self.fbind_existing_callback(new_callback)
         return new_callback
 
     cdef inline void fbind_existing_callback(self, BoundCallback callback):
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -230,7 +230,8 @@ cdef class EventDispatcher(ObjectWithUid):
 
         # then auto register
         for event in events:
-            self.__event_stack[event] = EventObservers(1, 0)
+            self.__event_stack[event] = EventObservers.__new__(
+                EventObservers, 1, 0)
 
     def __init__(self, **kwargs):
         cdef basestring func, name, key
@@ -300,7 +301,8 @@ cdef class EventDispatcher(ObjectWithUid):
 
         # Add the event type to the stack
         if event_type not in self.__event_stack:
-            self.__event_stack[event_type] = EventObservers(1, 0)
+            self.__event_stack[event_type] = EventObservers.__new__(
+                EventObservers, 1, 0)
 
     def unregister_event_types(self, basestring event_type):
         '''Unregister an event type in the dispatcher.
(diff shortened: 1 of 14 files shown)


kivy/_event.pyx
```diff
@@ -1022,6 +1051,31 @@ cdef class EventObservers:
...
+        '''
+        cdef BoundCallback new_callback = self.make_callback(
+            observer, largs, kwargs, is_ref)
+
+        if self.first_callback is None:
+            self.last_callback = self.first_callback = new_callback
+        else:
+            self.last_callback.next = new_callback
+            new_callback.prev = self.last_callback
+            self.last_callback = new_callback
```
AndiEcker, line 1068: DRY: why not replace these 6 lines with a call of `fbind_existing_callback(new_callback)`

AndiEcker: thanks a ton, @matham for your great work and for to keep the kivy universe expanding <3

gottadiveintopython:
That's a lot of work. I have a question.
Does this PR mean the following functions should not be in the `global_maps`?

https://github.com/kivy/kivy/blob/3e37301b291cfbd8e4cde5f1d0e7ce739700d8da/kivy/lang/parser.py#L104-L105

Because If the user codes something like this:

```yaml
Label:
    size_hint_x: None
    width: self.texture_size[0] + dp(10)
```

it cannot react to the change of the screen density, right?

kivy/_event.pyx
matham: Good point, fixed.

matham:
> Does this PR mean the following functions should not be in the global_maps?

Your understanding if it is correct. It is true that if you use `dp(x)` directly, it won't be updated. However, I don't think we should remove it because it'll break backwards compat and removing it won't actually fix anything. Instead we will need to make it clear in the docs about the advantages of using `10dp` over `dp(10)`.

However, this will not really be an issue until a future PR, when `dp` defaults to some value before Window is imported. Because even with this PR, when dpi is initially computed, Window is imported so we get the correct system `dpi` value, even if you call `dp(10)`. It's only an issue if you choose to manually change the DPI because then it won't be updated.

So, for the future, It'll be a backwards-compat break when we don't aggressively import `Window` upon computing `dp`. And we would need to make sure to make it clear in the docs about this issue when that happens. But as a point of why I think it's probably not going to be a big issue, by the time you create a Widget and kv rules are applied, you probably have a Window open and the correct `dpi` value. So it'll only be an issue if you manually call `dp(x)` in python before than.

I considered changing kv-lang so `width: self.texture_size[0] + dp(10)` would cause the rule to be tirggered when `dpi` is updated. However, I think this would be a special case and really not consistent with the rest of kv-lang so I dropped it.

If you really want to do that, I'd suggest doing something like:

```
Label:
    size_hint_x: None
    offset: '10dp'
    width: self.texture_size[0] + self.offset
```

Or a somewhat sillier way,
```
<Offsets>:
    dp1: '1dp'
    dp2: '2dp'
    dp10: '10dp'

Label:
    size_hint_x: None
    width: self.texture_size[0] + offsets.dp10
```

Or, a more realistic fix is to add `dp` etc. as properties of `Metrics`, because after all they are just a scaling multiplication factor, so then you could do `width: self.texture_size[0] + 10 * Metrics.dp`. I think I'll add this change.

gottadiveintopython:
> However, I don't think we should remove it because it'll break backwards compat and removing it won't actually fix anything

I agree. I should have said in different way e.g. "The following functions should be deprecated when they are called from kv."
But looks like it was an unnecessary worry because of:

> by the time you create a Widget and kv rules are applied, you probably have a Window open and the correct dpi value.

My concern has gone, thank you.

> a more realistic fix is to add dp etc. as properties of Metrics

:+1:

matham:
Ok, I added the feature I mentioned. thanks for bringing it up. Now both of the following are the same:

```
Label:
    size_hint_x: None
    offset: '10dp'
    width: self.texture_size[0] + self.offset

Label:
    size_hint_x: None
    width: self.texture_size[0] + 10 * Metrics.dp
```

PR #7296, opened 2020-12-30
Added on_destroy event to Widget class

quitegreensky:
New `on_destroy` event is added to `Widget` class and fires when the widget is removed from tree.


matham:
Thanks for the code contribution, however, the canonical way to do this in Kivy is to bind to `parent` and when it changes check if it's None. If it is None, we know that the parent has been removed. I.e.

```
Label:
    on_parent: if self.parent is None: xyz
```

So I don't think we need to add a special event just for this condition.

PR #7297, opened 2021-01-01
Disable the use of root logger

0x4A-0x41-0x4B:
Configure kivy's logger not to use the root logger as it may collide with other modules or user loggers.

Maintainer merge checklist

### DIFF
--- a/kivy/logger.py
+++ b/kivy/logger.py
@@ -359,9 +359,6 @@ Logger = logging.getLogger('kivy')
 Logger.logfile_activated = None
 Logger.trace = partial(Logger.log, logging.TRACE)
 
-# set the Kivy logger as the default
-logging.root = Logger
-
 # add default kivy logger
 Logger.addHandler(LoggerHistory())
 file_log_handler = None


matham:
I don't think it's quite as simple as that, unfortunately :(

The idea in the past was to capture all the logging sources, since we were operating under the assumption that Kivy is the framework so everything that is logged should be recorded automatically in the Kivy log files. With this change it won't happen anymore, only if you explicitly used Kivy's `Logger`.

So first, this will be a backward compat break. We can break compat, however, we need to add upgrade paths for people who want this functionality. 

Additionally, this is not the only issue with the logger module. Logger creates other global variables and hooks which should be optional somehow. E.g. recently I ran into an issue with the line `sys.stderr = LogFile('stderr', Logger.warning)` wherre pytest was looking for a `fp` attribute on `sys.stderr` (although perhaps this was a pytest issue). So ideally, we'd come up with a single approach to resolve all the issues.

E.g. one thought is to make Kivy `App` a context manager so that when you do:

```
with app:
   ...
```
then we activate this functionality while the app is running, but you disable it beforehand using a app attribute or something. I.e. until that moment Kivy accumulates a limited amount of logs in memory and when you enter the context, kivy makes itself the global logger (if enabled) and dumps all accumulated logs so far. Until the context is exited. Such a approach may also work with Builder/Factory/Cache/Clock and other kivy globals.

0x4A-0x41-0x4B:
Logging with a context manager in that case may make sense. The easiest solution would still be to either:
1. Check for a environ variable (like the KIVY_NO_FILELOG, i would suggest something like "KIVY_ROOT_LOGGER"), and **override the root logger if not set** (This would not break backwards compatibility in any way)
OR:
2. **Don't override the root logger** (which breaks backwards compatibility). If the kivy logger is needed it can be imported (from kivy.logger import Logger)

Opting for the first solution seems to be the best of both worlds (In addition, the context manager could be implemented).

matham:
`KIVY_NO_FILELOG` would certainly solve this specific issue, however, it does not solve in in a way that moves kivy to be a more stable and robust framework.

For starters, this is python, so it's not typical to be using environmental variables to control very specific code paths. Yes kivy makes use of environmental variables to control higher level features (e.g. the specific provider), but I don't think using env variables here makes sense as a pythonic way to control program flow.

But also, as I mentioned, this is part of a larger issue with aggressive imports. So the solution needs to be comprehensive for logger, not just a band-aid for this particular logger sub-problem. E.g. much of the issues we ran into in https://github.com/kivy/kivy/pull/7307 making the builds and tests not fail is because of aggressive imports and kivy's making use of global variables, which may or may not be set. So calling any function results in indeterminate side effects that is hard to predict and then figure out what is going wrong if something breaks because of e.g. circular imports/side effects.

Hence, I think whatever we do needs to be comprehensive for logger and not just solve this specific issue. Because I'd rather we keep the status-quo than add a band-aid that doesn't really solve the overall issue and may not even be useful in most situations. E.g. by default it would still set the logger when imported, so it won't help people causally importing kivy and getting their logger changed. And if we are going to break backward compat and not claim the logger, then we need to add a way for people to enable back the current functionality in a systematic way, but not with a env variable.

As for option 2, we can't just do that, because the current behavior is useful, in particular for deployment to get the error logs in case something outside Kivy crashes (so kivy's logger can record it to file). So there will need to be a way to easily enable the current functionality.

matham:
I'm going to close, since as I explained, removing it is not correct. But if you think of a way to incorporate my feedback, please create a new PR.

Issue #7094, opened 2020-09-19
Opening a matplotlib figure, causes kivy desktop app to shrink (size & font)


abedhammoud:
**Software Versions**
* Python: 3.8.5
* OS: win10
* Kivy: 2.0.0rc3
* Kivy installation method: pip

**Describe the bug**
When I call any of the matplotlib plotting routines, as the matplotlib figure is displayed, the Kivy desktop application shrinks in size (window size and font)

**Expected behavior**
The display of the matplotlib should not affect the size/font of the Kivy application.

**To Reproduce**
Please find the code below.

**Code and Logs and screenshots**
```python
import kivy
from kivy.app import App
from kivy.uix.boxlayout import BoxLayout

import matplotlib.pyplot as plt

kivy.lang.Builder.load_string('''

<DataView>:

    orientation: 'vertical'
    spacing: 5
    padding: 5

    Label:
        text: 'HELLO WORLD'
''', filename='dataview.kv')

class DataView(BoxLayout):

    def __init__(self, **kwargs):
        super().__init__(**kwargs)

class TestApp(App):

    def build(self):
        DV = DataView()

        # Commenting/uncommenting this line causes the kivy application
        # font size / window size to change.
        fig, ax = plt.subplots(2, 2)

        return DV

if __name__ == '__main__':
    TestApp().run()
```

**Additional context**

Kivy debugging dump:

[INFO   ] [deps        ] Successfully imported "kivy_deps.gstreamer" 0.2.0
[INFO   ] [deps        ] Successfully imported "kivy_deps.glew" 0.2.0
[INFO   ] [deps        ] Successfully imported "kivy_deps.sdl2" 0.2.0
[INFO   ] [Kivy        ] v2.0.0rc3, git-20c14b2, 20200615
[INFO   ] [Kivy        ] Installed at "C:\Users\abed\anaconda3\lib\site-packages\kivy\__init__.py"
[INFO   ] [Python      ] v3.8.5 | packaged by conda-forge | (default, Sep 16 2020, 17:19:16) [MSC v.1916 64 bit (AMD64)]
[INFO   ] [Python      ] Interpreter at "C:\Users\abed\anaconda3\python.exe"
[INFO   ] [Factory     ] 185 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_sdl2, img_pil (img_ffpyplayer, img_gif ignored)
[INFO   ] [Window      ] Provider: sdl2
[...]

Issue #3705, opened 2015-10-18
Kivy blurs pixels in Windows when UI scaling is enabled


BenMcLean:
In Windows 10, my resolution is 1920x1080 and UI scaling is set to 175%. With these settings, (and possibly others) the pixels in kivy apps appear blurred around the edges. Not sure if this is true across all Windows versions or just Windows 10.

Talking about this in IRC led me to discover that using kivy.metrics.dp(1) returns 1.0 and should ((unless I am mistaken) return 1.75 in this case.

Kivy needs to be made sensitive to where that information is exposed in the Windows API to fix this, so that this type of scaling can be applied with the nearest neighbor algorithm to keep the pixels sharp.

I can supply code that reproduces this issue if it will help.

As a temporary workaround, I have checked the "Disable display scaling on high DPI settings" option in the Windows Compatibility settings on my python.exe for all users. That results in sharp unscaled pixels. It would still be nice to get this fixed so we can have sharp scaled pixels.

BillBridge:
I am having the same problems. See http://stackoverflow.com/questions/38049695/kivy-gui-does-not-adapt-to-customize-display-on-windows?noredirect=1#comment63553443_38049695

What is the status for Windows as it seems to be resolved for OsX?

verata-veritatis:
Still no updates huh? Having issues with this on Win10 currently.

kuguma:
I also encountered the same problem on Win10. From the metrics.py dpi acquisition code, it seems that the code for windows is not implemented yet.

In conclusion, I was able to solve this problem via ctypes.
My code gets the window handle from tkinter and gets the DPI from there.
It would be better to get the handle from SDL. But I don't know how.
If someone tells me, I can send a pull request.

dangpzanco:
This is still a problem after +4 years....

CoziestYew804:
@kuguma can you please explain how you did that ?

probablyfine:
I've got this blurry pixel issue as well on Windows 10 with UI scaling at 150%. Another related problem is that the Window core module reports the wrong size/position.

@kuguma is correct, on Windows the metrics.py code always returns 1.0 in its `density` method. Based on their suggestion to solve this using ctypes (plus some googling), I hacked together a fix. I don't know how "proper" this is; there could be issues I'm not aware of and I don't know if it will work in every environment. Hopefully someone who understands this better can weigh in and let me know if this is a reasonable approach.

[...]

PR #7299, opened 2021-01-02
Window: Handle DPI Windows messages until SDL2 handles them

matham:
Maintainer merge checklist

Fixes https://github.com/kivy/kivy/issues/3705, fixes https://github.com/kivy/kivy/issues/7094.

**TLDR**; to use these changes, you have to make our window DPI aware until SDL2 does it for us. Just do the following before you import Kivy:
```py
from ctypes import windll
windll.user32.SetProcessDpiAwarenessContext(-4)
```

This PR sits on top of https://github.com/kivy/kivy/pull/7293. So to see the changes specific to this PR, only look at the last commit. This should **not** be merged until https://github.com/kivy/kivy/pull/7293 is merged.

Background
------------

The issue is that SDL2 does not handle any DPI awareness issues on Windows. So as the window DPI changes (e.g. scaling changed or monitor changed), SDL2 pretends it never happened. This leads to the above linked issues because there's no way to automatically adjust the density/dpi, even if the window is capable of adjusting to it, because SDL doesn't report it.

Some background. On Windows, a window can be either DPI aware, in which case Windows tells it when its DPI needs to change. Or the window is DPI unaware and Windows will scale the window in response to DPI changes, leading to blurry text. A window can be DPI aware in two ways
1. it reads the current DPI when the window starts - afterwards it becomes DPI unaware. I.e. if the scaling changes, Windows will scale the window, again leading to blurry text. This is the old version of DPI-aware.
2. The window is always DPI aware and is told when the DPI changes every time and Windows does no scaling. This is the new method and is set using the tldr code above. It's only available in Windows 10+.

SDL2
-----

Unfortunately, SDL2 is completely DPI unaware and it ignores all DPI messages. This means that our text will either look blurry, as in https://github.com/kivy/kivy/issues/3705. Or, if somehow it's made aware, like in https://github.com/kivy/kivy/issues/7094 when matplotlib-qt sets that mode, or if a user sets it e.g. in the above tldr code, Kivy will still not respond to DPI changes but Windows will stop scaling the window leading to tiny text etc.

This PR
--------

This PR does the following: **if** the Kivy window is DPI aware, either because you set it with the tldr code above or because matplotlib-qt set it, then Kivy **will** be responsive and update its UI with appropriate scaling factors as the DPI changes. It's done by hooking into the Window message queue.

Caveats
---------

However, you still need to set the window to be DPI-aware. We can't do that as that will likely lead to unintended consequences. And in any-case, this PR is only a stop-gap until SDL2 [gains support](https://bugzilla.libsdl.org/show_bug.cgi?id=3281) for DPI awareness and then the changes from this PR as it relates to the Windows message queue can be removed. So we can leave to them the design decisions as it relates to making the window DPI aware. But I'm open to the idea of us setting it in the first place.

We could add a config option (`dpi_aware`) that defaults to true and makes the window DPI aware, but that can be turned off if needed.

Personally, I have been using a 4k monitor on Windows side by side with a 2k monitor and the scaling issues has always been annoying. Hopefully this should make it better.


probablyfine:
Wow, this is super cool- thank you for all the work you put into this! Got my hands on another monitor for testing and it seems to be working great for me. I can clearly see the UI rescale as I drag it between monitors, and text always looks clear and crisp. Thank you again!

probablyfine:
I've been playing around with this a bit today and I'm running into an odd issue. I'm using the `Window.top` and `Window.left` properties in my application, and it seems like these become invalid once the window is dragged to a second monitor with a different DPI. I think this makes sense since the documentation suggests these are SDL2 properties, and so it'd make sense that they wouldn't track DPI changes. So I thought maybe I could just grab the current DPI and manually correct those two properties before using them.

However, when I tried to do this I found I was unable to get an updated DPI. I tried a few things, but I always got the same value before and after dragging the window to the other monitor:
- `Metrics.get_density(force_recompute=True)`
- `Metrics.get_dpi(force_recompute=True)`
- I tried just calling `dp(1)` since I thought that should give a different answer if the DPI changes (is that correct?)
- I tried getting the DPI manually without asking kivy, using: `windll.user32.GetDpiForWindow(windll.user32.GetActiveWindow())`
- I tried printing a message to console by adding a print statement when the WM_DPICHANGED message is received, at [this line](https://github.com/matham/kivy/blob/3878b1a05116a8a14f2c327f69678a1c79d0af7c/kivy/core/window/window_sdl2.py#L869), but the message never gets printed.

I am calling `windll.user32.SetProcessDPIAware(-4)` before importing kivy, and visually everything looks perfect. Any idea what I might be doing wrong?

matham:
I'm pretty mystified now. Yesterday, I was able to trigger a `WM_DPICHANGED` message whenever I dragged the window between monitors or changed the scaling while Kivy was open. But today, no matter what I do, I can't get `WM_DPICHANGED` to trigger.

I think Windows must still be pretty buggy. In fact, I have noticed for a quite a while now that when I drag firefox to my 4k monitor, it'll disappear off screen and I can't get it back. Now I understand why that happens, because Windows is returning a negative Window size and Firefox is following Windows reccomendations to position the window upon a DPI change to a Windows provided pos in the `WM_DPICHANGED` message. I ignored it because it was sometimes negative, but I guess firefox doesn't.

So maybe Windows was in a specific configuration when I was able to get the message, but I must have done something in between to break that. And I did try restarting, unplugging the monitor etc to get it back but no luck.

probablyfine:
I think I figure it out. In the tldr code you posted, it's actually using the system-wide DPI awareness function, not the per-monitor. It should be: `windll.user32.SetProcessDpiAwarenessContext(-4)`

I replaced your line with that one and now the `WM_DPICHANGED` message is coming through as intended. One thing that helped tip me off was finding out that you can view DPI awareness in the task manager. You can see it in the 'Details' tab, you just have to tell it to display the 'DPI Awareness' column. I saw that python was running in 'system' rather than 'per-monitor' mode and re-checked the docs on these functions.

There's a quote on the doc pages for the `WM_DPICHANGED` message that might give a hint as to what was happening: 

> It may be received on certain DPI changes if your top-level window or process is running as DPI unaware or system DPI aware, but in those situations it can be safely ignored.

So I wonder if you're right that Windows was in some weird state/configuration where that message was coming through even though the wrong awareness level was set, but maybe in that state it isn't guaranteed that the message will be sent.

matham:
Right, that was me typoing because I did have it right yesterday with `SetProcessDpiAwarenessContext`. But then I tried the system mode using `SetProcessDPIAwareness`, and I thought the other version simply had `Dpi` in lowercase 😬 Thanks for figuring it out.

But even with that, I remember Windows getting the height wrong (it was a very large negative number) or it was zero, as it was dragged into the other monitor. But it was not an issue if I just changed the scaling. So instead I ignored the Windows suggested resizing and just resized it based on the DPI ration changed.

probablyfine:
I saw this had been merged into the master branch, and I needed a new venv so thought I'd give it a try using the latest master. I'm having a weird problem: when I use the TLDR code above, kivy fails to start with the following: `Unable to get a Window, abort.`

If I comment out those two lines, everything works fine (albeit blurry since not DPI-aware). I don't seem to have this issue with your `matham:dpi-touch` branch.

I installed kivy using: `python -m pip install "kivy[base] @ https://github.com/kivy/kivy/archive/master.zip"`

And the full output when I try to run looks like:
```
[WARNING] [Config      ] Older configuration version detected (22 instead of 23)
[WARNING] [Config      ] Upgrading configuration in progress.
[INFO   ] [Logger      ] Record log in C:\Users\steve\.kivy\logs\kivy_21-01-24_35.txt
[INFO   ] [deps        ] Successfully imported "kivy_deps.angle" 0.3.0
[INFO   ] [deps        ] Successfully imported "kivy_deps.glew" 0.3.0
[INFO   ] [deps        ] Successfully imported "kivy_deps.sdl2" 0.3.1
[INFO   ] [Kivy        ] v2.1.0.dev0
[INFO   ] [Kivy        ] Installed at "C:\Users\steve\Documents\kivy-master-01-24-21\lib\site-packages\kivy\__init__.py"
[INFO   ] [Python      ] v3.7.1 (v3.7.1:260ec2c36a, Oct 20 2018, 14:57:15) [MSC v.1915 64 bit (AMD64)]
[INFO   ] [Python      ] Interpreter at "C:\Users\steve\Documents\kivy-master-01-24-21\Scripts\python.exe"
[INFO   ] [Logger      ] Purge log fired. Processing...
[INFO   ] [Logger      ] Purge finished!
[INFO   ] [Factory     ] 187 symbols loaded
[INFO   ] [Image       ] Providers: img_tex, img_dds, img_sdl2, img_pil (img_ffpyplayer ignored)
[INFO   ] [Window      ] Provider: sdl2
[CRITICAL] [App         ] Unable to get a Window, abort.
```
Any idea what might be going wrong?

matham:
I'd try undoing the change in the linked file https://github.com/kivy/kivy/pull/7299/commits/3878b1a05116a8a14f2c327f69678a1c79d0af7c#diff-108f63aba6f107e58f19f67019baf7c154ea2ca730f9803ba4b2da467cba8533. It's possible that density and therefore dpi is read before Window is ready for Windows leading to a circular import.

probablyfine:
Unfortunately no luck with that suggestion. It looks like the problem is occurring before then; if I add a print statement around those lines (having commented out the setting of `value`) it only gets printed when the TLDR code is disabled.

matham:
I've tried master both in my existing env and from a clean virtualenv, but I had no issues. Could you try a clean virtualenv?

probablyfine:
Thanks for trying it out. I gave it another try in a fresh virtual env, but still no luck. Here's how I set things up first:
```
python -m pip install --upgrade pip setuptools virtualenv
python -m virtualenv kivy-master-01-25-21
kivy-master-01-25-21\Scripts\activate
python -m pip install "kivy[base] @ https://github.com/kivy/kivy/archive/master.zip"
python dpi-test-app.py
```
And my minimal `dpi-test-app.py`:
```
from ctypes import windll
windll.user32.SetProcessDpiAwarenessContext(-4)
import kivy
from kivy.app import App

class TestApp( App):
    def build( self):
        return Label( text='hello')
        
if __name__ == '__main__':
    TestApp().run()
```
which crashes as before unless I comment out the second line.

I think your previous suggestion is exactly right, now I'm wondering if I misunderstood your suggestion. So, I can hack my way to the correct functionality using this:
```
from ctypes import windll
import kivy
from kivy.app import App
from kivy.clock import Clock
from kivy.metrics import Metrics
from functools import partial

class TestApp( App):
    def build( self):
        my_label = Label( text='hello')
    
        hwnd = windll.user32.GetActiveWindow()
        print( hwnd)
        windll.user32.SetProcessDpiAwarenessContext(-4)
        dpi = float(windll.user32.GetDpiForWindow(hwnd))
        print( dpi)
        def set_density(density, *args):
            Metrics.density = density
        Clock.schedule_once(partial(set_density, dpi/96.0), -1)
        
        return my_label
        
if __name__ == '__main__':
    TestApp().run()
```
And when that runs, I get the window handle "1180862" and dpi "144.0" printed to console, and the app starts perfectly with the correct DPI. However, if I change this slightly so that the `Label` is created _after_ the DPI-related lines, I get "0" and "0.0" printed to console for hwnd and dpi, respectively, and then it of course crashes as before.

So again I think you were exactly right. I had understood (maybe wrongly) your suggestion to essentially mean: go into kivy/metrics.py, and comment out these two lines at 262 and 263:
```
elif platform in ('macosx', 'win'):
    value = self.dpi / 96.
```

Is that what you had intended or did I get it wrong?

matham:
Yeah, I did intend for you to comment out those two lines. But if commenting out those lines doesn't help, why do you say I was right?

probablyfine:
Sorry the way I wrote it was confusing. Not _exactly_ right, but I think you must be on the right track. From the test I posted above it seems like it must be trying to get the window handle before one exists. So maybe not a circular import issue, but more generally a race condition.

matham:
Ok, I understand.

Here's another thing to try , comment out [this line](https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L894): `Clock.schedule_once(clock_callback, -1)`. It'll stop the new windows only functionality, but that will help narrow down the issue.

If that fixes the crash, try replacing `-1` with e.g. `0` or `1`. Perhaps the clock is executed before the window is ready causing the problem.

probablyfine:
I tried commenting out that line but it didn't change anything unfortunately.

I've been poking around a little and I think I found the point of failure, it seems to be at [this line](https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L300).

If I `pdb.set_trace()` at the start of that `try` clause and step through, the `hwnd` actually is valid. So that executes later than the same call in my second example above. Getting the DPI from the ctypes call also works correctly (144 on my system). But as soon as I try to assign something to `self.dpi` it crashes. 

So if I understand this correctly I think your original thought was actually right, the problem just happens in a different place.

matham:
Right, I thought I solved this issue [here](https://github.com/kivy/kivy/blob/master/kivy/core/window/__init__.py#L997) by delaying registering the `dpi` callback (`_reset_metrics_dpi`), but I guess it's still too early. 

Kivy's aggressive imports are the bane of my existence... :) That's why I did this in the first place to enable eventually to turn OFF these aggressive imports once DPI is dynamic.

Could you try replacing `self.fbind('dpi', self._reset_metrics_dpi)` with `Clock.schedule_once(lambda *x: self.fbind('dpi', self._reset_metrics_dpi) and self._reset_metrics_dpi(), 0)`?

probablyfine:
Okay, I tried that out. It actually didn't work, but I found the issue, it's nearby. It's [this line](https://github.com/kivy/kivy/blob/master/kivy/core/window/__init__.py#L909). I stepped through the __init__ method in the debugger and saw that fbind call that's too early. Commenting it out solves the issue so I think the later call is fine where it is. Wonder why it worked on your system?

matham:
Ahhhh I had moved that line below in `matham:dpi-touch` and then dutifully added the note about the circular import. But then when I had to manually fix merge conflicts from the two PRs, I must have overlooked this so got added to both places :(

https://github.com/kivy/kivy/pull/7347 should fix it. Thanks for tracking it down.

probablyfine:
Really glad to get this working, thanks for all your help.

lucasgerads:
I am trying to get this to work. But the text in my label still looks blurry. This is my test code: 
`from ctypes import windll
windll.user32.SetProcessDpiAwarenessContext(-4)
import kivy
from kivy.app import App

class TestApp( App):
    def build( self):
        return Label( text='this is blurry')
        
if __name__ == '__main__':
    TestApp().run()`

Maybe I have to do more here? You mention that I have to "set the window to be DPI-aware". I am not quite sure how to do that.

lucasgerads:
Ok, it seems like I fixed it. I am not sure what exactly is going on, but when I did the following: 

´´´
from ctypes import windll
windll.user32.SetProcessDpiAwarenessContext(-4)
lastError = windll.kernel32.GetLastError()
print(lastError)
´´´

I got the error ´87´ which corresponds to ´ERROR_INVALID_PARAMETER´. So I assumed something had to be wrong with the -4. I dug around the windll and found the following fix: 

´´´
from ctypes import windll, c_int64
windll.user32.SetProcessDpiAwarenessContext(c_int64(-4))
´´´

PR #7303, opened 2021-01-02
Camera: Fix AVFoundation provider to release the camera, start it async, and check if started before stopping it

misl6:
This PR fixes the following issues on AVFoundation Camera provider:

- Camera not released.
- `Camera` destructor crashes the App if the Camera is already stopped.
- UI blocked for ~ 1.5s during  `[mCaptureSession startRunning]` (See https://developer.apple.com/documentation/avfoundation/avcapturesession/1388185-startrunning?language=occ "Important" section for reference)

Maintainer merge checklist

### DIFF
--- a/kivy/core/camera/camera_avfoundation.pyx
+++ b/kivy/core/camera/camera_avfoundation.pyx
@@ -59,7 +59,9 @@ class CameraAVFoundation(CameraBase):
 
     def _release_camera(self):
         cdef _AVStorage storage = <_AVStorage>self._storage
-        avf_camera_deinit(storage.camera)
+        if storage.camera != NULL:
+            avf_camera_deinit(storage.camera)
+            storage.camera = NULL
 
     @property
     def _scheduled_rate(self):
--- a/kivy/core/camera/camera_avfoundation.pyx
+++ b/kivy/core/camera/camera_avfoundation.pyx
@@ -49,10 +49,17 @@ class CameraAVFoundation(CameraBase):
         self._framerate = 30
         super(CameraAVFoundation, self).__init__(**kwargs)
 
+    def __del__(self):
+        self._release_camera()
+
     def init_camera(self):
         cdef _AVStorage storage = <_AVStorage>self._storage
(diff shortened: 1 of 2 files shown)


kivy/core/camera/camera_avfoundation.pyx
```diff
@@ -49,11 +49,18 @@ class CameraAVFoundation(CameraBase):
...
+        self._release_camera()
+
     def init_camera(self):
         cdef _AVStorage storage = <_AVStorage>self._storage
         storage.camera = avf_camera_init(
             self._index, self.resolution[0], self.resolution[1])
 
+    def _release_camera(self):
+        cdef _AVStorage storage = <_AVStorage>self._storage
+        avf_camera_deinit(storage.camera)
```
matham, line 62: Is it not a problem calling `delete (Camera *)(camera);` if the camera object has been previously deleted already? We probably need to set `storage.camera` to NULL and check whether it's NULL before deleting?

misl6: Well, the only way to call `delete (Camera *)(camera);` on kivy side is when (from user or gc) `__del__` is called on `kivy.core.camera.camera_avfoundation.CameraAVFoundation`.
Upon deletion, `CameraAVFoundation` doesn't exist anymore, so there's no chance that someone is able to call the `avf_camera_deinit again.

matham: Oh, I assumed that there's somewhere a stop camera method that may also "release" the camera.

Still, even if one doesn't exist right now, someone may add one in the future so it's probably best to set it to null and also first check if it's null. It's good practice anyway because it makes it clear. Because a null pointer access gives a clearer error then accessing invalid memory.

misl6: Did it. It makes sense.
Also, I see a lot of potential issues on not checking `storage.camera` in other methods, but that's another story and another PR :)

matham (approved):

PR #7304, opened 2021-01-02
Window clearcolor default value to tuple

AndiEcker:
Ensure tuple value for Window.clearcolor.

The Window clearcolor property documentation (see kivy.core.window, line 473)
is declared as a tuple and not as a list.

Most places in the kivy code base are using tuples for Window.clearcolor.

Maintainer merge checklist


matham:
Why convert to tuple? You can see in the code (`if type(value) not in (list, tuple):`) that it accepts either a tuple or list. Is it documented somewhere that it must be a tuple? If so, we should probably correct it to say it can accept either.

AndiEcker:
My intention was to use tuples for colors because they would be more efficient than using lists, and tuples are used for colors in most places of the Kivy codebase (in the documentation as well as in the used default values).

Diving deeper I (e.g. into graphics/gl_instructions.pyx, lines 60-82) I can see that there are some more places where lists are used.

So I agree with you to leave the `Window.clearcolor` default value like it was (also to not break backward compatibility).

And I can see also the advantage of using lists when you want to change only one component of a color.

Thanks for your time/review and sorry for the noise.

PR #7307, opened 2021-01-03
Remove purge log's randomized behavior

0x4A-0x41-0x4B:
Remove purge logs from being fired with a percent chance, use a deterministic approach instead.

Maintainer merge checklist


kivy/logger.py
```diff
@@ -150,47 +151,36 @@ class FileHandler(logging.Handler):
...
 
     def purge_logs(self):
-        '''Purge log is called randomly to prevent the log directory from being
-        filled by lots and lots of log files.
-        You've a chance of 1 in 20 that purge log will be fired.
-        '''
-        if randint(0, 20) != 0:
-            return
-        if not self.log_dir:
-            return
```
matham, line 160: Please add back this check. It's needed when Kivy is running in certain configurations where there's no config.

matham (changes requested):
Please also add unit-tests. Specifically, in the pytest-provided folder, create a sub-folder with files, and then manually call purge to make sure the supposed number of files are deleted, but the parent folder isn't etc.

kivy/logger.py
matham: This must be checked *before* you import config.

0x4A-0x41-0x4B: Check for self.log_dir is now issued before the import of config. Unittests were also created. (Including test logs in the test_logger subfolder)

kivy/tests/test_logger.py
```diff
@@ -0,0 +1,25 @@
...
+        log_path = pathlib.Path("test_logs")
+
+        Config.set("kivy", "log_dir", log_path)
+        Config.set("kivy", "log_maxfiles", maxfiles)
+
+        FileHandler().purge_logs()
+
+        current_files = [f for f in log_path.iterdir() if f.is_file()]
+
+        self.assertEqual(maxfiles, len(current_files))
```
matham, line 25:
```suggestion
import pytest
import pathlib
import time

@pytest.mark.parametrize('n', [0, 1, 5])
def test_purge_logs(tmp_path, n):
    from kivy.config import Config
    from kivy.logger import FileHandler

    Config.set("kivy", "log_dir", str(tmp_path))
    Config.set("kivy", "log_maxfiles", n)

    # create the default file first so it gets deleted so names match
    handler = FileHandler()
    handler._configure()
    # must close file so it can be deleted
    handler.fd.close()
    # wait a little so the timestamps are different for different files
    time.sleep(.001)

    names = [f'log_{i}.txt' for i in range(n + 2)]
    for name in names:
        p = tmp_path / name
        p.write_text('some data')
        time.sleep(.001)

    handler.purge_logs()

    # files that should have remained after purge
    expected_names = list(reversed(names))[:n]
    files = {f.name for f in tmp_path.iterdir()}
    assert expected_names == files

```

I re-wrote the test because we're using pytest when possible now. It has the benefit of it providing us with a temp directory for the test.

Also, we shouldn't create these files ahead of time, because it just adds space to the repo, instead, we can dynamically create them, like I do. So please remove the log files.

kivy/logger.py
```diff
@@ -150,47 +151,40 @@ class FileHandler(logging.Handler):
...
 
-            # sort by date
-            lst = sorted(lst, key=lambda x: x['ctime'])
+        Logger.info("Logger: Purge log fired. Processing...")
 
-            # get the oldest (keep last maxfiles)
-            lst = lst[:-maxfiles] if maxfiles else lst
-            Logger.info('Logger: Purge %d log files' % len(lst))
+        # Get all files from log directory and corresponding creation timestamps
+        files = [(item, item.stat().st_ctime) for item in log_dir.iterdir() if item.is_file()]
```
matham, line 173:
```suggestion
        files = [(item, item.stat().st_ctime) 
                   for item in log_dir.iterdir() if item.is_file()]
```
Line was too long (>80).

kivy/logger.py
```diff
@@ -150,47 +151,40 @@ class FileHandler(logging.Handler):
...
-            # now, unlink every file in the list
-            for filename in lst:
-                try:
-                    unlink(filename['fn'])
-                except PermissionError as e:
-                    Logger.info('Logger: Skipped file {0}, {1}'.
-                                format(filename['fn'], e))
+        for file in files[:-maxfiles]:
+            # More log files than allowed maximum,
+            # delete files, starting with oldest creation timestamp (or edit-timestamp on Linux)
```
matham, line 181:
```suggestion
            # delete files, starting with oldest creation timestamp
            # (or edit-timestamp on Linux)
```

Line too long.

matham (changes requested):

0x4A-0x41-0x4B (commented):
Fixed tests and added missing_ok to file deletion process as it could raise an exception otherwise.

matham:
`missing_ok` as a parameter does not exist in python 3.7 or lower. So we should catch that specific exception, if one is raised.

matham (approved):
Unfortunately RTD is failing and I can't see why. I'm going to merge as I think it could be a cache issue on their end (I think they cache builds for each PR).

PR #7326, opened 2021-01-11
Fix/disable kivy no args

pwoolvett:
Enable the environment value `KIVY_NO_ARGS` to contain a "falsy" value without
disabling kivy cli parsing.

Kivy only checks for `KIVY_NO_ARGS` environment variable to be existent, without
checking its value. Setting it to a "falsy" value (eg 0, "false", etc) would still disable
cli arg parsing, which is could be misleading.

With this fix, only setting it to one of ('true', '1', 'yes') actually disables it.

Maintainer merge checklist

### DIFF
--- a/kivy/tests/test_environ_cli.py
+++ b/kivy/tests/test_environ_cli.py
@@ -25,11 +25,12 @@ def _patch_env(*filtered_keys, **kw):
 
 
 def _kivy_subproces_import(env):
-    return subprocess.check_output(
+    return subprocess.run(
         shlex.split(f"{sys.executable} -c 'import kivy' --help"),
+        stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
         env=env,
-    ).decode("utf8")
+    ).stdout.decode("utf8")
 
 
 @pytest.mark.parametrize("value", SAMPLE_VALUES)
--- a/kivy/tests/test_environ_cli.py
+++ b/kivy/tests/test_environ_cli.py
@@ -9,7 +9,7 @@ import sys
 
 import pytest
 
-ENV_NAME="KIVY_NO_ARGS"
+ENV_NAME = "KIVY_NO_ARGS"
 KIVY_ENVS_TO_EXCLUDE = ('KIVY_UNITTEST', 'KIVY_PACKAGING')
 
 EXPECTED_STR = "Kivy Usage"
@@ -40,7 +40,7 @@ def _kivy_subproces_import(env):
 
 @pytest.mark.parametrize("value", SAMPLE_VALUES)
 def test_env_exist(value):
-    env = _patch_env(*KIVY_ENVS_TO_EXCLUDE, **{ENV_NAME:value})
+    env = _patch_env(*KIVY_ENVS_TO_EXCLUDE, **{ENV_NAME : value})
     stdout = _kivy_subproces_import(env)
 
     if value in TRUTHY:
--- a/kivy/tests/test_environ_cli.py
+++ b/kivy/tests/test_environ_cli.py
@@ -34,7 +34,7 @@ def _kivy_subproces_import(env):
 
 @pytest.mark.parametrize("value", SAMPLE_VALUES)
 def test_env_exist(value):
-    env = _patch_env(*KIVY_ENVS_TO_EXCLUDE, **{ENV_NAME : value})
+    env = _patch_env(*KIVY_ENVS_TO_EXCLUDE, **{ENV_NAME: value})
     stdout = _kivy_subproces_import(env)
 
     if value in TRUTHY:
--- a/kivy/tests/test_environ_cli.py
+++ b/kivy/tests/test_environ_cli.py
@@ -26,7 +26,7 @@ def _patch_env(*filtered_keys, **kw):
 
 def _kivy_subproces_import(env):
     return subprocess.run(
-        [sys.executable, "-c", "'import kivy'", "--help"],
+        [sys.executable, "-c", "import kivy", "--help"],
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
         env=env,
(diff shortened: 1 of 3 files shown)


kivy/tests/test_environ_cli.py
```diff
@@ -0,0 +1,54 @@
...
+        env=env
+    ).decode("utf8")
+
+
+@pytest.mark.parametrize("value", SAMPLE_VALUES)
+def test_env_exist(value):
+    env = _patch_env(**{ENV_NAME:value})
+    stdout = _kivy_subproces_import(env)
+
+    if value.lower() in TRUTHY:
```
matham, line 46: why `lower`?

kivy/tests/test_environ_cli.py
```diff
@@ -0,0 +1,54 @@
...
+SAMPLE_VALUES = {
+    *TRUTHY, *FALSY
+}
+
+def _patch_env(**kw):
+    env = {
+        k: v
+        for k, v in environ.items()
+        if "kivy" not in k.lower()
+    }
```
matham, line 28:
```suggestion
def _patch_env(*filtered_keys, **kw):
    env = {
        k: v
        for k, v in environ.items()
        if k not in filtered_keys
    }
```

And just add `KIVY_NO_ARGS` as args. Because I'd feel more comfortable if we removed specific env variables for specific tests (e.g. for new tests in the future), rather than all kivy variables for all tests.

matham (changes requested):

kivy/tests/test_environ_cli.py
pwoolvett: used to accept TRUE  as well, but removed it later in init code to be consistent. should be removed here, you're right

kivy/tests/test_environ_cli.py
pwoolvett: also added minimal kivy envs to implement test

kivy/tests/test_environ_cli.py
```diff
@@ -0,0 +1,52 @@
...
+SAMPLE_VALUES = {
+    *TRUTHY, *FALSY
+}
+
+def _patch_env(*filtered_keys, **kw):
+    env = {
+        k: v
+        for k, v in environ.items()
+        if k not in filtered_keys
+    }
```
matham, line 29:
```suggestion
    }

    return env
```

matham (changes requested):

kivy/tests/test_environ_cli.py
```diff
@@ -0,0 +1,52 @@
...
+import itertools
+from functools import partial
+from os import environ
+import shlex
+import subprocess
+import sys
+
+import pytest
+
+ENV_NAME="KIVY_NO_ARGS"
```
matham, line 12:
```suggestion
ENV_NAME = "KIVY_NO_ARGS"
```

kivy/tests/test_environ_cli.py
```diff
@@ -0,0 +1,52 @@
...
+    return subprocess.check_output(
+        shlex.split(f"{sys.executable} -c 'import kivy' --help"),
+        stderr=subprocess.PIPE,
+        env=env
+    ).decode("utf8")
+
+
+@pytest.mark.parametrize("value", SAMPLE_VALUES)
+def test_env_exist(value):
+    env = _patch_env(ENV_NAME, *KIVY_ENVS_TO_EXCLUDE, **{ENV_NAME:value})
```
matham, line 41:
```suggestion
    env = _patch_env(ENV_NAME, *KIVY_ENVS_TO_EXCLUDE, **{ENV_NAME : value})
```

matham (changes requested):

kivy/tests/test_environ_cli.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def _patch_env(*filtered_keys, **kw):
+    env = {k: v for k, v in environ.items() if k not in filtered_keys}
+    env.update(kw)
+    return env
+
+
+def _kivy_subproces_import(env):
+    return subprocess.check_output(
```
matham, line 28: It's failing because kivy returns a non-zero value (https://github.com/kivy/kivy/blob/master/kivy/__init__.py#L372). So you need to disable checking the error return value.

matham (changes requested):

kivy/tests/test_environ_cli.py
pwoolvett: roger

kivy/tests/test_environ_cli.py
```diff
@@ -0,0 +1,50 @@
...
+
+def _patch_env(*filtered_keys, **kw):
+    env = {k: v for k, v in environ.items() if k not in filtered_keys}
+    env.update(kw)
+    return env
+
+
+def _kivy_subproces_import(env):
+    return subprocess.run(
+        shlex.split(f"{sys.executable} -c 'import kivy' --help"),
```
matham, line 29:
```suggestion
        [sys.executable, "-c", "'import kivy'", "--help"],
```

matham (changes requested):
Formatting as a string is not safe on Windows because of backslash in path.

kivy/tests/test_environ_cli.py
```diff
@@ -0,0 +1,50 @@
...
+
+def _patch_env(*filtered_keys, **kw):
+    env = {k: v for k, v in environ.items() if k not in filtered_keys}
+    env.update(kw)
+    return env
+
+
+def _kivy_subproces_import(env):
+    return subprocess.run(
+        [sys.executable, "-c", "'import kivy'", "--help"],
```
matham, line 29:
```suggestion
        [sys.executable, "-c", "import kivy", "--help"],
```

matham (changes requested):
I was trying to be too clever. But it's better to let python handling quoting. This fixed it locally on windows. Hopefully the sdist failure was similar.

matham (approved):

matham:
Thanks for your patience :)

pwoolvett:
> Thanks for your patience :)

glad to contribute

PR #7339, opened 2021-01-22
Camera: Android camera focus mode fix

Auskas:
An Android camera is initialized with the focus mode set to 'continuous-picture' which leads to a crash if a particular camera does not support this mode (that's true for some front-facing cameras). 
The way I see it, the correct way to avoid the crash is checking the focus capabilities of a camera and setting the focus mode to 'continuous-picture' only if it is supported.

Maintainer merge checklist


matham (approved):
Seems ok to me, although I didn't test.

PR #7356, opened 2021-02-01
Widget: Widget.clear_widgets empty widget list does not remove all children

pydsigner:
Maintainer merge checklist
* [x] Applied the `Component: core-widget` label.
* [x] Applied the `api-break` label?

### DIFF
--- a/kivy/tests/test_widget.py
+++ b/kivy/tests/test_widget.py
@@ -38,6 +38,7 @@ class WidgetTestCase(unittest.TestCase):
         c3 = self.cls()
         root.add_widget(c1)
         root.add_widget(c2)
+        root.add_widget(c3)
         self.assertEqual(root.children, [c1, c2, c3])
         root.clear_widgets([c2])
         self.assertEqual(root.children, [c1, c3])
--- a/kivy/tests/test_widget.py
+++ b/kivy/tests/test_widget.py
@@ -47,6 +47,14 @@ class WidgetTestCase(unittest.TestCase):
         root.clear_widgets()
         self.assertEqual(root.children, [])
 
+    def test_clear_widgets_children(self):
+        root = self.root
+        for _ in range(10):
+            root.add_widget(self.cls())
+        self.assertEqual(len(root.children), 10)
+
+        root.clear_widgets(root.children)
+        self.assertEqual(root.children, [])
     def test_position(self):
         wid = self.root
         wid.x = 50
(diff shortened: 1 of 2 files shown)


pydsigner:
As noted in Discord discussion, this change *does* alter the API behavior, though not in a way that should break existing users.

matham:
Good catch about the possibility of someone passing in `children`.

Could you add unittests? You don't have to, but it would be nice if you added tests.

pydsigner:
> Good catch about the possibility of someone passing in `children`.

The case was being handled before, but not as well; if a list other than `children` was passed in, it would still be duplicated for no gain, and sequences other than lists (like iterators and sets) were being needlessly excluded.

> Could you add unittests? You don't have to, but it would be nice if you added tests.

Done.

kivy/tests/test_widget.py
```diff
@@ -30,6 +30,23 @@ def test_invalid_add_widget(self):
...
+        root.add_widget(c2, index=1)
+        root.add_widget(c3, index=2)
+        self.assertEqual(root.children, [c1, c2, c3])
+        root.clear_widgets([c2])
+        self.assertEqual(root.children, [c1, c3])
+        root.clear_widgets([])
+        self.assertEqual(root.children, [c1, c3])
+        root.clear_widgets()
+        self.assertEqual(root.children, [])
+
```
matham, line 49:
```suggestion

    def test_clear_widgets_children(self):
        root = self.root
        for _ in range(10):
            root.add_widget(self.cls())
        self.assertEqual(len(root.children), 10)

        root.clear_widgets(root.children)
        self.assertEqual(root.children, [])
```

To make sure passing in `children` works.

matham (changes requested):

matham (approved):

kivy/uix/widget.py
```diff
@@ -684,11 +684,11 @@ def clear_widgets(self, children=None):
             The `children` argument can be used to specify the children you
             want to remove.
         '''
```
matham, line 686:
```suggestion
        .. versionchanged:: 2.1.0

            Specifying an empty ``children`` list leaves the widgets unchanged.
            Previously it was treated like ``None`` and all children were
            removed.
        '''
```

matham (changes requested):
Forgot to document changes.

matham (approved):

Issue #6108, opened 2019-01-21
Kivy does not watch keys only appearing in f-strings


brady-ds:
### Versions

* Python: 3.6.5
* OS: OpenSUSE 15.0
* Kivy: 1.10.1
* Kivy installation method: `pip3`

### Description

Kivy does not watch keys that only appear in `f`-strings because `ParserRuleProperty.precompile` is written to ignore strings, as seen [here](https://github.com/kivy/kivy/blob/master/kivy/lang/parser.py#L168) and [here](https://github.com/kivy/kivy/blob/master/kivy/lang/parser.py#L191).  For example, in the attached code both buttons should behave identically, disabling themselves and showing "True" when clicked.  But the second button never updates its text because it isn't watching `self.diabled`.

### Code and Logs

`main.py`:
```python
from kivy.app import App


class TestApp(App):
    pass


if __name__ == '__main__':
    TestApp().run()
```

`test.kv`:
```kv
BoxLayout:
    Button:
        text: str(self.disabled)
        on_press: self.disabled = True
    Button:
        text: f'{self.disabled}'
        on_press: self.disabled = True
```

adamwolf:
Is there a workaround that still uses f-strings?

PR #7371, opened 2021-02-07
KV: Allow using f-strings in KV-lang

tshirtman:
…and binding to properties in them

This approach uses the ast module to parse the content of
fstrings and getting any name in them, to add that to the list of
keys to watch.

fixes: #6108

Maintainer merge checklist

### DIFF
--- a/kivy/lang/__init__.py
+++ b/kivy/lang/__init__.py
@@ -176,6 +176,12 @@ the value can use the values of other properties using reserved keywords.
             TextInput:
                 on_focus: self.insert_text("Focus" if args[1] else "No focus")
 
+
+.. versionchanged:: 2.1.0
+
+    f-strings are now parsed in value expressions, allowing to bind to the
+    properties that they contain.
+
 .. kv-lang-ids:
 
 ids
--- a/kivy/lang/parser.py
+++ b/kivy/lang/parser.py
@@ -9,6 +9,7 @@ import os
 import re
 import sys
 import traceback
+import ast
 from re import sub, findall
 from types import CodeType
 from functools import partial
@@ -35,11 +36,16 @@ Cache.register('kv.lang')
 __KV_INCLUDES__ = []
 
 # precompile regexp expression
-lang_str = re.compile(
-    "((?:'''.*?''')|"
+str_re = (
+    "(?:'''.*?''')|"
     "(?:(?:(?<!')|''')'(?:[^']|\\\\')+?'(?:(?!')|'''))|"
     '(?:""".*?""")|'
-    '(?:(?:(?<!")|""")"(?:[^"]|\\\\")+?"(?:(?!")|""")))', re.DOTALL)
+    '(?:(?:(?<!")|""")"(?:[^"]|\\\\")+?"(?:(?!")|"""))'
+)
+
+lang_str = re.compile(f"({str_re})", re.DOTALL)
+lang_fstr = re.compile(f"([fF](?:{str_re}))", re.DOTALL)
+
 lang_key = re.compile('([a-zA-Z_]+)')
 lang_keyvalue = re.compile(r'([a-zA-Z_][a-zA-Z0-9_.]*\.[a-zA-Z0-9_.]+)')
 lang_tr = re.compile(r'(_\()')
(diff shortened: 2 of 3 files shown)


matham:
This is great! However, it will need comprehensive tests to be sure it doesn't break other parts that are not f-strings (e.g. `self.x + f'{self.y}'`) and that the f-string part works like the rest of kv.

tshirtman:
Yes, i'll see what i can do for the testing, but i didn't expect to get away with it ;).
Normally this can only add, as these strings were previously ignored, and we only go over them, so unless our regex are wrong and can, in some situation, match more than the string itself, i don't think it could break anything that currently works, what's harder to make sure of is is that all the kind of expressions currently supported are also matched inside f-strings.
But yes, testing is the answer :).

tshirtman:
I think the doc build failure is unrelated, and i’m not sure if/where to add versionchanged/versionadded, so i think it’s ready to merge. Suggestion for more cases to test are welcome though, i’m not sure the existing ones cover all the code branches (although i tested disabling a couple ones and it did break some of the tests, so most of the code should be useful).

matham:
Can you please test that `text: f"{' '.join([original.text for _ in range(2)])}"` does not bind (because it doesn't in normal kv I believe. And the same for `text: f"{original.text.upper()}"` which I believe also doesn't bind in normal kv. Additionally, that `text: f"{original.text}" + other.text` works. And finally that nested attributes work (i.e. `text: f"{' '.join(p.upper() for p in other.original.text)}"`), including `rebind` if possible).

>I think the doc build failure is unrelated

Yep, it has been happening for a bit. Not sure why.

>and i’m not sure if/where to add versionchanged/versionadded

At the end here: https://github.com/kivy/kivy/blob/master/kivy/lang/__init__.py is probably the best place.

tshirtman:
I added the suggested "shouldn't bind" examples, and i was surprised by the result, do you think that's what we should expect?
```py
    def test_bind_fstring_expressions_should_not_bind(self):
         from kivy.lang import Builder              
         root = Builder.load_string(dedent('''
         FloatLayout:
             Label:
                 id: original
                 text: 'perfect'
             Label:
                 id: target1
                 text: f"{' '.join([original.text for _ in range(2)])}"
             Label:
                 id: target2
                 text: f"{original.text.upper()}"
         '''))
         assert root.ids.target1.text == ' '
         assert root.ids.target2.text == ''
         root.ids.original.text = 'new text'
         assert root.ids.target1.text == ' '
         assert root.ids.target2.text == '
```

matham:
Well, more accurately I think `f"{original.text.upper()}"` *should* bind. But I don't think `original.text.upper()` binds, so the f-string version also shouldn't bind until it's fixed for the regular version.

Similarly, `f"{' '.join([original.text for _ in range(2)])}"` really should bind to `original.text`, however, what I want to exclude is `f"{sum(obj.x for obj in self.objs)}"`, because it shouldn't bind to the last `obj`. And I don't know that we can do the former without the latter. Plus I don't think regular kv binds to anything in comprehension so it also shouldn't work for f-strings.

Eventually we should use ast for everything. I had done that for the kv-compiler (https://github.com/matham/kivy/blob/ast_compile_kv_lang/kivy/lang/compiler/ast_parse.py#L1151), and kv does a lot of things wrong, or at least partially. But that's why we needed a real spec and a proper discussion about the exact syntax to support. In that branch I had optionally supported comprehensions and other things. But for now, we need to stick to what kivy already supports I think.

matham:
But to answer the question directly

>do you think that's what we should expect?

yes, for now, to be consistent with the rest of kv.

tshirtman:
Ok, i'm unsure about making that a "specification" (and making a test for that does feel like that to me) because it feels more like a bug, or at least a current limitation, that would be nice to break out of at some point, and i think you agree. 

I added the other case, and indeed it doesn't bind.

Yes, about ast, making it for that part also made me think the whole expression should be parsed with it, not just f-strings, and indeed i remember your branch, which i wasn't able to digest at the time, hopefully i would have a better understanding of it now, or maybe at least we can go in that direction more gradually (the size of that change really didn't help, imho).

matham:
>Ok, i'm unsure about making that a "specification" (and making a test for that does feel like that to me) because it feels more like a bug, or at least a current limitation, that would be nice to break out of at some point, and i think you agree.

I'm assuming you're talking about `original.text.upper()`. And what you say makes sense. But you did add a test for it?

matham (approved):

tshirtman:
> > Ok, i'm unsure about making that a "specification" (and making a test for that does feel like that to me) because it feels more like a bug, or at least a current limitation, that would be nice to break out of at some point, and i think you agree.
> 
> I'm assuming you're talking about `original.text.upper()`. And what you say makes sense. But you did add a test for it?

yes, but for some reason the result is empty for these two cases you suggested, which did surprise me, i should have been clearer earlier.

https://github.com/kivy/kivy/pull/7371/files#diff-efce534e4947a1105d10041d9cf02a608630c875d8b5cb0ff380c63e751c467fR389

matham:
Ok, I guess when we fix it we can change the tests.

LGTM to merge.

matham (changes requested):
Oh, it's still missing the versionadded.

tshirtman:
Added it at the end of the ValueExpressions section.

matham:
You got a trailing white-space somewhere in there.

matham (approved):

PR #7372, opened 2021-02-07
Focus: Allow modifiers (e.g. numlock) be present to tab cycle focus

### LINKED ISSUE #7364: Allow `get_focus_previous` with `numlock` enabled
**Is your feature request related to a problem? Please describe.**
When `numlock` is enabled, `shift+tab` doesn't call `get_focus_previous` for widgets inheriting `FocusBehavior`

Change this:
https://github.com/kivy/kivy/blob/ed68be56e4d580b00abd8f89705339e824c3b43d/kivy/uix/behaviors/focus.py#L534-L535
to:
```python
if 'shift' in modifiers:
    next = self.get_focus_previous()
```
or if `ctrl+shift+tab` is undesired:
```python
if 'shift' in modifiers and 'ctrl' not in modifiers:
    next = self.get_focus_previous()
```

If the devs approve, I'll submit a pull request

xayhewalo:
fixes #7364 

Maintainer merge checklist

### DIFF
--- a/kivy/uix/behaviors/focus.py
+++ b/kivy/uix/behaviors/focus.py
@@ -531,7 +531,7 @@ class FocusBehavior(object):
         key was consumed.
         '''
         if keycode[1] == 'tab':  # deal with cycle
-            if ['shift'] == modifiers:
+            if 'shift' in modifiers:
                 next = self.get_focus_previous()
             else:
                 next = self.get_focus_next()
--- a/kivy/uix/behaviors/focus.py
+++ b/kivy/uix/behaviors/focus.py
@@ -532,8 +532,8 @@ class FocusBehavior(object):
         '''
         if keycode[1] == 'tab':  # deal with cycle
             if {'ctrl', 'alt', 'meta', 'super', 'compose'} & modifiers:
-                return false
-            if {'shift'} & modifiers:  # or "in"
+                return False
+            if 'shift' in modifiers:
                 next = self.get_focus_previous()
             else:
                 next = self.get_focus_next()
--- a/kivy/uix/behaviors/focus.py
+++ b/kivy/uix/behaviors/focus.py
@@ -531,7 +531,9 @@ class FocusBehavior(object):
         key was consumed.
         '''
         if keycode[1] == 'tab':  # deal with cycle
-            if 'shift' in modifiers:
+            if {'ctrl', 'alt', 'meta', 'super', 'compose'} & modifiers:
+                return false
+            if {'shift'} & modifiers:  # or "in"
                 next = self.get_focus_previous()
             else:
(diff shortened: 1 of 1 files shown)


kivy/uix/behaviors/focus.py
```diff
@@ -531,7 +531,7 @@ def keyboard_on_key_down(self, window, keycode, text, modifiers):
         key was consumed.
         '''
         if keycode[1] == 'tab':  # deal with cycle
-            if ['shift'] == modifiers:
+            if 'shift' in modifiers:
```
matham, line 534:
```suggestion
            if {'alt', 'rctrl', 'lctrl', 'meta', 'super', 'alt-gr',
                    'compose'} & modifiers:
                return False

            if {'shift', 'rshift'} & modifiers:
```

E.g. on Windows, if the ctrl or these other special keys are pressed, tab means something else, like switch window (Windows will not even transmit to Kivy these keys when pressed since it uses it). So we shouldn't use them for tabbing, either.

matham (changes requested):

kivy/uix/behaviors/focus.py
xayhewalo: `rctrl`, `lctrl`, and `alt-gr` don't show up in `modifiers` on the platforms I have access too, only in `keycode`. I think what we want is
```suggestion
            if {'ctrl', 'alt', 'meta', 'super', 'compose'} & modifiers:
                return false
            if {'shift'} & modifiers:  # or "in"
```

matham: I was going by the list from https://github.com/kivy/kivy/blob/master/kivy/core/window/__init__.py#L63. But I guess you're right so feel free to use your suggestion.

kivy/uix/behaviors/focus.py
```diff
@@ -531,7 +531,9 @@ def keyboard_on_key_down(self, window, keycode, text, modifiers):
         key was consumed.
         '''
         if keycode[1] == 'tab':  # deal with cycle
-            if ['shift'] == modifiers:
+            if {'ctrl', 'alt', 'meta', 'super', 'compose'} & modifiers:
+                return false
+            if {'shift'} & modifiers:  # or "in"
```
matham, line 536:
```suggestion
                return False
            if 'shift' in modifiers:
```

It's probably faster to use `in`. And False was not capitalized.

matham (changes requested):

xayhewalo:
I'm not sure why CI tests are failing

matham (approved):

matham:
The docs failed because of https://github.com/sphinx-doc/sphinx/issues/8880.

PR #7383, opened 2021-02-17
fix Dropdown reposition in scrollview/recycleview bug.

### LINKED ISSUE #7382: Dropdown opening from data item in RecycleView is jumping down at refresh of RV-data/-layout
**Software Versions**
* Python: 3.6.9
* OS: Ubuntu 18.04
* Kivy: 2.0.0

A `Dropdown` attached to a `data` item of a `RecycleView` (RV) is jumping down and up if `data` of the `RecycleView` gets updated/refreshed.

**Expected behavior**
The `Dropdown` `y` position should not be recalculated if the `parent` of the item widget where the `Dropdown` is attached is `None`.

**To Reproduce**
Create a `RecycleView` with `Button` items, which are displaying a `Dropdown` when the user clicks on the `Button` item. Create a `Clock` timer to periodically update the `data` property of the RV. Now click on an item to open the `Dropdown`. Now you can see that the `Dropdown` is not staying fixed under the attaching item/button-widget - instead, it is jumping down and up again every time the `data` gets updated.

RV_Dropdown_repos_bug_gimp_opti

As you can see in the animation above, the `y` position of the item gets wrongly calculated if `parent` is `None`.

For to workaround overwrite the `_reposition` method of the `Dropdown` widget to prevent the recalculation of the `y` position if the `parent` is None:

```python
    def _reposition(self, *largs):
        if self.attach_to and not self.attach_to.parent:
            return
        super()._reposition(*largs)
```

**Additional context**
On refresh of RV's `data` property, the items will temporarily be removed from the RV, which sets the `parent` property of the item to `None`. This breaks the calculation of the item screen position (not adding the y-offset of the `ScrollView`) and the `Dropdown` is jumping down. As soon as the `parent` of the item is getting reset, the screen position gets calculated again and the `Dropdown` is jumping back up (to its original/correct position, underneath the attaching item).

I will create PR to fix this within `Dropdown._reposition`.

AndiEcker:
Fixes #7382

Maintainer merge checklist


tshirtman (approved):
Looks good to me.

AndiEcker:
thank you very much @tshirtman for your review and approval!!!

The change is small and logical and allows me to remove the workaround from my libraries... only the RTD/Sphinx check is failing - but I assume this has nothing to do with this PR as I've seen this recently on other PRs

tshirtman:
Yes, the doc error happens on other PRs, so no worries.

I didn' test the code yet, but i don't see how this could have any bad effect*, so i think we could merge it anyway, and fix later if something happens.

Do you think it's possible to make unit test showing what this bug fixes? It's probably more work than the change itself, but it would be nice to have.

* famous last words!

AndiEcker:
It is on my wishlist to dive into Kivy unit testing and add some unit tests
to the Kivy framework, but I'm currently very busy - so this will not
happen soon... sorry

Am So., 21. Feb. 2021 um 20:48 Uhr schrieb Gabriel Pettier <
notifications@github.com>:

tshirtman (approved):

matham:
If I understood correctly, this checks whether there is no parent for the widget, meaning the widget is not part of the tree. However, while this shouldn't cause harm, I don't think it solves the issue fully!? What if the widget does have a parent, but the parent doesn't have a parent?

Perhaps a better check would be whether the widget's ultimate parent is Window (`win`)?

AndiEcker:
@matham: good point/finding.

Would it be ok for you to make a PR, changing the check (extended by the
last fix) to something like::

```
--- if not widget or not widget.parent or not win:
+++ if not widget or not widget.get_parent_window() or not win:
```
???


Am So., 14. März 2021 um 17:51 Uhr schrieb matham ***@***.***

matham:
>changing the check (extended by the last fix) to something like

Yeah, that seems like a good fix.

AndiEcker:
Thanks @matham!!!

The PR is on the way .... will push it in the next days.

Am So., 14. März 2021 um 20:36 Uhr schrieb matham ***@***.***

PR #7387, opened 2021-02-24
Mouse: Update MouseMotionEventProvider to dispatch hover events

pythonic64:
Proposed changes will make `MouseMotionEventProvider` dispatch hover `MouseMotionEvent` event by reacting on properties/events:
- `mouse_pos`
- `system_size`
- `on_cursor_enter`
- `on_cursor_leave`
- `on_close`
- `on_rotate`

Changes:
- `MotionEvent.__init__` method now accept `is_touch` as argument and that change was needed to make `depack` work with touch and hover (non-touch) events
- Method `grab` in `MotionEvent` is changed to enable grab of non-touch events.
- Attribute `button` is added to `MotionEvent` because it was missing from `__init__` and is used by `is_mouse_scrolling` property.
- If window provider dispatches `on_cursor_xxx` events, then `on_cursor_enter` will create `begin` hover event and `on_cursor_leave` will create `end` hover event. Otherwise, the first change of `mouse_pos` will create `begin` event and `end` will be dispatched on `on_close` event. In both cases change to `mouse_pos` will dispatch `update` events.
- Updated `MouseMotionEventProvider` to use `fbind/funbind` methods.

Maintainer merge checklist:

### DIFF
--- a/kivy/tests/test_mouse_hover_event.py
+++ b/kivy/tests/test_mouse_hover_event.py
@@ -52,13 +52,7 @@ class MouseHoverEventTestCase(GraphicUnitTest):
         self.etype = etype
         self.motion_event = event
 
-    def on_touch_down(self, _, touch):
-        self.touch_event = touch
-
-    def on_touch_move(self, _, touch):
-        self.touch_event = touch
-
-    def on_touch_up(self, _, touch):
+    def on_any_touch_event(self, _, touch):
         self.touch_event = touch
 
     def to_relative_pos(self, win, x, y):
@@ -79,9 +73,9 @@ class MouseHoverEventTestCase(GraphicUnitTest):
         win = EventLoop.window
         if with_window_children:
             from kivy.uix.button import Button
-            button = Button(on_touch_down=self.on_touch_down,
-                            on_touch_move=self.on_touch_move,
-                            on_touch_up=self.on_touch_up)
+            button = Button(on_touch_down=self.on_any_touch_event,
+                            on_touch_move=self.on_any_touch_event,
+                            on_touch_up=self.on_any_touch_event)
             self.button_widget = button
             win.add_widget(button)
         return win, self.mouse
--- a/kivy/tests/test_mouse_hover_event.py
+++ b/kivy/tests/test_mouse_hover_event.py
@@ -6,11 +6,15 @@ class MouseHoverEventTestCase(GraphicUnitTest):
     '''Tests hover event from `MouseMotionEventProvider`.
     '''
 
+    framecount = 3
+    '''Must be equal of max number of `self.advance_frame` in test method.'''
+
     def setUp(self):
         super().setUp()
         self.etype = None
-        self.hover_event = None
+        self.motion_event = None
         self.touch_event = None
+        self.button_widget = None
         self.mouse = mouse = MouseMotionEventProvider('mouse', '')
         from kivy.base import EventLoop
         win = EventLoop.window
@@ -19,24 +23,34 @@ class MouseHoverEventTestCase(GraphicUnitTest):
         win.system_size = (320, 240)
         mouse.start()
         EventLoop.add_input_provider(mouse)
+        win.fbind('on_motion', self.on_motion)
+        # Patch `win.on_close` method to prevent EventLoop from removing
+        # window from event listeners list.
+        self.old_on_close = win.on_close
+        win.on_close = lambda *args: None
 
     def tearDown(self, fake=False):
         super().tearDown(fake)
         self.etype = None
-        self.hover_event = None
+        self.motion_event = None
         self.touch_event = None
         from kivy.base import EventLoop
         win = EventLoop.window
-        for widget in win.children[:]:
-            win.remove_widget(widget)
+        if self.button_widget:
+            win.remove_widget(self.button_widget)
+            self.button_widget = None
         mouse = self.mouse
         mouse.stop()
-        EventLoop.remove_event_listener(mouse)
+        EventLoop.remove_input_provider(mouse)
         self.mouse = None
+        win.funbind('on_motion', self.on_motion)
+        # Restore method `on_close` to window
+        win.on_close = self.old_on_close
+        self.old_on_close = None
 
-    def dispatch_fn(self, etype, event):
+    def on_motion(self, _, etype, event):
         self.etype = etype
-        self.hover_event = event
+        self.motion_event = event
 
     def on_touch_down(self, _, touch):
         self.touch_event = touch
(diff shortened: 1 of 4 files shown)


matham (dismissed):
I like the logic. We just create a hover event (basically a mouse event that is not a touch) in addition to normal mouse touch events. But currently they'll just be ignored and dropped in Window `on_motion` which only processes `is_touch` events.

Then we leave the actual processing and dispatching of this event to the window tree to managers, right? I had initially assumed that the manager may be able to get this info from the existing motion events and perhaps window properties, that's why I was initially puzzled why we have to add a hover type event (or more accurately a non-touch mouse event). But this makes sense now.

I didn't double check the exact details of starting/stopping/updating the events are correct, but I'll trust your implementation. Although, having tests would be nice, including checking that it currently doesn't effect normal widgets.

Perhaps this also needs docs, but I'm not sure where.

matham (dismissed):
Looks ok, but we should add a test eventually (perhaps in a subsequent PR) that the new non-touch mouse events are not dispatched to any widgets by default.

kivy/tests/test_mouse_hover_event.py
```diff
@@ -0,0 +1,224 @@
+from kivy.input.providers.mouse import MouseMotionEventProvider
```
matham, line 1: Actually, it's probably better to not import these globally but rather where you need it; that could be responsible for the failed tests.

matham (changes requested):

kivy/tests/test_mouse_hover_event.py
pythonic64: Failing rtd is because of `compoundselection.py` and that should be fixed by https://github.com/kivy/kivy/pull/7441 which is merged in master. 

The import of `MouseMotionEventProvider` is in global scope in `test_mouse_multitouchsim.py` too, but I'll move it in local scope for both modules in the next commit.

matham:
Ok, but the windows tests are failing!?

pythonic64:
Yes, for Python3.9.
```
__________________ AsyncImageTestCase.test_reload_asyncimage __________________

self = <kivy.tests.test_uix_asyncimage.AsyncImageTestCase testMethod=test_reload_asyncimage>

    def test_reload_asyncimage(self):
        from tempfile import mkdtemp
        from os import remove
        from shutil import copyfile, rmtree
    
        fn = resource_find('data/logo/kivy-icon-16.png')
        t = mkdtemp()
        source = join(t, 'source.png')
        copyfile(fn, source)
        image = AsyncImage(source=source)
        self.render(image, framecount=2)
>       self.assertEqual(image.texture_size, [16, 16])
E       AssertionError: [32, 32] != [16, 16]

kivy\tests\test_uix_asyncimage.py:125: AssertionError
```
They pass with Python3.7 on my Windows10.

matham (approved):

pythonic64:
Well, I expected the tests to fail on the CI with the latest commit as they failed on the previous one, but they passed. Seems to me that there are side-effects in some tests and it's probably due to some cleanup missing. Like `Window.children` isn't always empty between test classes or Cache doesn't get cleared.

Here is the traceback from test which failed, but with latest commit it passed although there weren't any changes that would fix this failure:
```
___ MultitouchSimulatorTestCase.test_multitouch_on_demand_scatter_rightmove ___

self = <kivy.tests.test_mouse_multitouchsim.MultitouchSimulatorTestCase testMethod=test_multitouch_on_demand_scatter_rightmove>

    def test_multitouch_on_demand_scatter_rightmove(self):
        self.multitouch_dot_move(
>           'right', on_demand=True, scatter=True
        )

kivy\tests\test_mouse_multitouchsim.py:541: 

self = <kivy.tests.test_mouse_multitouchsim.MultitouchSimulatorTestCase testMethod=test_multitouch_on_demand_scatter_rightmove>
button = 'right', kwargs = {'on_demand': True, 'scatter': True}
eventloop = <kivy.base.EventLoopBase object at 0x00000212EDB35128>
win = <kivy.core.window.window_sdl2.WindowSDL object at 0x00000212EE2ABCF8>
mouse = <kivy.input.providers.mouse.MouseMotionEventProvider object at 0x0000021283EA7240>
wid = <kivy.uix.scatter.Scatter object at 0x00000212867D6908>
event_id = 'unittest1'

    def multitouch_dot_move(self, button, **kwargs):
        # touch -> dot appears -> move touch -> dot moves
        # -> release touch -> touch & dot disappear
        eventloop, win, mouse, wid = self.mouse_init(**kwargs)
    
        # register mouse provider
        mouse.start()
        eventloop.add_input_provider(mouse)
    
        # no mouse touch anywhere
        self.assertEqual(mouse.counter, 0)
        self.assertEqual(mouse.touches, {})
    
        # right button down, red dot should appear
        # if the 'multitouch_on_demand' is disabled
        win.dispatch(
            'on_mouse_down',
            10, self.correct_y(win, 10),
            'right', {}
        )
        event_id = next(iter(mouse.touches))
        self.assertEqual(mouse.counter, 1)
    
        if 'on_demand' in kwargs and 'scatter' not in kwargs:
            # doesn't do anything on a pure Button
            self.render(wid)
    
            # cleanup!
            # remove mouse provider
            mouse.stop()
            eventloop.remove_input_provider(mouse)
            return
    
        # XXX right button up
        # first release the touch then check, so that we
        # have the red dot drawn in on_demand and in the
        # default (multitouch everywhere) because in the
        # multitouch_on_demand is the circle drawn after
        # the touch is released (in on_mouse_release)
        elif 'on_demand' in kwargs and 'scatter' in kwargs:
            # on_demand works after the touch is up
            self.assertIn(
                'multitouch_sim',
                mouse.touches[event_id].profile
            )
            self.assertTrue(mouse.multitouch_on_demand)
    
            # multitouch_sim is changed in on_touch_down
            # method of the widget that's able to handle
            # multiple touches, therefore for Scatter we
            # need to dispatch the method and because we
            # triggered only on_mouse_down directly i.e.
            # without ME dispatch, on_touch_down was not
            # called == multitouch_sim is False
            self.advance_frames(1)  # initialize stuff
            wid.on_touch_down(mouse.touches[event_id])
            wid.on_touch_up(mouse.touches[event_id])
            self.assertTrue(mouse.touches[event_id].multitouch_sim)
    
            win.dispatch(
                'on_mouse_up',
                10, self.correct_y(win, 10),
                'right', {}
            )
            color = mouse.touches[
                event_id
>           ].ud.get('_drawelement')[0].proxy_ref
E           TypeError: 'NoneType' object is not subscriptable

kivy\tests\test_mouse_multitouchsim.py:263: TypeError
```

matham:
These kinds of things are often due to timing issues. In my own app, I've noticed that the Windows CI is much slower than Linux CI. So I've noticed that often (some of the) Windows CI may fail, but not linux. So yeah, probably the tests assumes something about time, which is not obvious and fails when things are slower.

But it should be fine to merge, if it keeps on happening on master, we can troubleshoot then.

pythonic64:
Ok, merging. We should note that tests on CI can report a false positive and it that case developer submitting the pull request should run them on their machine to verify the failing cases.

matham:
Looks like that test now consistently fails on the Windows CI for all the new PRs. So we should either disable the test on Windows (CI), or figure out where it goes wrong.

From a quick look I didn't see any timing code there so I'm not sure why it happens.

Zen-CODE:
@pythonic64 Any feedback on this? To 2nd @matham , we need to revert or fix?

pythonic64:
I'm out of ideas to fix the failing test https://github.com/kivy/kivy/pull/7387#issuecomment-812091417 or reproduce it on my machine.
@matham I can submit a request to skip `MouseHoverEventTestCase` class, at least for now, until someone suggests a solution.

matham:
Yes, a PR to disable the test for now on Windows CI specifically, so we can see if it fails on elsewhere, would be good. I wonder if it fails due to change in tests, or due to the change in the provider code.

Zen-CODE:
Interestingly, it passed on other windows test, so I'd guess it's inconsistent provider behavior? Let me see if I can reproduce it in a VM....

PR #7390, opened 2021-02-25
Fix SoundAndroidPlayer state persisting after playback completion

allerter:
This PR adds an `OnCompletionListener` that will call its callback once the playback is finished. Before this, once the playback had finished, the `state` would still stay on `play` and `on_stop` wouldn't be dispatched since there was no call to `Sound.stop`. This takes care of that.

Maintainer merge checklist


tshirtman (approved):
Didn't test, but looks good to me :)

PR #7418, opened 2021-03-11
Fixes video thread that sleep before dispatching eos.

### LINKED ISSUE #7417: Video lag between loop when eos=loop is used
If you instanciate a Video using eos to loop, you can see it lagging every restart, using ffpyplayer video core.

```python
video = Video(
    state="play",
    options={"eos": "loop"},
    source=filename)
```

This is due to the sleep(0.2) here: https://github.com/kivy/kivy/blob/master/kivy/core/video/video_ffpyplayer.py#L303

tito:
Sleeping here was to avoid a busyloop when waiting the user/code to do something when eos happened.
I am not trying to refactor this code, but this Event allow a instant reactivity from the mainthread to the frame's thread, and reduce the lag, visually. 

Fixes #7417

### DIFF
--- a/kivy/core/video/video_ffpyplayer.py
+++ b/kivy/core/video/video_ffpyplayer.py
@@ -53,7 +53,8 @@ except:
     raise
 
 
-from threading import Thread, Event
+from threading import Thread
+from queue import Queue, Empty
 from kivy.clock import Clock, mainthread
 from kivy.logger import Logger
 from kivy.core.video import VideoBase
@@ -109,13 +110,25 @@ class VideoFFPy(VideoBase):
         self._next_frame = None
         self._seek_queue = []
         self._ffplayer_need_quit = False
-        self._ev_trigger = Event()
+        self._wakeup_queue = Queue(maxsize=1)
         self._trigger = Clock.create_trigger(self._redraw)
 
         super(VideoFFPy, self).__init__(**kwargs)
 
     def __del__(self):
         self.unload()
+    
+    def _wakeup_thread(self):
+        try:
+            self._wakeup_queue.put(None, False)
+        except Empty:
+            pass
+    
+    def _wait_for_wakeup(self, timeout):
+        try:
+            self._wakeup_queue.get(True, timeout)
+        except Empty:
+            pass
 
     def _player_callback(self, selector, value):
         if self._ffplayer is None:
@@ -135,7 +148,7 @@ class VideoFFPy(VideoBase):
 
     def _set_volume(self, volume):
         self._volume = volume
-        if self._ffplayer:
+        if self._ffplayer is not None:
             self._ffplayer.set_volume(self._volume)
 
     def _get_duration(self):
@@ -150,14 +163,21 @@ class VideoFFPy(VideoBase):
         elif self.eos == 'stop':
             self.stop()
         elif self.eos == 'loop':
+            # this causes a seek to zero
             self.position = 0
 
         self.dispatch('on_eos')
 
     @mainthread
-    def _change_state(self, state):
-        self._state = state
-        self._ev_trigger.set()
+    def _finish_setup(self):
+        # once setup is done, we make sure player state matches what user
+        # could have requested while player was being setup and it was in limbo
+        # also, thread starts player in internal paused mode, so this unpauses
+        # it if user didn't request pause meanwhile
+        if self._ffplayer is not None:
+            self._ffplayer.set_volume(self._volume)
+            self._ffplayer.set_pause(self._state == 'paused')
+            self._wakeup_thread()
 
     def _redraw(self, *args):
         if not self._ffplayer:
@@ -213,48 +233,48 @@ class VideoFFPy(VideoBase):
 
             self.dispatch('on_frame')
 
-    def _next_frame_run(self):
-        ffplayer = self._ffplayer
+    def _next_frame_run(self, ffplayer):
         sleep = time.sleep
         trigger = self._trigger
         did_dispatch_eof = False
+        wait_for_wakeup = self._wait_for_wakeup
         seek_queue = self._seek_queue
+        # video starts in internal paused state
 
         # fast path, if the source video is yuv420p, we'll use a glsl shader
         # for buffer conversion to rgba
+        # wait until we get frame metadata
         while not self._ffplayer_need_quit:
             src_pix_fmt = ffplayer.get_metadata().get('src_pix_fmt')
             if not src_pix_fmt:
-                sleep(0.005)
+                wait_for_wakeup(0.005)
                 continue
 
             if src_pix_fmt == 'yuv420p':
                 self._out_fmt = 'yuv420p'
                 ffplayer.set_output_pix_fmt(self._out_fmt)
-            self._ffplayer.toggle_pause()
             break
 
         if self._ffplayer_need_quit:
+            ffplayer.close_player()
             return
 
         # wait until loaded or failed, shouldn't take long, but just to make
         # sure metadata is available.
-        s = time.perf_counter()
         while not self._ffplayer_need_quit:
             if ffplayer.get_metadata()['src_vid_size'] != (0, 0):
                 break
-            # XXX if will fail later then?
-            if time.perf_counter() - s > 10.:
-                break
-            sleep(0.005)
+            wait_for_wakeup(0.005)
 
         if self._ffplayer_need_quit:
+            ffplayer.close_player()
             return
 
-        # we got all the informations, now, get the frames :)
-        self._change_state('playing')
+        self._ffplayer = ffplayer
+        self._finish_setup()
+        # now, we'll be in internal paused state and loop will wait until
+        # mainthread unpauses us when finishing setup
 
-        ev_trigger = self._ev_trigger
         while not self._ffplayer_need_quit:
             seek_happened = False
             if seek_queue:
@@ -297,6 +317,8 @@ class VideoFFPy(VideoBase):
                     frame, val = ffplayer.get_frame(force_refresh=True)
                 finally:
                     ffplayer.set_pause(bool(self._state == 'paused'))
+                    # todo: this is not safe because user could have updated
+                    # volume between us reading it and setting it
                     ffplayer.set_volume(self._volume)
             # Get next frame regular:
             else:
@@ -304,13 +326,12 @@ class VideoFFPy(VideoBase):
 
             if val == 'eof':
                 if not did_dispatch_eof:
-                    ev_trigger.clear()
                     self._do_eos()
                     did_dispatch_eof = True
-                self._ev_trigger.wait(.2)
+                wait_for_wakeup(None)
             elif val == 'paused':
                 did_dispatch_eof = False
-                sleep(0.2)
+                wait_for_wakeup(None)
             else:
                 did_dispatch_eof = False
                 if frame:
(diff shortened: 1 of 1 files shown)


tshirtman (approved):
LGTM

matham (commented):
Strictly speaking we could still get into a situation where it waits .2s (e.g. the user clicks play -> event is set in main thread -> event is cleared in second thread -> we wait the .2 timeout, this can happen in each loop iteration). However, it is still strictly better than before, so it should be fine.

But, I think `ev_trigger.clear()` can be moved under the `if not did_dispatch_eof:` section, which should happen only once when we hit `eof`.

tito:
> Strictly speaking we could still get into a situation where it waits .2s (e.g. the user clicks play -> event is set in main thread -> event is cleared in second thread -> we wait the .2 timeout, this can happen in each loop iteration). However, it is still strictly better than before, so it should be fine.

That's the goal, not being a busy loop. You could honestly push to to 60s for example, it triggers only when there is a change that requires an action from the second thread.
I'm just to scared to push it, and as i said, didn't want to go refactor, just reduce the lag.

kivy/core/video/video_ffpyplayer.py
```diff
@@ -252,7 +254,10 @@ def _next_frame_run(self):
         # we got all the informations, now, get the frames :)
         self._change_state('playing')
 
+        ev_trigger = self._ev_trigger
         while not self._ffplayer_need_quit:
+            ev_trigger.clear()
+
```
matham, line 260:
```suggestion
```

kivy/core/video/video_ffpyplayer.py
```diff
@@ -300,10 +305,10 @@ def _next_frame_run(self):
                 frame, val = ffplayer.get_frame()
 
             if val == 'eof':
-                sleep(0.2)
                 if not did_dispatch_eof:
                     self._do_eos()
```
matham, line 309:
```suggestion
                    ev_trigger.clear()
                    self._do_eos()
```

matham (changes requested):
If the goal is to reduce lag, then this change will eliminate the lag when the situation I described above occurs. It is safe to move it under there because it's only cleared the first time *before* the other thread is made aware of eof so before the user has a chance to react. So the first time the user reacts it'll still set it and then there will no lag.

With this change, the only possibility of lag is if the user hits play/pause just as we hit eof and before we handle it.

But, if you're uncomfortable with the change, you can leave it as is and the PR is fine.

kivy/core/video/video_ffpyplayer.py
```diff
@@ -256,8 +256,6 @@ def _next_frame_run(self):
 
         ev_trigger = self._ev_trigger
         while not self._ffplayer_need_quit:
-            ev_trigger.clear()
```
rnixx: this is actually the correct place for clearing the event if the remaining sleep calls are replaced. but it should be checked against is_set(), because clear() always acquires a lock.

kivy/core/video/video_ffpyplayer.py
```diff
@@ -306,6 +304,7 @@ def _next_frame_run(self):
 
             if val == 'eof':
                 if not did_dispatch_eof:
+                    ev_trigger.clear()
```
rnixx, line 307: see above

kivy/core/video/video_ffpyplayer.py
```diff
@@ -109,6 +109,7 @@ def __init__(self, **kwargs):
         self._next_frame = None
         self._seek_queue = []
         self._ffplayer_need_quit = False
+        self._ev_trigger = Event()
```
rnixx, line 112: I think _ev_trigger is not a very good name for the purpose of the event

rnixx (commented):
All sleep() calls should be replaced with event.wait() in _next_frame_run(). i did and tested this in the referenced PR.

the quit flag and the event might be cleared before the frame fetching thread gets started for better readability.

the thread should not be a daemon, better would be to add an event handler which a least at application exit should call unload(), in the remaining cases the application should gracefully call unload to ensure resources are freed properly.

since the load() function simply calls unload(), the play function might call unload directly for better readability, and the load function can me matched directly on class by setting load = unload.

the unload function might call _ffplayer.close_player() explicitly for better readability.

matham:
I switched to using a queue because with an event, we have to clear it, but that would lead to having to wait if the user e.g. stopped just after we cleared it. We could check the event before clearing it in the thread and only clear it if the event is set, but then we could potentially clear the event again *after* the user set it so now we'd have to wait longer again until it times out, The queue makes the one way directionality (kivy adds to the queue, thread only removes from the queue) clearer so we don't have to deal with having to clear and the issues that come with it because with the current implementation we can be sure that it never has to wait once the user asks to wakeup.

And this way, we can wait after eof or when paused, indefinitely without having to periodically check by timing out. We only need to do that with events because it is cleared in the thread so we can't be sure we didn't clear it after a user set it. But, with the queue, this kind of issue doesn't exist.

Also, sorry @tito for hijacking your PR.

kivy/core/video/video_ffpyplayer.py
matham: I did rename the variable

kivy/core/video/video_ffpyplayer.py
matham: I changed the code entirely so this should not be an issue.

matham:
After talking on Discord I can see that what I wrote isn't clear it all. So here's a more specific example of why I had an issue with tito's original PR and suggested changes, and how my current commit fixed it.

The problem with master is that ideally we would not like to wait at all once the kivy thread signals to unpause, exit, etc. Currently in master it sleeps for some ms in a busy loop and responds after this ms timeout, adding up to 200ms latency.

This PR instead originally replaced some of the `sleep` calls with an event that waited the 200ms and then timed out after those ms. So it is still a busy loop of 200ms resolution, but when kivy signals a change, we respond faster because we wake up from the event wait immediately. However, this only fixed it in one place so we still have some sleep calls.

But an even bigger issue is with the design of the event. I tried to explain this in my original response with the suggested changes, but it probably wasn't clear. So I'll assume that we took the approach of the PR and replaced all sleep calls with event waiting, like in https://github.com/kivy/kivy/pull/7469. This design is that the kivy thread sets the event whenever it needs a change from the player, e.g. unpasuing, seeking, stopping. The 2nd thread then checks if it has been set and clears itself if it was set. Originally in this PR, the event always cleared itself, without even checking.

The problem is that this leads to sometimes having to wait the full 200ms timeout when multiple events happen right after the other. It is still strictly better than master, but I thought it could be done more optimally. Here's the problem situation copied from discord:
```
2nd thread: waits for unpausing 
Kivy thread: sets pause and wakes up thread
2nd thread: wakes up
Kivy thread: sets player to stop, wakes up thread again but event is still set so no change
2nd thread: after woken up it goes around the loop to `ensure_idle` which checks event was set, yes, so it resets it
2nd thread: gets next frame, next frame is not yet ready so in `no_idle.wait(val)` it sleep for the full frame duration because event is cleared
```

We could do better using e.g. the queue implementation that I pushed to this PR. Because with a queue, we never have to reset the event so we never lose any signals from the Kivy thread. So with this PR currently, there will never be any latency from the waiting and the thread will respond immediately (except if the thread is busy doing something of course). With the current implementation of the PR, the sequence of events from above will instead happens in one of two ways:

```
2nd thread: waits for unpausing 
Kivy thread: sets pause and wakes up thread by adding to queue
2nd thread: wakes up
Kivy thread: sets player to stop, wakes up thread again but 2nd thread hasn't read the queue yet so no change
2nd thread: after woken up it empties the queue and checks that need close is true so it returns
```
Or in an alternative version,
```
2nd thread: waits for unpausing 
Kivy thread: sets pause and wakes up thread by adding to queue
2nd thread: wakes up and clears queue and goes around loop to get next frame
Kivy thread: sets player to stop, wakes up thread again because queue is now empty so it adds item to queue
2nd thread: gets next frame, next frame is not yet ready so it calls `wait_for_wakeup(val)`, but that returns immediately because there's something on queue so it goes around loop again, sees condition of loop is False because need close is True, so it returns immediately
```

rnixx:
thank you for your work. one thing remaining is that with the change of removing the daemon flag from the frame fetching thread the app won't exit if unload function not gets called.

How about adding an event listener for Window.on_close when the thread starts (and of course removing when the thread stops) to solve this case?

matham:
Hmm, doesn't `__del__` take care of that?

I'm hesitant adding any callbacsk to on_close, because kivy widgets generally aren't monitoring the state of globals such as the current app or window. So this would be a significant change in that direction, which ultimately makes separate modules (core player and window/app) more entangled which makes things harder in the long term. Ideally providers/widgets etc are self contained completely.

But, if `__del__` doesn't fix this issue, then I think we could add something like what you suggest, but instead of here in core video, we should add it to the video widget, because that widget should be responsible for cleaning up after using the core video resource. But also, ideally we would clean up after the app closes, rather than when the window closes, because that makes more sense in terms of the lifecycle of kivy widgets and the app, but there may not be an app present so I guess window is the next best thing, but it still makes me a little uncomfortable. This is part of the general kivy issue with cleanup and global variables/aggressive imports.

So I'm really hoping `__del__` fixes this, can you confirm that it doesn't?

rnixx:
I am currently testing the code in my app. just ran into the double lock acquisition with self._trigger.cancel() problem again when calling unload and currently try to find out what causes this issue (probably just another thing in my app). One this is fixed i can check the general video lifecycle

matham:
Hmm, I wonder though how you could run into double locking with `cancel`. If you look at the method: https://github.com/kivy/kivy/blob/master/kivy/_clock.pyx#L142, it executes no user code, only internal code, which calls no user code. So I don't see how it could ever run into a double lock situation.

The only possibility is that you have a `__del__` method somewhere, which gets called by python while `cancel` holds the lock during it's garbage collection cycle and in your code you call a Clock method. If that's the case, please take a look at https://github.com/kivy/kivy/blob/master/kivy/_clock.pyx#L455.

rnixx:
I have no explicit __del__ function somewhere in my code, but actually the problem indeed appears in the situation where unload() gets called from VideoFFPy.__del__. I try to reproduce with a minimalistic app, but no sucess so far. Maybe a dump question, but is there a situation apart from garbage collection where the __del__ function is supposed to be called in kivy?

rnixx:
And about the application exit. the unload on ``__del__`` works as long as the frame fetching thread is not running. as soon as the video is running, stopping the application simply hangs because the thread still holds a reference to the video instance and never returns, thus we're in the situation where we explicitely need to interrupt the thread.

I tried with this code:

    if __name__ == '__main__':
        from kivy.app import App
        from kivy.uix import video
        import sys

        video.CoreVideo = VideoFFPy

        if len(sys.argv) != 2:
            print("usage: %s file" % sys.argv[0])
            sys.exit(1)

        class VideoApp(App):

            def build(self):
                v = self.v = video.Video(source=sys.argv[1], play=True)
                return v

        VideoApp().run()

rnixx:
And again about the double lock. if i move the trigger.cancel() to _next_frame_run, it works as expected in my app. Would this be an option to you to cancel the trigger from withing the thread?

rnixx:
> ideally we would clean up after the app closes, rather than when the window closes, because that makes more sense in terms of the lifecycle of kivy widgets and the app, but there may not be an app present so I guess window is the next best thing, but it still makes me a little uncomfortable.

I agree. i just tried to bypass kivy entirely for the cleanup task with ``atexit.register``, but unfortunately atexit happens too late. digging further.

rnixx:
This is a diff that the current PR is actually working in all cases i tested.

https://gist.github.com/rnixx/02b632f58d16e884a7d15341c0743241

Maybe there is a cleaner solution to react on application close than binding to window event, but actually this solutions works.

Also it might be better to introduce such behavior in video base class, or imho even better an event on the application object which is supposed to be bound for cleanup tasks.

rnixx:
``self.eos == 'pause'`` in ``_do_eos`` seems to not be reachable any more due to this change:

https://github.com/kivy/kivy/commit/27b36c393502a3468abee86fbccfeb76a52022e0#diff-c5dc0033e87690d91aaa74f321c0ff059fbf6b8260610e3673c145eef2b1ff1dR70

kivy/core/video/video_ffpyplayer.py
```diff
@@ -294,65 +317,101 @@ def _next_frame_run(self):
...
             # Get next frame regular:
             else:
                 frame, val = ffplayer.get_frame()
 
             if val == 'eof':
-                sleep(0.2)
                 if not did_dispatch_eof:
                     self._do_eos()
                     did_dispatch_eof = True
+                wait_for_wakeup(None)
```
rnixx, line 331: If waiting here forever, loop eos does sometimes not work. it seems to start video again, only audio is playing and the loop after never happens.

eof also seems only be retured one time per video loop, so i changed to wait_for_wakeup(1 / 30.) and the video loops since then as it should.

rnixx: UPDATE: actually sometimes it gets val == 'eof' twice and did_dispatch_eof is True, then it runs into the wait which never gets signalled.

matham: I just had to reset `did_dispatch_eof` when seeked to fix this I believe. Can you check if that fixes the multiple eof issue?

matham:
About eos=pause being deprecated, I didn't remove it because I saw it wasn't removed from other video providers, including `__init__.py` and it doesn't hurt to leave for now. But that can be properly deprecated in another PR (so if users manually sets it, it emits a warning, right now it only warns if passed through `__init__`).

About cancling the trigger from under `__del__`, I just removed the cancel entirely because it's not needed. We set next frame and _ffplayer to None just below, meaning that if the clock calls `_redraw` once player is dead, it will just return immediately.

Finally, I reset daemon to True I would rather deal with that in a different PR once this is merged to not further complicate this PR.

kivy/core/video/video_ffpyplayer.py
rnixx: Just tested this change, it seems to work now

rnixx:
> Finally, I reset daemon to True I would rather deal with that in a different PR once this is merged to not further complicate this PR.

Ready to merge then?

PR #7420, opened 2021-03-12
Update pong.rst

devnev39:
Just a spell change.


matham:
Thanks for your contribution, however, the original spelling is actually correct.

In this context "run" means started (i.e. "it is loaded when the application is run/started"). "Running" implies it is loaded sometime during the application runtime, which is correct, but not what is intended to be said (that it is loaded when the application is initially run/started).

devnev39:
Ohh. Sorry Nad thank you very much for replying 👍

devnev39:
> Thanks for your contribution, however, the original spelling is actually correct.
> 
> In this context "run" means started (i.e. "it is loaded when the application is run/started"). "Running" implies it is loaded sometime during the application runtime, which is correct, but not what is intended to be said (that it is loaded when the application is initially run/started).

Ohh. It just slipped through my mind. Thanks for your reply.

PR #7424, opened 2021-03-14
Properties: Speed up bare widget creation (3X) and property dispatching/setting

matham:
Maintainer merge checklist

This modifies EventDispatcher and Property to speed up widget creation and dispatching/setting. Below are some numbers and graphs showing the speedup. For bare widget creation we get a 3x speedup! For other operations the speedup is more modest or is not changed.

You'll see referenced **3 optimization branches** compared to master, following is an explanation of the optimization in each branch. I can separate them into three PRs, if desired, and use this PR to track them overall because each branch builds on the other but are forward-independent, but I'm not sure it'll be easier. For review, **I'm linking the diff** so you can see more clearly what each branch does and that's probably the easiest way to review:

Branches
======

Following are the three independent branches (but they build on each other sequentially).

lazy
----

Diff: https://github.com/kivy/kivy/compare/master...matham:lazy. This is the main change that speeds up widget creation.

Currently, every property added to widget is initialized when the widget is created. However, many properties are not used at all for some widgets, and there's no reason to pay this cost upfront anyway before the properties are used in any way. **This branch initializes properties lazily**. 

Every time we get/set/dispatch a property, we load its property storage and use it to change the property state specific to the widget. This PR changes the direct storage access to instead use a `get_property_storage` method. This allows us to delay initializing the storage for every property until the first time `get_property_storage` is called. Hence, we don't pay the initialization cost before we need the property.

Backward compat
******************

This should not have backwards compat implications, unless you created custom properties in cython and are not updating to the new API. Anything accessible from python public API will remain unchanged/unbroken.

Previously we did:
```py
prop.link(widget, name)
prop.link_deps(widget, name)
```

for each property when initializing the eventdispatcher. `link` is supposed to tell the property its name and create the storage, then when all are "linked" we can use/link to any properties we depend on using our `link_deps`.

[Since 3.6](https://docs.python.org/3/reference/datamodel.html#object.__set_name__) (yay 3.6), descriptors now have a `__set_name__` method that is called by python before the class is fully created and before `__cinit__` is called. It does the name association, so that's what I'm using (we call a new `set_name` name if we create the property manually, e.g. from kv when `__set_name__` is not called by python). So we don't have to call `link` anymore for the name association. 

So all we have to do is create the storage and link to our dependencies. This is all done lazily the first time the property is used and that's when we call `link` and `link_deps` from `get_property_storage` internally. However, for backward compat, if you do call `link` and `link_deps` directly, which we allowed previously from python, it immediately initializes everything so old code using them from python is not broken.

link_eagerly
************

There is one slight niggle. `ReferenceListProperty` binds to other properties, say `NumericProperty`. Now, imagine the `ReferenceListProperty` was not used yet, so it doesn't track the `NumericProperty` because of the lazy binding. But `NumericProperty` changes yet it doesn't cause a `ReferenceListProperty` dispatch. So far so good because given that `ReferenceListProperty` hasn't been accessed in any way, it doesn't matter that it's not dispatched (does a tree falling in the woods make a sound kinda thing). The real issue is, given `ReferenceListProperty` has a getter, this getter would be dispatched normally when `NumericProperty` changes, but now it isn't. And if the getter has side-effects, it doesn't happen.

So, I added a `link_eagerly` method. This is called when the eventdispatcher is created for all properties, instead of `link`/`link_deps`. For most properties it does nothing. For `ReferenceListProperty` however it does make sure that it and the properties it depends on are initialized early.

But, most `ReferenceListProperty` getters don't have side-effects, including the ones in `widget.py`. So I added `watch_before_use` as a parameter to `ReferenceListProperty`. It defaults to True for safety. But if set to False (e.g. like I did in in widget.py) then `ReferenceListProperty` becomes lazy and `link_eagerly` does nothing. This gained back a lot of perf for `Widget`. I don't exactly like this, because it's exposing an internal detail, but users do not have know or use this with `ReferenceListProperty` and it'll just be a bit slower than it could be. See the `watch_before_use` docs for more details.

So, in conclusion. Previously you did 
```py
prop.link(widget, name)
prop.link_deps(widget, name)
```

Now, you can still do that. But to make it faster, you can instead do

```py
prop.link_eagerly(widget)
```

or if you manually create the property you also have to do (e.g. create_property)
```py
prop.set_name(widget, name)
prop.link_eagerly(widget)
```

Tests
*****

This PR also changes properties to use pytest and adds comprehensive testing for all the changes.

type
----

Diff: https://github.com/matham/kivy/compare/lazy...matham:prop-types. This is a small, but significant optimization. In some places we used `isinstance(x, str)` and in others `tp is str`. The latter is far faster and we really shouldn't have to worry about subclasses of str!?? So I used the latter. Similarly, I optimized the order of some type comparison after profiling. There's no reason to check for types that are slow and uncommon first. So instead I check for the common cases first.

args
----

Diff: https://github.com/matham/kivy/compare/prop-types...matham:pass-args. For every interaction with a property, we have to resolve and get the property storage from the widget. This costs cycles. So now, once we have the property storage we just pass it around to sub-methods where it's possible.

This part will break backward compat if you inherit and customize properties from cython because e.g. `cdef convert(self, EventDispatcher obj, value)` became `cdef convert(self, EventDispatcher obj, value, PropertyStorage property_storage)`. However, I don't think we guarantee that our cython purely `cdef` API is stable.

Performance
========

I tested the various branches using our `pytest-benchmark` test. You'll see numbers for each of the 4 branches. I had to test on linux to get stable results, but still the stdev is large so use these numbers with lots of salt. I'm looking more at the overall patterns rather than specific numbers.

Tables
------

The following tables show the time it took for eventdispatcher creation, Widget creation, Widget with little kv creation, Widget with multiple kv lines creation and then property setting, and finally kv parsing (which as expected remains mostly unchanged).

```
--------------------- benchmark 'test_event_dispatcher_creation': 4 tests ---------------------
Name (time in ns)                      Median                Mean              StdDev          
-----------------------------------------------------------------------------------------------
event_dispatcher_creation (lazy)       550.5000 (1.0)      563.1129 (1.0)      175.2989 (1.30)   
event_dispatcher_creation (master)     561.9000 (1.02)     581.0382 (1.03)     182.4781 (1.35)   
event_dispatcher_creation (type)       652.4000 (1.19)     662.1550 (1.18)     135.1069 (1.0)    
event_dispatcher_creation (args)       674.1000 (1.22)     679.0375 (1.21)     157.0804 (1.16)   
-----------------------------------------------------------------------------------------------


------------------- benchmark 'test_widget_creation': 4 tests -------------------
Name (time in us)           Median               Mean            StdDev          
---------------------------------------------------------------------------------
widget_creation (type)       12.1040 (1.0)      12.1666 (1.0)      2.4863 (1.0)    
widget_creation (args)       12.5190 (1.03)     12.6577 (1.04)     3.0162 (1.21)   
widget_creation (lazy)       12.5875 (1.04)     12.7279 (1.05)     2.9898 (1.20)   
widget_creation (master)     35.7390 (2.95)     36.5566 (3.00)     4.7394 (1.91)   
---------------------------------------------------------------------------------


------------------- benchmark 'test_kv_widget_creation': 4 tests -------------------
Name (time in us)              Median               Mean            StdDev          
------------------------------------------------------------------------------------
kv_widget_creation (args)       43.5850 (1.0)      45.8367 (1.0)      8.5321 (1.03)   
kv_widget_creation (type)       43.9140 (1.01)     46.1919 (1.01)     8.9500 (1.08)   
kv_widget_creation (lazy)       44.1890 (1.01)     46.4092 (1.01)     8.2754 (1.0)    
kv_widget_creation (master)     92.5850 (2.12)     94.6561 (2.07)     9.7891 (1.18)   
------------------------------------------------------------------------------------


-------------------- benchmark 'test_complex_kv_widget[create]': 4 tests ---------------------
Name (time in us)                      Median                Mean             StdDev          
----------------------------------------------------------------------------------------------
complex_kv_widget[create] (args)       106.7740 (1.0)      110.7523 (1.0)      13.5875 (1.0)    
complex_kv_widget[create] (lazy)       106.8340 (1.00)     111.2876 (1.00)     14.4386 (1.06)   
complex_kv_widget[create] (type)       107.0610 (1.00)     111.3111 (1.01)     14.6871 (1.08)   
complex_kv_widget[create] (master)     150.1250 (1.41)     156.5556 (1.41)     17.9079 (1.32)   
----------------------------------------------------------------------------------------------


------------------- benchmark 'test_complex_kv_widget[set]': 4 tests -------------------
Name (time in us)                  Median               Mean            StdDev          
----------------------------------------------------------------------------------------
complex_kv_widget[set] (args)       55.4995 (1.0)      57.8086 (1.0)      7.3582 (1.0)    
complex_kv_widget[set] (type)       56.8090 (1.02)     59.9338 (1.04)     8.1425 (1.11)   
complex_kv_widget[set] (lazy)       60.4445 (1.09)     63.2736 (1.09)     7.9872 (1.09)   
complex_kv_widget[set] (master)     75.1813 (1.35)     78.4721 (1.36)     9.6921 (1.32)   
----------------------------------------------------------------------------------------


---------------------- benchmark 'test_parse_kv': 4 tests ---------------------
Name (time in us)       Median                Mean             StdDev          
-------------------------------------------------------------------------------
parse_kv (args)          99.6970 (1.0)      103.4618 (1.0)      11.9808 (1.0)    
parse_kv (type)          99.7370 (1.00)     103.4681 (1.00)     12.2278 (1.02)   
parse_kv (lazy)         101.7080 (1.02)     105.7064 (1.02)     13.2168 (1.10)   
parse_kv (master)       103.4210 (1.04)     107.1135 (1.04)     12.7558 (1.06)   
-------------------------------------------------------------------------------
```

Mean graphs
--------------

The following are graphs showing the mean duration (mean over repeated runs) for other measures. The numbers e.g. 1, 10, 100, 1000 mean e.g. that the test creates 1, 10, 100, etc labels in each run. So e.g. 100 should take approx 10x longer than 10.

mean_Create property
Property creation is now super fast.

mean_Set property
Passing storage as args  instead of accessing each time really speeds up setting properties.

mean_Create label
Creating a label is also faster, although not quite the 3x as a bare Widget. Because label does other slow operations (textures and whatnot).

mean_Dispatch touch
Dispatching is also faster

mean_Render empty
Rendering empty widget canvas is basically the same.

Median graphs
---------------

This is the same as the mean graph, but rather than reporting the mean over repeated runs, it's the median, which is more stable maybe!?

median_Create property

median_Set property

median_Create label

median_Dispatch touch

median_Render empty

Here's the full raw data
[all.txt](https://github.com/kivy/kivy/files/6137768/all.txt)

Overall, while the benchmarks show the performance improvements, I'm not sure this would translate a lot to real-world apps, where often the slowness comes from rendering textures or other operations (e.g. as you can see that creating a label has much less benefit than creating bare widget).

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -694,7 +694,8 @@ cdef class NumericProperty(Property):
     cdef check(self, EventDispatcher obj, value):
         if Property.check(self, obj, value):
             return True
-        if type(value) not in (int, float, long):
+        tp = type(value)
+        if tp is not int and tp is not float and tp is not long:
             raise ValueError('%s.%s accept only int/float/long (got %r)' % (
                 obj.__class__.__name__,
                 self.name, value))
@@ -714,6 +715,8 @@ cdef class NumericProperty(Property):
             ps.numeric_fmt = 'px'
             ps.original_num = x
             return x
+        if tp is str:
+            return self.parse_str(obj, x, ps)
 
         if tp is tuple or tp is list:
             if len(x) != 2:
@@ -721,8 +724,6 @@ cdef class NumericProperty(Property):
                     obj.__class__.__name__,
                     self.name, x))
             return self.parse_list(obj, x[0], x[1])
-        elif isinstance(x, str):
-            return self.parse_str(obj, x)
         else:
             raise ValueError('%s.%s has an invalid format (got %r)' % (
                 obj.__class__.__name__,
@@ -773,6 +774,7 @@ cdef class StringProperty(Property):
             raise ValueError('%s.%s accept only str' % (
                 obj.__class__.__name__,
                 self.name))
+
 
 cdef inline void observable_list_dispatch(object self) except *:
     cdef Property prop = self.prop
@@ -912,6 +914,7 @@ cdef class ListProperty(Property):
         if value is not None:
             value = ObservableList(self, obj, value)
         Property.set(self, obj, value)
+
 
 cdef inline void observable_dict_dispatch(object self) except *:
     cdef Property prop = self.prop
@@ -1068,18 +1071,19 @@ cdef class ObjectProperty(Property):
         `baseclass` parameter added.
     '''
     def __init__(self, defaultvalue=None, rebind=False, **kw):
-        self.baseclass = kw.get('baseclass', object)
+        self.baseclass = kw.get('baseclass')
         super(ObjectProperty, self).__init__(defaultvalue, **kw)
         self.rebind = rebind
 
     cdef check(self, EventDispatcher obj, value):
         if Property.check(self, obj, value):
             return True
-        if not isinstance(value, self.baseclass):
+        if self.baseclass is not None and not isinstance(value, self.baseclass):
             raise ValueError('{}.{} accept only object based on {}'.format(
                 obj.__class__.__name__,
                 self.name,
                 self.baseclass.__name__))
+
 
 cdef class BooleanProperty(Property):
     '''Property that represents only a boolean value.
@@ -1420,7 +1424,8 @@ cdef class ReferenceListProperty(Property):
         self.dispatch(obj)
 
     cdef convert(self, EventDispatcher obj, value):
-        if not isinstance(value, (list, tuple)):
+        tp = type(value)
+        if tp is not list and tp is not tuple:
             raise ValueError('%s.%s must be a list or a tuple' % (
                 obj.__class__.__name__,
                 self.name))
@@ -1767,7 +1772,7 @@ cdef class VariableListProperty(Property):
         # reset here, it'll be changed in parse is we use anything that is not px
         ps.uses_scaling = 0
         try:
-            if isinstance(x, (list, tuple)):
+            if tp is tuple or tp is list:
                 original = list(x)
                 l = len(x)
                 if l == 1:
@@ -1804,7 +1809,7 @@ cdef class VariableListProperty(Property):
                     elif self.length == 2:
                         err = '%s.%s must have 1 or 2 components (got %r)'
                     raise ValueError(err % (obj.__class__.__name__, self.name, x))
-            elif tp is int or tp is long or tp is float or isinstance(x, str):
+            elif tp is int or tp is long or tp is float or tp is str:
                 y = self._convert_numeric(obj, x)
                 if self.length == 4:
                     return [y, y, y, y]
@@ -1828,14 +1833,14 @@ cdef class VariableListProperty(Property):
         tp = type(x)
         if tp is int or tp is float or tp is long:
             return x
+        if tp is str:
+            return self.parse_str(obj, x)
         if tp is tuple or tp is list:
             if len(x) != 2:
                 raise ValueError('%s.%s must have 2 components (got %r)' % (
                     obj.__class__.__name__,
                     self.name, x))
             return self.parse_list(obj, x[0], x[1])
-        elif isinstance(x, str):
-            return self.parse_str(obj, x)
         else:
             raise ValueError('%s.%s has an invalid format (got %r)' % (
                 obj.__class__.__name__,
@@ -2199,7 +2204,7 @@ cdef class ColorProperty(Property):
             return x
         cdef object color = x
         try:
-            if isinstance(x, str):
+            if type(x) is str:
                 color = self.parse_str(obj, x)
             color = self.parse_list(obj, color)
         except (ValueError, TypeError) as e:
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -564,7 +564,7 @@ cdef class Property:
                 raise e
 
         ps.value = value
-        self.dispatch(obj)
+        self._dispatch(obj, ps)
         return True
 
     cpdef get(self, EventDispatcher obj):
@@ -602,6 +602,9 @@ cdef class Property:
         '''
         return x
 
+    cdef _dispatch(self, EventDispatcher obj, PropertyStorage ps):
+        ps.observers.dispatch(obj, ps.value, None, None, 0)
+
     cpdef dispatch(self, EventDispatcher obj):
         '''Dispatch the value change to all observers.
 
@@ -619,7 +622,7 @@ cdef class Property:
 
         '''
         cdef PropertyStorage ps = self.get_property_storage(obj)
-        ps.observers.dispatch(obj, ps.value, None, None, 0)
+        self._dispatch(obj, ps)
 
 
 cdef class NumericProperty(Property):
@@ -1420,7 +1423,7 @@ cdef class ReferenceListProperty(Property):
                     [prop.get(obj) for prop in p],
                     update_properties=False)
 
-        self.dispatch(obj)
+        self._dispatch(obj, ps)
 
     cdef convert(self, EventDispatcher obj, value, PropertyStorage property_storage):
         tp = type(value)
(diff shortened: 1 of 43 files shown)


kivy/properties.pyx
```diff
@@ -387,15 +387,55 @@ cdef class Property:
...
+    cdef PropertyStorage get_property_storage(self, EventDispatcher obj):
+        cdef PropertyStorage ps = obj.__storage.get(self._name)
+        if ps is None:
+            self.link(obj, self._name)
+            self.link_deps(obj, self._name)
+            ps = obj.__storage[self._name]
+        return ps
+
+    def __set_name__(self, owner, name):
+        if name == 'touch_down' or name == 'touch_move' or name == 'touch_up':
```
pythonic64, line 407: Can we move this `touch_xxx` and other event name tests in `register_event_type` method? Then `register_event_type` can check if property name or event name already exists and raise an error?

kivy/properties.pyx
```diff
@@ -1699,53 +1773,54 @@ cdef class VariableListProperty(Property):
         # reset here, it'll be changed in parse is we use anything that is not px
         ps.uses_scaling = 0
         try:
-            if isinstance(x, (list, tuple)):
+            if tp is int or tp is long or tp is float or isinstance(x, str):
```
pythonic64, line 1776: Why not testing with `type(x) is str` instead of `isinstance(x, str)` when former is faster?
Type `long` is merged with `int` in Python3, so can we remove all `long` occurrences?

kivy/properties.pyx
matham: While I understand what you mean and it does seem like a better idea, there are multiple issues with it. Currently we only ensure that someone doesn't create a *kivy* property that would have a event handler with the same name. But we don't and shouldn't stop people from creating properties with that name. I.e.:

```py
class A(Widget):
    def __init__(self):
        self.touch_down =True
```

But perhaps that's not what you meant. But rather that when creating a Kivy prop we check whether an event with that name exists and raise an exception, and similarly raise an exception when creating an event when a property with that name already exists. Currently we don't do it, except for these few special names so doing it now could easily break existing code.

Perhaps we should only do it when registering these automatic `on_xxx` handlers when the instance is created, because that's the only reason it's a problem (and perhaps for kv `on_xxx: ` handlers) because of the ambiguity. But that couldn't be part of this PR and would need a discussion to figure out if the benefits are worth the potential breakage, and I'm not sure about it.

kivy/properties.pyx
matham: I guess you're right and I overlooked it. See https://github.com/kivy/kivy/pull/7439.

kivy/properties.pyx
pythonic64: I meant we check if event with property name already exists before registering the event and check if event name exist before registering the property. If any exists then raise the error.

matham: This would have to happen in EventDispatcher init as we register events, because this code is executed before any events are registered. But we could raise an exception if an event is registered for an existing property.

Similarly, we could also raise an exception if KV or user manually create a property and an event with the name exists.

But that would need a PR of its own with tests etc. And, the bigger issue, that would be a breaking change. Perhaps it's not a problem given that if both a property and event with same name exists the user code may not function as they expected. But it would need discussion to see if the break is worth it. I'm on the fence.

farzadpanahi:
@matham could you provide the machine specs you used to run the benchmarks on? We are using Kivi on RPi3 and our widget creation time is around 20ms each !

PR #7389, opened 2021-02-25
Mouse: Fix mouse being offset by 2 pixels vertically

pythonic64:
Fixes https://github.com/kivy/kivy/issues/7388.

Maintainer merge checklist:
* [x] Title is descriptive/clear for inclusion in release notes.
* [x] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [ ] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #7425, opened 2021-03-15
Mouse: Fix computation of relative touch position in MouseMotionEventProvider

pythonic64:
This changes fix computation of relative touch position in `MouseMotionEventProvider` and scaling of events in `MotionEvent.scale_for_screen` method which gives correct positions when events are dispatched through `on_touch_xxx` methods.

The fix changes the position range from `[0, Window.system_size[0]]` and `[0, Window.system_size[1]]` to `[0, Window.system_size[0] - 1]` and `[0, Window.system_size[1] - 1]`.

We most likely need the same fix for other providers which compute relative touch position when creating motion events.

Maintainer merge checklist

### DIFF
--- a/kivy/input/motionevent.py
+++ b/kivy/input/motionevent.py
@@ -355,27 +355,32 @@ class MotionEvent(MotionEventBase):
 
     def scale_for_screen(self, w, h, p=None, rotation=0,
                          smode='None', kheight=0):
-        '''Scale position for the screen
+        '''Scale position for the screen.
+
+        .. versionchanged:: 2.1.0
+            Max value for `x`, `y` and `z` is changed respectively to `w` - 1,
+            `h` - 1 and `p` - 1.
         '''
+        x_max, y_max = w - 1.0, h - 1.0
         sx, sy = self.sx, self.sy
         if rotation == 0:
-            self.x = sx * float(w)
-            self.y = sy * float(h)
+            self.x = sx * x_max
+            self.y = sy * y_max
         elif rotation == 90:
             sx, sy = sy, 1 - sx
-            self.x = sx * float(h)
-            self.y = sy * float(w)
+            self.x = sx * y_max
+            self.y = sy * x_max
         elif rotation == 180:
             sx, sy = 1 - sx, 1 - sy
-            self.x = sx * float(w)
-            self.y = sy * float(h)
+            self.x = sx * x_max
+            self.y = sy * y_max
         elif rotation == 270:
             sx, sy = 1 - sy, sx
-            self.x = sx * float(h)
-            self.y = sy * float(w)
+            self.x = sx * y_max
+            self.y = sy * x_max
 
-        if p:
-            self.z = self.sz * float(p)
+        if p is not None:
+            self.z = self.sz * (p - 1.0)
 
         if smode:
             if smode == 'pan':
--- a/kivy/input/motionevent.py
+++ b/kivy/input/motionevent.py
@@ -361,7 +361,7 @@ class MotionEvent(MotionEventBase):
             Max value for `x`, `y` and `z` is changed respectively to `w` - 1,
             `h` - 1 and `p` - 1.
         '''
-        x_max, y_max = w - 1.0, h - 1.0
+        x_max, y_max = max(0.0, w - 1.0), max(0.0, h - 1.0)
         sx, sy = self.sx, self.sy
         if rotation == 0:
             self.x = sx * x_max
@@ -380,7 +380,7 @@ class MotionEvent(MotionEventBase):
             self.y = sy * x_max
 
         if p is not None:
-            self.z = self.sz * (p - 1.0)
+            self.z = self.sz * max(0.0, p - 1.0)
 
         if smode:
             if smode == 'pan':
--- a/kivy/input/providers/mouse.py
+++ b/kivy/input/providers/mouse.py
@@ -228,9 +228,9 @@ class MouseMotionEventProvider(MotionEventProvider):
         cur.clear_graphics(EventLoop.window)
 
     def on_mouse_motion(self, win, x, y, modifiers):
-        width, height = EventLoop.window.system_size
-        rx = x / float(width)
-        ry = 1. - y / float(height)
+        width, height = win.system_size
+        rx = x / (width - 1.0)
+        ry = 1.0 - y / (height - 1.0)
         if self.current_drag:
             cur = self.current_drag
             cur.move([rx, ry])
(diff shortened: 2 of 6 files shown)


tshirtman:
i don't understand, why would system_size be 1px less than size in each direction?

pythonic64:
For example, if window is [800, 600] in size, values for `x` and `y` in provider take values from [0, 799] and [0, 599] but they never take 800 or 600. For height of 600 value for `y` is between 0 and 599, but never 600 and therefore motion event will never get 1.0 as relative value for `y` as `y` will never be 600. 

This was partially fixed by https://github.com/kivy/kivy/pull/7389 and you can see this by inspecting values for `mouse_pos`:
```
from kivy.app import App
from kivy.lang import Builder

kv = '''
#:import win kivy.core.window.Window

BoxLayout:
    orientation: 'vertical'
    Label:
        font_size: '42sp'
        text: 'System size: ' + str(win.system_size)
    Label:
        font_size: '42sp'
        text: 'Mouse pos: ' + str(win.mouse_pos)
'''


class MousePosBugApp(App):

    def build(self):


if __name__ == '__main__':
    MousePosBugApp().run()
```

matham:
Do you think that it should instead be `max(0, val - 1)`, in case the size is zero?

pythonic64:
I did the zero check in the provider, but on the other hand do we expect screen (window) width or height to ever be 0? If we remove zero checks we do get a better performance considering that single/multi touch (and later hover) do get dispatched a lot.

Currently, error is raised when 0 is set as width or height of window. 
```
   File "D:\Work\Kivy Framework\kivy-event-manager\kivy\core\window\__init__.py", line 967, in <lambda>
     keyboard_height=lambda *dt: self.update_viewport())
   File "D:\Work\Kivy Framework\kivy-event-manager\kivy\core\window\__init__.py", line 1509, in update_viewport
     projection_mat.view_clip(0.0, w, 0.0, h, -1.0, 1.0, 0)
   File "kivy\graphics\transformation.pyx", line 275, in kivy.graphics.transformation.Matrix.view_clip
     cpdef Matrix view_clip(Matrix self, double left, double right,
   File "kivy\graphics\transformation.pyx", line 301, in kivy.graphics.transformation.Matrix.view_clip
     raise ValueError('invalid frustrum')
 ValueError: invalid frustrum
```

So we have two options:
1. Stay safe and assume that window width or height can be 0
2. Assume that window height width is always greater that 0 and state that in `MotionEvent.scale_for_screen` docs

tshirtman:
yeah, the error makes sense, i don’t think it’s possible to create an opengl context with a width or height of 0, at least for the intent and purpose of kivy, so i wouldn’t bother too much. If we get in that situation, we have bigger issues to fix before.

I understand that the coordinate will never be WIDTH or HEIGHT, it can only be on the last pixel before that, or it’s out of the window, but should the relative value be 1, or is it in `[0, 1[` ? If 1 can be reached, then ok, we certainly need to remove 1 from the mesure, but if 1 is excluded, then it’s fine.

matham:
Well, even if we cannot practically have a gl context of size zero, that is not really part of the assumption (at least explicitly) of the input modules so it'd be more robust if we use `max` as suggested. I don't think it would contribute much to performance issues (touch dispatching only happens once per clock frame so in general that part of the code is not as performance sensitive FYI). So I would vote for option 1.

pythonic64:
Without changes in the provider 1 as a relative value will never be reached unless going out of the window. If value close to 1 is used and touch is rotated then 0 will never be reached, but only close to 0.

I'll add code for the option 2.

matham:
I guess the tests are breaking, probably because they assumed the old behavior. Makes me wonder if we should label this as an api-break in case people assumed the old behavior in their apps (although I can't quite see where).

matham (dismissed):

AndiEcker:
> yeah, the error makes sense, **i don’t think it’s possible to create an opengl context with a width or height of 0**, at least for the intent and purpose of kivy, so i wouldn’t bother too much. If we get in that situation, we have bigger issues to fix before.
> 
> I understand that the coordinate will never be WIDTH or HEIGHT, it can only be on the last pixel before that, or it’s out of the window, but should the relative value be 1, or is it in `[0, 1[` ? If 1 can be reached, then ok, we certainly need to remove 1 from the mesure, but if 1 is excluded, then it’s fine.

Agreeing that a window with a width or height of 0 makes no sense, but it will for opengl contexts respectively widget-canvases. like e.g. in the [ModalView in style.kv](https://github.com/kivy/kivy/blob/master/kivy/data/style.kv#L501) or for widgets that are used as placeholder/distance or are dynamically hideable.

tshirtman:
Oh of course we can have invisibly small widgets, i used that a lot of times, but it’s different, a canvas is not an opengl context with its matrices (projection, etc) and all, it just a group of instructions in an opengl context. The frustrum error is really about math, an opengl context needs a geometry to project 3D objects on a plane, and if the plane is actually a line or a point, (no with and/or no height) then the math breaks lose (division / 0, kind of, i’m not deep into the details), so as far as i know, this situation won’t happen.
Now, as matham said, there is no need to bring that technical detail into the considerations of the input system, that’s just not relevant to it, and the input system could potentially do these computations at a point where there is no consideration about the opengl context, and the window happens to be infinitely thin.

AndiEcker:
thanks @tshirtman for your clarifications - I really appreciate your helping comments and enjoy learning and discovering the huge Kivy framework more and more.

Thanks also to all maintainers and contributors like @matham and @pythonic64 for your great work!!!

pythonic64:
Should we round `x`, `y` and `z` values in motion event? Converting from absolute to relative values in provider and then relative to absolute in `scale_for_screen` method does not produce the same absolute values, for some cases they are equal and for other they are almost equal.

A test app to reproduce:
```
from kivy.app import App
from kivy.lang import Builder

kv = '''
#:import win kivy.core.window.Window

BoxLayout:
    orientation: 'vertical'
    Label:
        font_size: '32sp'
        text: 'System size: ' + str(win.system_size)
    Label:
        font_size: '32sp'
        on_touch_down: self.text = str(args[1].pos)
        on_touch_move: self.text = str(args[1].pos)
        on_touch_up: self.text = ''
'''


class TouchPosBugApp(App):

    def build(self):


if __name__ == '__main__':
    TouchPosBugApp().run()
```

If we round the values then when comparing positions we can test if they are exactly equal instead of almost equal. This can be useful in unit tests, like when testing expected positions of touch events or when testing collisions. Without rounding, comparing exact positions should not be considered safe and we should advise usage of almost equal with some delta or rounding before the equality test.

pythonic64:
You're welcome @AndiEcker

matham:
I think even if we round we should still use approximate equality tests for floats, exact comparison is never a great idea for this kind of thing (unless we know for sure it'll work). I don't like converting to int because that reduces information and doesn't feel right to me. We haven't made that assumption in the past that touch would be limited to the int grid, so we shouldn't now.

So I'd prefer the inexact equality (`isclose`) solution.

pythonic64:
I fixed the tests.
@matham you can review and merge.

matham (approved):

PR #7428, opened 2021-03-16
Modalview win code cleanup

AndiEcker:
This is commit 1 of 2 - if you agree with the changes done in this first step,
I'd like to go even further and replace the current `_window` property of
`ModalView` with the main window `kivy.core.window.Window`, because
`_window` is currently only used as a flag if the modal view is
opened/visible (holding `None` if not opened or the `Window` instance
if opened).

The only place where the window instance is used is in the declaration
of `ModalView` (in style.kv), which could easily be replaced with `Window`.

The opening state of the modal view will then be represented by a an
added `is_opened` property (will be added in commit 2 of 2).

This will make the modal view faster, more readable/understandable and
easier to maintain/test.

Maintainer merge checklist


AndiEcker:
sorry for the noise - but I accidentally mixed up this branch/PR with an old one - therefore closed this one and did a new PR #7434

PR #7433, opened 2021-03-18
Fix source typos

luzpaz:
Changes to API

### DIFF
--- a/kivy/lib/pango/pangoft2.pxi
+++ b/kivy/lib/pango/pangoft2.pxi
@@ -354,7 +354,7 @@ cdef extern from "pango/pango-layout.h" nogil:
     void pango_layout_set_height(PangoLayout *layout, int height)
     int pango_layout_get_height(PangoLayout *layout)
     void pango_layout_set_spacing(PangoLayout *layout, int spacing)
-    int pango_layout_get_spacing(PangoLayout *laout)
+    int pango_layout_get_spacing(PangoLayout *layout)
     void pango_layout_set_indent(PangoLayout *layout, int indent)
     int pango_layout_get_indent(PangoLayout *layout)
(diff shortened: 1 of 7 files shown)


kivy/lib/pango/pangoft2.pxi
```diff
@@ -354,7 +354,7 @@ cdef extern from "pango/pango-layout.h" nogil:
     void pango_layout_set_height(PangoLayout *layout, int height)
     int pango_layout_get_height(PangoLayout *layout)
     void pango_layout_set_spacing(PangoLayout *layout, int spacing)
-    int pango_layout_get_spacing(PangoLayout *laout)
+    int pango_layout_get_spacing(PangoLayout *layout)
```
luzpaz, line 357: couldn't find this typo upstream. I believe it is a typo. I have not tested it.

kivy/tools/report.py
```diff
@@ -60,8 +60,8 @@ def send_report(dict_report):
                 "content": "\n".join(dict_report['Configuration']),
                 "type": 'text'
             },
-            "Input Availablity.txt": {
-                "content": "\n".join(dict_report['InputAvailablity']),
+            "Input Availability.txt": {
+                "content": "\n".join(dict_report['InputAvailability']),
```
luzpaz, line 64: I'm just bringing these to you attention. I can revert if necessary.

tshirtman: i think it's fine to fix this one as well, it's just a tool, not part of the kivy api, not worth keeping an error around for the sake of backward compatibility :)

tshirtman (approved):
I don't think there is any issue with fixing these as well, sure, there is the off-chance that someone used the arguments as keywords in their code, but as these were documented as positional, i doubt that's the case, especially with misspelling, that would have been pretty unnatural.

matham (approved):

Issue #7412, opened 2021-03-09
Support for multiple windows on Windows 10 / MacOS


YarostheLaunchpadder:
**Is your feature request related to a problem? Please describe.**
I am creating different types of software with **Kivy** and some of them I would prefer to use multiple windows for settings for example.

**Describe the solution you'd like**
Ability of creating multiple windows inside the main App class like you can do in _Tkinter_ for example.

**Describe alternatives you've considered**
If I absolutely need it, I just create two python files and one launches another one, then when compiling, I just compile both and place in the same folder

matham:
I'm afraid this just isn't something we ever plan to implement. The suggested approach is to use subprocess/multiprocess to start another kivy window and use a queue or pipes to communicate between them.

AndiEcker:
Looks like there were plans in the past to support multiple windows - e.g.
at least within Kivy's `ModalView` there is the _search_window method
<https://github.com/kivy/kivy/blob/fec099e6ecf19365d54086c8c42e33ae3c028744/kivy/uix/modalview.py#L189>
using the Widget methods `get_parent_window` and `get_root_window` which
are "climbing" up the widget tree to find *a* window instance and only if
not found it will use the main window instance (kivy.core.window.Window).

But if there are no plans for the near future to support multiple windows,
I'd like to make a PR to speedup and simplify the code by removing the
`_search_window` method of ' ModalView'  and change the only usage of this
private method (in the 'open' method) to something simpler like e.g.::
`self._window = Window`.

Am So., 14. März 2021 um 19:06 Uhr schrieb matham ***@***.***

matham:
>`self._window = Window`.

I think that's a fair change given that there isn't likely to be multiple windows in the near or far future. Unless I'm missing something about modal @tshirtman ?

PR #7434, opened 2021-03-18
ModalView: code cleanup regarding detection of main-Window:

AndiEcker:
#7412 -discussion-follow-up from 14-Mar-2021 with @matham (pinging @tshirtman).

- removed unneeded _search_window() method and attach_to property.
- added undocumented events on_pre_open and on_pre_dismiss to module docstring.
- allow stacking of multiple ModalView instances.
- simplified demo code.
- simplified unit test (test_uix_modal.py).

This is a WIP commit (1 of 2) - if you agree with the changes done in this first step, I'd like to go even further and replace the current `_window` property of ModalView with the main window `kivy.core.window.Window`, because `_window` is currently only used as a flag if the modal view is opened/visible (holding `None` if not opened or the main window instance if opened).

The only place where the window instance is accessed is in the declaration of ModalView (in style.kv), which could also be replaced with `Window`.

The opening state of the modal view could then be represented by an added `is_open` property (will be added in commit 2 of 2).

These changes would make the modal view faster, more readable/understandable, and easier to maintain/test.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/modalview.py
+++ b/kivy/uix/modalview.py
@@ -77,7 +77,6 @@ returning `True` from your callback::
 __all__ = ('ModalView', )
 
 from kivy.animation import Animation
-from kivy.core.window import Window
 from kivy.properties import (
     StringProperty, BooleanProperty, ObjectProperty, NumericProperty,
     ListProperty, ColorProperty)
@@ -119,7 +118,6 @@ class ModalView(AnchorLayout):
     defaults to True.
     '''
 
-    # noinspection PyArgumentEqualDefault
     attach_to = ObjectProperty(None, deprecated=True)
     '''If a widget is set on attach_to, the view will attach to the nearest
     parent window of the widget. If none is found, it will attach to the
@@ -180,7 +178,6 @@ class ModalView(AnchorLayout):
 
     # Internals properties used for graphical representation.
 
-    # noinspection PyArgumentEqualDefault
     _anim_alpha = NumericProperty(0)
 
     _anim_duration = NumericProperty(.1)
@@ -206,6 +203,7 @@ class ModalView(AnchorLayout):
             view.open(animation=False)
 
         """
+        from kivy.core.window import Window
         if self._is_open:
             return
         self._window = Window
@@ -255,7 +253,7 @@ class ModalView(AnchorLayout):
 
     def _align_center(self, *_args):
         if self._is_open:
-            self.center = Window.center
+            self.center = self._window.center
 
     def on_touch_down(self, touch):
         """ touch down event handler. """
@@ -316,6 +314,7 @@ class ModalView(AnchorLayout):
 if __name__ == '__main__':
     from kivy.base import runTouchApp
     from kivy.uix.button import Button
+    from kivy.core.window import Window
     from kivy.uix.label import Label
     from kivy.uix.gridlayout import GridLayout
 
--- a/kivy/uix/modalview.py
+++ b/kivy/uix/modalview.py
@@ -1,11 +1,11 @@
-'''
+"""
 ModalView
 =========
 
 .. versionadded:: 1.4.0
 
 The :class:`ModalView` widget is used to create modal views. By default, the
-view will cover the whole "parent" window.
+view will cover the whole "main" window.
 
 Remember that the default size of a Widget is size_hint=(1, 1). If you don't
 want your view to be fullscreen, either use size hints with values lower than
@@ -32,7 +32,7 @@ the ModalView instance::
 
     view.dismiss()
 
-Both :meth:`ModalView.open` and :meth:`ModalView.dismiss` are bindable. That
+Both :meth:`ModalView.open` and :meth:`ModalView.dismiss` are bind-able. That
 means you can directly bind the function to an action, e.g. to a button's
 on_press ::
 
@@ -51,10 +51,12 @@ on_press ::
 ModalView Events
 ----------------
 
-There are two events available: `on_open` which is raised when the view is
-opening, and `on_dismiss` which is raised when the view is closed.
-For `on_dismiss`, you can prevent the view from closing by explicitly returning
-True from your callback. ::
+There are four events available: `on_pre_open` and `on_open` which are raised
+when the view is opening; `on_pre_dismiss` and `on_dismiss` which are raised
+when the view is closed.
+
+For `on_dismiss`, you can prevent the view from closing by explicitly
+returning `True` from your callback::
 
     def my_callback(instance):
         print('ModalView', instance, 'is being dismissed, but is prevented!')
@@ -70,19 +72,20 @@ True from your callback. ::
     keyboard if the :attr:`ModalView.auto_dismiss` property is True (the
     default).
 
-'''
+"""
 
 __all__ = ('ModalView', )
 
-from kivy.logger import Logger
 from kivy.animation import Animation
+from kivy.core.window import Window
+from kivy.properties import (
+    StringProperty, BooleanProperty, ObjectProperty, NumericProperty,
+    ListProperty, ColorProperty)
 from kivy.uix.anchorlayout import AnchorLayout
-from kivy.properties import StringProperty, BooleanProperty, ObjectProperty, \
-    NumericProperty, ListProperty, ColorProperty
 
 
 class ModalView(AnchorLayout):
-    '''ModalView class. See module documentation for more information.
+    """ModalView class. See module documentation for more information.
 
     :Events:
         `on_pre_open`:
@@ -102,23 +105,14 @@ class ModalView(AnchorLayout):
     .. versionchanged:: 2.0.0
         Added property 'overlay_color'.
 
-    '''
+    """
 
-    auto_dismiss = BooleanProperty(True)
+    auto_dismiss = BooleanProperty()
     '''This property determines if the view is automatically
     dismissed when the user clicks outside it.
 
     :attr:`auto_dismiss` is a :class:`~kivy.properties.BooleanProperty` and
     defaults to True.
-    '''
-
-    attach_to = ObjectProperty(None)
-    '''If a widget is set on attach_to, the view will attach to the nearest
-    parent window of the widget. If none is found, it will attach to the
-    main/global Window.
-
-    :attr:`attach_to` is an :class:`~kivy.properties.ObjectProperty` and
-    defaults to None.
     '''
 
     background_color = ColorProperty([1, 1, 1, 1])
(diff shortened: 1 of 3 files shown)


AndiEcker:
@matham @tshirtman 

could you please review this PR.

The last/3rd commit did revert the changes in style.kv.
Had to restore the `_window` property because neither
`Window` nor `app` can be used in style.kv.

modalview.py is now 20 lines shorter and the
unit tests got 20 lines more.

kivy/uix/modalview.py
```diff
@@ -102,25 +105,16 @@ class ModalView(AnchorLayout):
...
-    auto_dismiss = BooleanProperty(True)
+    auto_dismiss = BooleanProperty()
     '''This property determines if the view is automatically
     dismissed when the user clicks outside it.
 
     :attr:`auto_dismiss` is a :class:`~kivy.properties.BooleanProperty` and
     defaults to True.
     '''
 
-    attach_to = ObjectProperty(None)
```
tshirtman, line 115: it might not be an api break, because if anyone sets a value for this, the parent window will be the main one anyway, but do we need to document that the property now does nothing? It's unlikely that people bound anything to this property, but it would break their code if they did, so maybe leaving it in with a statement that it has no behavior now and will be removed in the future would be better?

tshirtman (approved):
Looks good to me, just not completely comfortable about the small api break, but it might be just fine.

kivy/uix/modalview.py
AndiEcker: Thank you very much @tshirtman  for your review/approval!!! :+1: 

Getting your concerns and therefore just added (in the 4rd/last commit) a `versionchanged` to
the docstring of the modalview module regarding the removal of `attach_to` (instead of marking
it as deprecated, to keep the Kivy code base more compact and clean).

As you said setting/using this property will make no difference, so an old project using it could
easily be fixed by just removing any reference to `ModalView.attach_to` - hope you agree!!

kivy/uix/modalview.py
```diff
@@ -70,19 +72,27 @@ def my_callback(instance):
     keyboard if the :attr:`ModalView.auto_dismiss` property is True (the
     default).
 
-'''
+.. versionchanged:: 2.1.0
+    Removed `attach_to` property. In contrary to :attr:`DropDown.attach_to`
```
matham, line 76:
```suggestion
.. versionchanged:: 2.1.0

    Removed `attach_to` property. In contrary to :attr:`DropDown.attach_to`
```

I think you need a line break there.

kivy/uix/modalview.py
```diff
@@ -102,25 +112,16 @@ class ModalView(AnchorLayout):
     .. versionchanged:: 2.0.0
         Added property 'overlay_color'.
 
-    '''
+    """
 
-    auto_dismiss = BooleanProperty(True)
+    auto_dismiss = BooleanProperty()
```
matham, line 117: While it does default to True, it should be explicitly provided. It just makes it harder to follow if the default isn't provided and doesn't really help. So please add it back.

kivy/uix/modalview.py
```diff
@@ -172,11 +173,13 @@ class ModalView(AnchorLayout):
 
     # Internals properties used for graphical representation.
 
-    _anim_alpha = NumericProperty(0)
+    _anim_alpha = NumericProperty()
```
matham, line 176: Please put back the default property values. There's no reason to remove them, so removing them can only contribute to bugs and it helps with code clarity being able to see the default value, even if we know what it defaults to when not provided. Same for all the other places.

kivy/uix/modalview.py
```diff
@@ -285,32 +270,38 @@ def on_touch_up(self, touch):
...
             self._real_remove_widget()
 
     def _real_remove_widget(self):
-        if self._window is None:
+        if not self._is_open:
             return
-        self._window.remove_widget(self)
-        self._window.unbind(
+        Window.remove_widget(self)
+        Window.unbind(
```
matham, line 282: While technically your change is ok, I would feel better if you leave this unchanged. Because although `self._window` is `Window`, it is possible for the Window to have changed, e.g. during testing when we may create a new window for each test (although currently we actually re-use it). 

Also, on principle, it's better to remove/unbind from from `self._window` because we know that is the original object we use for sure, while Window, although very likely to also be the same object, ould possibly be changed in the future so it's less robust.

kivy/uix/modalview.py
```diff
@@ -186,70 +189,49 @@ def __init__(self, **kwargs):
...
-            window = Window
-        return window
-
-    def open(self, *largs, **kwargs):
-        '''Show the view window from the :attr:`attach_to` widget. If set, it
-        will attach to the nearest window. If the widget is not attached to any
-        window, the view will attach to the global
-        :class:`~kivy.core.window.Window`.
+    def open(self, *_args, **kwargs):
+        """ attach this view to the main window.
```
matham, line 193:
```suggestion
        """Display the modal in the Window.
```

kivy/uix/modalview.py
```diff
@@ -186,70 +189,49 @@ def __init__(self, **kwargs):
...
-        if self._window is not None:
-            Logger.warning('ModalView: you can only open once.')
-            return
-        # search window
-        self._window = self._search_window()
-        if not self._window:
-            Logger.warning('ModalView: cannot open view, no window found.')
-            return
+        """
+        self._window = Window
```
matham, line 201:
```suggestion
        """
        if self._is_open:
            return
        self._window = Window
```
Don't we still need to check if it's already open?

kivy/uix/modalview.py
matham: Yeah, I am similarly concerned by the API break.

And like @tshirtman suggested, it's better to leave the property in place saying it's not used anymore and remove it in a latter version. rather than abruptly dropping it. 

It can be done by adding `deprecated` arg as follows: `attach_to = ObjectProperty(None, deprecated=True)`. And update the docs that it is not yet dropped, but will be in the future.

matham (changes requested):

kivy/uix/modalview.py
AndiEcker: Thank you very much @matham for your review!!! Sphinx does not need a line break here (see e.g. [https://kivy.org/doc/stable/api-kivy.core.text.html](https://kivy.org/doc/stable/api-kivy.core.text.html).

kivy/uix/modalview.py
AndiEcker: Ok, although I personally prefer short/clean code, especially in cases like this, where the default value is **explicitly** documented.

kivy/uix/modalview.py
AndiEcker: Ok

kivy/uix/modalview.py
AndiEcker: good catch

kivy/uix/modalview.py
AndiEcker: Reworded and moved it to the `ModalView` class comment, also to resolve your other change request (keep the property, and mark it as deprecated).

tshirtman (approved):
LGTM

kivy/uix/modalview.py
```diff
@@ -112,7 +119,8 @@ class ModalView(AnchorLayout):
     defaults to True.
     '''
 
-    attach_to = ObjectProperty(None)
+    # noinspection PyArgumentEqualDefault
```
matham, line 122:
```suggestion
```

Can you please remove these? They shouldn't be in the repo. It's not required to pass our tests and this is a pycharm specific false-positive warning because this code is perfectly good python.

If we start accepting IDE specific flags, or even general flags soon they'll be all over the codebase. The point of linting is to help, not to become an additional piece of code that needs to be managed. I understand it's annoying to see these pycharm highlighting warnings, but they are meant to help and if the goal then becomes to make pycharm show no warnings then it kinda lost the point of being there to help write better code, not wrangle with the IDE.

I'll link to a video that @tshirtman loves to quote about this topic: https://www.youtube.com/watch?v=wf-BqAjZb8M.

kivy/uix/modalview.py
```diff
@@ -172,11 +180,14 @@ class ModalView(AnchorLayout):
 
     # Internals properties used for graphical representation.
 
+    # noinspection PyArgumentEqualDefault
```
matham, line 183:
```suggestion
```

kivy/uix/modalview.py
```diff
@@ -261,56 +253,61 @@ def dismiss(self, *largs, **kwargs):
             self._anim_alpha = 0
             self._real_remove_widget()
 
-    def _align_center(self, *l):
-        if self._window:
-            self.center = self._window.center
+    def _align_center(self, *_args):
+        if self._is_open:
+            self.center = Window.center
```
matham, line 258:
```suggestion
            self.center = self._window.center
```

kivy/uix/modalview.py
```diff
@@ -70,19 +72,20 @@ def my_callback(instance):
...
     default).
 
-'''
+"""
 
 __all__ = ('ModalView', )
 
-from kivy.logger import Logger
 from kivy.animation import Animation
+from kivy.core.window import Window
```
matham, line 80:
```suggestion
```
I glossed over this before, but generally we try not to import `Window` at the module level because that causes a Window to be created upon import. Sometimes it's unavoidable and I know that importing kivy will already generally create the window, but it's better to reduce this for now until we can figure out how to resolve this issue of Kivy creating a Window when Window is imported.

kivy/uix/modalview.py
```diff
@@ -186,70 +197,51 @@ def __init__(self, **kwargs):
...
-        '''
-        if self._window is not None:
-            Logger.warning('ModalView: you can only open once.')
-            return
-        # search window
-        self._window = self._search_window()
-        if not self._window:
-            Logger.warning('ModalView: cannot open view, no window found.')
+        """
+        if self._is_open:
```
matham, line 209:
```suggestion
        """
        from kivy.core.window import Window
        if self._is_open:
```

kivy/uix/modalview.py
```diff
@@ -321,25 +318,19 @@ def _handle_keyboard(self, window, key, *largs):
     from kivy.uix.button import Button
```
matham, line 318:
```suggestion
    from kivy.uix.button import Button
    from kivy.core.window import Window
```

matham (changes requested):

matham (approved):

matham:
I hope you don't mind me merging the suggested changes, but I didn't want to make you wait for another review round. Thanks for your effort!

PR #7437, opened 2021-03-22
TextInput: Remove broken and confusing `suggestion_text` property

tshirtman:
Seems the code was never really functionnal for this case, making the
feature quite weird to use, as it would hide te whole text if the
suggested text was set to an empty string.
Also adapted the example to make it slightly more realistic

Maintainer merge checklist

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -3093,8 +3093,7 @@ class TextInput(FocusBehavior, Widget):
             lbl = MarkupLabel(
                 text=txt + "[b]{}[/b]".format(value), **kw)
         else:
-            lbl = Label(**kw)
-            text = txt
+            lbl = Label(text=txt, **kw)
 
         lbl.refresh()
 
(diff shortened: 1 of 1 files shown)


tshirtman:
Actually thinking more about it, this feature might have been a bad idea from the start, and the implementation is not good:
- it abuses Markup to hapend in bold at the end of the text, which means that if the text contains markup but we don't want to render it, then the feature will activate it anyway
-  it's also only able to display the suggestion at the end of the text, not at the cursor's position (which would require a relayout). Maybe these things are fixable, but it shows the feature:
- wasn't really though through
- wasn't used much by anyone, because the behavior before this PR makes the text completly disappears when `suggestion_text` is set to an empty string, which should certainly have bothered users.

matham:
Do you want to remove it altogether?

tshirtman:
I think it's worth discussing, i didn't try to make it fully workable, maybe it can be, but i'm not sure it currently provides much value, and it can be confusing as it's now aside the actual suggestion systems provided by android (that we have some support for again, and i wanted to look into in the first place).

matham:
I'd be for dropping this then.

matham (approved):

PR #7439, opened 2021-03-23
Properties: Drop long number type and document numpy issues with NumericProperty

matham:
Maintainer merge checklist

This resolves the issues discussed at https://github.com/kivy/kivy/pull/6984 and https://github.com/kivy/kivy/pull/7424#discussion_r599689345 and fixes https://github.com/kivy/kivy/issues/6982.

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -628,6 +628,25 @@ cdef class Property:
 cdef class NumericProperty(Property):
     '''Property that represents a numeric value.
 
+    It only accepts the int or float numeric data type or a string that can be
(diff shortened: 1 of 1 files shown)


tshirtman (approved):

pythonic64:
I would also add example about using `ObjectProperty`. Something like this:
```
import numbers
import numpy

a = ObjectProperty(baseclass=numbers.Number) # all numbers
b = ObjectProperty(baseclass=int) # only integers
c = ObjectProperty(baseclass=float) # only float
d = ObjectProperty(baseclass=numpy.float64) # only numpy.float64 numbers
```

PR #7445, opened 2021-03-24
EventDispatcher: Rename method unregister_event_types to unregister_event_type

pythonic64:
Maintainer merge checklist

### DIFF
--- a/kivy/_event.pyx
+++ b/kivy/_event.pyx
@@ -13,6 +13,7 @@ handlers.
 
 __all__ = ('EventDispatcher', 'ObjectWithUid', 'Observable')
 
+cimport cython
 from libc.stdlib cimport malloc, free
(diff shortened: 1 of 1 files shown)


matham:
I'm not sure this is worth code breakage, because it's not an egregious typo, although I can see it makes the API inconsistent.

If you think we should do this, a better approach is to make `unregister_event_types` call/an alias for `unregister_event_type` and deprecate `unregister_event_types` ([with](https://github.com/kivy/kivy/blob/master/kivy/utils.py#L318) `@deprecated(msg="deprecated in 2.1.0, use unregister_event_type instead. Will be removed after two releases.")`) and remove it eventually. And although I'm not suggesting we change it now, these really should have been named `register_event`/`unregister_event`/`is_event`.

matham (approved):

PR #7447, opened 2021-03-26
fix code quality issues

withshubh:
### Description
Hi :wave: I work at [DeepSource](https://deepsource.io), I ran DeepSource analysis on the forked copy of this repo and found some interesting [code quality issues](https://deepsource.io/gh/withshubh/kivy/issues/?category=recommended) in the codebase, opening this PR so you can assess if our platform is right and helpful for you.

### Summary of changes
- Remove unnecessary use of comprehension
- Remove length check in favour of truthiness of the object
- Remove unused global variable
- Remove methods with unnecessary super delegation.
- Iterate dictionary directly
- Add .deepsource.toml to continuously analyze the repo for code quality issues

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,15 @@
+version = 1
+
+test_patterns = ["kivy/tests/**/test_*.py"]
+
+exclude_patterns = [
+    "doc/**",
+    "examples/**"
+]
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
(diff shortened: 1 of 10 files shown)


matham:
Sorry, but I don't think we want to add config for another service to the repo, If we do want these kind of code checks we would do it using our existing git hook, but expand the things we check for.

Additionally, I don't think these changes are worth the risk to creating a bug, which admittedly is not very likely, but is still not worth it, especially when these changes don't really seem important. I can see removing the functions that just call super, but everything else doesn't seem an issue, just because one form is used over another. In new code maybe I'd suggest to call `if not x` rather than `if not len(x)`, but even there it's not a big deal.

PR #7449, opened 2021-03-27
TextInput: Fix readonly mode preventing using cursor keys

tshirtman:
Maintainer merge checklist

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -655,7 +655,7 @@ class TextInput(FocusBehavior, Widget):
             if count >= index:
                 return index - i, row
             i = count
-        return index, row
+        return int(index), int(row)
 
     def select_text(self, start, end):
         ''' Select a portion of text displayed in this TextInput.
@@ -1335,7 +1335,7 @@ class TextInput(FocusBehavior, Widget):
                 cursor_x = i
                 break
 
-        return cursor_x, cursor_y
+        return int(cursor_x), int(cursor_y)
 
     #
     # Selection control
@@ -1356,29 +1356,31 @@ class TextInput(FocusBehavior, Widget):
         if self.readonly:
             return
         self._hide_handles(EventLoop.window)
-        scrl_x = self.scroll_x
-        scrl_y = self.scroll_y
+        scroll_x = self.scroll_x
+        scroll_y = self.scroll_y
         cc, cr = self.cursor
         if not self._selection:
             return
         text = self.text
-        a, b = self._selection_from, self._selection_to
-        if a > b:
-            a, b = b, a
-        self.cursor = cursor = self.get_cursor_from_index(a)
-        start = cursor
-        finish = self.get_cursor_from_index(b)
-        cur_line = self._lines[start[1]][:start[0]] +\
-            self._lines[finish[1]][finish[0]:]
+        a, b = sorted((self._selection_from, self._selection_to))
+
+        self.cursor = (start_col, start_row) = self.get_cursor_from_index(a)
+        end_col, end_row = self.get_cursor_from_index(b)
+
+        cur_line = (
+            self._lines[start_row][:start_col]
+            + self._lines[end_row][end_col:]
+        )
         lines, lineflags = self._split_smart(cur_line)
-        len_lines = len(lines)
-        if start[1] == finish[1]:
-            self._set_line_text(start[1], cur_line)
+
+        if start_row == end_row:
+            self._set_line_text(start_row, cur_line)
         else:
-            self._refresh_text_from_property('del', start[1], finish[1], lines,
-                                             lineflags, len_lines)
-        self.scroll_x = scrl_x
-        self.scroll_y = scrl_y
+            self._refresh_text_from_property(
+                'del', start_row, end_row, lines, lineflags, len(lines)
+            )
+        self.scroll_x = scroll_x
+        self.scroll_y = scroll_y
         # handle undo and redo for delete selection
         self._set_unredo_delsel(a, b, text[a:b], from_undo)
         self.cancel_selection()
@@ -3424,9 +3426,11 @@ class TextInput(FocusBehavior, Widget):
     defaults to None.
     '''
 
-    base_direction = OptionProperty(None,
-                     options=['ltr', 'rtl', 'weak_rtl', 'weak_ltr', None],
-                     allownone=True)
+    base_direction = OptionProperty(
+        None,
+        options=['ltr', 'rtl', 'weak_rtl', 'weak_ltr', None],
+        allownone=True
+    )
     '''Base direction of text, this impacts horizontal alignment when
     :attr:`halign` is `auto` (the default). Available options are: None,
     "ltr" (left to right), "rtl" (right to left) plus "weak_ltr" and
@@ -3526,15 +3530,20 @@ class TextInput(FocusBehavior, Widget):
     '''
 
     def _get_min_height(self):
-        return (len(self._lines) * (self.line_height + self.line_spacing) +
-                self.padding[1] + self.padding[3])
+        return (
+            len(self._lines) * (self.line_height + self.line_spacing)
+            + self.padding[1]
+            + self.padding[3]
+        )
 
-    minimum_height = AliasProperty(_get_min_height,
-                                   bind=('_lines', 'line_spacing', 'padding',
-                                         'font_size', 'font_name', 'password',
-                                         'font_context', 'hint_text',
-                                         'line_height'),
-                                   cache=True)
+    minimum_height = AliasProperty(
+        _get_min_height,
+        bind=(
+            '_lines', 'line_spacing', 'padding', 'font_size', 'font_name',
+            'password', 'font_context', 'hint_text', 'line_height'
+        ),
+        cache=True
+    )
     '''Minimum height of the content inside the TextInput.
 
     .. versionadded:: 1.8.0
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -2518,13 +2518,15 @@ class TextInput(FocusBehavior, Widget):
         yield text[oldindex:]
 
     def _split_smart(self, text):
-        # Do a "smart" split. If autowidth or autosize is set,
-        # we are not doing smart split, just a split on line break.
-        # Otherwise, we are trying to split as soon as possible, to prevent
-        # overflow on the widget.
+        """
+        Do a "smart" split. If not multiline, or if wrap is set,
+        we are not doing smart split, just a split on line break.
+        Otherwise, we are trying to split as soon as possible, to prevent
+        overflow on the widget.
+        """
 
         # depend of the options, split the text on line, or word
-        if not self.multiline:
+        if not self.multiline or not self.do_wrap:
             lines = text.split(u'\n')
             lines_flags = [0] + [FL_IS_LINEBREAK] * (len(lines) - 1)
             return lines, lines_flags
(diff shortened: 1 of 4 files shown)


kivy/uix/textinput.py
```diff
@@ -2482,9 +2482,8 @@ def keyboard_on_key_down(self, window, keycode, text, modifiers):
                     self.copy()
             elif key == 27:
                 self.focus = False
-            return True
```
matham, line 2485: Wouldn't this now allow the things below to execute, such as tab-insertion and copy-paste, which is not supposed to happen in read-only mode?

XtremeGood: I actually asked for this thing on discord. My goal is to create a textInput of fixed size(say 10 rows). This would be the user friendly  logs window. Since logs can be 1000's of lines , the user should be able to scroll this textinput to see more lines and copy the text. Also they should be able to move inside the textInput using arrow keys. In my case I would actually need the text to readonly so that the users while moving through doesn't messes up the logs but should be able to copy them.

XtremeGood:
```

class InfoArea(TextInput):
    def insert_text(self, substring, from_undo=False):
        return True

    def keyboard_on_key_down(self, window, keycode, text, modifiers):
        if keycode[0] != 8:
            return super().keyboard_on_key_down(window, keycode, text, modifiers)
        return True



```

This is the code that I required.

tshirtman: @matham i did some manual testing after the change, and i couldn’t insert a tab, or paste using ctrl-v, maybe i should add some unittesting for good mesure, but it seemed fine to me.

matham: Hmm then that code is counterintuitive and it needs comments to explain what is going on and possibly a fix. Because upon a simple reading it looks like it should insert tabs, or that tab insertion line at the bottom and possibly the others doesn't do anything.

Also, when `_editable` is False, it could be because the widget is disabled or readonly and possibly other modes (keyboard related?). And I think we may want to interact with it only if readonly, but not when it's not editable for other reasons. And even so, I'm not sure people who set it to readonly always want it to be inter-actable? So it may need another option, but I'm not sure.

So you'll probably need instead something like:
```
if self.readonly and self.readonly_interactive:
    ...
    return True
elif not self._editable:
    ...
    return True
else:
    ...
```

Additionally, with your change, it seems like while you can interact it with partially, you can't fully interact with it. E.g. you can shift select, but ctrl-a doesn't work. So it'll probably need some code from under the `elif text and not is_interesting_key` condition extracted and moved under the `if self.readonly` condition. So it's clear what is happening and only specific to that condition.

matham:
Since you're working on textinput, I noticed when testing rtl with harfbuzz that the cursor appears in the wrong side of the text (on the right, rather than left). Feel free to ignore this though if it's unrelated to your changes. This would probably become hard quickly, especially since you can have rtl and ltr in the same text. But I guess rn we assume it just has one direction!?

tshirtman:
I'll see if i spot the issue, currently we assume one line has a direction, but it might be different from the lines before and after, and this assume that the provider can tell us the direction of a string of character (the base implementation always return 'rtl', and pango is the only core text provider to provide an alternative implementation, but maybe sdl2 could have its own override when harfbuzz is present.

matham:
Harfbuzz doesn't seem to have an API for getting the direction of text, so only pango can currently provide it. We'd need fribidi to implement it ourselves.

kivy/uix/textinput.py
tshirtman: Hopefuly the logic should be clearer now, i extracted handling of commands and shortcuts to specific methods, shortcut being used only at the top, and checking itself if text is editable or not, only allowing a subset of operations if text is not editable.
I also discovered a cute little bug with inserting tab while having a selection, it kept the selection alive, while other insertions did not, behavior has been changed to delete selection in this situation like any other insertion.

kivy/uix/textinput.py
```diff
@@ -2460,32 +2705,27 @@ def _key_up(self, key, repeat=False):
...
 
     def keyboard_on_key_down(self, window, keycode, text, modifiers):
-        # Keycodes on OS X:
-        ctrl, cmd = 64, 1024
-        key, key_str = keycode
+        key, _ = keycode
         win = EventLoop.window
 
         # This allows *either* ctrl *or* cmd, but not both.
         modifiers = set(modifiers) - {'capslock', 'numlock'}
```
matham, line 2712:
```suggestion
        modifiers = set(modifiers)
```

kivy/uix/textinput.py
```diff
@@ -2460,32 +2705,27 @@ def _key_up(self, key, repeat=False):
...
-        is_shortcut = (modifiers == {'ctrl'} or (
-            _is_osx and modifiers == {'meta'}))
-        is_interesting_key = key in (list(self.interesting_keys.keys()) + [27])
-
-        if not self.write_tab and super(TextInput,
-            self).keyboard_on_key_down(window, keycode, text, modifiers):
-            return True
+        is_shortcut = (
+            modifiers == {'ctrl'}
+            or _is_osx and modifiers == {'meta'}
```
matham, line 2715:
```suggestion
            'ctrl' in modifiers and 'meta' not in modifiers
            or _is_osx and 'ctrl' not in modifiers and 'meta' in modifiers
```
I think this is safer because numlock/capslock may not be the only modifiers active, but we can't know what they would be. But I don't feel very strongly about this.

kivy/uix/textinput.py
```diff
@@ -2571,6 +2770,7 @@ def keyboard_on_key_down(self, window, keycode, text, modifiers):
             self.focus = False
             return True
         elif key == 9:  # tab
+            self.delete_selection()
```
matham, line 2773: What puzzled me is that we insert tab, even if readonly. However, I guess `insert_text` returns early if readonly, but still it would be nice if it was guarded with a check for readonly. But it's not very important.

The other thing that puzzled me is that editable includes when it is disabled. But users may not want disabled text to be interactive, including copying. But, I guess when it's disabled all touch is disabled for the widget anyway.

Also, it was confusing to me that we explicitly allow ctrl-a when readonly from `_handle_shortcut`. But, what about up-arrow scrolling. But I guess that is handled in `_key_down` seperately.

matham (approved):
This is your revenge for the giant properties PR, isn't it 😅🤣

One thing I noticed, which is probably a old bug, but since you fixed wrapping you may want to look at it, but don't feel obligated:

* With a long line that wraps, if you place cursor at start of second line where it wraps and backspace, wrapping will break and it'll be a long line again
  * In the above situation, hitting home/end keys brings you to end of line but it's off screen, but right/left arrow does scroll to the right.

tshirtman:
I think we still have issues with cursor management indeed, that i would like to fix, regarding cursor management, i've seen some when do_wrap is False, and i'll look for the one you mention when it's True.

It's not a revenge PR, it's just that you said the code wasn't clear, and i very much agreed, so i tried to improve the situation, a bit all over the place (yes, it is revenge).

I'm not sure i can consider this done yet, i've also seen some glitches when backspacing on long lines, that would need to be addressed, hopefully i'll find the cause.

Thanks for the suggestions, i'll look at what they mean :).

kivy/uix/textinput.py
tshirtman: yes, there are various handling for various kind of typi typi, and it's not clear to me why we segregate "shortcuts", "interesting keys", and input, it makes *some* sense, but it's still a bit arbitrary.
To avoid testing in multiple places when functions are called, i did move the test inside the function, the performance cost seems negligible as users don't type thousands of keys per second, and the benefit is that you don't have to wonder if you fill all the conditions to call it, it's its responsability to early exit if not, but i guess it could be considered a smell, considering the complexity of this widget, i'd rather move downward when possible.

tshirtman: just to check i made a TextInput where disabled is based on time, so it "blinks" disabled, and even if it had focus before being disabled, it's not possible to type into it when it gets disabled, it also loses focus in that situation, which i'm not sure should be considered a bug or not.

matham:
Looks like all the tests are failing. Lemme know if you're at a point that needs review (it looks ok to me besides the `set` comments above, but I didn't test the recent changes).

tshirtman:
I had missed a part of the commit and forgot to check if the tests were passing >_>

kivy/uix/textinput.py
```diff
@@ -176,10 +176,12 @@ def insert_text(self, substring, from_undo=False):
 
 
 if 'KIVY_DOC' in environ:
-    def triggered(*_):
+    def triggered(*_, **__):
         def func(*args, **kwargs):
             return func(*args, **kwargs)
         return func
```
matham, line 182: I think the problem is that it is named `func` but both return `triggered`.

tshirtman: yeah, i was tired i guess, max recursion 😆

kivy/uix/textinput.py
```diff
@@ -173,6 +172,18 @@ def insert_text(self, substring, from_undo=False):
     BooleanProperty, AliasProperty, OptionProperty, \
     ListProperty, ObjectProperty, VariableListProperty, ColorProperty
 
+__all__ = ('TextInput', )
+
+
+if 'KIVY_DOC' in environ:
+    def triggered(*_, **__):
+        def func(*args, **kwargs):
```
tshirtman, line 180:
```suggestion
        def decorator_func(func):
            def decorated_func(*args, **kwargs):
                return func(*args, **kwargs)
            return decorated_func
        return decorator_func
```

kivy/uix/textinput.py
```diff
@@ -173,6 +172,18 @@ def insert_text(self, substring, from_undo=False):
...
+            return func(*args, **kwargs)
```
tshirtman, line 181:
```suggestion
```

kivy/uix/textinput.py
```diff
@@ -173,6 +172,18 @@ def insert_text(self, substring, from_undo=False):
...
+        return func
```
tshirtman, line 182:
```suggestion
```

tshirtman:
Ok, just the mouse test failing now.

matham (approved):

matham:
There's a few more things I noticed as I was playing with it, but you don't have to fix it in this PR if you don't feel like it, these have been there for a while probably.

1. Look at the top box as I change the window width: 
   image
2. With a long line that wraps, if you backspace before the first character on second line, the previous line stops wrapping.
3. With a line that wraps across multiple lines or multi-line text, if you use up-arrow, it doesn't keep its position horizontally based on the pixel number offset from left, but rather based on number of characters. So with non-uniform character font, it looks weird and is different than e.g. firefox.

tshirtman:
> 
> 
> There's a few more things I noticed as I was playing with it, but you don't have to fix it in this PR if you don't feel like it, these have been there for a while probably.

Yeah, this one is large enough :D

> 
>     1. Look at the top box as I change the window width:
>        image
> 
>     2. With a long line that wraps, if you backspace before the first character on second line, the previous line stops wrapping.
> 

Funnily enough, i was able to reproduce this once a few days ago, but not on subsequent attempts, will have to try again.

>     3. With a line that wraps across multiple lines or multi-line text, if you use up-arrow, it doesn't keep its position horizontally based on the pixel number offset from left, but rather based on number of characters. So with non-uniform character font, it looks weird and is different than e.g. firefox.

Nice catch, i guess we could use the get_cursor_from_xy at the new y pos and the old x pos and move to that new index, or maybe there is a way to only do one jump.

tshirtman:
Merged, i hope i didn’t break anything, pretty sure that did more good than harm though :D, at least it should be a tad easier to fix bugs in places the code was clarified.

Alspb:
There is a problem if do_wrap is False: long lines are represented as black rectangles.
E.g. TextInput('aaaaa'+'\n'+'b'*3000+'\n'+'ccccc') looks like:

Long lines bug

The reason is that the same holds true for Label (and textinput lines are in fact Labels). But for Label it's not a big issue (don't see any reason for using such long lines), while for TextInput it seems to be significant.

Issue #7149, opened 2020-10-14
Rendering Hindi(Indian) Language Wrong in Kivy for Android App 


kriyanation:
**Software Versions**
* Python:3.7
* OS:Android
* Kivy:1.11
* Kivy installation method: P4A

When I render the following text with the google NotoSansfont for Devanagari script in kivy android, I get it rendered wrongly. I have seen similar messages in SO for other languages and it says use pango text provider but it also mentions it is not available for android. Is there a possibility to render the font correctly in android in a kivy App. This will be a make or break for using kivy as the app development platform for us as we need to support Indian languages in our mobile app.

Example Text Expected output ->
![image](https://user-images.githubusercontent.com/2444073/96020641-f636ad00-0e6b-11eb-9a01-7561af2576f9.png)

Example Text Rendered Output -> 
![image](https://user-images.githubusercontent.com/2444073/96020606-e8812780-0e6b-11eb-988d-8fc79ec859b8.png)

This happens in all the controls - Label and TextInputs and we have tried different ttf files - NotoSans, Akshar, atleast 10 other font files.

The only font which works correctly is GNU Unifont for us but that is because it has all the glyphs within and the problem we have is it does not look good and readable although that is now our current and only option.

Can we get some help here please?

example code:
from kivy.app import App
from kivy.uix.image import Image
from kivy.uix.label import Label
from kivy.uix.boxlayout import  BoxLayout
from kivy.uix.button import Button

class MainApp(App):
    def build(self):
        label = Label(text="अंतरिक्ष", font_name="notosans.ttf", size_hint=(0.5,0.5),pos_hint={'center_x':.5,'center_y':.5})
        layout = BoxLayout(padding =10)
        layout.add_widget(label)
        return layout

if __name__=='__main__':
    app = MainApp()
    app.run()

Font file link : https://fonts.google.com/specimen/Noto+Sans?subset=devanagari

trivedigaurav:
Let me know if I understand it correctly, the notosans font has the 'क्ष' glyph but is not displayed when you use it for the kivy label? I will try looking at this issue if that is the case.

kriyanation:
[...]

Issue #2669, opened 2014-11-14
Issues with printing languages in unicode.


jsphonorio:
I am trying to print in Tamil and running into an issue where Kivy misprints the unicode language. Here is a short script and an Image showing the issue. To show comparison, here is same text print in Pyside and how it looks correctly.

```
  #!/usr/bin/python
  # -*- coding: utf-8 -*-

  from kivy.app import App
  from kivy.uix.button import Button

  class TestApp(App):
      def build(self):
          return Button(text='விகடகவி'.decode('utf-8'),font_name='Uni Ila.Sundaram-01.ttf',font_size=50)

  TestApp().run()
```

![kivy-example](https://cloud.githubusercontent.com/assets/9734945/5040671/a741f07a-6bf5-11e4-9a2b-2aca504b953e.png)
![pyside-example](https://cloud.githubusercontent.com/assets/9734945/5040672/abc1b8ce-6bf5-11e4-8f29-ffc73a3ab1a8.png)

The language combines letters together and sometimes rearranges consonants around vowels. Kivy has trouble parsing this.

These last two pictures just provide another example of a different phrase on a different computer. Same issue.

![kivy-example2](https://cloud.githubusercontent.com/assets/9734945/5040712/a4aac584-6bf6-11e4-9dc3-a4647d6809ee.png)

![pyside-example2](https://cloud.githubusercontent.com/assets/9734945/5040713/aa0fa27e-6bf6-11e4-81de-122b0c016aca.png)
##

pyprof:
I am facing same problem. But to write Bengali. Please be refereed to the screenshot to see how Kivy is breaking combined letters.

Here is the code.... https://bpaste.net/show/92402933377e
Here are the screenshots... 
- I am expecting something like this....
![a](https://cloud.githubusercontent.com/assets/12394079/7915229/0a6ac65a-089e-11e5-84f6-51d77de8c8d5.png)

- Getting something like this...
![screenshot from 2015-06-01 20 30 57](https://cloud.githubusercontent.com/assets/12394079/7915274/478eded6-089e-11e5-9b99-34e94fae05fa.png)

chintal:
As an additional data point, if it helps, vanilla pillow (PIL) Image, ImageDraw, and ImageFont have the same rendering problems as kivy. However, with libraqm also installed, pillow's renders are correct. Kivy's renders are not changed when libraqm is installed.

chintal:
[...]

PR #7452, opened 2021-03-28
Text: Add SDL_ttf + HarfBuzz support for better text language support

matham:
Maintainer merge checklist

This adds support for SDL2_ttf with HarfBuzz, now possible with the recent SDL_ttf changes: https://github.com/libsdl-org/SDL_ttf/commit/c71596421625e8a73d568e2fab27179c0fc41d5f. Fixes https://github.com/kivy/kivy/issues/7149.

There are a couple of API considerations. Both pango and harfbuzz share the direction and language property. For direction, `'ltr'`, and `'rtl'` are supported by HarfBuzz and Pango, while `'weak_ltr'`, and `'weak_rtl'` is Pango only. However, even for pango the weak is currently the same as strong direction. Similarly, for language, pango supports `RFC-3066`, which is quite broad, while harfbuzz supports the much stricter ISO 15924.

So the issue arises because they share the same property name, but the values can be slightly different. I don't see a way around this, besides creating duplicate properties, but that is even worse. So I think this will be fine, especially since I'm not sure the pango backend is used much.

To use this, you need a sdl_ttf version that has been compiled with `TTF_USE_HARFBUZZ=1` (this has to be `#define`, e.g. in configure) and harfbuzz present (because there's no sdl_ttf release with harfbuzz, you'll have to use their main branch). And then you have to compile kivy with `use_harfbuzz=1`. And finally you need to make sure the text provider is set to sdl2 (`KIVY_TEXT=sdl2`).

I expect it'll be a while before they make a release, so we'd need to update kivy-ios and py4a to use master and set the above variables. For OSX with hombrew and linux the same. OSX frameworks does not currently support it because there's no sdl+harfbuzz framework, if there will ever be. If they do release it with harfbuzz, we can rework the code to enable, For Windows, I already made a kivy_deps.sdl2 release with manually compiled stl_ttf/harfbuzz.

Here's a demo:
image

from the following code:
```py
from os import environ
environ['KIVY_TEXT'] = 'sdl2'

from kivy.app import App

class MainApp(App):
    def build(self):
        text = "Hello अंतरिक्ष!"
        font_name = r"G:\test\NotoSans-Regular.ttf"
        layout = GridLayout(cols=2)

        label = Label(text=text, font_name=font_name, font_size='25dp')
        layout.add_widget(label)

        label = Label(
            text=text, font_name=font_name,
            text_language='deva', font_size='25dp'
        )
        layout.add_widget(label)

        layout.add_widget(
            TextInput(text=text, font_name=font_name, font_size='25dp'))

        layout.add_widget(
            TextInput(
                text=text, font_name=font_name,
                text_language='deva', font_size='25dp'))
        return layout


if __name__=='__main__':
    app = MainApp()
    app.run()
```


kivy/core/text/_text_sdl2.pyx
```diff
@@ -220,3 +236,40 @@ def _get_fontdescent(container):
...
 def _get_fontascent(container):
     return TTF_FontAscent(_get_font(container))
+
+
+cdef _set_font_direction(str direction):
+    global last_base_direction
+    if direction == last_base_direction:
+        return
+
+    IF USE_HARFBUZZ:
```
tshirtman, line 246: all the calls being conditional, maybe this one is not necessary? Or maybe the whole method declaration could be conditional?

tshirtman (commented):
I'll try asap on linux, that's fantastic!

kivy/core/text/_text_sdl2.pyx
matham: Yeah, is is not necessary, however, it would be weird if a function is conditionally declared and just doesn't exist at all depending on a condition. And long term it we may call the function even if there's no harfbuzz if we don't want to guard the call under an if in multiple places. So it just safer and more consistent to always have the function defined.

matham:
Unfortunately, there seems to be an incompatibility between the CI compiled ttf/harfbuzz and gstreamer. I didn't test audio so I didn't see it. I guess we can merge it before we have pre-compiled windows wheels with harfbuzz and figure out later if we can compile it without issues for Windows.

matham:
I was able to resolve the binary incompatibility between harfbuzz/sdl_ttf compiled on CI and gstreamer by removing some unimportant harfbuzz sub-packages, but only for x64. So the most recent kivy_deps.sdl2 sdl_ttf only includes harfbuzz support in the x64 wheels and not the x86 wheels.

I suspect that gstreamer is including some mingw64 compiled dlls in their distribution, even though it's labeled as msvc. And that is causing the incompatibility.

tshirtman:
I tried but didn’t manage to build sdl2_ttf with harfbuzz on ubuntu yet, I’ll give it another shot later.

matham:
Hmm the only thing I can think of, is of you enable harfbuzz with the define, then it'll try to link with harfbuzz. But, SDL_TTF doesn't include harfbuzz.a/so in its list of libraries to link to so it'll probably complain about missing binaries.

So you'll need to add them to the link options I think. Otherwise I'm not sure sure what could be the issue!?

tshirtman:
It was kind of similar to that, but with headers, it didn’t find the `<hb.h>` despite it being installed in the prefix i gave to sdl2_ttf.

matham:
That's possibly because sdl just `#include <hb.h>`, but `hb.h` is probably under `harfbuzz/hb.h`. So you may need to add `#include/harfbuzz` to the include path.

tshirtman:
build on ubuntu in `~/.local/` to avoid changing the rest of the system.

step 0 install requirements
```
sudo apt-get install pkg-config gcc ragel gcovr gtk-doc-tools libfreetype6-dev libglib2.0-dev libcairo2-dev libicu-dev libgraphite2-dev python3 python3-setuptools ninja-build gobject-introspection libgirepository1.0-dev
pip install --user -U meson
```

step 1 clone the projects
```
git clone https://github.com/harfbuzz/harfbuzz
git clone https://github.com/libsdl-org/SDL
git clone https://github.com/libsdl-org/SDL_image
git clone https://github.com/libsdl-org/SDL_ttf
git clone https://github.com/libsdl-org/SDL_mixer
```

step 2 build harfbuzz
```
cd harfbuzz
meson build --auto-features=enabled -Doptimization=2 --prefix=$HOME/.local/
meson compile -C build
cd -
```

step 3 build SDL
```
cd SDL
./configure --prefix=$HOME/.local/
make
cd -
```

step 4 build SDL_ttf
```
cd SDL_ttf
export PKG_CONFIG_PATH="$HOME/.local/lib/x86_64-linux-gnu/pkgconfig"
env CFLAGS=$(pkg-config --cflags harfbuzz) LDFLAGS=$(pkg-config --libs harfbuzz) ./configure --prefix=$HOME/.local/
make install
cd -
```

step 4 build other SDL components
```
cd SDL_mixer
./configure --prefix=$HOME/.local/
make install
cd ../SDL_image
./configure --prefix=$HOME/.local/
make install
cd -
```

Build kivy with this SDL version.

```
env USE_HARFBUZZ=1 pip install --verbose --editable  .[dev]
```

but it still tries to use the system version, while the PKG_CONFIG_PATH is indeed in the environment, and even if i force setup.py to pass environ (instead of None) when calling pkg-config

```
    SDL2 found via pkg-config
    SDL2: found SDL header at /usr/include/SDL2/SDL.h
    SDL2: found SDL_mixer header at /usr/include/SDL2/SDL_mixer.h
    SDL2: found SDL_ttf header at /usr/include/SDL2/SDL_ttf.h
    SDL2: found SDL_image header at /usr/include/SDL2/SDL_image.h
    HarfBuzz: unable to find header
```
so a bit stuck for now.

tshirtman:
ok, some changes needed in setup.py, but it seem to work :)

capture-2021-04-03T00-21-08

```diff
--- a/setup.py
+++ b/setup.py
@@ -67,10 +67,14 @@ def pkgconfig(*packages, **kw):
 
     if isdir(pconfig):
         lenviron = environ.copy()
-        lenviron['PKG_CONFIG_PATH'] = '{};{}'.format(
-            environ.get('PKG_CONFIG_PATH', ''), pconfig)
+        lenviron['PKG_CONFIG_PATH'] = '{}:{}'.format(
+            environ.get('PKG_CONFIG_PATH', ''),
+            pconfig
+        )
+
     cmd = 'pkg-config --libs --cflags {}'.format(' '.join(packages))
-    results = getoutput(cmd, lenviron).split()
+    print(cmd)
+    results = getoutput(cmd, lenviron or environ).split()
     for token in results:
         ext = token[:2].decode('utf-8')
         flag = flag_map.get(ext)
@@ -555,8 +559,9 @@ if c_options['use_sdl2'] and c_options['use_harfbuzz'] is None \
     # 2.0.15 is last version without harfbuzz API. If sdl2 dist is compiled
     # without it it's not an issue for us as it'll just error in runtime
     harf_flags = pkgconfig('harfbuzz')
-    if 'libraries' in pkgconfig('SDL2_ttf > 2.0.15') and \
-            'libraries' in harf_flags:
+    pkg_sdl = pkgconfig('SDL2_ttf > 2.0.15')
+
+    if 'libraries' in pkg_sdl and  'libraries' in harf_flags:
         print('HarfBuzz found via pkg-config')
         c_options['use_harfbuzz'] = True
         sdl2_flags = merge(sdl2_flags, harf_flags)
@@ -739,6 +744,9 @@ def determine_sdl2():
             if isdir(harf_inc):
                 sdl2_paths.append(harf_inc)
 
+        sdl2_config = pkgconfig('sdl2', 'SDL2_ttf', 'SDL2_mixer', 'SDL2_image')
+        harfbuzz_config = pkgconfig('harfbuzz')
+
         for d in ['/usr/local/include/SDL2', '/usr/include/SDL2',
                   '/usr/local/include/harfbuzz', '/usr/include/harfbuzz']:
             if isdir(d):
@@ -902,7 +910,17 @@ sources = {
 }
 
 if c_options["use_sdl2"]:
-    sdl2_flags = determine_sdl2()
+    sdl2_flags = pkgconfig('sdl2', 'SDL2_ttf', 'SDL2_mixer', 'SDL2_image')
+    if sdl2_flags:
+        print("sdl2 found via pkgconfig: %s", sdl2_flags)
+        if c_options['use_harfbuzz']:
+            harf_flags = pkgconfig('harfbuzz')
+            if harf_flags:
+                print("harfbuzz found via pkgconfig: %s", harf_flags)
+                sdl2_flags = merge(sdl2_flags, harf_flags)
+
+    if not sdl2_flags:
+        sdl2_flags = determine_sdl2()
```

matham:
Hmm, the picture you're showing doesn't seem to work. The text should look differently on the right if harfbuzz is used.

I think you need to add `-DTTF_USE_HARFBUZZ=1` to sdl_ttf configure. Because otherwise sdl won't use it/link with it.

I pushed changes incorporating some of your changes, hopefully that'll fix it for you (that sdl2 discovery code is too complicated 🤦‍♀️)

tshirtman:
ah, forgot to put it in the procedure, but i did set `TTF_USE_HARFBUZZ=1` in `SDL_ttf.c`
```
gabriel@king:~/d/SDL_ttf> grep "TTF_USE_HARFBUZZ" SDL_ttf.c 
#ifndef TTF_USE_HARFBUZZ
#  define TTF_USE_HARFBUZZ 1
```
and indeed, I see now the result is not as expected, i'll check again if i missed something.

matham:
Try not removing harfbuzz from the paths and see if sdl_ttf complains. It should refuse to compile if you enabled it but harfbuzz is not present.

matham:
The test failure looks to be due to a failed mouse test unrelated to these changes.

tshirtman:
hm, i edited ld config file to add the paths to my local libs, and i can see they are loaded from there by running through strace
```
…
openat(AT_FDCWD, "/home/gabriel/.local/lib/libSDL2_ttf-2.0.so.0", O_RDONLY|O_CLOEXEC) = 3
…
openat(AT_FDCWD, "/home/gabriel/.local/lib/x86_64-linux-gnu/libharfbuzz.so.0", O_RDONLY|O_CLOEXEC) = 3
…
```


but the result still looks the same :|

image

matham:
Huh. Can you try the c code posted in the issue? That skips kivy so  it would be clearer if kivy is the issue. Or maybe try docker with forwarded graphics and install to the standard places overwriting apt SDL, rather than to local!?

On Sun, Apr 4, 2021, at 11:11 AM, Gabriel Pettier wrote:
> 

> hm, i edited ld config file to add the paths to my local libs, and i can see they are loaded from there by running through strace

> `…
openat(AT_FDCWD, "/home/gabriel/.local/lib/libSDL2_ttf-2.0.so.0", O_RDONLY|O_CLOEXEC) = 3
…
openat(AT_FDCWD, "/home/gabriel/.local/lib/x86_64-linux-gnu/libharfbuzz.so.0", O_RDONLY|O_CLOEXEC) = 3
…
`

matham:
So, shall we merge it?

chintal:
> So the issue arises because they share the same property name, but the values can be slightly different. I don't see a way around this, besides creating duplicate properties, but that is even worse. So I think this will be fine, especially since I'm not sure the pango backend is used much.

The pango backend is the only way at present to display Indian languages on Kivy. The only other 'option' is to externally render to an image and use that. Therefore, given that any kivy code that does display other languages is probably already using pango.  I think assuming that the pango backend is not used much is a mistake.

As such, please ensure that any change you make does not break existing code that relies on the pango backend. If the issue is only that the effect of the overlapping properties may be different depending on which backend is enabled, that should be alright. And that has absolutely nothing to do with how often the pango backend is used or not.

@tshirtman : the rendering shown in the screenshots you have are incorrect. 

@matham @tshirtman : I did write up a quick script + kivy GUI to test renderers across multiple Indian languages some time ago. You may consider reusing it. ( See https://github.com/kivy/kivy/issues/2669#issuecomment-526899628 )

matham:
@chintal thanks for your feedback.

You are right, but I was not considering breaking pango. The main point I was making is that harfbuzz and pango re-use the same properties for `direction` and `language`, however, the meaning of the values are slightly different. So an app written with pango in mind using a `language` property value from `RFC-3066` will not work when using harfbuzz because it's language is restricted to `ISO 15924`.

So basically, it is not quite portable between the two backends, but each backend works fine and we won't break either. My point was that the slightly different API meaning is not a problem because pango has limited support in terms of OS, so in the future, while we continue to support pango, I expect users will be happy using just sdl2+harfbuzz as the dominant API for multi-lang support because it's the path of least resistance. So most users could just ignore the pango semantics and remember/use the harfbuzz semantics. But ultimately, even if users use both pango and harfbuzz equally in different apps, re-using the same API, but with backend specific meaning is still a better choice than introducing `direction` and `language` properties specific to each backend. That was all I was saying.

I don't have a easy way to test pango. Would you consider testing this PR with your pango app using the pango backend? And maybe test also harfbuzz using your script, since I'm not sure I could easly parse the output of the script to understand whether the rendering is correct?

chintal:
@matham 

I can confirm that code at the head of the matham:harfbuzz branch renders indic fonts with the pango backend as before. 

Simply changing the text provider to 'sdl2' breaks the rendering. I will have to look into what is necessary to install harfbuzz, but I suppose it should be possible in principle (ubuntu). Are there specific versions of SDL2/harfbuzz required, or will a relatively recent one coming from the package manager do?

As an aside, as long as you're able to run indic_font_test.py and get the kivy window to open up, you should see the same string rendered by the various methods I was examining at the time. I know it is difficult to do, but one could compare renderings across the columns and see if it is accurate. The comment I linked to earlier also has a screenshot, in case there are missing fonts or so and the other backends don't work either. If they do work, though, 

I find that pyvips provides for the most reliable rendering with my (mostly default) set of system fonts on ubuntu. 

To the extent possible I have got the renderings reviewed by native speakers, those that I could not are marked with an asterix. 
I have since got a couple of others reviewed as well since, so on the whole I have fairly high confidence in the pyvips rendering and the kivy(pango) rendering when using the "Sakalbharati' font (non-commercial use license), a copy of which i have included in the repository and should not need installation.

matham:
I was able to create a kivy manylinux wheel that contains harfbuzz. Get it from the zip file named [manylinux_wheels ](https://github.com/kivy/kivy/actions/runs/803388017).

I have been unable to test it because I don't have linux within easy reach so I'm not sure if it works, but you're welcome to try it.

tshirtman:
Nice job! That i can report as working!
image
Thanks for making it so easy :D

VICTORVICKIE:
Will this be rendered in Android too?

tshirtman:
> Will this be rendered in Android too?

It might require updating the sdl2 recipe in p4a.

chintal:
I have tried the manylinux wheel with python 3.8.5. Unfortunately, I can't say it's working. The rendering has changed from the earlier SDL2 rendering, but it still is not correct. If you require version information or some other tests here I can do them and provide the results. 

The wheel was installed by itself into a fresh python 3.8.5 virtual environment. Additionally, I installed pillow, pyvips, and wand into that virtualenv. No other packages were explicitly installed.

I've stitched together the results from the various kivy versions being discussed using gimp.

render-tests

First column : kivy + harfbuzz manylinux wheel with 3.8.5, text provider = sdl2
Second column : kivy from matham:harfbuzz, text provider = sdl2 (no harfbuzz or latest sdl2 installed)
Third column : kivy from matham:harfbuzz, text provider = pango
Fourth column : pyvips rendering

The pyvips rendering is the most readable and is correct. The pango rendering is correct, but is slightly less readable because of the font size being a little small.

The sdl2 renderings for Hindi and Telugu show overlapping glyphs which should not be overlapping, extra marks which should have been hidden, and misplaced glyphs. These should be easy to see when you compare them to the pyvips / pango renderings. The other languages have some variations as well.

matham:
I just tried it on Windows and this is what I got:

image


Notice that the first one now looks correct. The difference is that I added the missing language string which you don't have in your demo: `text_language='deva'`.

The other rows don't look correct, but that's probably because the language strng of deva is incorrect for them. You'd need to know the correct one to set for each language. Unfortunately, harfbuzz doesn't include code to automatically figure out the language, while I think pango does.

matham:
>It might require updating the sdl2 recipe in p4a.

You'll need to:
1. Add harfbuzz recipe,
2. Modify sdl2_ttf to use the main branch and sed to enable building with harfbuzz and set the correct harfbuzz flags. My manylinux building changes in the PR can be used as a guide (partially copied from @tshirtman code above :D ).
3. Update Kivy recipe to set `USE_HARFBUZZ=1` and update the `KIVY_SDL2_PATH` by adding the directory path to the `hb.h` file so we can find it.

The same would go for ios.

tshirtman:
there is an harfbuzz recipe for p4a already, i don't know if changes are needed in it
https://github.com/kivy/python-for-android/blob/develop/pythonforandroid/recipes/harfbuzz/__init__.py

happy documenting my tentative local build was not in vain :D

matham:
The harfbuzz recipe can be updated to 2.8.0 from [github](https://github.com/harfbuzz/harfbuzz/releases), but otherwise should be good (maybe we need to set the build type to release? But it's probably the default with configure).

chintal:
@matham You are correct. It seems the text_language is required for rendering to work correctly. I expect 'deva' stands for devanagri, so that would probably work for Hindi, Marathi, and maybe Punjabi in that set of languages. 

This is unfortunate. Also somewhat strange that this should be required. That said, assuming this requirement is coming from harfbuzz through sdl2, the kivy bits of it are likely working. 

I have updated the github repository with the test script to reflect these changes. For future reference and for those who might come looking later on the  text_language seems to be those defined in https://docs.microsoft.com/en-us/typography/opentype/spec/scripttags and for the indian languages I have tested, the following text_language work : 

```
_language_strings = {
    'English': (u'English Keyboard', None),
    'Hindi': (u'हिंदी कीबोर्ड ', 'deva'),
    'Telugu': (u'తెలుగులో టైప్', 'telu'),
    'Kannada': (u'ಕನ್ನಡ ಕೀಲಿಮಣೆ', 'knda'),
    'Bengali': (u'বাংলা কিবোর্ড', 'beng'),
    'Marathi': (u'मराठी कळफलक', 'deva'),
    'Punjabi*': (u'ਪੰਜਾਬੀ ਦੇ ਬੋਰਡ', 'deva'),           # Source text unverified
    'Tamil*': (u'தமிழ் விசைப்பலகை', 'taml'),     # Source text unverified
    'Urdu': (u'اردوبورڈ', 'arab'),
    'Malyalam*': (u'മലയാളം കീബോര്‍ഡ് ', 'mlym'),  # Source text unverified
    'Oriya*': (u'ଉତ୍କଳଲିପି', 'orya'),                  # Source text unverified
}
```

matham:
Thanks for testing!

We don't do anything special for pango to guess the text language other than using the locale default, so I'm not sure how it figures it out automatically. But pango is a higher level library so it makes sense that it does it. With harfbuzz, from what I read, we'd need to do the guessing ourselves using ICU and then tell harfbuzz the right language. But that's too much additional work for me to do.

VICTORVICKIE:
@matham i can verify that Tamil is getting rendered correctly AFAICT from the given example.

chintal:
@matham On closer inspection, it seems the Urdu / Arabic text rendering is incorrect. It seems to be an issue with the text direction (rtl / ltr). I don't know if there is an additional switch that is needed to make it work correctly.

image

The font I am using itself is fine, and the text renders correctly with Kivy+Pango, Pillow+raqm and pyvips. The direction is incorrect with Kivy+SDL2+Harfbuzz and Wand.

matham:
There is a direction field which I did not set, so that makes sense it would be reversed. But I can't readily test on my computer at the moment to see if setting the direction field fixes it.

matham:
The newest Windows sdl2-ttf release does include harfbuzz, however, it statically included it in the dll. This means that some of the harfbuzz functions used in the PR won't be available, unless we copy the code from harfbuzz. It's not a lot of code, but it's not great. The alternative is to manually compile sdl2-ttf, which is also not great long-term.

damascene:
@matham Hey, I'm interested in testing this PR on Ubuntu Linux. Would appreciate if you couldse share instruction for testing. My system already has this version:
`ii  libsdl2-2.0-0:amd64                    2.0.10+dfsg1-3                        amd64        Simple DirectMedia Layer`

I'm looking forward to see RTL and Arabic support in Kivy.

damascene:
@tshirtman 
I've followed your instruction but hit this error:
```
SDL_ttf.c:28:10: fatal error: ft2build.h: No such file or directory
   28 | #include <ft2build.h>
      |          ^~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:1534: libSDL2_ttf_la-SDL_ttf.lo] Error 1
```

while running `SDL_ttf$ make install` 

I've looked for a solution but it seems like I've all the packages required.

tshirtman:
> @tshirtman 
> I've followed your instruction but hit this error:
> ```
> SDL_ttf.c:28:10: fatal error: ft2build.h: No such file or directory
>    28 | #include <ft2build.h>
>       |          ^~~~~~~~~~~~
> compilation terminated.
> make: *** [Makefile:1534: libSDL2_ttf_la-SDL_ttf.lo] Error 1
> ```
> 
> while running `SDL_ttf$ make install` 
> 
> I've looked for a solution but it seems like I've all the packages required. 

You are probably missing a freetype 2 dev package. Or maybe the one you have is too old.

damascene:
> You are probably missing a freetype 2 dev package. Or maybe the one you have is too old.

I installed a fresh Ubuntu daily build distro to test this. I still get the same error. 
```
:~/SDL_ttf$ pkg-config --cflags freetype2
-I/usr/include/freetype2 -I/usr/include/libpng16
```
Which supposed to be ok.

misl6:
Long story short:
- I'm upgrading to latest SDL2 release as I need a feature of SDL 2.24.0
- I took the occasion to also upgrade SDL_ttf, SDL_image and SDL_mixer
- I've seen in release notes of SDL_ttf, that a couple of interesting things that may help us to complete this PR have been added (may avoid us to include harfbuzz functions and definitions which are not exposed).

See: https://github.com/libsdl-org/SDL_ttf/blob/fc8ff6cbdc332ba5ad150d0fedfc7be94c448339/SDL_ttf.h#L2287 and https://github.com/libsdl-org/SDL_ttf/blob/fc8ff6cbdc332ba5ad150d0fedfc7be94c448339/SDL_ttf.h#L2306

Also, `harfbuzz` is now built by default, and we don't even need to build it separately.

@matham any thoughts?

The PR regarding the SDL2 update should land in a couple of hours.

misl6:
> @misl6
> 
> Sadly, the functions we need from harfbuzz, `hb_direction_from_string` and `hb_script_from_string` are not included in the header. And on Windows at least, harfbuzz is statically included in the sdl dll, so there's no harfbuzz dll to get it from. We'd need to separately build harfbuzz or copy the relevant harfbuzz code still, to use it.

Yep, but the newly added `TTF_SetFontDirection` and `TTF_SetFontScriptName` may help to avoid to separately build harfbuzz or copy the relevant harfbuzz code, as SDL now does the needed conversions to "harfbuzz-enums" under the hood.

`TTF_SetDirection` is now deprecated in favor of `TTF_SetFontDirection` which takes the following direction values ( instead of an harfbuzz direction ):

 *  `TTF_DIRECTION_LTR` (Left to Right)
 *  `TTF_DIRECTION_RTL` (Right to Left)
 *  `TTF_DIRECTION_TTB` (Top to Bottom)
 *  `TTF_DIRECTION_BTT` (Bottom to Top)
 
 `TTF_SetScript` is now deprecated in favor of `TTF_SetFontScriptName ` which creates (and uses) an `HB_TAG`  from a 4-char null-terminated string.

misl6:
Closing as the feature has been introduced via https://github.com/kivy/kivy/pull/8115 (and now GA via Kivy 2.2.0)

PR #7461, opened 2021-04-01
AsyncImageTestCase: Fix for test_reload_asyncimage method and cleanup

pythonic64:
Fixes sync bug in `test_reload_asyncimage` which can cause this test to fail.

```
__________________ AsyncImageTestCase.test_reload_asyncimage __________________

self = <kivy.tests.test_uix_asyncimage.AsyncImageTestCase testMethod=test_reload_asyncimage>

    def test_reload_asyncimage(self):
        from tempfile import mkdtemp
        from os import remove
        from shutil import copyfile, rmtree
    
        fn = resource_find('data/logo/kivy-icon-16.png')
        t = mkdtemp()
        source = join(t, 'source.png')
        copyfile(fn, source)
        image = AsyncImage(source=source)
        self.render(image, framecount=2)
>       self.assertEqual(image.texture_size, [16, 16])
E       AssertionError: [32, 32] != [16, 16]

kivy\tests\test_uix_asyncimage.py:125: AssertionError
```

Maintainer merge checklist

### DIFF
--- a/kivy/tests/common.py
+++ b/kivy/tests/common.py
@@ -12,7 +12,7 @@ The screenshots live in the 'kivy/tests/results' folder and are in PNG format,
 
 __all__ = (
     'GraphicUnitTest', 'UnitTestTouch', 'UTMotionEvent', 'async_run',
-    'requires_graphics')
+    'requires_graphics', 'ensure_web_server')
 
 import unittest
 import logging
--- a/kivy/tests/test_uix_asyncimage.py
+++ b/kivy/tests/test_uix_asyncimage.py
@@ -1,25 +1,26 @@
-from kivy.tests.common import GraphicUnitTest, ensure_web_server
-
-from kivy.uix.image import AsyncImage
-from kivy.config import Config
-from kivy import kivy_examples_dir
-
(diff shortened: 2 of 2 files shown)


matham (dismissed):
I see, so the problem was that it wasn't waiting?

BTW, once approved, you should have permissions to merge the PR yourself.

pythonic64:
Yes, and because of it `image.texture_size` is `(32, 32)` which is the size of `image-loading.zip`.

I assume that "Squash and merge" is the default merge option for Kivy pull requests?
When "Create a merge commit" should be used?

matham:
I tend to squash only when there's more than one commit in the PR, except if the individual commits are meaningful and then I leave them (which is rare, e.g. in my recent properties PR there were multiple commits, each meaningful because they were building on each other so preserving them individually could later help debugging).

If there's only one commit, I tend to merge rather than squash, but I'm not sure there's really a preference. I suppose that squashing helps see the changes more clearly in the log as opposed to merging where the changes could have been committed further back in time so it wouldn't be as clear when it got into master 🤷‍♀️

matham (approved):

PR #7467, opened 2021-04-05
Text: Raise when registering a font_regular with None

tshirtman:
Currently, if the user tries to pass None for font_regular, then using it as fallback for other types won't work, raising an IndexError, of which the cause might not be clear to users.
Preemptively raising in this case seems better.

Maintainer merge checklist

### DIFF
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -305,6 +305,9 @@ class LabelBase(object):
         fn_regular will be used instead.
         '''
 
+        if font_regular is None:
+            raise valueError("font_regular cannot be None")
+
         fonts = []
 
         for font_type in fn_regular, fn_italic, fn_bold, fn_bolditalic:
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -306,7 +306,7 @@ class LabelBase(object):
         '''
 
         if font_regular is None:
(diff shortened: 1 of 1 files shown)


kivy/core/text/__init__.py
```diff
@@ -305,6 +305,9 @@ def register(name, fn_regular, fn_italic=None, fn_bold=None,
         fn_regular will be used instead.
         '''
 
+        if font_regular is None:
+            raise valueError("font_regular cannot be None")
```
tshirtman, line 309:
```suggestion
            raise ValueError("font_regular cannot be None")
```

kivy/core/text/__init__.py
```diff
@@ -305,6 +305,9 @@ def register(name, fn_regular, fn_italic=None, fn_bold=None,
         fn_regular will be used instead.
         '''
 
+        if font_regular is None:
```
tshirtman, line 308:
```suggestion
        if fn_regular is None:
```

matham (commented):
Is it not possible to have a font which is only bolded/italicized?

tshirtman:
Maybe, but in that case i don't think we support that use case.

matham (approved):

tshirtman:
seems that mouse test is flacky…

matham:
Yep https://github.com/kivy/kivy/pull/7468.

Issue #7470, opened 2021-04-07
Proposal: Change Video Widget to support Placeholder Image


rnixx:
**Is your feature request related to a problem? Please describe.**
When loading videos, a white background is displayed and an error message gets logged that the image cannot be loaded (source is set to the video path actually)

**Describe the solution you'd like**
I want to define a placeholder image if the video is not loaded yet or stopped. This can be then a single frame from the video, some icon or simply a transparent image.

**Describe alternatives you've considered**
As workaround i subclassed ``kivy.uix.video.Video``, introducing a ``video_source`` property, leaving the ``source`` property to my desired placeholder image:

```python
from kivy.uix import video
from kivy.properties import StringProperty
from kivy.resources import resource_find

class Video(video.Video):
    video_source = StringProperty('')

    def __init__(self, **kwargs):
        self._video = None
        super(Video, self).__init__(**kwargs)
        self.fbind('video_source', self._trigger_video_load)

        if "eos" in kwargs:
            self.options["eos"] = kwargs["eos"]
        if self.video_source:
            self._trigger_video_load()

    def _do_video_load(self, *largs):
        if CoreVideo is None:
            return
        self.unload()
        if not self.video_source:
            self._video = None
            self.texture = None
        else:
            filename = self.video_source
            # Check if filename is not url
            if '://' not in filename:
                filename = resource_find(filename)
            self._video = CoreVideo(filename=filename, **self.options)
            self._video.volume = self.volume
            self._video.bind(on_load=self._on_load,
                             on_frame=self._on_video_frame,
                             on_eos=self._on_eos)
            if self.state == 'play' or self.play:
                self._video.play()
            self.duration = 1.
            self.position = 0.

v = Video(source='blank.png', video_source='my_video.mp4')
```

**Additional context**
Implementing this as in the workaround above would be an API break. It would be good if this can be avoided. Therefor we would need to "reroute" some properties, e.g. internally set ``source`` to ``video_source`` and introduce ``placeholder`` which gets mapped to ``source`` before super is called in ``__init__``. Further we would need to add event handlers or alias attributes or similar for this property routing to keep current behavior sane when setting properties from outside. I tested the workaround above only with ffpyplayer video provider, so there might be more implications with concrete core video providers than stated.

in the end, for the API user there should be only one additional property ``placeholder``

```python
v = Video(source=my_video.mp4', placeholder='blank.png')
```

Any feedback or ideas are highly appreciated.

tshirtman:
[...]

PR #7471, opened 2021-04-08
Add support for preview image to Video

rnixx:
Maintainer merge checklist

Here's the PR to https://github.com/kivy/kivy/issues/7470

The behavior is as follows:

- If ``Video`` gets created with ``source`` argument, a deprecation warning is logged that this usage is deprecated, and passed ``source`` value is set as ``video_source``. Same applies if setting ``source`` property on instance.
- If ``Video`` gets created with ``video_source`` argument only, the underlying ``Image.source`` keeps untouched, thus stays ``None``. This way we avoid the annoying Image not found log message.
- If ``Video`` gets created with ``video_source``  and ``preview_source`` arguments, latter is passed as ``source`` to superclass and gets displayed as texture until video is loaded/started.

### DIFF
--- a/kivy/uix/video.py
+++ b/kivy/uix/video.py
@@ -56,9 +56,9 @@ class Video(Image):
     defaults to None.
 
     .. versionchanged:: 2.1.0
-        While the use of ``source`` is the historical way for defining
+        While the use of :attr:`source` is the historical way for defining
         video source and stays supported for backward compatibility, it is
-        recommended to use ``video_source`` instead.
+        recommended to use :attr:`video_source` instead.
     '''
 
     video_source = StringProperty(None, allownone=True)
@@ -74,7 +74,7 @@ class Video(Image):
     '''Filename / source of a preview image displayed before video starts.
 
     :attr:`preview_source` is a :class:`~kivy.properties.StringProperty` and
-    defaults to None. If None/empty string and :attr:`source ` is a
+    defaults to None. If None/empty string and :attr:`source` is a
     filename, that will be used for the preview filename.
     Otherwise, :attr:`source` will be set to :attr:`preview_source`.
 
@@ -180,7 +180,7 @@ class Video(Image):
         self._video = None
         super(Video, self).__init__(**kwargs)
         self.fbind('video_source', self._trigger_video_load)
-        self.fbind('preview_source', self._bc_video_load)
+        self.fbind('preview_source', self._update_source_from_preview)
         if self.preview_source:
             self.source = self.preview_source
 
@@ -213,7 +213,7 @@ class Video(Image):
             raise Exception('Video not loaded.')
         self._video.seek(percent, precise=precise)
 
-    def _bc_video_load(self, inst, val):
+    def _update_source_from_preview(self, inst, val):
         self.source = self.preview_source
 
     def _trigger_video_load(self, *largs):
--- a/kivy/uix/video.py
+++ b/kivy/uix/video.py
@@ -214,7 +214,7 @@ class Video(Image):
         self._video.seek(percent, precise=precise)
 
     def _bc_video_load(self, inst, val):
-        self.video_source = val
+        self.source = self.preview_source
 
     def _trigger_video_load(self, *largs):
         ev = self._video_load_event
--- a/kivy/uix/video.py
+++ b/kivy/uix/video.py
@@ -42,7 +42,7 @@ from kivy.uix.image import Image
 from kivy.core.video import Video as CoreVideo
 from kivy.resources import resource_find
 from kivy.properties import (BooleanProperty, NumericProperty, ObjectProperty,
-                             OptionProperty)
+                             OptionProperty, StringProperty)
 from kivy.logger import Logger
 
 
--- a/kivy/uix/video.py
+++ b/kivy/uix/video.py
@@ -173,11 +173,8 @@ class Video(Image):
     _video_load_event = None
 
     def __init__(self, **kwargs):
-        # B/C case source is given. use it as video source
-        if 'source' in kwargs:
-            kwargs['video_source'] = kwargs['source']
         # Case preview is given
-        if 'preview_source' in kwargs:
+        if kwargs.get('preview_source'):
             kwargs['source'] = kwargs['preview_source']
 
         self._video = None
--- a/kivy/uix/video.py
+++ b/kivy/uix/video.py
@@ -43,7 +43,6 @@ from kivy.core.video import Video as CoreVideo
 from kivy.resources import resource_find
 from kivy.properties import (BooleanProperty, NumericProperty, ObjectProperty,
                              OptionProperty, StringProperty)
-from kivy.logger import Logger
 
 
 class Video(Image):
@@ -57,8 +56,9 @@ class Video(Image):
     defaults to None.
 
     .. versionchanged:: 2.1.0
-        The use of ``source`` property is deprecated. Use ``video_source``
-        instead.
+        While the use of ``source`` is the historical way for defining
+        video source and stays supported for backward compatibility, it is
+        recommended to use ``video_source`` instead.
     '''
 
     video_source = StringProperty(None)
@@ -173,10 +173,6 @@ class Video(Image):
     def __init__(self, **kwargs):
         # B/C case source is given. use it as video source
         if 'source' in kwargs:
-            Logger.warning(
-                'The use of ``source`` property is deprecated. '
-                'Please use ``video_source`` instead.'
-            )
             kwargs['video_source'] = kwargs['source']
         # Case preview is given
         if 'preview_source' in kwargs:
@@ -217,10 +213,6 @@ class Video(Image):
         self._video.seek(percent, precise=precise)
 
     def _bc_video_load(self, inst, val):
-        Logger.warning(
-            'The use of ``source`` property is deprecated. '
-            'Please use ``video_source`` instead.'
-        )
         self.video_source = val
 
     def _trigger_video_load(self, *largs):
--- a/kivy/uix/video.py
+++ b/kivy/uix/video.py
@@ -16,18 +16,22 @@ the video is loaded (when the texture is created)::
 
     def on_position_change(instance, value):
         print('The position in the video is', value)
+
     def on_duration_change(instance, value):
         print('The duration of the video is', value)
+
     video = Video(video_source='PandaSneezes.avi')
-    video.bind(position=on_position_change,
-               duration=on_duration_change)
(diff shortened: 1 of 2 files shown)


kivy/uix/video.py
```diff
@@ -31,21 +42,51 @@ def on_duration_change(instance, value):
...
-                             OptionProperty)
+                             OptionProperty, StringProperty)
+from kivy.logger import Logger
 
 
 class Video(Image):
     '''Video class. See module documentation for more information.
     '''
 
+    source = StringProperty(None)
```
matham, line 53: I think this should default to `""`, rather than None. This is supposed to be strings so there's no reason to use `None` as signal to mean not set.

kivy/uix/video.py
```diff
@@ -31,21 +42,51 @@ def on_duration_change(instance, value):
...
+    '''Filename / source of your video.
+
+    :attr:`source` is a :class:`~kivy.properties.StringProperty` and
+    defaults to None.
+
+    .. versionchanged:: 2.1.0
+        The use of ``source`` property is deprecated. Use ``video_source``
```
matham, line 60: I don't think we should deprecate `source`. I can't see a real reason why using source as the default for both video and image should be removed eventually and break existing programs. Especially considering we won't be able to actually remove `source` because it's part of Image so at best we could say to not use it, but it's simpler to continue supporting it.

So, I'd rather we keep it officially supported, but rather in the docs point out that we have the two new properties that offer more control.

matham (changes requested):

kivy/uix/video.py
rnixx: I copied it over from Image. Thus i left it None for consistency reasons

matham: Oh, ok. Then we should add `allownone=True`, I think. Otherwise it's weird that you can initially set it to None, but not afterwards.

kivy/uix/video.py
rnixx: > for both video and image

Using video path for image source just cause the image error log message and has no further advantage ;). But I'm fine with removing the deprecation warning if desired.

matham: Hmm, I assumed that the player would actually be able to get a preview image for some videos when a video is specified as source. Is that not the case? Did it never work in the past?

rnixx: Well, at least I've never seen what you describe. I also not see how this should integrate into the superclass on __init__, there's just the source property set and image tries to load an image texture as far as i can tell.

kivy/uix/video.py
```diff
@@ -31,21 +42,51 @@ def on_duration_change(instance, value):
...
+    defaults to None.
+
+    .. versionadded:: 2.1.0
+    '''
+
+    preview_source = StringProperty(None)
+    '''Filename / source of a preview image displayed before video starts.
+
+    :attr:`preview_source` is a :class:`~kivy.properties.StringProperty` and
+    defaults to None.
```
matham, line 77:
```suggestion
    defaults to None. If None/empty string and :attr:`source ` is a
    filename, that will be used for the preview filename.
    Otherwise, :attr:`source` will be set to :attr:`preview_source`.
```

kivy/uix/video.py
rnixx: done

kivy/uix/video.py
```diff
@@ -130,13 +171,21 @@ class Video(Image):
     _video_load_event = None
 
     def __init__(self, **kwargs):
+        # B/C case source is given. use it as video source
+        if 'source' in kwargs:
+            kwargs['video_source'] = kwargs['source']
+        # Case preview is given
+        if 'preview_source' in kwargs:
+            kwargs['source'] = kwargs['preview_source']
+
```
matham, line 180:
```suggestion
        # Case preview is given
        if kwargs.get('preview_source'):
            kwargs['source'] = kwargs['preview_source']

```

I think it's better if we don't override `video_source`, but leave it to whatever the user set it to. And instead use the appropriate value later.

Also, we need to be careful not to overwrite in case `preview_source` is empty.

kivy/uix/video.py
```diff
@@ -130,13 +171,21 @@ class Video(Image):
...
         self._video = None
         super(Video, self).__init__(**kwargs)
-        self.fbind('source', self._trigger_video_load)
+        self.fbind('video_source', self._trigger_video_load)
+        self.fbind('source', self._bc_video_load)
```
matham, line 184:
```suggestion
        self.fbind('preview_source', self._bc_video_load)
        if self.preview_source:
            self.source = self.preview_source
```
We only need to update `source` when `preview_source` changes. We shouldn't update `video_source` from whatever the user set it to. But we need to make sure it's updated here in case it was changed in co-inherited class in `super`.

kivy/uix/video.py
```diff
@@ -130,13 +171,21 @@ class Video(Image):
...
 
         if "eos" in kwargs:
             self.options["eos"] = kwargs["eos"]
-        if self.source:
+        if self.video_source:
```
matham, line 188:
```suggestion
        if self.source or self.video_source:
```

kivy/uix/video.py
```diff
@@ -163,6 +212,9 @@ def seek(self, percent, precise=True):
             raise Exception('Video not loaded.')
         self._video.seek(percent, precise=precise)
 
+    def _bc_video_load(self, inst, val):
+        self.video_source = val
```
matham, line 216:
```suggestion
        self.source = self.preview_source
```

kivy/uix/video.py
```diff
@@ -174,11 +226,11 @@ def _do_video_load(self, *largs):
         if CoreVideo is None:
             return
         self.unload()
-        if not self.source:
+        if not self.video_source:
```
matham, line 229:
```suggestion
        if not self.source and not self.video_source:
```

kivy/uix/video.py
```diff
@@ -174,11 +226,11 @@ def _do_video_load(self, *largs):
...
             self._video = None
             self.texture = None
         else:
-            filename = self.source
+            filename = self.video_source
```
matham, line 233:
```suggestion
            filename = self.video_source or self.source
```

matham (changes requested):

kivy/uix/video.py
matham: ffpyplayer for example can load a gif as an image, I think. So I it may be able to similarly load a image from a video, but I'm not sure.

But either way, I guess even if it couldn't, I'd prefer to not deprecate. When a user sees that warning they'll hopefully look at the docs and use the new properties. If we could remove source altogether, I would probably feel differently, but since we can't I think as it is now is better.

rnixx: I already removed the deprecation warnings.

matham (dismissed):

kivy/uix/video.py
```diff
@@ -163,6 +213,9 @@ def seek(self, percent, precise=True):
             raise Exception('Video not loaded.')
         self._video.seek(percent, precise=precise)
 
+    def _update_source_from_preview(self, inst, val):
+        self.source = self.preview_source
```
rnixx, line 217: I fear we introduce an unwanted behavior change here.

in my initial implementation it was possible to set ``source`` and ``video_source`` on an instane of ``Video`` and both triggered a video load. Now, setting ``source`` has no effect any mode (L135) but instead we cross-map ``source`` and ``preview_source``. Thats what i wanted to avoid in the first place. ``preview_source`` is useful only on widget ``__init__`` in the initial implementation (by intent)

rnixx: To be honest, i also not see why keeping ``video_source`` and ``source`` not in sync is a bad thing. The ``preview_image`` in it's current implementation is meant for initial placeholder only when the video gets instanciated to prevent the white canvas while loading (or as long as the video not gets started).

An additional case would be to display this image again once the player stops, but that would require acting on eos and is another aspect.

All in all it feels to me that the initial implementation is less immersive

rnixx:
I did an overhaul of this PR which tries to address issues stated above. Internally, the real video source now indeed refers to ``video_source`` only and ``preview_source`` no longer has any relationship to ``source``. If ``source`` is used, the widget behaves as it does right now in the original implementation. To address implicit magic behavior with the use of the ``source`` property, i introduced ``set_texture_from_resource`` on ``Image`` widget, prevent automatic ``texture_update`` on ``Image `` when ``source`` gets set on instance of ``Video`` (this should not be a behavioral problem because video handles the image texture on it's own) and set the preview image explicitly. I think this is way cleaner in the end than the hack with the ``source`` property being used for different purposes in the first place.

rnixx:
cc @tshirtman

tshirtman:
I didn’t think this would require an api change, it wasn’t possible to simply add `preview_source` and use it to load the texture until it’s replaced by the textures from the video once it start streaming? I’m not strongly against the change, and i might see the necessity on further examination, but i wonder what i missed that prevents this from working.

rnixx:
@tshirtman hm, yes, since now video_source and source are essentially the same thing, with the change in ``Image`` the implementation can further be simplified. Thanks for the hint. When staring at the same thing for hours things can get complicated in mind ;)

rnixx:
here we go ;)

tshirtman:
Ah, looks better to me indeed :). I’m a bit puzzled about why the change in `texture_update` is enough, but i guess i’m just rusty on how the video providers are interfacing with the video player, maybe the name of the function is a bit misleading, as i would expect it to be called everytime the stream sends a new texture, but it’s probably not the case, or it wouldn’t work at all, anyway, that’s outside of the concerns of this PR, if it works, then LGTM.

tshirtman (approved):

rnixx:
``texture_update`` is bound in ``Image.__init__`` to changes of the ``source`` property, so is only called if ``source`` gets changed. The video updates ``Image.texture`` directly here https://github.com/kivy/kivy/blob/master/kivy/uix/video.py#L214

tshirtman:
Right, thanks for the heads up :)

matham:
Nice solution!

To answer the original question why I didn't like overwriting the `video_source` from `source`. As a user I find it hostile when a library overwrites a property that I set. Because when I set the property of a instance, I expect it to not suddenly change from under me, but stay what I set it. Instead, I expect the library to use the property and create perhaps internal properties that depend on public properties and the internal properties change. Or use `if` etc to do stuff based on public API properties. But it should make the best effort to not change something set by a user (this is sometimes called round-tripping in other context).

But, I was grudgingly ok with changing `source` based on `preview_source` because we had no other option, seemingly.

PR #7475, opened 2021-04-09
MouseHoverEventTestCase: Skip testing on Windows platform

pythonic64:
Class `MouseHoverEventTestCase`, added in https://github.com/kivy/kivy/pull/7387, is causing `MultitouchSimulatorTestCase` to fail on CI on Windows platform. This pull request adds code to skip `MouseHoverEventTestCase`.

For more info see https://github.com/kivy/kivy/pull/7387#issuecomment-812091417.

Maintainer merge checklist:

### DIFF
--- a/kivy/tests/test_mouse_hover_event.py
+++ b/kivy/tests/test_mouse_hover_event.py
@@ -1,6 +1,14 @@
+import os
+import pytest
+
+from kivy import platform
 from kivy.tests.common import GraphicUnitTest
 
 
+@pytest.mark.skipif(
+    platform == 'win' and 'CI' in os.environ,
+    reason='Causes test_mouse_multitouchsim.py to fail on CI.'
+)
 class MouseHoverEventTestCase(GraphicUnitTest):
     '''Tests hover event from `MouseMotionEventProvider`.
     '''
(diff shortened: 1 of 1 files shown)


kivy/tests/test_mouse_hover_event.py
```diff
@@ -1,6 +1,11 @@
+import pytest
```
matham, line 1:
```suggestion
import pytest
import os
```

kivy/tests/test_mouse_hover_event.py
```diff
@@ -1,6 +1,11 @@
...
+
+from kivy import platform
 from kivy.tests.common import GraphicUnitTest
 
 
+@pytest.mark.skipif(platform == 'win',
+                    reason='Causes test_mouse_multitouchsim.py to fail on CI.')
```
matham, line 8:
```suggestion
@pytest.mark.skipif(
  platform == 'win' and 'CI' in os.environ,
  reason='Causes test_mouse_multitouchsim.py to fail on CI.'
)
```

I'd rather it's only [skipped on CI](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables).

matham (changes requested):

matham (approved):

PR #7466, opened 2021-04-05
Extend test coverage - 1

Zen-CODE:
This PR is (hopefully) the first of many to extend test coverage. Even through the tests my not be thorough, at least it adds coverage and gives as starting point for more meaningful coverage.

My current strategy is to generate coverage report stats and to systematically start targeting test at lines of uncovered code. Rather than accumulate a massive PR, bit my bit makes it easier to review and manage.

Maintainer merge checklist
* [ ] Title is descriptive/clear for inclusion in release notes.
* [ ] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [ ] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #7476, opened 2021-04-09
[WIP] MultitouchSimulatorTestCase: Fix CI failing test

pythonic64:
Possible fix for `MultitouchSimulatorTestCase` which fails if `MouseHoverEventTestCase` is enabled.

Maintainer merge checklist:


dotrichard:
Funnily enough, I am able to reproduce in this in a VirtualBox VM running Mint 20.1. If it would help, I could breakpoint and get you debugging info, if that would help?
```
>           color = mouse.touches[
                event_id
            ].ud.get('_drawelement')[0].proxy_ref
E           TypeError: 'NoneType' object is not subscriptable

kivy/tests/test_mouse_multitouchsim.py:261: TypeError
```

dotrichard:
As it turns out, the `color` variable that is assigned to here is not even used? Could this be just a case of an invalid test?

https://github.com/kivy/kivy/blob/53b655b0ef4bdfe8bf81a2bef8f09179a4917076/kivy/tests/test_mouse_multitouchsim.py#L261

matham:
>As it turns out, the color variable that is assigned to here is not even used? Could this be just a case of an invalid test?

I assumed it is there to test that the key exists, rather to use the color in some way.

dotrichard:
@pythonic64 So, I've pushed changes (accidentally. I'm an idiot. What can I say. But we previously had branch protection, which we need to re-instate. Will follow up...). These changes remove redundant value retrievals which directly affect this issue.

May I suggest you merge these changes from master, and report back? Do you think there is a real issue here, or were they just invalid tests? 

Thanks

matham:
Well, that's the question. Is the failing test on Windows a real issue, or is it just an artifact. To answer I think we'd need to understand what is being accessed and why it's failing (only on VM). Since we can't easily do it on the CI, it's good to know it can be replicated on VM because that'll make it easier to isolate the issue,

Zen-CODE:
I'm going to suggest we just leave that removed. It effectively tests a hidden property, so is already of dubious value. Tests should ideally be limited to very specific things.  @matham  Agreed?

To exploit this opportunity, please could you quickly check this: https://github.com/kivy/kivy/pull/7466
I saw you added the pytest configuration, so I'd be grateful for your thumbs up/suggestions. Thanks

matham:
I agree we should leave it removed for now, because it's important for the tests to work and not show unrelated errors in submitted PRs. However, it would be nice if eventually someone works out why it failed because it makes me uncomfortable to have random failures without knowing why.

pythonic64:
@dotrichard Removing unused `color` variable won't solve the issue of test failing on CI because it will still fail with the same error on variable `ellipse`. Test can/should be changed to assert that red dot (for multi touch sim) exits, by asserting that both `color` and `ellipse` exist. 

I run the tests on VM Ubuntu 18 and it did not reproduce the failing test. Can you collect values from `mouse.touches` and `mouse.hover_event` ?

Zen-CODE:
Since removing that `color` check,I no longer get errors. That seems to solve the problem?

pythonic64:
@Zen-CODE I don't think so, it just hides the problem. Both `color` and `ellipse` variables should work on CI, as same as they work on local Windows/Ubuntu.
```
color = mouse.touches[
    event_id
].ud.get('_drawelement')[0].proxy_ref
ellipse = mouse.touches[
    event_id
].ud.get('_drawelement')[1].proxy_ref
```

matham:
>I run the tests on VM Ubuntu 18 and it did not reproduce the failing test

I think you'd probably only be able to reproduce on Windows VM (maybe only on mint). Because it originally doesn't happen on Ubuntu CI.

pythonic64:
@matham Yes, but I remembered that I have virtual Ubuntu 18, so I gave it a try to see if test will fail. But they didn't :)

pythonic64:
Still failing, but now with referencing ellipse fails.
```
___ MultitouchSimulatorTestCase.test_multitouch_on_demand_scatter_rightmove ___

self = <kivy.tests.test_mouse_multitouchsim.MultitouchSimulatorTestCase testMethod=test_multitouch_on_demand_scatter_rightmove>

    def test_multitouch_on_demand_scatter_rightmove(self):
>       self.multitouch_dot_move(
            'right', on_demand=True, scatter=True
        )

kivy\tests\test_mouse_multitouchsim.py:531: 

self = <kivy.tests.test_mouse_multitouchsim.MultitouchSimulatorTestCase testMethod=test_multitouch_on_demand_scatter_rightmove>
button = 'right', kwargs = {'on_demand': True, 'scatter': True}
eventloop = <kivy.base.EventLoopBase object at 0x0000022A9AD7E350>
win = <kivy.core.window.window_sdl2.WindowSDL object at 0x0000022A9B2C5AC0>
mouse = <kivy.input.providers.mouse.MouseMotionEventProvider object at 0x0000022AA1FDFC10>
wid = <kivy.uix.scatter.Scatter object at 0x0000022AA4F8F0B0>
event_id = 'unittest1'

    def multitouch_dot_move(self, button, **kwargs):
        # touch -> dot appears -> move touch -> dot moves
        # -> release touch -> touch & dot disappear
        eventloop, win, mouse, wid = self.mouse_init(**kwargs)
    
        # register mouse provider
        mouse.start()
        eventloop.add_input_provider(mouse)
    
        # no mouse touch anywhere
        self.assertEqual(mouse.counter, 0)
        self.assertEqual(mouse.touches, {})
    
        # right button down, red dot should appear
        # if the 'multitouch_on_demand' is disabled
        win.dispatch(
            'on_mouse_down',
            10, self.correct_y(win, 10),
            'right', {}
        )
        event_id = next(iter(mouse.touches))
        self.assertEqual(mouse.counter, 1)
    
        if 'on_demand' in kwargs and 'scatter' not in kwargs:
            # doesn't do anything on a pure Button
            self.render(wid)
    
            # cleanup!
            # remove mouse provider
            mouse.stop()
            eventloop.remove_input_provider(mouse)
            return
    
        # XXX right button up
        # first release the touch then check, so that we
        # have the red dot drawn in on_demand and in the
        # default (multitouch everywhere) because in the
        # multitouch_on_demand is the circle drawn after
        # the touch is released (in on_mouse_release)
        elif 'on_demand' in kwargs and 'scatter' in kwargs:
            # on_demand works after the touch is up
            self.assertIn(
                'multitouch_sim',
                mouse.touches[event_id].profile
            )
            self.assertTrue(mouse.multitouch_on_demand)
    
            # multitouch_sim is changed in on_touch_down
            # method of the widget that's able to handle
            # multiple touches, therefore for Scatter we
            # need to dispatch the method and because we
            # triggered only on_mouse_down directly i.e.
            # without ME dispatch, on_touch_down was not
            # called == multitouch_sim is False
            self.advance_frames(1)  # initialize stuff
            wid.on_touch_down(mouse.touches[event_id])
            wid.on_touch_up(mouse.touches[event_id])
            self.assertTrue(mouse.touches[event_id].multitouch_sim)
    
            win.dispatch(
                'on_mouse_up',
                10, self.correct_y(win, 10),
                'right', {}
            )
>           ellipse = mouse.touches[
                event_id
            ].ud.get('_drawelement')[1].proxy_ref
E           TypeError: 'NoneType' object is not subscriptable

kivy\tests\test_mouse_multitouchsim.py:261: TypeError
```
I enabled the test with the latest master in this pull request: https://github.com/kivy/kivy/pull/7483.

pythonic64:
This error doesn't occur on specific Python/Windows version, it's almost "random" as tests on some Python/Windows combination will pass and then on one will fail.

Zen-CODE:
+1 for the effort. It is appreciated.

pythonic64:
Fixed in the following requests:
- https://github.com/kivy/kivy/pull/7505
- https://github.com/kivy/kivy/pull/7512
- https://github.com/kivy/kivy/pull/7493

PR #7481, opened 2021-04-16
Doc: Properties spelling fix

dirtbirb:
I hope this is the right way to contribute a tiny fix, please let me know if not!

Wanted to fix Variabel -> Variable, but also wasn't sure how to pluralize a class name, so I went by what was done elsewhere in the code ("NumericProperties" in kivy/_metrics.pyx line 17).

Maintainer merge checklist

### DIFF
--- a/kivy/properties.pyx
+++ b/kivy/properties.pyx
@@ -1692,7 +1692,7 @@ cdef class VariableListProperty(Property):
             Not currently used.
 
     Keeping in mind that the `default` list is expanded to a list of length 4,
-    here are some examples of how VariableListProperties are handled.
+    here are some examples of how VariableListProperty is handled.
 
     - VariableListProperty([1]) represents [1, 1, 1, 1].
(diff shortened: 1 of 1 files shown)


kivy/properties.pyx
```diff
@@ -1692,7 +1692,7 @@ cdef class VariableListProperty(Property):
             Not currently used.
 
     Keeping in mind that the `default` list is expanded to a list of length 4,
-    here are some examples of how VariabelListProperty's are handled.
+    here are some examples of how VariableListProperties are handled.
```
matham, line 1695:
```suggestion
    here are some examples of how VariableListProperty is handled.
```

Good catch. But I think the singular avoids the issues you mentioned and seems correct.

matham (changes requested):

kivy/properties.pyx
dirtbirb: Agreed -- I think I accepted the suggestion, but it's kinda hard to tell!

matham (approved):

PR #7483, opened 2021-04-16
MouseHoverEventTestCase: Enabled on Windows CI

pythonic64:
Maintainer merge checklist


matham:
Were you able to figure out the issue, or we just accepting that'll we'll have to disable the test on win CI (which is ok if we can't figure it out)?

pythonic64:
Seems that patching of `Window.on_close` method or dispatching of `on_close` event is the problem.

pythonic64:
@matham You can merge this request too. We are skipping those three tests until someone finds a solution to why running those tests causes tests in `MultitouchSimulatorTestCase` to fail.

matham (approved):

PR #7484, opened 2021-04-16
WindowBase: Add to_normalized_pos method

pythonic64:
Method `to_relative_pos` to be used in motion event providers to transform absolute to relative coordinates. 

I didn't use name `normalize_pos` as suggested at https://github.com/kivy/kivy/pull/4238#discussion_r613552441 because it's close to vector normalize which has a different meaning (https://github.com/kivy/kivy/blob/master/kivy/vector.py#L265).

Maintainer merge checklist:

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -1389,8 +1389,8 @@ class WindowBase(EventDispatcher):
         return (x, y)
 
     def to_normalized_pos(self, x, y):
-        '''Transforms absolute to normalized coordinate using
-        :attr:`system_size`.
+        '''Transforms absolute coordinates to normalized (0-1) coordinates
+        using :attr:`system_size`.
 
         .. versionadded:: 2.1.0
         '''
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -1388,6 +1388,19 @@ class WindowBase(EventDispatcher):
     def to_window(self, x, y, initial=True, relative=False):
         return (x, y)
 
+    def to_relative_pos(self, x, y):
+        '''Transforms absolute to relative coordinate using
+        :attr:`system_size`.
+
+        .. versionadded:: 2.1.0
+        '''
+        x_max = self.system_size[0] - 1.0
+        y_max = self.system_size[1] - 1.0
+        return (
+            x / x_max if x_max > 0 else 0.0,
+            y / y_max if y_max > 0 else 0.0
+        )
+
     def _apply_transform(self, m):
         return m
 
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -1388,8 +1388,8 @@ class WindowBase(EventDispatcher):
     def to_window(self, x, y, initial=True, relative=False):
         return (x, y)
 
-    def to_relative_pos(self, x, y):
-        '''Transforms absolute to relative coordinate using
+    def to_normalized_pos(self, x, y):
+        '''Transforms absolute to normalized coordinate using
         :attr:`system_size`.
 
         .. versionadded:: 2.1.0
--- a/kivy/tests/test_window_base.py
+++ b/kivy/tests/test_window_base.py
@@ -1,0 +1,16 @@
+from itertools import product
+
+from kivy.tests import GraphicUnitTest
+
+
+class WindowBaseTest(GraphicUnitTest):
+
+    def test_to_relative_pos(self):
+        for x, y in product([0, 319, 50, 51], [0, 239, 50, 51]):
+            self.Window.system_size = (320, 240)
+            w, h = self.Window.system_size
+            expected_sx = x / (w - 1.0)
+            expected_sy = y / (h - 1.0)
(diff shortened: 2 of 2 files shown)


matham:
Hmm I see what you mean about normalize, but I don't like relative because it's not clear what it's being relative to so it wouldn't be clear what it does from the name. I.e. it's not making it relative to a corner or widget (it's already relative), instead it's normalizing it to the 0-1 range.

I think it's ok to have names that have contextual meaning, but may clash if seen outside of any context. Because in the Window context, normalize doesn't have to mean what it means for vectors (although if you think about vectors a lot I can see you'd think about that). So I think normalize is ok here.

But if you still don't like that, we should try to find a better name. Maybe `to_unity`, but it feels weird.

pythonic64:
Maybe `to_relative_normalized_pos` or shorten version `to_norm_relative_pos`?

matham:
Personally I would go with `to_normalized_pos` rather than `to_norm_relative_pos`, because the pos passed in would already be relative and would only be normalized.

So this function would not make it further relative (unless this function also handles rotation or does some kind of offset computation).

But I'm ok with whichever you choose.

pythonic64:
Ok, `to_normalized_pos` it is!

kivy/core/window/__init__.py
```diff
@@ -1388,6 +1388,19 @@ def to_widget(self, x, y, initial=True, relative=False):
     def to_window(self, x, y, initial=True, relative=False):
         return (x, y)
 
+    def to_normalized_pos(self, x, y):
+        '''Transforms absolute to normalized coordinate using
```
matham, line 1392:
```suggestion
        '''Transforms absolute pos to normalized (0-1) coordinates using
```

kivy/tests/test_window_base.py
```diff
@@ -0,0 +1,16 @@
+from itertools import product
+
+from kivy.tests import GraphicUnitTest
+
+
+class WindowBaseTest(GraphicUnitTest):
+
+    def test_to_normalized_pos(self):
+        for x, y in product([0, 319, 50, 51], [0, 239, 50, 51]):
+            self.Window.system_size = (320, 240)
```
matham, line 10: Probably should save/restore original size after test with try/finally. So other tests use correct size.

matham (changes requested):

kivy/core/window/__init__.py
pythonic64: Maybe it's better to use "coordinates" instead of "pos" because that wording is used in `Widget.to_parent` and `Widget.to_local`?
Therefore:  "Transforms absolute coordinates to normalized (0-1) coordinates using :attr:`system_size`."

kivy/tests/test_window_base.py
pythonic64: Agree. I'll add that change.

kivy/core/window/__init__.py
matham: yeah, although `to_norm_coord` is probably better!? Rather than having to type it all out and spell it correctly...

pythonic64: I would rather use `to_normalized_pos`.

matham: That's OK with me.

pythonic64:
Another error on Python 3.8 and Windows 32bit:
```
__________________ ScrollViewTestCase.test_smooth_scroll_end __________________

self = <kivy.tests.test_uix_scrollview.ScrollViewTestCase testMethod=test_smooth_scroll_end>

    def test_smooth_scroll_end(self):
        EventLoop.ensure_window()
        win = EventLoop.window
        grid = _TestGrid()
        scroll = ScrollView(smooth_scroll_end=10)
    
        assert scroll.smooth_scroll_end == 10
        scroll.add_widget(grid)
    
        # XXX this shouldn't be needed, but previous tests apparently
        # don't cleanup
        while win.children:
            win.remove_widget(win.children[0])
    
        win.add_widget(scroll)
    
        # get widgets ready
        EventLoop.idle()
    
        e = scroll.effect_y
        assert e.velocity == 0
    
        touch = UTMotionEvent("unittest", next(touch_id), {
            "x": scroll.center_x / float(win.width),
            "y": scroll.center_y / float(win.height),
        })
    
        touch.profile.append('button')
        touch.button = 'scrollup'
    
        EventLoop.post_dispatch_input("begin", touch)
        # EventLoop.post_dispatch_input("update", touch)
        assert e.velocity == 10 * scroll.scroll_wheel_distance
        EventLoop.idle()
        assert 0 < e.velocity < 10 * scroll.scroll_wheel_distance
        EventLoop.post_dispatch_input("end", touch)
        EventLoop.idle()
>       assert 0 < e.velocity < 10 * scroll.scroll_wheel_distance
E       AssertionError: assert 0 < -12.671674735189558
E        +  where -12.671674735189558 = <kivy.effects.dampedscroll.DampedScrollEffect object at 0x12E64648>.velocity

kivy\tests\test_uix_scrollview.py:333: AssertionError
```

PR #7485, opened 2021-04-18
Partially fixes #7415, input works better

### LINKED ISSUE #7415: Text Input typing bug
* Python: 3.8
* OS: android
* Kivy: master

When you enter text in the text field, after pressing the space bar, it disappears somewhere, if you start erasing it, it appears, but only after erasing the visible characters.

https://user-images.githubusercontent.com/40869738/110664448-745b6d80-81d8-11eb-8b8c-928131a487bc.mp4

tshirtman:
backspace still incorrect

Maintainer merge checklist


kivy/uix/textinput.py
```diff
@@ -2850,12 +2842,19 @@ def keyboard_on_key_up(self, window, keycode):
...
-            self._key_up(key)
+            self._key_up(k)
 
     def keyboard_on_textinput(self, window, text):
         if self._selection:
             self.delete_selection()
+        if self._ime_composition:
+            self._replace_ime(text)
+            self._ime_composition = ''
+        if text.endswith('\n') and not self.multiline:
```
matham, line 2853: Do we handle `\r\n`?

tshirtman: I don’t know if we need to, this happens on android when you press enter after swiping a word, the word is sent again, but with \n at the end, but maybe we can add it to be safe.

matham: Ok, either works fine I guess (windows doesn't have swipe I imagine).

Issue #4387, opened 2016-06-12
on_mouse_down (and up) not called when bound to the window directly in build


CreamyCookie:
However, compared to `on_mouse_down`, the `on_key_down` event is triggered.

This dosen't work:

``` python
from kivy.app import App
from kivy.core.window import Window
from kivy.uix.widget import Widget


class MyWidget(Widget):
    pass


class MouseScrollApp(App):
    def build(self):
        Window.bind(on_key_down=self.on_key_down)
        Window.bind(on_mouse_up=self.on_mouse_down)

        return MyWidget()

    def on_key_down(self, *args):
        print("I work with keys: ", *args)

    def on_mouse_down(self, *args):
        print(*args)


if __name__ == '__main__':
    MouseScrollApp().run()
```

But this does:

``` python
from kivy.app import App
from kivy.core.window import Window
from kivy.uix.widget import Widget
from kivy.clock import Clock


class MyWidget(Widget):
    pass


class MouseScrollApp(App):
    def setup_binds(self, *l):
        Window.bind(on_key_down=self.on_key_down)
        Window.bind(on_mouse_up=self.on_mouse_down)

    def build(self):
        Clock.schedule_once(self.setup_binds)
        return MyWidget()

    def on_key_down(self, *args):
        print("I work with keys: ", *args)

    def on_mouse_down(self, *args):
        print(*args)


if __name__ == '__main__':
MouseScrollApp().run()
```

pythonic64:
Because `MouseMotionEventProvider` binds to `on_mouse_xxx` events and handlers will always accept mouse events (by returning `True`), so in the first case `MouseScrollApp` binds to `on_mouse_up` before `MouseMotionEventProvider` does and because event dispatching order is reversed, provider gets first to handle the event and accepts it which leaves `MouseScrollApp` without that same event. In the second case `MouseScrollApp` gets to handle the event first, but because it does not accept it event gets passed to provider and provider handles it. 

If you add a button in your second case:
```
from kivy.app import App
from kivy.clock import Clock
from kivy.core.window import Window
from kivy.uix.button import Button


class MouseScrollApp(App):

    def setup_binds(self, *l):
        Window.bind(on_key_down=self.on_key_down)
        Window.bind(on_mouse_down=self.on_mouse_down)

    def build(self):
        Clock.schedule_once(self.setup_binds)
        return Button(size_hint=(1, None))

    def on_key_down(self, *args):
        print("I work with keys: ", *args)

[...]

PR #7492, opened 2021-04-22
MouseMotionEventProvider: Refactor of provider and tests

pythonic64:
Changes:
- Uses Window.to_normalize_pos method
- Uses `win` argument in methods instead of `Window` 
- Does not accept `on_mouse_xxx` events so that user can handle them which fixes issue https://github.com/kivy/kivy/issues/4387

Maintainer merge checklist:

### DIFF
--- a/kivy/input/providers/mouse.py
+++ b/kivy/input/providers/mouse.py
@@ -209,55 +209,55 @@ class MouseMotionEventProvider(MotionEventProvider):
                 return True
         return False
 
-    def find_touch(self, x, y):
-        factor = 10. / EventLoop.window.system_size[0]
-        for t in self.touches.values():
-            if abs(x - t.sx) < factor and abs(y - t.sy) < factor:
-                return t
-        return False
+    def find_touch(self, win, x, y):
+        factor = 10. / win.system_size[0]
+        for touch in self.touches.values():
+            if abs(x - touch.sx) < factor and abs(y - touch.sy) < factor:
+                return touch
+        return None
 
     def create_event_id(self):
         self.counter += 1
         return self.device + str(self.counter)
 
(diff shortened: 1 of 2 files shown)


pythonic64:
Tests still fail, now on Python 3.8 64bit
```
___ MultitouchSimulatorTestCase.test_multitouch_on_demand_scatter_lefttouch ___

self = <kivy.tests.test_mouse_multitouchsim.MultitouchSimulatorTestCase testMethod=test_multitouch_on_demand_scatter_lefttouch>

    def test_multitouch_on_demand_scatter_lefttouch(self):
>       self.multitouch_dot_touch(
            'left', on_demand=True, scatter=True
        )

kivy\tests\test_mouse_multitouchsim.py:516: 
kivy\tests\test_mouse_multitouchsim.py:123: in multitouch_dot_touch
    self.assertIsNotNone(
E   AssertionError: unexpectedly None
```

matham (dismissed):
Great!

I guess `create_touch` etc is not considered part of the public API so we should not be worried about changing it!? So it should be fine.

pythonic64:
I also removed `return True` from `on_mouse_xxx` handlers to fix https://github.com/kivy/kivy/issues/4387.
Do you agree?

matham:
Yeah, I think that's fine.

In general I'm not a fan of returning True to stop a dispatch chain as an API choice because it's very limiting and makes cooperating with other callbacks hard. Although it's fine for widgets and their events because the on_touch_xxx events are specifically designed around it.

So if later we do need to stop dispatching (like current behavior) we could find a better way of doing it (e.g. marking the event somehow).

pythonic64:
I agree, later we can have a method which can set an internal flag in motion event.

I had to rebase the branch again because of conflicts in master branch.

matham (approved):

PR #7502, opened 2021-04-27
MultitouchSimulatorTestCase: Don't render widgets in tests

pythonic64:
Overrides method `render` to skip rendering of widget which causes `EventLoop` to start already started input providers which causes failures on CI. Attribute `framecount` is now set to 3 to avoid `EventLoop.stop` call in `on_window_filp` method which in some tests can stop the event loop before the test ends.

Maintainer merge checklist:

### DIFF
--- a/kivy/tests/test_mouse_multitouchsim.py
+++ b/kivy/tests/test_mouse_multitouchsim.py
@@ -2,7 +2,11 @@ from kivy.tests.common import GraphicUnitTest
 
 
 class MultitouchSimulatorTestCase(GraphicUnitTest):
-    framecount = 0
+
+    framecount = 3
+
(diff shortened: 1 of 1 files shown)


pythonic64:
@matham Can you merge this request? Then I can rebase other related requests and CI will re-run the tests.

matham (approved):
Makes sense that if you manually start some app components you don't want kivy to start them for you. However, ideally kivy would be smart enough to not start an already started component!?

pythonic64:
Can you suggest how to check in `start/stop` methods of `EventLoopBase` if loop is started/stopped?
There are `status`, `quit`, `stopping` attributes and it seems that combination of their values have to be tested to know if event loop is running or not.

PR #7504, opened 2021-04-28
kivy.cache.Cache : added a callback to be called when we're about to remove a key

olivier-boesch:
Added a per category callback function that is called when a key is about to be removed.

use case:
this may be useful if the cache is used for files. 
We create the file and add it the cache. When the key is about to be removed, we call the function (and delete the file).

Added code and docs.

Personally, I plan to use it with the mapview widget to offer a limited caching system for tiles.

Hope you'll find it useful

Maintainer merge checklist

### DIFF
--- a/kivy/cache.py
+++ b/kivy/cache.py
@@ -52,9 +52,9 @@ class Cache(object):
                 Time after which to delete the object if it has not been used.
                 If None, no timeout is applied.
             `remove_callback`: function (optional)
-                add a callback that's called when an object of this category is removed
-                callback signature : f(category, objet)
-                filled with the category and the object about to be removed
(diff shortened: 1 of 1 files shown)


pythonic64:
In my opinion it would be better for `Cache` to inherit from `EventDispatcher` and declare events `on_key_add` and `on_key_remove` which are dispatched with arguments  `category, key, value` when keys are added/removed. This way multiple callbacks can be added and using `bind` only weak references will be kept.

olivier-boesch:
I agree but the Cache class is used in the very inner core of kivy. I tried some changes and the consequences are not very good. I think I will a kind of user cache inspired by your suggestions.

I close this by now.

PR #7508, opened 2021-05-01
App: Process app quit event while paused

syrykh:
Hi,

While I was testing foreground services on Android, I found an issue when the kivy-based app hangs.

This issue happened to me in two cases (ref. [minimal demo app](https://github.com/syrykh/kivy-pause-issue) with apks available).
1. When foreground service is started and the app is swiped-out

https://user-images.githubusercontent.com/7023634/116788963-2581d400-aab5-11eb-8396-0b1e3ebeba08.mp4

2. When the app is paused

https://user-images.githubusercontent.com/7023634/116788985-3f231b80-aab5-11eb-8ded-d2c5ff222084.mp4

In both of these cases the app was still shown in process list (`ps -A`) and the last log line in logcat was:
```
05-01 17:24:32.843 12822 12822 V SDL     : onDestroy()
```

The hang is not actually visible to end user, but when there's an attempt to start the app again, only a blank screen is shown.

The issue was reproduced on Android 10@Pixel 1, Android 11@Pixel 2 and Android 11@AVD emulator, however it was not reproduced on Pocophone F1 with MIUI (probably because of the MIUI).

SDK: 30, NDK: r19c, kivy: master@fe1ab549e and [the app itself](https://github.com/syrykh/kivy-pause-issue)

While debugging this issue I came to mainloop of `core/window/window_sdl2.py` and added some logs there with [window_sdl2.py-log.patch.txt](https://github.com/kivy/kivy/files/6409976/window_sdl2.py-log.patch.txt). When the issue happens the logs are:
```
05-01 19:03:56.569 12902 12930 I python  : [INFO   ] [window_sdl2.py] mainloop: _pause_loop: wait_event done
05-01 19:03:56.569 12902 12930 I python  : [INFO   ] [window_sdl2.py] mainloop: _pause_loop: invoking self._win.poll()
05-01 19:03:56.569 12902 12930 I python  : [INFO   ] [window_sdl2.py] mainloop: _pause_loop: self._win.poll() done, event = ('quit',)
05-01 19:03:56.569 12902 12930 I python  : [INFO   ] [window_sdl2.py] mainloop: _pause_loop: invoking wait_event
```

That means that the pause loop receives a 'quit' event, ignores it, and waits for the next event, which will never come.
My PR fixes this issue by breaking the mainloop in this case.

Logs after the fix became:
```
05-01 19:34:23.435 13556 13556 V SDL     : onDestroy()
05-01 19:34:23.441 13556 13584 I python  : [INFO   ] [Base        ] Leaving application in progress...
05-01 19:34:23.442 13556 13584 I python  : Python for android ended.
```

**Demos with the fix**
1. When foreground service is started and the app is swiped-out

https://user-images.githubusercontent.com/7023634/116789638-7f37cd80-aab8-11eb-8417-737f61b5f377.mp4

2. When the app is paused

https://user-images.githubusercontent.com/7023634/116789648-8b238f80-aab8-11eb-9130-32582d076cdd.mp4

Not sure how to add unittests for this issue. The fix was verified on Android 10@Pixel 1, Android 11@Pixel 2 and Android 11@AVD emulator.

Please accept this PR. Thanks in advance!

Maintainer merge checklist

### DIFF
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -514,9 +514,9 @@ class WindowSDL(WindowBase):
             if action == 'dropfile':
                 dropfile = args
                 self.dispatch('on_dropfile', dropfile[0])
-            # A quit event might be received while the app is paused
-            elif action == 'quit':
-                EventLoop.quit = True
+            # app_terminating event might be received while the app is paused
+            # in this case EventLoop.quit will be set at _event_filter
+            elif EventLoop.quit:
                 return
 
         while True:
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -514,6 +514,10 @@ class WindowSDL(WindowBase):
             if action == 'dropfile':
                 dropfile = args
                 self.dispatch('on_dropfile', dropfile[0])
+            # A quit event might be received while the app is paused
+            elif action == 'quit':
+                EventLoop.quit = True
+                return
 
         while True:
             event = self._win.poll()


kivy/core/window/window_sdl2.py
```diff
@@ -514,6 +514,10 @@ def mainloop(self):
             if action == 'dropfile':
                 dropfile = args
                 self.dispatch('on_dropfile', dropfile[0])
+            # A quit event might be received while the app is paused
+            elif action == 'quit':
+                EventLoop.quit = True
```
matham, line 519:
```suggestion
                if self.dispatch('on_request_close'):
                    continue
                EventLoop.quit = True
```

matham (changes requested):

matham:
@akshayaurora btw, we should only have one component label, as that is used for release notes.

kivy/core/window/window_sdl2.py
syrykh: @matham, I was thinking about this. My concern is if the user will return `True` from `on_request_close`, it will hang.

matham: Yes, and I agree, but I'm wondering if that would introduce an inconsistency. Because why can a user ignore a quit event when *not* paused without it hanging? Or maybe it would also hang on android (it doesn't on desktop)?

Could you test whether returning True from `on_request_close` hangs the app if it's not paused? Or maybe android always pauses an app before closing it (I'm not super familiar with android)?

syrykh: > Could you test whether returning True from on_request_close hangs the app if it's not paused?

Sure.

syrykh: @matham, I did some tests and it appeared that returning True from on_request_close **does not hangs** the app if it's not paused. Then I continued my investigation and came to libsdl's SDL_android.d:
```c
/* Quit */
JNIEXPORT void JNICALL SDL_JAVA_INTERFACE(nativeQuit)(
                                    JNIEnv* env, jclass cls)
{
    /* Discard previous events. The user should have handled state storage
     * in SDL_APP_WILLENTERBACKGROUND. After nativeQuit() is called, no
     * events other than SDL_QUIT and SDL_APP_TERMINATING should fire */
    SDL_FlushEvents(SDL_FIRSTEVENT, SDL_LASTEVENT);
    /* Inject a SDL_QUIT event */
    SDL_SendQuit();
    SDL_SendAppEvent(SDL_APP_TERMINATING);
    /* Resume the event loop so that the app can catch SDL_QUIT which
     * should now be the top event in the event queue. */
    if (!SDL_SemValue(Android_ResumeSem)) SDL_SemPost(Android_ResumeSem);
}
```

I thought that the quit event might not be the right thing to analyze, because we cannot distinguish, whether it's when user wanted to close the app or because the app is terminated while paused. And then I realized that SDL_APP_TERMINATING event is the one.

I've added some more logs and saw:
```
05-02 20:06:02.964  7689  7689 V SDL     : onDestroy()
05-02 20:06:02.968  7689  7689 I python  : [INFO   ] [window_sdl2.py] _event_filter: action = 'app_terminating'
05-02 20:06:02.969  7689  7717 I python  : [INFO   ] [window_sdl2.py] mainloop: _pause_loop: wait_event done
05-02 20:06:02.970  7689  7717 I python  : [INFO   ] [window_sdl2.py] mainloop: _pause_loop: invoking self._win.poll()
05-02 20:06:02.970  7689  7717 I python  : [INFO   ] [window_sdl2.py] mainloop: _pause_loop: self._win.poll() done, event = ('quit',)
05-02 20:06:02.970  7689  7717 I python  : [INFO   ] [window_sdl2.py] mainloop: _pause_loop: invoking wait_event
```

And that's what I based my second commit on. I'm checking whether the `EventLoop.quit` was set by `_event_filter`. Not there should be no inconsistencies regarding quit event.

matham (approved):
Perfect! Thanks!

syrykh:
@matham, thanks a lot!

RobertFlatt:
Thank you, works great.

PR #7510, opened 2021-05-02
Removed python 2 support (behaviors)

Kulothungan16:
Maintainer merge checklist

### DIFF
--- a/kivy/uix/behaviors/button.py
+++ b/kivy/uix/behaviors/button.py
@@ -24,7 +24,7 @@ behaves like a button::
 
     class MyButton(ButtonBehavior, Image):
         def __init__(self, **kwargs):
-            super(MyButton, self).__init__(**kwargs)
+            super().__init__(**kwargs)
             self.source = 'atlas://data/images/defaulttheme/checkbox_off'
 
         def on_press(self):
@@ -118,7 +118,7 @@ class ButtonBehavior(object):
         if 'min_state_time' not in kwargs:
             self.min_state_time = float(Config.get('graphics',
                                                    'min_state_time'))
-        super(ButtonBehavior, self).__init__(**kwargs)
+        super().__init__(**kwargs)
         self.__state_event = None
         self.__touch_time = None
         self.fbind('state', self.cancel_event)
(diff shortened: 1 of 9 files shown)


matham:
Thank you for your contribution and working on this!

However, before you add additional commits and spend more time on this, we should pause and discuss whether we will merge it. In general it's a good idea to ask or touch base with maintainers before working on large changes to a project because it's possible it may not be accepted, which isn't a great feeling after you spent a lot of effort on it. So it's generally better to first ask whether changes would be accepted, and if agreed, then start working on them.

Now specifically about these changes. We generally don't merge these kind of changes, what people call stylistic changes. We generally only merge changes that either fix a bug or adds a feature. Only very rarely do we merge PRs that only contain stylistic changes - changes that don't fix a bug or add a feature.

The reason is that there are so many different styles to choose from, and styles change over time. So, if were to accept PRs that only change the code style, it would never end. Because styles always change and everyone has different preferences. E.g. the style of `super` (`super(...)`) you're replacing - that is not deprecated or in frowned upon, it's just an older style. But it's perfectly legal in Python3. So while it looks old, there's no reason to change it. The same for `class X(object)`.

An additional thing is that any change, no matter how simple introduces a possibility of a bug. So when merging a bug fix or new feature this possible cost of introducing a bug is nothing compared to the cost of not applying the bug fix/feature. But when merging a stylistic change, the cost of the possible bug is tends to outweigh other consideration so hence we don't merge them for fear of causing a bug.

Finally, such changes tend to touch many files, which is discouraged for PRs because it makes review harder and is much more likely to introduce bugs.

So overall I'd be against merging this PR. But I do very much appreciate your desire to contribute and improve Kivy.

Kulothungan16:
Yeah! I can understand now about stylistic changes, anyway I will try to avoid these kind of PRs

AndiEcker:
Ohhhh..... I'm a big fan of clean code
<https://www.martinfowler.com/tags/clean%20code.html>.

This PR would help to make Kivy's code cleaner, shorter and easier to
understand/read.

IMHO, rejecting stylistic changes, in general, is not a good idea for an
open-source project like Kivy.

Would you accept it, if there were a separate PR for every single
file/module? (I actually working on some code clean-ups for Kivy's layout
classes, but if you not accepting it I'll give up with these kinds of
contributions in the future)...

Am So., 2. Mai 2021 um 07:49 Uhr schrieb Kulothungan U.G <
***@***.***>:

PR #7517, opened 2021-05-05
Core: Use importlib's __import__ for compatibility with patching

hiaselhans:
This uses importlib.__import__ instead of plain __import__

The package 'pythonnet' breaks kivy by patching python's global __import__.
Of course it's not kivy to blame here, but using importlib doesn't hurt and as a side effect is better documented.

https://github.com/pythonnet/pythonnet/issues/1245
https://github.com/pythonnet/pythonnet/issues/547

### DIFF
--- a/kivy/core/__init__.py
+++ b/kivy/core/__init__.py
@@ -25,6 +25,7 @@ import sys
 import traceback
 import tempfile
 import subprocess
+import importlib
 import kivy
 from kivy.logger import Logger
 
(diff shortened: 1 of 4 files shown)


matham:
Thanks for the fix. There are some PEP8 issues though (please see the failed test).

hiaselhans:
thanks @matham for the quick response and title fix! :)

just pushed the pep8 fix..

matham (approved):

PR #7523, opened 2021-05-08
Video: Fix ffpyplayer if unload() called after play() but thread not started yet

rnixx:
When a ffpyplayer based video instance gets started and unload before the frame fetching thread is alive, the thread never stops because ``_ffplayer_need_quit`` gets reset in ``unload()``.

The fix resets thread controls just before thread gets started and additionally makes sure no other thread for this instance is running.

Maintainer merge checklist

### DIFF
--- a/kivy/core/video/video_ffpyplayer.py
+++ b/kivy/core/video/video_ffpyplayer.py
@@ -211,7 +211,7 @@ class VideoFFPy(VideoBase):
                 self._texture = fbo.texture
             else:
                 self._texture = Texture.create(size=self._size,
-                                                colorfmt='rgba')
+                                               colorfmt='rgba')
 
             # XXX FIXME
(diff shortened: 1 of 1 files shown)


matham:
I don't understand how this could be an issue. In unload, we block until the thread is dead (through join). Are you saying that if you call join on a thread after it started but perhaps before it had a chance to run, it doesn't really "join" the thread? That seems like a issue with Python if that's the case.

rnixx:
> I don't understand how this could be an issue.

At least it's no bug with python threading. In my case it is an issue when stressing the application with a lot of touch events (mtdev). I need to double check whether i forgot to delegate some call to the main thread (UI actions may be triggered via network as well).

As you said, the ffpyplayer implementation should not cause the problem, at least not if no concurrent access occurs, which seems to be the case in my app.

PR #7536, opened 2021-05-19
EventDispatcher: Add nicer error message for non-existing properties

sanderland:
Currently passing unknown properties to e.g. a Widget will cause a rather puzzling error message about the init function of `object` not taking this.
This PR attempts to rectify this and give helpful error messages.

Maintainer merge checklist


matham:
Unfortunately, we cannot do this because it will break multiple inheritance.

Consider the following classes, replicating a typical multiple inheritance scenario:

```py
class Event:
    def __init__(self, **kwargs):
        print(f'Event: {self}, {kwargs}')
        super().__init__(**kwargs)

class Behavior:
    def __init__(self, name):
        print(f'Behavior: {self}, name={name}')
        super().__init__()

class Widget(Event):
    pass

class Widget2(Behavior, Event):
    pass

class Widget3(Event, Behavior):
    pass
```

Then doing the following:
```py
Widget(name='Pasta')
Event: <__main__.Widget object at 0x000001E703F04518>, {'name': 'Pasta'}
  File "E:\Python\Python37\lib\site-packages\IPython\core\interactiveshell.py", line 3326, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-16-6063f054cac4>", line 1, in <module>
    Widget(name='Pasta')
  File "<ipython-input-4-48d2fe342363>", line 4, in __init__
    super().__init__(**kwargs)
TypeError: object.__init__() takes exactly one argument (the instance to initialize)

Widget2(name='Pasta')
Behavior: <__main__.Widget2 object at 0x000001E703F04588>, name=Pasta
Event: <__main__.Widget2 object at 0x000001E703F04588>, {}
Out[17]: <__main__.Widget2 at 0x1e703f04588>

Widget3(name='Pasta')
Event: <__main__.Widget3 object at 0x000001E703F04780>, {'name': 'Pasta'}
Behavior: <__main__.Widget3 object at 0x000001E703F04780>, name=Pasta
Out[18]: <__main__.Widget3 at 0x1e703f04780>
```

The problem you're trying to solve is the vague error at `Widget(name='Pasta')`. However, if you don't allow EventDispatcher to forward kwargs to super, then `Widget3(name='Pasta')` won't work anymore because it will raise an exception, when it shouldn't.

sanderland:
> Unfortunately, we cannot do this because it will break multiple inheritance.

I see, that's unfortunate. The error is really quite horrible, both on typos and on properties added between versions.
Would checking for the specific exception and translating it work?

matham:
> Would checking for the specific exception and translating it work?

I think that could be a good idea. But the question is how will you distinguish other TypeErrors? The only way is by message content, but that can change between versions.

So, I think it's ok to catch `TypeError`, then if the msg start matches `object.__init__()`, then `raise TypeError("Maybe you set the wrong prop etc.") from e`.

But, at the same time we need at least 3 unit tests. One that tests that the base message of passing a kwargs to object starts with `object.__init__()` in case it changes in a future version. Another that tests that other `TypeError` is not swallowed, and a third testing the typical issue of passing in a non-existent prop.

kivy/_event.pyx
```diff
@@ -228,7 +228,16 @@ cdef class EventDispatcher(ObjectWithUid):
         prop_args = {
             k: kwargs.pop(k) for k in list(kwargs.keys()) if k in properties}
         self._kwargs_applied_init = set(prop_args.keys()) if prop_args else set()
-        super(EventDispatcher, self).__init__(**kwargs)
+
+        try: # at this point any kwargs passed may go to object which will error and cause confusion
+           super(EventDispatcher, self).__init__(**kwargs)
+        except TypeError as e:
+           if kwargs and "object.__init__() takes exactly one argument" in e:
```
matham, line 235:
```suggestion
           if kwargs and str(e).startswith("object.__init__()"):
```

kivy/_event.pyx
```diff
@@ -228,7 +228,16 @@ cdef class EventDispatcher(ObjectWithUid):
...
+               raise ValueError(
+                  'Unexpected properties {}, valid properties are {}'.format(
+                      list(kwargs.keys()), sorted(properties.keys()))
```
matham, line 238:
```suggestion
                      sorted(kwargs.keys()), sorted(properties.keys())) from e
```

matham (changes requested):

kivy/_event.pyx
```diff
@@ -228,7 +228,16 @@ cdef class EventDispatcher(ObjectWithUid):
...
             k: kwargs.pop(k) for k in list(kwargs.keys()) if k in properties}
         self._kwargs_applied_init = set(prop_args.keys()) if prop_args else set()
-        super(EventDispatcher, self).__init__(**kwargs)
+
+        try: # at this point any kwargs passed may go to object which will error and cause confusion
+           super(EventDispatcher, self).__init__(**kwargs)
+        except TypeError as e:
+           if kwargs and "object.__init__() takes exactly one argument" in e:
+               raise ValueError(
+                  'Unexpected properties {}, valid properties are {}'.format(
```
matham, line 237:
```suggestion
                  'Properties "{}" past to __init__ may not be valid '
                  'property names. Valid properties are "{}"'.format(
```

matham (changes requested):

kivy/tests/test_widget.py
```diff
@@ -118,3 +118,58 @@ def test_export_to_png(self):
...
         self.root.remove_widget(wid)
+
+    def test_unknown_property(self):
+        from kivy.uix.widget import Widget
+        with self.assertRaises(ValueError) as cm:
+            Widget(width=12, unkn="abc")
+        self.assertIn("Unexpected properties ['unkn']", str(cm.exception))
+
+    def test_known_property_multiple_inheritance(self):
+        class Event:
```
matham, line 129: This should just be `EventDispatcher` directly, because that's what we're testing.

kivy/tests/test_widget.py
```diff
@@ -118,3 +118,58 @@ def test_export_to_png(self):
...
+
+        with self.assertRaises(ValueError) as cm:
+            Widget(name='Pasta')
+        self.assertIn("Unexpected properties ['name']", str(cm.exception))
+
+        Widget2(name='Pasta')  # does not raise a ValueError
+        Widget3(name='Pasta')  # does not raise a ValueError
+
+    def test_pass_other_typeerror(self):
+        class Event:
```
matham, line 156: Same here.

matham (changes requested):

kivy/_event.pyx
matham: Please use the message I suggested because it's clearer and hedges in case the exception is because of an different error, just in case.

kivy/tests/test_properties.py
```diff
@@ -1266,3 +1266,58 @@ def callback(obj, val):
     event.b = 'goodbye'
     assert event.b == 'goodbye'
     assert args == (event, 'goodbye')
+
+
+def test_unknown_property(self):
```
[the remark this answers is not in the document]
sanderland: self is a fixture because this file is a mess

kivy/tests/test_properties.py
```diff
@@ -1266,3 +1266,58 @@ def callback(obj, val):
...
+    from kivy.uix.widget import Widget
+    with self.assertRaises(ValueError) as cm:
+        EventDispatcher(width=12, unkn="abc")
+    self.assertIn("Unexpected properties ['unkn']", str(cm.exception))
```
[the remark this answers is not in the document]
sanderland: I was trying to stick to existing style, but have no idea what that is anymore

kivy/tests/test_properties.py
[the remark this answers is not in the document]
sanderland: pytest styled it

matham:
You have a mystery `kivy/graphics/cgl.h`  file. Probably because you did `git add -a` rather than `git add -p` which lets you select what you want to add :)

sanderland:
> You have a mystery `kivy/graphics/cgl.h` file. Probably because you did `git add -a` rather than `git add -p` which lets you select what you want to add :)

Actually I have a tool around git which very aggressively assumes gitignore to be correct, adds all, generates commit messages, pulls/precommit hook retries when needed etc :)

kivy/tests/test_properties.py
```diff
@@ -1266,3 +1266,65 @@ def callback(obj, val):
...
     assert args == (event, 'goodbye')
+
+
+def test_unknown_property():
+    from kivy.properties import NumericProperty
+
+    class MyWidget(EventDispatcher):
+        width = NumericProperty(0)
+
+    with pytest.raises(ValueError) as cm:
```
matham, line 1277:
```suggestion
    with pytest.raises(TypeError) as cm:
```

kivy/tests/test_properties.py
```diff
@@ -1266,3 +1266,65 @@ def callback(obj, val):
...
+            print(f'Behavior: {self}, name={name}')
+            super().__init__()
+
+    class Widget2(Behavior, EventDispatcher):
+        pass
+
+    class Widget3(EventDispatcher, Behavior):
+        pass
+
+    with pytest.raises(ValueError) as cm:
```
matham, line 1295:
```suggestion
    with pytest.raises(TypeError) as cm:
```

kivy/tests/test_properties.py
```diff
@@ -1266,3 +1266,65 @@ def callback(obj, val):
...
+
+def test_object_init_error():  # the above 3 test rely on this
+    class TestCls(object):
+        def __init__(self, **kwargs):
+            super(TestCls, self).__init__(**kwargs)
+
+    with pytest.raises(TypeError) as cm:
+        TestCls(name='foo')
+    assert str(cm.value).startswith(
+        "object.__init__() takes exactly one argument")
```
matham, line 1330:
```suggestion
        "object.__init__() takes")
```

kivy/_event.pyx
```diff
@@ -228,7 +228,19 @@ cdef class EventDispatcher(ObjectWithUid):
...
             k: kwargs.pop(k) for k in list(kwargs.keys()) if k in properties}
         self._kwargs_applied_init = set(prop_args.keys()) if prop_args else set()
-        super(EventDispatcher, self).__init__(**kwargs)
+
+        # at this point any kwargs passed may go to object
+        # which will raise a TypeError and cause confusion
+        try:
+           super(EventDispatcher, self).__init__(**kwargs)
+        except TypeError as e:
+           if kwargs and str(e).startswith("object.__init__() takes exactly one argument"):
```
matham, line 237:
```suggestion
           if kwargs and str(e).startswith("object.__init__() takes"):
```

The CI shows the following failed test:

```
>       assert str(cm.value).startswith(
            "object.__init__() takes exactly one argument")
E       AssertionError: assert False
E        +  where False = <built-in method startswith of str object at 0x17CE8660>('object.__init__() takes exactly one argument')
E        +    where <built-in method startswith of str object at 0x17CE8660> = 'object.__init__() takes no parameters'.startswith
E        +      where 'object.__init__() takes no parameters' = str(TypeError('object.__init__() takes no parameters',))
E        +        where TypeError('object.__init__() takes no parameters',) = <ExceptionInfo TypeError('object.__init__() takes no parameters',) tblen=2>.value

```

Because the message after `takes` can be different.

matham (changes requested):

kivy/_event.pyx
sanderland: Already fixed it :)

kivy/tests/test_properties.py
sanderland: No, we raise a ValueError, which I think is correct

matham: Oh, I did not notice this. We definitely should not change the error type from `TypeError` to `ValueError` because this kind of error is supposed to raise a `TypeError`.

sanderland: I see, thanks :)

kivy/tests/test_properties.py
```diff
@@ -1266,3 +1266,64 @@ def callback(obj, val):
...
+
+def test_unknown_property():
+    from kivy.properties import NumericProperty
+
+    class MyWidget(EventDispatcher):
+        width = NumericProperty(0)
+
+    with pytest.raises(TypeError) as cm:
+        MyWidget(width=12, unkn="abc")
+    assert "Unexpected properties ['unkn']" in str(cm.value)
```
matham, line 1279:
```suggestion
    assert 'Properties "[\'unkn\']"' in str(cm.value)
```

kivy/tests/test_properties.py
```diff
@@ -1266,3 +1266,64 @@ def callback(obj, val):
...
+
+    class Widget2(Behavior, EventDispatcher):
+        pass
+
+    class Widget3(EventDispatcher, Behavior):
+        pass
+
+    with pytest.raises(TypeError) as cm:
+        EventDispatcher(name='Pasta')
+    assert "Unexpected properties ['name']" in str(cm.value)
```
matham, line 1297:
```suggestion
    assert 'Properties "[\'name\']"' in str(cm.value)
```

matham (changes requested):
The error message was changed before. This should fix the tests.

sanderland:
> The error message was changed before. This should fix the tests.

Yeah was just waking up and fixing. Maybe check back when tests pass :)

PR #7541, opened 2021-05-26
SoundLoader: Fix play calls not working in ffpyplayer after the first

sanderland:
It seems that calling .play repeatedly on a Sound works in sdl2, but not in ffpyplayer.
I asked about this in the discord, but I think it got lost. Here's an attempt at a fix.
Not sure how you'd test this.

Maintainer merge checklist


matham (approved):
Seems ok

PR #7544, opened 2021-05-28
[WIP] Improvement of the Switch widget to allow better customization of its properties and bug fixing in the animation

DexerBR:
The switch widget is one of the most complicated to customize, among the kivy widgets.
When I started using kivy I had a hard time customizing the switch, so I had decided to use ToogleButton, although it wasn't quite what I wanted.
When searching the internet, I saw several people showing ways to customize the switch, but none of them led to the visual aspect of other custom widgets.
Therefore, I decided to update the Switch in order to allow greater control over its properties and enable a better styling without great efforts.
I believe it is an important reform of the Switch, which currently seems to be a little behind other Kivy widgets.

As for the documentation, I followed the model that was already on the switch. However, I was unable to test the preview on my machine after several attempts to configure my machine to view the .rst files. But I believe that the adjustments are small.

Maintainer merge checklist


kivy/uix/switch.py
```diff
@@ -73,19 +143,188 @@ class Switch(Widget):
...
 
     active_norm_pos = NumericProperty(0)
     '''(internal) Contains the normalized position of the movable element
     inside the switch, in the 0-1 range.
-
     :attr:`active_norm_pos` is a :class:`~kivy.properties.NumericProperty`
     and defaults to 0.
     '''
 
+    switch_background_relative_size_hint_y = NumericProperty(1, allownone=True)
```
matham, line 157: I think the names can be simplified a bit. E.g. for all the new properties, please remove the `switch_` prefix. The class is a switch so there's no need to name the properties with the class name.

Also, you can remove the `relative` part from these properties. So e.g. this would just be `background_size_hint_y`.

Also, I'm not sure that using a size hint for an image is a good idea. We don't generally do this in Kivy because images are generally designed for a fixed size (e.g. the text would look weird if scaled up). So it may be better to not add these size hints.

Also, I saw that you removed some blank lines in the docs, but I'm not sure why? They are needed there to properly divide the docs.

DexerBR: I agree with the names, I will simplify them.

As for the blank lines, I removed them at the beginning of the edition and ended up not realizing that they were necessary when I finished. As I couldn't test it, I didn't see that it would be missed. But I will add them again.

Regarding the size hint, it was a way I found to make the Switch components honor the properties of the Widget class. Because in my view, as all Switch components are images, having a size hint for the Switch widget itself, while its components are not responsive, seems a bit unexpected. What do you think about that?

matham: >Because in my view, as all Switch components are images, having a size hint for the Switch widget itself, while its components are not responsive, seems a bit unexpected. What do you think about that?

I do understand that. But I wonder if anyone uses icon images with a size hint. Personally, all my buttons and icons use a fixed size because making it responsive just didn't look right, unless I switch images depending on the size. So while what you're saying is right, if people don't use size hints, it may not be worth adding support for it, even the widget technically supports it. 

A user could always manually set the image size based on the widget size so it would be too hard for them to make it responsive. E.g.: `button_size: self.size` or something. But I'd be interested to hear if there are uses cases that use size hints for icons.

DexerBR: Really, using size hint to scale icons can be risky.
I remember using size hint on icons when I started using kivy. I defined size_hint_x, for example, and associated the height to it, obeying the actual proportion of the image. It worked fine, but everything changed when I found out that using density-independent pixels would give me more control.

Well, I think it’s better then to remove those size hints.
Soon I will commit with the changes.

DexerBR:
@matham Please take a look at these modifications. I added other useful properties and fixed a bug that prevented the switch from animating correctly when clicked.
I needed to use the on_kv_post event in this approach to solve the bug, I don't know if it's the best or most suitable, as I haven't seen it being used in other widgets. But, without using this event, I was unable to properly load all the properties of the Switch to display it as expected as soon as the kv rules are applied. Without using this approach, the position of the sliding button would only be correct after the first touch.
I found an important fix, because the current Switch looks more like a ToogleButton, with "static" behavior, with the exception of the slide action.

DexerBR:
I would like to mention that now the internal property ``active_norm_pos``  will not vary between 0 and 1 in all cases, it will depend on the ``norm_spacing`` property. For example, if ``norm_spacing = 0.2``, then the range ``of active_norm_pos`` will be ``0.2 - 0.8``.
If a property that always varies in the range ``0 - 1`` is needed (after all we are talking about internal variables), for whatever the purpose (maybe binding), but I don't know if someone will bind something to switch position, I can add it easily. But above all, I await your opinion

With my last commit, it's possible to get a styling of the switch at the level shown below, just by changing the **default** properties.

https://user-images.githubusercontent.com/73297572/120571263-1eb9ea00-c3f0-11eb-97d8-1590190079c4.mp4

Ignore dropped frames due to screen recording

Kulothungan16:
What about using `__init__` method instead of `on_kv_post`?

DexerBR:
> What about using `__init__` method instead of `on_kv_post`?

In my tests, I tried to make a replacement for
```python
def on_kv_post(self, *args):
```
per:
```python
def __init__(self, **kwargs):
      super(Switch, self).__init__(**kwargs)
```

But dependent variables (e.g. `norm_spacing`, `active_norm_pos`) were not updated after instantiation. Maybe there's a better approach, but I haven't been able to come up with one yet. But tips/ideas are welcome.

PR #7549, opened 2021-06-07
MouseMotionEventProvider: Add disable_hover property

pythonic64:
Request adds `disable_hover` property so that dispatching of mouse hover events can be disabled in config or at runtime.

Maintainer merge checklist:

### DIFF
--- a/kivy/input/providers/mouse.py
+++ b/kivy/input/providers/mouse.py
@@ -50,7 +50,7 @@ touch will simulate a multi-touch event. For example::
 
 Provider dispatches hover events by listening to properties/events in
 :class:`~kivy.core.window.Window`. Dispatching can be disabled by setting
-`True` to :attr:`MouseMotionEventProvider.disable_hover` property or by adding
+:attr:`MouseMotionEventProvider.disable_hover` to ``True`` or by adding
 `disable_hover` in the config::
 
     [input]
@@ -194,12 +194,9 @@ class MouseMotionEventProvider(MotionEventProvider):
             self._disable_hover = value
 
     disable_hover = property(_get_disable_hover, _set_disable_hover)
-    '''Disables dispatching of hover events if set to `True`.
+    '''Disables dispatching of hover events if set to ``True`` (default).
 
     .. versionadded:: 2.1.0
-
-    :attr:`disable_hover` is a :class:`property` instance and defaults to
-    `False`.
     '''
 
     def start(self):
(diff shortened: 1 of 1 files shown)


kivy/input/providers/mouse.py
```diff
@@ -46,6 +46,19 @@
     if 'multitouch_sim' in touch.profile:
         touch.multitouch_sim = True
 
+.. versionchanged:: 2.1.0
+
+Provider dispatches hover events by listening to properties/events in
+:class:`~kivy.core.window.Window`. Dispatching can be disabled by setting
+`True` to :attr:`MouseMotionEventProvider.disable_hover` property or by adding
```
matham, line 53:
```suggestion
:attr:`MouseMotionEventProvider.disable_hover` to ``True`` or by adding
```

Switching order, but also, you need to double the backticks for sphinx to display the text as code.

kivy/input/providers/mouse.py
```diff
@@ -159,11 +174,34 @@ def __init__(self, device, args):
...
+                    self._start_hover_events()
+            self._disable_hover = value
+
+    disable_hover = property(_get_disable_hover, _set_disable_hover)
+    '''Disables dispatching of hover events if set to `True`.
+
+    .. versionadded:: 2.1.0
+
+    :attr:`disable_hover` is a :class:`property` instance and defaults to
+    `False`.
```
matham, line 202:
```suggestion
    '''Disables dispatching of hover events if set to ``True`` (default).

    .. versionadded:: 2.1.0
```

There's no need to say it's a `property`, we only do that for kivy style properties because otherwise it's not visible in the docs. But for normal properties, it's not very important.

matham (changes requested):

kivy/input/providers/mouse.py
pythonic64: It is meant as highlight and not as a code. Like in `ModalView` https://github.com/kivy/kivy/blob/master/kivy/uix/modalview.py#L59

kivy/input/providers/mouse.py
pythonic64: Ok 👍

kivy/input/providers/mouse.py
pythonic64: Or like in https://github.com/kivy/kivy/blob/master/kivy/animation.py#L74 . 

So keep \`True\` or use \`\`True\`\`?

matham: Oh, I see. Personally I would still use ` ``True`` ` because it's code (it's capitalized). If you just wanted to highlight it then I'd write `` `true` `` because that's more regular english. But whichever you choose is fine.

pythonic64: In that case I would rather use ` ``True`` ` as that is the exact value.

pythonic64:
@matham I made the changes. You can squash & merge.

matham (approved):

PR #7553, opened 2021-06-13
Simple PoS Data Handling Example

r4d10n:
Improved example based on basic_data.py, suited for PoS GUI....

Maintainer merge checklist

### DIFF
--- a/examples/widgets/recycleview/pos_data.py
+++ b/examples/widgets/recycleview/pos_data.py
@@ -1,0 +1,320 @@
+import datetime, re
+from random import sample, randint
+from string import ascii_lowercase
+
+from kivy.app import App
+from kivy.lang import Builder
+from kivy.uix.boxlayout import BoxLayout
+from kivy.uix.popup import Popup
+from kivy.uix.label import Label
+from kivy.uix.textinput import TextInput
+
+kv = """
+#: import Clock kivy.clock.Clock
+<PopupLabelCell>
+    size_hint: (None, None)
+    height: 30
+    text_size: self.size
+    halign: "right"
+    valign: "middle"
+
+<EditStatePopup>:
+    container: container
+    size_hint: None, None
+    size: 400, 300
+    title_size: 20
+    auto_dismiss: True
+
+    BoxLayout:
+        orientation: "vertical"
+        padding: dp(10)    
+        
+        GridLayout:
+            id: container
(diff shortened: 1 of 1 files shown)


tshirtman:
Not bad! but a couple feedback, in no particular order
- don't leave trailing spaces at the end of lines
- you probably want to make the TextInputs `multiline: False` 
- `size_hint` on the popup buttons is not useful and breaks their apparence
- making the delete button use `on_release` instead of `on_press` avoids the bug of another button briefly showing as selected
- the sort logic is weird, to avoid it i would instead keep the items in a list in no particular order (or order of addition, so orderno of the last one is always the biggest one), and use sorted(items, key=app.sort_order) on the recycleview, so it's a matter of presentation rather than app logic, to avoid the weird multiple calls to self.sort in places.
- you probably want to disable auto_dismiss for the popup, as it's a small one and easy to leave by mistake
- a filter text input at the top could be nice, and thanks to separating the actual items from what is passed to the recycleview, it would be a matter of filtering at the same level as sorting, `data: sorted(item for app.items if search.text in item['customer.text'])` or something like it.
- `select` could add the id of the order to a list of selected items, and an item being selected could change its appearance (color or border or what else).

Anyway, it's nice already, you don't necessarily need to do all of that, although i'd prefer the extra spaces to be removed, and the small glitches (button's size_hint, on_press…) to be resolved at least.

r4d10n:
Hi Gabriel... Much thanks for the detailed review and suggestions... I've given a shot at trying those stuff you suggested... Kindly test and let me know. It's a learning effort ! 

Some issues I noticed: 

* Alternate Row coloring works only for sorting by OrderID. On sorting using other params, order of coloring changes and I am unable to modify widget parameters (bkcolor) for dynamic classes (Row). Any suggestions for modifying dynamic class params ?

* Filtering works almost okay with on_text_validate. For dynamic updating, on_text method does not seem to get invoked on all text changes esp backspace. Even now, some issues while trying to delete items during Filtered view. Is there something wrong with my approach ? 

* Tried Togglebutton / CheckBox instead of Button for Selecting rows. On clicking it, several other rows seems to get selected too. Also background_color property is not being applied on them. Select button fails to work on some random rows. Some presses shows blue colour elsewhere. 

* Ideally Column Headers with click sorting would be nice and aligned.

PR #7555, opened 2021-06-14
Exporting to BytesIO in same fashion as to file

HyTurtle:
Popping the fmt out of the kwargs to pass it on to the save method; which is needed to save to a BytesIO object.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -725,7 +725,7 @@ class Widget(WidgetBase):
 
                 .. versionadded:: 1.11.0
         '''
-        self.export_as_image(*args, **kwargs).save(filename, flipped=False)
+        self.export_as_image(*args, **kwargs).save(filename, fmt=kwargs.pop('fmt','png'), flipped=False)
 
     def export_as_image(self, *args, **kwargs):
         '''Return an core :class:`~kivy.core.image.Image` of the actual


kivy/uix/widget.py
```diff
@@ -725,7 +725,7 @@ def export_to_png(self, filename, *args, **kwargs):
 
                 .. versionadded:: 1.11.0
         '''
-        self.export_as_image(*args, **kwargs).save(filename, flipped=False)
+        self.export_as_image(*args, **kwargs).save(filename, fmt=kwargs.pop('fmt','png'), flipped=False)
```
tshirtman, line 728: I think i would rather put fmt='png' in the method signature, and document that usage in the docstring, especially as export_as_image doesn't need it, so no need to pass it.

I'm not sure why we do it this way for export_as_image with scale, seems like a bad pattern to me.

tshirtman (commented):
That sounds like a useful change indeed, i think the way the the option is exposed could be improved, but aside that, i'd agree to this change.

kivy/uix/widget.py
HyTurtle: Was trying to save the need for documenting, but agree though that this looks out of place.
Would you default to png or None and require the fmt to be stated for BytesIO.
I'll edit - may need to double check valid fmt formats for documenting

tshirtman: I would go for the one that doesn't break previous usage, which i think actually would be `fmt=None`? as currently if you pass a filename but not fmt, the format is going to be deduced from the filename, while if we force fmt='png' by default, then it's going to save as png, even if the filename is `.jpeg`, which would require any such usage to now pass the desired value of `fmt`.

kivy/uix/widget.py
```diff
@@ -696,7 +696,7 @@ def clear_widgets(self, children=None):
         for child in children:
             remove_widget(child)
 
-    def export_to_png(self, filename, *args, **kwargs):
+    def export_to_png(self, filename, format='png', *args, **kwargs):
```
tshirtman, line 699: we usually use fmt to avoid shadowing `format` which is a builtin function (not a big deal, in itself, but it's better to be consistent).

HyTurtle: That's fair, was avoiding fmt=fmt.

kivy/uix/widget.py
HyTurtle: I sort of took the method name literally there, expecting to get a png format back, will change.

PR #7587, opened 2021-07-28
Fix PermissionError when reconnecting mtdev input devices

duelafn:
This is an alternate fix to PR #6339. I encountered and solved this issue independently then when preparing this pull request discovered Cheaterman's #6339 which solves the problem in the same way. I'm not sure how (or the proper etiquette) to update that patch, so am creating a competing PR -- Apologies, if I chose wrong, let me know if there was a better way to submit this.

In any case, Cheaterman is correct that the Linux kernel creates input devices then hands permission changes off to udev. This results in a period of time when the device is readable only by root.

For instance, ls -l in a loop during an unplug / replug (or renegotiation of a flaky USB connection) will show something like:

    crw-rw---- 1 root input 13, 67 Jul 27 15:41 /dev/input/event3
    crw-rw---- 1 root input 13, 67 Jul 27 15:41 /dev/input/event3
    crw-rw---- 1 root input 13, 67 Jul 27 15:41 /dev/input/event3
    crw-rw---- 1 root input 13, 67 Jul 27 15:41 /dev/input/event3
    crw-rw---- 1 root input 13, 67 Jul 27 15:41 /dev/input/event3
    ls: cannot access '/dev/input/event3': No such file or directory
    ls: cannot access '/dev/input/event3': No such file or directory
    ls: cannot access '/dev/input/event3': No such file or directory
    ls: cannot access '/dev/input/event3': No such file or directory
    ls: cannot access '/dev/input/event3': No such file or directory
    crw------- 1 root root 13, 67 Jul 27 15:41 /dev/input/event3
    crw------- 1 root root 13, 67 Jul 27 15:41 /dev/input/event3
    crw------- 1 root root 13, 67 Jul 27 15:41 /dev/input/event3
    crw------- 1 root root 13, 67 Jul 27 15:41 /dev/input/event3
    crw------- 1 root root 13, 67 Jul 27 15:41 /dev/input/event3
    crw-rw---- 1 root input 13, 67 Jul 27 15:41 /dev/input/event3
    crw-rw---- 1 root input 13, 67 Jul 27 15:41 /dev/input/event3
    crw-rw---- 1 root input 13, 67 Jul 27 15:41 /dev/input/event3
    crw-rw---- 1 root input 13, 67 Jul 27 15:41 /dev/input/event3

MTDMotionEventProvider in kivy/input/providers/mtdev.py handles reconnects at line 293 with an infinite spin waiting for the device to return (this code happens after only for a reconnect, so the device was known to be good once).

Device reconnects can be processed by MTDMotionEventProvider faster than udev can get a chance to run, so it can happen that the loop in kivy/input/providers/mtdev.py detects the presence of the device, then when Device in kivy/lib/mtdev.py attempts to connect, it receives PermissionError.

Like Cheaterman, I put the spinning code in the Device object to be applied all the time, both the initial connection and any reconnect attempts. I did this because it is possible that the initial connection attempt happens when the permissions are not yet set. My patch includes a timeout so that a PermissionError can still be raised in the case where the system is actually misconfigured (e.g., user is not a member of the input group in the example above).

As an alternative, one could put an infinite spin in kivy/input/providers/mtdev.py after line 293, but that wouldn't address the race possibility at initial connection. One could also pass a parameter that alters the spin duration depending on whether we are making an initial or repeat connection. I can update this patch to any of those approaches if you like.

In my experiments, udev can take quite a while to update permissions. It takes about 0.6s to correct the permissions on a Raspberry Pi 4. Therefore, I set the spin timeout to 3 seconds which should cover all reasonable systems.

Maintainer merge checklist

### DIFF
--- a/kivy/lib/mtdev.py
+++ b/kivy/lib/mtdev.py
@@ -21,6 +21,7 @@ documentation for further details.
 # flake8: noqa
 
 import os
+import time
 from ctypes import cdll, Structure, c_ulong, c_int, c_ushort, \
                    c_void_p, pointer, POINTER, byref
 
@@ -141,7 +142,22 @@ class Device:
         self._fd = -1
         self._device = mtdev()
 
-        self._fd = os.open(filename, os.O_NONBLOCK | os.O_RDONLY)
+        # Linux kernel creates input devices then hands permission changes
+        # off to udev. This results in a period of time when the device is
+        # readable only by root. Device reconnects can be processed by
+        # MTDMotionEventProvider faster than udev can get a chance to run,
+        # so we spin for a period of time to allow udev to fix permissions.
+        # We limit the loop time in case the system is misconfigured and
+        # the user really does not (and will not) have permission to access
+        # the device.
+        # Note: udev takes about 0.6 s on a Raspberry Pi 4
+        permission_wait_until = time.time() + 3.0
+        while self._fd == -1:
+            try:
+                self._fd = os.open(filename, os.O_NONBLOCK | os.O_RDONLY)
+            except PermissionError:
+                if time.time() > permission_wait_until:
+                    raise
         ret = mtdev_open(pointer(self._device), self._fd)
         if ret != 0:
             os.close(self._fd)


tshirtman:
Oopsy, indeed we apparently let that PR slip, sorry you had to duplicate the work, i think i'll prefer yours because it contains a timeout and the nice comment about why the code does that. So sorry about letting @Cheaterman's PR rot, thanks to both of you!

akshayaurora (approved):

PR #7594, opened 2021-08-10
Typo on refreshing process.

AnderSchofield:
Guys! On line 63, I've noticed that _Clock is schedule_once, which blocks the kv file refreshing._
Probably a typo, since there's an interval set on args.
Sorry If I misunderstood anything!
Keep up the excellent work!
Best regards.

Maintainer merge checklist

### DIFF
--- a/kivy/tools/kviewer.py
+++ b/kivy/tools/kviewer.py
@@ -60,7 +60,7 @@ class KvViewerApp(App):
         o = Observer()
         o.schedule(KvHandler(self.update, TARGET), PATH)
         o.start()
-        Clock.schedule_once(self.update, 1)
+        Clock.schedule_interval(self.update, 1)
         return super(KvViewerApp, self).build()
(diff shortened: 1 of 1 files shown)


tshirtman:
No, it shouldn't be needed, because the watchdog module will call refresh every time the file changes, so no need to use a schedule_interval here, unless watchdog doesn't work. Does this change fixes something for you?

AnderSchofield:
BTW, I've done some research, and looks like there’s some current issues with Watchdog. Not certain if  it should affect this script's functionality, JPEG.

AnderSchofield:
@tshirtman Hello there! Sorry for bothering! I just wasn't able to confirm if thr problem was indeed watch dog, or my system. (Without the change, It doesn't refresh). We're you able to check? 
If it's my system, I can close the thread!
Thanks in advance!

Issue #7381, opened 2021-02-16
TextField is not scrollable on android. It selects text


kengoon:
**Software Versions**
* Python:3.8
* OS:android
* Kivy:master
* Kivy installation method:buildozer

**Describe the bug**
 whenever i type in series of character that exceed the height of a `textfield` and i want to scroll through it on my android phone, instead of scrolling, it selects the text

njk639:
The default behavior for a text input in Kivy doesn't include scrolling. You could try to implement a solution using a ScrollView and text input box together. Here's a discussion on StackExchange that I found. I haven't tested rhe code, but in theory the approach makes sense to me. https://stackoverflow.com/questions/14117317/kivy-scrollview-implementation-in-textinput

Alspb:
ScrollView solution doesn't work properly, please find the details here: https://github.com/kivy/kivy/issues/7320

njk639:
> ScrollView solution doesn't work properly, please find the details here: #7320
Hmm. I see. Does this happen both when the font size is small and when large relative to the overall text field size? Does it happen if you overwrite the default on_double_tap method to pass?

Neither are a solution, but the might help narrow down one. I'm hoping they will help answer if it 1) behaves properly when the initial tap is on white space, which might be more feasible at different text sizes or 2) if this is a bug with double-tap detection on Android.

misl6:
Fixed via #7610

Issue #7320, opened 2021-01-09
Using swipes in TextInput to scroll on Android


Alspb:
**Feature request**

Currently in TextInput swipes (i.e. on_touch_move) are used to select text. That's usual and convenient for desktop OS, but on Andorid swipes are typicaly used to scroll text (up/down - for multiline, left/right - for single-line).
Additionally, there is only one way to scroll text on Android (without selecting it) - by moving middle handle, and this option is not really convenient.

So it seems reasonable to use swipes in TextInput for scrolling if use_handles is True, and for text selection if use_handles is False (because without handles one need some way to select text).
It'd also be great to add scroll bar and some scrolling effects, but it's not too important.

There is a way to use ScrollView:
```
ScrollView:
    id: scrlv
    do_scroll_x: False
    do_scroll_y: True
    TextInput:
        size_hint: 1, None
        use_handles: True
        height: max(self.minimum_height, scrlv.height)
```
But there are several problems here:
1) On Android app stops responding for a really huge text (like '\n'.join(['abcd{}'.format(i) for i in range(10000)])).
2) Moving cursor outside of the visible area (with keyboard arrows or middle handle) doesn't perform scrolling to the cursor.

Alspb:
ScrollView solution above doesn't work because of the huge TextInput height (which causes the Memory error even on desktop and also problems with scrolling to cursor).
Haven't found a way to deal with this problem yet. 

Также лучше дописать, что вариант со ScrollView при большем объеме текста на Linux тоже ломается - MemoryError. Дело в большой height (которая определяется по minimum_height). При ряд методов TextInput также использует self.height, из-за чего курсор и уходит из области видимости (предположительно проблема в методе _set_cursor).

misl6:
Fixed via #7610

PR #7610, opened 2021-09-03
Added scroll from swipe feature in TextInput

DexerBR:
Currently TextInput does not have the scroll-from-swipe behavior. This makes it difficult to use on mobile devices, in which the main interaction interface is touch.

It is possible to scroll horizontally and vertically. I didn't try to implement velocity features, such as ScrollView, but this can be implemented in the future

The implemented behavior is: if after starting the touch the user doesn't swipe the scroll_distance inside the scroll_timeout, the mode will be changed, from scroll to selection.

Implementation of the feature request: #7381 and #7320

</br>

TODO:
* [x] Keep the selection when scrolling from swipe

</br>

Maintainer merge checklist

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -211,8 +211,12 @@ _is_osx = sys.platform == 'darwin'
 
 # When we are generating documentation, Config doesn't exist
 _is_desktop = False
+_scroll_timeout = _scroll_distance = 0
 if Config:
     _is_desktop = Config.getboolean('kivy', 'desktop')
+    _scroll_timeout = Config.getint('widgets', 'scroll_timeout')
+    _scroll_distance = '{}sp'.format(Config.getint('widgets',
+                                                   'scroll_distance'))
 
 # register an observer to clear the textinput cache when OpenGL will reload
 if 'KIVY_DOC' not in environ:
@@ -1445,9 +1449,15 @@ class TextInput(FocusBehavior, Widget):
     def long_touch(self, dt):
         self._long_touch_ev = None
         if self._selection_to == self._selection_from:
-            pos = self.to_local(*self._long_touch_pos, relative=False)
+            pos = self.to_local(*self._touch_down.pos, relative=False)
             self._show_cut_copy_paste(
                 pos, EventLoop.window, mode='paste')
+
+    def cancel_long_touch_event(self):
+        # schedule long touch for paste
+        if self._long_touch_ev is not None:
+            self._long_touch_ev.cancel()
+            self._long_touch_ev = None
 
     def on_double_tap(self):
         '''This event is dispatched when a double tap happens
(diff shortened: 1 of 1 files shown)


DexerBR:
An update of the current state of development.

https://user-images.githubusercontent.com/73297572/132382151-40c37198-073a-4191-950b-5095c1d22467.mp4

kivy/uix/textinput.py
```diff
@@ -1562,13 +1565,61 @@ def on_touch_move(self, touch):
...
+            if self.multiline:
+                _scroll_timeout = touch.time_update - touch.time_start
+                self._scroll_distance_x += abs(touch.dx)
+                self._scroll_distance_y += abs(touch.dy)
+
+                # disable scroll and start selection mode if scroll distance
+                # isn't reached within scroll_timeout
+                if _scroll_timeout >= self.scroll_timeout/1000 and\
+                   self._scroll_distance_x <= self.scroll_distance and\
+                   self._scroll_distance_y <= self.scroll_distance:
```
tshirtman, line 1580: this indent is not obvious enough, one way to make it more apparent is to parenthesize the expression like this
```suggestion
                if (
                    _scroll_timeout >= self.scroll_timeout/1000
                    and self._scroll_distance_x <= self.scroll_distance
                    and self._scroll_distance_y <= self.scroll_distance
                ):
```

kivy/uix/textinput.py
```diff
@@ -1562,13 +1565,61 @@ def on_touch_move(self, touch):
...
+            # multiline = False
+            else:
+                _scroll_timeout = touch.time_update - touch.time_start
+                self._scroll_distance_x += abs(touch.dx)
+
+                # disable scroll and start selection mode if scroll distance
+                # isn't reached within scroll_timeout
+                if _scroll_timeout >= self.scroll_timeout/1000 and\
+                   self._scroll_distance_x <= self.scroll_distance:
+                    self._enable_scroll = False
```
tshirtman, line 1599: similarly 
```suggestion
                if (
                    _scroll_timeout >= self.scroll_timeout/1000
                    and self._scroll_distance_x <= self.scroll_distance
                ):
                    self._enable_scroll = False
```

tshirtman (changes requested):

tshirtman:
That's a cool change, tested on desktop only for now, by editing the showcase code to enable the feature, but works well here, including selection on hold.

DexerBR:
I've tested it on desktop and mobile and it seems to be working fine, but I'm awaiting your review and testing to see the need for changes.

kivy/uix/textinput.py
```diff
@@ -1574,22 +1600,42 @@ def on_touch_up(self, touch):
...
-        if self._long_touch_ev is not None:
-            self._long_touch_ev.cancel()
-            self._long_touch_ev = None
+        self.cancel_long_touch_event()
 
         if not self.focus:
             return False
 
+        _scroll_timeout = touch.time_update - touch.time_start
+        any_condition_true = any((
```
tshirtman, line 1609: there are not very descriptive names, and there is no comment to explain the logic, i don't think extracting them from the conditions clarifies it much, also although it's a minor point, using any/all here is less efficient than using `and` or `or`, as it'll need to evaluate each expression before passing them to the function call, instead of stopping at the first value allowing to determine the expression (in an `and` the first falsy value, in an `any` the first truthy one), and function calls in python are not free, the performances are not a big deal there, but i find it less idiomatic than just using the python keywords.

tshirtman (approved):
LGTM, little feedback, but no deal breakers imho. Good job.

kivy/uix/textinput.py
DexerBR: I agree with you, I believe I focused more on making the feature work than on the aspects you mentioned.  😂
I will modify

tshirtman (approved):
Looks good to me!

PR #7614, opened 2021-09-05
Fix install command for zsh

wxllow:
`python -m pip install kivy[base] kivy_examples --no-binary kivy` throws the error `zsh: no matches found: kivy[base]`, you need to use `python -m pip install "kivy[base]" kivy_examples --no-binary kivy` instead

Maintainer merge checklist


Zen-CODE (approved):
Still works in  bash, so why not. Thanks  +1

PR #5576, opened 2018-01-15
window_x11: implement get_window_info()

jakogut:

PR #7619, opened 2021-09-10
Core: Window: Fix for a broken on Windows Platform get_window_info() call

FI-Mihej:
An initial commits set with the `get_window_info()` implementation states: "The implementation
only returns information for Linux (X11/Wayland) and **Windows**". At the same time an implementation of get_window_info() initially had a bug: forgotten `return` for a Windows platform. As result this functionality is not working on Windows.

This fix turnes on initially planned functionality for Windows platform (tested). This is necessary for a platform depend manipulations with main kivy window by getting it's HWND from `Window.get_window_info().window` property.

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -413,6 +413,7 @@ cdef class _WindowSDL2Storage:
                 windows_info = WindowInfoWindows()
                 windows_info.window = wm_info.info.win.window
                 windows_info.hdc = wm_info.info.win.hdc
+                return windows_info
 
     # Transparent Window background
     def is_window_shaped(self):


FI-Mihej:
An initial implementation PR: https://github.com/kivy/kivy/pull/5576

FI-Mihej:
Commit with an initial implementation: https://github.com/kivy/kivy/commit/d1ebb76343ca3ce306901f7af4e403fcab7b9f6e

tito (approved):

kivy/core/window/_window_sdl2.pyx
```diff
@@ -413,6 +413,7 @@ cdef class _WindowSDL2Storage:
                 windows_info = WindowInfoWindows()
                 windows_info.window = wm_info.info.win.window
                 windows_info.hdc = wm_info.info.win.hdc
+                return windows_info
```
tito, line 416: Hi @FI-Mihej , it look like the tests written by the previous contributor was not executed, but with your fixes, now they are :)
And they break. Would you mind give a shot and fixes them too ?

I think just changing https://github.com/kivy/kivy/blame/c28c47b39ae57c97c70cc0398d74774d73a6894b/kivy/tests/test_window_info.py#L28 -> using lowercase attribute should work, as the definition of theses variables are lowercase. https://github.com/kivy/kivy/blob/c28c47b39ae57c97c70cc0398d74774d73a6894b/kivy/core/window/window_info.pxd#L18

Thanks,

tito (changes requested):

PR #2959, opened 2015-01-31
Simplify pip install (resolves kivy/kivy#2958)

Samureus:
Resolves kivy/kivy#2958

PR #7623, opened 2021-09-13
Sphinx: Use class instead of instance in add_lexer + Fixes search on sphinx>1.7.9

misl6:
Maintainer merge checklist

Since Sphinx-2.1.0 (released Jun 02, 2019) ``Sphinx.add_lexer()`` takes a Lexer class instead of instance.
An instance of lexers was still supported until Sphinx-3.x.

### DIFF
--- a/doc/doc-requirements.txt
+++ b/doc/doc-requirements.txt
@@ -1,6 +1,5 @@
 Cython>=0.24
 # Frozen Sphinx requirements for easier pip installation
-sphinx==1.7.9
 sphinxcontrib-actdiag
 sphinxcontrib-blockdiag
 sphinxcontrib-nwdiag
--- a/doc/sources/.templates/layout.html
+++ b/doc/sources/.templates/layout.html
@@ -106,12 +106,14 @@
 
       };
     </script>
+{%- block scripts %}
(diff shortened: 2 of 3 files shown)


misl6:
Ok, missed this one: https://github.com/kivy/kivy/pull/6352
Investigating about it.

misl6:
Ok, the issue related to `sphinx>1.7.9` is now fixed.

doc/doc-requirements.txt
```diff
@@ -1,6 +1,5 @@
 Cython>=0.24
 # Frozen Sphinx requirements for easier pip installation
-sphinx==1.7.9
```
tshirtman, line 3: if we don't need this line, maybe also remove the commented line before as well?

tshirtman (approved):

doc/doc-requirements.txt
misl6: I initially thought the same when removing that line, but looks like that comment was inserted in https://github.com/kivy/kivy/pull/2959/files so it doesn't belong to the line just removed

tshirtman: Ah indeed, well spotted.

PR #7629, opened 2021-09-20
Remove key errors in _widget_destructors when finishing app

LESSSE:
Maintainer merge checklist

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -261,8 +261,7 @@ def _widget_destructor(uid, r):
     # Internal method called when a widget is deleted from memory. the only
     # thing we remember about it is its uid. Clear all the associated callbacks
     # created in kv language.
-    if uid in _widget_destructors:
-      del _widget_destructors[uid]
+    _widget_destructors.pop(uid, None)
     Builder.unbind_widget(uid)
 
 


kivy/uix/widget.py
```diff
@@ -261,7 +261,8 @@ def _widget_destructor(uid, r):
     # Internal method called when a widget is deleted from memory. the only
     # thing we remember about it is its uid. Clear all the associated callbacks
     # created in kv language.
-    del _widget_destructors[uid]
+    if uid in _widget_destructors:
```
pythonic64, line 264: You can do this in one line: `_widget_destructors.pop(uid, None)`

p0lygun: @LESSSE pls do the requested changes so that PR can be merged

LESSSE: Done, Thanks

p0lygun (commented):
LGTM!

pythonic64:
Was this part of the code raising an error and if so do you have the code to reproduce that error?
My concern is that we may hide the actual issue instead of fixing it.

p0lygun:
@pythonic64 it is also happening in my project too, i'll try to make a minimal runnable

LESSSE:
The error log example:

> Exception ignored in: functools.partial(<function _widget_destructor at 0x7ff2cf053a60>, 8952)
   File "/home/lessse/Documents/facing/repos/deadpan_prototype/venv/lib/python3.8/site-packages/kivy/uix/widget.py", line 264, in _widget_destructor
     del _widget_destructors[uid]
 KeyError: 8952

This happens in a project using settings (SettingsWithSidebar). If during the app execution I open settings page, this error will spawn after closing the window. I suppose any small example using a settings page should show the error.

p0lygun:
ooof forgot about this, I had a minimal runnable just forgot to update this issue 
Minimal runnable 
```py
import json
from kivy.app import App
from kivy.config import ConfigParser
from kivy.lang import Builder
from kivy.uix.settings import SettingsWithSidebar, Settings

settings_json = [
    {
        "type":"numeric",
        "title":"test",
        "key":"test",
        "section":"test"
    }
]

kv = """
Button:
    text:'Open settings'
    on_release:app.open_settings()
"""


class Test(App):
    def build(self):
        self.settings_cls = SettingsWithSidebar
        self.config.read('test.ini')
        self.config.write()

    def build_settings(self, settings: Settings):
        settings.add_json_panel(
            "App Settings",
            self.config,
            data=json.dumps(settings_json)
        )

    def build_config(self, config: ConfigParser):
        config.read('test.ini')
        config.adddefaultsection('test')
        config.setdefaults('test', {'test': 0})
        config.write()


if __name__ == '__main__':
    Test().run()
```
The error occurs the the app is closed

matham:
Ok, this is actually a bug in kivy, where we seemingly overwrite the widget uid in `label = SettingSidebarLabel(text=name, uid=uid, menu=self)` so we end up creating multiple widgets with the same uid. We need to fix the underlying issue.

matham:
We may want to consider disabling setting the `uid` as a property, except with a specific method so it's not accidentally set. That would have prevented the issue as the current code would raise an exception rather than quietly accepting a custom uid.

Techmartian:
I am facing the same issue when I click logout button from a different screen, and call login screen. For me I was looking for the equivalent of PHP where you destroy session after logout button is clicked. 

So, anyone here has a solution for that, kindly share with me as this has given me sleepless nights trying to figure it out but so far I haven't figured it out yet. Am using Python  3.12.0, Kivy 2.3.0, and Kivymd 1.1.0

PR #7631, opened 2021-09-20
Fixes some bugs in the TextInput if the text is right-aligned or center-aligned and not multiline.

DexerBR:
Master's TextInput has a problem that makes it practically unusable when `multiline` is set to `False` and `halign` is set to `center` or `right`. The behavior is most noticeable when the text is larger than the viewport and occur when clicking, typing, or deleting text.

Below is an example of current bugs

https://user-images.githubusercontent.com/73297572/134050738-e66f9e91-641d-4013-b2b8-f8522f486503.mp4

https://user-images.githubusercontent.com/73297572/134050794-87f63ea7-87a7-4a95-a622-64312e1284de.mp4

This PR solves these problems, almost entirely, with the exception that when starting the deletion of characters from a text larger than the viewport, small spaces appear on the left side (`halign=right`) and on both sides (`halign=center`).

Below is the demonstration of the behavior I mentioned

https://user-images.githubusercontent.com/73297572/134051264-0a2cc26c-9c2d-4724-a3f3-4290207a5b37.mp4

https://user-images.githubusercontent.com/73297572/134051288-556dedcd-0b53-4d76-8c14-ac1f3da997bd.mp4

</br>

Maintainer merge checklist


Alspb:
You're right, seems to be a separate bug. In addition to your description I'd add that:

1. When several characters are deleted simultaneously (using _delete_selection_) these spaces become really large, so the widget looks ugly.
2. The problem also exists for hailng='left', but in that case the spaces appear on the right side.
3. The bug exists on both stable and master versions.

I've found a solution, the idea is to add maximal values for scroll_x and scroll_y:
```python
class TextInput(FocusBehavior, Widget):
    ...
    def __init__(self, **kwargs):
        ...
        fbind('scroll_max_x', self.adjust_scroll)
        fbind('scroll_x', self.adjust_scroll)
        fbind('scroll_max_y', self.adjust_scroll)
        fbind('scroll_y', self.adjust_scroll)
    ...
    def _get_scroll_max_x(self):
        xoffset_max = max([self._get_text_width(line, self.tab_width, self._label_cached) for line in self._lines])
        viewport_width = self.width - self.padding[0] - self.padding[2]
        return max(xoffset_max - viewport_width, 0)
    scroll_max_x = AliasProperty(_get_scroll_max_x, None, bind=('height', 'width', 'padding', '_lines', 'line_height', 'line_spacing'))
    def _get_scroll_max_y(self):
        viewport_height = self.height - self.padding[1] - self.padding[3]
        text_height = len(self._lines) * (self.line_height + self.line_spacing)
        return max(text_height - viewport_height, 0)
    scroll_max_y = AliasProperty(_get_scroll_max_y, None, bind=('height', 'width', 'padding', '_lines', 'line_height', 'line_spacing'))
    def adjust_scroll(self, *largs):
        self.scroll_x = min(self.scroll_x, self.scroll_max_x)
        self.scroll_y = min(self.scroll_y, self.scroll_max_y)
```
You should also remove your __draw_line_ edits, the above is sufficient.

Please double check whether it works!

Note: I've been testing with #7641 fixes applied, but it doesn't seem to be significant.

Alspb:
A bit of an off topic:
At first I tried to fix the bug by modifying the __draw_line_ method and found that I don't really understand this part of the code:
```python
if texture_width * (1 - tcx) < viewport_width:
    tcw = tcw - tcx
    texture_width = tcw * texture_width
elif viewport_width < texture_width:
    tcw = (viewport_width / texture_width) * tcw
    texture_width = viewport_width
```
What does the first condition mean? Isn't the second (elif) condition always True when the first is False? 
Seems that this code works fine, but probably it could be refactored a bit.

DexerBR:
I believe my recent commits will definitely fix the bug. 🙂

misl6 (changes requested):
- Code looks good to me. (I just left a minor suggestion)
- Tested via `KivyCatalog` and looks fine.

Thank you!

misl6 (approved):
Thank you! LGTM.

Issue #7246, opened 2020-12-09
TextInput: app crashes when selecting using handles after deleting some text


Alspb:
**Software Versions**
* Python: 3.8.5
* OS: Linux Mint 20 (based on Ubuntu 20.04 LTS) or Android 10
* Kivy: 2.0.0rc4

**Describe the bug**
App crashes when the following actions are performed in the TextInput (described in the code below):

1. Select 'aaaa bbbb' substring (by double tap and then using selection handles), then delete it.
2. Select 'cccc' word (by double tap).
3. Pull the right selection handle right.

The traceback is placed below.

**Code**
```python
from kivy.app import App
from kivy.uix.textinput import TextInput
class Sample(App):
    def build(self):
        return TextInput(text='aaaa bbbb cccc '+'d'*200, use_handles=True)
Sample().run()
```

**Traceback**
```
Traceback (most recent call last):
   File "/usr/lib/python3.8/pdb.py", line 1704, in main
     pdb._runscript(mainpyfile)
   File "/usr/lib/python3.8/pdb.py", line 1573, in _runscript
     self.run(statement)
   File "/usr/lib/python3.8/bdb.py", line 580, in run
     exec(cmd, globals, locals)
   File "<string>", line 1, in <module>
   File "/home/al/Desktop/KivyBugs/bug6.py", line 1, in <module>
     from kivy.app import App
   File "/home/al/kivy_venv/lib/python3.8/site-packages/kivy/app.py", line 950, in run
     runTouchApp()
   File "/home/al/kivy_venv/lib/python3.8/site-packages/kivy/base.py", line 573, in runTouchApp
     EventLoop.mainloop()
   File "/home/al/kivy_venv/lib/python3.8/site-packages/kivy/base.py", line 347, in mainloop
     self.idle()
   File "/home/al/kivy_venv/lib/python3.8/site-packages/kivy/base.py", line 397, in idle
     Clock.tick_draw()
   File "/home/al/kivy_venv/lib/python3.8/site-packages/kivy/clock.py", line 783, in tick_draw
     self._process_events_before_frame()
   File "kivy/_clock.pyx", line 658, in kivy._clock.CyClockBase._process_events_before_frame
   File "kivy/_clock.pyx", line 701, in kivy._clock.CyClockBase._process_events_before_frame
   File "kivy/_clock.pyx", line 697, in kivy._clock.CyClockBase._process_events_before_frame
[...]

PR #7621, opened 2021-09-13
Fixed inconsistent behavior of TextInput bubble and handles

DexerBR:
The entirety of the bugs resolution is conditioned to the approval of #7618

This PR resolves several bugs that appear when using handles and bubble to select/copy/paste/cut text. Fixes: #7247, #7396

Fixed bugs:
- Selection is not updated even after moving handles
- In multiline mode, if you move the handles beyond the top and bottom edges, texts beyond those lines cannot be selected, which results in the selection limited to the viewport.
- The cut/copy/paste bubble is not shown after the handles' on_release event
- Bubbles are not displayed relative to textinput/handles position, after correcting the above error
- Bug that did not limit selection to the current line, while using handles and in multiline mode

Now the cursor follows the last handle dragged

</br>

Maintainer merge checklist
* [x] Title is descriptive/clear for inclusion in release notes.
* [x] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [x] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #7632, opened 2021-09-21
Fixed enormously large cursor_x returned by get_cursor_from_xy when touch is righter then the last symbol of the line.

Alspb:
Fixed enormously large cursor_x returned by get_cursor_from_xy when touch is righter then the last symbol of the line
Fixed selecting using handles bug (#7246). More accurately, this bug was already fixed, and here the fix is modified a bit.

Maintainer merge checklist


DexerBR:
Could you demonstrate the bug through some example? I tested the master and couldn't reproduce the cited bug.
As for selection using handles, the bug that didn't limit selection to the current cursor line, returning high values, has been fixed in #7621. I believe I have also fixed the bug reported in #7246, although I couldn't reproduce the bug.
Looking at your modifications, although I'm not sure, I believe they need to be revised as they seem to be potential to break something else.

Alspb:
Indeed, currently for sufficiently large touch.x _get_cursor_from_xy_ returns cursor_x = `int(x - self.x)`, which is incorrect. I've just fixed it.
As for #7246, I suppose the crash was caused by non-int cursor_x returned by get_cursor_from_xy, so it was fixed in [7449](https://github.com/kivy/kivy/pull/7449/) (by replacing `return cx, cy` with `return int(cursor_x), int(cursor_y)`). Here I just slightly modified this fix - now cursor_x is always int before return statement is executed, so I removed the excessive conversion to int.

akshayaurora (approved):

akshayaurora:
@Alspb thanks for the fix, apologies for the late merge.

PR #7642, opened 2021-09-25
TextInput loading time optimisation for large texts

Alspb:
TextInput with large text (e.g. 'a' * 100000) loads faster after the fix.
The idea is to save symbols/words widths returned by __get_text_width_ into a local dictionary inside __split_smart_ method to prevent subsequent __get_text_width_ calls for the same symbol/word.
Currently __get_text_width_ results are already cached, but calling __get_text_width_ itself leads to unnecessary operations (calling __get_line_options_, converting its results to strings, etc.), which significantly slows __split_smart_ down in case of large text.
At the same time it seems that for all the other TextInput methods current caching is sufficient.

Maintainer merge checklist


akshayaurora:
@Alspb nice work! would you please post some benchmarks. https://github.com/kivy/kivy/blob/master/kivy/tests/perf_test_textinput.py

Would be nice to test the difference in perf before and after this test

akshayaurora (approved):

akshayaurora:
Before:

```
Test: load_large_text
=====================
loading uix/textinput.py....
putting text in textinput
mem usage after test
198528.0 MB
------------------------------------------
Loaded 4004 lines 0.13971200000000294 secs
------------------------------------------
=====================
Test: stress_insert
=====================
Done!
pasted 4004 lines 9 times
mem usage after test
200688.0 MB
total lines in text input: 40031
--------------------------------------
total time elapsed: 1.042116667000002
--------------------------------------
=====================
Test: stress_del
=====================
Done!
deleted 210 characters 9 times
mem usage after test
203408.0 MB
total lines in text input: 39954
--------------------------------------
total time elapsed: 0.450933292000002
--------------------------------------
=====================
Test: stress_selection
=====================
Done!
mem usage after test
203408.0 MB
```

After

```
=====================
Test: load_large_text
=====================
loading uix/textinput.py....
putting text in textinput
mem usage after test
109152.0 MB
------------------------------------------
Loaded 4011 lines 0.1250300000000002 secs
------------------------------------------
=====================
Test: stress_insert
=====================
[INFO   ] [Clipboard   ] Provider: sdl2(['clipboard_nspaste'] ignored)
Done!
pasted 4011 lines 9 times
mem usage after test
176592.0 MB
total lines in text input: 40101
--------------------------------------
total time elapsed: 0.6837495839999996
--------------------------------------
=====================
Test: stress_del
=====================
Done!
deleted 210 characters 9 times
mem usage after test
189072.0 MB
total lines in text input: 40024
--------------------------------------
total time elapsed: 0.4659492079999996
--------------------------------------
=====================
Test: stress_selection
=====================
Done!
mem usage after test
189072.0 MB

```

That's a big diff for the text insertion, nice job @Alspb

tshirtman:
Interresting, actually maybe could reap even more benefit by decorating `_get_text_width` with `@lru_cache` (from functools). so the caching would work even accross multiple layouting, and but wouldn't grow forever, because of the size of the cache (https://docs.python.org/3/library/functools.html#functools.lru_cache).

tshirtman:
Reading the file is actually pretty fast, if i move that to a thread, 
```python
from threading import Thread
import time

import kivy
from kivy.app import App
from kivy.clock import mainthread


class MyApp(App):

    def build(self):
        self.count = 0
        layout = BoxLayout()

        # add the label
        self.label1 = Label(text='0')
        layout.add_widget(self.label1)

        # add the button
        layout.add_widget( Button(text='Increment', on_release=self.increment) )

        # add the textinput
        self.textinput1 = TextInput(text='Click button to load file.')
        layout.add_widget(self.textinput1)

        # add the button
        layout.add_widget( Button(text='Load File', on_release=self.load_file) )

        return layout

    def load_file(self, *args):
        Thread(target=self._load_file_thread).start()

    @mainthread
    def _set_text(self, text):
        start = time.time()
        self.textinput1.text = text
        end = time.time()
        print("finished displaying text in " + str(end-start) + " seconds")

    def _load_file_thread(self):
        start = time.time()
        print("loading file")
        with open( 'big_file.txt' ) as big_file:
            text = big_file.read()
        end = time.time()
        print("finished loading file in " + str(end-start) + " seconds")
        self._set_text(text)

    def increment(self, *args):
        self.count+=1
        self.label1.text = str(self.count)


if __name__ == "__main__":
    MyApp().run()
```
i get:
```
loading file
finished loading file in 0.07073855400085449 seconds
finished displaying text in 18.885766983032227 seconds
```

So the issue is in laying out the text
running with `py-spy top -- python main.py` and click the "load" button asap, i get as a final result.
```
  %Own   %Total  OwnTime  TotalTime  Function (filename)                                                                                                                                                                                                                                                                       
  0.00%   0.00%    3.53s    16.71s   _create_line_label (kivy/uix/textinput.py)
  0.00%   0.00%    3.10s    22.70s   _refresh_text (kivy/uix/textinput.py)
  0.00%   0.00%    2.75s    10.12s   refresh (kivy/core/text/__init__.py)
  0.00%   0.00%    2.23s     6.13s   render (kivy/core/text/__init__.py)
  0.00%   0.00%    1.47s     2.32s   _get_font_id (kivy/core/text/text_sdl2.py)
  0.00%   0.00%    1.39s     3.71s   get_extents (kivy/core/text/text_sdl2.py)
  0.00%   0.00%    1.22s     1.64s   __init__ (kivy/core/text/__init__.py)
  0.00%   0.00%    1.11s     1.11s   __init__ (kivy/weakmethod.py)
```
(or this flamegraph, with `py-spy record -- python main.py`)
![flamegraph](https://user-images.githubusercontent.com/22759/178367126-4fc785c5-0b87-455d-8baa-089669e48a45.svg)

so almost all the time is spent in `_refresh_text` unsurprisingly, and `_create_line_label` in it.

My guess, looking at the code from `_create_line_label` is that most of the time is spent looking for the ideal length of the line, while we start with the logical line length if it's too long, we cut it in half, until it fits and then grows again (half the length) until it does't fit anymore, etc, following a bisection until the stop is less than 2. Any improvement to the initial guess of the line's length could make a significant difference, for example, just by enlarging the window of the example, so most logical lines fit in one display line, the rendering time was cut by more than 50%.

Another idea could be to be able to compute how many words of the line can fit, without rendering the full line, that could be achieved by rendering + caching each unique word encountered, with line options as well (font, etc), so the words texture could be looked up from cache, adding their width until we don't fit the current line anymore, allowing us to either render the line from these words at this point, or reworking the rendering system so a line can just use these textures directly through multiple rectangle instructions, in this technique it would be important to properly keep track of the space between words as well, which depends on the specific character(s) used.

I'm sure there is research about how to do these things efficiently as well, that could be worth looking into (famously, Knuth spent a lot of time on efficient and correct ways to layout text).

maltfield:
Very interesting! My actual use-case is to display a log file in my app, which is rendered with a monospaced font. So I guess such optimizations would work especially well in my use case.

I'm not the only one who tried to do this with Kivy

 * https://groups.google.com/g/kivy-users/c/k1E379mME9U

I ended-up using a RecycleView as suggested [here[(https://stackoverflow.com/a/72923330). The consequence is that I user's can't select bits of the text (eg spanning multiple lines). And in RecycleView, very long lines turn to a solid black line.

My actual implementation is here:

 * https://github.com/BusKill/buskill-app/blob/d30d27264a26031844be76b4ad2ea7a9267d15b5/src/buskill_gui.py#L430-L531
 * https://github.com/BusKill/buskill-app/blob/d30d27264a26031844be76b4ad2ea7a9267d15b5/src/buskill.kv#L230-L322

Self-contained app builds of the above implementation are here:

 * Linux AppImage https://github.com/BusKill/buskill-app/releases/tag/2651008107_linux
 * MacOS .dmg https://github.com/BusKill/buskill-app/releases/tag/2651008107_mac
 * Windows .exe https://github.com/BusKill/buskill-app/releases/tag/2651008107_windows

But in the long-term, such optimizations would really be appreciated so I could do this in an actual TextInput. It's kinda silly that you can't make a decent text editor with Kivy that can handle files a few MB large

 * https://github.com/duranbe/Kedit

tshirtman:
Yes, monospace fonts are indeed a special case allowing for much faster computation of these, sadly the TextInput doesn’t currently account for them, or, (as far as i know), have an easy way to do so, but if we could detect this case and have specific implementations of the intensive parts, triggered in this case, this could be a big win.

Indeed RecycleView with Labels allows much better performance, but splitting long lines introduce more complexity, but with a monospace font, knowing the width of the RV, you can set the expected height of an item in its data, using `key_size` to set it on the item, allows for correct computation of the total scrollable size, etc.

Implementing text selection on the RV would indeed be an "interesting" project, but i’m confident its possible, adding data about selection start/end in the item, and updating it in interactions.

Maybe a common `LogView` (though the name is a bit more generic than all the possible use cases) could implement this to avoid people reinventing this need.

PR #7645, opened 2021-10-01
Fixed TextInput _move_cursor_word_left / right for wrapped words

Alspb:
Fixed two bugs:
1) TextInput and CodeNavigationBehavior __move_cursor_word_left_ , __move_cursor_word_right_  methods move cursor only to the beginning / end of the current line for wrapped words, not to the beginning / end of the word itself.
2) For the case below CodeNavigationBehavior._move_cursor_word_right (i.e. Ctrl + Right shortcut) doesn't do anything at all.
CodeInput move word right bug

#### Note:
CodeNavigationBehavior.__move_cursor_word_right_ moves cursor to the beginning of the next word.
So for the following example 
CodeInput move word right note
('a...a' and 'b...b' are different words)
A) When cursor is at the end of the first line it should move to the beginning of the second line (i.e. beginning of the word 'b...b'),
B) When cursor is at the beginning of the second line it should move to the end of the text.
But since __move_cursor_word_left_ gets cursor index as an argument, it acts the same in both cases above (because cursor indices are the same). That's why currently in case B cursor also doesn't move at all.
The direct way to fix it is to change cursor index argument to cursor position. But not sure that it's worth it taking backward compatibility into account.
So in the fix for both cases __move_cursor_word_left_ acts as in the case B. But it's not a big deal, cause in case A cursor is already _almost_ at the beginning of the word 'b...b'.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/behaviors/codenavigation.py
+++ b/kivy/uix/behaviors/codenavigation.py
@@ -15,6 +15,7 @@ __all__ = ('CodeNavigationBehavior', )
 
 from kivy.event import EventDispatcher
 import string
+import kivy.uix.textinput as textinput
 
 
 class CodeNavigationBehavior(EventDispatcher):
@@ -35,6 +36,7 @@ class CodeNavigationBehavior(EventDispatcher):
 
         col, row = self.get_cursor_from_index(pos)
         lines = self._lines
+        lines_flags = self._lines_flags
 
         ucase = string.ascii_uppercase
         lcase = string.ascii_lowercase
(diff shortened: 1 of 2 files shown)


Alspb:
As I understand, the fail is caused by the line
`import kivy.uix.textinput as textinput`
in _codenavigation.py_.
Does anyone have any idea why?

For now I see only two alternatives:
A) Set
```
self.FL_IS_LINEBREAK = FL_IS_LINEBREAK
self.FL_IS_WORDBREAK = FL_IS_WORDBREAK
```
inside `TextInput.__init__` and use `self.FL_IS_LINEBREAK`, `self.FL_IS_WORDBREAK` in _codenavigation.py_, but it seems a bit odd.
B) Move `_move_cursor_word_left` and `_move_cursor_word_right` to `CodeInput` class, and therefore remove _codenavigation.py_ (as it becomes empty). Actually currently I don't see any disadvantages.

Alspb:
Decided to put `FL_IS_LINEBREAK, FL_IS_WORDBREAK` into CodeInput attributes.

PR #7677, opened 2021-11-04
Textinput cursor fixes

Alspb:
Fixed:

1. All TextInput instances have the same __adjust_viewport_ trigger, so if one changes cursors in several TextInput widgets simultaneously, it only scrolls to the new cursor position in the last widget.
2. Text "trembling" when resizing TextInput.
The reason is that currently it scrolls to the beginning before the next frame and then to the cursor after the next frame.
textinput resize trembling
3. Minor fixes.
a) self.cursor is in fact a tuple, so it should be initially set as tuple, not list;
b) When widget size changes `self._trigger_refresh_text()` is already called in `self._update_text_options` (which is bound to 'size' in `__init__`), so no need to duplicate it.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -490,7 +490,7 @@ class TextInput(FocusBehavior, Widget):
             self._update_graphics, -1)
         self._trigger_adjust_viewport = triggered(timeout=-1)(self._adjust_viewport)
         self.is_focusable = kwargs.get('is_focusable', True)
-        self._cursor = [0, 0]
+        self._cursor = 0, 0
         self._selection = False
         self._selection_finished = True
         self._selection_touch = None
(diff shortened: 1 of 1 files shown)


Alspb:
Seems that Read the docs build fails with Sphinx 4.2.0, see [the bug description](https://github.com/sphinx-doc/sphinx/issues/9817).

Alspb:
Addition: fixed 'del' action for wrapped words.
Currently if cursor is at the end of the line and word is wrapped, 'del' deletes character to the left from the cursor (i.e. acts like backspace).
Reason: 'del' currently equals "move cursor one char right plus backspace", but for wrapped words cursor just moves to the beginning of the next line. so it's index doesn't change.

PR #7679, opened 2021-11-05
MotionEvent: Fix calculation of z values in scale_for_screen method

pythonic64:
`MotionEvent` attributes `z`, `oz`, `pz`, `dz` will be set to `0` when `p` is `None`. Follow-up on https://github.com/kivy/kivy/pull/7659.

Maintainer merge checklist


tshirtman (approved):

tshirtman:
LGTM, if possible adding a test for what was broken before that change would be nice.

pythonic64:
@tshirtman This change is to reset `z` values for some extreme case where `p` argument changes between `scale_for_screen` calls. So nothing is broken, we are just covering this extreme case. Merge if you agree.

tshirtman (approved):
LGTM

Issue #7685, opened 2021-11-12
Scatter multitouch bug introduced in e8725996ac8cc8d533ba91f1e8325b74928d0b7d


Lauszus:
**Software Versions**
* Python: 3.8.10
* OS: Ubuntu 20.04
* Kivy: Latest from Github
* Kivy installation method: pip install from local Git repo

**Describe the bug**

After the latest commit to master: https://github.com/kivy/kivy/commit/e8725996ac8cc8d533ba91f1e8325b74928d0b7d scatter interactions does not work properly. Interactions work fine in the commit before that: https://github.com/kivy/kivy/commit/f40dd28ed6614d5e1a972fd5a3eeb99e12fa8578

**Expected behavior**

The scatter widget should rotate and scale child widgets according to the multitouch input.

**To Reproduce**

Run the following code:

```python
from kivy.app import App
from kivy.lang import Builder
from kivy.uix.relativelayout import RelativeLayout


scatter_layout = '''
<ScatterLayout>:
    canvas:
        Color:
            rgb: .8, .8, .8
        Rectangle:
            size: self.size
            pos: self.pos

    Scatter:
        Widget:
            size: 100, 100
            canvas:
                Color:
                    rgb: 1, 0, 0
                Rectangle:
                    size: self.size
                    pos: (root.width - self.width) / 2, (root.height - self.height) / 2
'''


class ScatterLayout(RelativeLayout):
    def __init__(self, **kwargs):
        Builder.load_string(scatter_layout)
        super().__init__(**kwargs)


class ScatterApp(App):
    def build(self):
        return ScatterLayout()


if __name__ == '__main__':
    ScatterApp().run()
```

And then try to interact with the widget either using a touchscreen or by emulating it by pressing middle click on the mouse.

**Code and Logs and screenshots**

Video showing the bug:

https://user-images.githubusercontent.com/1098847/141456713-29ae2360-2888-4728-a2b9-b60ddd193142.mp4

Before the commit it worked as intended:

https://user-images.githubusercontent.com/1098847/141456760-8413ae2e-3b35-4a30-9f09-e4342ae8793a.mp4

FYI the videos does not seem to work in Firefox, but they work fine when using Chrome.

pythonic64:
@Lauszus Thanks for reporting this issue.

Breaking of previous position also broke the delta values (`dx`, `dy`, `dz`). Pull request https://github.com/kivy/kivy/pull/7691 will fix the problem.

PR #7691, opened 2021-11-16
MotionEvent: Fix keeping of the previous normalized position

pythonic64:
Because input provider can make many calls to `MotionEvent.move`  before the event dispatch to the listeners we have to keep first previous position instead of latest one to have a consistent previous and current values.

Fixes https://github.com/kivy/kivy/issues/7685.

### Additional info:

This update is needed to synchronize normalized values and their absolute counterparts because even before https://github.com/kivy/kivy/pull/7659 only sx, sy, sz were in sync with x, y, z with others being in sync only for case when method `move` is called once per event dispatch.

Before the https://github.com/kivy/kivy/pull/7659 method `move` would assign current absolute x, y, z values to px, py, pz and scale_for_screen would set ox, oy, oz on the first call and make them equal to px, py, pz and x, y, z respectively. While this approach does work, it has issues:
- Method `move` is used by input provider and should only work with normalized values because providers work with normalized values. It should not change absolute values as those are set in `scale_for_screen`.
- Method `scale_for_screen` should only read normalized values and assign absolute values to respective attributes for current screen size, rotation and other arguments. Otherwise, because different arguments can be used to "scale the event" and then origin/previous values will be incorrect between the event update/move.
- It relies on `WindowBase.on_motion` method not to wrap transformation of the event with `push/pop` calls. Link https://github.com/kivy/kivy/blob/6fcf41729ce4abc6625bc732846f2c4d1d92ea3f/kivy/core/window/__init__.py#L1532-L1546

This changes will ensure that:
- Origin/previous/current normalized positions are the same at the time of the first dispatch and are equal to latest current position. This means that latest position will be used if `move` is called multiple time before the first dispatch.
- Previous normalized position is from the previous dispatch.
- Event managers can manipulate normalized positions to re-dispatch events or to do other behavior (example: https://github.com/pythonic64/hover/blob/14ddf18b9012af0b97aa00b690dff228333b9004/hover_manager.py#L115-L119).

We can remove attribute `sync_with_dispatch` as its value is never changed and just assume synchronized behavior.

Maintainer merge checklist:

### DIFF
--- a/kivy/input/motionevent.py
+++ b/kivy/input/motionevent.py
@@ -309,8 +309,10 @@ class MotionEvent(MotionEventBase):
 
     def depack(self, args):
         '''Depack `args` into attributes of the class'''
-        # set initial position and last position
-        if self.osx is None:
+        if self.osx is None \
+                or self.sync_with_dispatch and not self._first_dispatch_done:
+            # Sync original/previous/current positions until the first
+            # dispatch (etype == 'begin') is done.
             self.psx = self.osx = self.sx
             self.psy = self.osy = self.sy
             self.psz = self.osz = self.sz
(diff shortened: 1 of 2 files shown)


Lauszus:
I can confirm that this fixes https://github.com/kivy/kivy/issues/7685

kivy/input/motionevent.py
```diff
@@ -344,15 +365,25 @@ def ungrab(self, class_instance):
...
+        Called by the :meth:`EventLoopBase.dispatch_input`.
+
+        .. versionadded:: 2.1.0
+        '''
+        self._keep_prev_pos = True
+        self._first_dispatch_done = True
+
     def move(self, args):
-        '''Move the touch to another position
+        '''Move the touch to an another position.
```
tshirtman, line 379:
```suggestion
        '''Move the touch to another position.
```

tshirtman (approved):

tshirtman:
Ok, thanks for explanation and code simplification, way easier to appreciate like this.

tshirtman (approved):

PR #7701, opened 2021-11-22
deprecate 'kivy.utils.SafeList'

gottadiveintopython:
Maintainer merge checklist

`SafeList` is not used from Kivy itself or any garden flowers. And it's probably no longer necessary after Kivy dropped python2. So I believe we can safely remove it.


pythonic64:
I think we should deprecated class in 2.1.0 version and then remove in the next release.

gottadiveintopython:
Yeah that may be better.

tshirtman (approved):

PR #7703, opened 2021-11-22
[WIP] refactor kivy.lang

gottadiveintopython:
Maintainer merge checklist

Not sure this kind of PR is welcomed since it doesn't fix any bugs nor adds any features. But this includes couple of changes that may slightly improve performance, which is why I decided to open the PR. I'm not 100% sure all the changes are desirable, please point out if there are undesirable changes.

### DIFF
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -578,11 +578,11 @@ class BuilderBase(object):
             _ids = dict(rctx['ids'])
             _root = _ids.pop('root')
             _new_ids = _root.ids
-            for _key in _ids.keys():
-                if _ids[_key] == _root:
+            for _key, _value in _ids.items():
+                if _value == _root:
                     # skip on self
                     continue
-                _new_ids[_key] = _ids[_key]
+                _new_ids[_key] = _value
             _root.ids = _new_ids
 
         # first, ensure that the widget have all the properties used in
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -4,9 +4,7 @@ Builder
 
 Class used for the registering and application of rules for specific widgets.
 '''
-import codecs
 import sys
-import types
 from os import environ
 from os.path import join
 from copy import copy
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -967,7 +967,7 @@ if 'KIVY_PROFILE_LANG' in environ:
             '<style type="text/css">\n',
             'pre { margin: 0; }\n',
             '</style>']
-        files = set([x[1].ctx.filename for x in Builder.rules])
+        files = {x[1].ctx.filename for x in Builder.rules}
         for fn in files:
             try:
                 with open(fn) as f:
@@ -983,7 +983,7 @@ if 'KIVY_PROFILE_LANG' in environ:
                 for psn, rule in Builder.rules:
                     matched_prp += list(match_rule(fn, index, rule))
 
-                count = sum(set([x.count for x in matched_prp]))
+                count = sum({x.count for x in matched_prp})
 
                 color = (255, 155, 155) if count else (255, 255, 255)
                 html += ['<tr style="background-color: rgb{}">'.format(color),
--- a/kivy/lang/builder.py
+++ b/kivy/lang/builder.py
@@ -991,7 +991,7 @@ if 'KIVY_PROFILE_LANG' in environ:
                          '</tr>']
             html += ['</table>']
         html += ['</body></html>']
-        with open('builder_stats.html', 'w') as fd:
+        with open('builder_stats.html', 'w', encoding='utf-8') as fd:
             fd.write(''.join(html))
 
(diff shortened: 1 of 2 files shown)


matham (dismissed):

kivy/lang/builder.py
```diff
@@ -978,12 +976,12 @@ def dump_builder_stats():
...
             for index, line in enumerate(lines):
                 line = line.rstrip()
-                line = cgi.escape(line)
+                line = escape(line)
                 matched_prp = []
                 for psn, rule in Builder.rules:
                     matched_prp += list(match_rule(fn, index, rule))
 
-                count = sum(set([x.count for x in matched_prp]))
+                count = sum({x.count for x in matched_prp})
```
tshirtman, line 984: i agree it's similar to the previous code, but i'm now surprised by the logic, why do we want the sum of unique numbers in the counts? Granted, it's only the stats display, so it might have gone broken without us noticing it, but maybe there is a reason i fail to notice.

tshirtman (dismissed):

kivy/lang/builder.py
gottadiveintopython: >  why do we want the sum of unique numbers in the counts?

I too don't understand why `set` is used there. It's been there from the [very first](https://github.com/kivy/kivy/commit/20d594dfddf7ce4d9f801927c6777de3ccc1b6dd#diff-610290e2c0753e034f5ad7186190ce166cef9db020c0b514f50d1e36d7ca58c4R1633). I believe we shouldn't change the behaviour unless we are 100% sure it's wrong.

tshirtman (approved):

kivy/lang/builder.py
gottadiveintopython: I investegated it a little bit. When I put the lines below:

```diff
                for psn, rule in Builder.rules:
                    matched_prp.extend(match_rule(fn, index, rule))
+                if len({x.count for x in matched_prp}) != len(matched_prp):
+                    for rule in matched_prp:
+                        print(rule, rule.count)
                count = sum({x.count for x in matched_prp})
```

I got the follwing output:

```text
<ParserRuleProperty name='state_image' filename=path/to/style.kv:13 value="self.background_normal if self.state == 'normal' else self.background_down" watched_keys=[['self', 'background_normal'], ['self', 'state'], ['self', 'background_down']]> 28
<ParserRuleProperty name='state_image' filename=path/to/style.kv:13 value="self.background_normal if self.state == 'normal' else self.background_down" watched_keys=[['self', 'background_normal'], ['self', 'state'], ['self', 'background_down']]> 28
<ParserRuleProperty name='disabled_image' filename=path/to/style.kv:14 value="self.background_disabled_normal if self.state == 'normal' else self.background_disabled_down" watched_keys=[['self', 'background_disabled_normal'], ['self', 'state'], ['self', 'background_disabled_down']]> 23
<ParserRuleProperty name='disabled_image' filename=path/to/style.kv:14 value="self.background_disabled_normal if self.state == 'normal' else self.background_disabled_down" watched_keys=[['self', 'background_disabled_normal'], ['self', 'state'], ['self', 'background_disabled_down']]> 23
<ParserRuleProperty name='rgba' filename=path/to/style.kv:17 value='self.background_color' watched_keys=[['self', 'background_color']]> 0
<ParserRuleProperty name='rgba' filename=path/to/style.kv:17 value='self.background_color' watched_keys=[['self', 'background_color']]> 0
<ParserRuleProperty name='border' filename=path/to/style.kv:19 value='self.border' watched_keys=[['self', 'border']]> 0
<ParserRuleProperty name='border' filename=path/to/style.kv:19 value='self.border' watched_keys=[['self', 'border']]> 0
<ParserRuleProperty name='pos' filename=path/to/style.kv:20 value='self.pos' watched_keys=[['self', 'pos']]> 772
<ParserRuleProperty name='pos' filename=path/to/style.kv:20 value='self.pos' watched_keys=[['self', 'pos']]> 772
<ParserRuleProperty name='size' filename=path/to/style.kv:21 value='self.size' watched_keys=[['self', 'size']]> 1325
<ParserRuleProperty name='size' filename=path/to/style.kv:21 value='self.size' watched_keys=[['self', 'size']]> 1325
.
.
```

As you can see, there are duplicated lines. So maybe the code is correct? idk.

gottadiveintopython: I also tried the following, and the output was similar (each line was doubled).

```diff
                for psn, rule in Builder.rules:
                    matched_prp.extend(match_rule(fn, index, rule))
+                if len(matched_prp) > 1:
+                    for prp in matched_prp:
+                        print(prp, prp.count)
                count = sum({x.count for x in matched_prp})
```

Which means When `if len(matched_prp) > 1`,

- `matched_prp` always has two ParserRulePropeties
- and their `count` is the same. (`matched_prp[0].count == matched_prp[1].count`)

I don't know why there are doubled ParserRuleProperties but the code probably is not broken I guess.

PR #7714, opened 2021-12-03
MouseMotionEventProvider: Update simulated touch graphics on window resize or rotate

pythonic64:
Simulated touch graphics are not updated on window resize/rotate which makes graphics keep same position and thus makes difficult to left-click on touch graphics to cancel the simulated touch. This request fixes that case by binding to window's `on_rotate` event and `system_size` property and invoking `update_graphics` method for all touches once the binded callbacks are triggered. 

Should we update touch graphics when window's `softinput_mode` or `keyboard_height` change? Currently only window's size and rotation are used to compute absolute position for the touch graphics. Code reference: https://github.com/kivy/kivy/blob/6fcf41729ce4abc6625bc732846f2c4d1d92ea3f/kivy/input/providers/mouse.py#L132-L142

Maintainer merge checklist:

### DIFF
--- a/kivy/input/providers/mouse.py
+++ b/kivy/input/providers/mouse.py
@@ -211,6 +211,8 @@ class MouseMotionEventProvider(MotionEventProvider):
         fbind('on_mouse_down', self.on_mouse_press)
         fbind('on_mouse_move', self.on_mouse_motion)
         fbind('on_mouse_up', self.on_mouse_release)
+        fbind('on_rotate', self.update_touch_graphics)
+        fbind('system_size', self.update_touch_graphics)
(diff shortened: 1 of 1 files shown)


tshirtman (approved):

pythonic64:
If we ever need to update touch graphics on `keyboard_height` and `softinput_mode` changes then, beside doing `fbind/funbind` on those properties, lines:
```
     w, h = win._get_effective_size() 
     self.scale_for_screen(w, h, rotation=win.rotation) 
```
should be replaced with:
```
win.transform_motion_event_2d(self)
```

Issue #7700, opened 2021-11-19
Window.softinput_mode = 'below_target' does not work on platforms when keyboard mode is dock


brentpicasso:
Problem: 
Kivy does not get the keyboard height if dock mode is enabled for the keyboard. 

Expected behavior: If kivy is configured to use any on-screen keyboard - even if it is the kivy-generated keyboard generated on a desktop computer (for example) - the screen should move out of the way if softinput_mode is configured to do so (example "below_target")

This is the code that only looks for keyboard height on mobile platforms - it should also check if the docked keyboard is enabled and account for that.
https://github.com/kivy/kivy/blob/8ae4378e8517b23d8e9e35647ebf4e518a1c0883/kivy/core/window/__init__.py#L632

PR #7726, opened 2021-12-28
Window.softinput_mode fix for "pan" and "below_target" modes when using kivy virtual keyboard.

DexerBR:
Fixes inconsistent Kivy Virtual Keyboard display when using `Window.softinput_mode = 'below_target'`  or `Window.softinput_mode = "pan"`, with the docked keyboard mode enabled.

Fixes: #7700

Maintainer merge checklist

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -640,10 +640,8 @@ class WindowBase(EventDispatcher):
                 if isinstance(w, VKeyboard):
                     vkeyboard_height = w.height * w.scale
                     if smode == 'pan':
-                        w.top = 0
                         return vkeyboard_height
                     elif smode == 'below_target':
-                        w.top = w.target.y
                         return vkeyboard_height - w.target.y
         return 0
 
@@ -2151,6 +2149,13 @@ class WindowBase(EventDispatcher):
             # only after add, do dock mode
             keyboard.widget.docked = self.docked_vkeyboard
             keyboard.widget.setup_mode()
+
+            # sets vkeyboard position according to Window.softinput_mode
+            smode = self.softinput_mode
+            if smode == 'pan':
+                keyboard.widget.top = 0
+            elif smode == 'below_target':
+                keyboard.widget.top = keyboard.target.y
 
         else:
             # system keyboard, just register the callback.
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -631,18 +631,20 @@ class WindowBase(EventDispatcher):
 
     def _get_kivy_vkheight(self):
         mode = Config.get('kivy', 'keyboard_mode')
-        smode = self.softinput_mode
         if (
-            (mode == 'dock' or mode == 'systemanddock')
+            mode in ['dock', 'systemanddock']
             and self._vkeyboard_cls is not None
         ):
             for w in self.children:
                 if isinstance(w, VKeyboard):
                     vkeyboard_height = w.height * w.scale
-                    if smode == 'pan':
+                    if self.softinput_mode == 'pan':
                         return vkeyboard_height
-                    elif smode == 'below_target':
-                        return vkeyboard_height - w.target.y
+                    elif (
+                        self.softinput_mode == 'below_target'
+                        and w.target.y < vkeyboard_height
+                    ):
+                            return vkeyboard_height - w.target.y
         return 0
 
     def _get_kheight(self):
@@ -1627,7 +1629,7 @@ class WindowBase(EventDispatcher):
         w2, h2 = w / 2., h / 2.
         r = radians(self.rotation)
 
-        x, y = 0, 0
+        y = 0
         _h = h
         if smode == 'pan':
             y = kheight
@@ -2149,10 +2151,9 @@ class WindowBase(EventDispatcher):
             keyboard.widget.setup_mode()
 
             # sets vkeyboard position according to Window.softinput_mode
-            smode = self.softinput_mode
-            if smode == 'pan':
+            if self.softinput_mode == 'pan':
                 keyboard.widget.top = 0
-            elif smode == 'below_target':
+            elif self.softinput_mode == 'below_target':
                 keyboard.widget.top = keyboard.target.y
 
         else:
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -644,7 +644,7 @@ class WindowBase(EventDispatcher):
                         self.softinput_mode == 'below_target'
                         and w.target.y < vkeyboard_height
                     ):
-                            return vkeyboard_height - w.target.y
+                        return vkeyboard_height - w.target.y
         return 0
 
     def _get_kheight(self):
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -629,11 +629,31 @@ class WindowBase(EventDispatcher):
             import android
         return android.get_keyboard_height()
 
+    def _get_kivy_vkheight(self):
+        mode = Config.get('kivy', 'keyboard_mode')
+        smode = self.softinput_mode
+        if (
+            (mode == 'dock' or mode == 'systemanddock')
+            and self._vkeyboard_cls is not None
+        ):
+            for w in self.children:
+                if isinstance(w, VKeyboard):
+                    vkeyboard_height = w.height * w.scale
+                    if smode == 'pan':
+                        w.top = 0
+                        return vkeyboard_height
+                    elif smode == 'below_target':
+                        w.top = w.target.y
+                        return vkeyboard_height - w.target.y
+        return 0
+
     def _get_kheight(self):
         if platform == 'android':
             return self._get_android_kheight()
-        if platform == 'ios':
+        elif platform == 'ios':
             return self._get_ios_kheight()
+        else:
+            return self._get_kivy_vkheight()
         return 0
 
     keyboard_height = AliasProperty(_get_kheight, bind=('_keyboard_changed',))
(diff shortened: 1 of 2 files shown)


kivy/core/window/__init__.py
```diff
@@ -629,11 +629,31 @@ def _get_android_kheight(self):
...
+        smode = self.softinput_mode
+        if (
+            (mode == 'dock' or mode == 'systemanddock')
+            and self._vkeyboard_cls is not None
+        ):
+            for w in self.children:
+                if isinstance(w, VKeyboard):
+                    vkeyboard_height = w.height * w.scale
+                    if smode == 'pan':
+                        w.top = 0
```
pythonic64, line 643: Can we avoid assigning `top` property in this method as it should only return the height of virtual keyboard?

DexerBR: @pythonic64 Okay, so I think the code snippet below would be a better alternative:

```python
def _set_vkeyboard_top(self, vkeyboard, top=0):
    vkeyboard.top = top

...
if isinstance(w, VKeyboard):
    vkeyboard_height = w.height * w.scale
    if smode == 'below_target':
        self._set_vkeyboard_top(w, w.target.y)
        return vkeyboard_height - w.target.y
...


    

```

pythonic64: What I meant is that `_get_kivy_vkheight` (and any other getter) should not assign any values, but only get the requested value, in this case a height of virtual keyboard. 

Object opening the keyboard widget should assign its position.

DexerBR: You are right! My last commit fixes this.

kivy/core/window/__init__.py
```diff
@@ -629,11 +629,29 @@ def _get_android_kheight(self):
...
+                        return vkeyboard_height - w.target.y
+        return 0
+
     def _get_kheight(self):
         if platform == 'android':
             return self._get_android_kheight()
-        if platform == 'ios':
+        elif platform == 'ios':
             return self._get_ios_kheight()
+        else:
```
pythonic64, line 653: These three lines can merge into one: `return self._get_kivy_vkheight()`

kivy/core/window/__init__.py
pythonic64: Looking good 👍

kivy/core/window/__init__.py
DexerBR: Replace the conditions?
I don't think I understand, could you be more specific?

pythonic64:
```
        else:
            return self._get_kivy_vkheight()
        return 0
```
replace it with: `return self._get_kivy_vkheight()`

DexerBR: Understood!

misl6 (changes requested):
- Something unexpected is happening during runtime with click locations (See video)
- Added some minor suggestions

I have to check the cause, but maybe you're more updated than me about this argument right now:

https://user-images.githubusercontent.com/8177736/149671922-7281939f-a8bc-4e60-a2cc-af28ac9b0ff8.mov




```
Config.set('kivy', 'keyboard_mode', 'systemanddock')

from kivy.app import App
from kivy.lang import Builder

kv = """
BoxLayout:
    orientation: 'vertical'
    spacing: dp(20)
    TextInput:
        size_hint_y: None
        height: dp(45)
        hint_text: "input_type:null (default) and keyboard_suggestions: False"
        write_tab: False
        keyboard_suggestions: False
    TextInput:
        size_hint_y: None
        height: dp(45)
        hint_text: "input_type:null (default) and keyboard_suggestions: True"
        write_tab: False
        keyboard_suggestions: True
    TextInput:
        size_hint_y: None
        height: dp(45)
        hint_text: "input_type:text and keyboard_suggestions: False"
        write_tab: False
        input_type: 'text'
        keyboard_suggestions: False
    TextInput:
        size_hint_y: None
        height: dp(45)
        hint_text: "input_type:text and keyboard_suggestions: True"
        write_tab: False
        input_type: 'text'
        keyboard_suggestions: True
    BoxLayout:
        size_hint: 1, 1/2
"""


class Test(App):
    def build(self):
        Window.softinput_mode = 'below_target'


Test().run()
```

DexerBR:
> * Something unexpected is happening during runtime with click locations (See video)
> * Added some minor suggestions
> 
> I have to check the cause, but maybe you're more updated than me about this argument right now:
> 
>  Screen.Recording.2022-01-16.at.18.51.09.mov 
> ```
> from kivy.config import Config
> Config.set('kivy', 'keyboard_mode', 'systemanddock')
> 
> from kivy.app import App
> from kivy.core.window import Window
> 
> kv = """
> BoxLayout:
>     orientation: 'vertical'
>     spacing: dp(20)
>     TextInput:
>         size_hint_y: None
>         height: dp(45)
>         hint_text: "input_type:null (default) and keyboard_suggestions: False"
>         write_tab: False
>         keyboard_suggestions: False
>     TextInput:
>         size_hint_y: None
>         height: dp(45)
>         hint_text: "input_type:null (default) and keyboard_suggestions: True"
>         write_tab: False
>         keyboard_suggestions: True
>     TextInput:
>         size_hint_y: None
>         height: dp(45)
>         hint_text: "input_type:text and keyboard_suggestions: False"
>         write_tab: False
>         input_type: 'text'
>         keyboard_suggestions: False
>     TextInput:
>         size_hint_y: None
>         height: dp(45)
>         hint_text: "input_type:text and keyboard_suggestions: True"
>         write_tab: False
>         input_type: 'text'
>         keyboard_suggestions: True
>     BoxLayout:
>         size_hint: 1, 1/2
> """
> 
> 
> class Test(App):
>     def build(self):
>         Window.softinput_mode = 'below_target'
>         return Builder.load_string(kv)
> 
> 
> Test().run()
> ```

Fixed! Thanks!

kivy/core/window/__init__.py
```diff
@@ -629,12 +629,30 @@ def _get_android_kheight(self):
             import android
         return android.get_keyboard_height()
 
+    def _get_kivy_vkheight(self):
+        mode = Config.get('kivy', 'keyboard_mode')
+        if (
+            mode in ['dock', 'systemanddock']
```
pythonic64, line 635: Individual testing would be better as `mode in ['dock', 'systemanddock']` will just create a list to do a contains test and then trash it.

DexerBR: I agree with you. Because it's a comparison with only two different values, I don't think there's a big advantage compared to the individual test. Which would be true if there was a test for 5 different values, for example.
I will wait for the opinion of @misl6, before modifying again.

misl6: At a first glance, could look that `x == "dock" or x =="systemanddock"` is the optimised way to write that condition, but if we analyse that, at least on recents `CPython` implementations, the result is not what could be expected :)

```
>>> from timeit import timeit

>>> x = "systemanddock"
>>> print(timeit(lambda: x == "dock" or x == "systemanddock", number=10000000))
0.8471034320000399
>>> print(timeit(lambda: x in ["dock", "systemanddock"], number=10000000))
0.7669299230000206
>>> print(timeit(lambda: x in ("dock", "systemanddock"), number=10000000))
0.7826992520000431
>>> print(timeit(lambda: x in {"dock", "systemanddock"}, number=10000000))
0.709562496999979

>>> x = "system"
>>> print(timeit(lambda: x == "dock" or x == "systemanddock", number=10000000))
0.8592906410000296
>>> print(timeit(lambda: x in ["dock", "systemanddock"], number=10000000))
0.8776267410000855
>>> print(timeit(lambda: x in ("dock", "systemanddock"), number=10000000))
0.873066920000042
>>> print(timeit(lambda: x in {"dock", "systemanddock"}, number=10000000))
0.7129542200000287 


>>> import dis
>>> dis.dis(compile('x == "dock" or x == "systemanddock"', '<stdin>', 'eval'))
  1           0 LOAD_NAME                0 (x)
              2 LOAD_CONST               0 ('dock')
              4 COMPARE_OP               2 (==)
              6 JUMP_IF_TRUE_OR_POP      8 (to 16)
              8 LOAD_NAME                0 (x)
             10 LOAD_CONST               1 ('systemanddock')
             12 COMPARE_OP               2 (==)
             14 RETURN_VALUE
        >>   16 RETURN_VALUE
>>> dis.dis(compile('x in ["dock", "systemanddock"]', '<stdin>', 'eval'))
  1           0 LOAD_NAME                0 (x)
              2 LOAD_CONST               0 (('dock', 'systemanddock'))
              4 CONTAINS_OP              0
              6 RETURN_VALUE
>>> dis.dis(compile('x in ("dock", "systemanddock")', '<stdin>', 'eval'))
  1           0 LOAD_NAME                0 (x)
              2 LOAD_CONST               0 (('dock', 'systemanddock'))
              4 CONTAINS_OP              0
              6 RETURN_VALUE
>>> dis.dis(compile('x in {"dock", "systemanddock"}', '<stdin>', 'eval'))
  1           0 LOAD_NAME                0 (x)
              2 LOAD_CONST               0 (frozenset({'dock', 'systemanddock'}))
              4 CONTAINS_OP              0
              6 RETURN_VALUE
```

DexerBR: Thanks for the clarification @misl6!
So, I think everything is ok with my last commit.

pythonic64: @misl6 Nice 👍 

I ran the timeit test on my machine and `x in {"dock, "systemanddock"}` (which compiles to `frozenset`) provides the best performances. 

Python 3.7, Windows 10:
1. Case when function result is true (x = "systemanddock"):
   - 1.782105
   - 1.4844757
   - 1.4534518999999997
   - 1.3753492999999999

2. Case when function result is false (x = "system"):
   - 1.9317027000000002
   - 1.5773074
   - 1.6985533000000004
   - 1.3939842000000002

misl6 (approved):
Thank you. LGTM.

PR #7727, opened 2021-12-28
scrolleffect issue resolved

### LINKED ISSUE #7721: ScrollView effect_cls bug
**Software Versions**
* Python: 3.9.7
* OS: Win 10
* Kivy: v2.0.0

**Describe the bug**
I have a ScrollView with ScrollEffect and a simple BoxLayout on it, but if I try to scroll and resize later, the BoxLayout scrolls down regardless of whether it has like or not.
I showed this in a video I posted on youtube (not listed):
 - https://youtu.be/j3qYRGJ2BI8

**Expected behavior**
BoxLayout couldn't do this, I want it to be at the top of the screen when I resize the window.

**To Reproduce**
```
from kivy.app import App
from kivy.lang import Builder

KV = """
<BugScroll>:
    ScrollView:
        effect_cls:'ScrollEffect'
        BoxLayout:
            size_hint_y:None
            height:self.minimum_height
            orientation:'vertical'
            id:box
"""
Builder.load_string(KV)

class BugScroll(BoxLayout):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        for i in range(100):
            self.ids.box.add_widget(Button(size_hint_y=None, height=55, text=str(i)))

class Program(App):
    def build(self):
        return BugScroll()

if __name__ == '__main__':
    Program().run()

```

SrGambiarra:
I solved the scroll effect problem, issue #7721 is well explained what the problem was.

fix #7721
closes #7721

### DIFF
--- a/kivy/effects/scroll.py
+++ b/kivy/effects/scroll.py
@@ -25,13 +25,13 @@ from kivy.properties import NumericProperty, ObjectProperty
 
 
 class ScrollEffect(KineticEffect):
-    '''ScrollEffect class. See the module documentation for more informations.
+    '''ScrollEffect class. See the module documentation for more information.
     '''
 
     drag_threshold = NumericProperty('20sp')
     '''Minimum distance to travel before the movement is considered as a drag.
 
-    :attr:`velocity` is a :class:`~kivy.properties.NumericProperty` and
+    :attr:`drag_threshold` is a :class:`~kivy.properties.NumericProperty` and
     defaults to 20sp.
     '''
 
--- a/kivy/effects/scroll.py
+++ b/kivy/effects/scroll.py
@@ -25,13 +25,13 @@ from kivy.properties import NumericProperty, ObjectProperty
 
 
 class ScrollEffect(KineticEffect):
-    '''ScrollEffect class. See the module documentation for more information.
+    '''ScrollEffect class. See the module documentation for more informations.
     '''
 
     drag_threshold = NumericProperty('20sp')
     '''Minimum distance to travel before the movement is considered as a drag.
 
-    :attr:`drag_threshold` is a :class:`~kivy.properties.NumericProperty` and
+    :attr:`velocity` is a :class:`~kivy.properties.NumericProperty` and
     defaults to 20sp.
     '''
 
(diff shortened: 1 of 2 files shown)


kivy/effects/scroll.py
```diff
@@ -99,11 +99,13 @@ def on_value(self, *args):
...
         if self.value < scroll_min:
-            self.overscroll = self.value - scroll_min
-            self.reset(scroll_min)
+            if self.value > scroll_max:
+                self.overscroll = self.value - scroll_min
+                self.reset(scroll_min)
         elif self.value > scroll_max:
-            self.overscroll = self.value - scroll_max
-            self.reset(scroll_max)
+            if self.value < scroll_max:
```
misl6, line 106: How `self.value` could be less than `scroll_max` here, is a typo?

misl6 (changes requested):

kivy/effects/scroll.py
SrGambiarra: scroll_max is always 0 in ScrollEffect and self.value decreases along scroll from top to bottom

misl6: Considered that to reach line `106` the condition at line `105` is `True`, how can be `self.value < scroll_max` if it was `self.value > scroll_max` ?

SrGambiarra: I understand what you're saying, it doesn't really make sense, it's so that it never resets when it's in this condition, as it has no effect

SrGambiarra: if it resets, the behavior will not be as expected.

gottadiveintopython: I agree with @misl6 .

```python
elif A > B:
    if A < B:
```

The if statement looks never be True.

SrGambiarra: I was wrong guys, sorry!! I encountered the ScrollEffect problem.
I had to change the ScrollEffect and the DampedScrollEffect
I will send the corrections

gottadiveintopython: I don't fully understand the `kivy/effects/*` (I'm currently trying to), but I think the PR is appropriate. :+1:

gottadiveintopython:
I don't fully understand the `kivy/effects/*` (I'm currently trying to), but I think the PR is appropriate. :+1:

kivy/effects/scroll.py
```diff
@@ -106,6 +107,15 @@ def on_value(self, *args):
...
             self.scroll = self.value
+    
+    def apply_distance(self, distance):
+        if abs(distance) < self.min_distance:
+            self.velocity = 0
+
+        if distance < 0 and self.value > self.max:
+            self.value += distance
+        elif distance > 0 and self.value < self.min:
+            self.value += distance
```
gottadiveintopython, line 118: It's kinda confusing because if you look this part alone, it looks wrong. And the following code looks correct:

```python
        if distance < 0 and self.value > self.min:
            self.value += distance
        elif distance > 0 and self.value < self.max:
            self.value += distance
```

But when I actually changed the code like that, ScrollEffect didn't work correctly.

gottadiveintopython:
Looks like the issue is still there. On my pc, it sometimes occurs sometimes doesn't.

https://user-images.githubusercontent.com/26050135/149689800-bd49044f-abac-4eb3-9090-87658d563c93.mp4

gottadiveintopython:
Which is great work 👍 . But in order to be approved by the core developers, it might need to be perfect so I'll try to find a way to do it.

gottadiveintopython:
@SrGambiarra 
Can you check if the issue still occurs using [this branch](https://github.com/gottadiveintopython/kivy/tree/fix_7221)?
On my environment, it doesn't occur.

SrGambiarra:
@gottadiveintopython
Yes, it's working!! What did you change?

SrGambiarra:
Changes don't show up for me... I've been using GitHub for 1 year.
DumpedScrollEffect overrides some ScrollEffect functions, maybe that's the problem.

gottadiveintopython:
That's strange. This is the link to the diff.
https://github.com/gottadiveintopython/kivy/commit/babc7b69a0b5eb55500c19b6a06b9b603a3c9585

PR #7736, opened 2022-01-05
TextInput: Fixed #7474, issue 3

### LINKED ISSUE #7474: TextInput multiline wrap/cursor issues
Look at the top box as I change the window width: 
   image

- With a long line that wraps, if you backspace before the first character on second line, the previous line stops wrapping.
- With a line that wraps across multiple lines or multi-line text, if you use up-arrow, it doesn't keep its position horizontally based on the pixel number offset from left, but rather based on number of characters. So with non-uniform character font, it looks weird and is different than e.g. firefox.

_Originally posted by @matham in https://github.com/kivy/kivy/issues/7449#issuecomment-815529292_

Alspb:
Fixed #7474, issue 3:

> With a line that wraps across multiple lines or multi-line text, if you use up-arrow, it doesn't keep its position horizontally based on the pixel number offset from left, but rather based on number of characters. So with non-uniform character font, it looks weird and is different than e.g. firefox.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -1186,6 +1186,37 @@ class TextInput(FocusBehavior, Widget):
             / (self.line_height + self.line_spacing) - 1
         )
 
+    def _get_col_by_width(self, row, col, width):
+        if col == 0:
+            return col
+        _get_text_width = self._get_text_width
+        _tab_width = self.tab_width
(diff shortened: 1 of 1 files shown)


misl6 (changes requested):
Hi @Alspb 
Unit tests are failing, can you please take care of it?

PR #7737, opened 2022-01-05
CodeInput: fixed disappearing lines after inserting text

Alspb:
Fixed: Last CodeInput line disappears when inserting text ended with "\n" to the end of the CodeInput's text.

How to reproduce:
1) Select part of the text ended with the linebreak:
CodeInput insert bug
2) Put cursor at the end of the text (after the last "b" character) and press Ctrl+V (paste). Last line disappeared:
CodeInput insert bug 2

Maintainer merge checklist

### DIFF
--- a/kivy/uix/codeinput.py
+++ b/kivy/uix/codeinput.py
@@ -129,7 +129,7 @@ class CodeInput(CodeNavigationBehavior, TextInput):
             ntext = u'*' * len(ntext)
         ntext = self._get_bbcode(ntext)
         kw = self._get_line_options()
-        cid = u'{}\0{}\0{}'.format(ntext, self.password, kw)
+        cid = u'{}\0{}\0{}'.format(text, self.password, kw)
         texture = Cache_get('textinput.label', cid)
 
         if texture is None:


kivy/uix/codeinput.py
```diff
@@ -129,7 +129,7 @@ def _create_line_label(self, text, hint=False):
             ntext = u'*' * len(ntext)
         ntext = self._get_bbcode(ntext)
         kw = self._get_line_options()
-        cid = u'{}\0{}\0{}'.format(ntext, self.password, kw)
+        cid = u'{}\0{}\0{}'.format(text, self.password, kw)
```
misl6, line 132: What about the `ntext` variable and all the `bbcode` part?
Is something that we would like to change?

Alspb: Seems that everything else works fine, in particular without using cache (line 133).
The problem is that different `text`s can give the same `ntext`. At the same time `text` also used (line 141) after setting `ntext` (in contrast to TextInput._create_line_label).

misl6: Tested. And now makes more sense to me. Thank you!

misl6 (approved):

Issue #7337, opened 2021-01-19
Incorrect behavior keyboard_suggestions


idevo89:
**Software Versions**
* Python: 3.8
* OS: Android 9
* Kivy: master
* Kivy installation method: python-for-android develop with kivy==master

**Describe the bug**
Keyboard suggestions for TextInput do not work correctly on android. Words are inputted twice when you use a suggestion or swipe as you can notice when you delete the words or press enter. Setting keyboard_suggestions to false does not deactivate suggestions.

**Expected behavior**
Keyboard suggestions or swiping should input words correctly and it should be possible to deactivate the suggestions feature with keyboard_suggestions set to false.

**To Reproduce**
```python
from kivy.app import App
from kivy.lang import Builder


kv = """
Screen:
    TextInput:
        size_hint_y: None
        height: dp(45)
        hint_text: "keyboard_suggestions: False"
        write_tab: False
        keyboard_suggestions: False
        pos_hint: {"top": 0.9}
    TextInput:
        size_hint_y: None
        height: dp(45)
        hint_text: "keyboard_suggestions: True"
        write_tab: False
        keyboard_suggestions: True
        pos_hint: {"top": 0.7}
"""


class Test(App):
    def build(self):
        return Builder.load_string(kv)


Test().run()

```

**Gif**

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/48392636/105105400-51ec8400-5ab4-11eb-80cf-7c3ab896c822.gif)

matham:
I'm wondering if it's due to https://github.com/kivy/kivy/pull/7231. Can you try a commit before that? Or maybe 2.0.0?

@dwmoffatt could you take a look?

dwmoffatt:
I'll take a look

misl6:
GBoard:

- Issue confirmed when using swipe or suggestions.

Samsung Keyboard:

- The first inputted character, keeps into TextInput.

Reverting #7231 fixed the issue.

It could potentially disrupt the UX for multiple users. @dwmoffatt any news?

dwmoffatt:
I think I know what the problem is. There are some other variations/flags that need to get set for some of the autocomplete and suggestions to work properly based on how you are entering the data. 

https://developer.android.com/reference/android/text/InputType

I haven't had a chance to try anything out yet. Hopefully next week I will have some time.

misl6:
Update:
[...]

PR #7744, opened 2022-01-09
Change default input_type to null. Add some warning regarding TYPE_TEXT_FLAG_NO_SUGGESTIONS

misl6:
Maintainer merge checklist

Some info regarding this PR:

- Issue https://github.com/kivy/kivy/issues/7337 has been introduced via https://github.com/kivy/kivy/pull/7231.
- Some (major) soft keyboard vendors are de-facto ignoring the `TYPE_TEXT_FLAG_NO_SUGGESTIONS`, so the workaround of disabling the suggestions doesn't work.
- There are some issues on SDL regarding IME and `SDL_TEXTEDITING`, that we should track: https://github.com/libsdl-org/SDL/issues/4847, https://github.com/libsdl-org/SDL/issues/3385, https://github.com/libsdl-org/SDL/issues/3377
- `input_type` is a great addition and is a step forward to support non-latin languages.
- Some events from SDL2 are missing in order to fully support suggestions and swipes.
- This PR should be considered as a temp fix, that reverts the default keyboard behaviour to the one we had prior to https://github.com/kivy/kivy/pull/7231, without completely reverting the great additions that have been made by @dwmoffatt. A new Kivy release, will likely land quite soon, and this is a main blocker.

The following code snipped could be used for testing:

```
from kivy.app import App
from kivy.lang import Builder


kv = """
BoxLayout:
    orientation: 'vertical'
    spacing: dp(20)
    TextInput:
        size_hint_y: None
        height: dp(45)
        hint_text: "input_type:null (default) and keyboard_suggestions: False"
        write_tab: False
        keyboard_suggestions: False
    TextInput:
        size_hint_y: None
        height: dp(45)
        hint_text: "input_type:null (default) and keyboard_suggestions: True"
        write_tab: False
        keyboard_suggestions: True
    TextInput:
        size_hint_y: None
        height: dp(45)
        hint_text: "input_type:text and keyboard_suggestions: False"
        write_tab: False
        input_type: 'text'
        keyboard_suggestions: False
    TextInput:
        size_hint_y: None
        height: dp(45)
        hint_text: "input_type:text and keyboard_suggestions: True"
        write_tab: False
        input_type: 'text'
        keyboard_suggestions: True
    BoxLayout:
        size_hint: 1, 1/2
"""


class Test(App):
    def build(self):


Test().run()
```

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -2070,7 +2070,7 @@ class WindowBase(EventDispatcher):
 
                 .. note::
 
-                    `input_type` is currently only honored on mobile devices.
+                    `input_type` is currently only honored on Android.
 
                 .. versionadded:: 1.8.0
 
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -2065,13 +2065,17 @@ class WindowBase(EventDispatcher):
 
             `input_type`: string
                 Choose the type of soft keyboard to request. Can be one of
-                'text', 'number', 'url', 'mail', 'datetime', 'tel', 'address'.
+                'null', 'text', 'number', 'url', 'mail', 'datetime', 'tel',
+                'address'.
 
                 .. note::
 
                     `input_type` is currently only honored on mobile devices.
 
                 .. versionadded:: 1.8.0
+
+                .. versionchanged:: 2.1.0
+                    Added `null` to soft keyboard types.
 
             `keyboard_suggestions`: bool
                 If True provides auto suggestions on top of keyboard.
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -554,6 +554,7 @@ cdef class _WindowSDL2Storage:
                 TYPE_CLASS_NUMBER = 2
                 TYPE_CLASS_PHONE = 3
                 TYPE_CLASS_TEXT = 1
+                TYPE_CLASS_NULL = 0
 
                 TYPE_TEXT_VARIATION_EMAIL_ADDRESS = 32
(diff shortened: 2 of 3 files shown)


matham:
What is the implication of having null be the default? What is the difference between null and text? Is it just the same as text, except that it can't support suggestions? I don't have android setup for testing so can't test.

misl6:
1) Prior to #7231 the default configuration for keyboard was `InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD`. That configuration shows a latin-only keyboard without suggestions. The same behaviour looks like can be achieved via [TYPE_NULL](https://developer.android.com/reference/android/text/InputType#TYPE_NULL)  (at least on my configuration, so tests on multiple configurations are really appreciated)
2) #7337 still remains open for `input_type="text"`
3) `keyboard_suggestions` only works on certain keyboards, when `input_type="text"`. Unfortunately GBoard, is not one of them, cause https://developer.android.com/reference/android/text/InputType#TYPE_TEXT_FLAG_NO_SUGGESTIONS is ignored.

matham:
Ok, we'll have to make sure to document this clearly in the release notes, and possibly in the docs itself.

Because the change to default to null is technically a backward compat break, and for people who manually set it to text before, even if it was the default, it'll still be broken. Unless we change `keyboard_suggestions` to default to False, which we probably shouldn't.

misl6:
Considering that for the vast majority of users `keyboard_suggestion=False` acts as the same of `keyboard_suggestion=True` when `input_type="text"`, I can only totally agree about clearly specifying this behaviour in docs. The same applies for the breaking change.

I will be back here to update it just after the documentation review (WIP 🧐).

matham (approved):

Issue #5540, opened 2017-12-17
Popup's on_dismiss is called before its on_open


omermazig:
### Versions

* Python: 3.5.2
* OS: linux
* Kivy: 1.10.0

### Description

When I do two actions which should open and close a popup fast enough (click on 'save' button which opens the popup, and then click somewhere else on the screen which closes it), I see in debug mode that on_dismiss is called **before** on_open, and that causes all kind of problems. How can this be?

I don't use the keyboard in this use-case, so I don't think the inheritance from my KeyboardListener is unrelated to the problem, but I added it here for any case.

### Code and Logs

```python
from kivy.uix.popup import Popup
from kivy.event import EventDispatcher

class TinyPopup(Popup, KeyboardListener):
    '''
    A base class for popups in our systems
    '''

    dismiss_callback = ObjectProperty()
    '''Call this callback when the popup is closed
    '''

    cancel_callback = ObjectProperty()
    '''Call this callback when the popup is canceled
    '''

    def on_open(self):
        '''
        Called when the popup is opened
        :return: Notnig
        :rtype: None
        '''
        # Register for keyboard events
        self.register_for_keyboard_listening()

    def on_dismiss(self):
        ''''Call this when the popup is dismissed
        '''
        # Unregister for keyboard events
        self.stop_listening_to_keyboard()

        # Call the dismiss_callback if we have it
        if self.dismiss_callback is not None:
            self.dismiss_callback()

    def on_keyboard_down(self, keyboard, keycode, text, modifiers):
        '''
        Act when a key is pressed down
        :param keyboard: The keyboard object
        :param keycode: The pressed key code
        :param text: The text that was pressed
        :param modifiers: A list of keyboard modifiers
        :return: True to accept the key, or False to ignore it
        '''
        if keycode[1] == 'escape':
            # Dismiss popup upon 'Esc'
            self.dismiss()

            # Call the cancel_callback if we have it
            if self.cancel_callback is not None:
                self.cancel_callback()

        # Grag all keyboard events so that the base app doesn't get anything
        return True


class KeyboardListener(EventDispatcher):
    '''
    A Keyboard listener
    '''

    def register_for_keyboard_listening(self, *args):
        '''
        Start listening to the keyboard
        '''
        if get_top_widget() is not None:
            get_top_widget().add_keyboard_listener(self)
        else:
            Clock.schedule_once(self.register_for_keyboard_listening, 0)

    def stop_listening_to_keyboard(self):
        '''
[...]

PR #7748, opened 2022-01-11
ModalView: on_dismiss runs before on_open if user dismisses popup before opening animation completes

Alspb:
The reason is that `on_open` only runs when opening animation completes, which can take some time if `_anim_duration` parameter is high.
Such a behavior seems counterintuitive, it's expected that `on_open` should always precede `on_dismiss`.
Also it worth noting that dismissing doesn't stop opening animation, so both animations run together.

To avoid these issues it's suggested not to dismiss the popup if opening animation is in progress and vice versa not to open if dismissing animation still runs.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/modalview.py
+++ b/kivy/uix/modalview.py
@@ -204,7 +204,7 @@ class ModalView(AnchorLayout):
 
         """
         from kivy.core.window import Window
-        if self._is_open:
+        if self._is_open or self._anim_alpha > 0:
             return
         self._window = Window
         self._is_open = True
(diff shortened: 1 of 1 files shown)


matham (changes requested):
I don't think it's a good idea to rely on the animation value. Instead, it'd be better to save the animation when it's run and clear it when it's stopped, and then check whether the animation exists.

Also, tests for this condition would be nice to have.

pythonic64:
Related issue https://github.com/kivy/kivy/issues/5540.

NomadDemon:
How about cancel animation and cancel "animation_done" callback ?

PR #7749, opened 2022-01-13
collections fix for python 3.10

sanderland:
Issue: https://github.com/sanderland/katrain/issues/499

Fix from: https://gitlab.freedesktop.org/gstreamer/gst-devtools/-/merge_requests/257/diffs?commit_id=2495a9d910877a944d59b37449ad0aa14037fb5b

### DIFF
--- a/kivy/uix/filechooser.py
+++ b/kivy/uix/filechooser.py
@@ -101,15 +101,12 @@ from kivy.uix.screenmanager import ScreenManager, Screen
 from kivy.properties import (
     StringProperty, ListProperty, BooleanProperty, ObjectProperty,
     NumericProperty, AliasProperty)
+import collections
 from os import listdir
 from os.path import (
     basename, join, sep, normpath, expanduser, altsep,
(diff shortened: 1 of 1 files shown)


kivy/uix/filechooser.py
```diff
@@ -106,7 +106,10 @@
     basename, join, sep, normpath, expanduser, altsep,
     splitdrive, realpath, getsize, isdir, abspath, isfile, dirname)
 from fnmatch import fnmatch
-import collections
+try:
+    import collections.abc as collections_abc
+except ImportError:
+    import collections as collections_abc
```
misl6, line 112: > Deprecated since version 3.3, will be removed in version 3.9: Moved Collections Abstract Base Classes to the collections.abc module. For backwards compatibility, they continue to be visible in this module through Python 3.8."

I think that should be safe for us to just change it to `import collections.abc` and then use `is instance(file, collections.abc.Callable)`, cause we don't mark `kivy` as supported on `Python<=3.3`.

misl6 (dismissed):

matham (approved):

Issue #7767, opened 2022-01-24
Kivy ProgressBar-value doesn't get updated when kwargs' position switched.


ApuCoder:
I'm using Kivy (v-2.0.0) [Python (v-3.9.7)].

**The problem**
I noticed the unexpected behaviour (behavior) when initialized the `ProgressBar` with the attrs. `max` and `value` with different order such as `ProgressBar(max = 1000, value = 450)` or `ProgressBar(value = 450, max = 1000)`. They both render different result.

**What I tried but didn't work**
I tried to bind the attr. "max" with the `AliasProperty` `value`, setting different `cache` values, but none of them seem to work for me. I think (not quite sure) this problem arises due to the `setter` method.

**What seems to work**
I changed the attr. `value` to `BoundedNumericProperty` as `BoundedNumericProperty(0.0, min = 0, errorvalue = 0)` and redefine it in the __init__ as
```python
    def __init__(self, **kwargs):
        #self._value = 0.
        super(ProgressBar, self).__init__(**kwargs)
        self.value = min(self.max, self.value) # To limit the upper-value of the attr: value within attr: max.
```
Now I got the result in both cases as expected.

**Executable (for the original problem)**
```python
from kivy.app import runTouchApp
from kivy.uix.progressbar import ProgressBar

pb = ProgressBar(max = 1000, value = 450) # This works as intended.
#pb = ProgressBar(value = 450, max = 1000) # This does not work as intended.
print(f"{pb.max = }, {pb.value = }")

runTouchApp(pb)
```

ApuCoder:
Well, I noticed again this doesn't solve the issue completely, as the attr. `value` can be (made) unbounded above at runtime.

kuzeyron:
Looks to me like you are calling the super() in the wrong class. You should be seeing something in the logs about this.

ApuCoder:
No, actually I did it in source code of `ProgressBar` after observing that odd behaviour. Did you also notice that same behaviour after altering the position of kwargs ?

kuzeyron:
> No, actually I did it in source code of `ProgressBar` after observing that odd behaviour. Did you also notice that same behaviour after altering the position of kwargs ?

Not the best practice on editing the source code of Kivy. Here is one example that works:
```python
from kivy.app import App
from kivy.properties import NumericProperty
from kivy.uix.progressbar import ProgressBar


class MyProgressBar(ProgressBar):
    max = NumericProperty(20)
    value = NumericProperty(50)

[...]

PR #7772, opened 2022-01-27
Fix for ProgressBar when value is not set

kuzeyron:
This PR solves the issue when value isn't set for ProgressBar.
Issue ticket: https://github.com/kivy/kivy/issues/7767

### DIFF
--- a/kivy/uix/progressbar.py
+++ b/kivy/uix/progressbar.py
@@ -43,12 +43,12 @@ class ProgressBar(Widget):
         return self._value
 
     def _set_value(self, value):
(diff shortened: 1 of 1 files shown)


kuzeyron:
This was only working temporary.

ApuCoder (changes requested):
Is it really necessary to have a limit (especially the upper limit) for the attr. `value` ? Why not let the user decide and manage it ? In other word, will it not be sufficient to set `value` a `BoundedNumericProperty` with `min=0` or atleast a `NumericProperty` ?

PR #7778, opened 2022-01-31
WindowBase: Don't return motion event in transform_motion_event_2d method

pythonic64:
Motion event is transformed inplace, new instance is not created so returning the same one is not needed.

Maintainer merge checklist:


matham:
But that is a backward compat break. What is the benefit of not returning it. Returning a transformed, but original object is part of a [fluent design](https://en.wikipedia.org/wiki/Fluent_interface) and is often helpful.

matham (approved):

Issue #7670, opened 2021-10-26
Add support to release linux aarch64 wheels


odidev:
Problem
---------
On aarch64, ‘pip install kivy’ builds the wheels from source code and then installs it. It requires the user to have a development environment installed on their system. Also, it takes more time to build the wheels than downloading and extracting the wheels from PyPi.

Resolution
-----------
On aarch64, ‘pip install kivy’ should download the wheels from PyPi

@tshirtman, Please let me know your interest in releasing aarch64 wheels. I can help with this.

sysnux:
+1
I just received my Librem5 and building kivy on the phone is not ideal.
Once installed, kivy apps work fine, except maybe multitouch (I need to investigate on that).

odidev:
@tshirtman, manylinux2014 has added Linux AArch64 support and also PyPI supports Linux AArch64 wheels.

GHA Doesn't provides Linux AArch64 builders so we need to build the wheels in emulated environment. below are the commit ID and build logs- _Because of emulation it's slow and takes almost 2 Hours to build single wheel._
https://github.com/odidev/kivy/commit/0dc3ad69e108a83d767daf287f9c63ce5cc2e992
https://github.com/odidev/kivy/actions/runs/1698079472

We can also use Circle CI as it provides Linux AArch64 builders and it requires just 21 minutes to build all the wheels. Below are the commit ID and builds logs.
https://github.com/odidev/kivy/commit/b47e66e95b5e498fe156b4eba1f36559a09a694c
https://app.circleci.com/pipelines/github/odidev/kivy/1/workflows/eb6c4921-5206-439d-a0d7-57d12d916261/jobs/1?invite=true#step-106-4441

Which one looks good to you? Please let me know your thoughts on this. Thanks.

tshirtman:
Thanks for the research and proof of concept!

I like the simplicity of using Github actions for everything, and i wonder if the slow build is bad enough to warrant reaching for CircleCi again, it's true it'll run every day for nightly builds, but it's not going to impact the formal CI that would delay merge requests. It's certainly wasteful though, but if that's an issue we could only do this build on demand, for release or specific tests.

[...]

PR #7784, opened 2022-02-02
Linux AArch64 wheel build optimization

odidev:
Wheel build time optimization- [background](https://github.com/kivy/kivy/issues/7670#issuecomment-1015953647)

Maintainer merge checklist

### DIFF
--- a/.github/workflows/manylinux_wheels.yml
+++ b/.github/workflows/manylinux_wheels.yml
@@ -42,14 +42,14 @@ jobs:
       CIBW_BUILD:  ${{ matrix.cibw_build }}
       CIBW_ARCHS: ${{ matrix.cibw_archs }}
       CIBW_BEFORE_ALL_LINUX: >
-        source .ci/ubuntu_ci.sh &&
+        mv `pwd`/kivy_build $HOME &&
         yum install -y epel-release &&
-        build_and_install_linux_kivy_sys_deps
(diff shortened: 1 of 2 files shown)


.github/workflows/manylinux_wheels.yml
```diff
@@ -42,14 +42,14 @@ jobs:
       CIBW_BUILD:  ${{ matrix.cibw_build }}
       CIBW_ARCHS: ${{ matrix.cibw_archs }}
       CIBW_BEFORE_ALL_LINUX: >
-        source .ci/ubuntu_ci.sh &&
+        mv `pwd`/kivy_build $HOME &&
         yum install -y epel-release &&
-        build_and_install_linux_kivy_sys_deps
+        yum -y install autoconf automake cmake gcc gcc-c++ git make pkgconfig zlib-devel portmidi portmidi-devel xorg-x11-server-devel mesa-libEGL-devel mtdev-devel mesa-libEGL freetype freetype-devel openjpeg openjpeg-devel libpng libpng-devel libtiff libtiff-devel libwebp libwebp-devel dbus-devel dbus ibus-devel ibus libsamplerate-devel libsamplerate libudev-devel libmodplug-devel libmodplug libvorbis-devel libvorbis flac-devel flac libjpeg-turbo-devel libjpeg-turbo wget
```
matham, line 47: Looks good, but this should be in its own function in `.ci/ubuntu_ci.sh` so you can call it here as well as in https://github.com/kivy/kivy/blob/master/.ci/ubuntu_ci.sh#L166. The idea is to have minimal actual code in these actions so it's easier to understand and in case we want to migrate from github actions in the future.

matham (changes requested):

matham (approved):

PR #7786, opened 2022-02-03
WindowBase: Add on_drop_begin, on_droptext and on_drop_end events

pythonic64:
This adds `on_drop_begin` and `on_drop_end` events to `WindowBase` which are useful when dropping multiple files on the window.

I unable to test `on_droptext` because SDL_DROPTEXT event is not working for me and therefore `on_droptext` is not being dispatched. This could be an SDL issue which leaves us with question should we keep this event or remove it?

Should I rename `on_dropfile` to `on_drop_file` but still dispatch `on_dropfile` for backward compatibility?

Maintainer merge checklist:


matham:
Hmm yeah, adding `on_drop_xxx` and leaving the old `on_dropxx` as well sounds like a good plan. Maybe we can mark it as deprecated, but we won't remove it for a while!?

About it not working for you, I'd make sure to try dragging in plain text without any formatting, because I don't think it supports dragging in rich text.

pythonic64:
I'm thinking of dispatching `on_dropfile` in `on_drop_file` method and `on_dropfile` will have `deprecated` decorator.

I did drag plain text from the web page and it didn't work.

pythonic64:
Seems that SDL_DROPTEXT is currently only supported on x11 window https://github.com/libsdl-org/SDL/blob/release-2.0.20/src/video/x11/SDL_x11events.c#L1497.

kivy/core/window/__init__.py
```diff
@@ -2017,9 +2051,37 @@ def on_dropfile(self, filename):
             (ios, android etc.)
 
         .. versionadded:: 1.2.0
+
+        .. versionchanged:: 2.1.0
+            Renamed from `on_dropfile` to `on_drop_file`.
         '''
+        if self.get_property_observers('on_dropfile'):
+            self.dispatch('on_dropfile', filename)
```
matham, line 2059: `get_property_observers` is not particularly efficient. I don't think it's worth checking if anything is bound and instead always dispatch.

I also don't like when widgets do any work in `on_xxx` methods, except in user code. Because it's possible someone will inherit the class and overwrite the method and not call `super` - we have not trained our users to call `super` in these callbacks. It's less likely to be a problem, but still.

A better solution is to do the following in `__init__`: `self.fbind('on_drop_file', self.dispatch, 'on_dropfile')` and it should work,

kivy/core/window/__init__.py
```diff
@@ -2017,9 +2051,37 @@ def on_dropfile(self, filename):
...
         pass
 
+    def on_drop_text(self, text):
+        '''Event called when a text is dropped on the application.
+
+        .. note::
+            This event works with sdl2 window provider on x11 window.
+
+        .. versionadded:: 2.1.0
+        '''
```
matham, line 2073: I'd add a `pass` here and below, although technically we don't need it.

matham (changes requested):

kivy/core/window/__init__.py
pythonic64: `self.fbind('on_drop_file', self.dispatch, 'on_dropfile')` doesn't work because `on_drop_file` passes a window instance, but wrapping with `lambda` does.

kivy/core/window/__init__.py
pythonic64: Done.

matham (approved):

pythonic64:
@matham What do you think about this `mouse_pos` fix https://github.com/kivy/kivy/pull/7786/files#diff-2412482658d195db4f3a0a6204b1473c0184368e4a0352c7ad37a0f586d4ec14R742-R748 ?

matham:
I'm not sure why it originally didn't work on linux so I'm not sure if this is needed. But if this updates the mouse position correctly than it's good!?

pythonic64:
It needs that code for Windows too. With changes from this pr, `mouse_pos` is updated first and `on_drop_begin` follows and position is correct when file/text is dropped on the window content, but if drop occurred on the window edges or title bar then `mouse_pos` is set to the nearest value (using max/min).

Without `mouse_pos` update when I drag a file, drop it on the window and don't move the mouse:

On Windows:
- on_drop_begin (incorrect `mouse_pos`)
- on_drop_file (single or multiple)
- on_drop_end
- on_cursor_enter (incorrect `mouse_pos`)
- on_mouse_move (correct `mouse_pos`, assigned first and then `on_mouse_move` is dispatched)

On Linux:
- on_cursor_enter (incorrect `mouse_pos`)
- on_drop_begin
- on_drop_file (single or multi)
- on_drop_end
- no on_mouse_move, `mouse_pos` stays incorrect, until mouse moves

With `mouse_pos` update, we have `mouse_pos` assign (and `mouse_pos` event) and then `on_drop_begin` with correct `mouse_pos` value.

pythonic64:
So the question is when should we update `mouse_pos` and should we also dispatch `on_mouse_move` after it? 

Maybe this flow is a good idea, when action `windowenter` is received, before `on_cursor_enter` event https://github.com/kivy/kivy/pull/7786/files#diff-42f1340bfa2c476e54b6bd2c1fc18ca99241460bd2b56ccb5b5fb28cf6d84565R679:
- update `mouse_pos`
- dispatch `on_cursor_enter`
- dispatch `on_mouse_move`

pythonic64:
Ok, I made the changes to always assign correct value to `mouse_pos` and then dispatch `on_cursor_enter`. Works on both Windows and Linux.

Also fixed `mouse_pos` when cursor is static, app is started and wheel scroll up/down as it would dispatch `on_mouse_down` and `on_mouse_up` with -1 for x and y.

kivy/core/window/window_sdl2.py
```diff
@@ -678,9 +689,14 @@ def mainloop(self):
                 self._focus = False
 
             elif action == 'windowenter':
-                self.dispatch('on_cursor_enter')
+                x, y = self._win.get_nearest_mouse_pos()
```
matham, line 692: `windowenter` is not proceeded by any other mouse events, right? And is it specific to mouse?

kivy/core/window/window_sdl2.py
```diff
@@ -759,6 +775,21 @@ def mainloop(self):
             else:
                 Logger.trace('WindowSDL: Unhandled event %s' % str(event))
 
+    def _dispatch_drop_event(self, action, args):
+        if action == 'dropfile':
+            self.dispatch('on_drop_file', args[0])
+        elif action == 'dropbegin':
```
matham, line 781: I'm assuming no mouse events happen before this, hence the original issue!? However, could we not use e.g. touch or some other non-mouse input to drop stuff? In which case we should not update the mouse!?

kivy/core/window/window_sdl2.py
pythonic64: From sdl doc:
```
SDL_WINDOWEVENT_ENTER - window has gained mouse focus
SDL_WINDOWEVENT_LEAVE - window has lost mouse focus
```

SDL reports the following events:

On Windows if app starts with cursor on the window:
- `windowexposed` is received
- no `windowenter`
- if mouse moves, `mousemotion` is received, `mouse_pos` is updated
- if any of mouse buttons is pressed, `mousebuttondown` is received, `mouse_pos` is updated, `on_mouse_down` is dispatched

If apps starts with cursor outside of window:
- `windowexposed` is received
- if cursor is moved on window, `windowenter` is received
- `mousemotion` is received, `mouse_pos` is updated

On Linux if app starts with cursor on the window:
- `windowmoved` is received - window is centered on the desktop
- `windowexposed` is received
- `windowenter` is received
- `windowfocusgained` is received
- if mouse moves, `mousemotion` is received, `mouse_pos` is updated
- if any of mouse buttons is pressed, `mousebuttondown` is received, `mouse_pos` is updated, `on_mouse_down` is dispatched

If app starts with cursor outside of the window:
- `windowmoved` is received
- `windowexposed` is received
- `windowfocusgained` is received
- if cursor is moved on window, `windowenter` is received
- `mousemotion` is received, `mouse_pos` is updated


After this on both platforms, moving cursor outside of the window (or minimizing it) will receive `windowleave` and moving cursor on the window will receive `windowenter`. 

For drop text/file case, on Linux we have `windowenter` and then drop events, on Windows we have drop events and then `windowenter` and `mousemotion`.

kivy/core/window/window_sdl2.py
pythonic64: SDL does not report `mousemotion` events and therefore `mouse_pos` does not get updated before `dropxxx` events. On Linux `windowenter` is received first and then `dropxxx` events, but on Windows `dropxxx` event will be first and then `windowenter`. 

I don't know if non-mouse devices can trigger `dropxxx` events from SDL. Also, how to handle the case when app is paused and receives `dropxxx` events?

pythonic64:
Another approach would be to expose `get_nearest_mouse_pos` and `get_global_mouse_pos` method to `WindowBase` and let the user decide when to use them. Doc of `mouse_pos` would state that property is updated on mouse motion only and for other events like `on_drop_begin` and `on_mouse_enter` user should call `get_nearest_mouse_pos` and `get_global_mouse_pos` to get the correct mouse pos.

kivy/core/window/window_sdl2.py
matham: Looking at the code in sdl2, there's no direct connection between a drop event and the mouse. However, it would be better if it was a mouse event that caused it to keep the mouse position updated. So, what you're doing is good, the only change I would make is to check if the mouse is actually in the window, and only update the position then.

pythonic64: We can check if mouse cursor is within window bounds, but I don't know how could we check if the drop is caused by the mouse.

matham: We don't have to check whether it's caused by a mouse. The goal is to make sure the mouse is up to date. So if it's not caused by the mouse but the mouse is in the window, then it's ok if we still update it (although it'd probably keep the same value). The thing to avoid is updating the mouse when it's outside the window.

pythonic64: Ok. I'll add the changes.

pythonic64: Problem with "if within the window" is that on Windows user can drop a file on title bar and on window edges, but those are not within the window bounds so old `mouse_pos` remains.

matham: On the edge/title bar counts as being outside the window, right? Then maybe we need to update the mouse pos to `-1` when it exits the screen, so when we drop it on the edges the value will be negative indicating it wasn't dropped inside.

p0lygun: How is this supposed to behave with custom title bar ? As in that there is no title bar or border

matham: If there's no title bar, then it can be dropped on the edge or bar, meaning it will be the same as dropping in the center of the window.

p0lygun: sdl might take care of it by itself will test it once the or Is ready for the final merge

pythonic64: Negative values for `mouse_pos` would be a breaking change.

We can add `is_mouse_pos_within_window` method to `WindowBase` and note in `on_drop_begin` about edges/title bar case on Windows.

pythonic64: @p0lygun With custom title bar `mouse_pos` will be updated with correct value because title bar is within the window.

matham: I dislike adding a new function just for this. In the ideal API, we should not have gotten the pos from `mouse_pos`, but the drop event would have included the dropped position in the event dispatch. How about resetting it to zero when it leaves the window?

pythonic64: I would go with dispatching it with `on_drop_begin` and updating `mouse_pos` if it is within the window. Resetting to zero would indicate a mouse_pos changed (within the window) but it didn't.

pythonic64: And mouse pos value dispatched with `on_drop_begin` would be relative to window position (left-top corner without title bar).

matham: Yeah, that sounds good!

pythonic64: I've discovered that if:
- the cursor is on the title bar or the edges
- mouse wheel scroll up/down
- `windowenter` is received
- `mousewheel` is received

In this case `mouse_pos` would be out of the window bounds, so I've added a test to skip `mousewheel` and `windowenter` if cursor is not within the window. This could be the issue in the SDL.

pythonic64: To add, click on the title bar or mouse motion over it does not trigger any events.

matham (approved):

pythonic64:
Should we dispatch mouse position with other `on_drop_xxx` events? We can reuse the same one from the `on_drop_begin`.

matham:
Yeah, good idea.

PR #7789, opened 2022-02-04
Updated url for editable install docs

p0lygun:
the latest URL now points to setup tools development mode

Maintainer merge checklist


matham:
I'd prefer we keep the original link to pip because that includes the meta info about how to use the command, as opposed to the link you are replacing it with that is basically a wall of text describing how it work. In particular since the pip link already links to the link you're replacing it with.

PR #7790, opened 2022-02-04
made code examples user friendly; fixes #7720 [WIP]

### LINKED ISSUE #7720: kivy.org code examples are not user friendly
https://kivy.org/doc/stable/tutorials/firstwidget.html
you cant just copy the code snippets without  line numbers.

p0lygun:
Currently when copying from a `include` source code the line numbers are copied, this PR fixes that 
## current 
image
## After fix
image

Maintainer merge checklist

### DIFF
--- a/doc/sources/.static/fresh.css
+++ b/doc/sources/.static/fresh.css
@@ -558,6 +558,9 @@ code.docutils{
 	border-radius: .5em;
 }
 
+.linenos::before {
+	content: attr(data-line-number);
+}
 
 @media (min-width: 640px) {
 
@@ -634,15 +637,6 @@ code.docutils{
 		font-size: 0.9em;
 	}
 
-}
-.linenos::before {
-	line-height: 0;
-	font-size: 14px;
-	content: attr(data-line-number);
-}
-.linenos {
-	font-size: 0;
-	line-height: 80%;
 }
 
 @media (min-width: 1280px) {
--- a/doc/sources/.static/fresh.css
+++ b/doc/sources/.static/fresh.css
@@ -635,13 +635,14 @@ code.docutils{
 	}
 
 }
-span.linenos {
-    -webkit-touch-callout: none;
-    -webkit-user-select: none;
-    -khtml-user-select: none;
-    -moz-user-select: none;
-    -ms-user-select: none;
-    user-select: none;
+.linenos::before {
+	line-height: 0;
+	font-size: 14px;
+	content: attr(data-line-number);
+}
+.linenos {
+	font-size: 0;
+	line-height: 80%;
 }
 
 @media (min-width: 1280px) {
--- a/doc/sources/.static/fresh.css
+++ b/doc/sources/.static/fresh.css
@@ -636,7 +636,12 @@ code.docutils{
 
 }
 span.linenos {
-	user-select: none;
+    -webkit-touch-callout: none;
+    -webkit-user-select: none;
+    -khtml-user-select: none;
+    -moz-user-select: none;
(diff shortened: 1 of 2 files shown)


misl6 (commented):
`user-select` only disables selection, but unfortunately doesn't prevent copying. (At least on Webkit browsers)

doc/sources/.static/kivy.js
```diff
@@ -224,5 +224,12 @@ $(function() {
...
 	}
-
+	// hack to add a attrib to all linenos
+    $(".linenos").each(function(){
+        let $this = $(this);
+        $this.attr("data-line-number", $this.text());
+		$this.contents().filter(function(){
+    			return this.nodeType === 3; //https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType
+		}).remove();
+  });
```
misl6, line 234: Not a super-fan of this kind of hack. Maybe we will change it in the future,  but right now, does its job (at least on `kivy.org`), so I'm fine with it.

doc/sources/.static/kivy.js
```diff
@@ -224,5 +224,12 @@ $(function() {
 		var section_title = $('li.toctree-l1.current > a').text();
 		$('div.body h1:eq(0)').prepend(section_title + ' &raquo; ');
 	}
-
+	// hack to add a attrib to all linenos
```
misl6, line 227: Can we clarify why is that needed?

doc/sources/.static/fresh.css
```diff
@@ -635,6 +635,15 @@ code.docutils{
...
 }
+.linenos::before {
+	line-height: 0;
+	font-size: 14px;
+	content: attr(data-line-number);
+}
+.linenos {
+	font-size: 0;
+	line-height: 80%;
+}
```
misl6, line 646: Is really needed?

doc/sources/.static/fresh.css
```diff
@@ -635,6 +635,15 @@ code.docutils{
 	}
 
 }
+.linenos::before {
+	line-height: 0;
+	font-size: 14px;
```
misl6, line 640: Are `line-height` and `font-size` needed?

doc/sources/.static/fresh.css
```diff
@@ -635,6 +635,15 @@ code.docutils{
 	}
 
 }
+.linenos::before {
```
misl6, line 638: Can be moved on top of the `@media` queries? (Just after `code.docutils`)

misl6 (changes requested):
Tested on Safari, Chrome, Firefox and the hack works as expected.
I've added some comments.

doc/sources/.static/kivy.js
p0lygun: the comment ?

doc/sources/.static/fresh.css
p0lygun: wait actually no, as there is no text in linenos now i'll remove it

doc/sources/.static/kivy.js
misl6: Nope, the comment is fine, just needs to be a little bit verbosely about it.
As an example (but feel free to change it): "Hack to add an attrib to all linenos, this is needed in order to avoid that certain browsers (Ex: Safari) treat the line-number as a copyable item."

p0lygun: ahh i see u mean it should be more descriptive, correct

misl6 (approved):

PR #7793, opened 2022-02-07
WindowBase|WindowSDL: Add drop position for all on_drop_xxx events

pythonic64:
Adds dispatching of the drop position (in coords relative to the window left/top) with all `on_drop_xxx` events. Documentation about the events is updated. This is a follow-up on https://github.com/kivy/kivy/pull/7786.

Maintainer merge checklist:

### DIFF
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -159,8 +159,9 @@ class WindowSDL(WindowBase):
 
     def __init__(self, **kwargs):
         self._pause_loop = False
-        self._win = _WindowSDL2Storage()
         self._cursor_entered = False
+        self._drop_pos = None
+        self._win = _WindowSDL2Storage()
         super(WindowSDL, self).__init__()
         self.titlebar_widget = None
         self._mouse_x = self._mouse_y = -1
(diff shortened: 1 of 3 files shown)


kivy/core/window/window_sdl2.py
```diff
@@ -772,18 +775,21 @@ def mainloop(self):
                 Logger.trace('WindowSDL: Unhandled event %s' % str(event))
 
     def _dispatch_drop_event(self, action, args):
+        x, y = (0, 0) if self._drop_pos is None else self._drop_pos
```
matham, line 778: I think we should bake in to the docs that the default if it's not in the window or if it's not a mouse but e.g. a touch, that it will be `-1` rather than `0`!?

pythonic64: Negative values are valid as they can be used to test if the drop occurred on the title bar or on the left window edget. Docs do state that `x` and `y` are the cursor positions at the time of the drop and SDL does not send those values with the `dropfile` event, only `filename` is sent so we don't know from where the drop originated.

matham (approved):

pythonic64:
Maybe we need more in the doc for case when/if the drop is not coming from the mouse.

matham:
>Maybe we need more in the doc for case when/if the drop is not coming from the mouse?

Possibly. But we can always update it later if we figure out it's possible to get a drop without the mouse!? It's up to you.

akshayaurora:
looks good,  LGTM!

matham:
@akshayaurora please make sure to squash before merging, especially if there are multiple commits that are not separate features.

@pythonic64 it may be a good idea to mark PRs as WIP until you feel it's ready to merge so we don't merge before you're ready if you're still considering changes. We can still review it before that though. There's a button that makes it a draft to block merging until you're ready!

pythonic64:
@matham At the time of the creation the request looked ready to me, but now I think we should have added `*args` to those events, just in case that SDL might pass a drop origin.

PR #7795, opened 2022-02-08
WindowBase: Add *args to on_drop_xxx events

pythonic64:
Add the `*args` to all `on_drop_xxx` event and update the documentation to clearly state that user should not rely on `x` and `y` arguments if the drop is not from the mouse.

Follow up on https://github.com/kivy/kivy/pull/7793

Maintainer merge checklist:

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -310,13 +310,13 @@ class WindowBase(EventDispatcher):
                 The *unicode* parameter has be deprecated in favor of
                 codepoint, and will be removed completely in future versions.
 
-        `on_drop_begin`: x, y
+        `on_drop_begin`: x, y, *args
             Fired when text(s) or file(s) drop on the application is about to
             begin.
 
             .. versionadded:: 2.1.0
 
-        `on_drop_file`: filename (bytes), x, y
+        `on_drop_file`: filename (bytes), x, y, *args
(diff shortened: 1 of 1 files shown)


kivy/core/window/__init__.py
```diff
@@ -2032,17 +2032,25 @@ def on_textinput(self, text):
...
+        mouse.
+
         :Parameters:
             `x`: `int`
-                Cursor x position, relative to window :attr:`left`.
+                Cursor x position, relative to the window :attr:`left`, at the
+                time of the drop.
             `y`: `int`
-                Cursor y position, relative to window :attr:`top`.
+                Cursor y position, relative to the window :attr:`top`, at the
```
matham, line 2050: Is it really relative to the top? E.g. the bottom left corner is `-h`?

pythonic64: It's the same origin (0, 0) position as for the `on_mouse_xxx` events, which is the window left/top.

matham: Ok, didn't remember that that was the origin.

matham (approved):

kivy/core/window/__init__.py
pythonic64: It is and imho it should be changed to left/bottom be inline with the touch/motion events, but that would be a breaking change.

PR #7799, opened 2022-02-10
Dark Theme support for docs

p0lygun:
# TODO
- [x] add a switch to toggle between themes
- [x] use localstorage to store user preference
- [x] decide best way to switch theme

### Light Theme
<details>  
<summary> Click to show </summary>  

image

</details>  

### Dark Theme
<details>
<summary> Click to show </summary>  

image  

</details>

Maintainer merge checklist


doc/sources/.static/dark.css
```diff
@@ -0,0 +1,514 @@
...
+}
+::selection {
+    background-color: #004daa !important;
+    color: #e8e6e3 !important;
+}
+::-moz-selection {
+    background-color: #004daa !important;
+    color: #e8e6e3 !important;
+}
+
```
misl6, line 56: Did you add these cause you noticed an issue with `color-scheme` ?

doc/sources/.static/dark.css
```diff
@@ -0,0 +1,514 @@
...
+    background-color: rgb(36, 39, 41);
+    background-image: none;
+    color: rgb(232, 230, 227);
+}
+@media (min-width: 1024px) {
+    .toc {
+        background-color: rgb(30, 32, 33);
+        border-color: rgb(57, 61, 64);
+    }
+}
```
misl6, line 268: This one shouldn't be needed cause the one without the media query gets applied also on `width>=1024px`.

doc/sources/.static/dark.css
```diff
@@ -0,0 +1,514 @@
+html {
+    background-color: #181a1b !important;
+}
```
misl6, line 3: Is this one needed?

misl6 (commented):
Nice work. I really like the color palette.

I've left some comments in order to simplify the CSS file, but I expect to add more during the review.

Instead of having an additional CSS, is better to include these rules in the default one, and then act as follow:

If the `dark` theme is either selected from the toggle button, or by browser configuration, we add a class on the `html` tag that selects the `dark` theme:
```html
<html class="dark-theme">
```

```css
.dark-theme html {
    color-theme: dark;
}
```

Also, in order to keep the CSS file small and clean, is better to use [Custom Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*https://developer.mozilla.org/en-US/docs/Web/CSS/--*).

doc/sources/.templates/layout.html
```diff
@@ -177,6 +185,14 @@ <h3>Related Topics</h3>
                   <li><a class="nav-api" href="{{ pathto('api-kivy') }}">API Reference</a></li>
                   <li><a class="nav-pdf" href="https://media.readthedocs.org/pdf/kivy/latest/kivy.pdf">PDF</a></li>
                   <li><a class="nav-wiki" href="http://wiki.kivy.org">Wiki</a></li>
+                  <li>
+                      <div id="themeToggleSwitchContainer">
```
p0lygun, line 189: Need some ideas for how the switch should look 
currently 
image

doc/sources/.templates/layout.html
```diff
@@ -158,6 +159,13 @@ <h3>Related Topics</h3>
 {%- endblock %}
 {%- block extrahead %} {% endblock %}
   </head>
+  <script>
+    var link = document.createElement('link');
+        link.rel = 'stylesheet';
+        link.href = "{{ pathto('_static/dark.css', 1) }}";
+            link.id = "darkModeS";
+        document.head.appendChild(link);
+    </script>
```
p0lygun, line 168:
```suggestion

```
remove unnecessary and outdated code

doc/sources/.static/fresh.css
```diff
@@ -1,157 +1,293 @@
+:root {
```
p0lygun, line 1: minify all html on production ?

p0lygun (commented):
@misl6 some thoughts on these pls

doc/sources/.static/fresh.css
```diff
@@ -179,387 +315,768 @@ a:hover {
...
+    color: var(--highlight-c);
+}
+
+.highlight .c1 {
+    color: var(--highlight-c);
+}
+
+.highlight .cs {
+    color: var(--highlight-c);
+}
```
misl6, line 839: Can we merge these rules?

doc/sources/.static/fresh.css
```diff
@@ -179,387 +315,768 @@ a:hover {
...
+    color: var(--neutral-secondary-text);
+}
+
+.highlight .kt {
+    color: var(--neutral-secondary-text);
+}
+
+.highlight .ld {
+    color: var(--neutral-secondary-text);
+}
```
misl6, line 907: Can we merge these rules?

doc/sources/.templates/layout.html
misl6: Just thinking out loud:

- Instead of placing it vertically,  can be placed horizontally with a couple of icons? (sun and moon)
- Desktop sizes: placed at the left of the wiki button
- Smartphone sizes: Into this area?
Image 11-02-22 at 22 15

misl6 (changes requested):
Looks good!

I left some comments regarding DRY violations, I only commented on some examples. Can you take care of it?

doc/sources/.static/fresh.css
```diff
@@ -1,641 +1,938 @@
...
+        border: 1px solid var(--dark-border-color);
+        padding: 10px;
+        float: right;
+        margin: 0 0 10px 10px;
+        font-size: .9em
+    }
+
+    .toc ul li li li {
+    border-left-color: var(--dark-border-color);
+}
```
misl6, line 935: Why can't be placed outside of the media query?

p0lygun: its how it is currently 
https://github.com/kivy/kivy/blob/master/doc/sources/.static/fresh.css#L630
and i forgot to add 
https://github.com/kivy/kivy/blob/master/doc/sources/.static/fresh.css#L630

p0lygun: nvm its there

misl6 (changes requested):

doc/sources/.static/kivy.js
```diff
@@ -233,4 +233,19 @@ $(function() {
...
-});
+
+	// to have the current "active" element be always be centred on the left side bar
+	const currApiIndex_li = $(".api-index li.current");
+	if (currApiIndex_li.length){
+		const target = currApiIndex_li[0];
+		$(target.parentNode).animate({"scrollTop": target.offsetTop - 2*target.parentNode.offsetTop }, 'normal');
+	}
+
+	document.documentElement.className = localStorage.getItem('theme') || "light";
```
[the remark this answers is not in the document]
p0lygun: using `(prefers-color-scheme` and `matchMedia` ?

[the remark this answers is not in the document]
misl6: Yes, and that line, that doesn't uses jQuery could also be placed into the <head>, in order to avoid that at refresh, the page doesn't flash between the two themes.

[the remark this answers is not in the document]
misl6: Adding:
```html
<script>
     const currTheme = localStorage.getItem("theme");
     if(currTheme){
        document.documentElement.className = currTheme;
     } else if(window.matchMedia){
        document.documentElement.className = window.matchMedia('(prefers-color-scheme: dark)').matches ? "dark": "light";
     } else { // neither localStorage or window.matchMedia exists
        document.documentElement.className = "light"; //todo should dark theme be the fallback value instead ?
     }
</script>
````

just after the line `122` of `layout.html` and deleting the related part in `js` code should fix the issue.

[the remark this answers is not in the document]
p0lygun: what if we add it above line `97` ?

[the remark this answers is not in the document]
misl6: Better to keep it inside the `{%- block scripts %}` block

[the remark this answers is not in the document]
p0lygun: i did but the code in not being included into the final rendered page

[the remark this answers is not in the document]
p0lygun: nvm it was supposed to be on LN `124`

p0lygun (commented):
@misl6 PR up for final review

PR #7804, opened 2022-02-14
Use the `KIVY_RPI_VERSION` env variable to force the build of `egl_rpi` in non Raspi CI builds

### LINKED ISSUE #7797: Setup Platform Override for Raspberry Pi CI Builds
**Is your feature request related to a problem? Please describe.**

The Raspberry Pi support is only enabled if `setup.py` detects it's running on a Raspberry Pi via the `/opt/vc/include/bcm_host.h` file, but then goes on to check the `/proc/cpuinfo` to identify that it's not running on a Raspberry Pi 4 hardware. That is fair enough, but does not detect correctly in the `pi-gen` environment as the CPU is not a Raspberry Pi.

There is the flag / env variable `use_rpi`/`USE_RPI`, which can be used to force cythonization of the `egl` and `bcm` modules, but the compilation fails, as the [setup.py:631-663](https://github.com/kivy/kivy/blob/master/setup.py#L631-L663) requires `platform` to be set to `rpi` and ignores `use_rpi`.

Allow overriding the `platform` in setup.py via ENV variable, or include a check for `use_rpi`, so the compilation works.

There should be a warning in the Docs that the Raspberry Pi version with `egl_rpi` support _must_ be compiled locally on a Raspberry Pi at the moment.
If the solution as described above is implemented, the docs could instead mention to use the override flag for a QEMU enabled CI build.

**Describe alternatives you've considered**

* Building on the Raspberry Pi was the originally foreseen use case for this, but is not practical for creating a pre-built image.
* I could prebuild Kivy on a Raspberry Pi and copy the build outputs over / put them in a repo somewhere and use them from there.

I will try the above suggestion and submit a pull request.

**Additional context**
I am trying to build a Raspberry Pi image based on the raspberrypi/pi-gen toolchain. The image is built in a container on some x64 server using QEMU's emulation for ARM processors in order to build a disk image for a Raspberry Pi. This is what is also used for building the official Raspbian / RaspberryOS images.

Installing `kivy[base]` as wheel comes without `egl_rpi` python prebuilts, so I can't use it for my current use case.

peanball:
Forces the `pi_version` variable in `setup.py` to a user provided value when the `FORCE_RPI_VERSION` environment variable is set.

Resolves #7797.

A new environment variable is introduced that allows circumventing the Raspberry Pi version detection on non-raspberry Pi hardware that still has the full Raspberry Pi software stack.

The [RPi-Distro/Pi-gen](https://github.com/RPi-Distro/Pi-gen) project is used for building bootable images for Raspberry Pi and internally uses `qemu-static` to execute ARM binaries on x86 hardware in order to run compilers, packages, etc. in the correct format. 

Currently, the `setup.py` uses the `/proc/cpuinfo` pseudofile to determine, which Raspberry Pi version it is running on. In the `pi-gen` environment, this pseudofile is used from the host machine (or Docker container) though, and will not contain any Raspberry Pi relevant CPU information.

By forcing the Raspberry Pi version, the rest of setup.py behaves as if it is running on a raspberry pi, and given all the required dependencies creates a runnable image for Raspberry Pi with full `egl_rpi` window support, GLES2, etc. This is suitable for Raspberry Pi < 4, as `egl_rpi` is not supported from Raspberry Pi 4 on. The version check for Raspberry Pi < 4 remains intact, even with the override.

When the `FORCE_RPI_VERSION` is not set there is no change in behaviour.

Maintainer merge checklist

### DIFF
--- a/doc/sources/installation/installation-rpi.rst
+++ b/doc/sources/installation/installation-rpi.rst
@@ -83,15 +83,17 @@ Cross-Compilation for Raspberry Pi 1-3 headless installation on Raspbian Buster
 The Raspberry OS project uses `pi-gen` project to create bootable images for Raspberry PI.
 
 Kivy determines automatically the sub-packages to build based on the environment it is compiled within. By default, the `egl_rpi` renderer that uses the (now deprecated but still useful) DISPMANX API is only compiled when running on a Raspberry Pi.
+In order to build Kivy in such `pi-gen` environment, the auto-detection of the Raspberry Pi hardware version needs to be disabled.
 
-When cross-compiling using e.g. `pi-gen`, the build system can be forced into compiling for Raspberry Pi with `egl_rpi` support by setting the environment variabel `USE_RPI` to `1`.
+When cross-compiling using e.g. `pi-gen`, the build system can be forced into compiling for Raspberry Pi with `egl_rpi` support by setting the environment variabel `FORCE_RPI_VERSION` to any number < 4, e.g. `3`.
 
 The install command then looks something like this::
 
-    apt install build-essential libraspberrypi-dev raspberrypi-kernel-headers 
-    USE_RPI=1 python -m pip install "kivy[base]" kivy_examples --no-binary kivy
+    apt install build-essential libraspberrypi-dev raspberrypi-kernel-headers
+    FORCE_RPI_VERSION=3 python -m pip install "kivy[base]" kivy_examples --no-binary kivy
 
 Please note that the `egl_rpi` window handler is not supported on Raspberry Pi 4 and higher.
+The existing version check will refuse to compile the `egl_rpi` provider when detecting or forcing the Raspberry Pi version to 4 or higher.
 
 Raspberry Pi 4 headless installation on Raspbian Buster
 *******************************************************
--- a/doc/sources/installation/installation-rpi.rst
+++ b/doc/sources/installation/installation-rpi.rst
@@ -77,6 +77,21 @@ then install SDL2 from apt::
 
     sudo apt install libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev libsdl2-ttf-dev
 
+Cross-Compilation for Raspberry Pi 1-3 headless installation on Raspbian Buster
+*******************************************************************************
+
+The Raspberry OS project uses `pi-gen` project to create bootable images for Raspberry PI.
+
+Kivy determines automatically, which libraries to add based on the environment it is run. The `egl_rpi` renderer that uses the (now deprecated but still useful) DISPMANX API is only compiled in when running on a Raspberry Pi.
+
+When cross-compiling using e.g. `pi-gen`, the build system can be forced into compiling for Raspberry Pi with `egl_rpi` support by setting the environment variabel `USE_RPI` to `1`.
+
+The install command then looks something like this::
+
+    apt install build-essential libraspberrypi-dev raspberrypi-kernel-headers 
+    USE_RPI=1 python -m pip install "kivy[base]" kivy_examples --no-binary kivy
+
+Please note that the `egl_rpi` window handler is not supported on Raspberry Pi 4 and higher.
 
 Raspberry Pi 4 headless installation on Raspbian Buster
 *******************************************************
--- a/doc/sources/installation/installation-rpi.rst
+++ b/doc/sources/installation/installation-rpi.rst
@@ -82,7 +82,7 @@ Cross-Compilation for Raspberry Pi 1-3 headless installation on Raspbian Buster
 
 The Raspberry OS project uses `pi-gen` project to create bootable images for Raspberry PI.
 
-Kivy determines automatically, which libraries to add based on the environment it is run. The `egl_rpi` renderer that uses the (now deprecated but still useful) DISPMANX API is only compiled in when running on a Raspberry Pi.
+Kivy determines automatically the sub-packages to build based on the environment it is compiled within. By default, the `egl_rpi` renderer that uses the (now deprecated but still useful) DISPMANX API is only compiled when running on a Raspberry Pi.
 
 When cross-compiling using e.g. `pi-gen`, the build system can be forced into compiling for Raspberry Pi with `egl_rpi` support by setting the environment variabel `USE_RPI` to `1`.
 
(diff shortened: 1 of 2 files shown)


doc/sources/installation/installation-rpi.rst
```diff
@@ -77,6 +77,21 @@ then install SDL2 from apt::
 
     sudo apt install libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev libsdl2-ttf-dev
 
+Cross-Compilation for Raspberry Pi 1-3 headless installation on Raspbian Buster
+*******************************************************************************
+
+The Raspberry OS project uses `pi-gen` project to create bootable images for Raspberry PI.
+
+Kivy determines automatically, which libraries to add based on the environment it is run. The `egl_rpi` renderer that uses the (now deprecated but still useful) DISPMANX API is only compiled in when running on a Raspberry Pi.
```
matham, line 85:
```suggestion
Kivy determines automatically the sub-packages to build based on the environment it is compiled within. By default, the `egl_rpi` renderer that uses the (now deprecated but still useful) DISPMANX API is only compiled when running on a Raspberry Pi.
```

doc/sources/installation/installation-rpi.rst
```diff
@@ -77,6 +77,21 @@ then install SDL2 from apt::
...
+
+When cross-compiling using e.g. `pi-gen`, the build system can be forced into compiling for Raspberry Pi with `egl_rpi` support by setting the environment variabel `USE_RPI` to `1`.
+
+The install command then looks something like this::
+
+    apt install build-essential libraspberrypi-dev raspberrypi-kernel-headers 
+    USE_RPI=1 python -m pip install "kivy[base]" kivy_examples --no-binary kivy
```
matham, line 92: Would you not also need to specify `KIVY_CROSS_SYSROOT`? Or does that work automatically?

matham (changes requested):

doc/sources/installation/installation-rpi.rst
peanball: It works automatically because the complete user space is a Raspbian when using the Pi-gen project, including BCM headers and libs. It just happens to run in a docker container via QEMU. So the detection of all libraries (e.g. bcm_host.h, egl & bcm libs) works like on a real raspberry pi.

peanball: It would not work correctly on a 4, you're right. Forcing the raspberry pi version instead, possibly with a dedicated env variable would be cleaner. I will try it out and update the PR.

setup.py
```diff
@@ -112,9 +112,17 @@ def get_isolated_env_paths():
...
 # proprietary broadcom video core drivers
 if exists('/opt/vc/include/bcm_host.h'):
+    used_pi_version = pi_version
+    # Force detected Raspberry Pi version for cross-builds, if needed
+    if 'FORCE_RPI_VERSION' in environ:
+        try:
+            used_pi_version = int(environ.get('FORCE_RPI_VERSION'))
+        except:
+            print("Could not override Raspberry Pi version "
+                 f"to {environ.get('FORCE_RPI_VERSION')}")
```
matham, line 122:
```suggestion
    if 'KIVY_RPI_VERSION' in environ:
        used_pi_version = int(environ['KIVY_RPI_VERSION'])
```

doc/sources/installation/installation-rpi.rst
```diff
@@ -77,6 +77,23 @@ then install SDL2 from apt::
...
 
+Cross-Compilation for Raspberry Pi 1-3 headless installation on Raspbian Buster
+*******************************************************************************
+
+The Raspberry OS project uses `pi-gen` project to create bootable images for Raspberry PI.
+
+Kivy determines automatically the sub-packages to build based on the environment it is compiled within. By default, the `egl_rpi` renderer that uses the (now deprecated but still useful) DISPMANX API is only compiled when running on a Raspberry Pi.
+In order to build Kivy in such `pi-gen` environment, the auto-detection of the Raspberry Pi hardware version needs to be disabled.
+
+When cross-compiling using e.g. `pi-gen`, the build system can be forced into compiling for Raspberry Pi with `egl_rpi` support by setting the environment variabel `FORCE_RPI_VERSION` to any number < 4, e.g. `3`.
```
matham, line 88:
```suggestion
When cross-compiling using e.g. `pi-gen`, the build system can be forced into compiling for Raspberry Pi with `egl_rpi` support by setting the environment variabel `KIVY_RPI_VERSION` to any number < 4, e.g. `3`.
```

doc/sources/installation/installation-rpi.rst
```diff
@@ -77,6 +77,23 @@ then install SDL2 from apt::
...
+
+The install command then looks something like this::
+
+    apt install build-essential libraspberrypi-dev raspberrypi-kernel-headers
+    FORCE_RPI_VERSION=3 python -m pip install "kivy[base]" kivy_examples --no-binary kivy
```
matham, line 93:
```suggestion
    KIVY_RPI_VERSION=3 python -m pip install "kivy[base]" kivy_examples --no-binary kivy
```

matham (changes requested):
`KIVY_RPI_VERSION` is a better name for the  variable.

Also, there's no reason to do the try/except there. If there's a problem reading the variable as an int, python will show an appropriate exception, depending on the problem. But in general, a bare except is not a good idea.

doc/sources/installation/installation-rpi.rst
matham: You forgot to include these changes.

peanball: Yes, still working on it and has marked the PR as draft. I had/have very little time and was also running a test in my build pipeline for these changes to make sure they do what I want. Once it's ready I'm happy to move it from `Draft` and re-request your review again.

Thanks for the review and patience :)

peanball:
> `KIVY_RPI_VERSION` is a better name for the variable.
> 
> Also, there's no reason to do the try/except there. If there's a problem reading the variable as an int, python will show an appropriate exception, depending on the problem. But in general, a bare except is not a good idea.

Yes, fully agree with the name and that the bare except was not a good idea.

Thanks again for your review. I've now applied all the changes and consider the PR ready from my side.

matham (approved):

PR #7808, opened 2022-02-20
Add the LINK_SUFFIX to Sphinx template to restore search result links

peanball:
Corrects the broken page links on the search results page, which contain `undefined` because the LINK_SUFFIX is currently not set.

The setting was introduced with Sphinx 0.6.0 and likely the template was not adapted. This PR fixes that.

Tested by:
- not changing any settings in the `docs/conf.py` and getting links ending with `.html` (instead if `undefined`, confirming the fix)
- setting `html_link_suffix` to `.link` in the `doc/conf.py` getting links ending in `.link`
- setting `html_file_suffix` to `.xhtml` in the `doc/conf.py` getting links ending in `.xhtml`.

When no setting is provided in the conf.py, sphinx defaults to `.html`, which is the desired and expected behaviour.

Maintainer merge checklist

### DIFF
--- a/doc/sources/.templates/layout.html
+++ b/doc/sources/.templates/layout.html
@@ -113,6 +113,7 @@
                   LANGUAGE:'{{ language }}', 
                   COLLAPSE_INDEX:false, 
                   FILE_SUFFIX:'{{ '' if no_search_suffix else file_suffix }}', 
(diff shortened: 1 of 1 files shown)


peanball:
Already fixed in `master`...

PR #7810, opened 2022-02-23
Track whether the clock has started

matham:
Maintainer merge checklist

This adds a variable that tracks whether the kivy clock has started. Previously I added a similar variable to track whether the clock has ended. This is required for interfacing with async code (and threading) where we need to be sure that the async code or thread is not stuck forever waiting for kivy to finish a callback that will never happen because the clock has not started (and potentially never will) or if it finished.

Introducing `has_ended` allowed us to be notified by special callback when the clock finishes allowing any waiting async code that are waiting on a kivy callback to be canceled/timed out. By adding `has_started`, it will enable async code to raise an error if it checks that the clock has not started, meaning it won't be safe for it to schedule a callback and wait, as it may wait forever.

While this is technically a new feature, I would really like if we can get this into 2.1.0 because it's needed for my [kivy_trio](https://github.com/matham/kivy-trio/) library that interfaces kivy with trio async code. And it's not really much of a functional change (it's not even publicly documented as I don't expect users to use it) so it can't break anything existing. But if people think it's not a good idea we can wait until after the release.

### DIFF
--- a/kivy/_clock.pxd
+++ b/kivy/_clock.pxd
@@ -95,6 +95,7 @@ cdef class CyClockBase(object):
     cdef public object _lock_acquire
     cdef public object _lock_release
 
+    cdef public int has_started
     cdef public int has_ended
     cdef public object _del_safe_lock
     cdef public int _del_safe_done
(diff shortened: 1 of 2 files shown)


misl6:
We're in feature-freeze, but IMHO we can make exceptions for non-impacting changes such as this one, so looks good to me.

Issue #7755, opened 2022-01-18
Video Player Widget Not Displaying Thumbnail if its not in the same directory with the e same filename as video file


Noor-Kalibbala:
**Software Versions**
* Python:
* OS:
* Kivy:
* Kivy installation method:

Am trying to play a video and display its thumbnail before it starts but the thumbnail is not displayed at first, but when i create the thumbnail with the same video file name within the folder , then the thumbnail is displayed. i dont know why the video player is behaving strangely like this
```
import kivy
kivy.require('1.8.0')

from kivy.app import App
from kivy.lang import Builder

root = Builder.load_string('''
VideoPlayer:
    source: '/home/noor/Videos/sunshine.mp4'
    thumbnail:"/home/noor/Videos/screen.png"
''')

class TestApp(App):
    def build(self):
        return root

if __name__ == '__main__':
    TestApp().run()
```

Noor-Kalibbala:
hello @misl6 
Adding the video widget like in the example below does not display the thumbnail
```
widget_to_add = VideoPlayer(source=value[0], thumbnail=thumbnail_path)
                print(thumbnail_path)
                widget_to_add.allow_stretch = True
                # add it to the boxlayout
                self.ids.media.add_widget(widget_to_add)
```

PR #7816, opened 2022-02-26
VideoPlayer: Defer before the next frame the default thumbnail and annotations loading

misl6:
Fixes issue #7755 

Maintainer merge checklist

### DIFF
--- a/kivy/uix/videoplayer.py
+++ b/kivy/uix/videoplayer.py
@@ -488,8 +488,6 @@ class VideoPlayer(GridLayout):
         self._annotations = ''
         self._annotations_labels = []
         super(VideoPlayer, self).__init__(**kwargs)
-        self._load_thumbnail()
-        self._load_annotations()
 
         if self.source:
             self._trigger_video_load()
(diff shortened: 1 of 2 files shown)


kivy/uix/videoplayer.py
```diff
@@ -501,47 +499,58 @@ def _trigger_video_load(self, *largs):
...
+        Clock.schedule_once(self._try_load_default_annotations, -1)
         if self._video is not None:
             self._video.unload()
             self._video = None
         if value:
             self._trigger_video_load()
 
+    def on_thumbnail(self, instance, value):
+        if value:
+            self._load_thumbnail(value)
```
matham, line 529: Do we need to reset to default if value is None? Or maybe clear it?

matham (changes requested):

kivy/uix/videoplayer.py
```diff
@@ -501,47 +499,58 @@ def _trigger_video_load(self, *largs):
...
+        # with the same filename (except extension) of the source (video) file.
+        Clock.schedule_once(self._try_load_default_thumbnail, -1)
+        Clock.schedule_once(self._try_load_default_annotations, -1)
         if self._video is not None:
             self._video.unload()
             self._video = None
         if value:
             self._trigger_video_load()
 
+    def on_thumbnail(self, instance, value):
```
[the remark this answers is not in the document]
misl6: IIRC: On docs, we have a warning regarding `on_xxx` inheritance
`fbind` looks a great approach for things inside the framework BTW, applied the fix.

kivy/uix/videoplayer.py
misl6: Now the user can clear it. Added a test that checks it.

kivy/uix/videoplayer.py
```diff
@@ -501,47 +504,57 @@ def _trigger_video_load(self, *largs):
...
             return
         self._annotations_labels = []
-        annotations = self.annotations
-        if not annotations:
-            filename = self.source.rsplit('.', 1)
-            annotations = filename[0] + '.jsa'
-        if exists(annotations):
-            with open(annotations, 'r') as fd:
-                self._annotations = load(fd)
+        with open(annotations, 'r') as fd:
```
matham, line 556: Couldn't `annotations` now be the empty string? If so we also need to check if it's empty!?

misl6: 🤦 You're right. Added the check and a test.

matham (dismissed):

PR #7820, opened 2022-03-01
Patch gst current release to look for dlls in correct place for win store

matham:
…tore.

Maintainer merge checklist

The current release of all deps, except for gsteamer is fixed to look for the dlls in the right place (https://github.com/kivy/kivy-sdk-packager/commit/fad7271a7c938473115a699cc278b132d5fb9365). We couldn't release gstreamer with the fix because we exceeded the pypi size. This patches it so the current gst release will have the fix. 

Additionally, we're deleting on pypi gst releases 0.3.0, 0.14, 0.15, and 0.16 as discussed on discord (they should not be used by anyone). This PR will serve as the backward compat warning to be included in the release notes that we deleted them.

### DIFF
--- a/kivy/__init__.py
+++ b/kivy/__init__.py
@@ -31,6 +31,7 @@ __all__ = (
 import sys
 import shutil
 from getopt import getopt, GetoptError
+import os
 from os import environ, mkdir
 from os.path import dirname, join, basename, exists, expanduser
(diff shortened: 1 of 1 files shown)


misl6 (approved):
Can't test on a Windows PC, but looks good to me.

PR #7846, opened 2022-03-14
Fix VKeyboard missing with custom keyboard class

Gobbel2000:
In `window/__init__.py` the VKeyboard class is only imported when needed, which only happens in `request_keyboard()`. If a custom keyboard class is set using `set_vkeyboard_class()`, it is never imported. In that case the function `_get_kivy_vkheight()` that was introduced in #7726 fails, because VKeyboard is still None. This change instead compares to `self._vkeyboard_cls` which is ensured above to not be None. If no custom class is set, this should still be just VKeyboard.

Maintainer merge checklist

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -647,7 +647,7 @@ class WindowBase(EventDispatcher):
             and self._vkeyboard_cls is not None
         ):
             for w in self.children:
-                if isinstance(w, VKeyboard):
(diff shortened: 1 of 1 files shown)


D4SK:
This fix works great for me as well. I hope this will be merged soon. Seems like a very safe change. Maybe @DexerBR can comment as he added the change that is affected.

@misl6 @pythonic64

misl6 (approved):
LGTM. Thank you!

PR #7663, opened 2021-10-17
Add python3.10 in the ci configuration [build wheel linux] [build wheel osx] [build wheel win]

tshirtman:
Maintainer merge checklist
* [ ] Title is descriptive/clear for inclusion in release notes.
* [ ] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [ ] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #7847, opened 2022-03-19
Tests: ffpyplayer now ships cp310-* and Apple Silicon compatible wheels, so tests on the full version can be re-introduced.

misl6:
`ffpyplayer` now ships `cp310-*` and `Apple Silicon` compatible wheels, so tests on the full version can be re-introduced.

Maintainer merge checklist

### DIFF
--- a/.github/workflows/test_osx_python.yml
+++ b/.github/workflows/test_osx_python.yml
@@ -45,13 +45,13 @@ jobs:
         source .ci/osx_ci.sh
         arm64_set_path_and_python_version ${{ matrix.python }}
         build_and_install_universal_kivy_sys_deps
-        install_kivy_test_run_pip_deps dev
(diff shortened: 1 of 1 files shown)


.github/workflows/test_osx_python.yml
```diff
@@ -45,13 +45,13 @@ jobs:
         source .ci/osx_ci.sh
         arm64_set_path_and_python_version ${{ matrix.python }}
         build_and_install_universal_kivy_sys_deps
-        install_kivy_test_run_pip_deps dev
```
matham, line 48: This doesn't do anything, does it?

misl6: Absolutely nothing. Is probably something we left into during https://github.com/kivy/kivy/pull/7663 tests

matham (approved):

PR #7857, opened 2022-04-04
iOS camera provider enhancements

RobertFlatt:
iOS camera provider enhancements, these are iOS only but implemented in the camera provider shared by iOS and MacOS.

The additional apis implement basic mobile camera functionality, this is not currently implemented.

To provide basic camera services on iOS, this PR adds
```
 def get_device_orientation(self):

 def change_camera_input(self, index):

 def zoom_level(self, level):

 def get_app_documents_directory(self):

 def save_texture(self, texture, filepath = '', quality = 0.9):
   # saves to iOS Photos App if filepath not specified,
   # or app documents directory if filepath is specified.
```

The fork also builds and runs on MacOS, indicating it is unlikely the changes have broken the MacOS camera provider.

TODO, but not part of this PR:

~~1) `kivy-ios` does not load the camera provider, the Camera widget fails. Assuming this PR is accepted in a timely fashion, I will implement `kivy-ios` PR that fixes the load and references this enhanced camera provider.~~

2) iOS has native photo and video capture apis, these are not currently implemented in this camera provider.

### DIFF
--- a/kivy/core/camera/camera_avfoundation_implem.m
+++ b/kivy/core/camera/camera_avfoundation_implem.m
@@ -5,6 +5,8 @@
  *
  * TODO:
  * - add iOS native photo and video capture
+ * - add support for telephoto and wide angle lenses
+ *     needs iOS10 AVCaptureDeviceDiscoverySession   discoverySessionWithDeviceTypes:mediaType:position: 
  * I've let the code concerning caps, even if it's not yet used. uncomment
  * WITH_CAMERA_CAPS to compile with it.
  */
--- a/kivy/core/camera/camera_avfoundation_implem.m
+++ b/kivy/core/camera/camera_avfoundation_implem.m
@@ -796,8 +796,8 @@ bool avf_camera_have_new_metadata(camera_t camera){
     return ((Camera *)camera)->haveNewMetadata();
 }
 
-bool avf_camera_set_video_orientation(camera_t camera, int orientation){
-    return ((Camera *)camera)->setVideoOrientation(orientation);
+void avf_camera_set_video_orientation(camera_t camera, int orientation){
+    ((Camera *)camera)->setVideoOrientation(orientation);
 }
 
 #if TARGET_OS_IPHONE || TARGET_IPHONE_SIMULATOR
--- a/kivy/core/camera/camera_avfoundation_implem.m
+++ b/kivy/core/camera/camera_avfoundation_implem.m
@@ -4,12 +4,8 @@
  * by Mathieu Virbel
  *
  * TODO:
- * - add interface for setting some capabilities as focus/exposure/...
- * I've let the code concerning caps, even if it's not yet used. uncomment
- * WITH_CAMERA_CAPS to compile with it.
+ * - add iOS native photo and video capture
  */
-
-//#define WITH_CAMERA_CAPS
 
 
 #import <AVFoundation/AVFoundation.h>
--- a/kivy/core/camera/camera_avfoundation_implem.m
+++ b/kivy/core/camera/camera_avfoundation_implem.m
@@ -5,15 +5,17 @@
  *
  * TODO:
  * - add iOS native photo and video capture
+ * I've let the code concerning caps, even if it's not yet used. uncomment
+ * WITH_CAMERA_CAPS to compile with it.
  */
 
+//#define WITH_CAMERA_CAPS
 
 #import <AVFoundation/AVFoundation.h>
 #import <Foundation/NSException.h>
 
 #if TARGET_OS_IPHONE || TARGET_IPHONE_SIMULATOR
 #import <UIKit/UIKit.h>
-#define WITH_CAMERA_CAPS
 #endif
 
 #ifdef WITH_CAMERA_CAPS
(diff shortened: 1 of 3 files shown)


RobertFlatt:
~~kivy-ios does not load the camera provider, the Camera widget fails. Assuming this PR is accepted in a timely fashion, I will implement kivy-ios PR that fixes the load and references this enhanced camera provider.~~

Not required with kivy-ios master.

misl6 (commented):
Code LGTM at a first glance!

I will test it on runtime (and will try to write a demo App, unless you already have one to share) during the upcoming weekend 😃

Super-minor: How about adding some logging when functionalities are not expected to be available but are called on macOS or vice-versa? Comments are super-fine, but users tend to not read them immediately. 😆

RobertFlatt:
This example uses the new features https://github.com/Android-for-Python/INDEX-of-Examples#c4k_photo_example

It depends on the Preview widget not the Camera widget. For iOS only, the **master** version of Preview is required https://github.com/Android-for-Python/Camera4Kivy . Release waits on this PR and a kivy-ios that points to the new Kivy master. 

The readme has not been updated for ios. Example shows switch between  cameras, capture photo or screenshot, pinch-spread for zoom (on my SE only the back camera), device rotation behavior, mirrored front camera. 

To enable use of the Camera 
```
        <key>NSCameraUsageDescription</key>
	<string> </string>
```
To enable save to Photos App
```
	<key>NSPhotoLibraryAddUsageDescription</key>
	<string> </string>
```
To make contents of local storage visible to File Manager
```
	<key>UIFileSharingEnabled</key>
	<true/>
        <key>LSSupportsOpeningDocumentsInPlace</key>
	<true/>
```
Image saves default to Photos App. The Preview widget storage selection is the same as for Android, so to save to app storage in place of Photos App  `capture_photo(location = 'private')` https://github.com/Android-for-Python/Camera4Kivy#location

[I started this because I wanted to try to port the tflite recipe, but my test example used a camera, so ...]

kivy/core/camera/camera_avfoundation_implem.m
```diff
@@ -426,13 +441,56 @@ The user can override these settings by calling the attemptCapturePreset() funct
...
+        if (camNum != _cameraNum) {
+            NSLog(@"Warning: Max Camera Num is %lu; Using camera %d\n",
+                  (unsigned long)([devices count] - 1), camNum);
+        }
+        device = [devices objectAtIndex:camNum];
+        cameraNum = _cameraNum;
+    } else {
+        device = [AVCaptureDevice defaultDeviceWithMediaType:AVMediaTypeVideo]  ;
+        cameraNum = 0;
+    }
```
misl6, line 470: The new  [discoverySessionWithDeviceTypes:mediaType:position:](https://developer.apple.com/documentation/avfoundation/avcapturedevicediscoverysession/2361539-discoverysessionwithdevicetypes) allows a fairly more granular selection of the camera. 

The `devices` array may be not consistent of returning the "correct" order of the cameras, so could be nice to create an enumerator on Kivy side, and pass the configuration requested from the user to  `discoverySessionWithDeviceTypes:mediaType:position:`

misl6 (changes requested):
Tested it on my iPhone 12 Pro, and generally works good!

Unfortunately, after some minor edits to your example and `Camera4Kivy`, I wasn't able to attach the ultra-wide or telephoto (back) lenses due to how camera selection is handled.

I've added some comments explaining what could be done to improve it.

RobertFlatt:
1) The api mentioned depends on ios10+ , (last time I looked) we build with minimum  9 not 10. And this is not adjustable https://github.com/kivy/kivy-ios/issues/691 On the subject of min api, the photo capture api gets good with ios11.

2) I don't understand enough about deviceTypes https://developer.apple.com/documentation/avfoundation/avcapturedevicetype or Apple devices  to implement this. It is not clear to me what the options do, or any side effects. 
I expect it is possible to synthesize some useful abstraction that extends Kivy's camera = [0,1] but at this point I don't know what it should be, or even what the usage model would be. So I would not try. 

3) This PR provides basic iOS functionality, which we don't have without the PR. We already know the camera provider could be improved, see the TODO at the top of the file. Support for ultrawide and telephoto should be added to that list.  Approve this PR as a starting point. Anybody who understands the additional enhancement design issues can then jump in.

misl6:
1) Yeah, the API mentioned depends on iOS 10+, and we're still targeting iOS 9 minimum. (Maybe is a good idea to higher that requirement?). But, considering that on iOS is always a good practice to check for the availability of APIs (or for the iOS version that introduced that feature), we can easily switch between the two implementations during runtime.

2) Decrypting Apple's docs is always a nightmare, so not your fault 😄 

3) I agree that this PR is an improvement to the current situation, and will merge it as-is if you don't expect to continue working on it in the near/mid-term. In that case, I will open an issue that we'll need to likely target before releasing `2.2.x` regarding the deprecated code and missing support for non-standard cameras

RobertFlatt:
Updated the comments at the top.
I have other priorities so this is as far as I will take it. 
In the future maybe further enhancements will pop to the top of my stack.

RobertFlatt:
@misl6 Can you approve this, please.

misl6 (approved):
LGTM. Thank you!
I will open a couple of issues targeting what we discussed during the review.

RobertFlatt:
@misl6 You open to a kivy-ios PR that bumps the kivy version to include this change?

https://github.com/RobertFlatt/kivy-ios/blob/master/kivy_ios/recipes/kivy/__init__.py#L10

misl6:
> LGTM. Thank you!
> I will open a couple of issues targeting what we discussed during the review.

See:
https://github.com/kivy/kivy/issues/7871

PR #7860, opened 2022-04-11
Ellipse: update angle_start, angle_end to explicit floats

seevee:
Ellipse parameters `angle_start` and `angle_end` are now called as `float` in docs instead of `int`

This is effectively a typo fix, see the `cdef` definitions directly below these changes:

```cpython
    cdef float _angle_start
    cdef float _angle_end
```

EDIT: additionally, the default values for these parameters have been explicitly set as floats for clarity.


misl6 (approved):
LGTM. Thank you!
(sorry for the wait)

PR #7876, opened 2022-04-30
`Line`/`SmoothLine`: Fixes rendering issues related to corner radius and updates its order (`rounded_rectangle`) + add getter methods for `rounded_rectangle`, `rectangle`, `ellipse`, `circle`.

DexerBR:
This PR is intended to solve some common problems of the `rounded_rectangle` used with `Line`/`SmoothLine`:

- Improve segment control to avoid unwanted shapes if any corner radius = 0 or too larger. Internally, the minimum radius will be 1.

- Updates the default resolution to 45. The current value results in slightly imprecise shapes.

- Prevents rendering issues if width or height is less than 2px when using SmoothLine.

- Corner radius order changed to match `RoundedRectangle` `radius` property (clockwise). There is no standardization between the radius of `rounded_rectangle` and `RoundedRectangle`. Previously `rounded_rectangle` received bottom-left, bottom-right, top-right, top-left. With this PR, it defaults to RoundedRctangle: top-left, top-right, bottom-right, bottom-left (clockwise).

Also includes:

- Unittests for `rounded_rectangle` (Line/SmoothLine).
- Getter methods to `rounded_rectangle`, `rectangle`, `ellipse` and `circle`.

Test code:
```python
from kivy.app import App
from kivy.lang import Builder

Builder.load_string("""
<LineTest>:
    w: 100
    h: 100
    r1: 1
    r2: 1
    r3: 1
    r4: 1
    segs: 45
    _width: 1
    ov: 2
    canvas:

        Color:
            rgba: 0, 0, 1, 1
        SmoothLine:
            width: root._width
            overdraw_width: root.ov
            rounded_rectangle: (50, 300, root.w * 2, root.h, root.r1, root.r2, root.r3, root.r4, root.segs)

        Color:
            rgba: 1, 0, 0, 1
        SmoothLine:
            width: root._width
            overdraw_width: root.ov
            rounded_rectangle: (100, 350, root.w, root.h * 2, root.r1, root.r2, root.r3, root.r4, root.segs)

        Color:
            rgba: 0, 1, 0, 1
        SmoothLine:
            width: root._width
            overdraw_width: root.ov
            rounded_rectangle: (150, 425, root.w, root.h, root.r1, root.r2, root.r3, root.r4, root.segs)

        Color:
            rgba: 1, 1, 1, 1
        Line:
            width: root._width
            rounded_rectangle: (275, 450, root.w, root.h, root.r1, root.r2, root.r3, root.r4, root.segs)

    BoxLayout:
        orientation: "vertical"
        size_hint_y: None
        height: self.minimum_height
        Label:
            text: f"line_width:{root._width:.2f}, ov:{root.ov:.2f}, w:{root.w:.2f}, h:{root.h:.2f}, r1:{root.r1:.2f}, r2:{root.r2:.2f}, r3:{root.r3:.2f}, r4:{root.r4:.2f}, segs:{root.segs:.2f}"
        Slider:
            size_hint: 1, None
            height: 30
            min: 1
            max: 50
            on_value:
                root._width = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 1
            max: 50
            on_value:
                root.ov = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: -50
            max: 500
            value: 0
            step: 1
            on_value:
                root.w = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: -50
            max: 500
            value: 0
            step: 1
            on_value:
                root.h = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 0
            max: 500
            on_value:
                root.r1 = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 0
            max: 500
            on_value:
                root.r2 = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 0
            max: 500
            on_value:
                root.r3 = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 0
            max: 500
            on_value:
                root.r4 = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 0
            max: 500
            on_value:
                root.segs = int(self.value)
""")


class LineTest(FloatLayout):
    ...


class TestApp(App):
    def build(self):
        return LineTest()

TestApp().run()
```

Maintainer merge checklist

### DIFF
--- a/kivy/graphics/vertex_instructions_line.pxi
+++ b/kivy/graphics/vertex_instructions_line.pxi
@@ -1112,8 +1112,8 @@ cdef class Line(VertexInstruction):
 
         * x and y represent the bottom-left position of the rectangle
         * width and height represent the size
-        * corner_radius is the number of pixels between two borders and the center of the circle arc joining them
-        * resolution is the number of line segment that will be used to draw the circle arc at each corner (defaults to 30)
+        * corner_radius specifies the radius used for the rounded corners clockwise: top-left, top-right, bottom-right, bottom-left.
+        * resolution is the number of line segment that will be used to draw the circle arc at each corner (defaults to 90)
 
         The line is automatically closed.
 
@@ -1122,6 +1122,10 @@ cdef class Line(VertexInstruction):
             Line(rounded_rectangle=(0, 0, 200, 200, 10, 20, 30, 40, 100))
 
         .. versionadded:: 1.9.0
+
+        .. versionchanged:: 2.2.0
+        Default value of `resolution` changed to 90.
+        `corner_radius` order changed to match RoundedRectangle radius property (clock-wise). It was bottom-left, bottom-right, top-right, top-left in previous versions.
         '''
         def __set__(self, args):
             if args == None:
@@ -1135,13 +1139,13 @@ cdef class Line(VertexInstruction):
             self.flag_data_update()
 
     cdef void prebuild_rounded_rectangle(self):
-        cdef float a, px, py, x, y, w, h, c1, c2, c3, c4
-        cdef resolution = 30
+        cdef float a, px, py, x, y, w, h, c1, c2, c3, c4, min_wh_half, step
+        cdef resolution = 90
         cdef int l = <int>len(self._mode_args)
 
         self._points = []
-        a = <float>-PI
         x, y, w, h = self._mode_args [:4]
+        w, h = <float>max(2.0, w), <float>max(2.0, h)  # avoid issues with smoothline if width/height < 2.0
 
         if l == 5:
             c1 = c2 = c3 = c4 = self._mode_args[4]
@@ -1154,41 +1158,66 @@ cdef class Line(VertexInstruction):
             c1, c2, c3, c4 = self._mode_args[4:8]
             resolution = self._mode_args[8]
 
-        px = x + c1
-        py = y + c1
+        step = PI / resolution
 
-        while a < - PI / 2.:
-            a += pi / resolution
-            self._points.extend([
-                px + cos(a) * c1,
-                py + sin(a) * c1])
+        # Limits the maximum radius allowed. Minimum radius = 1 avoid shape
+        # draw issues (especially with smoothline)
+        min_wh_half = <float>(min(w, h) / 2)
+        c1 = <float>max(1.0, min(min_wh_half, c1))
+        c2 = <float>max(1.0, min(min_wh_half, c2))
+        c3 = <float>max(1.0, min(min_wh_half, c3))
+        c4 = <float>max(1.0, min(min_wh_half, c4))
 
+        # Segment calculations are based on the quadrants of the trigonometric
+        # circle: 0 (0º) → PI/2 (90º) → PI (180º) → -PI (270º) → 0 (360º).
+
+        # top-right (quadrant 1)
+        a = 0.0
         px = x + w - c2
-        py = y + c2
-
-        while a < 0:
-            a += PI / resolution
-            self._points.extend([
-                px + cos(a) * c2,
-                py + sin(a) * c2])
-
-        px = x + w - c3
-        py = y + h - c3
+        py = y + h - c2
 
         while a < PI / 2.:
-            a += PI / resolution
+            a += step
             self._points.extend([
-                px + cos(a) * c3,
-                py + sin(a) * c3])
+                px + cos(a) * c2,
+                py + sin(a) * c2
+            ])
 
-        px = x + c4
-        py = y + h - c4
+        # top-left  (quadrant 2)
+        a = <float>PI / 2.0
+        px = x + c1
+        py = y + h - c1
 
         while a < PI:
-            a += PI / resolution
+            a += step
+            self._points.extend([
+                px + cos(a) * c1,
+                py + sin(a) * c1
+            ])
+
+        # bottom-left (quadrant 3)
+        a = <float>- PI
+        px = x + c4
+        py = y + c4
+
+        while a < - PI / 2.0:
+            a += step
             self._points.extend([
                 px + cos(a) * c4,
-                py + sin(a) * c4])
+                py + sin(a) * c4
+            ])
+
+        # bottom-right (quadrant 4)
+        a = <float>- PI / 2.0
+        px = x + w - c3
+        py = y + c3
+
+        while a < 0.0:
+            a += step
+            self._points.extend([
+                px + cos(a) * c3,
+                py + sin(a) * c3
+            ])
 
         self._close = 1
 
--- a/kivy/graphics/vertex_instructions_line.pxi
+++ b/kivy/graphics/vertex_instructions_line.pxi
@@ -1253,11 +1253,10 @@ cdef class Line(VertexInstruction):
         px = x + c1
         py = y + h - c1
 
-        min_c2_x = x + w - c2
         while a < PI / 2. - step:
             a += step
             self._points.extend([
-                min(min_c2_x, px - cos(a) * c1),
+                px - cos(a) * c1,
                 py + sin(a) * c1,
             ])
 
@@ -1266,12 +1265,11 @@ cdef class Line(VertexInstruction):
         px = x + w - c2
         py = y + h - c2
 
-        min_c3_y = y + c3
         while a < PI / 2. - step:
             a += step
             self._points.extend([
                 px + sin(a) * c2,
-                max(min_c3_y, py + cos(a) * c2),
+                py + cos(a) * c2,
             ])
 
         # bottom-right
@@ -1279,12 +1277,10 @@ cdef class Line(VertexInstruction):
         px = x + w - c3
         py = y + c3
 
-        
-        max_c4_x = x + c4
         while a < PI / 2. - step:
             a += step
             self._points.extend([
-                max(max_c4_x, px + cos(a) * c3),
+                px + cos(a) * c3,
                 py - sin(a) * c3,
             ])
 
@@ -1293,12 +1289,11 @@ cdef class Line(VertexInstruction):
         px = x + c4
         py = y + c4
 
-        min_c1_y = y + h - c1
         while a < PI / 2. - step:
             a += step
             self._points.extend([
                 px - sin(a) * c4,
-                min(min_c1_y, py - cos(a) * c4),
+                py - cos(a) * c4,
             ])
 
         self._close = 1
--- a/kivy/graphics/vertex_instructions_line.pxi
+++ b/kivy/graphics/vertex_instructions_line.pxi
@@ -1254,7 +1254,7 @@ cdef class Line(VertexInstruction):
         py = y + h - c1
 
         min_c2_x = x + w - c2
-        while a < PI / 2.:
+        while a < PI / 2. - step:
             a += step
             self._points.extend([
                 min(min_c2_x, px - cos(a) * c1),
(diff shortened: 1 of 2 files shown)


akshayaurora:
Considering this changes current behaviour, We need to document the change, plus unit tests checking the corner cases would be nice.

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -1097,7 +1098,7 @@ cdef class Line(VertexInstruction):
 
     property rounded_rectangle:
         '''Use this property to build a rectangle, without calculating the
-        :attr:`points`. You can only set this property, not get it.
```
akshayaurora, line 1100: Did you change this? If not then we need to retain this message.

akshayaurora (approved):

kivy/graphics/vertex_instructions_line.pxi
DexerBR: To perform the tests, it is necessary to access the resulting rounded rectangle, to know if it matches expectations.
Therefore, it is now possible to obtain the resulting rounded rectangle and to know, for example, if it was generated.

Another option would also be to keep this rectangle get unavailable for the user and create a specific method just to get the resulting rounded rectangle.
The first option seemed more interesting to me, as it would give more control to the user. But I would like to know your opinion on this.

tshirtman: That's potentially useful indeed, but would be good to define what will be the result if the line was defined otherwise.

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -1310,6 +1355,11 @@ cdef class SmoothLine(Line):
         VertexInstruction.apply(self)
         return 0
 
+    # FIXME: Some artifacts can be observed, depending on the line width,
+    # overdraw_width and radius. The minimum width and/or height of
+    # rounded_rectangle also needs to be limited to 2px.
```
tshirtman, line 1360: Is this now addressed? As the doc says that it won't render https://github.com/kivy/kivy/pull/7876/files#diff-2e9fcb864c08a6bf13e14b06a577a4f19f1de1cc6274381f03112eea4b854442R1117 the radius = 0 issue seems to be addressed as well (comment + code change).

tshirtman (approved):
Looks good to me, some docs improvement appreciated, and comments being possibly outdated, but good job overall :)

kivy/graphics/vertex_instructions_line.pxi
DexerBR: 1 - If the line does not fit within the requirements to be rendered, the rounded rectangle would be None (width or height less than 2 pixels).
2 - If the radius is set to 0, it will automatically be set to 1, to avoid internal issues with SmoothLine line construction. Furthermore, the difference between 0 and 1 would be almost imperceptible.
3 - And the maximum radius would be cut to half of the smallest dimension. It's the same behavior as RoundedRectangle. So if the smallest dimension is 100 and the user sets the radius to 200, the rounded rectangle will be generated with a radius of 50.

Would it be necessary to document this? I added a note regarding item 1, but I don't know if it would be necessary for the others. Since, for item 3, for example, there is no mention in the RoundedRectangle. And it's kind of a consensus that this is the expected behavior for high radius values in rounded rectangle shapes.

tshirtman: 1 - Sorry, i mean if the line was not defined as a rectagle (say, `Line(points=[0, 0, 100, 100])` then the user tries to read the property, or even if it was at some point defined as a rectangle (`line = Line(rounded_ectangle=[0, 0, 100, 100, 10, 10, 10, 10])`) but then later reset as something else (`line.ellipse = [0, 0, 100, 100]`) then the user tries to read the value, from the comment i infer that it would be `None` at least in the first case, but will it be reset to `None` in the second?
2 & 3 - i wasn't sure and so i just double checked, and it was previously possible to do such a shape 
image
which is indeed not possible with `RoundedRectangle` and won't be possible with the new implementation.
Contrary to `RoundedRectangle` it's not possible to have a different radius per direction for each angle, so the APIs are not exactly similar (though that could be addressed in a different PR if someone wishes to make it so), but some people might have used the possibility of making the radius more than half the width/height, although it doesn't always make sense, it does in some cases.

DexerBR: 1 - With my last commits, it is now possible to get the generated shape for: `rounded_rectangle`, `rectangle`, `ellipse` and `circle`. Setting one of these properties will automatically reset the others to None, after the new shape parameters are determined.

An important note (not sure if it needs to be included in the documentation) is that the values obtained when trying to access a shape will not be immediately available, as with `points`, for example. The reason for this is that some shapes internally manage the radii, segments, and other parameters, according to the values passed. After that, the resulting shape is returned, which corresponds to the displayed shape.
Returning a value before that could, in some cases, create different values than what would actually be rendered. Even if the property value were updated later, the previous value would be inconsistent.
Of the new shapes that can be accessed, the only one that can be accessed in advance is `rectangle`. But a different logic just for this property, could be a little inconsistent. So I kept them all the same.

2 & 3 - The minimum radius value is still `1px`. This doesn't seem to make any difference from the radius `0`. But the maximum radius has been scaled up to the smallest dimension between width and height.
I addressed a bug that prevented the radius from being the maximum of the smallest width (minus `1px` of the min radius because of the rendering issues I mentioned above, of course). Now it's possible to get this shape from the image you shared, and it's still possible to get to a circle just passing values `greater than or equal to the smallest dimension` for all angles.

In general, fortunately Line and especially SmoothLine are much more reliable!

kivy/graphics/vertex_instructions_line.pxi
DexerBR: Updated. Would it be important to keep this note that smoothline will not render at 2px dimensions?

DexerBR:
A useful code to test the behavior with different sizes, line widths, radii and segments:

```python
from kivy.app import App
from kivy.lang import Builder

Builder.load_string("""
<LineTest>:
    w: 0
    h: 0
    r1: 1
    r2: 1
    r3: 1
    r4: 1
    segs: 45
    _width: 1
    canvas:

        Color:
            rgba: 0, 0, 1, 1
        SmoothLine:
            width: root._width
            overdraw_width: 2
            rounded_rectangle: (50, 250, 200 + root.w, 150 + root.h, root.r1, root.r2, root.r3, root.r4, root.segs)

        Color:
            rgba: 1, 0, 0, 1
        SmoothLine:
            width: root._width
            overdraw_width: 2
            rounded_rectangle: (100, 300, 100 + root.w, 200 + root.h, root.r1, root.r2, root.r3, root.r4, root.segs)

        Color:
            rgba: 0, 1, 0, 1
        SmoothLine:
            width: root._width
            overdraw_width: 2
            rounded_rectangle: (150, 350, 200 + root.w, 200 + root.h, root.r1, root.r2, root.r3, root.r4, root.segs)

        Color:
            rgba: 1, 1, 1, 1
        Line:
            width: root._width
            rounded_rectangle: (200, 400, 200 + root.w, 200 + root.h, root.r1, root.r2, root.r3, root.r4, root.segs)

    BoxLayout:
        orientation: "vertical"
        size_hint_y: None
        height: self.minimum_height
        Slider:
            size_hint: 1, None
            height: 30
            min: 1
            max: 50
            on_value:
                root._width = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: -500
            max: 500
            value: 0
            on_value:
                root.w = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: -500
            max: 500
            value: 0
            on_value:
                root.h = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 0
            max: 500
            on_value:
                root.r1 = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 0
            max: 500
            on_value:
                root.r2 = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 0
            max: 500
            on_value:
                root.r3 = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 0
            max: 500
            on_value:
                root.r4 = self.value
        Slider:
            size_hint: 1, None
            height: 30
            min: 1
            max: 500
            on_value:
                root.segs = self.value

""")


class LineTest(FloatLayout):
    ...


class TestApp(App):
    def build(self):
        return LineTest()

TestApp().run()
```

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -954,6 +962,12 @@ cdef class Line(VertexInstruction):
             segments = int(abs(angle_end - angle_start) / 2) + 3
             if segments % 2 == 1:
                 segments += 1
+        
+        # Resulting ellipse
+        self._ellipse = (x, y, w, h, angle_start, angle_end, segments)
+        # Reset another properties
```
tshirtman, line 968:
```suggestion
        # Reset other properties
```

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -1036,6 +1056,11 @@ cdef class Line(VertexInstruction):
         if segments == 0:
             segments = int(abs(angle_end - angle_start) / 2) + 3
 
+        # Resulting circle
+        self._circle = (x, y, r, angle_start, angle_end, segments)
+        # Reset another properties
```
tshirtman, line 1061:
```suggestion
        # Reset other properties
```

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -1094,14 +1126,19 @@ cdef class Line(VertexInstruction):
             x, y, width, height = args
         else:
             x = y = width = height = 0
-            assert 0
+            assert(0)
+        
+        # Resulting rectangle
+        self._rectangle = (x, y, width, height)
+        # Reset another properties
```
tshirtman, line 1133:
```suggestion
        # Reset other properties
```

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -1094,14 +1126,19 @@ cdef class Line(VertexInstruction):
             x, y, width, height = args
         else:
             x = y = width = height = 0
-            assert 0
+            assert(0)
```
tshirtman, line 1129: why assert(0) ? debug leftover?

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -1154,47 +1209,100 @@ cdef class Line(VertexInstruction):
...
+                px - sin(a) * c4,
+                py - cos(a) * c4
+            ])
 
         self._close = 1
 
     property bezier:
         '''Use this property to build a bezier line, without calculating the
-        :attr:`points`. You can only set this property, not get it.
+        :attr:`points`.
```
tshirtman, line 1305: did you also mean to save `_bezier` and to reset the others in this setter? i don't see it.

kivy/graphics/vertex_instructions_line.pxi
DexerBR: I think it was an unintentional change.

kivy/graphics/vertex_instructions_line.pxi
DexerBR: It was also an unintentional change. Thanks for identifying!

DexerBR:
Hi @tshirtman !
Is there anything else that needs to be addressed before merging this one?

Thanks in advance!

tshirtman (approved):
Looks good to me! I didn't test it again though, but i'd rather not block this further, if there are important issues, they will surely be found before release.

misl6:
Did a quick test, and also LGTM.

Agree with @tshirtman :
> if there are important issues, they will surely be found before release.

PR #7903, opened 2022-06-20
Migrate from probot/no-response to lee-dohm/no-response

misl6:
- https://github.com/probot/no-response has been marked as archived.
- `probot/no-response`  is not working anymore.
- https://github.com/lee-dohm/no-response has been marked as the recommended replacement.

Maintainer merge checklist


matham (dismissed):
It's a little scary to give permissions to someone we don't know, but I guess that's how this works 🤷‍♀️

misl6:
> It's a little scary to give permissions to someone we don't know, but I guess that's how this works 🤷‍♀️

AFAIK:

✅ Considering that the action runs on a `ubuntu-latest` runner (unlike the old approach, where was a Github App), the `github.token` is only available to the runner, and not the original author of the `action`.

🔴 The author may add some malicious code to the action, and release a new version with the same tag.

💡 We may want to fork the action, and use something like `uses: kivy/no-response@version`, so we can make sure that the pinned version code doesn't actually change. (See: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses)

matham:
>We may want to fork the action, and use something like uses: kivy/no-response@version, so we can make sure that the pinned version code doesn't actually change

I think this is a good idea, and we can periodically fetch from upstream.

What do you think?

tshirtman:
Isn't it enough to pin on the commit id instead of tag, and to update that from time to time?

misl6:
Targeting a commit ID seems reasonable.

I've added a comment which explains why we're targeting a specific commit and the origin of that commit (doesn't belong to any branch).

PS: I previously missed moving the `no-response.yml` file into the `workflows` folder. 👓

PR #7909, opened 2022-07-01
Feature/add text input spinner

AnthonyZi:
Implement TextInputSpinner

Seperates SpinnerBehavior from Spinner and moves it into uix/behaviors.
Adds TextInputSpinner, a spinner based on SpinnerBehavior with
a TextInput instead of a Button. This allows the user to either choose one
of the Spinner dropdown options or to type a new value into the entry. As
typing information into the entry delivers more information to narrow the
dropdown options down, an optional live_update feature for TextInputSpinner
has been implemented.

Maintainer merge checklist


AnthonyZi:
Do you have a strategy on resolving these circular imports? I had to fix another circular import by making use of the Factory. I think it was the SpinnerOption in the behavior. Is this the right way to do this?

PR #7913, opened 2022-07-10
fix(UrlRequest): Add "on_finish" and add alternative implementation

NomadDemon:
- Added "on_finish" callback to URlRequest, it fires always when the request is done
- Added alternative implementation for UrlRequest based on REQUESTS lib
- It have also implemented "on_finish" and "auth" params for easy BASICAUTH handling
- Refactored tests of new implementation to be mocked by RESPONSES, so we dont need real network to test nor we dont ping real servers during tests

Maintainer merge checklist


NomadDemon:
Iam not sure if my changes in "setup" are ok, never changed stuff in setup of any package
I just wanted the project to download requests and responses with it [they are wide used in real world]

NomadDemon:
this new implementation have same interface as old "UrlRequest" just changed backend to more flexible [and tests are now made as they should be made]

kivy/network/newurlrequest.py
```diff
@@ -0,0 +1,584 @@
+'''
+UrlRequest
+==========
+
+.. versionadded:: 1.0.8
```
tshirtman, line 5: need to update the name and version added.

But more generally, i think it would be better to avoid so much duplication, and rather to have one class inheriting from the other, (and possibly an abstract implementation), as it needs a better name than `newurlrequest` i think something like `urlrequest_requests.py` would be better, even i the name of the class stays the same so only import needs to be changed for it to be a drop-in replacement.

tshirtman (changes requested):

kivy/network/newurlrequest.py
tshirtman: I agree it would be better to only have one module, hence i think having in the same module, the base implementation, and the two subclasses, but making `UrlRequest` an alias to `UrllibUrlRequest` would be a good option, people can still import either class if they want to, and we can change the alias to point to `RequestsUrlRequest` in 1-2 releases, deprecating the legacy version, (but still available ,if people did something weird and have compatibility issues), and then removing it entirely further down the line, so people have enough headway to do any updates they need.

kivy/network/urlrequest.py
```diff
@@ -633,26 +555,239 @@ def cancel(self):
...
+            cert=ca_file,
+            **kwargs
+        )
 
-    def on_error(req, error):
-        pprint('Got an error:')
-        pprint(error)
+        return None, response
 
-    Clock.start_clock()
```
NomadDemon, line 648: i dont know what is this part, without it still works well [even tests]

setup.cfg
```diff
@@ -52,8 +52,10 @@ dev =
     kivy_deps.glew_dev~=0.3.1; sys_platform == "win32"
     flake8
     pre-commit
+    responses
 base =
     pillow
+    requests
     docutils
```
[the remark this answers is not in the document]
tshirtman: I think it's ok. requests is now a hard dependency because you didn't make the import conditional, it could be done, but at this point most users will have it directly or through dependencies anyway.

kivy/network/urlrequest.py
tshirtman: then it should be fine to remove, if it breaks something, at least we'll know why we need it :)

tshirtman (commented):
Looks mostly good to me, a few little remarks.

Cooperative subclassing like this (parent method requiring small methods to be implemented by subclasses) can be a pain to maintain, but i don't really see a better solution here, and this shoud be a temporary solution, as we could deprecate the Urllib in a future release, before removing it entirely.

One think that could be added actually, since we have a "network" section in config, would be to decide which implementation to use depending on the value set in that. The main question being of the default value, we can't pick the requests implementation yet for obvious reasons, but setting the default value to the old one would mean user not switching over automatically in the future, so we could have an "auto" value, by default, so people could set either value by default, or just wait for the transition to happen automatically when we deprecate the urllib implementation.

anyway, aside minor points, looks good to me!

kivy/config.py
```diff
@@ -369,14 +373,15 @@
...
 from os.path import exists
+from weakref import ref
+
 from kivy import kivy_config_fn
+from kivy.compat import PY2, string_types
 from kivy.logger import Logger, logger_config_update
-from collections import OrderedDict
 from kivy.utils import platform
-from kivy.compat import PY2, string_types
-from weakref import ref
```
NomadDemon, line 379: isort.. sorry ^^'

kivy/config.py
```diff
@@ -335,6 +338,7 @@
     to the `graphics` section.
     `kivy_clock` has been added to the kivy section.
     `default_font` has beed added to the kivy section.
+    `useragent` has been added to the network section.
```
NomadDemon, line 341: was missing

kivy/network/urlrequest.py
```diff
@@ -58,43 +58,35 @@ def bug_posted(req, result):
...
-    from urlparse import urlparse, urlunparse
-else:
-    from http.client import HTTPConnection
-    from urllib.parse import urlparse, urlunparse
+import requests
+from kivy.clock import Clock
+from kivy.config import Config
+from kivy.logger import Logger
+from kivy.utils import platform
+from kivy.weakmethod import WeakMethod
```
NomadDemon, line 72: isort again.. sorry

tshirtman (approved):

PR #7919, opened 2022-07-21
Fix float argument in `range` in `calculate_normals` and remove empty return from `MTL`

Neizvestnyj:
`rang` must accept an argument of the `int` type. `MTL` class must return `dict` not `NoneType`

### DIFF
--- a/examples/3Drendering/objloader.py
+++ b/examples/3Drendering/objloader.py
@@ -9,7 +9,7 @@ class MeshData(object):
         self.indices = []
 
     def calculate_normals(self):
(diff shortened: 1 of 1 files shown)


matham (approved):

examples/3Drendering/objloader.py
```diff
@@ -132,7 +132,7 @@ def __init__(self, filename, swapyz=False):
 def MTL(filename):
     contents = {}
     mtl = None
-    return
```
matham, line 135: Actually, did you test this with a mtl file? Because I'm not sure the MTL function is currently working as is.

PR #7927, opened 2022-07-28
Fix minor typo in pong tutorial code comments

paultcochrane:
While reading through the pong tutorial code, I noticed that `of` should
have been `off` in code comments.  This commit fixes the issue.

This PR is submitted in the hope that it is helpful.  If you would like anything changed, please don't hesitate to let me know and I'll update the PR as appropriate and resubmit.

Maintainer merge checklist

### DIFF
--- a/examples/tutorials/pong/steps/step5/main.py
+++ b/examples/tutorials/pong/steps/step5/main.py
@@ -40,7 +40,7 @@ class PongGame(Widget):
     def update(self, dt):
         self.ball.move()
 
-        # bounce of paddles
+        # bounce off paddles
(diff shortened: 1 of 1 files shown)


tshirtman (approved):

paultcochrane:
@tshirtman thanks for approving this! I'm not sure why the unit tests are failing though; the change is only a documentation change and should be affecting that, should it?

tshirtman:
Yeah i don’t think it has anything to do with it, the CI might be flacky, that will need checking later, but no reason to block what’s just a comment change

PR #7942, opened 2022-08-04
Config unicode support on Windows

SanAndreyas:
On Windows platform without this it cannot write correctly unicode characters

Maintainer merge checklist

### DIFF
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -571,7 +571,7 @@ class ConfigParser(PythonConfigParser, object):
         if self.filename is None:
             return False
         try:
-            with open(self.filename, 'w') as fd:
+            with open(self.filename, 'w', encoding="utf-8") as fd:
                 PythonConfigParser.write(self, fd)
         except IOError:
             Logger.exception('Unable to write the config <%s>' % self.filename)
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -479,7 +479,7 @@ class ConfigParser(PythonConfigParser, object):
         #    self.readfp(f)
         old_vals = {sect: {k: v for k, v in self.items(sect)} for sect in
                     self.sections()}
-        PythonConfigParser.read(self, filename, encoding="utf-8")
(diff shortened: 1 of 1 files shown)


matham (dismissed):
Should be ok, I think. Currently it uses locale, but probably utf8 should be safe as a kivy default.

tshirtman:
Shouldn't it also default reading the configuration to utf-8, for consistency?
Currently it defers to ConfigParser implementation, which does accept an optional encoding
https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read
and without it, default to the default for `open()` of which the docs says
>  if encoding is not specified the encoding used is platform-dependent: locale.getpreferredencoding(False) is called to get the current locale encoding
which probably won't cause any issue on linux/osx, but very much might on windows, as utf8 is not the default encoding.

SanAndreyas:
@tshirtman your're 100% right. The following shall also be modified 'PythonConfigParser.read(self, filename, encoding="utf-8")' @ line 482.

matham (dismissed):

Berserker66:
I suggest to use "utf-8-sig" instead, which will correctly skip the BOM if present, and ignore it if not present.

SanAndreyas:
utf-8-sig would be nice only for reading. But it looks kind of inconsistent to me to have the reading like utf-8-sig and the writing as utf-8

matham:
>utf-8-sig would be nice only for reading. But it looks kind of inconsistent to me to have the reading like utf-8-sig and the writing as utf-8

I don't think we should generate the BOM. So I think writing with just utf8 is ok, even if reading is done with `utf-8-sig`. I don't think that's inconsistent. Rather, it makes the reading more robust.

matham (approved):

misl6:
@matham I was checking the pending PRs marked as milestone `2.2.0` since we're approaching a feature freeze.
Anything is blocking us to merge this one?

PR #7946, opened 2022-08-09
Add doc for `Canvas.add()`

### LINKED ISSUE #7945: 'Canvas.add()' checks boundary but  'Canvas.insert()' doesn't
**Software Versions**
* Python: 3.8.12
* OS: Linux Mint 20.2 Mate Edition
* Kivy: master 714f4935085d4ac0c220fdf7183fb35175f8bcad
* Kivy installation method: `git clone` then `python -m pip install -e ".[dev,full]"`

### Issue

There was a canvas containing a `before` group and an `after` group.

```python

c = Widget().canvas

def print_children(c):
    print([i.__class__.__name__ for i in c.children])

print_children(c)
c.before
c.after
print_children(c)
```

```
[]
['CanvasBase', 'CanvasBase']
```

When  I called `c.insert(0, PushMatrix())` and `c.add(PopMatrix())`, I was expecting the `c.children` to be either of

- `['CanvasBase', 'PushMatrix', 'PopMatrix', 'CanvasBase']` or
- `['PushMatrix', 'CanvasBase', 'CanvasBase', 'PopMatrix']`

but neither was an actual result.

```python
c.insert(0, PushMatrix())
c.add(PopMatrix())
print_children(c)
```

```
['PushMatrix', 'CanvasBase', 'PopMatrix', 'CanvasBase']
```

### Cause

The cause of the issue is that `Canvas` overrides `InstructionGroup.add()` but doesn't  override `InstructionGroup.insert()`.
https://github.com/kivy/kivy/blob/714f4935085d4ac0c220fdf7183fb35175f8bcad/kivy/graphics/instructions.pyx#L656-L662

As you can see, `Canvas.add()` does some check so that the `after` group always remains the last one, which is why I couldn't add a PopMatrix behind the `after` group. On the other hand, `InstructionGroup.insert()` doesn't do that (Of course, `InstructionGroup` is not even aware of the existance of the `before` group and the `after` group),

https://github.com/kivy/kivy/blob/714f4935085d4ac0c220fdf7183fb35175f8bcad/kivy/graphics/instructions.pyx#L191-L195

which is why I was able to insert a PushMatrix in font of the `before` group.

### Solution

I think it's better to

- document about this unintuitive behavior or
- have `Canvas` override `insert()` and make it work in the same way as `add()`

### to reproduce the issue

```python
from kivy.graphics import Canvas, PushMatrix, PopMatrix


def print_children(c: Canvas):
    print([i.__class__.__name__ for i in c.children])

c = Widget().canvas

print("\n---- initial state ----")
print_children(c)

c.before
c.after
print("\n---- added 'after' and 'before' ----")
print_children(c)

c.add(PopMatrix())
print("\n---- after c.add(PopMatrix()) ----")
print_children(c)

c.insert(0, PushMatrix())
print("\n---- after c.insert(0, PushMatrix()) ----")
print_children(c)
```

gottadiveintopython:
Maintainer merge checklist

Fixes #7945

### DIFF
--- a/kivy/graphics/instructions.pyx
+++ b/kivy/graphics/instructions.pyx
@@ -654,6 +654,11 @@ cdef class Canvas(CanvasBase):
         return 0
 
     cpdef add(self, Instruction c):
+        '''Append an :class:`Instruction` to our list. If the canvas contains
+        an `after` group, then this instruction is inserted just before the
+        after group, which remains last. This is different from how
+        :meth:`insert` works, which can insert anywhere.
+        '''
         # the after group must remain the last one.
         if self._after is None:
             c.radd(self)


tshirtman (dismissed):

gottadiveintopython:
@tshirtman  Thanks for the approval. I might edit the doc later because I'm not confident in my English so don't merge yet. And it would be great if you give me an advice about this pr.

kivy/graphics/instructions.pyx
```diff
@@ -654,6 +654,11 @@ cdef class Canvas(CanvasBase):
         return 0
 
     cpdef add(self, Instruction c):
+        '''Add a new :class:`Instruction` to our list. If the list contains
+        the after group, it remains the last one. This is different to how
+        :meth:`insert` works, which doesn't care the before/after group at
```
tshirtman, line 659:
```suggestion
        :meth:`insert` works, which doesn't care about the before/after group at
```
i think this is clear enough, but if you want, you could try something like
```
"""
Append an :class:`Instruction` to our list. If the canvas contains an `after`
group, then this instruction is inserted just before the after group, which
remains last. This is different from how :meth:`insert works, which can
insert anywhere.
"""
```

gottadiveintopython: I really appriciate :+1: I'll take your one.

gottadiveintopython: @tshirtman I'm ok with the current state of the pr.

misl6:
@gottadiveintopython

In case you want to have a green check (I've seen your reaction on the PEP8 PR which I completely forgot about), feel free to force-push as it's now merged.

gottadiveintopython:
@misl6 Thanks, I'll do that later.

misl6 (approved):
LGTM, thank you!

(Now tests are not failing as `setuptools` has been fixed)

PR #7969, opened 2022-08-23
Config is not available when generating docs + Use `getdefault` instead of `has_option` + `get`

misl6:
Maintainer merge checklist

- `gen_docs` step started to fail after #7913 (See: https://github.com/kivy/kivy/runs/7962835314)
- `Config` is not available during docs generation
- Use `Config.getdefault` instead of `Config.has_option` + `Config.get` to get the `prefered_implementation `

### DIFF
--- a/kivy/network/urlrequest.py
+++ b/kivy/network/urlrequest.py
@@ -56,6 +56,7 @@ If you want a synchronous request, you can call the wait() method.
 
 '''
 
+import os
 from base64 import b64encode
 from collections import deque
 from http.client import HTTPConnection
@@ -783,9 +784,11 @@ implementation_map = {
     "urllib": UrlRequestUrllib,
 }
 
-if Config.has_option('network', 'implementation'):
-    prefered_implementation = Config.get("network", "implementation")
-    UrlRequest = implementation_map.get(prefered_implementation)
-
+if not os.environ.get("KIVY_DOC_INCLUDE"):
+    prefered_implementation = Config.getdefault(
+        "network", "implementation", "default"
+    )
 else:
-    UrlRequest = implementation_map["default"]
+    prefered_implementation = "default"
+
+UrlRequest = implementation_map.get(prefered_implementation)


kivy/network/urlrequest.py
```diff
@@ -783,9 +784,11 @@ def call_request(self, body, headers):
     "urllib": UrlRequestUrllib,
 }
 
-if Config.has_option('network', 'implementation'):
-    prefered_implementation = Config.get("network", "implementation")
-    UrlRequest = implementation_map.get(prefered_implementation)
-
+if not os.environ.get("KIVY_DOC_INCLUDE"):
```
NomadDemon, line 787: just askin:

`os.environ.get("KIVY_DOC_INCLUDE")`

how it works in big picture?

misl6: Sometimes code speaks a better language, so I will try to explain with code chunks 😄 .

https://github.com/kivy/kivy/blob/484142781244cdb1de6f158c4e752733397c72a0/kivy/__init__.py#L334-L344
and
https://github.com/kivy/kivy/blob/484142781244cdb1de6f158c4e752733397c72a0/kivy/config.py#L712

Does it answer your question?

NomadDemon: and what the difference between 
`environ['KIVY_DOC'] = '1' `
and
`environ['KIVY_DOC_INCLUDE'] = '1' `
?

and whats the problem with config, when generating docs? do they collide somehow?

i see this construction in many places, thats why i think about how to solve this problem without "hacks" :)

misl6: `KIVY_DOC` acts as a conditional to avoid exceptions while documenting code which includes platform-specific code or code executed in threads, etc ...

`KIVY_DOC_INCLUDE` AFAIK is only used to skip the configuration part when `sphinx-build` is in action.

`sphinx-build` takes its own set of configuration parameters, and without having this hack, the config will also be parsed from kivy.

BTW, that's certainly something we can improve in the future in order to avoid these conditionals in the codebase 😄

matham (approved):

kivy/network/urlrequest.py
NomadDemon: after i will fix inputs autosuggestions on android i will check what we can do here :P

Issue #7970, opened 2022-08-24
Add environment variable to make logger code testable.


Julian-O:
**Is your feature request related to a problem? Please describe.**

`kivy.logger` can be broken into two parts:

* Definitions: Define all of the classes and other definitions required to implement the Kivy logger system.
* Installation: Install it as the primary logging code for the entire system (as a side-effect of import) by monkey-patching the `logging.root` and `sys.stderr`.

Unfortunately, the installation part makes it very difficult to unit-test the definitions. The classes are already installed and running before they can be broken out individually.

I am working on a number of changes to the logging code, and I want to be sure it is properly tested - it is part of the fundamental core that every Kivy app uses.

**Describe the solution you'd like**

1. Define an undocumented, unsupported environment variable *for testing only* that turns off the installation of logging.
2. Add a skipif marker to logging tests that only work if the environment variable is appropriately set,
3. Add a custom pytest marker to those tests so that they can be selectively run when the environment variable is set.
4. Modify the ci scripts:
   * to run the unit tests twice:
        * Once with the variable not set, which will run the majority of the tests, and skip the special logging ones.
        * Once with the environment variable and the custom marker set, so that only those tests will be run.
   * merge the coverage data so only one report is produced.

**Describe alternatives you've considered**

* My first approach was to create a separate file, `logginglib.py`, and gradually migrate the definitions into it, as it was refactored, eventually leaving only the installation code in `kivy.logger`. `test_logginglib.py` could then test the definitions in isolation.

  However, this was destined to failure for two reasons:

  1. In `test_logginglib.py`, when the test-case runs and imports `kivy.logginglib`, it implicitly imports `kivy.__init__.py` which imports `kivy.logging`, and runs the installation code.

  2. Even worse, `pytest` is importing `kivy.__init__.py` implicitly before the tests run. I haven't figured exactly where. It is before `conftest.py` is called, and even before `tests.__init__.py` is called.

* This can't be controlled by the config file because the logger is installed before the config file is read.

* This can't be controlled in conftest.py or as a pytest fixture, because the logging is installed so early.

[...]

PR #7971, opened 2022-08-24
Support KivyLogMode environment variable for logging testing

Julian-O:
Address #7970 by setting up an (undocumented, unsupported) environment variable to allow logging unit tests to run without logging being "installed".

A single test added, to show the new scheme works.

This has no impact on the API.

Also:
* Remove unused imports.
* Remove Python 2 only code.
* Run through `black` (with line-length overridden to match PEP8) to canonicalize formatting.

Maintainer merge checklist

### DIFF
--- a/.ci/ubuntu_ci.sh
+++ b/.ci/ubuntu_ci.sh
@@ -101,11 +101,14 @@ install_kivy_sdist() {
 
 test_kivy() {
   rm -rf kivy/tests/build || true
-  KIVY_NO_ARGS=1 python3 -m pytest --maxfail=10 --timeout=300 --cov=kivy --cov-report term --cov-branch "$(pwd)/kivy/tests"
+  # Tests with default environment variables.
+  env KIVY_NO_ARGS=1 python3 -m pytest --maxfail=10 --timeout=300 --cov=kivy --cov-branch --cov-report= "$(pwd)/kivy/tests"
+  # Logging tests, with KIVY_LOG_MODE=TEST.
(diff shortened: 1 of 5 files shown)


misl6:
Absolutely nothing against `black` formatting, and we also discussed applying `black` to all the codebase, in a single commit, in the future (In order to hide this single commit from the blame history, when the feature will be widely supported), but applying black formatting on a whole file were only a few lines were touched is something I'm quite afraid, as it breaks history and makes the PR less readable.

Can we avoid that?

misl6 (approved):
I was requesting to document `KIVY_LOG_MODE` somewhere, but:
> This is temporary scaffolding, that is intended to be replaced by a properly documented feature in the short-term.

As for the unused imports and the Python 2 code, I've wrote it down in order to keep in mind it for the next time.

LGTM. Thank you!

Julian-O:
Thanks for the code review.

> I've wrote it down in order to keep in mind it for the next time.

Oh, you can check that item off! The unused imports and Python 2 code were removed in the first commit (and not replaced in the second).

Julian-O:
> I was requesting to document KIVY_LOG_MODE somewhere.

Yes, I certainly plan to when it is ready for external use. My plan is for it to stop accepting "TEST", and start accepting "KIVY" (default), "PYTHON" and "MIXED". I didn't want to give people the impression that "TEST" would be supported for long.

misl6:
> Thanks for the code review.
> 
> > I've wrote it down in order to keep in mind it for the next time.
> 
> Oh, you can check that item off! The unused imports and Python 2 code were removed in the first commit (and not replaced in the second).

I was referring to `KIVY_LOG_MODE` , the import and python2 code reminder has already been deleted 😄

PR #7975, opened 2022-08-25
Light clean up of stderr handling code.

Julian-O:
I have a confession. I think the LogFile and ConsoleHandler code is ugly. I completely rewrote it with some re-usable classes, with meaningful names, low-coupling, and lots of unit tests. It was beautiful.

But, at the last minute, when I went to write a detailed explanation of what the old code did and justify why my solution was better, despite being 5x the amount of code, I decided the problems weren't *that* bad. I was being unfair.

So I took a much lighter approach than I had intended.

What I did do:

	* LogFile is neither a file, nor does it do any logging. It is a stream-like object that processes each line with a function. Renamed to ProcessingStream.
	* Generally added comments and docstrings.
	* ConsoleHandler relied on a global constant, despite having a
	  local reference to the same value. Removed the constant.
	* Added a comment to not rely on KIVY_LOG_MODE=TEST.
	* Don't catch keyboard interrupt exceptions and the like.
	* 'l' is considered a poor choice of variable name, not least because it lookes like 1 and I.
	* Updated test to not rely on constant.
	* Added missing test to ensure that root logger *is* overridden.
	

Maintainer merge checklist

### DIFF
--- a/kivy/logger.py
+++ b/kivy/logger.py
@@ -99,8 +99,6 @@ __all__ = (
 
 Logger = None
 
-previous_stderr = sys.stderr
-
 
 logging.addLevelName(9, 'TRACE')
 logging.TRACE = 9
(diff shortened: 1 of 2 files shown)


misl6 (approved):
> But, at the last minute, when I went to write a detailed explanation of what the old code did and justify why my solution was better, despite being 5x the amount of code, I decided the problems weren't that bad. I was being unfair.

That's part of the process 😄 

LGTM. Thank you!

PR #7982, opened 2022-08-27
Use `SDL_WINDOWEVENT_DISPLAY_CHANGED` to notice about window switching display to update `_density` an `dpi`

misl6:
Maintainer merge checklist

🔴 It was supposed to be "Ready to Merge", but the CI is failing. We should consider how the SDL2 dep is managed on Ubuntu and other Linux distros.

⚠️ `SDL_WINDOWEVENT_DISPLAY_CHANGED` has been added on SDL `2.0.18` but needs SDL `2.24.0` to work as expected on macOS, as previously was using a non-macOS compatible logic to compute the display on which the window is really attached to (https://github.com/kivy/kivy/pull/7974 bumps SDL2 to `2.24.0`).

💡 Since now `SDL2` seems to be aware of `DPI` on all platforms (including Windows), we may want to update how `dpi` gets computed.

- Added missing `SDL_WINDOWEVENT_*` declarations for future usage.
- Handle `SDL_WINDOWEVENT_DISPLAY_CHANGED` so we can notice about a window that changed the display is attached to.
- Extract a method `_update_density_and_dpi` which updates `_density` and `dpi` to avoid DRY violations

Previously:
no_dpi

After:
ok_dpi

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -720,6 +720,11 @@ cdef class _WindowSDL2Storage:
                     'windowmoved',
                     event.window.data1, event.window.data2
                 )
+            elif event.window.event == SDL_WINDOWEVENT_DISPLAY_CHANGED:
+                action = (
+                    'windowdisplaychanged',
+                    event.window.data1, event.window.data2
+                )
             else:
                 #    print('receive unknown sdl window event', event.type)
                 pass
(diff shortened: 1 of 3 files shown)


tshirtman (approved):

matham (dismissed):

FilipeMarch:
Hello, since 2022  I would like to use my external monitor, but because of this dpi bug I can't because the screen gets completely distorted here using macOS. Please could you merge this PR? This would really really help me a lot.

PR #7985, opened 2022-08-27
Handle non-strings in logs.

### LINKED ISSUE #7984: Kivy Logger doesn't take into account non-string messages
**Software Versions**
* Python: *
* OS: *
* Kivy: `master`
* Kivy installation method: `source`

**Describe the bug**

Kivy Logger doesn't take into account non-string messages, and fails with:
`AttributeError: ... object has no attribute 'replace'`

**Expected behavior**
As specified in [Python Docs](https://docs.python.org/3/howto/logging.html#using-arbitrary-objects-as-messages), also an arbitrary object could be used as a message.

**To Reproduce**
```python
from kivy.logger import Logger
...
...
Logger.info(dict())
...
...
```

```python
   File "/Users/mirko/Documents/projects/kivy/kivy/logger.py", line 433, in format
     self._coloring_cls(ColonSplittingLogRecord(record)))
   File "/Users/mirko/Documents/projects/kivy/kivy/logger.py", line 390, in __init__
     self.msg = self._format_message(self.msg)
   File "/Users/mirko/Documents/projects/kivy/kivy/logger.py", line 364, in _format_message
     return message.replace(
 AttributeError: 'dict' object has no attribute 'replace'
```

**Additional context**

CC: @Julian-O

Julian-O:
Fixes #7984 

LogRecords may contain non-string messages (not just non-string args) that aren't converted to strings until the last moment (to save expensive effort if the log message is never emitted).

Bug introduced recently with Colored/UncoloredLogRecords did not convert to str before replacing color markings (if any).

* str() added.
* Unit tests added (used to fail, now pass)

Maintainer merge checklist

### DIFF
--- a/kivy/logger.py
+++ b/kivy/logger.py
@@ -361,7 +361,7 @@ class ColoredLogRecord(logging.LogRecord):
 
     @classmethod
     def _format_message(cls, message):
-        return message.replace(
+        return str(message).replace(
             "$RESET", cls.RESET_SEQ).replace("$BOLD", cls.BOLD_SEQ)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


misl6 (approved):
LGTM. Thank you!

Issue #7172, opened 2020-10-23
kivy import overwrites sys.stderr


rambo:
**Software Versions**
* Python: 3.8.5
* OS: Ubuntu 20.04
* Kivy: git master
* Kivy installation method: pip install -e

**Describe the bug**
```
import kivy
```

Has a lot of side-effects, one of them is that sys.stderr is now kivy LogFile

**Expected behavior**
Just importing the module must not have side effects like this in the operating environment (there are plenty of others too but let's keep them each in a separate issue)

**To Reproduce**
```
import kivy
```

**Additional context**

Also replacing the root logger is a sin, it's been discussed for example in #4113 and #4322, this messes up logging for every library that uses standard logging and is not aware of kivy (and why should they be, using the standard logger should just work) so any app using libraries like those suddenly loses all logs from them.

Anyway I'm going to look what I can do about these issues, PR will come later but progress can be seen at https://github.com/advian-oss/kivy/tree/logging

Julian-O:
I have been working on this issue. I have an (insufficiently-tested yet) prototype that tidies up the stderr redirect to be cleaner.

In an ideal world, the redirect would be available to those who wanted it, but not get implemented by default upon import. However, I am worried about breaking existing apps if I remove it entirely. 

Would it be satisfactory to allow override of the the default behaviour with an environment variable? (Or config?)

rambo:
The proposal of using ENV/config variable sounds reasonable to maintain backwards compatibility while allowing integration of libraries that use the python standard logging systems correctly

Julian-O:
Fixed about a year ago. Closing.

Issue #7986, opened 2022-08-28
LoggerHistory stores one extra record.


Julian-O:
**Software Versions**
* Python: N/A
* OS: N/A
* Kivy: 2.10
* Kivy installation method: N/A

**Describe the bug**
The docstring for `kivy.logger` explains `LoggerHistory` stores the last 100 records.

It actually stores the last 101.

**Expected behavior**
Store exactly 100

**To Reproduce**

I haven't bothered actually running this, sorry, but:

    from kivy.logger import LoggerHistory

    for i in range

    print(LoggerHistory.history)

**Code and Logs and screenshots**
```python
def test_logger_history_size():
    from kivy.logger import Logger, LoggerHistory
    for x in range(200):
        Logger.info(x)
    assert len(LoggerHistory.history) == 100, \
        "Wrong size: %s" % len(LoggerHistory.history)
```

**Additional context**

Other alternatives:

* Document that it stores 101.
* Stop hard-coding the capacity with a magic number.
* Replace it with a subclass of MemoryLogger that has a buffer.
* Stop advertising it exists, so it can be deprecated. If it is needed for tests, add it as a handler in the test code.
    * It is also used as a cache before the log file is configured: that can be replaced with a subclass of BufferingHandler.

Issue #7981, opened 2022-08-27
Remove undocumented sys._kivy_logging_handler


Julian-O:
**Software Versions**
* Python: N/A
* OS: N/A
* Kivy: 2.10
* Kivy installation method: N/A

**Describe the bug**
`logger.py` looks to see if the client has monkey-patched `_kivy_logging_handler` into the `sys` module. If so, it will add it as a handler.

Despite being present since the initial commit, this attribute is not mentioned in any documentation.

**Expected behavior**

One approach would be to document the attribute.

However, with #7978 in the process of being completed, it is no longer necessary for a client to go to such extreme lengths as monkey patching a core Python module. They can simply directly add their favourite handler (or handlers!) straight to the root logger.

Furthermore, I tried to find examples of it being used. The attribute is not set or used anywhere else in the Kivy repository. After 10 minutes searching, I could only find [one public use of it in the wild](https://github.com/cropleyb/pentai/blob/bbd32864cabce9ba5cb1051fa9d78d69c8feb5e5/main.py).

So I instead recommend deprecating and removing the attribute. 

Does it make any sense to deprecate something that was never actually supported in the first place? I would be tempted to simply replace it with an assert statement (or a `warnings.warn` call?) that fails if the attribute is present, and gives instructions to the developer to call `logging.root.addHandler()` instead.

misl6:
Fixed via #7988

Issue #7963, opened 2022-08-19
Proposal: Break the logger-jam


Julian-O:
There have been several issues (stale and open) discussions about the default behaviour of Kivy.logger. The discussions have been intertwined, because the functionality is intertwined. This attempts to summarise them in one place and reach an opinion/recommendation that should allow progress to happen mor quickly.

These issues are in scope here:

1. logger.py has the side effect of monkey-patching logging.Logger.root. 
	- This is a non-standard approach and causes surprising outcomes.

2. logger.py has the side effect of monkey-patching a redirect of sys.stderr.

3. logger ignores sublogger names, and uses colons to separate the subsystem field.
	- The rest of Kivy follows suit.
   
4. As a general question of coding practice: Should importing a file had any side effects?

Note: The ColoredFormatter deepcopy issue is out of scope here. It caused its own cluster of issues. #7962 attempts to address this, along with other clean up.

Previous Discussions
--------------------

* #4113

	- Proposes to address (3) with a systemic change.
	- Proposes to maintain (1).
	- Discussions in comments:
	   - Is Kivy an app, framework or a library?
	   - Other packages have tried similar approaches to Kivy and failed.
		  - examples: tornado and celery
	   - Discussion on general question of (4).
	   - Are there any examples of the current system causing problems?
		  - Yes; flask.
		  - Adding to the discussion: 
		    - Note: #4933 is another answer.

* #4933 

	- User, perhaps unwisely, removed all the handlers from the root logger.
	- Due to (1) this sent all Kivy logs to logging.lastResort, which outputs to stderr.
	- Due to (2) this caused an infinite loop.
	
* #4322

	- Complains about (2)
	- Proposes making it (and more of Kivy.logger?) optional.

* #4940

	- PR overrode default behaviour to turn off (1) unless explicitly set with an environment variable. 
	- Incidentally fixed one source of infinite recursion in #4933.
	- There was a comment saying it caused duplicate logs
		- Needs further investigation, but I suspect this is related to (2).
	- Closed in 2018.

* #6733
	- Complains about (1)
	- Proposes removing it.
	- Discussion talks about infinite loops, due to (2)
	- Points to `logging_tree` package as a useful debugging tool.
	- Repeats some of the framework v library discussion.
	- Considers logger config dict.
		- Concludes just having the ability to turn it off would be sufficient.

* #7172
	- Argues against (4).
	- Proposes to remove (2).
	- Argues against (but doesn't propose to remove) (1).
[...]

PR #7988, opened 2022-08-29
Added support for KIVY_LOG_MODE

### LINKED ISSUE #7980: Logger documentation is incorrect about default destination.
**Software Versions**
* Python: N/A
* OS: N/A
* Kivy: Since 2021
* Kivy installation method: N/A

**Describe the bug**

#6821 improved the documentation of the kivy.logger.

Since then, the file docstring states that Logger's features include:

> * output to `stdout` by default

However, the ConsoleHandler inherits from logging.StreamHandler without overriding the destination, and logging.StreamHandler sends to sys.*stderr* by default, not *stdout*.

**Expected behavior**

The documentatin should be match the behaviour. (I considered whether the behaviour should change, but it has used stderr for a long time, and changing now would break existing systems.)

Julian-O:
* Fixes #7980, #7981
* Substantially addresses #7963, but doesn't close it.
* Adds a test (marked xfail) to detect #7986, but doesn't fix.

Includes:

* Adds:
   * Detection of the env var, and setting handlers and propagation appropriately.
   * Shares (via __all__) a lot more of the internal classes, so clients can use the Kivy definitions (e.g. I would like to re-use the colored formatters in my code, but include logger name and timestamps in the format)
   * Documentation of new modes.
   * Mention of env vars in config.py.
* Removed:
   * No longer supports (undocumented) sys._kivy_logging_handler (#7981)
      * assert added, in case anyone is using it. (I couldn't find any real examples.)
      * IMO not an "api-break", but others may differ.
   *  No longer supports KIVY_LOG_MODE=TEST which has been subsumed by PYTHON mode.

* Test changes:
   * New unit tests
   * Limiting some unit tests to some modes.
   * Adding pytest markers to allow quick testing of relevant tests, without repeating unaffected tests.
   * Triggering of these tests in ci.
   * Removing KIVY_LOG_MODE=TEST which has been subsumed by PYTHON mode.
   * Added an xfail test for #7986.

Maintainer merge checklist

### DIFF
--- a/.ci/ubuntu_ci.sh
+++ b/.ci/ubuntu_ci.sh
@@ -104,7 +104,7 @@ test_kivy() {
   # Tests with default environment variables.
   env KIVY_NO_ARGS=1 python3 -m pytest --maxfail=10 --timeout=300 --cov=kivy --cov-branch --cov-report= "$(pwd)/kivy/tests"
   # Logging tests, with non-default log modes
-  env KIVY_NO_ARGS=1 KIVY_LOG_MODE=PYTHON python3 -m pytest -m logmodepython --maxfail=10 --timeout=300 --cov=kivy --cov-append --cov-branch "$(pwd)/kivy/tests"
+  env KIVY_NO_ARGS=1 KIVY_LOG_MODE=PYTHON python3 -m pytest -m logmodepython --maxfail=10 --timeout=300 --cov=kivy --cov-append --cov-report= --cov-branch "$(pwd)/kivy/tests"
   env KIVY_NO_ARGS=1 KIVY_LOG_MODE=MIXED python3 -m pytest -m logmodemixed --maxfail=10 --timeout=300 --cov=kivy --cov-append --cov-report=term --cov-branch "$(pwd)/kivy/tests"
 }
 
--- a/.ci/ubuntu_ci.sh
+++ b/.ci/ubuntu_ci.sh
@@ -103,8 +103,9 @@ test_kivy() {
   rm -rf kivy/tests/build || true
   # Tests with default environment variables.
   env KIVY_NO_ARGS=1 python3 -m pytest --maxfail=10 --timeout=300 --cov=kivy --cov-branch --cov-report= "$(pwd)/kivy/tests"
-  # Logging tests, with KIVY_LOG_MODE=TEST.
-  env KIVY_NO_ARGS=1 KIVY_LOG_MODE=TEST python3 -m pytest -m logmodetest --maxfail=10 --timeout=300 --cov=kivy --cov-append --cov-report=term --cov-branch "$(pwd)/kivy/tests"
+  # Logging tests, with non-default log modes
+  env KIVY_NO_ARGS=1 KIVY_LOG_MODE=PYTHON python3 -m pytest -m logmodepython --maxfail=10 --timeout=300 --cov=kivy --cov-append --cov-branch "$(pwd)/kivy/tests"
+  env KIVY_NO_ARGS=1 KIVY_LOG_MODE=MIXED python3 -m pytest -m logmodemixed --maxfail=10 --timeout=300 --cov=kivy --cov-append --cov-report=term --cov-branch "$(pwd)/kivy/tests"
 }
 
 test_kivy_benchmark() {
--- a/.ci/windows_ci.ps1
+++ b/.ci/windows_ci.ps1
@@ -110,7 +110,7 @@ function Test-kivy {
     python -m pytest --timeout=400 --cov=kivy --cov-branch --cov-report= "$(pwd)/kivy/tests"
     # Logging tests, with KIVY_LOG_MODE=TEST.
     $env:KIVY_LOG_MODE = 'PYTHON'
-    python -m pytest -m logmodepython --timeout=400 --cov=kivy --cov-append --cov-report=term --cov-branch "$(pwd)/kivy/tests"
+    python -m pytest -m logmodepython --timeout=400 --cov=kivy --cov-append --cov-branch "$(pwd)/kivy/tests"
     $env:KIVY_LOG_MODE = 'MIXED'
     python -m pytest -m logmodemixed --timeout=400 --cov=kivy --cov-append --cov-report=term --cov-branch "$(pwd)/kivy/tests"
 }
--- a/.ci/windows_ci.ps1
+++ b/.ci/windows_ci.ps1
@@ -109,8 +109,10 @@ function Test-kivy {
     # Tests with default environment variables.
     python -m pytest --timeout=400 --cov=kivy --cov-branch --cov-report= "$(pwd)/kivy/tests"
     # Logging tests, with KIVY_LOG_MODE=TEST.
-    $env:KIVY_LOG_MODE = 'TEST'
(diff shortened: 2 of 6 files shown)


Julian-O:
Also fixes #7172, #6733, #4322

Julian-O:
Sorry to edit mid-review, but this sneaky little fella came up while I was introducing PYTHON mode to one of my apps. I configured logging just the way I wanted, and then the Config file snuck in and changed the levels.

Julian-O:
I see the failing test on the clipboard functionality. My first instinct is to deny all blame.

Julian-O:
Yeah, the same test passes on my machine, but more importantly, passes on the ci actions on *my branch* that is being pulled here.  This appears to be an unreliable test (no idea why) rather than a fault in the my (unrelated) code. Has it been seen before?

misl6:
> Yeah, the same test passes (repeatedly) on my machine, but more importantly, passes on the ci actions on my branch that is being pulled here. This appears to be an unreliable test (no idea why) rather than a fault in the my (unrelated) code. Has it been seen before?

Yeah, that's a flaky test on our side, sometimes it fails. Re-started the job.

About the addition: no worries. I did a quick read at the code, but have not started the full review yet (unfortunately).

kivy/logger.py
```diff
@@ -302,14 +397,18 @@ def isatty(self):
 
 
 def logger_config_update(section, key, value):
-    if LOG_LEVELS.get(value) is None:
-        raise AttributeError('Loglevel {0!r} doesn\'t exists'.format(value))
-    Logger.setLevel(level=LOG_LEVELS.get(value))
+    if KIVY_LOG_MODE != "PYTHON":
+        if LOG_LEVELS.get(value) is None:
+            raise AttributeError('Loglevel {0!r} doesn\'t exists'.format(value))
+            Logger.setLevel(level=LOG_LEVELS.get(value))
```
misl6, line 403: Should not look like the following?

```suggestion
    if KIVY_LOG_MODE != "PYTHON":
        if LOG_LEVELS.get(value) is None:
            raise AttributeError('Loglevel {0!r} doesn\'t exists'.format(value))
        Logger.setLevel(level=LOG_LEVELS.get(value))
```

kivy/logger.py
```diff
@@ -464,39 +575,42 @@ def is_color_terminal():
     else None
 )
 
+assert not hasattr(sys, '_kivy_logging_handler'), \
+    "Not supported. Try logging.root.addHandler()"
+
```
misl6, line 580: Maybe we can add a reference to issue https://github.com/kivy/kivy/issues/7981, so if it happens that someone looks at this chunk of code in 1/2 releases can take the occasion to just delete this line?

.ci/ubuntu_ci.sh
```diff
@@ -103,8 +103,9 @@ test_kivy() {
   rm -rf kivy/tests/build || true
   # Tests with default environment variables.
   env KIVY_NO_ARGS=1 python3 -m pytest --maxfail=10 --timeout=300 --cov=kivy --cov-branch --cov-report= "$(pwd)/kivy/tests"
-  # Logging tests, with KIVY_LOG_MODE=TEST.
-  env KIVY_NO_ARGS=1 KIVY_LOG_MODE=TEST python3 -m pytest -m logmodetest --maxfail=10 --timeout=300 --cov=kivy --cov-append --cov-report=term --cov-branch "$(pwd)/kivy/tests"
+  # Logging tests, with non-default log modes
+  env KIVY_NO_ARGS=1 KIVY_LOG_MODE=PYTHON python3 -m pytest -m logmodepython --maxfail=10 --timeout=300 --cov=kivy --cov-append --cov-branch "$(pwd)/kivy/tests"
```
misl6, line 107: Is `cov-report=` missing intentionally?

misl6 (changes requested):
Generally looks good!
I've left some comments and suggestions.

That opens some potential fixes and improvements on kivy "sister" projects like `pyobjus` and `pyjnius` which may take some advantage of it.

Feel free to ping me back when you're ready for a new review.

kivy/logger.py
```diff
@@ -1,19 +1,36 @@
...
-The Kivy `Logger` class provides a singleton logger instance. This instance
-exposes a standard Python
-`logger <https://docs.python.org/3/library/logging.html>`_ object but adds
-some convenient features.
+The Kivy ``Logger`` class provides a singleton logging.logger instance.
 
-All the standard logging levels are available : `trace`, `debug`, `info`,
-`warning`, `error` and `critical`.
+As well as the standard logging levels (``debug``, ``info``,
+``warning``, ``error`` and ``critical``), a new ``trace`` level is available.
```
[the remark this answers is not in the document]
Julian-O: Went with "additional"

kivy/logger.py
Julian-O: Thank you!

Julian-O:
Thank you very much for the review. I have submitted the fixes.

Happy to hear it may help on the other projects - I went and checked their issue-tracker to see if I could find any specifics, but couldn't recognise any.

Julian-O:
@misl6: This is an ever so gentle ping, just in case you missed my updates, while trying not to sound at all entitled to the volunteered time of a busy maintainer...

.ci/windows_ci.ps1
```diff
@@ -109,8 +109,10 @@ function Test-kivy {
     # Tests with default environment variables.
     python -m pytest --timeout=400 --cov=kivy --cov-branch --cov-report= "$(pwd)/kivy/tests"
     # Logging tests, with KIVY_LOG_MODE=TEST.
-    $env:KIVY_LOG_MODE = 'TEST'
-    python -m pytest -m logmodetest --timeout=400 --cov=kivy --cov-append --cov-report=term --cov-branch "$(pwd)/kivy/tests"
+    $env:KIVY_LOG_MODE = 'PYTHON'
+    python -m pytest -m logmodepython --timeout=400 --cov=kivy --cov-append --cov-branch "$(pwd)/kivy/tests"
```
misl6, line 113: Is `cov-report=` missing intentionally?

kivy/config.py
```diff
@@ -120,6 +120,12 @@
         Path of log directory.
     `log_enable`: int, 0 or 1
         Activate file logging. 0 is disabled, 1 is enabled.
+
+        .. note::
+            Logging outout can also be controlled by the environment variables
```
misl6, line 125:
```suggestion
        .. note::
            Logging output can also be controlled by the environment variables
```

misl6 (approved):
LGTM, I've only left 2 comments on something that got lost during the first review.

misl6:
> @misl6: This is an ever so gentle ping, just in case you missed my updates, while trying not to sound at all entitled to the volunteered time of a busy maintainer...

Don't worry, feel free to ping !

I guess we're 99.98% ready to merge :)

.ci/windows_ci.ps1
Julian-O: Julian-O: Well, `cov-report=` shouldn't be there because we don't want a...
[@misl6 raises eyebrow]
[Julian-O stops, re-reads the fine-print in the manual again. Oh!]
Julian-O: Ahem, as I was saying, `cov-report=` should definitely be there because we don't want a report.

Julian-O:
Should be good now.

misl6 (approved):
LGTM. Thank you!

PR #7989, opened 2022-08-30
Android Lifecycle convergence

RobertFlatt:
Maintainer merge checklist

### DIFF
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -284,9 +284,10 @@ The currently implemented Pause mechanism is:
        System due to the user switching to another application, a phone
        shutdown or any other reason.
     #. :meth:`App.on_pause` is called:
-    #. If False is returned or `App.on_pause` has no return statement, then :meth:`App.on_stop` is called.
-    #. If True is returned or `App.on_pause` is not defined, the application will sleep until
-       the OS resumes our App.
+    #. If False is returned or `App.on_pause` has no return statement, then
+       :meth:`App.on_stop` is called.
+    #. If True is returned or `App.on_pause` is not defined, the application
+       will sleep until the OS resumes our App.
     #. When the app is resumed, :meth:`App.on_resume` is called.
     #. If our app memory has been reclaimed by the OS, then nothing will be
        called.
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -284,8 +284,8 @@ The currently implemented Pause mechanism is:
        System due to the user switching to another application, a phone
        shutdown or any other reason.
     #. :meth:`App.on_pause` is called:
-    #. If False is returned, then :meth:`App.on_stop` is called.
-    #. If True is returned (default case), the application will sleep until
+    #. If False is returned or `App.on_pause` has no return statement, then :meth:`App.on_stop` is called.
+    #. If True is returned or `App.on_pause` is not defined, the application will sleep until
        the OS resumes our App.
     #. When the app is resumed, :meth:`App.on_resume` is called.
     #. If our app memory has been reclaimed by the OS, then nothing will be
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -995,6 +995,7 @@ Context.html#getFilesDir()>`_ is returned.
 
         On Android set OS state to pause, Kivy app state follows.
         No functionality on other OS.
+        .. versionadded:: 2.2.0
         '''
         if platform == 'android':
             from android import mActivity
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -284,10 +284,10 @@ The currently implemented Pause mechanism is:
        System due to the user switching to another application, a phone
        shutdown or any other reason.
     #. :meth:`App.on_pause` is called:
-    #. If False is returned or :meth:`App.on_pause` has no return statement, then
-       :meth:`App.on_stop` is called.
-    #. If True is returned or :meth:`App.on_pause` is not defined, the application
-       will sleep until the OS resumes our App.
+    #. If False is returned or :meth:`App.on_pause` has no return statement,
+       then :meth:`App.on_stop` is called.
+    #. If True is returned or :meth:`App.on_pause` is not defined, the
+       application will sleep until the OS resumes our App.
     #. When the app is resumed, :meth:`App.on_resume` is called.
     #. If our app memory has been reclaimed by the OS, then nothing will be
        called.
@@ -1001,6 +1001,8 @@ Context.html#getFilesDir()>`_ is returned.
         if platform == 'android':
             from android import mActivity
             mActivity.moveTaskToBack(True)
+        else:
+            Logger.info('App.pause() is not available on this OS.')
 
     def on_start(self):
         '''Event handler for the `on_start` event which is fired after
--- a/kivy/app.py
+++ b/kivy/app.py
@@ -989,7 +989,7 @@ Context.html#getFilesDir()>`_ is returned.
             for child in self._app_window.children:
                 self._app_window.remove_widget(child)
         App._running_app = None
-      
+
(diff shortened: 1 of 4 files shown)


RobertFlatt:
## Executive Summary

On a mobile device (unlike a desktop) the Kivy Lifecycle is an abstraction of a lifecycle implemented in the OS. The Kivy and OS state machines muct remain synchronized to avoid ambigious states.

This PR addresses differences between these state machines on Android, it may be a template for extending these observations to iOS.

## Issues

On Android the cases where the Kivy state machine diverges from the Android state machine are:

 - 1) Back Gesture/Button | App:  stop | Android : pause |

 - 2) on_pause False | App : stop | Android : no change |

 - 3) app.stop() | App: stop | Android : no change |

 - 4) app.pause() | App : N/A | Android : N/A |

 - 5) Android sleep->destroy | App : sleep | Android : destroyed |

 - 6) Android destroy app | App: service quit | Android : service no change  |


In this PR we address issues i thru iv. Issues v and vi are not Kivy issues.

## Discussion and proposed changes

 - Cases i-iii) In general the Kivy state machine follows the Android state machine via events from SDL2. However in these three cases Kivy sets its own state without informing Android. The cases are addressed by replacing Kivy state changes with calls to Android telling it to change state. The Android state change will in turn tell Kivy to change state via SDL2. 

Both Android api calls used in this PR are available on all current p4a supported Android versions.
https://developer.android.com/reference/android/app/Activity#moveTaskToBack(boolean)
https://developer.android.com/reference/android/app/Activity#finishAndRemoveTask()

 - Case iv) is currently this is not implemented in Kivy, it is implemented in this PR (in the same manner as the first 3 cases) for completeness.

 - Case v) Is not implemented in SDL2 https://github.com/libsdl-org/SDL/issues/3297 . Hence Kivy fails to issue an 'on_stop' - this is documented. There are no other known side effects. If it is an issue, it is an SDL2 issue. 

 - Case vi) The maximum lifetime of a Kivy Android service is the lifetime of the Kivy App (that is until the app is killed, not until the app is paused). This behavior is not changed by this PR. However this is not the default behavior of a service in a Java Android app. If it is an issue, it is a p4a issue. 

 - Case ii Revisited) The case is addressed in this PR, but I really think the 'on_pause return' feature should be removed. The utility is unclear to me (`self.stop()` exists and is more intuative) and there is an unfortunate side effect. As designed, if a user omits the `return` statement in `on_pause` there will be no effect on desktop execution (on_pause is not called) but on mobile the app will correctly dissappear from the screen and the task list. The connection between a simple coding error and the app behavior is not intuative. This PR logs this event, and corrects an ambiguity in the comment/documentation. If there is interest in removing the functionality I can create a PR for this.

kivy/core/window/__init__.py
```diff
@@ -1984,9 +1984,15 @@ def on_keyboard(self, key, scancode=None, codepoint=None,
...
+        # Consume the event and tell Android to pause.
         # TODO If just CMD+w is pressed, only the window should be closed.
         is_osx = platform == 'darwin'
-        if WindowBase.on_keyboard.exit_on_escape:
+        if key == 27 and platform == 'android':
+            from android import mActivity
+            mActivity.moveTaskToBack(True)
+            return True
+        elif WindowBase.on_keyboard.exit_on_escape:
             if key == 27 or all([is_osx, key in [113, 119], modifier == 1024]):
```
NomadDemon, line 1996: just question:
shouldnt we do some key mapper to constants for keys such "27" (android) and 113 like here etc?

RobertFlatt: No, because it is a change that would be mission creep for this PR. This PR uses the existing conventions for the express purpose of being clear about the functional changes. 

If this kind of thing is important to you, please submit a PR.
- Explain why you propose the changes.
- Show that the changes are consistent across the Kivy code base.
- Think about 27 which is not ESC on Android, it just looks that way to a user because of this https://github.com/kivy/kivy/blob/master/kivy/core/window/window_sdl2.py#L209-L211

NomadDemon: I was asking to make it in this PR :) but just asked question globally, in future [separate PR], whatever :)

But you are maybe right. I was thinking in same way pygame was handling keys, they were mapped like K_BACKSPACE, K_ESCAPE, K_SLASH etc.

I though it might improve readability of code somehow

kivy/app.py
```diff
@@ -982,6 +991,17 @@ def stop(self, *largs):
...
+    def pause(self, *largs):
+        '''Pause the application.
+
+        On Android set OS state to pause, Kivy app state follows.
+        No functionality on other OS.
+        .. versionadded:: 2.2.0
+        '''
+        if platform == 'android':
+            from android import mActivity
+            mActivity.moveTaskToBack(True)
```
misl6, line 1003: Minor: Maybe we can add a log that warns the developer about the non-availability?

misl6 (commented):
- Tested on runtime (on API 31) with all the possible combinations ✅ 
- Code LGTM. (we can do something to avoid some minor DRY violations, but we can target it later) ✅ 
- Left some minor suggestions 🧐 

Thank you!

Regarding:
> Case ii Revisited) The case is addressed in this PR, but I really think the 'on_pause return' feature should be removed. The utility is unclear to me (self.stop() exists and is more intuative) and there is an unfortunate side effect. As designed, if a user omits the return statement in on_pause there will be no effect on desktop execution (on_pause is not called) but on mobile the app will correctly dissappear from the screen and the task list. The connection between a simple coding error and the app behavior is not intuative. This PR logs this event, and corrects an ambiguity in the comment/documentation. If there is interest in removing the functionality I can create a PR for this.

I guess we can discuss it on a separate issue, and if we decide to remove that feature, it will need to be deprecated at first, and then removed in 2/3 releases.

RobertFlatt:
> but I really think the 'on_pause return' feature should be removed. 

> it will need to be deprecated at first, and then removed in 2/3 releases

Faced with inertia I'm less enthusiastic!

Its really about cleaning up some fuzzy design, not a bug. So not something I'd fight hard for. 

A quick search finds 3 cases:
```
kivy-master>grep -R \'on_pause\'
kivy/core/window/window_sdl2.py:            elif not app.dispatch('on_pause'):
kivy/core/window/window_sdl2.py:        elif not app.dispatch('on_pause'):
kivy/support.py:        if app.dispatch('on_pause'):
```
I could add a Logger.warning message in the return False case. The return True case is a NOP and of course very common, a warning in that case would  generate a storm of confusion.

kivy/app.py
RobertFlatt: Done

misl6 (approved):
LGTM. Thank you!

misl6:
> > but I really think the 'on_pause return' feature should be removed.
> 
> > it will need to be deprecated at first, and then removed in 2/3 releases
> 
> Faced with inertia I'm less enthusiastic!
> 
> Its really about cleaning up some fuzzy design, not a bug. So not something I'd fight hard for.
> 
> A quick search finds 3 cases:
> 
> ```
> kivy-master>grep -R \'on_pause\'
> kivy/core/window/window_sdl2.py:            elif not app.dispatch('on_pause'):
> kivy/core/window/window_sdl2.py:        elif not app.dispatch('on_pause'):
> kivy/support.py:        if app.dispatch('on_pause'):
> ```
> 
> I could add a Logger.warning message in the return False case. The return True case is a NOP and of course very common, a warning in that case would generate a storm of confusion.

Let's open an issue to keep it tracked! 😃

RobertFlatt:
For anybody reading this in my future, it turns out issue iv) 'service lifetime' has previously been fixed https://github.com/kivy/python-for-android/pull/2401

Issue #7995, opened 2022-09-03
Textinput does not render emojis correctly


RobinPicard:
**Software Versions**
* Python: 3.9
* OS: MacOS BigSur 11.4
* Kivy: v2.2.0.dev0
* Kivy installation method: development

**Describe the bug**
The textinput widgets render the colored emojis as completely black contrarily to labels that render them corrected

**Expected behavior**
The emojis should be colored

**To Reproduce**
```
from kivy.core.window import Window
from kivy.app import App
from kivy.uix.label import Label
from kivy.uix.boxlayout import BoxLayout
from kivy.uix.textinput import TextInput

Window.clearcolor = (1, 1, 1, 1)

class CustomBoxLayout(BoxLayout):
    def __init__(self, *args):
        super().__init__(*args)
        self.a_textinput = TextInput(background_normal='', font_name='NotoColorEmoji.ttf')
        self.a_label = Label(font_name='NotoColorEmoji.ttf')
        self.add_widget(self.a_textinput)
        self.add_widget(self.a_label)
        self.a_textinput.bind(text=self.update_label)

    def update_label(self, instance, text, *args):
        self.a_label.text = text

class MyApp(App):
    def build(self):
        return CustomBoxLayout()

if __name__ == '__main__':
    MyApp().run()
```

**Code and Logs and screenshots**
<img width="301" alt="Capture d’écran 2022-09-03 à 19 34 03" src="https://user-images.githubusercontent.com/83579270/188282047-57343112-f9dd-4e53-a337-586d8493c481.png">

**Additional context**
I think I found the origin of the problem, I will open a PR shortly

Julian-O:
@RobinPicard: Just a ping to see how you are going with the PR review comments.

Julian-O:
Related #7565

RobinPicard:
> @RobinPicard: Just a ping to see how you are going with the PR review comments.

I'm not working on Kivy-related things anymore, sorry

Julian-O:
To others reading this, we have a partially implemented PR, with review comments, that needs a volunteer to rebase, address the comments, and commit a new PR.. Could be an easy way to earn Internet points.

Minimuino:
> @DexerBR has stepped forward with #8491. Thank you muchly!

Thanks for your work! Hope this gets fixed soon!

Huxtech:
I’ve run into this issue before and was able to find a working solution for rendering emojis correctly in Kivy Textinput. 
[...]

PR #7997, opened 2022-09-04
Add colored emoji support for TextInputs

RobinPicard:
## Description

This PR aims at answering the following issue: https://github.com/kivy/kivy/issues/7995

Currently, Labels render colored emojis correctly but not TextInputs. I believe the origin of the problem is that when a TextInput creates Labels to get the textures for each line, it does not do it in the same way as if it were a standard stand-alone Label. More specifically, the Labels are created without a value for the `color` property. As a result, the texture is made of pixels in shades of grey (except for emojis). Later, when the textures are added to the Canvas of the TextInput, they are colored by a Color instruction based on the value of the `foreground_color` property of TextInput. That way, the whome texture is made in shades of that given color. This is not a problem for regular text but that does not work for emojis that end being recolored monochromatically.

The solution is to treat Labels used in TextInputs like regular Labels by setting a value the `color` property following the same logic that was contained in the Color instruction (the graphic provider will not apply the `color` to emojis by itself I believe). Then, the Color instruction of the canvas is set to (1, 1, 1, 1) such that the texture's color are not modified.

## Testing
The tests `test_uix_textinput` are successful and the few tests I've done with the example in the issue work fine

## Maintainer merge checklist

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -196,7 +196,7 @@
             pos: self._cursor_visual_pos
             size: root.cursor_width, -self._cursor_visual_height
         Color:
-            rgba: self.disabled_foreground_color if self.disabled else (self.hint_text_color if not self.text else self.foreground_color)
+            rgba: (1, 1, 1, 1)
 
 <TextInputCutCopyPaste>:
     content: content.__self__
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -566,6 +566,9 @@ class TextInput(FocusBehavior, Widget):
         fbind('font_family', refresh_line_options)
         fbind('base_direction', refresh_line_options)
         fbind('text_language', refresh_line_options)
+        fbind('foreground_color', refresh_line_options)
+        fbind('disabled_foreground_color', refresh_line_options)
+        fbind('hint_text_color', refresh_line_options)
 
         def handle_readonly(instance, value):
             if value and (not _is_desktop or not self.allow_copy):
@@ -2638,7 +2641,7 @@ class TextInput(FocusBehavior, Widget):
 
         kw = self._get_line_options()
         kw['color'] = self.disabled_foreground_color if self.disabled\
-            else (self.hint_text_color if not text else self.foreground_color)
+            else (self.hint_text_color if hint else self.foreground_color)
         cid = '%s\0%s' % (ntext, str(kw))
         texture = Cache_get('textinput.label', cid)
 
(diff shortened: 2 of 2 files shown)


kivy/data/style.kv
```diff
@@ -196,7 +196,7 @@
             pos: self._cursor_visual_pos
             size: root.cursor_width, -self._cursor_visual_height
         Color:
-            rgba: self.disabled_foreground_color if self.disabled else (self.hint_text_color if not self.text else self.foreground_color)
+            rgba: (1, 1, 1, 1)
```
tshirtman, line 199: The previous version had the advantage of automatically adjusting the rendering if any of the `xxx_color` properties, or `disabled` was updated, the new way means lines have to be rerendered, for it to work. so binding on these properties, probably around line 568, of textinput.py, would be necessary, otherwise these color/disabled changes won't be applied until the text (or some other property that is bound) is actually changed.

tshirtman (commented):
Looks promising! I think it creates an issue with binding, see comment, but shouldn't be too hard to check and fix.

kivy/data/style.kv
RobinPicard: Ah yes, thanks for the help!

RobinPicard: Does it look fine to you now @tshirtman?

kivy/uix/textinput.py
```diff
@@ -566,6 +566,9 @@ def __init__(self, **kwargs):
         fbind('font_family', refresh_line_options)
         fbind('base_direction', refresh_line_options)
         fbind('text_language', refresh_line_options)
+        fbind('foreground_color', refresh_line_options)
+        fbind('disabled_foreground_color', refresh_line_options)
+        fbind('hint_text_color', refresh_line_options)
```
DexerBR, line 571: I recommend binding to `update_text_options`, binding to `refresh_line_options` will move the cursor to the end of the text.

So you can move this code below, to the line below [line 588](https://github.com/kivy/kivy/pull/7997/files#diff-67d848ba06b2f91032fd6dc59a8956275118595997f9737872a39fb5983183d2R588)

```python
fbind('foreground_color', update_text_options)
fbind('disabled_foreground_color', update_text_options)
fbind('hint_text_color', update_text_options)
```

DexerBR (changes requested):

Julian-O:
OP has declared they are not working on Kivy any more. This is an orphan PR that needs fostering.

Issue #7998, opened 2022-09-05
Labels sometimes do not "cut lines" at the right place


RobinPicard:
**Software Versions**
* Python: 3.9
* OS: MacOS BigSur 11.4
* Kivy: v2.2.0.dev0
* Kivy installation method: development

**Describe the bug**
Labels with a width constraint do not "cut lines" at the right place in some specific cases (see code and picture below). It happens in some cases when there is no space in a series of characters wider than the width of the widget. In many languages that's a rare edge case but it's actually not that uncommon in languages like Chinese where the space is not a distinct but is "integrated" in the character for the coma and the period.

**To Reproduce**
```
from kivy.app import App
from kivy.lang import Builder

KV = '''
Label:
    text: 'Untextassezlongquiracontrenimportequoisansaucunespace.\\n\\na'
    font_size: 30
    size_hint: (None, None)
    size: 500, 500
    text_size: self.size
    canvas.before:
        Color:
            rgba: (1, 0, 0, 1)
        Rectangle:
            pos: self.pos
            size: self.size
'''
root = Builder.load_string(KV)

class WeatherApp(App):
    def build(self):
        return root

if __name__ == '__main__':
        WeatherApp().run()
```

***Expected behavior**
The `U` should not be on the first line alone. Instead, the first line should be full with a break happening after the `e`

**Screenshot**
<img width="250" alt="Capture d’écran 2022-09-05 à 23 38 21" src="https://user-images.githubusercontent.com/83579270/188514515-016ab01d-c639-4064-9ec3-d49635d3bd55.png">

matham:
https://github.com/kivy/kivy/pull/7999

PR #7999, opened 2022-09-05
Modify layout fix bug in how long text without space is cut 

RobinPicard:
This PR aims at answering the following issue: https://github.com/kivy/kivy/issues/7998

As explained in the issue, the function `layout_text` cuts a line at the wrong place in certain cases. Upon reading the code source for this function, I came to the conclusion that it's probably caused by the fact that the value of the variable `is_space` is not reset to 0 at each iteration over the lines. I think so because:
- The change I propose below solved the example presented in the issue and did not cause any problem in the tests I've done (which consisted in trying out many random strings)
- I can't make sense of why we would initially give the value 0 to `is_space` but then not set this starting value again for each line leading to a situation in which lines can influence each other's layout
However, there's a lot I do not understand in this file, so maybe there's just something I don't get

Maintainer merge checklist

### DIFF
--- a/kivy/core/text/text_layout.pyx
+++ b/kivy/core/text/text_layout.pyx
@@ -397,7 +397,7 @@ def layout_text(object text, list lines, tuple size, tuple text_size,
     cdef int lhh, lww, k, bare_h, dwn = append_down, pos = 0
     cdef object line, ln, val, indices
     cdef LayoutLine _line
-    cdef int is_space = 0
+    cdef int is_space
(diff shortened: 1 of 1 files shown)


tshirtman (approved):
Explanation looks logical to me and the fix quite straightforward, so I'm inclined to merge it.
If possible, a test for that case preventing regression would be nice.

matham (approved):
The fix makes sense to me.

PR #8002, opened 2022-09-08
Fixing the Persian and Arabic words problem in kivy

SaeidJavadi:
As you know Kivy does not support all languages well, I tried to solve the RTL problem by making these changes


misl6:
Hi @SaeidJavadi !

I guess you actually wanted to open a PR on master, not stable.

PR #8032, opened 2022-10-13
doc: Initial remarks on *BSD compatibility.

cederom:
Supplemented documentation with initial information on *BSD Unix (i.e. FreeBSD) compatibility.

Signed-off-by: Tomasz 'CeDeROM' CEDRO <tomek@cedro.info>

Maintainer merge checklist


doc/sources/gettingstarted/installation.rst
```diff
@@ -220,10 +227,14 @@ on Windows::
...
 
-or in bash, Linux and macOS::
+in bash, Linux and macOS::
 
     python kivy_venv/share/kivy-examples/demo/showcase/main.py
 
+on *BSD Unix:
+
+    python3 kivy_venv/share/kivy-examples/demo/showcase/main.py
+
```
misl6, line 237: Minor: Why a separate one?

cederom: hello @misl6 :-)
* it is about `python` vs `python2` vs `python3`.
* on *BSD (i.e. FreeBSD) base system installed in `/` location, while all userland is located in `/usr/local` that may include Python installations.
* base system cannot be modified by hand by the user, as it may come from ROM/NFS etc.. also that would lead to inconsistency in behavior among other systems, what is highly avoided here.
* thus using `python3` instead `python` as the most correct way.
* macOS is also kind of BSD but adapted to Linux non-standards, for instance using `python3` on macOS would work.. but this is unknown on Linux and depends on a distro.
* In a virtualenv `python` is the same as `python3` but outside virtualenc that would depend on the system configuration and packages.
* maybe the safest way is to simply use `python3` everywhere in the docs ?

misl6: > it is about python vs python2 vs python3.

Oh, now it's clear. I did not seen the `3` suffix. (A new pair of 👓 is needed)

> In a virtualenv python is the same as python3 but outside virtualenc that would depend on the system configuration and packages.

>maybe the safest way is to simply use python3 everywhere in the docs ?

Yeah, that's another kind of story, but we will likely need migrate to `python3` in docs, as if the user did not followed the advice of using a virtualenv, that may lead to an error.

misl6 (approved):
LGTM. Thank you!

doc/sources/gettingstarted/installation.rst
cederom: For venv this is fine, but I want to add system package on FreeBSD do Kivy apps will run out-of-the-box here :-)

Kivy app runs even on a terminal with no Xorg! I tried that around 20 years ago with other libs but that required some hacking.. now it seems standard feature :-)

PR #8035, opened 2022-10-16
Simplify Animation._unregister

nanouasyn:
Use a more idiomatic way to remove an element from a set if the element is in the set.

Maintainer merge checklist

### DIFF
--- a/kivy/animation.py
+++ b/kivy/animation.py
@@ -294,8 +294,7 @@ class Animation(EventDispatcher):
         Animation._instances.add(self)
 
     def _unregister(self):
-        if self in Animation._instances:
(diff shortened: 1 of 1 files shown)


gottadiveintopython:
I think this is good for both readability and maybe performance.

My concern is `set.discard()` is available on all the python versions that are supported by Kivy? I couldn't find that information from [the official doc](https://docs.python.org/3/tutorial/datastructures.html#sets). But I guess the answer is yes since the unittests didn't fail.

misl6 (approved):
LGTM. Thank you @nanouasyn , and thank you @gottadiveintopython for the review.

PR #8051, opened 2022-11-03
fix module kivy.modules.console

aorizondo:
When activated, the module console crash due a loop calling kivy.__init__
Both changes prevent the loop
`Component: kivy.modules.console`
`api-break`


misl6:
Hi @aorizondo !

Do you have a log snippet or something else that could help us to reproduce the issue?

PR #8056, opened 2022-11-12
New Feature: Add `BoxShadow` graphic instruction 🎉

DexerBR:
# BoxShadow

This PR inserts BoxShadow as a new graphical instruction. Also includes documentation.

boxshadow

## Usage

Like other graphical instructions, BoxShadow will follow the same patterns used for the declaration inside the canvas:

boxshadow_demo

```kvlang
<MyWidget>:
    Button:
        pos_hint: {"center_x": 0.5, "center_y": 0.5}
        size_hint: None, None
        size: 200, 150
        background_down: self.background_normal
        canvas.before:
            Color:
                rgba: 0, 0, 1, 0.85
            BoxShadow:
                pos: self.pos
                size: self.size
                offset: 0, -10
                spread_radius: -20
                border_radius: 10, 10, 10, 10
                blur_radius: 80 if self.state == "normal" else 50
```

## Key features

✅  **Fast** → The shader used uses the technique of building a rounded rectangle with a edge gradient, which delivers a result similar to the gaussian blur, however, much more efficiently.

  - Some code references used to build the BoxShadow shader:
    - https://www.shadertoy.com/view/3tj3Dm
    - https://www.shadertoy.com/view/NtVSW1
    - https://iquilezles.org/articles/distfunctions2d/

✅  **Efficient** → Fbo is created only at startup with a predefined size. Additional computation will not (unnecessarily) recreate the Fbo.

✅  **OpenGL ES2 GLSL compatible**

✅  **No FPS drop** → The base class used to construct the instruction is Fbo. Therefore, there is no need to compute the effect update every frame, but only when any of the properties requests the redraw. Also, the functions used in the fragment shader are fast enough to avoid performance drops.

✅  **Adaptability** → As it is a graphical instruction, the flexibility and ease of integration with other UI elements increases significantly.

## Points for improvement in the future

This is not a direct problem of BoxShadow, but creating thousands of Fbos may take a few seconds on some devices. This is due to the slow creation of the texture by the `Texture` module used for creating textures in Kivy.
Improvements in texture creation time or some improvements in Fbo creation time will benefit not only this `BoxShadow` instruction, but the entire Kivy framework.

Maintainer merge checklist


misl6 (commented):
Hi @DexerBR !

Thank you very much for this nice feature!
I've tested it and reviewed the code, and everything looks good!

Do yo mind about adding some test here?https://github.com/kivy/kivy/blob/master/kivy/tests/test_graphics.py

DexerBR:
> Hi @DexerBR !
> 
> Thank you very much for this nice feature! I've tested it and reviewed the code, and everything looks good!
> 
> Do yo mind about adding some test here?https://github.com/kivy/kivy/blob/master/kivy/tests/test_graphics.py

Hi @misl6! Thanks for the review!

Certainly! I was waiting for a review before writing/adding the tests... I will do it soon!

misl6 (approved):
LGTM. Thank you!

Issue #7983, opened 2022-08-27
Proposal: Stick to a specific `SDL2`, `SDL_ttf`, `SDL_image`, `SDL_mixer` version or add a minimum supported one.


misl6:
**Is your feature request related to a problem? Please describe.**

Currently, we don't specify a target or minimum supported version for the following dependencies on **Linux** and **macOS**:
- `SDL2`
- `SDL_ttf`
- `SDL_image`
- `SDL_mixer`

Or, at least, we specify a version, but only for the CI-generated wheels and `Kivy.app` builds.

That makes it hard to test changes or add new features exposed by newer versions of the `SDL2` libraries ecosystem. (See: https://github.com/kivy/kivy/pull/7982 CI failure on `Ubuntu Unittests`)

**Describe the solution you'd like**

Is hard to force the contributors (or someone who wants to build from sources) to use a specific version, but at least we need to consider adding a "minimum supported version", which should be checked at the very beginning of the build process.

If system-provided dependencies are too old (as it's happening on the CI), or the developer wants to stick to the same exact version that we use in the CI, a build script for the non-python-packages dependencies could be exposed and documented.

We already maintain a build script for SDL2 dependencies as the CI-generated wheels and `Kivy.app` already makes use of a self-built version, so should not be an issue.

**Describe alternatives you've considered**

1. **(Could be the easiest one to maintain and to avoid some support requests)**
Force the contributors (or someone who wants to build from sources) to use a specific self-built library set that can be built via the build script. System-provided deps may be forced via `KIVY_FORCE_SYSTEM_DEPS=1`. That allows us to only target a specific version of the SDL2 library, and even use a patched/unreleased version if it contains a specific bugfix/feature.

2. Do not specify a minimum supported version, but add guards to avoid using an unsupported feature.

**Additional context**
- On macOS that situation may be unlikely, as in the docs we state to use `brew` or directly download and install the `Framework`. `brew` usually supports the latest stable version for all the supported macOS versions, and the `Framework` links could be easily updated.
- On Windows, that's not an issue, as we provide our own version of `sdl2` via `kivy_deps`
- I love Linux distros, but Linux is where the headache starts. Unfortunately, the latest version of `sdl2` libraries are only packaged and distributed (via the default package manager and sources) for the latest version of the OS. (Or at least, that happens on Ubuntu, e.g. `18.04 LTS` is still on sdl2 `2.0.8`)

[...]

PR #8058, opened 2022-11-16
Makes Windows DPI aware of scale changes

DexerBR:
On Windows, changing the scale factor will result in the window stretching, rather than changing kivy's internal DPI metrics. As a result, the window appears with a blurred interface.

### Below is the behavior when setting the scale to 125%.

- **Previously (kivy → `2.1.0`):**

  master

- **After this PR:**

  pr

<br>

[SDL_HINT_WINDOWS_DPI_SCALING](https://github.com/libsdl-org/SDL/blob/0dfc829a6b75b5a3c4c1d49fb943b622c12d67bc/include/SDL_hints.h#L2082) was introduced in SDL `2.24.0`, with two options:

- `"0"` - SDL coordinates equal Windows coordinates. No automatic window resizing when dragging between monitors with different scale factors (unless this is performed by Windows itself, which is the case when the process is DPI unaware).
- `"1"` - SDL coordinates are in DPI-scaled points. Automatically resize windows as needed on displays with non-100% scale factors.

Setting SDL_HINT_WINDOWS_DPI_SCALING to "1" will automatically request process DPI awareness and enable the SDL_WINDOW_ALLOW_HIGHDPI flag for all windows.

This change updates the SDL coordinate system units to use DPI-scaled points instead of pixels on Windows. This ensures that windows are appropriately sized regardless of the display's DPI scaling settings, such as on high-DPI displays.

For example, requesting a 640x480px window on a display with 125% scaling will result in an window with 800x600px.

With these modifications, a previous workaround that resized the screen on every DPI change has been removed.

For testing:

```python
from kivy.app import App
from kivy.metrics import dp


class TestDPI(App):
    def build(self):
        return Button(text=f"1dp is equivalent to {dp(1)}dp", font_size=40)


TestDPI().run()
```

<br>

Maintainer merge checklist

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -420,7 +420,7 @@ class WindowBase(EventDispatcher):
     def _get_size(self):
         r = self._rotation
         w, h = self._size
-        if self._density != 1:
+        if platform == 'win' or self._density != 1:
             w, h = self._win._get_gl_size()
         if self.softinput_mode == 'resize':
             h -= self.keyboard_height
@@ -519,7 +519,7 @@ class WindowBase(EventDispatcher):
     # make some property read-only
     def _get_width(self):
         _size = self._size
-        if self._density != 1:
+        if platform == 'win' or self._density != 1:
             _size = self._win._get_gl_size()
         r = self._rotation
         if r == 0 or r == 180:
@@ -536,7 +536,7 @@ class WindowBase(EventDispatcher):
         '''Rotated window height'''
         r = self._rotation
         _size = self._size
-        if self._density != 1:
+        if platform == 'win' or self._density != 1:
             _size = self._win._get_gl_size()
         kb = self.keyboard_height if self.softinput_mode == 'resize' else 0
         if r == 0 or r == 180:
@@ -739,13 +739,13 @@ class WindowBase(EventDispatcher):
     '''
 
     def _get_effective_size(self):
-        '''On density=1 and non-ios displays, return :attr:`system_size`,
-        else return scaled / rotated :attr:`size`.
+        '''On density=1 and non-ios / non-Windows displays,
+        return :attr:`system_size`, else return scaled / rotated :attr:`size`.
 
         Used by MouseMotionEvent.update_graphics() and WindowBase.on_motion().
         '''
         w, h = self.system_size
-        if platform == 'ios' or self._density != 1:
+        if platform in ('ios', 'win') or self._density != 1:
             w, h = self.size
 
         return w, h
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -112,7 +112,7 @@ cdef class _WindowSDL2Storage:
         
         # makes dpi aware of scale changes
         if platform == "win":
-            SDL_SetHint(SDL_HINT_WINDOWS_DPI_AWARENESS , b'permonitorv2')
+            SDL_SetHint(SDL_HINT_WINDOWS_DPI_SCALING, b"1")
 
         if SDL_Init(SDL_INIT_VIDEO | SDL_INIT_JOYSTICK) < 0:
             self.die()
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -109,6 +109,10 @@ cdef class _WindowSDL2Storage:
 
         SDL_SetHintWithPriority(b'SDL_ANDROID_TRAP_BACK_BUTTON', b'1',
                                 SDL_HINT_OVERRIDE)
+        
+        # makes dpi aware of scale changes
+        if platform == "win":
+            SDL_SetHint(SDL_HINT_WINDOWS_DPI_AWARENESS , b'permonitorv2')
 
         if SDL_Init(SDL_INIT_VIDEO | SDL_INIT_JOYSTICK) < 0:
             self.die()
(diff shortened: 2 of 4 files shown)


misl6:
FYI:

CI is failing for the same reason of #7982.

The issue is tracked on #7983

misl6:
@DexerBR 

Good news! https://github.com/kivy/kivy/pull/8096 has been merged.

Can you please rebase on top of `master` so the CI can run again?

(Please be aware of the latest changes, and check the docs before testing it on your local development machine)

DexerBR:
Hi @misl6!

The PR has been rebased!

misl6:
Thank you @DexerBR !

I'm looking forward to review the PR this weekend when I will have access to a Windows laptop.

misl6:
Hi @DexerBR !

My old Windows laptop is still upgrading ... , but meanwhile, I had a chance to dig into it a little bit deeper.

I guess we can now remove the following workaround: https://github.com/kivy/kivy/pull/7299 ?

DexerBR:
> Hi @DexerBR !
> 
> My old Windows laptop is still upgrading ... , but meanwhile, I had a chance to dig into it a little bit deeper.
> 
> I guess we can now remove the following workaround: #7299 ?

I think some workarounds can be removed but not all. The dpi change monitor still needs to exist, as it looks like the SDL still doesn't handle the dpi change messages (just the monitor change one for now, which can be used to update the dpi, just like you did here: #7982). I don't know how it behaves on other platforms, but after some testing, I think this workaround still needs to exist on Windows.

misl6:
> > Hi @DexerBR !
> > My old Windows laptop is still upgrading ... , but meanwhile, I had a chance to dig into it a little bit deeper.
> > I guess we can now remove the following workaround: #7299 ?
> 
> I think some workarounds can be removed but not all. The dpi change monitor still needs to exist, as it looks like the SDL still doesn't handle the dpi change messages (just the monitor change one for now, which can be used to update the dpi, just like you did here: #7982). I don't know how it behaves on other platforms, but after some testing, I think this workaround still needs to exist on Windows.

Yeah, I do not know if you have already seen the messages on Discord #dev channel. **quanon** proposed to keep the workarounds in order to avoid being too tied with SDL2, and I now agree.

Moving `_WindowsSysDPIWatch` outside of the sdl2-window specific code looks like a good thing, but there's no need to make that change in this PR.

misl6 (approved):
LGTM. Thank you!

I also had a chance to test it over 2 monitors with 2 different scale settings and works like a charm!

Screenshot 2023-03-04 172812

Screenshot 2023-03-04 172839

jotoft:
I think there might be something else with the application scale that needs to be considered when doing this.

It looks like the text works pretty good here but there is a difference in how the scale of other elements are on different monitors.

For instance the pictures demo application.

# With PR
## ~144 DPI
image

## ~240 DPI
image

# Without PR
## ~144 
image

## ~240 DPI
image

DexerBR:
Hi @jotoft! Thanks for testing!

Do any of these monitors have a scale factor greater than 100%?

If so, does the behavior you mentioned occur when moving the screen from one monitor to another, or under some other circumstance?

jotoft:
> Hi @jotoft! Thanks for testing!
> 
> Do any of these monitors have a scale factor greater than 100%?
> 
> If so, does the behavior you mentioned occur when moving the screen from one monitor to another, or under some other circumstance?
@DexerBR 

Hi, yes both monitors have a scale factor of over 100%.

(250% and 150%)

For me it is like this also from the beginning when launching `examples/demo/pictures/main.py`, moving the window from monitor does not cause any further issues.

In this application the widgets for the pictures are added dynamically with root.add_widget.
I think it would be preferable if the application would stay the same relative scale, maybe the sizes of the widgets are calculated in pixels or something (actually I am not 100% on what kivy uses as its internal units. i.e. what does a width of 100 mean?).

The reason I am bringing this up is that I was also considering making a similar change to Windows but I ran into this demo application issue (and then I saw that you already had a pull request).

DexerBR:
@jotoft I tested the pictures demo app with this PR using the scales 100% and 125%, and I couldn't reproduce the behavior you mentioned.

jotoft:
@dexerbr I tested your PR a little bit more right now and it seems that everything looks good except the scatter widget and switches, (in the showcase app) so I think maybe this should be a separate issue. 

Maybe the scale difference is not very noticeable at 125%?

jotoft:
```python

:class:`Scatter` is used to build interactive widgets that can be translated,
rotated and scaled with two or more fingers on a multitouch system.

Scatter has its own matrix transformation: the modelview matrix is changed
before the children are drawn and the previous matrix is restored when the
drawing is finished. That makes it possible to perform rotation, scaling and
translation over the entire children tree without changing any widget
properties. That specific behavior makes the scatter unique, but there are some
advantages / constraints that you should consider:
```

I think the problem lies with this widget maybe. I think the Scatter widgets needs to be re-scaled according to DPI/96

jotoft:
That fixes my issue with this, but I am not sure how to implement it properly. I think it could maybe be done in a follow up, probably it already behaves this way on OS X and Linux as well.

jotoft:
Yes, tested master branch on Linux now and it has the same problem, and I suspect it is the same with OS X. An issue should be raised.
To conclude, I approve of this pull request!

DexerBR:
@jotoft Thank you for investigating! If it's a problem with `Scatter` handling scale changes, I agree it's worth opening an issue to track.

misl6:
> @jotoft Thank you for investigating! If it's a problem with `Scatter` handling scale changes, I agree it's worth opening an issue to track.

Unfortunately not only the `Scatter` widget. And this is something we likely need to fix before **2.2.0** .
As an example, the switch widget looks completely broken on macOS with a retina display.

misl6 (approved):
Hopefully `SDL3` will allow us to handle it in a cleaner way, with a cross-platform event that signals DPI/scale changes.

But, meanwhile:

LGTM. Thank you!

Issue #8022, opened 2022-09-29
Horizontal Scroll Emulation for Desktop OS.


AliSot2000:
### The Problem I have:

I want a standard implementation which is easy to enable (like two lines of code) to emulate horizontal scrolling with a modifier key.

### My Solution:

I've forked this project and modified the [`kivy/inputs/providers/mouses.py`](https://github.com/AliSot2000/kivy) file to make this work. The code itself works. 

I have four issues which I would like to discuss with a maintainer of this project to get my feature fully integrated and ready for release with a new version.
1. I've not invested enough time to fully understand the configuration and how it is read etc. I've copied the initialization of other attributes for my version but I've not fully tested it with everything kivy has to offer. The fastest solution would be to have someone who's more familiar with the project have a quick look at my changes and give me feedback.
2. I can create my implementation with two or one attribute. The version with two attributes is more verbose will be slower. The other implementation is faster but relies on the assumption that `None in modifiers` will always evaluate to `False`.
3. I've updated the doc strings already in the `mouse.py` file but two other files need to have their docs updated. This point should be resolved in conjunction with Point 1.
4. I'm not very familiar with the development of the kivy project. A quick scroll through the branches didn't reveal a development branch which is used to merge new features into. This is why I'm opening an issue and not a pull request.

### What I've tried:

At first, I tried to create a callback at Window level to listen for events of the mouse which I could filter and then proceed to send via `self.dispatch_children(...)`. However this method failed for now since I was not able to create a correctly formatted `MouseMotionEvent`. And even with guessed values filled in for the all the positional attributes, it wasn't working correctly, mostly because I believe the stop_event was never triggered. I only dispatched the start event. (With more time and more reading through the code, I'm probably able to get this solution working but it's hacky and I don't really like it.)

Then I've modified the source code of kivy and managed to get it working flawlessly by modifying the `kivy/inputs/providers/mouses.py` however I would like this feature to be incorporated in the main kivy repo so I don't have to modify the source with every update.

### Context:

I'm mostly using a mouse with only a vertical scroll wheel. I have a mouse with a horizontal scroll wheel but I don't like that mouse. So I would like to have my kivy app emulate the horizontal scroll wheel with a modifier key. 

Additionally, this feature is quite useful since a lot of mice don't have a horizontal scroll wheel, and the scrolling via 'dragging' the Widgets around seems unprofessional and less precise to me. So I'd think, having a global setting with which to enable / disable this feature would be a desirable addition to the kivy project.

[...]

PR #8064, opened 2022-12-01
Horizontal Scroll with Modifier Key for Mouse

AliSot2000:
Hi 

I ran into the issue that I needed to scroll horizontally and I didn't feel like using the multi-touch emulation. Many applications have a modifier key designated to switch from a vertical scrolling direction to horizontal scrolling direction. 

I added said capability into kivy. (I've two implementations, one is commented out. One is faster and relies on default behavior, the other is slower but more verbose.)
Since this is my first time working with the project I'm not very familiar with both the way you do the docs and how you like to have the config (more verbose vs faster). 

I'd appreciate it if a maintainer could glance at the code I wrote and help me figure out how it should be structured to integrate as good as possible with the existing project.

AliSot2000

PS: I've written an [issue](https://github.com/kivy/kivy/issues/8022) already concerning this but I've had no response since so I decided to open this pull request. 

Maintainer merge checklist


RobertFlatt:
FYI https://github.com/Android-for-Python/gestures4kivy#pan

AliSot2000:
> FYI https://github.com/Android-for-Python/gestures4kivy#pan

I tried that first. 

I ran into issues with nested boxes which scroll. (I do not remember what they were since it’s been some time that I worked on the project)

AliSot2000:
@RobertFlatt 

Hi, I've gotten around to testing again. I've tried a lot of things but I was unable to use _gestures4kivy_ to implement nested ScrollViews. The outer ScrollView would always capture the event and not propagate it to the nested one, thereby making it useless. From my googling and understanding of events in kivy, this seems to be the default 'way' to handle it. 

I don't see me rewriting the _gestures4kivy_ too if I've already implemented a solution with this pull request.

For your convenience, I've an example of nested ScrollViews which you can use to verify the behaviour yourself.
[Example](https://www.reddit.com/r/kivy/comments/11a1z96/nested_scroll_views_no_longer_scrolling/)

RobertFlatt:
From the gestures4kivy docs:
> CommonGestures callback methods detect gestures; they do not implement behaviors.

I'm not really the right person to ask, I know only a little about the ScrollView widget, which is enough to know I don't understand. Since I'm not a kv jockey either, your example is a mystery to me.  So I tried something more basic.

I tried a simpler Python example, based on https://kivy.org/doc/stable/api-kivy.uix.scrollview.html and a model of a slot machine, imagine each number is a different  fruit🍎🍌🍍

```

def buttons():
    layout = GridLayout(cols=1, spacing=10, size_hint_y=None)
    # Make sure the height is such that there is something to scroll.
    layout.bind(minimum_height=layout.setter('height'))
    for i in range(20):
        btn = Button(text=str(i), size_hint_y=None, height=40)
        layout.add_widget(btn)
    return layout

def child_scroll():
    child = ScrollView(size_hint=(1, None), size=(Window.width, Window.height))
    child.add_widget(buttons())
    return child

layout2 = GridLayout(cols=3, spacing=10, size_hint_y=None)    
layout2.add_widget(child_scroll())
layout2.add_widget(child_scroll())
layout2.add_widget(child_scroll())

root = ScrollView(size_hint=(1, None), size=(Window.width, Window.height))
root.add_widget(layout2)

runTouchApp(root)
```

Each of the three columns scrolls independently.

I can see that one might *also* want to have the three columns move together (though a slot machine presumably does not allow this)

But I don't understand how to disambiguate one control signal (button, gesture, big handle on the side of the slot machine) to do two different functions (independent scroll, and synchronized scroll) in the same screen area. 

My lack of understanding is about UI design, not a about code, and unrelated to how the control signal is generated.

RobertFlatt:
I did a little more experimenting. As far as I can tell, and I may be wrong: 

- Touch events are not propagated to the parent ScrollView.
- Wheel Touch events are not propagated to the parent or to child ScrollView

Which I assume is what it says here https://kivy.org/doc/stable/api-kivy.uix.scrollview.html#scrolling-behavior

Without an event g4k is not going to respond.

AliSot2000:
So my application has a ScrollView which scrolls horizontally. This Scroll View contains another GridLayout which has itself multiple children which scroll too.

Using the default ScrollView works. The problem is that I cannot scroll horizontally bc I don't have a horizontal mouse button and I don't like to use the touch emulation.

I want to emulate the behavior you get when you comment all `on_scroll_start`, `on_scroll_stop`, `on_scroll_move`, `cgb_scroll`, `cgb_pan` in the example. 

Now I tried to implement that with the gestures4kivy like in said example. But you should discover that the scrolling is only sent to the parent and not the child ScrollViews. Compare that to the behavior I got before where on a horizontal scroll _first_ the inner ScrollView scrolled until it reached its limit, then the _outer_ ScrollView scrolls until your cursor is on top of another ScrollView. Then the _inner_ one scrolls again, until the limit is reached...

It is this behavior I'm trying to replicated with gestures4kivy.

RobertFlatt:
g4k gestures have the scope of the widget that inherits CommonGestures.

You can of course use the callbacks to call methods in child widgets if you want to. 

There is no additional magic behavior for any co-inherited Widget.

I really don't understand enough about interacting ScrollViews to comprehend the mechanism behind the behavior you describe.  

Have fun figuring it out.

Issue #5275, opened 2017-07-13
Switch to setuptools, remove distutils


tito:
Distutils is still the main library used for our packaging.

I would like to propose switching now by default on setuptools. The previous issues with distutils and Cython integration on old debian could be just now unsupported (i don't remember exactly why, but we may recover the why in the commit history / issues / ml.)

The reason is today, i installed Kivy using "pip -e ." from my Kivy directory. When uninstalling, pip said:
```
    DEPRECATION: Uninstalling a distutils installed project (Kivy) has been deprecated and will be removed in a future version. This is due to the fact that uninstalling a distutils project will only partially uninstall the project.
```

So it's time to fully move.

KeyWeeUsr:
Are we going to remove distutils completely later, or is it only about the default tool? Also, I'm not sure, but weren't there some problems with setuptools and android/p4a?

matham:
We need setuptools to build wheels. However, setuptools apparently doesn't
work on android.

misl6:
Fixed via #8070

PR #8067, opened 2022-12-03
Fix "ext_modules option must be an Extension instance" error

duelafn:
Fixes build error:

  distutils.errors.DistutilsSetupError: each element of 'ext_modules' option must be an Extension instance or 2-tuple

I get this error if building the latest master branch in Debian testing.

Apparently, this can occur if distutils is loaded before loading setuptools. See:

  https://stackoverflow.com/questions/21594925/error-each-element-of-ext-modules-option-must-be-an-extension-instance-or-2-t

Maintainer merge checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,7 @@ if "--build_examples" in sys.argv:
     build_examples = True
     sys.argv.remove("--build_examples")
 
+from setuptools import setup, Extension, find_packages
 from kivy.utils import pi_version
 from copy import deepcopy
 import os
@@ -22,7 +23,6 @@ from time import sleep
 from sysconfig import get_paths
 from pathlib import Path
 import logging
-from setuptools import setup, Extension, find_packages
 
 
 if sys.version_info[0] == 2:


misl6:
Hi @duelafn !

Well, we will need to migrate away from `distutils`, soon or later as is deprecated with removal planned for Python 3.12 😅

Is this issue happening with a specific version of `setuptools` ?

matham:
I do remember that at some point the order of imports as it currently is was deliberate. I'm not sure if whatever the reason was is still relevant.

duelafn:
We're using setuptools 65.5.0 with python 3.10 (and experimentally with python 3.11). I could do some experiments with moving to setuptools, but I don't have a way to test all your build environments (#5275 suggests that there was once issues on Android).

misl6:
🥇Thank you @duelafn for the report and the PR which helped us to force the switch to `setuptools`.

Unfortunately, I'm not able to reproduce your issue on both CI and local environments.

But:

I'm running some tests at https://github.com/kivy/kivy/pull/8070 (even on Android and iOS) and looks like we can get rid of some complexity and of `distutils` (which is deprecated).

`setuptools` docs says that `setuptools` should be now imported before any `distutils` import, if any of it is really required.

I guess that previously we needed to explicitly import `distutils` before `setuptools` as `setuptools` wasn't implementing a logic to fallback on `distutils` (the vendored one) when `Cython` is not available, for `build_ext`.

misl6:
Fixed via #8070

Issue #8080, opened 2022-12-23
Bug in displaying shadows when moving the widget


HeaTTheatR:
**Software Versions**
* Python: 3.10
* OS: Mac OS Catalina 10.15
* Kivy: master

**Describe the bug**
`BoxShadow` instructions are not displayed when moving the widget. If we move a widget that uses `BoxShadow` instructions by at least a pixel, the canvas will be cleared of `BoxShadow` instructions.  `BoxShadow` instructions will be restored only when the parent window is resized.

**Expected behavior**
It is expected that the shadow that is drawn on the canvas when the widget is moved will be displayed..

**Code**
```python
from kivy.animation import Animation
from kivy.app import App
from kivy.lang import Builder
from kivy.uix.button import Button

KV = """
#:import Clock kivy.clock.Clock


<ShadowButton>
    size_hint: None, None
    size: 200, 150

    canvas.before:
        Color:
            rgba: 0, 0, 1, .85
        BoxShadow:
            pos: self.pos
            size: self.size
            offset: 0, -10
            spread_radius: -20
            border_radius: 10, 10, 10, 10
            blur_radius: 80


Screen:

    ShadowButton:
        on_release: Clock.schedule_once(lambda x: app.move_button(self), 2)
        pos_hint: {"center_x": .5, "center_y": .5}
"""


class ShadowButton(Button):
    pass


class Example(App):
    def move_button(self, instance_button):
        Animation(
            pos_hint={"center_x": .3, "center_y": .3}, d=1.5
        ).start(instance_button)

    def build(self):
        return Builder.load_string(KV)


Example().run()
```

**What the bug looks like**

https://user-images.githubusercontent.com/16930280/209276814-183ba456-580e-4d98-9f3e-5bb109df7ae8.mov

Also, any manipulation of the `BoxShadow` instructions using the `Animation` class leads to a complete cleaning of the canvas instructions:

```python
from kivy.animation import Animation
from kivy.app import App
from kivy.lang import Builder
from kivy.uix.button import Button
from kivy.properties import NumericProperty

KV = """
#:import Clock kivy.clock.Clock


<ShadowButton>
    size_hint: None, None
    size: 200, 150

    canvas.before:
        Color:
            rgba: 0, 0, 1, .85
        BoxShadow:
            pos: self.pos
            size: self.size
            offset: 0, -10
            spread_radius: -20
            border_radius: 10, 10, 10, 10
            blur_radius: root.blur_radius


Screen:

    ShadowButton:
[...]

PR #8098, opened 2023-01-05
Fix `BoxShadow` shader crashing issue on Adreno GPUs

DexerBR:
As reported by @yunus-ceyhan in this [issue](https://github.com/kivy/python-for-android/issues/2723) . The shader used in the `BoxShadow` graphics instruction may crash on some Adreno GPUs.

The specific cause is still not very clear, however, there are two possibilities:
- Issue in Adreno GPU driver
- Issue in kivy's Shader implementation

The workaround is to check if the alpha channel of the `frag_color` variable is greater than zero. If so, a pure vector can be returned, otherwise, a vector created with texture2D needs to be returned, to avoid crashes.

### As a reference, for future implementations using shaders:

Code that crashes on Adreno GPUs:
```GLSL
void main(void)
{
    vec4 color = vec4(1.0, 0.0, 1.0, 1.0);
    gl_FragColor = color;
}
```

Code that compiles successfully:
```GLSL
void main(void)
{
    vec4 texture = texture2D(texture0, tex_coord0);
    gl_FragColor = texture;
}
```

Workaround:
```GLSL
void main(void)
{
    vec4 texture = texture2D(texture0, tex_coord0);
    vec4 color = vec4(1.0, 0.0, 1.0, 1.0);

     if (frag_color.a > 0.0) {
         gl_FragColor = color;
     } else {
         gl_FragColor = texture;
     }
}
```

<br>

⚠️ Update:

As found by @misl6 , there seems to be a bug in kivy that does not allow a shader initialization/compilation if the fragment shader is not passed as an argument to Fbo during its creation.
If the shader is passed as the initial argument, then crashes will not occur, even using the code below:

```GLSL
void main(void)
{
     vec4 color = vec4(1.0, 0.0, 1.0, 1.0);
     gl_FragColor = color;
}
```

🟢Taking this into account, the recommended for implementations using Fbo and shader is:
```python
fbo = Fbo(size=(width, height), fs=fs_code)
```

❌ Not recommended (may break on Adreno GPUs):
```python
fbo = Fbo(size=(width, height))
fbo.shader.fs = fs_code
```

Maintainer merge checklist

### DIFF
--- a/kivy/graphics/boxshadow.pyx
+++ b/kivy/graphics/boxshadow.pyx
@@ -149,7 +149,6 @@ cdef class BoxShadow(Fbo):
         self._init_texture()
 
     cdef void _init_texture(self):
-        self.shader.fs = SHADOW_fs
         self._rect = Rectangle(size=(100, 100))
         self._rect.texture = self.texture
         with self:
--- a/kivy/graphics/boxshadow.pyx
+++ b/kivy/graphics/boxshadow.pyx
@@ -89,19 +89,21 @@ void main (void){
 
 float distShadow = sigmoid(
     roundedBoxSDF(
-        tex_coord0 * size - size/2.0, size/2.0 - blur_radius * 1.5 - vec2(2.0),
-        border_radius) / (max(1.0, blur_radius)/4.0
-    )
+        tex_coord0 * size - size/2.0,
+        size/2.0 - blur_radius * 1.5 - vec2(2.0),
+        border_radius
+    ) / (max(1.0, blur_radius) / 4.0)
 );
-
-
-// Some devices require the resulting color to be blended with the texture.
-// Otherwise there will be a compilation issue.
 
 vec4 texture = texture2D(texture0, tex_coord0);
 vec4 shadow = vec4(frag_color.rgb, 1.0 - distShadow) * (frag_color.a * 2.0);
 
-gl_FragColor = mix(texture, shadow, 1.0);
+// This check is required to prevent shader crashing on some Adreno GPUs.
+if (frag_color.a > 0.0) {
+    gl_FragColor = shadow;
+} else {
+    gl_FragColor = texture;
+}
 
 }
 """
--- a/kivy/graphics/boxshadow.pyx
+++ b/kivy/graphics/boxshadow.pyx
@@ -56,7 +56,7 @@ from kivy.graphics.gl_instructions import ClearBuffers, ClearColor
 
 cdef str SHADOW_fs = """
 #ifdef GL_ES
-precision highp float;
+    precision highp float;
 #endif
 
 /* Outputs from the vertex shader */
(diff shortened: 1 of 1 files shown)


misl6:
Maybe this workaround (we have to discuss about the real cause) also fixes: https://github.com/kivy/kivy/issues/8080 ?
@HeaTTheatR can you perform some tests?

HeaTTheatR:
> Maybe this workaround (we have to discuss about the real cause) also fixes: #8080 ? @HeaTTheatR can you perform some tests?

These changes unfortunately do not fix the problem described in https://github.com/kivy/kivy/issues/8080

moonpyx:
Hello, some time ago I had that problem and the solution was very similar to use texture2D, but recently I made a small implementation which I tested in samsung s22 ultra that uses Adreno to define the final color of the fragment shader which worked for me and I share it with you hoping it will be helpful, here it is:

vec4 HackColor(vec4 color){
    vec4 c=color + (texture2D(texture0, tex_coord0)*vec4(vec3(0.0),1.0));
    color.rgb=c.rgb; 
    color.a = color.a*c.a;
    return color;
}
...
gl_FragColor = HackColor(color);

Peter-Port:
I hope this will be a part of "master",soon.
Thanks for your work, especially to:
T-Dynamos
HeaTTheatR
DexerBR
(found these names during my research)

RobertFlatt:
Seeing several users a day with shader related issues, workaround is always KivyMD 1.0.2 Be nice to see these addressed.
There seem to be two mechanisms
- Android crash discussed here
- Desktop BoxShadow related Shader did not link. e.g. https://groups.google.com/g/kivy-users/c/A1Asirs1I60/m/X7eZDj-rAgAJ  
So far nobody will post an issue for this one.....

misl6:
> Seeing several users a day with shader related issues, workaround is always KivyMD 1.0.2 Be nice to see these addressed. There seem to be two mechanisms
> 
> * Android crash discussed here
> * Desktop BoxShadow related Shader did not link. e.g. https://groups.google.com/g/kivy-users/c/A1Asirs1I60/m/X7eZDj-rAgAJ
>   So far nobody will post an issue for this one.....

Unfortunately, the Google Groups issue reporter only posted a screenshot with ~8 lines of logs, which is not enough to have an idea regarding the cause  😔

About Android and Adreno: I'm silently working (and testing on multiple Adreno chips) with @DexerBR to have a fix / workaround that doesn't look hacky and that fixes the root cause of the issue.

From our findings, looks like that `texture2D(texture0, tex_coord0);` returns a `vec4` full of `NaN` values on Adreno chips, at least during the early setup phase.

misl6:
@DexerBR , thank you for helping with tests and updating the PR!

I'm still investigating about the issue, as it seems we had similar tracebacks that comes from similar conditions (Eg: 3DRendering Example), so we will likely create a patch to fix this issue / documenting it

misl6 (approved):

misl6:
@HeaTTheatR , I'm waiting for the CI to complete, then when merged into master, can you make some tests on your side?

HeaTTheatR:
> @HeaTTheatR , I'm waiting for the CI to complete, then when merged into master, can you make some tests on your side?

Yes, of course!

PR #8109, opened 2023-01-22
Fix for changes of Splitter.strip_cls having no effect

CdrSonan:
The current implementation of `kivy.uix.Splitter.strip_cls` only causes `strip_cls` to be instantiated once, when the `_strip` attribute of the Splitter is not yet set. Therefore, any subsequent changes of `strip_cls` have no effect, even though `on_sizable_from()` is called correctly. This also includes setting `strip_cls` in a .kv file, in which case `on_sizable_from()` is called twice: Once with the default strip class, and a second time with the correct one.

As a fix, I am suggesting to remove the else statement preventing `strip_cls` from being instantiated several times. This also matches the code to the implementation of `dropdown_cls` of the Spinner widget, and its refresh mechanism in `kivy.uix.Spinner,_build_dropdown()`.

However, this change also causes re-instantiation of `strip_cls` when the `sizable_from` property is changed. This could theoretically cause loss of data stored in the old instance, but since `sizable_from` is usually only set once during UI creation, this is unlikely to be a problem in practice.
A possible solution for this issue would be to check whether `strip_cls` matches the class of the current `_strip` attribute, and only perform the re-instantiation if that is not the case.
If this is worth the additional complexity, I can update this pull request accordingly.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/splitter.py
+++ b/kivy/uix/splitter.py
@@ -202,9 +202,10 @@ class Splitter(BoxLayout):
             sup.remove_widget(instance._strip)
 
         cls = instance.strip_cls
-        if isinstance(cls, string_types):
-            cls = Factory.get(cls)
-        instance._strip = _strp = cls()
+        if not isinstance(_strp, cls):
+            if isinstance(cls, str):
+                cls = Factory.get(cls)
+            instance._strip = _strp = cls()
 
         sz_frm = instance.sizable_from[0]
         if sz_frm in ('l', 'r'):
--- a/kivy/uix/splitter.py
+++ b/kivy/uix/splitter.py
@@ -200,11 +200,11 @@ class Splitter(BoxLayout):
             self.unbind(disabled=_strp.setter('disabled'))
 
             sup.remove_widget(instance._strip)
-        else:
-            cls = instance.strip_cls
-            if isinstance(cls, string_types):
-                cls = Factory.get(cls)
-            instance._strip = _strp = cls()
+
+        cls = instance.strip_cls
+        if isinstance(cls, string_types):
+            cls = Factory.get(cls)
+        instance._strip = _strp = cls()
 
         sz_frm = instance.sizable_from[0]
         if sz_frm in ('l', 'r'):
--- a/kivy/uix/splitter.py
+++ b/kivy/uix/splitter.py
@@ -45,7 +45,6 @@ if self.horizontal else 'path to vertical pressed image'
 
 __all__ = ('Splitter', )
 
-from kivy.compat import string_types
 from kivy.factory import Factory
 from kivy.uix.button import Button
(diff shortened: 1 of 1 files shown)


misl6:
> However, this change also causes re-instantiation of strip_cls when the sizable_from property is changed. This could theoretically cause loss of data stored in the old instance, but since sizable_from is usually only set once during UI creation, this is unlikely to be a problem in practice.
A possible solution for this issue would be to check whether strip_cls matches the class of the current _strip attribute, and only perform the re-instantiation if that is not the case.
If this is worth the additional complexity, I can update this pull request accordingly.

Maybe we can:
- Separate the binding / unbinding / size change logic into a separate method
- Update `instance._strip` when `instance.strip_cls` changes + trigger the binding / unbinding logic after that
- Trigger the binding / unbinding logic when `sizable_from` changes

Also, FYI, here we have the opportunity to remove `from kivy.compat import string_types` 😀

Let me know if you want to enhance this PR or if you're fine with it.

CdrSonan:
Thanks for your reply!
I am going to implement what you said and then update the pull request.

misl6:
> Thanks for your reply!
> I am going to implement what you said and then update the pull request.

Love it. Feel free to ping me by requesting a review when you're ready!

CdrSonan:
@misl6 
I've finally had time to get back to this.
Moving the reinstancing logic to a new function hasn't proven practical. Most parts of `on_sizable_from()` would also be necessary in the hypothetical `on_strip_cls()`, so there would either be duplicate code between the two, or additional utility functions for shared functionality. (Mainly various callback binds/unbinds)

So I have instead replaced the original else statement with an `isinstance()` call that checks whether `instance._strip` is indeed an instance of `strip_cls`. If it isn't, the strip is reinstanced, otherwise the reinstancing code is skipped. Since `None` is not an instance of `strip_cls`, this also means reinstancing is triggered if the strip hasn't been instanced yet,
Therefore, everything works as it's supposed to. Reinstancing is triggered when the widget is created or when `strip_cls` changes, but skipped when only `sizable_from` changes.
I have tested this implementation in both the example Splitter App and my own code using custom strip classes, and encountered no issues.

I have also removed the `string_types` import and replaced it with just `str`. I am not sure if this has the desired effect since it doesn't offer any additional backwards compatibility, but it is equivalent to the original code, given the current implementation of `string_types`.

misl6 (approved):
LGTM. Thank you!

misl6:
> I have also removed the string_types import and replaced it with just str. I am not sure if this has the desired effect since it doesn't offer any additional backwards compatibility, but it is equivalent to the original code, given the current implementation of string_types.

We still have a few `string_types` around from the Python 2 era, but we removed Python 2 compatibility a long time ago. We take the opportunity to clean up the code when possible 😀

PR #8115, opened 2023-01-29
Use `font_direction` and `font_script_name` from SDL2_ttf

akshayaurora:
This is intentionally kept separate from `base_direction` and `text_language` which are Pango specific.

Maintainer merge checklist


misl6:
@akshayaurora can you rebase on top of master since we now have support for `TTF_SetFontDirection` and `TTF_SetFontScriptName` ?

And... How do you feel about merging it? (unless something breaks during the CI runs)

Since we need testing from users from all over the globe and some users are not confident enough to build Kivy from source code, it would be helpful to suggest users install it via night wheels in order to test the feature.

PR #8123, opened 2023-02-06
initialize our logger inside of kivy/__init__.py

craigkai:
The logger can end up not being `_configured` if this is not called. This can occure when log level is set to a stricter level like `error` and no errors occur, then when we call the purge_logs function our handler could be un-configured and not have a log_dir attribute -- resulting in the function bailing out.

Maintainer merge checklist

### DIFF
--- a/kivy/__init__.py
+++ b/kivy/__init__.py
@@ -380,6 +380,16 @@ if not environ.get('KIVY_DOC_INCLUDE'):
     level = LOG_LEVELS.get(Config.get('kivy', 'log_level'))
     Logger.setLevel(level=level)
 
+    # Initialize our logger, if we do not log anything during this
+    # init sequence then the logger will not be `_configured`. This
+    # can occure when log level is set to a stricter level like `error`
+    # and no errors occur, then when we call the purge_logs function our
+    # handler could be un-configured and not have a log_dir attribute --
+    # resulting in the function bailing out.
+    from kivy.logger import file_log_handler
+    file_log_handler._configure()
+
+
     # Can be overridden in command line
     if ('KIVY_UNITTEST' not in environ and
             'KIVY_PACKAGING' not in environ and
--- a/kivy/__init__.py
+++ b/kivy/__init__.py
@@ -387,8 +387,8 @@ if not environ.get('KIVY_DOC_INCLUDE'):
     # handler could be un-configured and not have a log_dir attribute --
     # resulting in the function bailing out.
     from kivy.logger import file_log_handler
-    file_log_handler._configure()
-
+    if file_log_handler is not None:
+        file_log_handler._configure()
 
     # Can be overridden in command line
     if ('KIVY_UNITTEST' not in environ and


craigkai:
If this is the wrong place to configure the logger, we can also check inside the purge_logs function and call `_configure` if `log_dir` is not defined instead of just bailing.
```
def purge_logs(self):
        """Purge logs which exceed the maximum amount of log files,
        starting with the oldest creation timestamp (or edit-timestamp on Linux)
        """

        if not self.log_dir:
            return
```

kivy/__init__.py
```diff
@@ -380,6 +380,16 @@ def _patch_mod_deps_win(dep_mod, mod_name):
...
     Logger.setLevel(level=level)
 
+    # Initialize our logger, if we do not log anything during this
+    # init sequence then the logger will not be `_configured`. This
+    # can occure when log level is set to a stricter level like `error`
+    # and no errors occur, then when we call the purge_logs function our
+    # handler could be un-configured and not have a log_dir attribute --
+    # resulting in the function bailing out.
+    from kivy.logger import file_log_handler
+    file_log_handler._configure()
```
misl6, line 390: `file_log_handler` may be `None`, so we should check if is not, before calling `_configure`, as we do just before the call of `purge_logs`.

misl6 (changes requested):
Hi @craigkai !

I've left a comment regarding the proposed changes.

BTW, that looks quite hacky, and as you said is better to move this logic into the `FileHandler` class.
I think that we should ensure that all the configurations are made at initialization, and not during usage.

kivy/__init__.py
craigkai: I am not sure if it can be done exactly on initialization since it looks like importing config creates a circular import error:

```python
  File "/Users/craigkaiser/kaiserengineering/digitaldash/main.py", line 20, in <module>
    from kivy.logger import Logger, LOG_LEVELS
  File "/Users/craigkaiser/.local/share/virtualenvs/digitaldash-iuTs6a7o/lib/python3.11/site-packages/kivy/__init__.py", line 40, in <module>
    from kivy.logger import Logger, LOG_LEVELS
  File "/Users/craigkaiser/.local/share/virtualenvs/digitaldash-iuTs6a7o/lib/python3.11/site-packages/kivy/logger.py", line 634, in <module>
    add_kivy_handlers(logging.root)
  File "/Users/craigkaiser/.local/share/virtualenvs/digitaldash-iuTs6a7o/lib/python3.11/site-packages/kivy/logger.py", line 598, in add_kivy_handlers
  File "/Users/craigkaiser/.local/share/virtualenvs/digitaldash-iuTs6a7o/lib/python3.11/site-packages/kivy/config.py", line 387, in <module>
    from kivy import kivy_config_fn
ImportError: cannot import name 'kivy_config_fn' from partially initialized module 'kivy' (most likely due to a circular import) (/Users/craigkaiser/.local/share/virtualenvs/digitaldash-iuTs6a7o/lib/python3.11/site-packages/kivy/__init__.py)
```

PR #8131, opened 2023-02-13
Initialize logger if no log_dir is configured during log purge

craigkai:
The logger can end up not being `_configured` if this is not called. This can occure when log level is set to a stricter level like `error` and no errors occur, then when we call the purge_logs function our handler could be un-configured and not have a log_dir attribute -- resulting in the function bailing out.

Maintainer merge checklist

### DIFF
--- a/kivy/logger.py
+++ b/kivy/logger.py
@@ -218,7 +218,10 @@ class FileHandler(logging.Handler):
         """
 
         if not self.log_dir:
-            return
+            # Try configuring our file handler if none is configured
+            self._configure()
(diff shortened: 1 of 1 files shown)


misl6:
Hi @craigkai !

Sorry for the late reply, but we were really busy with the new release process.

How do you feel about adding a test that exercises this issue so we can ensure that does not happen again in the future?

craigkai:
I'll see what I can do!

misl6:
Ping @craigkai !

PR #8132, opened 2023-02-14
`BoxShadow`: Add `inset` feature

DexerBR:
### Adds `inset` feature, to allow drawing outside or inside boundaries.

boxshadow_inset

Maintainer merge checklist


kivy/graphics/boxshadow.pyx
```diff
@@ -87,16 +91,36 @@ float sigmoid(float x) {
...
+            ) - adjustmentFactor
+        ) / (max(1.0, blur_radius) / 4.0)
+    );
+
+    if (inset == 1){
+        gl_FragColor = vec4(1.0, 1.0, 1.0, distShadow);
+    }
+    else{
+        vec4 shadow = vec4(1.0, 1.0, 1.0, (1.0 - distShadow) * (frag_color.a * 2.0));
+        gl_FragColor = shadow;
```
misl6, line 122: I guess (after the debugging session we had last month) that we can now avoid to assign the `vec4` to `shadow`, before assigning it to `gl_FragColor`?

misl6: PS: I need to test it (deeply), but at a first glance the code looks good.
Expect a review in ~2 days.

DexerBR: > I guess (after the debugging session we had last month) that we can now avoid to assign the `vec4` to `shadow`, before assigning it to `gl_FragColor`?

Correct! It also simplifies the logic.

misl6:
Hi @DexerBR !

I was playing with the feature and the result is not coherent with my expectations. (And we have the same bug on `master`, so feel free to target it on a different PR).

Isn't the `BoxShadow` effect expected to been draw only outside of the `BoxShadow` rectangle for `inset: False` and inside of the `BoxShadow` rectangle for `inset: True` ?

With the following code:

```kv
Screen:
    canvas:
        Color:
            rgba: 0.5, 0.5, 0.5, 1.0
        Rectangle:
            pos: self.pos
            size: self.size

    BoxLayout:
        canvas.before:
            Color:
                rgba: 1,1,1,1
            Rectangle:
                pos: self.pos
                size: self.size
        canvas.after:
            Color:
                rgba: 0, 0, 1, 0.85
            BoxShadow:
                pos: self.pos
                size: self.size
                offset: 0, 0
                spread_radius: 0
                border_radius: 0, 0, 0, 0
                blur_radius: 80
                inset: False
        pos_hint: {"center_x": .5, "center_y": .75}
        size_hint: None, None
        size: 200, 150
    BoxLayout:
        canvas.before:
            Color:
                rgba: 1,1,1,1
            Rectangle:
                pos: self.pos
                size: self.size
        canvas.after:
            Color:
                rgba: 0, 0, 1, 0.85
            BoxShadow:
                pos: self.pos
                size: self.size
                offset: 0, 0
                spread_radius: 0
                border_radius: 0, 0, 0, 0
                blur_radius: 80
                inset: True
        pos_hint: {"center_x": .5, "center_y": .25}
        size_hint: None, None
        size: 200, 150
```

I was expecting something like that:

Image 18-02-23 at 14 35

Instead, I'm getting the following:

Image 18-02-23 at 14 29


Has that been done on purpose or we should consider it a bug?

DexerBR:
Acutally, this was done on purpose, because, like other graphical instructions, `BoxShadow` is not aware of the hierarchy of declarations made in the canvas object.

Therefore, it is up to the user to define in which layer a canvas instruction must be written. For example, if users want an insert effect, they needs to write the `BoxShadow` above the target canvas instruction, using `canvas`,  `canvas.after` or simply arranging the hierarchy of declarations, so the `BoxShadow` is above.

If the default mode is used (`inset=False`), then it should do the opposite.

misl6:
> Acutally, this was done on purpose, because, like other graphical instructions, `BoxShadow` is not aware of the hierarchy of declarations made in the `canvas` object.
> 
> Therefore, it is up to the user to define in which layer a canvas instruction must be written. For example, if users want an insert effect, they needs to write the `BoxShadow` above the target canvas instruction, using `canvas`, `canvas.after` or simply arranging the hierarchy of declarations, so the `BoxShadow` is above.
> 
> If the default mode is used (`inset=False`), then they should do the opposite.

Yeah, absolutely, `BoxShadow` is not aware of the hierarchy, and should not.

But, if we're comparing `BoxShadow (Kivy)` with `box-shadow (CSS)`, the result is not coherent and the user might not expect that `BoxShadow` with `inset: False` actually also "paints" something inside of the rectangle.

Maybe (in another PR), we can at least document this behavior?

misl6 (approved):
LGTM. Thank you!

But, please, consider documenting / changing the behaviour of what discussed during the review. (Which is unrelated to this PR)

DexerBR:
Thank you!
I think it's worth explaining this behavior more clearly in the documentation. I will do it soon.

DexerBR:
> > Acutally, this was done on purpose, because, like other graphical instructions, `BoxShadow` is not aware of the hierarchy of declarations made in the `canvas` object.
> > Therefore, it is up to the user to define in which layer a canvas instruction must be written. For example, if users want an insert effect, they needs to write the `BoxShadow` above the target canvas instruction, using `canvas`, `canvas.after` or simply arranging the hierarchy of declarations, so the `BoxShadow` is above.
> > If the default mode is used (`inset=False`), then they should do the opposite.
> 
> Yeah, absolutely, `BoxShadow` is not aware of the hierarchy, and should not.
> 
> But, if we're comparing `BoxShadow (Kivy)` with `box-shadow (CSS)`, the result is not coherent and the user might not expect that `BoxShadow` with `inset: False` actually also "paints" something inside of the rectangle.
> 
> Maybe (in another PR), we can at least document this behavior?

Hi @misl6 !

I added additional documentation about `BoxShadow` behavior in this PR:
- https://github.com/kivy/kivy/pull/8139

PR #8142, opened 2023-02-22
New Feature: Allows to hide the taskbar icon

DexerBR:
This PR introduces the possibility to hide the taskbar application icon.

Options:
- `0` → The icon will be hidden in the taskbar.
- `1` → default - The icon will be shown on the taskbar.

<br>

Test code:

```python
Config.set("graphics", "show_taskbar_icon", 0)

from kivy.app import App


class TestApp(App):

    def build(self):
        return Button(text="Hello World")

if __name__ == "__main__":
    TestApp().run()
```

Maintainer merge checklist

### DIFF
--- a/kivy/config.py
+++ b/kivy/config.py
@@ -405,7 +405,7 @@ from kivy.utils import platform
 _is_rpi = exists('/opt/vc/include/bcm_host.h')
 
 # Version number of current configuration format
-KIVY_CONFIG_VERSION = 26
+KIVY_CONFIG_VERSION = 27
 
 Config = None
 '''The default Kivy configuration object. This is a :class:`ConfigParser`
@@ -938,6 +938,8 @@ if not environ.get('KIVY_DOC_INCLUDE'):
 
         elif version == 25:
             Config.setdefault('graphics', 'always_on_top', '0')
+
+        elif version == 26:
             Config.setdefault("graphics", "show_taskbar_icon", "1")
 
         # WARNING: When adding a new version migration here,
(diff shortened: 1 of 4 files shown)


kivy/config.py
```diff
@@ -932,6 +938,7 @@ def name(self, value):
 
         elif version == 25:
             Config.setdefault('graphics', 'always_on_top', '0')
+            Config.setdefault("graphics", "show_taskbar_icon", "1")
```
misl6, line 941: Better to require another config migration, even if we bumped to `version = 26` just yesterday.

DexerBR: Ok, adressed!

kivy/core/window/window_sdl2.py
```diff
@@ -311,11 +311,15 @@ def create_window(self, *largs):
             # setup window
             w, h = self.system_size
             resizable = Config.getboolean('graphics', 'resizable')
+            show_taskbar_icon = Config.getboolean(
+                'graphics',
+                'show_taskbar_icon'
+            )
```
misl6, line 317: What about moving the logic into `setup_window` method  (Like we're doing for `multisamples = Config.getint('graphics', 'multisamples')`) ?

This will allow us to not add an additional argument to the `setup_window` method.

misl6 (changes requested):

kivy/config.py
```diff
@@ -253,6 +253,11 @@
         When enabled, the window will be brought to the front and will keep
         the window above the rest. Only works for the sdl2 window provider.
         ``0`` is disabled, ``1`` is enabled.
+    `show_taskbar_icon`: int, one of ``0`` or ``1``, defaults to ``0``
```
misl6, line 256:
```suggestion
    `show_taskbar_icon`: int, one of ``0`` or ``1``, defaults to ``1``
```

misl6 (changes requested):
Hi @DexerBR !

Seems there's a typo. I've left a suggestion block.

misl6 (approved):
LGTM. Thank you!

PR #8153, opened 2023-03-06
Use correct Path for Kivy Home Dir

JakobDev:
At the moment, Kivy saves all of it's data under `~/.kivy`, but this is wrong. Each Operating System has it's own Location to save such files. This PR make Kivy respect the Standard for each OS.

This is especially important for Linux, if you run a Kivy App in a Sandboxed Environment (e.g. Flatpak or Snap) which expects the Apps to save it's Data in a custom Directory (e.g. Flat Flatpak it's `~/.var/<appid>`) and restrict access to Home.

For macOS I#m not 100% sure, if I got the right Location, because I never used it, but this is the Location I found that are used by other Programs as well.

Maintainer merge checklist

### DIFF
--- a/doc/sources/guide/config.rst
+++ b/doc/sources/guide/config.rst
@@ -14,15 +14,29 @@ environment variable `KIVY_HOME`::
 
     <KIVY_HOME>/config.ini
 
-On desktop, this defaults to::
+On Windows, this defaults to::
 
-    <HOME_DIRECTORY>/.kivy/config.ini
+    %APPDATA%/kivy/config.ini
 
 Therefore, if your user is named "tito", the file will be here:
 
-- Windows: ``C:\Users\tito\.kivy\config.ini``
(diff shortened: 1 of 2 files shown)


pythonsus (approved):

leycec:
This is a substantial improvement to Kivy's current home directory handling. It'd be great if this could be merged at some point. Thanks for the tremendous volunteerism, @JakobDev! :hugs:

misl6:
@JakobDev , sorry for the late reply, but as you might have noticed, we were into feature-freeze for `2.2.0` (which is now released).

Some thoughts about this proposal:
- Should we need to offer an automated migration path? (We're currently targeting `2.3.0` as the next release, even if some major changes may require a `3.0.0`, so I'm not sure about the next release version yet)
- How about using https://platformdirs.readthedocs.io? (we can open a PR on `platformdirs` repo to support iOS if needed) + is a plain python package and will not need a recipe for Android or iOS

PR #8159, opened 2023-03-08
Added API preference to VideoCapture

DarkSparkAg:
Added a preferred capture API backend to the cv2 VideoCapture function call to greatly reduce lag when creating a kivy camera object in Windows.

Maintainer merge checklist

### DIFF
--- a/kivy/core/camera/camera_opencv.py
+++ b/kivy/core/camera/camera_opencv.py
@@ -106,7 +106,7 @@ class CameraOpenCV(CameraBase):
 
         elif self.opencvMajorVersion in (2, 3, 4):
             # create the device
-            self._device = cv2.VideoCapture(self._index)
+            self._device = cv2.VideoCapture(self._index, cv2.CAP_DSHOW)
             # Set preferred resolution
             self._device.set(PROPERTY_WIDTH,
                              self.resolution[0])


kivy/core/camera/camera_opencv.py
```diff
@@ -106,7 +106,7 @@ def init_camera(self):
 
         elif self.opencvMajorVersion in (2, 3, 4):
             # create the device
-            self._device = cv2.VideoCapture(self._index)
+            self._device = cv2.VideoCapture(self._index, cv2.CAP_DSHOW)
```
misl6, line 109: `cv2.CAP_DSHOW` is Windows only, and also on Windows the user may want to start the capture with another API.

Maybe you can introduce a logic that allows the user to select a specific API?

misl6 (changes requested):

kivy/core/camera/camera_opencv.py
DarkSparkAg: Well that took longer than I anticipated, but it's done.  Let me know if you have any other feedback.  I don't have a lot of experience with public projects on GitHub so suggestions are welcome.

kivy/uix/camera.py
```diff
@@ -78,6 +78,14 @@ class Camera(Image):
     :attr:`resolution` is a :class:`~kivy.properties.ListProperty` and defaults
     to [-1, -1].
     '''
+    
+    apiID = NumericProperty(0)
+    '''Preferred video capture API ID to use when invoking the camera. If you are using
+    0, the API will be the default one::
+
+    :attr:`apiID` is a :class:`~kivy.properties.NumericProperty` and defaults
+    to 0.
```
misl6, line 87: Hi @DarkSparkAg !

Here we need to add a `.. versionadded` snippet, so users can immediately know when this feature has been added.

kivy/uix/camera.py
```diff
@@ -78,6 +78,14 @@ class Camera(Image):
     :attr:`resolution` is a :class:`~kivy.properties.ListProperty` and defaults
     to [-1, -1].
     '''
+    
+    apiID = NumericProperty(0)
```
misl6, line 82: Maybe `apiID` should be `api_id` in order to follow the snake case convention?

kivy/uix/camera.py
```diff
@@ -78,6 +78,14 @@ class Camera(Image):
...
+    '''Preferred video capture API ID to use when invoking the camera. If you are using
+    0, the API will be the default one::
```
misl6, line 84: Should we mention that only specific core providers are supporting that?
How about using a `StringProperty`with default to `default` instead?

misl6 (changes requested):
Hi @DarkSparkAg !

I've added some comments regarding your PR. Feel free to share your thoughts about it.

kivy/uix/camera.py
DarkSparkAg: I'm afraid I don't understand OpenCV well enough to understand your question.  Is it connected to different API's only being available on specific platforms?

I believe a NumericProperty would be better as the cv2.VideoCapture() function expects an integer for the api Preference.  And that way users can simply use cv.CAP_DSHOW, cv.CAP_MSMF, cv.CAP_V4L2, etc when attempting to call out a specific api.  If you still believe a StringProperty would be better, I defer to your judgement.  However, I'm not sure how to properly implement the conversion of a StringProperty back to a integer for cv2.VideoCapture().

DarkSparkAg: Thank you for the explanation.  I have attempted to make the changes that you requested.

block0xhash:
This issue here https://github.com/kivy/kivy/issues/8243 was flagged to be fixed here, is this still the case. It seems the issue still exists on Kivy==2.3.1. Applying the changed suggested here https://github.com/kivy/kivy/issues/8243#issue-1708725494 cuts down the start time from +-30secs to +-5secs  this is on Windows 11` pip freeze
certifi==2025.4.26
charset-normalizer==3.4.2
docutils==0.21.2
ffpyplayer==4.5.2
filetype==1.2.0
idna==3.10
Kivy==2.3.1
Kivy-examples==2.3.1
Kivy-Garden==0.1.5
kivy_deps.angle==0.4.0
kivy_deps.glew==0.3.1
kivy_deps.gstreamer==0.3.4
kivy_deps.sdl2==0.8.0
numpy==2.2.5
opencv-python==4.11.0.86
pillow==10.4.0
Pygments==2.19.1
pypiwin32==223
pywin32==310
requests==2.32.3
urllib3==2.4.0
(kivy_venv)
`

PR #8162, opened 2023-03-11
`Label`: allow different values of left, top, right and bottom for `padding`.

DexerBR:
Previously, `padding `only accepted 2 values: `[padding_horizontal, padding_vertical]`, now it has the same behavior as `padding` in `BoxLayout`/`GridLayout`:

- `[10]` → `[10, 10, 10, 10]`
- `[10, 20]` → `[10, 20, 10, 20]`
- `[10, 20, 30, 40]` → `[10, 20, 30, 40]`

Test code:
```python
from kivy.app import App
from kivy.lang import Builder


class UI(FloatLayout):
    pass


Builder.load_string(
    """
<UI>:
    padding_left: 0
    padding_top: 0
    padding_right: 0
    padding_bottom: 0
    Label:
        text: "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
        size_hint: None, None
        center: root.center
        size: 300, 300

        text_size: self.size
        # size: self.texture_size

        # markup: True

        # base_direction: "rtl"

        halign: 'center'
        valign: 'center'

        padding: root.padding_left, root.padding_top, root.padding_right, root.padding_bottom

        canvas:
            Color:
                rgba: 0.2, 0.2, 0.8, 0.5
            Rectangle:
                pos: self.pos
                size: self.size

            Color:
                rgba: 1, 0, 0, 0.5
            Rectangle:
                pos: self.pos
                size: root.padding_left, self.height

            Rectangle:
                pos: self.x, self.top - root.padding_top
                size: self.width, root.padding_top

            Rectangle:
                pos: self.right - root.padding_right, self.y
                size: root.padding_right, self.height

            Rectangle:
                pos: self.pos
                size: self.width, root.padding_bottom

    BoxLayout:
        size_hint_y: 0.4
        orientation: "vertical"

        Slider:
            size_hint_y: None
            height: 50
            min: 0
            max: 400
            step: 1
            on_value:
                root.padding_left = self.value
        Slider:
            size_hint_y: None
            height: 50
            min: 0
            max: 400
            step: 1
            on_value:
                root.padding_top = self.value
        Slider:
            size_hint_y: None
            height: 50
            min: 0
            max: 400
            step: 1
            on_value:
                root.padding_right = self.value
        Slider:
            size_hint_y: None
            height: 50
            min: 0
            max: 400
            step: 1
            on_value:
                root.padding_bottom = self.value

"""
)


class Testapp(App):

    def build(self):
        return UI()


Testapp().run()
```

Maintainer merge checklist

### DIFF
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -721,10 +721,11 @@ class LabelBase(object):
         valign = options['valign']
 
         y = padding_top = options['padding'][1]  # pos in the texture
+        padding_bottom = options['padding'][3]
         if valign == 'bottom':
             y = size[1] - ih + padding_top
         elif valign == 'middle' or valign == 'center':
-            y = int((size[1] - ih) / 2 + padding_top)
+            y = int((size[1] - ih) / 2 + (padding_top + padding_bottom) / 2)
         self._render_begin()
         self.render_lines(lines, options, self._render_text, y, size)
 
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -302,7 +302,7 @@ class LabelBase(object):
         for padding_option in ('padding_x', 'padding_y'):
             if kwargs_get(padding_option):
                 Logger.warning(
-                    f"The use of the {padding_option} parameter is "
+                    f"LabelBase: The use of the {padding_option} parameter is "
                     "deprecated, and will be removed in future versions. Use "
                     "padding instead."
                 )
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -88,6 +88,7 @@ from kivy.core.text.text_layout import layout_text, LayoutWord
 from kivy.resources import resource_find, resource_add_path
 from kivy.compat import PY2
 from kivy.setupconfig import USE_SDL2, USE_PANGOFT2
+from kivy.logger import Logger
 
 
 if 'KIVY_DOC' not in os.environ:
@@ -274,7 +275,34 @@ class LabelBase(object):
         kwargs_get = kwargs.get
         options['color'] = color or (1, 1, 1, 1)
         options['outline_color'] = outline_color or (0, 0, 0, 1)
-        options['padding'] = kwargs_get('padding', (0, 0, 0, 0))
+
+        options['padding'] = kwargs_get('padding', [0, 0, 0, 0])
+        if isinstance(options['padding'], (int, float)):
+            options['padding'] = [options['padding']] * 4
+        elif (
+            isinstance(options['padding'], (list, tuple))
+            and len(options['padding']) != 4
+        ):
+            if len(options['padding']) == 1:
+                options['padding'] = options['padding'] * 4
+            elif len(options['padding']) == 2:
+                options['padding'] = options['padding'] * 2
+            else:
+                raise ValueError(
+                    "padding should be int|float or a list|tuple with 1, 2 or "
+                    f"4 elements, got {type(options['padding'])} with "
+                    f"{len(options['padding'])} elements."
+                )
+
+        options['padding_x'] = kwargs_get('padding_x')
+        options['padding_y'] = kwargs_get('padding_y')
+        for padding_option in ('padding_x', 'padding_y'):
+            if kwargs_get(padding_option):
+                Logger.warning(
+                    f"The use of the {padding_option} parameter is "
+                    "deprecated, and will be removed in future versions. Use "
+                    "padding instead."
+                )
 
         if 'size' in kwargs:
             options['text_size'] = kwargs['size']
@@ -293,6 +321,15 @@ class LabelBase(object):
         self.texture = None
         self.is_shortened = False
         self.resolve_font_name()
+        self._update_padding()
+
+    def _update_padding(self):
+        options = self.options
+        self.options['padding'] = list(self.options['padding'])
+        if options['padding_x']:
+            self.options['padding'][::2] = [options['padding_x']] * 2
+        if options['padding_y']:
+            self.options['padding'][1::2] = [options['padding_y']] * 2
 
     @staticmethod
     def register(name, fn_regular, fn_italic=None, fn_bold=None,
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -640,9 +640,10 @@ class LabelBase(object):
     def render_lines(self, lines, options, render_text, y, size):
         get_extents = self.get_cached_extents()
         uw, uh = options['text_size']
-        xpad = options['padding'][0]
+        padding_left = options['padding'][0]
+        padding_right = options['padding'][2]
         if uw is not None:
-            uww = uw - 2 * xpad  # real width of just text
+            uww = uw - padding_left - padding_right  # real width of just text
         w = size[0]
         sw = options['space_width']
         halign = options['halign']
@@ -659,14 +660,15 @@ class LabelBase(object):
                 line = last_word.text
                 if not cur_base_dir:
                     cur_base_dir = find_base_dir(line)
-            x = xpad
+            x = padding_left
             if halign == 'auto':
                 if cur_base_dir and 'rtl' in cur_base_dir:
-                    x = max(0, int(w - lw - xpad))  # right-align RTL text
+                    # right-align RTL text
+                    x = max(0, int(w - lw - padding_right))
             elif halign == 'center':
                 x = int((w - lw) / 2.)
             elif halign == 'right':
-                x = max(0, int(w - lw - xpad))
+                x = max(0, int(w - lw - padding_right))
 
             # right left justify
             # divide left over space between `spaces`
@@ -718,12 +720,11 @@ class LabelBase(object):
         size = self.size
         valign = options['valign']
 
-        y = ypad = options['padding'][1]  # pos in the texture
+        y = padding_top = options['padding'][1]  # pos in the texture
         if valign == 'bottom':
-            y = size[1] - ih + ypad
+            y = size[1] - ih + padding_top
         elif valign == 'middle' or valign == 'center':
-            y = int((size[1] - ih) / 2 + ypad)
-
+            y = int((size[1] - ih) / 2 + padding_top)
         self._render_begin()
         self.render_lines(lines, options, self._render_text, y, size)
 
--- a/kivy/core/text/__init__.py
+++ b/kivy/core/text/__init__.py
@@ -669,7 +669,7 @@ class LabelBase(object):
                 x = min(
                     int(w - lw),
                     max(
-                        padding_left,
+                        int(padding_left),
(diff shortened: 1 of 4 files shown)


kivy/uix/label.py
```diff
@@ -707,13 +710,23 @@ def on_ref_press(self, ref):
     .. versionchanged:: 1.9.0
         `padding_y` has been fixed to work as expected.
         In the past, the text was padded by the negative of its values.
+
+    .. deprecated:: 2.2.0
+        Please use :attr:`padding` instead.
     '''
```
misl6, line 716: I may agree with the deprecation, but we can't remove the support for `padding_x` and `padding_y`, without giving the users a quite large deprecation window.

Is fine to mark `padding_x` and `padding_y` as deprecated, but we should keep the functionality for at least ~3/4 releases, so users have time to migrate their code at their own pace.

misl6 (changes requested):
I like the feature, a lot.

But, unfortunately, we should keep the `padding_x` and `padding_y` properties for a long time, and these properties can't be removed without having them deprecated first.

kivy/uix/label.py
DexerBR: Understood! Sounds like a good approach!

kivy/core/text/__init__.py
```diff
@@ -263,11 +271,7 @@ def __init__(
         kwargs_get = kwargs.get
         options['color'] = color or (1, 1, 1, 1)
         options['outline_color'] = outline_color or (0, 0, 0, 1)
-        options['padding'] = kwargs_get('padding', (0, 0))
-        if not isinstance(options['padding'], (list, tuple)):
-            options['padding'] = (options['padding'], options['padding'])
-        options['padding_x'] = kwargs_get('padding_x', options['padding'][0])
-        options['padding_y'] = kwargs_get('padding_y', options['padding'][1])
+        options['padding'] = kwargs_get('padding', (0, 0, 0, 0))
```
misl6, line 274: `LabelBase` should be considered as part of the public API, and as `Label`, support for `padding_x` and `padding_y` needs to be deprecated, but not removed.

A migration path should also be applied to `padding`. If a user specifies an integer, instead of the new format, we need to print a deprecation warning and then convert the integer to the new format.

kivy/uix/label.py
```diff
@@ -303,8 +303,9 @@ class Label(Widget):
 
     _font_properties = ('text', 'font_size', 'font_name', 'bold', 'italic',
                         'underline', 'strikethrough', 'font_family', 'color',
-                        'disabled_color', 'halign', 'valign', 'padding_x',
-                        'padding_y', 'outline_width', 'disabled_outline_color',
+                        'disabled_color', 'halign', 'valign',
+                        'padding', 'padding_x', 'padding_y',
```
misl6, line 307: At line **345**:
```python
dkw = {x: getattr(self, x) for x in self._font_properties}
```

`padding_x` and `padding_y` are accessed even if the user did not specify them.
This is (correctly) printing a deprecation warning.

We should avoid to pass `padding_x` and `padding_y` to `CoreLabel` or `CoreMarkupLabel`, since are deprecated also of on `CoreLabel` and `CoreMarkupLabel` side.

The suggestion here is to remove `padding_x` and `padding_y` from `_font_properties`, and "manually" bind for changes (we are doing the same for `disabled` property)

misl6 (changes requested):
Hi @DexerBR !

The feature is great, and I really want to have it into **2.2.0** release, but we need to fix a few things regarding deprecations and migration paths.

I've left a few comments, which are trying to explain what needs to be done.

Feel free to ping me if you need any additional guidance.

kivy/core/text/__init__.py
```diff
@@ -182,6 +184,13 @@ class LabelBase(object):
         `text_language`: str, defaults to None (user locale)
             RFC-3066 format language tag as a string (Pango only)
 
+    .. deprecated:: 2.2.0
+        `padding_x` and `padding_y` have been deprecated. Please use `padding`
+        instead.
+
+    .. versionchanged:: 2.2.0
+        `padding` is now a list and defaults to [0, 0, 0, 0].
```
misl6, line 192: Minor: Maybe we can say that still accepts an `int` ?

kivy/core/text/__init__.py
```diff
@@ -263,11 +272,34 @@ def __init__(
...
+
+        options['padding'] = kwargs_get('padding', [0, 0, 0, 0])
+        if isinstance(options['padding'], (int, float)):
+            options['padding'] = [options['padding']] * 4
+        elif (
+            isinstance(options['padding'], (list, tuple))
+            and len(options['padding']) != 4
+        ):
+            if len(options['padding']) == 1:
+                options['padding'] = options['padding'] * 4
```
misl6, line 284: Previously `len(options['padding']) == 1` was not supported.

Do we really want to introduce this chance?

kivy/core/text/__init__.py
```diff
@@ -286,6 +318,15 @@ def __init__(
         self.texture = None
         self.is_shortened = False
         self.resolve_font_name()
+        self._update_padding()
+
+    def _update_padding(self):
```
misl6, line 323: Minor:

How about `_migrate_deprecated_padding_xy` ? (or something similar)

kivy/core/text/__init__.py
```diff
@@ -286,6 +318,15 @@ def __init__(
...
+        options = self.options
+        self.options['padding'] = list(self.options['padding'])
+        if options['padding_x']:
+            self.options['padding'][::2] = [options['padding_x']] * 2
```
misl6, line 327: Maybe we can also print a deprecation warning?

kivy/core/text/__init__.py
```diff
@@ -263,11 +275,34 @@ def __init__(
...
+
+        options['padding_x'] = kwargs_get('padding_x')
+        options['padding_y'] = kwargs_get('padding_y')
+        for padding_option in ('padding_x', 'padding_y'):
+            if kwargs_get(padding_option):
+                Logger.warning(
+                    f"The use of the {padding_option} parameter is "
+                    "deprecated, and will be removed in future versions. Use "
+                    "padding instead."
+                )
```
misl6, line 305:
```suggestion
                Logger.warning(
                    f"LabelBase: The use of the {padding_option} parameter is "
                    "deprecated, and will be removed in future versions. Use "
                    "padding instead."
                )
```

kivy/core/text/__init__.py
misl6: Ouch! My bad, seen it.

But it's missing the classname that allows the Kivy logger to print it with the proper format.
Left a suggestion on it.

misl6 (changes requested):

kivy/core/text/__init__.py
```diff
@@ -671,7 +715,7 @@ def _render_real(self):
         size = self.size
         valign = options['valign']
 
-        y = ypad = options['padding_y']  # pos in the texture
+        y = ypad = options['padding'][1]  # pos in the texture
```
misl6, line 718: Same applies here, probably better to just use `options['padding'][1]` instead of a generic `ypad` variable.

kivy/core/text/markup.py
```diff
@@ -165,7 +165,7 @@ def _pre_render(self):
         # flattened later when shortening and broken up lines if broken
         # mid-word will have space mid-word when lines are joined
         uw_temp = None if shorten else uw
-        xpad = options['padding_x']
+        xpad = options['padding'][0]
```
misl6, line 168: Since we also have another case at lines just below, that uses `xpad` for justifying text, that needs to be checked, is probably better to just use `options['padding'][0]` and `options['padding'][2]` instead of a generic `xpad` variable.
 (Would be nice to have some test that exercises this case)

misl6 (changes requested):

kivy/core/text/__init__.py
```diff
@@ -593,7 +637,7 @@ def find_base_direction(text):
     def render_lines(self, lines, options, render_text, y, size):
         get_extents = self.get_cached_extents()
         uw, uh = options['text_size']
-        xpad = options['padding_x']
+        xpad = options['padding'][0]
         if uw is not None:
             uww = uw - 2 * xpad  # real width of just text
```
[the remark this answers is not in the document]
DexerBR: In my last commit I am taking into account the "`xpad`" as `padding_left` and `padding_right` 👍

kivy/core/text/markup.py
```diff
@@ -463,11 +463,11 @@ def render_lines(self, lines, options, render_text, y, size):
 
         for layout_line in lines:  # for plain label each line has only one str
             lw, lh = layout_line.w, layout_line.h
-            x = xpad
+            x = padding_right
```
misl6, line 466: Should not be `padding_left` here, instead of `padding_right` ?

kivy/core/text/__init__.py
```diff
@@ -659,14 +660,15 @@ def render_lines(self, lines, options, render_text, y, size):
...
                     cur_base_dir = find_base_dir(line)
-            x = xpad
+            x = padding_left
             if halign == 'auto':
                 if cur_base_dir and 'rtl' in cur_base_dir:
-                    x = max(0, int(w - lw - xpad))  # right-align RTL text
+                    # right-align RTL text
+                    x = max(0, int(w - lw - padding_right))
             elif halign == 'center':
                 x = int((w - lw) / 2.)
```
misl6, line 669: Since `padding_left` and `padding_right` now can be different, we should also center the text appropriately? (by also taking into account `rtl` text)

misl6 (changes requested):

kivy/core/text/markup.py
DexerBR: This is interesting... I'm not sure if this code already expected to receive a value for the padding right (as previously the left and right values were the same), but passing the value of `padding_left` instead of `padding_right` will have the opposite effect of what is expected

misl6: > will have the opposite effect of what is expected

I tested it this morning (before having the first coffee of the day), and changing `x = padding_right` to `x = padding_left`, actually worked with `halign = 'left'`.

What is your expectation, and how it looks instead?

DexerBR: Oops, my bad, I think I got confused while testing the different alignment modes...

You're right! I tested again and it looks like changing `x = padding_right` to `x = padding_left` did work! Good catch!

kivy/core/text/__init__.py
DexerBR: Previously, only the right padding (the same as the left padding) was taken into account, so I think it's ok, or not?

misl6: I'm referring to the `halign == 'center'` case.

Previously, with only 1 `padding_x` padding value, the padding was in-influent to calculate the text positioning.
Now, instead, with `padding_left` and `padding_right`, we should take it into account.

misl6 (approved):
LGTM. Thank you!

PR #8163, opened 2023-03-11
`Line`/`SmoothLine`: `ellipse` - fix behavior and add feature to allow closing line through center of ellipse

DexerBR:
Considering:
- #8164

**The two graphical instructions will have representations that coincide with each other:**

https://user-images.githubusercontent.com/73297572/227750228-e3c6d068-ea64-4d9c-bec7-2d6480623178.mp4

<br>
<br>

Demo:

https://user-images.githubusercontent.com/73297572/227750230-6a03802f-acc1-4ae1-963e-03331387ba26.mp4

Source code:

```python
from kivy.app import App
from kivy.lang import Builder
from kivy.uix.boxlayout import BoxLayout


class UI(BoxLayout):
    pass


Builder.load_string(
    """
<UI>:
    start_angle: 0
    end_angle: 0
    segments: 100
    close: 0
    close_mode: "straight-line"

    BoxLayout:
        orientation: "vertical"
        canvas:
            Color:
                rgba: 0.2, 0.2, 0.2, 1
            Rectangle:
                pos: self.pos
                size: self.size

            # Color:
            #     rgba: 0, 0, 0, 1
            # Ellipse:
            #     pos: self.center_x - 100, self.center_y
            #     size: 200, 200
            #     angle_start: root.start_angle
            #     angle_end: root.end_angle
            #     segments: root.segments

            Color:
                rgba: 1, 1, 1, 1
            SmoothLine:
                width: 1
                overdraw_width: 2
                ellipse: (self.center_x - 100, self.center_y, 200, 200, root.start_angle, root.end_angle, root.segments)
                close: root.close
                close_mode: root.close_mode

        Label:
            text: f"start_angle: {root.start_angle} | end_angle: {root.end_angle} | segments: {root.segments}"
            size_hint_y: None
            height: 50
        
        BoxLayout:
            size_hint_y: None
            height: self.minimum_height
            Label:
                text: "close"
            Switch:
                size_hint_y: None
                height: 40
                on_active:
                    root.close = int(self.active)
        BoxLayout:
            size_hint_y: None
            height: self.minimum_height
            Label:
                text: "straight-line"
            CheckBox:
                group: "mode"
                size_hint_y: None
                height: 40
                allow_no_selection: False
                active: True
                on_active:
                    root.close_mode = "straight-line"
            Label:
                text: "center-connected"
            CheckBox:
                group: "mode"
                size_hint_y: None
                height: 40
                allow_no_selection: False
                on_active:
                    root.close_mode = "center-connected"
[...]

PR #8164, opened 2023-03-11
`Ellipse`: Handle the number of segments and avoid division by zero

DexerBR:
After a few divisions by zero, the app would crash immediately, probably due to segfault.

Considering:
- #8163

https://user-images.githubusercontent.com/73297572/224500939-b024dbdf-3278-4c1e-b3be-bf3ef96661d4.mp4

Source Code:

```python
from kivy.app import App
from kivy.lang import Builder


class UI(BoxLayout):
    pass


Builder.load_string(
    """
<UI>:
    start_angle: 0
    end_angle: 0
    segments: 100
    close: 0
    close_from_center: 0

    BoxLayout:
        orientation: "vertical"
        canvas:
            Color:
                rgba: 0.2, 0.2, 0.2, 1
            Rectangle:
                pos: self.pos
                size: self.size

            Color:
                rgba: 0, 0, 0, 1
            Ellipse:
                pos: self.center_x - 100, self.center_y
                size: 200, 200
                angle_start: root.start_angle
                angle_end: root.end_angle
                segments: root.segments

            Color:
                rgba: 1, 1, 1, 1
            SmoothLine:
                width: 1
                overdraw_width: 2
                ellipse: (self.center_x - 100, self.center_y, 200, 200, root.start_angle, root.end_angle, root.segments)
                close: root.close
                close_from_center: root.close_from_center

        Label:
            text: f"start_angle: {root.start_angle} | end_angle: {root.end_angle} | segments: {root.segments}"
            size_hint_y: None
            height: 50
        
        BoxLayout:
            size_hint_y: None
            height: self.minimum_height
            Label:
                text: "close"
            Switch:
                size_hint_y: None
                height: 40
                on_active:
                    root.close = int(self.active)
        BoxLayout:
            size_hint_y: None
            height: self.minimum_height
            Label:
                text: "close_from_center"
            Switch:
                size_hint_y: None
                height: 40
                on_active:
                    root.close_from_center = int(self.active)
        Slider:
            size_hint_y: None
            height: 50
            min: -360
            max: 360
            step: 1
            on_value:
                root.start_angle = self.value
        Slider:
            size_hint_y: None
            height: 50
            min: -360
            max: 360
            step: 1
            on_value:
                root.end_angle = self.value
        Slider:
            size_hint_y: None
            height: 50
            min: -10
            max: 150
            step: 1
            on_value:
                root.segments = int(self.value)
"""
)


class Testapp(App):

    def build(self):
        return UI()


Testapp().run()
```

Maintainer merge checklist

### DIFF
--- a/kivy/graphics/vertex_instructions.pyx
+++ b/kivy/graphics/vertex_instructions.pyx
@@ -1123,7 +1123,8 @@ cdef class Ellipse(Rectangle):
     :Parameters:
         `segments`: int, defaults to 180
             Define how many segments are needed for drawing the ellipse.
-            The drawing will be smoother if you have many segments.
+            You can use this property to create polygons with 3 or more sides.
+            The default value is calculated from the range between angle.
         `angle_start`: float, defaults to 0.0
             Specifies the starting angle, in degrees, of the disk portion.
         `angle_end`: float, defaults to 360.0
@@ -1148,10 +1149,13 @@ cdef class Ellipse(Rectangle):
         cdef double cx, cy, tangential_factor, radial_factor, fx, fy
         cdef vertex_t *vertices = NULL
         cdef unsigned short *indices = NULL
-        cdef int count = self._segments
+        cdef int segments = self._segments
 
         if self.w == 0 or self.h == 0:
             return
+
+        if segments < 3:
+            segments = max(1, int(abs(self._angle_end - self._angle_start) / 2))
 
         tx = tc[0]
         ty = tc[1]
@@ -1161,11 +1165,11 @@ cdef class Ellipse(Rectangle):
         rx = 0.5 * self.w
         ry = 0.5 * self.h
 
-        vertices = <vertex_t *>malloc((count + 2) * sizeof(vertex_t))
+        vertices = <vertex_t *>malloc((segments + 2) * sizeof(vertex_t))
         if vertices == NULL:
             raise MemoryError('vertices')
 
-        indices = <unsigned short *>malloc((count + 2) * sizeof(unsigned short))
+        indices = <unsigned short *>malloc((segments + 2) * sizeof(unsigned short))
         if indices == NULL:
             free(vertices)
             raise MemoryError('indices')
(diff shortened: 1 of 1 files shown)


kivy/graphics/vertex_instructions.pyx
```diff
@@ -1148,11 +1149,14 @@ cdef class Ellipse(Rectangle):
...
         cdef unsigned short *indices = NULL
-        cdef int count = self._segments
+        cdef int segments = self._segments
 
         if self.w == 0 or self.h == 0:
             return
 
+        if segments < 3:
+            segments = max(1, int(abs(self._angle_end - self._angle_start) / 2))
+
```
misl6, line 1159: Can we print a warning so the developer can notice it?

kivy/graphics/vertex_instructions.pyx
```diff
@@ -1123,7 +1123,8 @@ cdef class Ellipse(Rectangle):
     :Parameters:
         `segments`: int, defaults to 180
             Define how many segments are needed for drawing the ellipse.
-            The drawing will be smoother if you have many segments.
+            You can use this property to create polygons with 3 or more sides.
+            The default value is calculated from the range between angle.
```
misl6, line 1127: Since it defaults to `180`, why not using the default value (`180`) also if `segments < 3` ?

misl6 (commented):
Hi @DexerBR !

Thank you for addressing this issue.
I've left a couple of comments, let me know what do you think about it.

kivy/graphics/vertex_instructions.pyx
DexerBR: We can, however the calculation of segments based on the angle has the advantage of being more efficient.

For example, if the angle range is 90 degrees then we will have 45 segments instead of 180, keeping a good visual appearance. The value of 180 was probably intended for a range of 360 degrees (default).

DexerBR: Now the default value will be automatically calculated based on the angle range, as it is the most efficient approach 🙂

kivy/graphics/vertex_instructions.pyx
```diff
@@ -1148,11 +1152,16 @@ cdef class Ellipse(Rectangle):
...
         cdef unsigned short *indices = NULL
-        cdef int count = self._segments
+        cdef int segments = self._segments
 
         if self.w == 0 or self.h == 0:
             return
 
+        if segments == 0 or segments < 3:
+            if segments != 0:
+                Logger.warning('Ellipse: Number of segments not supported, they will be automatically set.')
```
misl6, line 1162:
```suggestion
                Logger.warning('Ellipse: A minimum of 3 segments is required. The default value will be used instead.')
```

misl6 (commented):
I left a couple of minor suggestions, but LGTM!

misl6 (approved):
LGTM. Thank you!

PR #8169, opened 2023-03-15
`Image`: add `fit_mode` feature

DexerBR:
This PR adds the `fit_mode` feature, as a replacement for `allow_stretch` and `keep_ratio`.

The previous behaviors using `allow_stretch` and `keep_ratio` will be present in `fit_mode`, plus a new mode is being introduced: `"cover"`. The `"cover"` mode will fill the widget box with the image regardless of the size and aspect ratio of both, keeping the aspect ratio of the image.

The available options will be
    
- ``"scale-down"``: the image will be scaled down to fit inside the widget box, **maintaining its aspect ratio and without stretching**. If the size of the image is smaller than the widget, it will be displayed at its original size. If the image has a different aspect ratio than the widget, there will be blank areas on the widget box.

- ``"fill"``: the image is stretched to fill the widget, **regardless of its aspect ratio or dimensions**. If the image has a different aspect ratio than the widget, this option can lead to distortion of the image.

- ``"contain"``: the image is resized to fit inside the widget box, **maintaining its aspect ratio**. If the image size is larger than the widget size, the behavior will be similar to ``"scale-down"``. However, if the size of the image size is smaller than the widget size, unlike ``"scale-down``, the image will be resized to fit inside the widget. If the image has a different aspect ratio than the widget, there will be blank areas on the widget box.

- ``"cover"``: the image will be stretched horizontally or vertically to fill the widget box, **maintaining its aspect ratio**. If the image has a different aspect ratio than the widget, then the image will be clipped to fit.

**Demo**:

https://user-images.githubusercontent.com/73297572/226389904-02e783b4-7705-47bd-bea8-a75d6eb75976.mp4

<br>

**image_test.png (250x250px)**:

image_test

**Code for testing** (It is recommended to test with images with different aspect ratios):
```python
from kivy.app import App
from kivy.lang import Builder


class UI(FloatLayout):
    pass


Builder.load_string(
    """
<UI>:
    Image:
        id: img
        source: "image_test.png"
        size_hint: None, None
        pos_hint: {"center": (0.5, 0.6)}
        mipmap: True
        canvas.after:
            Color:
                rgba: 1, 0, 0, 0.5
            Rectangle:
                pos: self.pos
                size: self.size

    BoxLayout:
        size_hint_y: 0.4
        orientation: "vertical"

        Label:
            text: f"fit_mode = {img.fit_mode} \\nwidth = {img.width} | height = {img.height}"
            outline_width: 1
            font_size: dp(18)

        Slider:
            size_hint_y: None
            height: 50
            min: 0
            max: 3
            step: 1
            on_value:
                img.fit_mode = ("scale-down", "fill", "contain", "cover")[int(self.value)]
        Slider:
            size_hint_y: None
            height: 50
            min: 0
            max: 1000
            step: 1
            on_value:
                img.width = self.value
        Slider:
            size_hint_y: None
            height: 50
            min: 0
            max: 1000
            step: 1
            on_value:
                img.height = self.value

"""
)


class Testapp(App):

    def build(self):
        return UI()


Testapp().run()
```

Maintainer merge checklist

### DIFF
--- a/kivy/uix/image.py
+++ b/kivy/uix/image.py
@@ -58,8 +58,16 @@ __all__ = ('Image', 'AsyncImage')
 from kivy.uix.widget import Widget
 from kivy.core.image import Image as CoreImage
 from kivy.resources import resource_find
-from kivy.properties import StringProperty, ObjectProperty, ListProperty, \
-    AliasProperty, BooleanProperty, NumericProperty, ColorProperty
+from kivy.properties import (
+    StringProperty,
+    ObjectProperty,
+    ListProperty,
+    AliasProperty,
+    BooleanProperty,
+    NumericProperty,
+    ColorProperty,
+    OptionProperty
+)
 from kivy.logger import Logger
 
 # delayed imports
@@ -67,8 +75,7 @@ Loader = None
 
 
 class Image(Widget):
-    '''Image class, see module documentation for more information.
-    '''
+    '''Image class, see module documentation for more information.'''
 
     source = StringProperty(None)
     '''Filename / source of your image.
@@ -104,7 +111,7 @@ class Image(Widget):
     def get_image_ratio(self):
         if self.texture:
             return self.texture.width / float(self.texture.height)
-        return 1.
+        return 1.0
 
     mipmap = BooleanProperty(False)
     '''Indicate if you want OpenGL mipmapping to be applied to the texture.
@@ -138,18 +145,22 @@ class Image(Widget):
         :class:`~kivy.properties.ColorProperty`.
     '''
 
-    allow_stretch = BooleanProperty(False)
+    allow_stretch = BooleanProperty(False, deprecated=True)
     '''If True, the normalized image size will be maximized to fit in the image
     box. Otherwise, if the box is too tall, the image will not be
     stretched more than 1:1 pixels.
 
     .. versionadded:: 1.0.7
 
+    .. deprecated:: 2.2.0
+        :attr:`allow_stretch` have been deprecated. Please use `fill_mode`
+        instead.
+
     :attr:`allow_stretch` is a :class:`~kivy.properties.BooleanProperty` and
     defaults to False.
     '''
 
-    keep_ratio = BooleanProperty(True)
+    keep_ratio = BooleanProperty(True, deprecated=True)
     '''If False along with allow_stretch being True, the normalized image
     size will be maximized to fit in the image box and ignores the aspect
     ratio of the image.
@@ -158,8 +169,35 @@ class Image(Widget):
 
     .. versionadded:: 1.0.8
 
+    .. deprecated:: 2.2.0
+        :attr:`keep_ratio` have been deprecated. Please use `fill_mode`
+        instead.
+
     :attr:`keep_ratio` is a :class:`~kivy.properties.BooleanProperty` and
     defaults to True.
+    '''
+
+    fill_mode = OptionProperty('none', options=['none', 'stretch', 'fit', 'fill'])
+    '''If the image size is smaller than the widget size, determines how the
+    image should be stretched to fill the widget box.
+
+    Available options:
+
+    - ``"none"``: the image will not be stretched.
+
+    - ``"stretch"``: the image is stretched to fill the widget, regardless of
+    its aspect ratio or dimensions. This can lead to distortion of the image if
+    it has a different aspect ratio than the widget.
+
+    - ``"fit"``: the image will be maximized to fit in the widget box,
+    **maintaining its aspect ratio**. This can result in blank areas on the
+    widget if the image has a different aspect ratio than the widget.
+
+    - ``fill``: the image will be stretched horizontally or vertically to
+    fill the widget box, **maintaining its aspect ratio**.
+
+    :attr:`fill_mode` is a :class:`~kivy.properties.OptionProperty` and
+    defaults to ``"none"``.
     '''
 
     keep_data = BooleanProperty(False)
@@ -172,7 +210,7 @@ class Image(Widget):
     defaults to False.
     '''
 
-    anim_delay = NumericProperty(.25)
+    anim_delay = NumericProperty(0.25)
     '''Delay the animation if the image is sequenced (like an animated gif).
     If anim_delay is set to -1, the animation will be stopped.
 
@@ -196,22 +234,25 @@ class Image(Widget):
     internal cache. The cache will simply ignore any calls trying to
     append the core image.
 
-    .. versionadded:: 1.6.0
+    .. versionadded:: 1.6.00
 
     :attr:`nocache` is a :class:`~kivy.properties.BooleanProperty` and defaults
     to False.
     '''
 
     def get_norm_image_size(self):
-        if not self.texture:
+        if not self.texture or self.fill_mode == "fill":
             return list(self.size)
+
         ratio = self.image_ratio
         w, h = self.size
         tw, th = self.texture.size
 
         # ensure that the width is always maximized to the container width
-        if self.allow_stretch:
-            if not self.keep_ratio:
+        # NOTE: self.allow_stretch and self.keep_ratio conditions should be
+        # removed along with the deprecated properties in the future
+        if self.fill_mode in ("stretch", "fit") or self.allow_stretch:
+            if self.fill_mode == "stretch" or not self.keep_ratio:
                 return [w, h]
             iw = w
         else:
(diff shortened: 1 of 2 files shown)


kivy/uix/image.py
```diff
@@ -196,22 +234,25 @@ def get_image_ratio(self):
     internal cache. The cache will simply ignore any calls trying to
     append the core image.
 
-    .. versionadded:: 1.6.0
+    .. versionadded:: 1.6.00
```
Kulothungan16, line 237: I think here should be only one zero

Kulothungan16:
@DexerBR Now, I can easily understand this features, `allow_stretch` and `keep_ratio` are always confusing in the past when I started learning kivy. Thanks

kivy/uix/image.py
DexerBR: Thank you for catching it!

DexerBR:
> @DexerBR Now, I can easily understand this features, `allow_stretch` and `keep_ratio` are always confusing in the past when I started learning kivy. Thanks ❤️

I agree! Sometimes it could be confusing to understand which combination is right to achieve a specific filling mode. Also, is it something that could be covered by just one property/option rather than two.

rnixx:
@DexerBR For covering images there exist https://github.com/kivy/kivy/blob/master/kivy/uix/behaviors/cover.py. Maybe this can be deprecated then.

kivy/uix/image.py
```diff
@@ -138,18 +145,22 @@ def get_image_ratio(self):
...
+    .. deprecated:: 2.2.0
+        :attr:`allow_stretch` have been deprecated. Please use `fit_mode`
+        instead.
+
     :attr:`allow_stretch` is a :class:`~kivy.properties.BooleanProperty` and
     defaults to False.
     '''
 
-    keep_ratio = BooleanProperty(True)
+    keep_ratio = BooleanProperty(True, deprecated=True)
```
misl6, line 163: `keep_ratio` has been deprecated, but is in use from Kivy widgets.
That will show a deprecation warning in the console. Better to migrate these widgets as well (even in a different PR, just after this one is merged, but before 2.2.0 release).

E.g.

https://github.com/kivy/kivy/blob/e2c8f0753337cd693ae5b34de0b3bd3a481d725c/kivy/data/style.kv#L76

(We also have a few examples which are using `keep_ratio`, but we can migrate them later, even after 2.2.0)

kivy/uix/image.py
```diff
@@ -138,18 +145,22 @@ def get_image_ratio(self):
         :class:`~kivy.properties.ColorProperty`.
     '''
 
-    allow_stretch = BooleanProperty(False)
+    allow_stretch = BooleanProperty(False, deprecated=True)
```
misl6, line 148: `allow_stretch` has been deprecated, but is in use from Kivy widgets.
That will show a deprecation warning in the console. Better to migrate these widgets as well (even in a different PR, just after this one is merged, but before **2.2.0** release).

E.g.
https://github.com/kivy/kivy/blob/e2c8f0753337cd693ae5b34de0b3bd3a481d725c/kivy/data/style.kv#L75

(We also have a few examples which are using `allow_stretch `, but we can migrate them later, even after **2.2.0**)

kivy/uix/image.py
```diff
@@ -138,18 +145,22 @@ def get_image_ratio(self):
...
     '''If True, the normalized image size will be maximized to fit in the image
     box. Otherwise, if the box is too tall, the image will not be
     stretched more than 1:1 pixels.
 
     .. versionadded:: 1.0.7
 
+    .. deprecated:: 2.2.0
+        :attr:`allow_stretch` have been deprecated. Please use `fit_mode`
```
misl6, line 156: We're still suggesting the usage of `allow_stretch`, even if is now deprecated:
https://github.com/kivy/kivy/blob/e2c8f0753337cd693ae5b34de0b3bd3a481d725c/kivy/uix/image.py#L35
https://github.com/kivy/kivy/blob/e2c8f0753337cd693ae5b34de0b3bd3a481d725c/kivy/uix/image.py#L83

kivy/uix/image.py
```diff
@@ -158,10 +169,49 @@ def get_image_ratio(self):
 
     .. versionadded:: 1.0.8
 
+    .. deprecated:: 2.2.0
+        :attr:`keep_ratio` have been deprecated. Please use `fit_mode`
```
misl6, line 173: We're still suggesting the usage of `keep_ratio`, even if is now deprecated:

https://github.com/kivy/kivy/blob/e2c8f0753337cd693ae5b34de0b3bd3a481d725c/kivy/uix/image.py#L35
https://github.com/kivy/kivy/blob/e2c8f0753337cd693ae5b34de0b3bd3a481d725c/kivy/uix/image.py#L83

kivy/uix/image.py
```diff
@@ -158,10 +169,49 @@ def get_image_ratio(self):
...
+        instead.
+
     :attr:`keep_ratio` is a :class:`~kivy.properties.BooleanProperty` and
     defaults to True.
     '''
 
+    fit_mode = OptionProperty(
+        "scale-down", options=["scale-down", "fill", "contain", "cover"]
+    )
+    '''If the size of the image is smaller than the size of the widget,
```
misl6, line 183:
```suggestion
    '''If the size of the image is different than the size of the widget,
```

kivy/uix/image.py
```diff
@@ -205,13 +255,23 @@ def get_image_ratio(self):
...
+            else:
+                return [(h * tw) / th, h]
+
         # ensure that the width is always maximized to the container width
-        if self.allow_stretch:
-            if not self.keep_ratio:
+        # NOTE: self.allow_stretch and self.keep_ratio conditions should be
+        # removed along with the deprecated properties in the future
+        if self.fit_mode in ("fill", "contain") or self.allow_stretch:
+            if self.fit_mode == "fill" or not self.keep_ratio:
```
misl6, line 274: This will print the following during runtime:
```bash
[WARNING] Deprecated property "<BooleanProperty name=allow_stretch>" of object "<kivy.uix.image.Image object at 0x10d63cc90>" was accessed, it will be removed in a future version
```

Is better to avoid this behavior, and introduce a different migration method, which is accessed only if the developer specifies the `allow_stretch` or `keep_ratio` properties.

As an example, you can bind for changes of `allow_stretch` and `keep_ratio` and select the proper `fit_mode` so the `get_norm_image_size` is easier to read, and does not access `allow_stretch` and `keep_ratio`.

misl6 (changes requested):
Hi @DexerBR !

The feature is super great (and works great), and will help a lot of newcomers (and also veterans) to have more (and easier) control over images!

Unfortunately, a few small things need to be changed (deprecations and migration paths) in order to merge this PR.

I've also added a few comments regarding things that could be addressed later (after this PR is merged)

misl6:
> > I wonder if this behavior would have use cases, outside of the Image widget.
> 
> Good question. In my poor imagination it has not.
> 
> Anyone?

Some imagination produced this: 😅

(I recall to have used it sometimes, even without an `Image` or `Video`, but I do not recall why.)

```python
from kivy.app import App
from kivy.lang import Builder
from kivy.uix.behaviors import CoverBehavior

class UI(BoxLayout):
    pass

class CoverLabel(CoverBehavior, Label):

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.fbind("texture_size", self._update_reference_size)

    def _update_reference_size(self, *args):
        self.reference_size = self.texture_size


Builder.load_string(
"""
<UI>:
    size_hint: 1, 1
    orientation: "vertical"
    CoverLabel:
        size_hint: 1, .5
        color: 1, 1, 1, 1
        font_size: 400
        markup: True
        text: "Hello World by Kivy [color=ff0000]Hello World by Kivy[/color] [color=00ff00]Hello World by Kivy[/color] [color=0000ff]Hello World by Kivy[/color]"
""")

class Testapp(App):

    def build(self):
        return UI()

Testapp().run()
```

kivy/uix/image.py
DexerBR: Hi @misl6! That makes sense! I think this migration can be done in another PR.

kivy/uix/image.py
DexerBR: https://github.com/kivy/kivy/pull/8169#discussion_r1148395736

DexerBR:
> > > I wonder if this behavior would have use cases, outside of the Image widget.
> > 
> > 
> > Good question. In my poor imagination it has not.
> > Anyone?
> 
> Some imagination produced this: 😅
> 
> (I recall to have used it sometimes, even without an `Image` or `Video`, but I do not recall why.)
> 
> ```python
> from kivy.app import App
> from kivy.uix.boxlayout import BoxLayout
> from kivy.uix.behaviors import CoverBehavior
> from kivy.uix.label import Label
> 
> class UI(BoxLayout):
>     pass
> 
> class CoverLabel(CoverBehavior, Label):
> 
>     def __init__(self, **kwargs):
>         super().__init__(**kwargs)
>         self.fbind("texture_size", self._update_reference_size)
> 
>     def _update_reference_size(self, *args):
>         self.reference_size = self.texture_size
> 
> 
> Builder.load_string(
> """
> <UI>:
>     size_hint: 1, 1
>     orientation: "vertical"
>     CoverLabel:
>         size_hint: 1, .5
>         color: 1, 1, 1, 1
>         font_size: 400
>         markup: True
>         text: "Hello World by Kivy [color=ff0000]Hello World by Kivy[/color] [color=00ff00]Hello World by Kivy[/color] [color=0000ff]Hello World by Kivy[/color]"
> """)
> 
> class Testapp(App):
> 
>     def build(self):
>         return UI()
> 
> Testapp().run()
> ```

So I think we don't have enough reasons to deprecate the module - (so far 😄)

misl6 (approved):
LGTM. Thank you!

Let me know if you also want to address the migrations on widgets / examples that are using `keep_ratio`  and `allow_stretch` 😄

DexerBR:
> LGTM. Thank you!
> 
> Let me know if you also want to address the migrations on widgets / examples that are using `keep_ratio` and `allow_stretch` 😄

Done in https://github.com/kivy/kivy/pull/8202

T-Dynamos:
I have a question, does it works with rounded rectangle?

PR #8170, opened 2023-03-17
Add svg rotation transform support

### LINKED ISSUE #8166: Svg transform:rotate() support
I wrote an svg identicon generator I was hoping to use in a Kivy app, but on displaying an svg in Kivy I quickly realized that Kivy does not yet support the SVG rotate() transformation.

I can see the relevant source that needs to be modified at around line 213 here: https://github.com/kivy/kivy/blob/master/kivy/graphics/svg.pyx

Another elif block with the appropriate Matrix transformation should do it. I'm not currently set up to edit and contribute to Kivy source code, else I would probably dive in, figure it out, and submit a pull request myself.

**Describe alternatives you've considered**
I probably will end up rewriting my identicon generator to draw natively in Kivy. I'd really prefer not to, as the library I wrote is much more widely useful generating SVGs rather than being locked into usefulness only in Kivy, and maintaining multiple libraries that effectively do the same thing feels like unnecessary tedium. Alas.

**Additional context**

quasistoic:
Fixes #8166 - ChatGPT wrote the matrix transformations. I merely fixed type and import errors and verified the code works.

Maintainer merge checklist

I didn't include unittests, as I didn't see any other unittests for svg support and so I couldn't pattern logic and organization off of that. Would be happy to add tests if I had another example to work off of. I did include an example svg, which rendered properly once I figured out the build-run-verify workflow.

I'm not certain about label adding and whatnot, as those feel like discretionary decisions that core developers should be making.


quasistoic:
These don't feel like real unittests, but at the very least these screenshot-based tests allow someone to check whether a specific change introduces a regression.

examples/svg/rotated.svg
misl6, line 1: Minor:
What about using something simpler to test this feature?
```svg
<svg width="200" height="200" xmlns="http://www.w3.org/2000/svg">
  <rect fill="#ff0000" width="100" height="100" x="50" y="50" transform="rotate(45 100 100)" />
  <rect fill="#00ff00" width="100" height="100" x="50" y="50" transform="rotate(-60 100 100)" />
  <rect fill="#0000ff" width="50" height="50" x="10" y="10" transform="rotate(45)" />
  <rect fill="#ff00ff" width="50" height="50" x="10" y="10" transform="rotate(-30)" />
</svg>
```

kivy/tests/test_graphics_svg.py
```diff
@@ -0,0 +1,92 @@
+'''
+Svg tests
+==============
+
+Testing Svg rendering.
+'''
+
+import sys
+import pytest
+from kivy.tests.common import GraphicUnitTest, requires_graphics
```
misl6, line 10: `sys`, `pytest` and `requires_graphics` are not needed here.

kivy/tests/test_graphics_svg.py
```diff
@@ -0,0 +1,92 @@
...
+        # create a root widget
+        wid = Widget()
+
+        # put some graphics instruction on it
+        with wid.canvas:
+            svg = Svg()
+            svg.set_tree(ET.ElementTree(ET.fromstring(SIMPLE_SVG)))
+
+        # render, and capture it directly
+        r(wid)
```
misl6, line 56: Minor:
`self.render(wid)` looks cleaner

misl6 (changes requested):
Hi @quasistoic ,

I left a couple of suggestions but looks good to me in general. Let me know!

kivy/tests/test_graphics_svg.py
quasistoic: Fixed.

kivy/tests/test_graphics_svg.py
quasistoic: Agreed, fixed.

examples/svg/rotated.svg
quasistoic: I switched to something simpler and cleaned up the formatting so it's easier to read what's going on. I kept the three distinct shapes as I feel it makes it easiest to see what's going on in the background of the ./examples/svg/main.py window. LMK if you think it requires further simplification. Thanks!

quasistoic:
Comments addressed and ready for re-review. Thanks for looking!

misl6 (approved):
LGTM. Thank you!

PR #8184, opened 2023-03-24
Revert "Detect High DPI on Linux Desktop"

akshayaurora:
Reverts kivy/kivy#8147

Previous way of using size difference between Window.size and Windows.get_DrwableSize is the correct way as outlined by SDL2 Documentation too.

The reason on why currently on Wayland window reports both to be the same cause apparently we are not properly using Wayland window provider.

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -266,13 +266,6 @@ cdef class _WindowSDL2Storage:
         SDL_GetWindowSize(self.win, &w, &h)
         return w, h
 
-
-    def _get_display_dpi(self):
-        cdef int display_index = SDL_GetWindowDisplayIndex(self.win)
-        cdef float horizontal_dpi = 0.
-        SDL_GetDisplayDPI(display_index, NULL, &horizontal_dpi, NULL)
-        return horizontal_dpi
-
     def _set_cursor_state(self, value):
         SDL_ShowCursor(value)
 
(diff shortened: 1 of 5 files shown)


jotoft:
@akshayaurora can you explain better what the issue with this pull request was? The window size vs drawable area size was still the primary way of detecting the DPI in this PR.

The display DPI was just a secondary way of getting the DPI in case the primary way fails.

Without this PR there is no support at all for detecting DPI so the scaling does not work at all on Linux.

akshayaurora:
The problem with that pr is that it tries to work around on get_gl_size not working. 

get_gl_size is supposed to give a dpi adjusted size. It only supposed to work on Linux when using wayland(SDL2 Window provider with the video driver set to wayland).

Your solution worked around the issue trying to detect display dpi directly.  Which is not supposed to be accurate even according to the documentation on SDL2 itself.

Correct solution here is to figure out why the get_gl_size is reporting same as Window.size/system_size instead of dpi adjusted size.(Probably need to properly enable wayland video driver but needs a bit more investigation)

1)Figure out why get_gl_size is not working/reporting correct size. i.e  If we are not accurately setting the video driver or not compiling the required wayland support then that needs to be fixed.(This would fix things for the binary wheel installations on linux).

2) We probably need to detect env i.e Wayland/X11/FBO  and then adjust scaling appropriately.

Your pr was a step in right direction just needs a bit adjustment to account for the above.

FilipeMarch:
The `dpi` calculation is clearly wrong on my laptop
```python

runTouchApp(Label(text="hello world"))
```
This is how `Hello World` app should look like
image

This is how it looks like on my laptop (Arch Linux, `kivy==master`)
image

This is my laptop screen config (not sure if this matters)
image

This is how `kivy==master` vs `kivy==2.1.0` looks like on my laptop
image

PR #8194, opened 2023-04-03
remove obvious comment

jelgun:
This PR removes an obvious comment (i.e. comment that restates what the code does in an obvious manner). The code itself is understandable that the clock is installed.


misl6:
Hi @jelgun !

PR #8195, opened 2023-04-03
remove beautification lines

jelgun:
This PR removes a beautification comment (i.e. a comment that includes an array of special characters for beautification).


misl6:
Hi @jelgun !

PR #8196, opened 2023-04-03
remove fixme comment

jelgun:
This PR removes a task comment (i.e. a comment referring to a work that could/should be done in the future or was already done). Instead, an issue was created in https://github.com/kivy/kivy/issues/8199


misl6:
Hi @jelgun !

PR #8207, opened 2023-04-13
Fix dropped motion events in mtdev provider.

rnixx:
Maintainer merge checklist

with a current setup on a 6.1.0 linux kernel and a multi touch display i encountered motion events getting dropped.

The mtdev input provider processes values if MTDEV_CODE_TRACKING_ID event is received and it's value denotes an unused slot. However, this not happens in my case, so i added processing on position changes as well to ensure on_touch_move events get fired properly.

Actually i have not tried with another kernel, so i have no idea if there are changes in libmtdev or if the multitouch display i use not follows standards with sending data.

Also i have not digged in detail into the mtdev docs (https://docs.kernel.org/6.1/input/multi-touch-protocol.html) nor did i studied the input provider implementation in depth, maybe there's a better place to fix this issue.

### DIFF
--- a/kivy/input/providers/mtdev.py
+++ b/kivy/input/providers/mtdev.py
@@ -283,7 +283,7 @@ else:
 
             invert_x = int(bool(drs('invert_x', 0)))
             invert_y = int(bool(drs('invert_y', 0)))
-            Logger.info('MTD: <%s> axes inversion: X is %d, Y is %d' %
+            Logger.info('MTD: <%s> axes invertion: X is %d, Y is %d' %
                         (_fn, invert_x, invert_y))
 
             rotation = drs('rotation', 0)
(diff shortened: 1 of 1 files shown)


rnixx:
According to the Multi-touch (MT) Protocol docs (https://www.kernel.org/doc/html/v4.19/input/multi-touch-protocol.html)

> Creation, replacement and destruction of contacts is achieved by modifying the ABS_MT_TRACKING_ID of the associated slot. A non-negative tracking id is interpreted as a contact, and the value -1 denotes an unused slot.

In the current implementation, changes are applied by interpreting ABS_MT_TRACKING_ID -1 as sync, which is not the way it is supposed to be done.

When taking a look at the protocol B example (https://www.kernel.org/doc/html/v4.19/input/multi-touch-protocol.html#protocol-example-b)

> ABS_MT_SLOT 0
ABS_MT_TRACKING_ID 45
ABS_MT_POSITION_X x[0]
ABS_MT_POSITION_Y y[0]
ABS_MT_SLOT 1
ABS_MT_TRACKING_ID 46
ABS_MT_POSITION_X x[1]
ABS_MT_POSITION_Y y[1]
SYN_REPORT

there can be changes for several slots followed by a EV_SYN/SYN_REPORT which should be used to process changes.

> All drivers mark the end of a multi-touch transfer by calling the usual input_sync() function. This instructs the receiver to act upon events accumulated since last EV_SYN/SYN_REPORT and prepare to receive a new set of events/packets.

So i changed the implementation/PR to process events every time a sync is received and there are changes of interest. i also did some minor polishing with the changes set to reduce noise.

IMO this PR is now ready to be merged. If someone has objections please let me know

rnixx:
anyone?

misl6 (approved):
> anyone?

I'm unable to test it on runtime as I do not have such a setup ready, but:
- Code looks good
- You digged into the MTDev code and made a perfect explanation on "why we should change it", and you made changes to the PR accordingly.

So, that's an LGTM.
Please just consider taking care of this specific provider if you're using it quite extensively in the future.

PR #8212, opened 2023-04-21
vkeyboard: fix touch inputs in outer margin closing the keyboard

D4SK:
Currently a touch event in the outer margin of the vkeyboard causes the touch event to be forwarded to other widgets. This causes a textinput to unfocus, therefore the keyboard disappears. Because the margin area is right next to the keys this can easily happen.

Maintainer merge checklist


D4SK:
Thinking about this further, this is probably not the right way to fix this. If anyone knows the root cause of why a press just below the keyboard can unfocus the textinput, even though the keyboard is at the bottom of the screen, let me know!

PR #8216, opened 2023-04-22
Added client certificate authentication to UrlRequest.py

elfarolab:
Hello,

This pull request is for adding client certificate authentication to UrlRequest.py.
The client certificate can be used for authentication with the API server where certificate verification takes place.
As per protocol, server and client certificates must be signed by the same CA.

Added parameters to UrlRequest:
- cert_file : the client certificate
- key_file : the key of the client certificate
- key_password : the password of the key if any. It could be also a function that results the password as text

Example:

        req = UrlRequest(webservices_url,
                on_success=self.req_made,
                on_error=self.req_error,
                on_failure=self.req_error,
                debug=True,
                ca_file=ca_file,
                verify=False,
                cert_file=cert_file,
                key_file=key_file,
                key_password=key_password,
                req_body=params,
                req_headers=headers)

Thank you so much to everybody!

Maintainer merge checklist

### DIFF
--- a/kivy/network/urlrequest.py
+++ b/kivy/network/urlrequest.py
@@ -120,6 +120,9 @@ class UrlRequestBase(Thread):
 
         Parameters `on_finish` added.
         Parameters `auth` added.
+    
+    .. versionchanged:: 2.3.0
+
         Parameters `cert_file` added.
         Parameters `key_file` added.
         Parameters `key_password` added.
--- a/kivy/network/urlrequest.py
+++ b/kivy/network/urlrequest.py
@@ -18,7 +18,7 @@ The syntax to create a request::
     req = UrlRequest(url, on_success, on_redirect, on_failure, on_error,
                      on_progress, req_body, req_headers, chunk_size,
                      timeout, method, decode, debug, file_path, ca_file,
-                     verify)
+                     verify, cert_file, key_file, key_password)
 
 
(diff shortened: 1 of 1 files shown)


misl6:
Hi @elfarolab !

Nice feature! Sorry for the wait, but as you might have noticed we were into a feature-freeze mode for `2.2.0` (which is now released).

How about adding some tests that exercise the new feature?

elfarolab:
Hello

it was a piece of code for SSL network client requests. I don't know if it is a good idea to write a test for that.

It will require a public SSL server with legal SSL certificate to validate the certificate of the client request. It will be a huge test, difficult to maintain, it does require constant maintenance. In the end is much more bigger effort than the 3 lines of code I added.

Any thoughts?

thx

misl6:
> To test the added 3 lines of options we need to create a public SSL server API with a legal signed SSL certificate to validate the certificate of the client requests

Absolutely agree that this kind of test is not something that we want to maintain. (even simple HTTP get tests are just too noisy on our CI pipeline, as network issues may badly affect tests.

I was instead suggesting something easier:
- With `cert_file`, `key_file` and `key_password` provided, is the SSL context prepared as we expect?
- Are `cert_file`, `key_file` and `key_password` used for all our UrlRequest implementations? (`urllib` and `requests`)

FYI: Seems is not used in `UrlRequestRequests`

PR #8233, opened 2023-05-05
Update setup.cfg

RANJITHROSAN17:
We needed to pin Sphinx ~=6.2.1 since we're incompatible with Sphinx 7.0.0


misl6:
Hi @RANJITHROSAN17 !

We specifically selected `sphinx<=6.2.1` as we should (ATM) keep the compatibility with **Python 3.7**, which is incompatible with `sphinx==6.x`.

Instead I do not get why the change you made in `kivy/__init__.py` is needed. Can you please open a separate PR describing why is that needed?

PR #8256, opened 2023-05-23
Fix errors when simulating (arm) on a Native M1 Mac (Design for Iphone/Ipad)

mrtruongleo:
Fix errors when simulating on a Native M1 Mac (Design for Iphone/Ipad):

`TypeError: unsupported operand type(s) for %: 'NoneType' and 'str' Application quit abnormally!`

With a native M1 Mac, we can run builds directly in My Mac (Design for Iphone/Ipad) simulator much faster compared to the iPhone simulator. We no longer need to build an Intel version in the production of recipes using the --arch=arm64 flag.
Xcode build setting:
-Architechtures: arm64
-Excluded Architechtures: x86_64

### DIFF
--- a/kivy/__init__.py
+++ b/kivy/__init__.py
@@ -381,44 +381,47 @@ if not environ.get('KIVY_DOC_INCLUDE'):
     Logger.setLevel(level=level)
 
     # Can be overridden in command line
-    if ('KIVY_UNITTEST' not in environ and
-            'KIVY_PACKAGING' not in environ and
-            environ.get('KIVY_NO_ARGS', "false") not in ('true', '1', 'yes')):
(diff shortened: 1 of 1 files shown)


kivy/__init__.py
```diff
@@ -381,44 +381,47 @@ def _patch_mod_deps_win(dep_mod, mod_name):
...
+            sys.argv = sys_argv[0:1] + args
+            if mp_fork is not None:
+                # Needs to be first opt for support_freeze to work
+                sys.argv.insert(1, '--multiprocessing-fork')
 
-    else:
+        else:
+            opts = []
+            args = []
+    except:
```
akshayaurora, line 422: This is for worse than failing, as it skips over the actual error. Without any warning /indication either.

Issue #8160, opened 2023-03-09
Throw exception for screens that do not have a name


benpayne:
I created a few screens under a screen manager in a kv file and forgot to specify the "name" property on all of them.  As a result I was not able to change to another screen once the program ran.  I suspect that it saw all screens with the same name of empty string and thought I was already on the screen I was asking for.  So it seems needed to create a unique name for each screen.  As such it would be ideal if an exception was thrown if a screen was created without a name.  Also it would be nice if a screen was added to a screen manager with a non-unique name that it also throw an exception.  

Example:

```
ScreenManager:
        id: top_screen_manager
        Screen:
            id: screen1
            _name: "screen1"_
            Label:
                text: 'This is screen 1'

        Screen:
            id: screen2
            _name: "screen2"_
            Label:
                text: 'This is screen 2'
```

Then in py file:

```
    self.ids.top_screen_manager.switch_to(self.ids.screen2)
```

If both names were omitted this switch_to function would not do anything.

matheussmoreira:
I am analyzing your suggestion.

benpayne:
Sounds great, let me know if you have questions.

matheussmoreira:
Can you send me the full python code?

matheussmoreira:
So, I've coded a ScreenManager creation with multiple Screens that have the same name in a different way. Really, when switching to a new screen, if it has the same name of another one this transition kinda does not happen as it should, so not-allowing repeated names seems to be a good solution, regardless of them be empty or not. I am sending a solution targeting ScreenManager's add_widget() method (not sure if it affects the problem originated in the KV file).

Issue #634, opened 2012-08-09
ScreenManager Screens can't be added in KV


hansent:
```
from kivy.app import App
from kivy.lang import Builder

KV = '''
<Screen>:
    Label:
        text: root.name

ScreenManager:
    Screen:
        name: "test"
    Screen:
        name: "another"
'''
class TestApp(App):
    def build(self):
        Builder.load_string(KV)

if __name__ == "__main__":
    TestApp().run()
```

gives the following error:

```
 Traceback (most recent call last):
   File "test.py", line 23, in <module>
     TestApp().run()
   File "/home/tehansen/code/kivy/kivy/app.py", line 522, in run
     root = self.build()
   File "test.py", line 20, in build
     Builder.load_string(KV)
   File "/home/tehansen/code/kivy/kivy/lang.py", line 1133, in load_string
     self._apply_rule(widget, parser.root, parser.root)
   File "/home/tehansen/code/kivy/kivy/lang.py", line 1262, in _apply_rule
     widget.add_widget(child)
   File "/home/tehansen/code/kivy/kivy/uix/screenmanager.py", line 573, in add_widget
     'Name %r already used' % screen.name)
 kivy.uix.screenmanager.ScreenManagerException: Name '' already used
```

So I think the problem is the Screen is instantiated by the builder and added to the screenmanager before name is set

PR #8260, opened 2023-05-29
Refactoring: ScreenManager should contain screens with unique names

matheussmoreira:
Adressing #8160.

Maintainer merge checklist


kivy/uix/screenmanager.py
```diff
@@ -980,8 +980,8 @@ def _screen_name_changed(self, screen, name):
 
     def add_widget(self, widget, *args, **kwargs):
         '''
-        .. versionchanged:: 2.1.0
-            Renamed argument `screen` to `widget`.
+        .. versionchanged:: 2.3.0
+            Raising exception for repeated screen name.
```
misl6, line 984: Hi @matheussmoreira !

The old `versionchanged` should be kept, instead of editing it, you should consider adding a new one.

misl6 (changes requested):
Hi @matheussmoreira !

This PR will re-introduce this bug: https://github.com/kivy/kivy/issues/634 which was fixed via https://github.com/kivy/kivy/pull/636 (which basically allowed multiple screens with the same name)

Isn't the warning enough?

kivy/uix/screenmanager.py
matheussmoreira: Ok, I will retrieve the old one.

matheussmoreira:
> Hi @matheussmoreira !
> 
> This PR will re-introduce this bug: #634 which was fixed via #636 (which basically allowed multiple screens with the same name)
> 
> Isn't the warning enough?

To make someone aware that the name repetition exists, in fact, it is enough. However, it seems to be better to relate this information more clearly with the undesired behavior of the screen switch ending up not occurring, otherwise confusions and complaints like the one raised on #8160 (created this March) will persist. I guess making the warning message a bit more descriptive, or adding a second warning, or even changing the screen name to a new one on `add_widget()` are satisfying proposals. What do you think?

matheussmoreira:
> > Hi @matheussmoreira !
> > This PR will re-introduce this bug: #634 which was fixed via #636 (which basically allowed multiple screens with the same name)
> > Isn't the warning enough?
> 
> To make someone aware that the name repetition exists, in fact, it is enough. However, it seems to be better to relate this information more clearly with the undesired behavior of the screen switch ending up not occurring, otherwise confusions and complaints like the one raised on #8160 (created this March) will persist. I guess making the warning message a bit more descriptive, or adding a second warning, or even changing the screen name to a new one on `add_widget()` are satisfying proposals. What do you think?

matheussmoreira:
> > Hi @matheussmoreira !
> > This PR will re-introduce this bug: #634 which was fixed via #636 (which basically allowed multiple screens with the same name)
> > Isn't the warning enough?
> 
> To make someone aware that the name repetition exists, in fact, it is enough. However, it seems to be better to relate this information more clearly with the undesired behavior of the screen switch ending up not occurring, otherwise confusions and complaints like the one raised on #8160 (created this March) will persist. I guess making the warning message a bit more descriptive, or adding a second warning, or even changing the screen name to a new one on `add_widget()` are satisfying proposals. What do you think?

PS: a screen name is already changed on `switch_to()` function in case of repetition (for those screens not yet present in the ScreenManager's `self.screens` list), but if I call `add_widget()` directly, we have this blind spot.

misl6:
> > Hi @matheussmoreira !
> > This PR will re-introduce this bug: #634 which was fixed via #636 (which basically allowed multiple screens with the same name)
> > Isn't the warning enough?
> 
> @misl6 thank you for making me aware of the bug and its fix that happened in the past.
> 
> To make someone aware that the name repetition exists, in fact, it is enough. However, it seems to be better to relate this information more clearly with the undesired behavior of the screen switch ending up not occurring, otherwise confusions and complaints like the one raised on #8160 (created this March) will persist. I guess making the warning message a bit more descriptive, or adding a second warning, or even changing the screen name to a new one on `add_widget()` are satisfying proposals. What do you think?

Maybe the second warning could be added when the user tries to switch to it?

matheussmoreira:
> > > Hi @matheussmoreira !
> > > This PR will re-introduce this bug: #634 which was fixed via #636 (which basically allowed multiple screens with the same name)
> > > Isn't the warning enough?
> > 
> > 
> > @misl6 thank you for making me aware of the bug and its fix that happened in the past.
> > To make someone aware that the name repetition exists, in fact, it is enough. However, it seems to be better to relate this information more clearly with the undesired behavior of the screen switch ending up not occurring, otherwise confusions and complaints like the one raised on #8160 (created this March) will persist. I guess making the warning message a bit more descriptive, or adding a second warning, or even changing the screen name to a new one on `add_widget()` are satisfying proposals. What do you think?
> 
> Maybe the second warning could be added when the user tries to switch to it?

Yes, it seems reasonable to me. I'll do it.

matheussmoreira:
Work done.

kivy/uix/screenmanager.py
```diff
@@ -1068,11 +1069,16 @@ def on_current(self, instance, value):
             screen.dispatch('on_pre_enter')
             screen.dispatch('on_enter')
 
+    def get_screens_with_name(self, name):
```
misl6, line 1072: We want to expose that and not to keep it for internal use?

How about `_get_screens_with_name` ?

misl6 (commented):
Hi @matheussmoreira !

I've left a comment + maybe we can add some tests so we can make sure that the warning is shown appropriately (now and in future)?

kivy/uix/screenmanager.py
matheussmoreira: We really are using this function internally only, so it is better the make the suggested change. About the tests, I'll do it in a few days!

matheussmoreira:
Hi @misl6! I added my test code, but I noticed that this one as well as those added before me get the error of _"App: Unable to get a Window, abort."_ when creating an instance of the ScreenManager class. Is the test module even able of doing this (having in mind that here we don't use any child of App)?

Issue #8177, opened 2023-03-19
Remove deprecated features


misl6:
The initial plan for this issue was to remove any deprecation `<=1.9.0`, but that was scheduled for `2.3.0`.

Since we entered in `3.0.0` milestone, we're now targeting any feature deprecated before or in `2.0.0`, and if is a game-changer, also features deprecated in `2.1.0`, `2.2.0`, `2.3.0` series.

For who is taking care of this issue, please add a detailed list of the removals as suggested by #8585 

Ref: #8167

Zerus97:
Hi, i would like to work on this issue

misl6:
Hi @Zerus97,

I've assigned the issue to you, please ping us when your PR is ready for a review.

Thank you!

Zerus97:
Hey @misl6 , how are you? Im having a hard time trying to find deprecated code and since this is my first time solving an issue i was hoping if you could help me with that. I really want to contribute with the projetct!

misl6:
Hi @Zerus97 !

The best way is to search for `.. deprecated::` into code 😀

Zerus97:
@misl6 Thanks, ill start looking into it!

misl6:
Deprecations yet to be removed:

- `VideoPlayer.play` https://github.com/kivy/kivy/blob/7e9affd6f21a15ded862f621e605d35f2cc83be5/kivy/uix/videoplayer.py#L347

- `Video.play` https://github.com/kivy/kivy/blob/7e9affd6f21a15ded862f621e605d35f2cc83be5/kivy/uix/video.py#L86

- `TextInput.padding_x` https://github.com/kivy/kivy/blob/7e9affd6f21a15ded862f621e605d35f2cc83be5/kivy/uix/textinput.py#L3325

- `TextInput.padding_y` https://github.com/kivy/kivy/blob/7e9affd6f21a15ded862f621e605d35f2cc83be5/kivy/uix/textinput.py#L3341

- `FileChooser.file_encodings` https://github.com/kivy/kivy/blob/7e9affd6f21a15ded862f621e605d35f2cc83be5/kivy/uix/filechooser.py#L486

- `KVLang Templates`: ⚠️ We may break some guides and examples, so we should be careful.

- `Rotate.set` https://github.com/kivy/kivy/blob/7e9affd6f21a15ded862f621e605d35f2cc83be5/kivy/graphics/context_instructions.pyx#L686

- `pygame` window provider: ⚠️ We really want to? We should be super careful?

[...]

PR #8270, opened 2023-06-03
Allow `Scale` and `Translate` to take `x`, `y` and `z` as `kwargs`. Removes deprecated `Scale().scale` property.

GreenJon902:
Added support to create scale and translate instructions with kwargs like this
```
Translate(x=0, y=1, z=2)
Translate(0, 1)  # Original method still works

Scale(x=2, y=1, z=23)  # Scale also works
Scale(1, 2, 3)  # But the original for this also works.
```
Unit tests were also added for both methods of creation.

P.S. If your want to know why I can't just use normal args: I have a system where I load designs from a JSON file which stores the type of instruction and kwargs (like pos and size), and it doesn't work with normal args.

Maintainer merge checklist

### DIFF
--- a/kivy/graphics/context_instructions.pyx
+++ b/kivy/graphics/context_instructions.pyx
@@ -901,13 +901,17 @@ cdef class Translate(Transform):
     '''
     def __init__(self, *args, **kwargs):
         cdef double x, y, z
-        Transform.__init__(self, **kwargs)
+        x, y, z = 0, 0, 0
         if len(args) == 3:
             x, y, z = args
-            self.set_translate(x, y, z)
         elif len(args) == 2:
             x, y = args
-            self.set_translate(x, y, 0)
+        x = kwargs.pop("x", x)
+        y = kwargs.pop("y", y)
+        z = kwargs.pop("z", z)
+
+        Transform.__init__(self, **kwargs)
+        self.set_translate(x, y, z)
 
     cdef set_translate(self, double x, double y, double z):
         self.matrix = Matrix().translate(x, y, z)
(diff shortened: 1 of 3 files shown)


kivy/graphics/context_instructions.pyx
```diff
@@ -764,9 +764,25 @@ cdef class Scale(Transform):
     .. deprecated:: 1.6.0
         Deprecated single scale property in favor of x, y, z, xyz axis
         independent scaled factors.
```
misl6, line 766: Considering https://github.com/kivy/kivy/issues/8177, I guess we can just remove the single scale property.

kivy/graphics/context_instructions.pyx
```diff
@@ -764,9 +764,25 @@ cdef class Scale(Transform):
...
         cdef double x, y, z
+
+        x, y, z = 1.0, 1.0, 1.0
+        if len(args) == 1:
+            s = args[0]
+            x, y, z = s, s, s
+        if len(args) == 2:
+            x, y = args
+        elif len(args) == 3:
+            x, y, z = args
```
misl6, line 781: Referring to my previous comment, I think we should only consider the `len(args) == 3` option.
Also, by removing the deprecated single-scale property, we can likely clean up the code a lot. 😀

misl6 (changes requested):
Hi @GreenJon902 !

I've left some comments regarding your PR, as there's the chance to do some clean-up 😀

kivy/graphics/context_instructions.pyx
GreenJon902: I have done this, and replaced the deprecation notice with a versionchanged message.

misl6 (approved):
LGTM. Thank you!

PR #8271, opened 2023-06-03
Fix `SmoothLine` rendering issues

DexerBR:
This PR fixes several rendering issues with the `SmoothLine` when drawing fixed or freehand shapes (as shown in the demonstration videos below). There is also code included in this PR, with some sets of points for testing. These points had many rendering issues (which are now resolved).

**Some issues and fixes applied:**

- Incorrect width rendering: 
  - When the angle between two line segments was very small, the line width value at the intersection was extrapolated, causing serious line rendering issues. The solution was to use fixed width values when this situation is identified.
  - If the distance between the points was very small, incorrect angle values were calculated. The solution was to remove points that were within a distance of less than 0.1 from each other (I believe it is insignificant to have points with a distance smaller than that - in the context of kivy graphics).

- Line closing: The generated closing line was distorted in several cases. The approach adopted to fix it was to add an extra point and perform the calculations of final and initial angles based on that point.

**In general, these approaches make the rendering of `SmoothLine` significantly more reliable and stable!**

### Before:

image

### After:

image

### Before:

https://github.com/kivy/kivy/assets/73297572/9d35291e-86c3-4666-9d62-e96402cc3df6

### After:

https://github.com/kivy/kivy/assets/73297572/8a8baba4-0458-41d8-b6cf-97ea08cef2a6

### Test code:

```python
from kivy.lang import Builder
from kivy.app import App

Builder.load_string(
    """
<UI>:
    points: []
    canvas.before:
        Color:
            rgba: 0.9, 0.9, 0.9, 1
        Rectangle:
            size: self.size
            pos: self.pos

        Color:
            rgba: 0, 0, 1, 1
        SmoothLine:
            overdraw_width: 2.5
            width: 1
            # close: 1

            points: root.points
            # points: [234.86195373535156, 400.0, 247.36195373535156, 412.5, 247.3619384765625, 412.5, 259.86195373535156, 400.0, 259.8619384765625, 400.0, 247.3619384765625, 387.5, 247.36195373535156, 387.5, 234.86195373535156, 400.0]
            # points:  [300, 300, 300, 300, 50, 160, 100, 200]
            # points:  [300, 300, 300, 300, 100, 200]
            # points:  [300, 300, 300, 250, 100, 200]
            # points:  [300, 300, 300, 280, 100, 200]
            # points:  [300, 300, 300, 200, 100, 200]
            # points:  [100, 100, 100, 200, 100, 200]
            # points:  [300, 300, 300, 200, 100, 200, 200, 300]
            # points:  [100, 230, 300, 200, 100, 200]
            # points: [150, 30, 200, 100, 300, 300]


    on_touch_move:
        self.points.extend(args[1].pos)

    # on_touch_down:
        # self.points.extend(args[1].pos)

"""
)


class UI(FloatLayout):
    ...


class TestSmoothLine(App):
    def build(self):
        return UI()


TestSmoothLine().run()
```

Maintainer merge checklist

### DIFF
--- a/kivy/graphics/vertex_instructions_line.pxi
+++ b/kivy/graphics/vertex_instructions_line.pxi
@@ -1481,19 +1481,37 @@ cdef class SmoothLine(Line):
         return 0
 
     # FIXME: Some artifacts can be observed, depending on the line width,
-    # overdraw_width and radius.
+    # overdraw_width and radius. This occurs due to the way the vertices are
+    # built. It is not noticeable when the alpha value of the color of the
+    # active context is equal to 1. One solution would be to avoid overlapping
+    # vertices.
     cdef void build_smooth(self):
         cdef:
             list p = self.points
+            int close = self._close
             double width = max(0, (self._width - 1.))
             double owidth = width + self._owidth
             vertex_t *vertices = NULL
             unsigned short *indices = NULL
             unsigned short *tindices = NULL
+            double min_angle_threshold, angle_diff
             double ax, ay, bx = 0., by = 0., rx = 0., ry = 0., last_angle = 0., angle, av_angle
             double cos1, sin1, cos2, sin2, ocos1, ocos2, osin1, osin2
             long index, icount, iv, ii, max_vindex, count
             unsigned short i0, i1, i2, i3, i4, i5, i6, i7, vindex, vcount
+
+
+        # Points that are very close (with a distance less than 0.1) will
+        # be discarded. This increases the reliability of line rendering.
+        p = self._remove_close_points(p)
+
+        # If it is just a line segment, there will be no support to close the line.
+        if len(p) <= 4:
+            close = False
+
+        # A new point needs to meet a minimum distance threshold before being added.
+        if close and not (abs(p[-2] - p[0]) < 0.1 and abs(p[-1] - p[1]) < 0.1):
+            p = p + p[:2]
 
         iv = vindex = 0
         count = <long>int(len(p) / 2.)
@@ -1503,8 +1521,6 @@ cdef class SmoothLine(Line):
 
         vcount = <unsigned short>(count * 4)
         icount = (count - 1) * 18
-        if self._close and self._close_mode == 'straight-line':
-            icount += 18
 
         vertices = <vertex_t *>malloc(vcount * sizeof(vertex_t))
         if vertices == NULL:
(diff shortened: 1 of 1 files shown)


misl6:
How about introducing some tests that exercise this issue?

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -1481,20 +1481,38 @@ cdef class SmoothLine(Line):
...
     # FIXME: Some artifacts can be observed, depending on the line width,
-    # overdraw_width and radius.
+    # overdraw_width and radius. This occurs due to the way the vertices are
+    # built. It is not noticeable when the alpha value of the color of the
+    # active context is equal to 1. One solution would be to avoid overlapping
+    # vertices.
     cdef void build_smooth(self):
         cdef:
             list p = self.points
+            int close = self._close
```
misl6, line 1491: Minor:
Instead of using `close` as a variable name, what about using `can_close`, so it's clearer the intent during the code below?

kivy/graphics/vertex_instructions_line.pxi
```diff
@@ -1680,43 +1710,41 @@ cdef class SmoothLine(Line):
...
         free(indices)
+    
+    cdef list _remove_close_points(self, list p):
+        cdef int index = 0
+        cdef double x1, y1, x2, y2
+
+        while index < len(p) - 2:
+            x1, y1 = p[index], p[index + 1]
+            x2, y2 = p[index + 2], p[index + 3]
+            if abs(x2 - x1) < 0.1 and abs(y2 - y1) < 0.1:
```
misl6, line 1727: Same here for the `0.1` (maybe we can add an argument), so that may be configurable in future.

misl6 (commented):
Since #8309 makes an extensive usage of `SmoothLine`, and this PR is still in a pending state, I feel is better to merge this one before.

I've left some minor comments, can you share your thoughts?

kivy/graphics/vertex_instructions_line.pxi
DexerBR: Maybe `must_close_line`, could be a better name?

DexerBR:
> Since #8309 makes an extensive usage of `SmoothLine`, and this PR is still in a pending state, I feel is better to merge this one before.
> 
> I've left some minor comments, can you share your thoughts?

Actually, the other PR is not related to this one, as it uses its own implementation (`AntialiasingLine`), which is independent of `SmoothLine`. But it will be good to have this PR merged! 🙂

misl6 (approved):
Still not optimal, as you said in `FIXME` note, but at least is a step forward.

LGTM. Thank you!

Issue #8280, opened 2023-06-12
_migrate_deprecated_padding_xy causing app to not load even when no padding is ever used within the app


JacekJagosz:
**Software Versions**
* Python: 3.10 and 3.9
* OS: Solus 4.3 and Raspberry Pi OS Bullseye
* Kivy: 2.2.0
* Kivy installation method: pip wheel for both devices

**Describe the bug**
My app that was created using Kivy 2.1.0 will crash at startup when trying to run with Kivy 2.2.0.
I am getting an error attached below, with function `_migrate_deprecated_padding_xy`, even though none of the lables in my code use padding.
The problematic line is https://github.com/DexerBR/kivy/blob/master/kivy/core/text/__init__.py#L331 introduced in pull #8162. If the replace `self.options['padding'] = list(self.options['padding'])` with `self.options['padding'] = [0, 0, 0, 0]` and the app launches and works just fine.
Could it be that if no padding is specified the default value of [0, 0, 0, 0] is not set early enough and this is why `self.options['padding']` is `NoneType`?

**Expected behavior**
If no padding is specified for a label, it should deafult to [0, 0, 0, 0] instead of crashing.

**To Reproduce**
I have no idea which part of the code causes this issue, as none of labels use padding.

**Code and Logs and screenshots**
```python
[INFO   ] [Kivy        ] v2.2.0
[INFO   ] [Kivy        ] Installed at "/home/jacek/.local/lib/python3.10/site-packages/kivy/__init__.py"
[INFO   ] [Python      ] v3.10.11 (main, Apr 27 2023, 08:33:53) [GCC 12.2.0]
[INFO   ] [Python      ] Interpreter at "/usr/bin/python3"
[INFO   ] [Logger      ] Purge log fired. Processing...
[INFO   ] [Logger      ] Purge finished!
[INFO   ] [Factory     ] 190 symbols loaded
[DEBUG  ] [Cache       ] register <kv.resourcefind> with limit=None, timeout=60
[DEBUG  ] [Cache       ] register <kv.lang> with limit=None, timeout=None
[TRACE  ] [Lang        ] load file /home/jacek/.local/lib/python3.10/site-packages/kivy/data/style.kv, using utf8 encoding
[TRACE  ] [Parser      ] parsing 1341 lines
[TRACE  ] [Parser      ] got directive <kivy 1.0>
[TRACE  ] [Parser      ] got directive <import kivy kivy>
[TRACE  ] [Builder     ] build rule for <Label>
[...]

PR #8291, opened 2023-06-21
Handle edge case when padding is None in _migrate_deprecated_padding_xy

JacekJagosz:
Fixes https://github.com/kivy/kivy/issues/8280

Maintainer merge checklist


JacekJagosz:
This is a regression with Kivy 2.2 from 2.1, where some code using KivyMD will fail with very indescriptive error.
This makes the code work just as it did with 2.1.

PR #8296, opened 2023-06-24
[SDL2] Window.opacity feature

mak8kammerer:
Hello everyone!

Today I want to add opacity feature to window.
Based on this on this PR: #[5169](https://github.com/kivy/kivy/pull/5169).

Maintainer merge checklist

### DIFF
--- a/examples/miscellaneous/opacitywindow.py
+++ b/examples/miscellaneous/opacitywindow.py
@@ -1,0 +1,25 @@
+from kivy.app import App
+from kivy.uix.slider import Slider
+from kivy.lang import Builder
+
+Builder.load_string('''
+#:import window kivy.core.window.Window
+<OpacitySlider>:
+    min: 0.0
+    max: 1.0
+    value: window.opacity
+    on_value: window.opacity = args[1]
+''')
+
+
+class OpacitySlider(Slider):
+    pass
+
+class OpacityWindowApp(App):
+
+    def build(self):
+        return OpacitySlider()
+
+
+if __name__ == '__main__':
+    OpacityWindowApp().run()
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -933,6 +933,39 @@ class WindowBase(EventDispatcher):
     the position set in :class:`~kivy.config.Config`.
     '''
 
+    def _get_opacity(self):
+        return self._get_window_opacity()
+
+    def _set_opacity(self, opacity):
+        if opacity < 1.0 or opacity > 0.0: 
+            self._set_window_opacity(opacity)
+        else:
+            Logger.warning('Window: The opacity value of '
+                           'the window should be in the '
+                           'range from 0.0 to 1.0.'
+                           )
+
+    def _get_window_opacity(self):
+        Logger.warning('Window: opacity is not implemented'
+                       ' in the current window provider.')
+
+    def _set_window_opacity(self, opacity):
+        Logger.warning('Window: opacity is not implemented'
+                       ' in the current window provider.')
+
+    opacity = AliasProperty(_get_opacity, _set_opacity)
+    '''Opacity of the window. Accepts a value between 0.0 (transparent) and
+    1.0 (opaque).
+
+    .. note::
+        This feature requires the SDL2 window provider.
+
+    .. versionadded:: 2.3.0
+
+    :attr:`opacity` is an :class:`~kivy.properties.AliasProperty` and defaults
+    to `1.0`.
+    '''
+
     @property
     def __self__(self):
         return self
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -394,6 +394,14 @@ cdef class _WindowSDL2Storage:
     def set_window_pos(self, x, y):
         SDL_SetWindowPosition(self.win, x, y)
 
+    def set_window_opacity(self, opacity):
+        SDL_SetWindowOpacity(self.win, opacity)
+
+    def get_window_opacity(self):
+        cdef float opacity
+        SDL_GetWindowOpacity(self.win, &opacity)
+        return opacity
+
     def get_window_info(self):
         cdef SDL_SysWMinfo wm_info
         SDL_GetVersion(&wm_info.version)
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -489,6 +489,12 @@ class WindowSDL(WindowBase):
     def _set_window_pos(self, x, y):
         self._win.set_window_pos(x, y)
 
+    def _get_window_opacity(self):
+        return self._win.get_window_opacity()
+
+    def _set_window_opacity(self, opacity):
+        self._win.set_window_opacity(opacity)
+
     # Transparent Window background
     def _is_shaped(self):
         return self._win.is_window_shaped()
--- a/kivy/lib/sdl2.pxi
+++ b/kivy/lib/sdl2.pxi
@@ -924,6 +924,10 @@ cdef extern from "SDL_audio.h":
     )
     cdef int SDL_ConvertAudio(SDL_AudioCVT *cvt)
 
+cdef extern from "SDL_video.h":
+    cdef int SDL_SetWindowOpacity(SDL_Window * window, float opacity)
+    cdef int SDL_GetWindowOpacity(SDL_Window * window, float* opacity)
+
 cdef extern from "SDL_mixer.h":
     cdef struct Mix_Chunk:
         int allocated


misl6:
Nice! How about adding some tests to our suite that check everything works as expected?

"Everything works as expected":
- `opacity` property changes 
- `opacity < 1.0 or opacity > 0.` check is performed and warning log is dispatched.

I do not expect you to test that the window is respecting the opacity set via `SDL_SetWindowOpacity`, as I expect this test has already been performed on SDL side. 

FYI: We're trying to encourage contributors to add tests in `2.3.0`, as an effort to make Kivy more stable and well-tested (so we can avoid potential issues in future)

mak8kammerer:
OK, I'll add the tests.

kivy/tests/test_window_opacity.py
```diff
@@ -0,0 +1,25 @@
...
+        self.assertEqual(LoggerHistory.history[0].msg, 
+                         'Window: The opacity value of '
+                         'the window should be in the '
+                         'range from 0.0 to 1.0.')
+        
+        LoggerHistory.clear_history()
+        
+        Window.opacity = -15
+        assert len(LoggerHistory.history) == 1
+
```
misl6, line 25: Hi!
It likely makes sense to have this test into `test_window_base.py` instead of using a specific file.
What do you think about it?

mak8kammerer: Ok, no problems

mak8kammerer: Done

misl6 (changes requested):
Hi @mak8kammerer !

The PEP8 checks are failing, can you take care of it?

misl6:
Ouch! Are failing again 😅

mak8kammerer:
I see, I'll figure it out now.

mak8kammerer:
@misl6, can you run checks again?

kivy/tests/test_window_base.py
```diff
@@ -18,3 +18,26 @@ def test_to_normalized_pos(self):
                 assert result_sy == expected_sy
         finally:
             win.system_size = old_system_size
+
+
+class WindowOpacityTest(GraphicUnitTest):
+
+    def test_window_opacity_property(self):
```
pythonic64, line 25: This method can be moved into `WindowBaseTest` as adding new class `WindowOpacityTest` doesn't seem necessary for just one method.

kivy/tests/test_window_base.py
```diff
@@ -18,3 +18,26 @@ def test_to_normalized_pos(self):
...
+        from kivy.core.window import Window
```
pythonic64, line 26: You can use `self.Window` to access window instance, so no need for import statement.

kivy/tests/test_window_base.py
```diff
@@ -18,3 +18,26 @@ def test_to_normalized_pos(self):
...
+
+        Window.opacity = 15
+        self.assertEqual(LoggerHistory.history[0].msg,
+                         'Window: The opacity value of '
+                         'the window should be in the '
+                         'range from 0.0 to 1.0.')
+
+        LoggerHistory.clear_history()
+
+        Window.opacity = -15
```
pythonic64, line 42: Original `Window.opacity` should be stored/restored in `setUp`/`tearDown` methods. 

```python
def setUp(self):
    super().setUp()
    self._prev_window_opacity = self.Window.opacity

def tearDown(self):
    self.Window.opacity = self._prev_window_opacity
    super().tearDown()
```

kivy/core/window/_window_sdl2.pyx
```diff
@@ -394,6 +394,14 @@ cdef class _WindowSDL2Storage:
     def set_window_pos(self, x, y):
         SDL_SetWindowPosition(self.win, x, y)
 
+    def set_window_opacity(self, opacity):
+        SDL_SetWindowOpacity(self.win, opacity)
```
pythonic64, line 398: From SDL2 docs https://wiki.libsdl.org/SDL2/SDL_SetWindowOpacity function `SDL_SetWindowOpacity` returns `-1` if opacity is not supported or (another) negative value in case of an error. We should use `SDL_GetError` to get the error message and log it.

kivy/core/window/__init__.py
```diff
@@ -933,6 +933,39 @@ def _set_top(self, value):
     the position set in :class:`~kivy.config.Config`.
     '''
 
+    def _get_opacity(self):
+        return self._get_window_opacity()
+
+    def _set_opacity(self, opacity):
+        if opacity > 0.0 and opacity < 1.0:
```
pythonic64, line 940: SDL2 will clamp the value if less than 0 or greater than 1, so I would remove range testing from this method (and logger warning) and let specific window implementation decide on what to do. Since SDL2 will clamp there is no need to range test in `WindowSDL`, but just test for errors https://github.com/kivy/kivy/pull/8296/files#r1248874160

pythonic64: @misl6 Do you agree?

pythonic64:
Regarding tests:
I suggest using `setUp`/`tearDown` to store/restore previous opacity value and `LoggerHistory.history` list and in that case we need a new class, so @mak8kammerer we should restore `WindowOpacityTest` and test for these cases:
1) test if opacity is supported -> no error log on opacity change
2) (if supported) test values within [0, 1] range
3) (if supported) test if values are clamped

kivy/core/window/__init__.py
misl6: @pythonic64 thanks for chiming in! ❤️

I partially agree:
- Testing for errors (and logging it) via `SDL_GetError` is absolutely something we should have. (As for any reason, even if the value is in the supported range, something bad may happen, and the user/developer needs to be informed)

- Defining a supported range on `WindowBase` looks like a good approach instead, as will force us in the future to standardize the opacity range along all the window providers. (the provider-specific implementation will take care of translating the `0.0 - 1.0` range to a supported value). By not defining a supported range, the risk is that in the future we may introduce a window provider with `0-100` range, which is incompatible with the `0.0-1.0` range.
By defining a supported range, Kivy is responsible to maintain compatibility by applying the needed translations.

misl6: > I agree on standardizing opacity range to 0.0 - 1.0 but imho that should be (and in this pull request already is) stated only in the docs. Similarly, we don't do range test for each component in ColorProperty, but we do state supported range in the docs. If in the future we have another window provider we will still respect the defined range from the docs.

Yep. Makes sense!

@pythonic64 and @mak8kammerer feel free to ping me when you need a final review or to run tests (as @mak8kammerer is a new contributor, and workflow runs need to be manually approved)

kivy/tests/test_window_base.py
pythonic64: @mak8kammerer To answer your Discord question:

Methods `setUp`/`tearDown` should be added in `WindowOpacityTest` class.

kivy/core/window/__init__.py
pythonic64: @mak8kammerer SDL will clamp the value, that's stated in the docs, "Remarks" section: https://wiki.libsdl.org/SDL2/SDL_SetWindowOpacity#remarks .

In our discussion above we agree that range should be specified in the docs only as we don't expect users assigning values outside of that range. Since SDL is clamping the value that's fine, but if in the future we have another window provider which doesn't clamp the value but raises an error then we will decide in that window specific implementation to either:
1. clamp and set the clamped value
2. catch and log the error, leaving the current opacity value unchanged

`Widget.opacity` is an another example where we don't do range check and assigning values outside of `[0.0-1.0]` does not raise an error.

kivy/core/window/__init__.py
```diff
@@ -933,6 +933,33 @@ def _set_top(self, value):
...
+
+    def _set_opacity(self, opacity):
+        self._set_window_opacity(opacity)
+
+    def _get_window_opacity(self):
+        Logger.warning('Window: opacity is not implemented'
+                       ' in the current window provider.')
+
+    def _set_window_opacity(self, opacity):
+        Logger.warning('Window: opacity is not implemented'
```
pythonic64, line 947: Maximize available width for the code:
```python
Logger.warning('Window: Opacity is not implemented in the current '
               'window provider')
```

Do the same in the `_get_window_opacity` method.

pythonic64 (changes requested):

kivy/core/window/_window_sdl2.pyx
```diff
@@ -394,6 +394,19 @@ cdef class _WindowSDL2Storage:
     def set_window_pos(self, x, y):
         SDL_SetWindowPosition(self.win, x, y)
 
+    def set_window_opacity(self, opacity):
+        result = SDL_SetWindowOpacity(self.win, opacity)
+        if result == -1:
```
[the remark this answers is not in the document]
pythonic64: I've updated the comment above, removed unused `error` variable.

kivy/lib/sdl2.pxi
```diff
@@ -924,6 +924,10 @@ cdef extern from "SDL_audio.h":
     )
     cdef int SDL_ConvertAudio(SDL_AudioCVT *cvt)
 
+cdef extern from "SDL_video.h":
+    cdef int SDL_SetWindowOpacity(SDL_Window * window, float opacity)
```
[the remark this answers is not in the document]
mak8kammerer: Ok, thanks

mak8kammerer:
@misl6, work done.

pythonic64 (changes requested):

kivy/core/window/_window_sdl2.pyx
```diff
@@ -394,6 +394,20 @@ cdef class _WindowSDL2Storage:
     def set_window_pos(self, x, y):
         SDL_SetWindowPosition(self.win, x, y)
 
+    def set_window_opacity(self, opacity):
+        if SDL_SetWindowOpacity(self.win, opacity):
+            message = (<bytes>SDL_GetError()).decode('utf-8', 'replace')
+            Logger.error(f'WindowSDL: Setting opacity failed - {message}')
```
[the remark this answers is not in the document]
mak8kammerer: No problems. Thank You!

pythonic64:
@mak8kammerer I think we missed dispatching of `opacity` on the value change, meaning callbacks like `window.bind(opacity=on_opacity_change)` won't be called.

Can you add `Label` in the example app to show opacity value from callback function and report if it works? It most-likely won't so we will need more code changes to cover that case.

mak8kammerer:
@pythonic64, done. Please, check.

examples/miscellaneous/opacitywindow.py
```diff
@@ -0,0 +1,32 @@
+from kivy.app import App
+from kivy.uix.boxlayout import BoxLayout
+from kivy.lang import Builder
+
+Builder.load_string('''
+#:import window kivy.core.window.Window
```
pythonic64, line 6: Global variable `win` should be available in kv code and if so then this import is not needed and you can use `win` instead of `window` for the rest of the kv code.

kivy/core/window/__init__.py
```diff
@@ -933,6 +933,34 @@ def _set_top(self, value):
     the position set in :class:`~kivy.config.Config`.
     '''
 
+    def _get_opacity(self):
+        return self._get_window_opacity()
+
+    def _set_opacity(self, opacity):
+        self._set_window_opacity(opacity)
+        return True
```
pythonic64, line 941: Always returning `True` is not a good idea as assigning the same value will always dispatch change events. For example, assigning 5 times value of `0.2` will dispatch change event 5 times but it should dispatch it only once.

I'll post suggestions on how to handle this case.

kivy/core/window/__init__.py
```diff
@@ -933,6 +933,34 @@ def _set_top(self, value):
     the position set in :class:`~kivy.config.Config`.
     '''
 
+    def _get_opacity(self):
+        return self._get_window_opacity()
+
+    def _set_opacity(self, opacity):
+        self._set_window_opacity(opacity)
```
pythonic64, line 940: Change to `return self._set_window_opacity(opacity)` so that window specific implementation can decide what to do.

kivy/core/window/__init__.py
```diff
@@ -933,6 +933,34 @@ def _set_top(self, value):
...
+
+    def _get_window_opacity(self):
+        Logger.warning('Window: Opacity is not implemented in the current '
+                       'window provider')
+
+    def _set_window_opacity(self, opacity):
+        Logger.warning('Window: Opacity is not implemented in the current '
+                       'window provider')
+
+    opacity = AliasProperty(_get_opacity, _set_opacity)
```
pythonic64, line 951: Change to `opacity = AliasProperty(_get_opacity, _set_opacity, cache=True)` so that opacity value is cached so that cached value inside of `AliasProperty` will be returned instead of calling getter method.

kivy/core/window/_window_sdl2.pyx
```diff
@@ -394,6 +394,21 @@ cdef class _WindowSDL2Storage:
     def set_window_pos(self, x, y):
         SDL_SetWindowPosition(self.win, x, y)
 
+    def set_window_opacity(self, opacity):
+        if SDL_SetWindowOpacity(self.win, opacity):
+            message = (<bytes>SDL_GetError()).decode('utf-8', 'replace')
+            Logger.error(f'WindowSDL: Setting opacity to {opacity} failed - '
+                         f'{message}')
+
```
pythonic64, line 402: Add `return False` in case of an error and `return True` for success. Append on line 402:
```python
    return False
return True
```

kivy/core/window/window_sdl2.py
```diff
@@ -489,6 +489,12 @@ def _get_window_pos(self):
     def _set_window_pos(self, x, y):
         self._win.set_window_pos(x, y)
 
+    def _get_window_opacity(self):
+        return self._win.get_window_opacity()
+
+    def _set_window_opacity(self, opacity):
+        self._win.set_window_opacity(opacity)
```
pythonic64, line 496: Change to:
```python
if self.opacity != opacity:
    return self._win.set_window_opacity(opacity)
```

This way `self.opacity` will use the cached value and if `True` is returned by `set_window_opacity` opacity getter method will be called to get new value which be cached.

kivy/tests/test_window_base.py
```diff
@@ -18,3 +19,37 @@ def test_to_normalized_pos(self):
...
+        self._prev_history = LoggerHistory.history[:]
+
+    def tearDown(self):
+        self.Window.opacity = self._prev_window_opacity
+        LoggerHistory.history[:] = self._prev_history
+        super().tearDown()
+
+    def check_opacity_support(self):
+        LoggerHistory.clear_history()
+        self.Window.opacity = 0.2
```
pythonic64, line 38: Won't work for the case when the current opacity value is set to `0.2`. You must ensure that you set the different value from the current one. The same goes for `test_window_opacity_property` method.

pythonic64 (changes requested):

examples/miscellaneous/opacitywindow.py
mak8kammerer: We cannot use `win` variable. I get an error:
`NameError: name 'win' is not defined`

kivy/core/window/__init__.py
mak8kammerer: Done.

kivy/core/window/__init__.py
mak8kammerer: Done.

kivy/core/window/_window_sdl2.pyx
mak8kammerer: Done.

kivy/core/window/window_sdl2.py
mak8kammerer: Done.

kivy/tests/test_window_base.py
pythonic64: @mak8kammerer Can add changes for the case above? With those changes in-place this pull request is pretty much done.

mak8kammerer: Done. Please check and let me know.

examples/miscellaneous/opacitywindow.py
```diff
@@ -0,0 +1,32 @@
...
+    Slider:
+        size_hint_y: 4
+        min: 0.0
+        max: 1.0
+        value: window.opacity
+        on_value: window.opacity = args[1]
+''')
+
+
+class Layout(BoxLayout):
```
pythonic64, line 21: `Layout` is a class in `kivy.uix.layout` and we should not re-define it and moreover we don't have to use it. Example can be written like this:
```python
from kivy.app import App
from kivy.lang import Builder

kv = '''
#:import window kivy.core.window.Window

BoxLayout:
    orientation: 'vertical'
    Label:
        text: f'Window opacity: {window.opacity}'
        font_size: 25
    Slider:
        size_hint_y: 4
        min: 0.0
        max: 1.0
        value: window.opacity
        on_value: window.opacity = args[1]
'''


class WindowOpacityApp(App):

    def build(self):


if __name__ == '__main__':
    WindowOpacityApp().run()
```

pythonic64:
@mak8kammerer Can you add more info in the description on how the opacity feature is implemented (eg. `opacity` is a property, cached, dispatch value on change)? Also, I would suggest changing the title to "WindowSDL: Add opacity feature".

mak8kammerer:
@pythonic64, work done.

pythonic64:
@mak8kammerer I meant in the description of this pull request https://github.com/kivy/kivy/pull/8296#issue-1772676144 . Current description of the property is fine and we don't have to add anything to it.

Also, did you missed this comment https://github.com/kivy/kivy/pull/8296#discussion_r1341655060?

examples/miscellaneous/opacitywindow.py
mak8kammerer: Done.

examples/miscellaneous/opacitywindow.py
```diff
@@ -0,0 +1,28 @@
...
+from kivy.lang import Builder
+
+kv = '''
+#:import window kivy.core.window.Window
+
+BoxLayout:
+    orientation: 'vertical'
+    Label:
+        text: f'Window opacity: {window.opacity}'
+        font_size: 25
```
pythonic64, line 11: Add `sp` units to font size. Change this line to: `font_size: '25sp'`.

mak8kammerer: Done.

pythonic64 (dismissed):
Request looks good to me.

@misl6 Can you approve and merge?

kivy/core/window/__init__.py
```diff
@@ -933,6 +933,33 @@ def _set_top(self, value):
     the position set in :class:`~kivy.config.Config`.
     '''
 
+    def _get_opacity(self):
+        return self._get_window_opacity()
+
+    def _set_opacity(self, opacity):
+        return self._win.set_window_opacity(opacity)
```
misl6, line 940: @pythonic64 and @mak8kammerer , here `return self._win.set_window_opacity(opacity)` should not be `self._set_window_opacity(opacity)` ?

mak8kammerer: @misl6, look [this](https://github.com/kivy/kivy/pull/8296/files/0da1eabd3f922185c8a9c85eba697532062dd276#r1303461335)

misl6: Seems that @pythonic64 was suggesting to change it to `return self._set_window_opacity(opacity)` (which is correct), and not `return self._win.set_window_opacity(opacity)` (at the link you provided)

mak8kammerer: Sorry, my blame.

mak8kammerer: @misl6, fixed! Thank You!

mak8kammerer:
@misl6, there was clearly something wrong with me. Yesterday, while editing, I lost `return`.

kivy/core/window/__init__.py
pythonic64: @misl6 Correct, that line should be `return self._set_window_opacity(opacity)`

pythonic64 (approved):

misl6 (approved):
LGTM. Thank you!

pythonic64:
@mak8kammerer Thank you for getting this feature done.

@misl6 I think you can close https://github.com/kivy/kivy/pull/5169.

PR #8308, opened 2023-07-01
Copy edits: Standardize on spelling

Julian-O:
Changed comments (only) to use American spellings of common words.

(As an Australian, it hurts me to use the American spelling, but it looks more professional to have one standard throughout, and the code itself uses American spellings, so I kept it consistent.)

Also trivial copy-edits to comments in RecycleView.

Maintainer merge checklist


Julian-O:
There were two locations in the *code* I didn't touch.

`kivy/lib/vidcore_lite/egl.pyx` defines an `Initialise()` function (which calls an `initialize` function).

`kivy\modules\_webdebugger` includes some JavaScript that checks for a value called "colour". I don't understand the code to know if that is right or wrong, so I assume it is right.

misl6 (approved):
LGTM. Thank you!

PR #8309, opened 2023-07-02
Add graphics with antialiasing: `SmoothRectangle`, `SmoothEllipse`, `SmoothRoundedRectangle`, `SmoothQuad` and `SmoothTriangle`

DexerBR:
This PR adds a new graphics section with antialiasing. In addition, some classes have been adjusted to expose the points used in their build (necessary for the antialiasing approach).

Similar to what we have with `SmoothLine`, the new introduced graphics will be:

- `SmoothRectangle`
- `SmoothRoundedRectangle`
- `SmoothEllipse`
- `SmoothQuad`
- `SmoothTriangle`

Although there was a desire to integrate antialiasing directly into the current graphics, it seemed better to maintain support for figures without antialiasing and add the new instructions as `Smooth___`.

🟡 All support for antialiasing is maintained regardless of the opacity of the context in which the graphic is drawn.
🟡 **There is still no support for texture antialiasing.**

---

Antialiasing is achieved through a combination of a graphic without antialiasing and `AntiAliasingLine` (a new instruction used internally only for the purpose of antialiasing).

`AntiAliasingLine` was created and designed to fit around a graphic without antialiasing, providing a smoothing effect to its edges. Although a similar effect could be achieved with `SmoothLine`, `AntiAliasingLine` has some advantages over `SmoothLine`, such as:

- Keeps the antialiasing implementation isolated, avoiding breaking changes when depending on SmoothLine.
- It is significantly more stable as it uses miter joints and has a point filter to ensure consistent behavior. Its point filter also discards irrelevant points, which can improve performance in some cases.
- Some properties, such as `width` and premultiplied texture, were specifically defined through extensive experimentation for the purpose of antialiasing.
- It has the ability to perform stencil operations based on the **target graphic** to avoid alpha channel overlap at the intersection between the target graphic and `AntiAliasingLine`.

🟡 ~~It is worth noting that a customized version of stencil instructions is used here~~. The reason for this is that the function `glClear(GL_STENCIL_BUFFER_BIT)` is somewhat expensive when it comes to constant updates, which can significantly increase GPU usage. However, by limiting the call to clear stencil functions only when an update is actually needed, performance increases significantly; otherwise, GPU consumption would be close to the maximum limit (when there are tens of thousands of instructions). **During the tests, no side effects were observed.**
  - _Perhaps this is an improvement that could be implemented in `stencil_instructions.pyx`, and although it requires more testing, the performance improvement would be significant!_

### Depends on:
- https://github.com/kivy/kivy/pull/8405

Here are some basic comparison data from a test performed with [canvas_stress.py](https://github.com/kivy/kivy/blob/master/examples/canvas/canvas_stress.py), using `64000` `RoundedRectangle`/`SmoothRoundedRectangle`:

Using `RoundedRectangle`:

- CPU ~20%
- GPU ~32%
- MEMORY ~1100MB

Using `SmoothRoundedRectangle`:

- CPU ~20%
- GPU ~38%
- MEMORY ~1800MB

The data was obtained by observing the resource monitor while resizing the window or dragging touch across the window using the multitouch simulator (red dot created by the right mouse button).

**In almost all cases, the difference between using an instruction with antialiasing and one without would be very small (or insignificant) in terms of resource consumption.**

---

### Some comparisons between graphics with and without antialiasing:

aa1

aa2

aa3

aa4

aa5

aa6

aa7

**Test code:**
```python
from kivy.app import App
from kivy.lang import Builder

KV = """
FloatLayout:
    # on_touch_down:
    #     self.export_to_png("img.png")
    canvas:
        Color:
            rgba: 0, 0, 0, 1
        Rectangle:
            size: self.size
            pos: self.pos

        Color:
            rgba: 0, 0, 1, 1
        Rectangle:
            size: self.width, 150
            pos: 0, 250

        Color:
            rgba: 1, 1, 1, 0.5

        SmoothQuad:
            points: [50, 350, 200, 375, 175, 500, 150, 500]
        
        SmoothTriangle:
            points: [250, 350, 345, 400, 325, 500]

        SmoothEllipse:
            pos: 400, 350
            size: 150, 150
            angle_start: 0
            angle_end: 300

        SmoothRoundedRectangle:
            pos: 500, 350
            size: 150, 150
            segments: 30
            radius: [(200, 200), (100, 50), (100, 50),(100, 250),]
        
        PushMatrix
        Rotate:
            angle: 30
            origin: 900, 450
        SmoothRectangle:
            pos: 800, 400
            size: 100, 100
        PopMatrix

        
                
        Quad:
            points: [50, 150, 200, 175, 175, 300, 150, 300]
        
        Triangle:
            points: [250, 150, 345, 200, 325, 300]

        Ellipse:
            pos: 400, 150
            size: 150, 150
            angle_start: 0
            angle_end: 300

        RoundedRectangle:
            pos: 500, 150
            size: 150, 150
            segments: 30
            radius: [(200, 200), (100, 50), (100, 50),(100, 250),]
        
        PushMatrix
        Rotate:
            angle: 30
            origin: 900, 250
        Rectangle:
            pos: 800, 200
            size: 100, 100
        PopMatrix

"""


class Example(App):
    def build(self):

Example().run()
```

Maintainer merge checklist


misl6:
Hi @DexerBR !

Did you managed to find out why the tests are failing?

DexerBR:
> Hi @DexerBR !
> 
> Did you managed to find out why the tests are failing?

@misl6  I noticed that the tests fail right after attempting to disable the antialiasing line (by setting `"kivy/tests/test_button.png"` to the `source` attribute of the target graphic instruction).

I wonder why the test doesn't fail on other platforms. Maybe it's a race condition on that specific platform?

DexerBR:
I found that the problem is on the platform in identifying the file path.

misl6:
Now all tests pass.
Unless you have new uncommitted changes to add, I will start a review tomorrow.

DexerBR:
> Now all tests pass. Unless you have new uncommitted changes to add, I will start a review tomorrow.

Hi! So far, I have no changes to add.

kivy/graphics/vertex_instructions.pyx
```diff
@@ -1592,3 +1626,728 @@ cdef class RoundedRectangle(Rectangle):
...
+        return 0
+
+
+cdef class AAStencilPop(Instruction):
+    cdef int apply(self) except -1:
+        cgl.glColorMask(True, True, True, True)
+        cgl.glDisable(GL_STENCIL_TEST)
+        cgl.glStencilFunc(GL_EQUAL, 1, 0xff)
+        cgl.glStencilOp(GL_KEEP, GL_KEEP, GL_KEEP)
+        return 0
```
misl6, line 1704: Since most of the code is duplicated from stencil instructions, and I agree that introducing the check to avoid calling `cgl.glClear(GL_STENCIL_BUFFER_BIT)` every time makes sense also on standard `StencilPush`, how about introducing the feature in `StencilPush`, but with a setting that enables it? (disabled by default)

That is actually great to conform to DRY principle, and developers and maintainers can take advantage of the speedup if they want to.

DexerBR: I think it's a good idea!

misl6 (commented):
Hi @DexerBR !

I've left some minor comments.

Unfortunately, with `Cython` we do not have a clean way to reduce the duplicated code, but if you have an idea to reduce (at least partially) the duplicated code between the figures, feel free to share.

DexerBR:
> Hi @DexerBR !
> 
> I've left some minor comments.
> 
> Unfortunately, with `Cython` we do not have a clean way to reduce the duplicated code, but if you have an idea to reduce (at least partially) the duplicated code between the figures, feel free to share.

Hi @misl6! I totally agree with you about `Cython`!

I did a small refactoring, moving all the code that could be defined centrally out of the "snooth instructions".

misl6:
> > Hi @DexerBR !
> > I've left some minor comments.
> > Unfortunately, with `Cython` we do not have a clean way to reduce the duplicated code, but if you have an idea to reduce (at least partially) the duplicated code between the figures, feel free to share.
> 
> Hi @misl6! I totally agree with you about `Cython`!
> 
> I did a small refactoring, moving all the code that could be defined centrally out of the "snooth instructions".

Love it!

misl6 (approved):
LGTM. Thank you!

Issue #8305, opened 2023-06-29
two minor Linux install bugs


auwsom:
Ubuntu 22.04 LTS 
Python 3.10.6

`python -m virtualenv kivy_venv` creates a dir with no ../bin
using `python -m venv kivy_venv` does.

`curl -O https://raw.githubusercontent.com/kivy/kivy/master/tools/build_linux_dependencies.sh -o build_kivy_deps.sh` does not create the intended file
`curl https://raw.githubusercontent.com/kivy/kivy/master/tools/build_linux_dependencies.sh -o build_kivy_deps.sh` does. (conflicting option flags)

misl6:
Hi @auwsom !

How about improving our docs via a PR? 😀

jfishzon:
Hello, 
created a PR related to this issue.
https://github.com/kivy/kivy/pull/8324

misl6:
Fixed via #8324

PR #8324, opened 2023-07-19
Fix linux dependencies build instructions

jfishzon:
#8305 
change from:
python -m virtualenv kivy_venv
to:
python -m venv kivy_venv

change from:
curl -O https://raw.githubusercontent.com/kivy/kivy/master/tools/build_linux_dependencies.sh -o build_kivy_deps.sh

to:
curl https://raw.githubusercontent.com/kivy/kivy/master/tools/build_linux_dependencies.sh -o build_kivy_deps.sh

change from:
https://virtualenv.pypa.io/en/latest/
to:
https://docs.python.org/3/library/venv.html

Maintainer merge checklist

### DIFF
--- a/doc/sources/gettingstarted/installation.rst
+++ b/doc/sources/gettingstarted/installation.rst
@@ -74,7 +74,7 @@ subsequent commands outside the virtual environment)::
 Create virtual environment
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Create a new `virtual environment <https://virtualenv.pypa.io/en/latest/>`_
+Create a new `virtual environment <https://docs.python.org/3/library/venv.html>`_
 for your Kivy project. A virtual environment will prevent possible installation conflicts
 with other Python versions and packages. It's optional **but strongly recommended**:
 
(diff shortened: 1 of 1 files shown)


doc/sources/gettingstarted/installation.rst
```diff
@@ -80,7 +80,7 @@ with other Python versions and packages. It's optional **but strongly recommende
 
 #. Create the virtual environment named ``kivy_venv`` in your current directory::
 
-       python -m virtualenv kivy_venv
+       python -m venv kivy_venv
```
misl6, line 83: I'm fine to suggest to use the built-in `venv` command instead of the fully featured `virtualenv`, but maybe in that case we should refer to `https://docs.python.org/3/library/venv.html` instead of `https://virtualenv.pypa.io/en/latest/` ?

misl6 (changes requested):
Hi @jfishzon !

I've left a comment regarding a reference we have to `virtualenv`, can you please take care of it?

doc/sources/gettingstarted/installation.rst
jfishzon: Heya, modified as per your request

misl6 (approved):
LGTM. Thank you!

Issue #8340, opened 2023-08-04
import causing window to open


metabyte144:
**Software Versions**
* Python: 3.10
* OS: Windows 10
* Kivy: 2.2.1
* Kivy installation method: pip

**Describe the bug**
A import from kivy.core.window import Window should not open a window. That means that the module init is executing 
code with massive unexpected side effects. Further this prevents the use of kivy in a GUI process, other than the main process and is a major obstacle for multiprocessing.

Constructor code execution that causes side effects, is a major design problem in all OOP languages and should be considered a bug.

**Expected behavior**
No kivy.core.window module init side effects

**To Reproduce**
A short, *runnable* example that reproduces the issue with *latest kivy master*.

**Code and Logs and screenshots**
```python
from kivy.app import App
from kivy.uix.gridlayout import GridLayout
from kivy.uix.button import Button
from kivy.core.window import Window
from multiprocessing import Process

class DoClose(App):

def build(self):
    self.window = GridLayout()
    self.window.cols = 1
    self.window.size_hint = (0.6, 0.7)
    self.window.pos_hint = {"center_x": 0.5, "center_y": 0.5}
    self.button = Button(text="SHUTDOWN", size_hint=(1, 0.4), bold=True, background_color='#00FFCE')
    self.button.bind(on_press=self.stop)
    self.window.add_widget(self.button)
    return self.window

def gogogo():
DoClose().run()

if name == "main":
proc = Process(name="gcon", target=gogogo)
proc.start()
```
![image](https://github.com/kivy/kivy/assets/78042651/8daf6891-094c-4ce2-986d-9bb029e04127)

**Additional context**
This issue has been suggested by @akshayaurora  as follow-up to #8338

Julian-O:
I think you have stumbled over a systemic issue in the Kivy architecture. However, I do not think your characterisation is accurate.

> Constructor code execution that causes side effects, is a major design problem in all OOP languages and should be considered a bug.

It isn't that the constructor has side-effects that is the problem. It is that the constructor (or indeed any code) is being run *implicitly on import*.

Unless it is the top-level code of a main script, Python modules should avoid having side effects when imported (apart from the obvious declaration of identifiers in the namespace).

However, there are several modules in Kivy that have side-effects - even in `__init__.py` files. As a result, merely importing Kivy modules can break your code in unexpected ways.

[...]

PR #8352, opened 2023-08-17
Converted `_initialized` to BooleanProperty, changed `initialized` to read only AliasProperty, added lazy initialization of window

Samael-TLB:
Removed side-effects during window import, added lazy initialization of window. This would also help advanced users who want to use multiprocessing with Kivy without side-effects. Removed the bug that `Window.initialized` was able to be changed/set by users, which should not be the case. Made ``Window.initialized` read-only property.

Reference https://github.com/kivy/kivy/issues/8340

Maintainer merge checklist

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -353,7 +353,23 @@ class WindowBase(EventDispatcher):
     '''
 
     __instance = None
-    __initialized = False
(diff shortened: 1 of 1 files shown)


Samael-TLB:
Merged into a single pull request by #8376

PR #8353, opened 2023-08-17
Added new conifg value `lazy_window_initialization`

Samael-TLB:
For removing side-effects during Window initialization, while maintaining backward behaviour compatibility, and also for advanced users who would want to have multiprocessing used with Kivy (without multiple window initialization), we added a config value named `lazy_window_initialization`, which when set would lazily do the window initialization.

Reference https://github.com/kivy/kivy/issues/8340

Maintainer merge checklist


Samael-TLB:
Merged into one pr by #8376

Issue #7704, opened 2021-11-22
Arabic support


cryptoone:
If Kivy is still alive and a strong project, Native arabic support is a must.

No native left to right language support; using Bidi and arabic reshaper customer modules is a long run, especially for big apps.
supporting it will surely give Kivy more power, and will ease multi-language development in Kivy.
**Describe the solution you'd like**

this BIDI-Algo and arabic reshaper; mixing it with I8n translation makes it hell of long run, to get a simple arabic/english interface.
PLease add this; this will give kivy a huge boost on supporters and users.

tshirtman:
Help is always welcome, we know about this issue, and there has been attempts at solutions, but no active core-dev speaks/read/write arabic as far as i know, or have the time to focus on that, contributions are welcome, it's probably not going to happen otherwise.

geohadab:
We need RTL and Arabic support. I am ready to help to resolve this issue. @tshirtman

tshirtman:
I think the first step would be getting this PR to be merged https://github.com/kivy/kivy/pull/7452 though i think it's mostly tested with indie, it would be nice to test it with arabic as well.

bambier:
I cant find which part of kivy compiles the first string. If we add `arabic_reshaper.reshape` and `bidi.algorithm.get_display` to that part it can convert non-readable RTL words to readable ones; In other hand it dosnt effects on non-RTL (or LTR) languages as you can see blow
![image](https://github.com/kivy/kivy/assets/22503761/2b4a4bb8-8ea2-4ec1-ac8a-9c63fea49d46)

misl6:
Hi @bambier, does the following works for you? (needs Kivy `2.2.0`)

```kv
Label:
    text: "تشاو موندو"
    font_direction: "rtl"
    font_script_name: "Arab"
```

bambier:
> Hi @bambier, does the following Its for you? (needs Kivy `2.2.0`)
> 
> ```python-kivy
> Label:
>     text: "تشاو موندو"
>     font_direction: "rtl"
>     font_script_name: "Arab"
> ```
> 
> EDIT: FYI, the feature I'm talking about has been added via #8115

Not exactly  what I want...
If we use 'arabic_reshaper' and 'bidi' we don't see that non-RTL fonts breakdown. Also, I feel that few pixels margins and paddings affect texts when we use `font_direction`

These aren't happening when we use those libraries.


[...]

PR #8367, opened 2023-09-09
Add support for RTL language due to the issue #7704

bambier:
Maintainer merge checklist

## Explanation of problem:
As I [described](https://github.com/kivy/kivy/issues/7704#issuecomment-1579971901) in #7704 Persian not supported very well when we use 
```
font_direction: "rtl"
font_script_name: "Arab"
```
I changed the text getting method to fix this problem.

In other hand I added [Vaziri](https://rastikerdar.github.io/vazirmatn/en) font which supports both Latin (Roboto) and Arabic (Vaziri) language by defualt that users can use them if they want.

Also changed requirements that needed for `python-bidi` and `arabic-reshaper`.


p0lygun:
i am not sure about adding dependencies to the project :thinking: , maybe @misl6 can share his thoughts ?

misl6:
Hi @bambier !

Enhancing support for more languages and scripts is something we are working on, and is really important for us.

However, new built-in fonts in our collection do not look like a good idea, as it will increase the package size for everyone.
(I do not even think that is really a good idea to ship fonts for latin languages, but I guess we can only remove them during a major version release (e.g. `3.x.x`) : https://github.com/kivy/kivy/tree/master/kivy/data/fonts)

Instead, I think we should add a section to our documentation (and you're encouraged to do so) that explains how to access system fonts and suggestions about free fonts to support specific languages or scripts.

**About the additional dependencies:**
Aren't [font_direction](https://kivy.org/doc/stable/api-kivy.uix.label.html#kivy.uix.label.Label.font_direction) and [font_script_name](https://kivy.org/doc/stable/api-kivy.uix.label.html#kivy.uix.label.Label.font_script_name) not enough for your use case?
If not, can you explain why?

bambier:
> Hi @bambier !
> 
> Enhancing support for more languages and scripts is something we are working on, and is really important for us.
> 
> However, new built-in fonts in our collection do not look like a good idea, as it will increase the package size for everyone. (I do not even think that is really a good idea to ship fonts for latin languages, but I guess we can only remove them during a major version release (e.g. `3.x.x`) : https://github.com/kivy/kivy/tree/master/kivy/data/fonts)
> 
> Instead, I think we should add a section to our documentation (and you're encouraged to do so) that explains how to access system fonts and suggestions about free fonts to support specific languages or scripts.
> 
> **About the additional dependencies:** Aren't [font_direction](https://kivy.org/doc/stable/api-kivy.uix.label.html#kivy.uix.label.Label.font_direction) and [font_script_name](https://kivy.org/doc/stable/api-kivy.uix.label.html#kivy.uix.label.Label.font_script_name) not enough for your use case? If not, can you explain why?

Hi
As I said earlier in issues (https://github.com/kivy/kivy/issues/7704#issuecomment-1577493312) and (https://github.com/kivy/kivy/issues/7704#issuecomment-1579971901) when we use `font_direction` and `font_script_name` unnecessary margin/paddings affects on text also when we use both RTL and NON-RTL words together we can obviously see that those are broken and looks weird but with these libraries words look good.

About fonts I dont think using only Latin supporting fonts is good idea, isn't it better use font that supports every language or just use phone internal font system (or something similar! All android phone has their own fonts in side them so it makes app lighter)

misl6:
> About fonts, I dont think using only Latin supporting fonts is a good idea. Isn't it better to use font that supports every language (Vaziri font supports both Latin and Arabic characters so we dont need Roboto font and we can delete them) or just use phone internal font system (or something similar! All android phone has their own fonts on them, so they make the app lighter)

As I said in my [previous comment](https://github.com/kivy/kivy/pull/8367#issuecomment-1742078174), accessing system fonts looks like a good idea. (and avoid us to include new fonts).
Deleting/replacing fonts, is not possible, at least not with a major release, cause it may break apps appearance. However, better to discuss that topic separately. (We can find a workaround)

From what I can see, here the main issue is that we're forcing the text direction for all the string, and not only for specific chunks (as it should be).

Can you confirm that the reshaper is not needed if `font_script_name` is set to `Arab` and the string does not mix rtl and non-rtl text, as our implementation with `harfbuzz` already provides a proper handling?

PR #8374, opened 2023-09-18
Avoid multiple values for `noadd` in `VertexInstruction`

DexerBR:
## Notes:

When adding an `Instruction`, there is a possibility of not adding the `Instruction` automatically through the canvas context manager. Therefore, the only way to add an instruction defined with `noadd = True` is from `canvas.add` or `canvas.insert`.

In some cases, it is not desired for the canvas context manager to automatically add instructions to the children of the canvas (as with `BindTexture`), because the instructions are managed (added in specific orders, or not added) within a specific instruction, as is the case from `VertexInstruction`. However, the code that prevents the `BindTexture` from being added also prevents a `Instruction` as a whole from not being added (because it duplicates the definition of the arguments in the `BindTexture`), which can be seen as a limitation.

This PR isolates the definition of `noadd` in `BindTexture` from the rest of the `VertexInstruction`.

---
Maintainer merge checklist

### DIFF
--- a/kivy/graphics/instructions.pyx
+++ b/kivy/graphics/instructions.pyx
@@ -289,7 +289,7 @@ cdef class VertexInstruction(Instruction):
     Triangles, Lines, Ellipse and so on.
     '''
     def __init__(self, **kwargs):
-        # avoid multiple values for 'noadd' in BindTexture bellow
+        # avoid multiple values for 'noadd' in BindTexture below
         noadd_value = kwargs.pop('noadd', False)
 
         # Set a BindTexture instruction to bind the texture used for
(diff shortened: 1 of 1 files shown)


misl6 (changes requested):
Hi @DexerBR !

I've left a comment on a small typo.

Can you please also add an explaination regarding this PR for the records?
When is that happening? Why is that needed?

DexerBR:
> Hi @DexerBR !
> 
> I've left a comment on a small typo.
> 
> Can you please also add an explaination regarding this PR for the records? When is that happening? Why is that needed?

Hi @misl6! Updated PR description!

misl6 (approved):
LGTM. Thank you!

PR #8385, opened 2023-09-29
Rename `CaptureDelegate` to `KivyCaptureDelegate` in `AVFoundationCamera` to avoid duplicated symbols

Asalerm:
kivy-ios camera backend was using same class layout/name as Opencv. Which resulted in duplicated symbols

Class name changed from CaptureDelegate to KivyCaptureDelegate to avoid this issue.


misl6 (approved):
LGTM. Thank you!

Issue #8390, opened 2023-10-06
ffpyplayer type error


cagsurgunu:
**Software Versions**
* Python: 3.10
* OS: Windows
* Kivy:2.1.1
* Kivy installation method: pip

Hello, i write a program with ffpyplayer for iptv. Sometimes iptv channels can have mistake url. if all urls are correct, channel switching(change filename) works perfect. but if there is a mistake url, my kivy app crash.
i get  Error :  TypeError: expected bytes, NoneType found. 
Can i except this error?

misl6:
Can you please post the full log?

cagsurgunu:
> Can you please post the full l


> Can you please post the full log?

Exception has occurred: AttributeError
'NoneType' object has no attribute 'startswith'
  File "D:\iptv\iptv_android\test1.py", line 140, in <module>
    AndroidIP_TV().run()
AttributeError: 'NoneType' object has no attribute 'startswith'

misl6:
Hi @cagsurgunu ,

Unfortunately, the provided log is not enough to triage the issue, can you please provide more? (the full output)

cagsurgunu:
> nately, the provided log is not enou

misl6 thank you very much. i change my code for this problem. So i can't get all error. But i change again old code. if i change to old i will get and send logs to you.

cagsurgunu:
details logs are below:
Traceback (most recent call last):
   File "C:\Python\Python39\lib\runpy.py", line 197, in _run_module_as_main
     return _run_code(code, main_globals, None,
   File "C:\Python\Python39\lib\runpy.py", line 87, in _run_code
     exec(code, run_globals)
   File "c:\Users\omer.yildirim\.vscode\extensions\ms-python.python-2023.18.0\pythonFiles\lib\python\debugpy\__main__.py", line 39, in <module>
     cli.main()
   File "c:\Users\omer.yildirim\.vscode\extensions\ms-python.python-2023.18.0\pythonFiles\lib\python\debugpy/..\debugpy\server\cli.py", line 430, in main
     run()
   File "c:\Users\omer.yildirim\.vscode\extensions\ms-python.python-2023.18.0\pythonFiles\lib\python\debugpy/..\debugpy\server\cli.py", line 284, in run_file
     runpy.run_path(target, run_name="__main__")
[...]

PR #8397, opened 2023-10-08
[WIP] Prevent `VideoFFPy` from crashing when `filename` is `None`

misl6:
Maintainer merge checklist

Fixes issue #8390


Julian-O:
I am coming in with only a very shallow understanding, so apologies if I have misunderstood but... I think this PR is tackling the wrong problem.

`core/video/__init__.py` defines a class, `VideoBase`.
It tries to define an instance variable, `self._filename`, but in line 84, it wrongly refers to it as `self.filename` (no leading underscore).

The wrong variable is initialised, and it is unaffected by later attempts to change it.

Meanwhile `VideoFFPy` also refers to `self.filename`, and chaos ensues...

I would change all references to self._filename.
Having the None check from the PR is probably a good idea too.

PR #8405, opened 2023-10-14
Introduce `clear_stencil` option in `StencilPush`, to allow disabling stencil clearing and improve GPU performance when using Stencil Instructions

DexerBR:
This PR applies the findings obtained from https://github.com/kivy/kivy/pull/8309, where a high GPU utilization was noticed, when using a huge amount of Stencil Instructions.

From the tests, no side effects were noticed (_could this be a potential bug in Kivy? This would mean that Kivy could be "rebuilding" the entire canvas context, even when not necessary, which makes cleaning the stencil unnecessary in all, or almost all the cases?_)

🟢  For comparison purposes, with the tests done in https://github.com/kivy/kivy/pull/8309, the difference in GPU usage went from ~98% (`clear_stencil=True`) to 38% (`clear_stencil =False`).

By default, the option is enabled, but disabling it is recommended unless any side effects are noticed.

Maintainer merge checklist

### DIFF
--- a/kivy/graphics/stencil_instructions.pxd
+++ b/kivy/graphics/stencil_instructions.pxd
@@ -5,6 +5,7 @@ cdef void restore_stencil_state(dict state)
 cdef void reset_stencil_state()
 
 cdef class StencilPush(Instruction):
+    cdef int _clear_stencil
     cdef int apply(self) except -1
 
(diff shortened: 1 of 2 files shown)


misl6 (approved):
LGTM. Thank you!

PR #8415, opened 2023-10-22
Fix cgl_backend build with clang 16

reckenrode:
Per https://registry.khronos.org/OpenGL-Refpages/gl4/html/glShaderSource.xhtml, `glShaderSource` should take a `const GLchar**` not a `const GLchar* const*`. This difference causes building Kivy with clang 16 to fail due to incompatible function pointer conversion errors. This was found while working on the Darwin clang 16 update for nixpkgs.

Maintainer merge checklist


misl6 (approved):
LGTM. Thank you!

cederom:
Yeah this PR broke build on CLANG17 and GCC14 for 2.3.0 release and git master :-(
PoC revert is here and it fixes the build for 2.3.0 and master: https://github.com/kivy/kivy/pull/8911

PR #8435, opened 2023-11-04
Fixes markup `[sup]` followed by `[sub]` doesn't work (issue #8077)

### LINKED ISSUE #8077: Markup [sup] followed by [sub] doesn't work
**Software Versions**
* Python: 3.10.6
* OS: Windows 10 
* Kivy: 2.1.0
* Kivy installation method: conda

**Describe the bug**
If you make a button with `button = Button(text='1[sup]2[/sup][sub]3[/sub]', markup=True)`, the 2 and 3 are both superscript.

image

With a space between them, it displays correctly:

image

**Expected behavior**
It should display the 2 as superscript and the 3 as subscript

**To Reproduce**

```python
from kivy.app import App

class MyApp(App):
    def build(self):
        button = Button(text= '1[sup]2[/sup][sub]3[/sub]', markup=True)
        return button

if __name__ == '__main__':
    MyApp().run()
```

azinneck0485:
Made changes to `markup.py` to fix #8077

Fix was to add a condition `wh < pph` to the logic used to calculate `script_pos` for sub-/superscript in `render_lines` method.  This means the "if condition" is only evaluated when nested, otherwise the "else condition" is evaluated and used in the calculation.

Tested with the following file, which builds on the test script in the original issue report:
[Issue8077_test.zip](https://github.com/kivy/kivy/files/13257789/Issue8077_test.zip)

Output display, showing multiple variations of sub and sup, including nested conditions:
image

Please let me know if there are any other tests I should run (I did not see anything in the documentation), additional changes that need to be make, or if I should make the pull request against a different branch.
 

Maintainer merge checklist

### DIFF
--- a/kivy/core/text/markup.py
+++ b/kivy/core/text/markup.py
@@ -484,12 +484,13 @@ class MarkupLabel(MarkupLabelBase):
                 wh = options['line_height'] * word.lh
                 # calculate sub/super script pos
                 if options['script'] == 'superscript':
-                    script_pos = max(0, psp if psp and (wh<pph) else self.get_descent())
+                    script_pos = max(0, psp if psp and (wh < pph)
+                                     else self.get_descent())
                     psp = script_pos
                     pph = wh
                 elif options['script'] == 'subscript':
                     script_pos = min(lh - wh, ((psp + pph) - wh)
-                                     if pph and (wh<pph) else (lh - wh))
+                                     if pph and (wh < pph) else (lh - wh))
                     pph = wh
                     psp = script_pos
                 else:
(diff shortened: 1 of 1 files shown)


misl6:
Hi @azinneck0485 ,

The automated PEP8 check is failing, can you take care of it?

azinneck0485:
I fixed the places in `markup.py` where the changes I made failed the PEP8 check in my latest commit

misl6 (approved):
LGTM. Thank you!

(merging the PR as-is, but if you want to add some tests and refactor that part to have a better readable code, no one will complain about it. 😅)

Issue #3597, opened 2015-08-22
Add support for multiple mouse button events


Zen-CODE:
Once the left mouse button is clicked, you cannot check for the right mouse button until that touch is finished.

See: https://groups.google.com/forum/#!topic/kivy-users/dlgPYB4Aui8
Minimal example:

    from kivy.app import App
    from kivy.uix.widget import Widget

    class MainWindow(Widget):
        def on_touch_down(self, touch):
            # Once the left mousebutton is pressed, this does not fire for the
            # right mouse button until the left is release
            print(touch.button, "is being pressed")

        def on_touch_move(self, touch):
            print("touch moving: {0}".format(touch.button))

        def on_touch_up(self, touch):
            print(touch.button, "was released")

    class TestingApp(App):
        def build(self):
            return MainWindow()

if __name__ == "__main__":
    TestingApp().run()

janssen:
Following the thread from the mailing list, it seems there is another issue here, too.  Which is that 'on_mouse_XXX' events on the Window are not being delivered, either.  Ever.  Which may have something to do with why on_touch_up events are being missed.

Julian-O:
Ran the code on Windows 11, Kivy 2.2.1.

Far worse than described. Right button releases are ignored. Pressing right button can make left button release instead.

ElliotGarbus:
>Far worse than described. Right button releases are ignored. Pressing right button can make left button release instead.

@Julian-O This is also related to #5898 These issues all seem to be cause by the implementation of the MouseMotionEventProvider(), https://github.com/kivy/kivy/blob/master/kivy/input/providers/mouse.py  

These look like a deceptively easy set of fixes to the on_mouse_press, on_mouse_release, and on_mouse_motion methods.  Are there tests associated with this file?

Julian-O:
@ElliotGarbus: I can't comment meaningfully on the technical side here. I don't know what an EventManager is, and I don't understand the cost of adding new MotionEvents. (That said, I do understand combinatorial explosion - is there a limit to the number of distinct buttons a mouse can have?)

Are you proposing that this prototype become part of the API, or more a proof-of-concept that users can build there own?

If the latter, I would like to see it documented (e.g. added as an example) because the current behaviour is astonishing (in the "Principle of Least Astonishment" sense.)

ElliotGarbus:
@Julian-O The EventManager is documented here: https://kivy.org/doc/stable/api-kivy.eventmanager.html#module-kivy.eventmanager

[...]

PR #8461, opened 2023-11-16
Removed Obsolete features from Mouse Provider, fixes issue

ElliotGarbus:
This PR removes obsolete features from the mouse provider.  This fixes issue #5898 The mouse provider had features that were used to interact with a scatter.  These features were made obsolete by the multitouch_sim.

The removed features:

- Pressing Shift + click would issue a double-tap
- Pressing Alt during a drag would create a new touch
- Holding Alt while pressing a mouse button would cancel drag

Maintainer merge checklist
* [ ] Title is descriptive/clear for inclusion in release notes.
* [ ] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [ ] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #8480, opened 2023-11-27
Added support for multiple mouse button events

ElliotGarbus:
This PR adds support for multiple mouse button events, fixing issue #3597 
The PR also adds some documentation for how the mouse multi-touch simulation works.

Implementation note:  The original implementation used a variable,  self.current_drag to hold the current touch. This limited the behavior to only work correctly when one button was pressed at a time.  In multi-touch simulation is this not an issue as the buttons have predefined roles.  When multi-touch simulation is disabled this did not allow multiple buttons to be used simultaneously.  This fix changes self.curent_drag to a dictionary, and holds a touch per button.  In the case that multi-touch simulation is enabled, a single key is used to access the self.current_drag dict, retaining the original behavior.

When multi-touch simulation is disabled.  Each mouse button will properly report their button up/down state changes.  If the mouse is dragged with multiple buttons pressed there will be an on_touch_move message for each button.

Maintainer merge checklist


kivy/input/providers/mouse.py
```diff
@@ -330,45 +351,40 @@ def create_hover(self, win, etype):
...
-            touch = self.current_drag
-            touch.move([nx, ny])
-            touch.update_graphics(win)
-            self.waiting_event.append(('update', touch))
-        elif self.alt_touch is not None and 'alt' not in modifiers:
-            # alt just released ?
-            is_double_tap = 'shift' in modifiers
-            self.create_touch(win, nx, ny, is_double_tap, True, [])
+        for button in self.current_drag.keys():
+            if self.current_drag.get(button, None):
```
AndiEcker, line 355: The expression in this if-statement will always be True. Therefore the if-statement can be omitted.

ElliotGarbus: Thanks! good catch... fixed.

ElliotGarbus:
**Rebased onto current `master` and reworked to resolve conflicts**

This PR was originally opened in Nov 2023 and had drifted significantly from `master`, since two other mouse-provider changes landed in the meantime and touched the exact same code:

- The removal of the obsolete `alt_touch`/scatter-interaction hooks and the `shift+click → is_double_tap` simulation ([#8461](https://github.com/kivy/kivy/pull/8461))
- The addition of `modifiers` tracking on `MouseMotionEvent` and the `button='all'` special case in `on_mouse_release` (used for Numpad 0 / Fn release events, [#9289](https://github.com/kivy/kivy/issues/9289))

Rather than mechanically resolving conflicts commit-by-commit across the 4 original commits, I reimplemented the fix as a single commit directly on top of current `master`, preserving the original design intent:

**What changed:**
- `self.current_drag` is now a `dict` instead of a single value/`None`.
- When `disable_multitouch` is set, each mouse button gets its own key in `current_drag`, so multiple buttons held down at once are now tracked as independent touches (fixes [#3597](https://github.com/kivy/kivy/issues/3597)).
- When multitouch simulation is *enabled* (the default), all drags continue to share a single `'mt_simulation'` key — behavior is unchanged from before, only one simulated "dot" touch is ever active at a time.

**Tests:**
- Added `test_multiple_buttons_tracked_simultaneously_when_disabled` — the core regression test for #3597: press-and-hold left, then press right without releasing left, and confirm both get independent touches (own `id`s), move independently, and release independently.
- Added `test_multiple_buttons_share_single_drag_when_enabled` — confirms multitouch-sim mode still shares one slot.
- Updated two existing tests (`test_button_all_removes_all_touches`, `test_button_all_with_multitouch_dot`) that asserted `current_drag is None`; it's now `{}`.
- Also manually verified interactively with a small scratch app (not part of the PR) clicking/holding multiple mouse buttons simultaneously with `disable_multitouch` on and off, to confirm the dict keys track the held buttons as expected.

**Docs:**
- Added a "Mouse Input Provider" section to `doc/sources/migration.rst` describing the behavior change and its (minimal) migration impact.

Current state: rebased cleanly onto `master`, `ruff check` clean, all tests in `test_mouse_multitouchsim.py` and `test_mouse_hover_event.py` passing (46/46).

kuzeyron (approved):

PR #8485, opened 2023-11-30
Refactor `filechoosertest_unicode.py`

Julian-O:
Was thinking about making changes to filechooser.py, but was not happy with the state of the existing tests.

This refactor includes:

* Moving imports to the top of the file.
* Removing unused imports. Importing names in a consistent way.
* Importing `platform` from where it is defined in `kivy.utils`, rather than `kivy` where it is explicitly not exported.
* Continue to skip the tests on some platforms, but use `pytest.mark.skipif` to do so, so pytest knows it has been skipped.
* Copy-edit comment
* bfile testing was only doing anything on Python 2. Removed.
* Renamed variables to be more meaningful now there is not unicode versus bytes.
* Ran through `black` so the formatting is neater.
* Removed spurious `for i in range(1):`
* Corrected misspelling `exitsfiles`
* Specified exceptions to ignore.
* Simplified deletion, and made it more robust - one file failure won't stop rest being deleted.

[There was no attempt to improve coverage here - filechooser.py is over 1,000 lines long, and is embarrassingly undertested.]

Maintainer merge checklist

### DIFF
--- a/kivy/tests/test_filechooser_unicode.py
+++ b/kivy/tests/test_filechooser_unicode.py
@@ -1,95 +1,82 @@
 # -*- coding: utf-8 -*-
 # XXX: please be careful to only save this file with an utf-8 editor
+
+from os import remove, rmdir, mkdir
+from os.path import join, dirname, isdir
 import unittest
+from zipfile import ZipFile
+
 import pytest
-from kivy import platform
+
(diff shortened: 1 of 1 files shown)


misl6 (approved):
LGTM. Thank you!

PR #8488, opened 2023-11-30
[WIP] Rectifying event propagation in Kivy 

Samael-TLB:
This commit tries to rectify the assumption that a disabled widget having a touch collision should consume the event. This shall only be true if the widget is visible and disabled and not when the widget is disabled and not visible at all. This helps in a lot of scenarios where due to design constraints and animations, one widget might come on top of another while being hidden completely. As Kivy doesn't have equivalent feature of pointer-events or z-index, so we shouldn't consume the event if the widget is hidden.

WIP: Not to be merged now till the entire event propagation rectified.
Maintainer merge checklist

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -693,7 +693,7 @@ class ScrollView(StencilView):
         if check_children:
             touch.push()
             touch.apply_transform_2d(self.to_local)
-            if self.dispatch_children('on_scroll_start', touch):
+            if self.dispatch_children('on_scroll_start', touch) or self.dispatch_children('on_touch_down', touch):# temporary fix for failing test will be removed in later refactoring
                 touch.pop()
                 return True
             touch.pop()
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -583,8 +583,13 @@ class Widget(WidgetBase):
             If False, the event will continue to be dispatched to the rest
             of the widget tree.
         '''
-        if self.disabled and self.collide_point(*touch.pos):
-            return True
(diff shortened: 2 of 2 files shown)


akshayaurora:
Is there an issue you are trying to solve with this pr? 

Please be specific when mentioning what you are trying to do in a pr,  `event propagation` is generic can include tons of different things.

Specifically here it seems you are having an issue with `touch` events being consumed by disabled widgets. 

Separate out the issue with disabled widgets from your issue with hidden widgets. 

This pr is trying to define how `hidden` widgets are handled in kivy. Kivy does not handle hidden widgets. It's the task of the programmer to manage this as there can be multiple different ways to handle it. 

If you want to add a way for kivy widgets to manage hidden widgets. Then it maybe better to have that as a separate pr that only focuses on managing Hidden widget behaviour that people can use or ignore depending on what they want.

There are good ideas in this pr I suggest you make your pr atomics,  separate out concerns and be more specific about what you are trying to do.

Looking forward to cleaner more atomic Pull Requests.

Samael-TLB:
Hi, @akshayaurora thanks for the advice. This pr was just started as part of a bigger refactoring of how event propagation and consumption is being handled by Kivy. 

You can refer to this https://github.com/orgs/kivy/discussions/14 for the context. The initial starting commit handled the hidden widget case necessary for the further works. I believe I forgot to update the pr context. Can you please kindly reopen this pr or you still suggest having atomic ones?

akshayaurora:
You are talking about too many things in that discussions, in a very unstructured way. As a general rule big prs are a strict no.

Atomic prs are always the way to go. Separation of concerns are important.

Issue #6233, opened 2019-04-04
ActionBar Documentation is Incorrect


DaytonaJohn:
### Versions

* Python: 3.6.7
* OS: Ubuntu 18.04
* Kivy: v1.11.0.dev0
* Kivy installation method: manual

### Description
The documentation for kivy.uix ActionBar states:

> An ActionView will contain an ActionPrevious ...

This implies that an ActionView will automatically include an ActionPrevious, but that is incorrect. The correct documentation should say:

> An ActionView must include a child ActionPrevious ...

DaytonaJohn:
Actually, there are more errors in the ActionBar documentation. Here is a replacement for the entire second paragraph in the actionbar.py docstring:

`An :class:`ActionBar` must contain an :class:`ActionView` with various
:class:`ContextualActionViews <kivy.uix.actionbar.ContextualActionView>`.
An :class:`ActionView` must contain a child :class:`ActionPrevious` which may have title,
app_icon and previous_icon properties. :class:`ActionView` children must be
subclasses of :class:`ActionItems <ActionItem>`. Some predefined ones include
an :class:`ActionButton`, an :class:`ActionToggleButton`, an
:class:`ActionCheck`, an :class:`ActionSeparator` and an :class:`ActionGroup`.
`

PR #8493, opened 2023-12-03
Correct `ActionBar` documentation

Julian-O:
Addresses #6233

Maintainer merge checklist


kivy/uix/actionbar.py
```diff
@@ -14,8 +14,9 @@
 
 An :class:`ActionBar` contains an :class:`ActionView` with various
```
misl6, line 15: We should change also ``` An :class:`ActionBar` contains an :class:`ActionView` ``` as suggested by the issue author?

Julian-O: Thank you. I have fixed this now.

PR #8498, opened 2023-12-07
Refactored touch handling in widget

Samael-TLB:
This pr is the starting of one of the many coming prs targetted to solve some touch event handling issues in `Kivy`.

This pr specifically refactors the touch handling events in `Widget` class and introduces three new checks to determine touch event consumption in the appropriate order taking inspiration from well established systems like web browsers, etc.

The idea is to separate out the event handling into three checks which are essentially easily override able functions to decide whether the event is of concern to the widget, whether any child of the widget did consume the event or not, etc. before the widget itself tries to handle the event.

Further context is at https://github.com/orgs/kivy/discussions/14 .
Maintainer merge checklist


akshayaurora:
As asked earlier please keep atomic prs,  you are trying to handle multiple issues.

ScrollView - separate pr and explain why the pr is needed with a issue that clearly explains the issue.
Touch down.touch up in widgets. Explain what issue you face with current implementation,  what your prs does to address the issue.

Visibility of Widgets: Same as above.

Samael-TLB:
my bad it seems, I think I need to be clearer in much more details.

akshayaurora:
On the contrary I am asking for less detail,  atomic self contained relevant prs touching only one subject at a time. 

Steps you should take::

1. Identify your problem by reducing it to atomic parts.
2.  Isolate it with a minimal isolated running kivy app and post it in support channels,  ask for others to confirm that it is a issue.
3.  Once others confirm the issue,  search for existing such issues if one exists,  try out the solutions mentioned there.
4. If no such issue exists,  open a issue with the self contained app that demonstrates the issue.
5. Provide a pr for that issue,  linked to that pr.

If you are trying to do a Improvement proposal, then follow similar steps only in the last steps instead of a issue you open a issue titled with a KEP(Kivy improvement proposal) 

A KEP should visibly demonstrate a issue with current implementation, less talk/discussion more code problems and solutions.

Hope this helps.

Samael-TLB:
i see, i presumed that the problem identification and other steps done through the discussions but may be I was wrong. Surely, I would follow your suggested ways.

Issue #5789, opened 2018-06-06
Apple Deprecating OpenCL/OpenGL on iOS & macOS


mbrockman1:
Obviously not an immediate issue due to deprecation and not immediate removal, but what does this mean for Kivy/Apple?

tshirtman:
It might become necessary to investigate in the future, if there are, like ANGLE on windows, free implementation of opengl/ES on top of Metal. I know there are commercial ones, but i'm not aware of free ones.

Rewriting kivy for Metal directly, or writting such a compatibility layer ourselves would certainly be a very big effort, and i don't see it happening without a significant funding.

stefanhuber:
Using MoltenVK might be an option: https://github.com/KhronosGroup/MoltenVK

mischievous:
Dumb question.... If I wanted to swap out the opengl for metal or just a c++ osx window (no cocoa) would it be straight forward? I see backends for "debug", "opengl", "mock", and "sdl2".  Would it be as simple as adding a backend for metal and then generating the libraries on osx?

tshirtman:
it's not trivial, but there has been work done at least for kivy-ios to be able to use a fork of Angle that supports the Metal api, and it works, https://github.com/kivy/kivy-ios/pull/448 though it requires more testing, i think maybe such changes should be integrated in kivy, not just kivy-ios, to support OSX, but i didn't look in details into what that entailed.

rpopovici:
@tshirtman what is the staus of this issue? From the looks of it, the latest official angle version(https://github.com/google/angle) has complete support for OpenGL ES2..

tshirtman:
That's great to hear, i didn't follow on this issue and i don't own a mac, but as the test with metalAngle apparently had good results, surely it should be possible to now use the official angle with kivy on osx. It just needs someone with osx and some time to do it, contributions are welcome :).

rpopovici:
@tshirtman I would do it, but I am new to kivy and I don't know exactly what needs to be done. It appears that kivy_deps.angle needs to be built for mac os and the binaries bundled with the kivy whl. Other than that this line need to be removed https://github.com/kivy/kivy/blob/c8698d40d2d2d9d0b9df8d65b5b393551b870b80/kivy/graphics/cgl.pyx#L108 and it should work..

misl6:
At this time, `MetalANGLE` has only been tested over iOS, and it needed some changes over `SDL2`, but seems pretty easy to port it on MacOS.
[...]

PR #8534, opened 2023-12-28
Introduce `ANGLE` GL provider on Apple platforms on top of SDL2 window backend

misl6:
Maintainer merge checklist

Apple deprecated OpenGL and OpenGL ES in 2018 and even if they have not announced yet a date for the complete removal (https://github.com/kivy/kivy/issues/5789).

This PR finally adds ANGLE support for Apple platforms (iOS, macOS, but potentially others), on top of the SDL2 window provider.

The `ANGLE` GL backend has been made the default one for both macOS and iOS, but the user still have the option to opt-out as specified in docs. However, in future, OpenGL via ANGLE will be the only supported option. therefore users are encouraged to use it and report any issues.

To better fit the new changes, as we need to manage the graphics context by ourselves, a new `egl_backend` provider logic has been added.

ATM, the only available EGL provider is `EGLANGLE`, which acts as a trampoline to the platform-specific implementation (`EGLMetalANGLE` for Apple platforms), so the whole logic is extensible to future EGL providers and platforms different than Apple ones.

**Performance topic:**
We can finally use iOS Simulators for Kivy apps, as ANGLE **works better** than Apple's OpenGLES implementation. I can feel some performance gain even on real devices, but that may vary between apps and devices. Please report your feelings and test results.

Some iOS / macOS specific features or providers, which were previously suffering from performance bottlenecks, now can take advantage of ANGLE features to have performance boosts. (`AVFoundationCamera` is expected to undergo an extensive rewrite for that specific reason)

`ANGLE` binaries are taken via https://github.com/kivy/angle-builder, releases, so users can avoid to build it from source (takes a lot of disk space, and also takes a while to complete)


doc/sources/gettingstarted/installation.rst
```diff
@@ -190,6 +192,14 @@ The script will download and build the SDL dependencies from source. It will als
 the dependencies into a directory named `kivy-dependencies`. This directory will be used
 by Kivy to build and install Kivy from source with SDL support.
 
+On macOS is still possible to build SDL2 with the Apple-provided OpenGL framework, but it is
```
AndreMiras, line 195: Minor typo /is still possible/it is still possible/

setup.py
```diff
@@ -991,6 +1031,29 @@ def determine_sdl2():
     else:
         print('AVFoundation cannot be used, OSX >= 10.7 is required')
 
+if c_options["use_angle_gl_backend"]:
+
+    # kivy.graphics.egl_backend.egl_angle is always compiled,
+    # but it only acts as a proxy to the real implementation.
+
+    if platform in ("darwin", "ios"):
```
AndreMiras, line 1039: Is it not already implied by `c_options['use_angle_gl_backend'] = platform in ['darwin', 'ios']` from the same file?

tools/build_macos_dependencies.sh
```diff
@@ -1,5 +1,10 @@
 set -e -x
 
+# If USE_LEGACY_OPENGL is not set, default to "0"
+if [ -z "$USE_LEGACY_OPENGL" ]; then
+        USE_LEGACY_OPENGL="0"
+fi
```
AndreMiras, line 6: We could potentially use the shortcut for it:
```
USE_LEGACY_OPENGL="${USE_LEGACY_OPENGL:-0}"
```

kivy/graphics/cgl_backend/cgl_angle.pyx
```diff
@@ -0,0 +1,153 @@
...
+    #endif
+    """
+    void *eglGetProcAddress(const char * procname)
+
+cpdef is_backend_supported():
+    return USE_ANGLE_GL_BACKEND
+
+def init_backend():
+    if not USE_ANGLE_GL_BACKEND:
+        raise Exception('ANGLE GL backend is not supported')
```
AndreMiras, line 23: This condition is confusing me :thinking: 
Am I reading it the wrong way?
I feel like there's a mismatch between the flag name and the exception message given that we're checking that the flag is not set. Is it because that condition would actually return False if the flag is set to anything including 0?

AndreMiras (approved):
Impressive, thanks for the effort :muscle:

setup.py
misl6: ATM yes, but the plan is to introduce the `ANGLE` backend even for platforms different than `darwin` and `ios`.
So, here we check if we can use the `angle_gl_backend` and then build the correct implementation (e.g. Metal, Android, ...)

That acts more like a placeholder for the future, but I'm open to removing it.

kivy/graphics/cgl_backend/cgl_angle.pyx
misl6: This raises an exception when the developer tries to use the `angle` provider even if `USE_ANGLE_GL_BACKEND!=True`.

We do something similar here:
https://github.com/kivy/kivy/blob/17319904d26f257c1a26b94689ca9f3f6f3f96ff/kivy/graphics/cgl_backend/cgl_glew.pyx#L26

tools/build_macos_dependencies.sh
misl6: Suggestion applied! Thank you!

doc/sources/gettingstarted/installation.rst
misl6: Suggestion applied! Thank you!

kivy/graphics/cgl_backend/cgl_angle.pyx
AndreMiras: All right now I get it thanks!

setup.py
AndreMiras: Ah yeah that makes sense let's keep it

AndreMiras (approved):

Issue #8519, opened 2023-12-20
Label text not updated after calling open_settings()


avonderlieth:
**Software Versions**
* Python: 3.9.2
* OS: WIN32 and Raspberry Pi (bullseye)
* Kivy: 2.2.1
* Kivy installation method: pip install kivy

Description:
After calling open_settings(), label texts are no longer updated on the screen. In the background the values and text properties do change and the color property will be updated, but the GUI still shows the text as it was before the first call to open_settings()

This gist shows the effect using the pong example expanded by a settings object and the call to open_settings() whenever player2 scores. All is well until the first time open_settings() is called, i.e. the label texts and colors are being updated after a score. However, after the first open_settings, only the label colors will be updated:
[https://gist.github.com/avonderlieth/12669a4b4ff68da8a09662c29927763f ](https://gist.github.com/avonderlieth/12669a4b4ff68da8a09662c29927763f)

This is just an example to illustrate the behavior. My original project (much larger) was working with Kivy-1.9.1 but after a (forced) upgrade to 2.2.1 the code was no longer working. The behavior is present in both WIN32 and Raspian (bullseye)

**Expected behavior**
label text and label color are being updated after a score

**To Reproduce**
See example in gist

ElliotGarbus:
I was able to replicate the issue on Windows 11, using the default kivy settings screen.  Here is a minimal reproducer.  While the F1 key can be pressed to open the settings panel, I added a button to as well.
The clock continues to run, but the screen does not update.  Note that setting the text attribute of the Label in the callback shows the same behavior.  

```python
from kivy.app import App
from kivy.lang import Builder
from kivy.clock import Clock
from kivy.properties import NumericProperty

kv = """
BoxLayout:
    orientation: 'vertical'
    Label:
        id: label
        text: str(app.count)
        font_size: 60
    Button:
        size_hint_y: None
        height: dp(48)
        text: 'Open Settings'
        on_release: app.open_settings()
"""


class CheckSettings(App):
    count = NumericProperty(0)

    def build(self):
        return Builder.load_string(kv)

    def on_start(self):
        Clock.schedule_interval(self._count, 1)

    def _count(self, _):
        self.count += 1
        print(f'Count is now {self.count}')


CheckSettings().run()
```

ElliotGarbus:
Interesting the bug does not appear if kivy.uix.settings is imported prior to opening the settings panel:
@avonderlieth this is a potential workaround to your issue.

[...]

Issue #8079, opened 2022-12-22
Spinner broken after settings display


julien6387:
**Software Versions**
* Python: 3.9.7
* OS: RedHat 8u2
* Kivy: 2.1.0
* Kivy installation method: pip install kivy[base]

**Describe the bug**
I'm using the demo/showcase provided in kivy-example.
The ActionSpinner is working fine until I click F1 to get the Kivy settings and close.
Then the ActionSpinner is not refreshed anymore when selecting another item from the DropDown list
I cannot repeat the issue with a Python 3.6.8.

**Expected behavior**
The ActionSpinner should still be working after the Kivy settings are displayed.

**To Reproduce**
the kivy-examples demo/showcase

entaangle:
I'm hitting this too with kivy 2.2.1.  Is forcing me to write my own settings panel.

misl6:
Fixed via #8535

Issue #8047, opened 2022-10-27
kivy settings frezzes GUI to be updated


jeff-dh:
**Software Versions**
* Python: Python 3.8.10
* OS: Linux Mate
          Linux *** 5.4.0-131-generic # 147-Ubuntu SMP Fri Oct 14 17:07:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
* Kivy: default pip wheel (v2.1.0, git_hash: 023bd79b90f9831b45bb8eb449346648aa5fe5f8) & head of master (90400b656e3749eb6f9cb912f224146f0b06082e)

* Kivy `pip install "kivy @ git+https://github.com/kivy/kivy.git@90400b656e3749eb6f9cb912f224146f0b06082e"` in venv

**Describe the bug**
https://stackoverflow.com/questions/73099115/python-kivy-settings-frezzes-gui-to-be-updated

I had the same issue and it still exists and looks to me like a bug.

The issue is:

The GUI (or at least parts of it) "freeze" after the Settings Panel has been opened and closed. `Widget.text` properties are not updated anymore after the settings panel has been opened and closed.

As mentioned in the stackoverflow post this fixes the issue:

`from kivy.uix.settings import SettingsWithSpinner`

But if the import is missing, no error is reported or similar.

**Expected behavior**
Either the import should not be necessary or I guess the missing import should raise an error.

**To Reproduce**

```python
from kivy.app import App
from kivy.uix.boxlayout import BoxLayout

#uncomment the following line to fix the issue
#from kivy.uix.settings import SettingsWithSpinner

from kivy.lang import Builder

Builder.load_string('''
<Interface>:
    Button:
        text: txb.text
        on_release: app.open_settings()

    TextInput:
        id: txb
        text: "aaa"
''')

class Interface(BoxLayout):
    pass

class InterfaceApp(App):
    def build(self):
        return Interface()


InterfaceApp().run()

```

Julian-O:
FYI: Not reproducible on Windows 11, Kivy 2.2.1

misl6:
Should have been fixed via #8535, please let us know if the issue persists on your side.

Issue #7895, opened 2022-06-08
Bug: "text" attribute at labes / buttons can't be changed after returning from settings in kivy 2.1


xvasek:
**Software Versions**
* Python: 3.10.5
* OS: Windows 10 / Linux / android
* Kivy: 2.1 or master
* Kivy installation method: via pip

Hello, 

if you run the code below in kivy 2.1, when you press "Settings" button, and then "Close", the clock on the label "device_label" stops refreshing.

Expected behavior: The clock should refresh after closing settings.

Actually you can't change any text on existing labels / buttons on screen. It is possible to remove widgets and add new ones and then new ones bevahe correctly.

Interesting fact 1: You can change other properties on existing labels / buttons, ie. you can change font size. But text is "frozen".
Interesting fact 2: As you can see in the console, the "text" property of label is actually changing, but it is not visible in the application.

This bug appeared somewhere between kivy 2.0 (program behaves normally) and kivy 2.1 (bug exists) and is still present in current kivy master (v2.2.0.dev0)

```
# -*- coding: utf-8 -*-
import kivy
from kivy.app import App
from kivy.config import Config
from kivy.core.window import Window
from kivy.uix.label import Label
from kivy.uix.button import Button
from kivy.properties import ObjectProperty
from kivy.properties import StringProperty
from kivy.uix.boxlayout import BoxLayout
from kivy.storage.jsonstore import JsonStore
from kivy.base import EventLoop
from kivy.clock import Clock
from time import sleep
from datetime import datetime

class MyTopBar(BoxLayout):
    app = ObjectProperty(None)
    status = StringProperty("Ready", allownone = True)

    def __init__(self, **kwargs):
        def settings(button):
            self.app.open_settings()
        kwargs['orientation'] = "horizontal"
        kwargs['size_hint'] = (1, 0.1)
        super(MyTopBar, self).__init__(**kwargs)
        self.device_label = (Label(text = "Device: ", halign = "center"))
        self.status_label = (Label(text = self.status))
        self.button_settings = Button(text = "Settings")
        self.button_settings.bind(on_release = settings)
        self.add_widget(self.device_label)
        self.add_widget(self.status_label)
        self.add_widget(self.button_settings)

    def set_status(self, text):
        self.status = text
        self.status_label.text = self.status

    def show_clock(self):
        self.device_label.text = "Device: " + self.app.config.get('skate', 'device') + "\n" + datetime.now().strftime('%H:%M:%S')
        print (self.device_label.text)

[...]

PR #8535, opened 2023-12-29
Fix text attributes not changing after the settings panel is opened

ElliotGarbus:
If the settings panel is opened text attributes will no longer get updated.  See issues: #8519, #8079, #8047, #7895, 

Below is a test case.  In version 2.2.1 of kivy, if the setting panel is opened the on screen count will stop, and will not resume.  The Clock callback continues.

```python
from kivy.app import App
from kivy.lang import Builder
from kivy.clock import Clock
from kivy.properties import NumericProperty, BooleanProperty

kv = """
BoxLayout:
    BoxLayout:
        orientation: 'vertical'
        Label:
            id: label
            text: str(app.count)
            font_size: 60
        BoxLayout:
            size_hint_y: None
            height: dp(48)
            Button:
                text: 'Open Settings'
                on_release: app.open()
                disabled: app.settings_open
    BoxLayout:
        id: box
"""


class CheckSettings(App):
    count = NumericProperty(0)
    settings_open = BooleanProperty(False)

    def __init__(self, **kwargs):
        super().__init__(**kwargs)

    def build(self):

    def display_settings(self, settings):
        self.root.ids.box.add_widget(settings)
        self.settings_open = True
        return True

    def close_settings(self, *args):
        self.root.ids.box.clear_widgets()
        self.settings_open = False

    def on_start(self):
        Clock.schedule_interval(self._count, 1)

    def _count(self, _):
        self.count += 1
        print(f'Count: {self.count}')

    def open(self):
        self.open_settings()   # added to use as a breakpoint


CheckSettings().run()
```

The issue: In settings.py the class SettingTitle was erroneously redefining Label.text.

Maintainer merge checklist

### DIFF
--- a/kivy/data/style.kv
+++ b/kivy/data/style.kv
@@ -880,6 +880,7 @@
         font_size: '15sp'
 
 <SettingTitle>:
+    text: self.title
     text_size: self.width - 32, None
     size_hint_y: None
     height: max(dp(20), self.texture_size[1] + dp(20))
--- a/kivy/uix/settings.py
+++ b/kivy/uix/settings.py
@@ -672,7 +672,7 @@ class SettingTitle(Label):
     '''A simple title label, used to organize the settings in sections.
     '''
 
-    title = Label.text
+    title = StringProperty()
 
     panel = ObjectProperty(None)
 


misl6 (approved):
LGTM. Thank you!

And `2.3.0rc2` on its way!

PR #8544, opened 2024-01-01
Fix typo and mistake at FAQ.md

mak8kammerer:
> `#### I get an "ImportError: No module named event". How come?`

The `event` module is a simple python module, while `_event` is a cython module which we cannot import if it has not been compiled. [Helpful link.](https://groups.google.com/g/kivy-users/c/kaJTLAvuqJ8)

~> `make force`~

~To build a Python extension, we need to run `setup.py` script, not the `make` command.~

Maintainer merge checklist

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -281,7 +281,7 @@ latest available version and retry.
 #### I get an "ImportError: No module named _event". How come?
 
   If you use Kivy from our development version, you must compile it before
-using it. Check :ref:`installation guide<installation-canonical>` for more information.
+using it. Check [installation guide](https://kivy.org/doc/stable/gettingstarted/installation.html) for more information.
 
 ## Android Questions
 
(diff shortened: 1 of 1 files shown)


misl6 (changes requested):

FAQ.md
```diff
@@ -278,12 +278,10 @@ execution.
...
 
 
-#### I get an "ImportError: No module named event". How come?
+#### I get an "ImportError: No module named _event". How come?
 
   If you use Kivy from our development version, you must compile it before
-using it. In the kivy directory, run
-
-      make force
+using it. Check :ref:`installation guide<installation-canonical>` for more information.
```
misl6, line 284: The `:ref:` does not work here, you should use a https link instead.

misl6 (changes requested):

misl6 (approved):
LGTM. Thank you!

PR #8562, opened 2024-01-14
Docs: fix typo in kivy.uix.image

mak8kammerer:
Removed unnecessary `-` in class rule (.kv)

Maintainer merge checklist

### DIFF
--- a/kivy/uix/image.py
+++ b/kivy/uix/image.py
@@ -57,7 +57,7 @@ want your image to be greater than the size of your widget, you could do::
 
 And in your kivy language file::
 
-    <-FullImage>:
+    <FullImage>:
         canvas:
             Color:
                 rgb: (1, 1, 1)


pythonic64:
Why do you mean "-" in "<-FullImage>:" is a typo? Using "-" means don't include canvas instructions from "Image" class in canvas of "FullImage".

mak8kammerer:
Wow! I don't know about this feature. Perhaps it should be mentioned in the documentation.

Julian-O:
@mak8kammerer: I went to raise an issue saying this should be documented, but stopped when I found it was!

https://kivy.org/doc/stable/api-kivy.lang.html#redefining-a-widget-s-style

> Sometimes we would like to inherit from a widget in order to use its Python properties without also using its .kv defined style. For example, we would like to inherit from a Label, but we would also like to define our own canvas instructions instead of automatically using the canvas instructions inherited from the Label. We can achieve this by prepending a dash (-) before the class name in the .kv style definition.

mak8kammerer:
It looks like I searched very badly. @Julian-O, thank you!

Issue #8457, opened 2023-11-15
OpenGL GetProcAddress function


bl1nch:
Hi, Kivy developers! I decided to try using the `MPV` video player as a backend for video rendering. There are `python-mpv` `ctypes bindings` for the `MPV core library` https://github.com/jaseg/python-mpv, that supports video rendering directly into `OpenGL teture`. There is an example in its documentation that shows how to use it with `OpenGl`.
https://github.com/dfaker/imgui_glfw_pythonmpv_demo/blob/main/main.py

But in this example `MPV` requires `GetProcAndress` `OpenGL` function that I can't find in `Kivy` graphics instructions.
```
def get_process_address(_, name):
    print(name)
    address = glfw.get_proc_address(name.decode('utf8'))
    return ctypes.cast(address, ctypes.c_void_p).value
```
As I understand `Kivy` based on `OpenGL`, so I would like to know if it possible to implement this function to support `MPV` integration, because as I understand it, now video rendering in `Kivy` implemented not the best way. I tried to figure out how it works, and as I understand it, decoded by `ffmpeg` frames copied from video memory to RAM and then rendered by `Kivy`, that gives a high CPU load. So it seems like the `MPV` can solve it.

Thanks in advance for your reply!

Julian-O:
I interpret this as a new feature request: "Expose OpenGL's `get_proc_address()` through `kivy.graphics.opengl` and/or `kivy.graphics.gl_instructions'"

(No comment on whether this is a good idea; I am out of my depth on that.)

bl1nch:
Hi, is there any news on this issue?

Julian-O:
Hi @bl1nch:

I think we might need to reset expectations here.

This is an Open Source project based on volunteers.

This request has been promptly added to the long list of new feature suggestions, where it will sit - hopefully garnering upvotes from people who agree it is useful - until a volunteer decides that they want to work for what I predict will be dozens of hours implementing (perhaps hundreds given the number of platforms to consider), and other volunteers can review it. 

At the moment, there doesn't seem to be a strong justification: making it eventually possible to use yet another video player, because you don't like the current implementations, for unclear reasons.

Now, maybe other developers will know more about this and see the value in spending that time, but I see a lot of other higher priority suggests, and would predict that this may take months or more likely years before it is attempted.

If that doesn't meet your needs, you need to provide your own resources to get it implemented - implement the change yourself, or persuade others to do so. (At the moment, there is no bounty system available here, or I would suggest that.)

[...]

PR #8589, opened 2024-01-23
OpenGL GetProcAddress function

bl1nch:
https://github.com/kivy/kivy/issues/8457

Maintainer merge checklist

### DIFF
--- a/kivy/graphics/cgl_backend/cgl_sdl2.pyx
+++ b/kivy/graphics/cgl_backend/cgl_sdl2.pyx
@@ -6,10 +6,17 @@ include "../common.pxi"
 include "../../include/config.pxi"
 
 from kivy.graphics.cgl cimport *
+from libc.stdint cimport intptr_t
 
 IF USE_SDL2:
     cdef extern from "SDL.h":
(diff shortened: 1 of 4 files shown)


kivy/core/window/_window_sdl2.pyx
```diff
@@ -320,6 +321,9 @@ cdef class _WindowSDL2Storage:
         cdef void * p = SDL_GL_GetProcAddress(c_name)
         return <intptr_t>p
 
+    def reset_gl_context(self):
```
baseplate-admin, line 324: Do we need this function? 
Why not call the function directly instead of making it like this

bl1nch: Do you mean the `SDL_GL_GetProcAddress `function or `reset_gl_context`?
Anyway `SDL_GL_GetProcAddress` returns void pointer so there's no point to call it directly because we need to cast it to intptr_t, and about `reset_gl_context ` I don't know how to call it directly because it can't be imported from `kivy.graphics.instructions` from python

misl6 (changes requested):

Issue #2156, opened 2014-05-06
Kivy Audio "on_eos" event.


marner2:
It would be very useful to have an "on_eos" event raised in the audio api similar to how the videos have. The default case could just be:
"Clock.schedule_once(lambda: dispatch('on_eos'), self._get_length())"
or some such construct.
##

PR #8609, opened 2024-02-11
[Audio]: Add 'on_eos' event

mak8kammerer:
Good evening!

I want to add `on_eos` event to Kivy audio system. It works the same way as in the [video](https://kivy.org/doc/stable/api-kivy.core.video.html), but is only available with `ffpyplayer` and `gstplayer` providers.

Related issue: [#2156](https://github.com/kivy/kivy/issues/2156).

P.S.: Could someone please advise me on how the tests for this PR should look? Or is it not necessary to write them for this case?

Maintainer merge checklist

### DIFF
--- a/kivy/core/audio/__init__.py
+++ b/kivy/core/audio/__init__.py
@@ -100,7 +100,10 @@ class Sound(EventDispatcher):
         `on_stop`: None
             Fired when the sound is stopped.
         `on_eos`: None
-            Fired when EOS is hit.
+            Fired when EOS is hit. Available only with 
+            `ffpyplayer` and `gstplayer` backends.
+
+            .. versionadded:: 3.0.0
     '''
 
     source = StringProperty(None)
@@ -147,7 +150,7 @@ class Sound(EventDispatcher):
     :attr:`loop` is a :class:`~kivy.properties.BooleanProperty` and defaults to
     False.'''
 
-    __events__ = ('on_play', 'on_stop')
+    __events__ = ('on_play', 'on_stop', 'on_eos')
 
     def on_source(self, instance, filename):
         self.unload()
--- a/kivy/core/audio/__init__.py
+++ b/kivy/core/audio/__init__.py
@@ -99,6 +99,8 @@ class Sound(EventDispatcher):
             Fired when the sound is played.
         `on_stop`: None
             Fired when the sound is stopped.
+        `on_eos`: None
+            Fired when EOS is hit.
     '''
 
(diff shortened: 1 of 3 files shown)


Julian-O:
As well as unittests, this needs some documentation. I've read the code, and I am still just guessing what EOS stands for.

mak8kammerer:
Tests are ready. If somebody has any suggestions, let me know. Tests are not my thing.

gottadiveintopython:
I'm not familiar with the audio providers so I have nothing to say about it, but you can use the `kivy_clock` pytest fixture instead of `kivy.clock.Clock`, which makes the unittests have less side-effects.

gottadiveintopython:
Nice! You don't need the `set_clock`.

```python
def test_on_eos_event(self, kivy_clock):
    ...
```

mak8kammerer:
> Nice! You don't need the `set_clock`.
> 
> ```python
> def test_on_eos_event(self, kivy_clock):
>     ...
> ```

It won't work. Take a look at [this StackOverflow question](https://stackoverflow.com/questions/68439799/typeerror-missing-1-required-positional-argument-while-using-pytest-fixture) and [this code snippet](https://github.com/kivy/kivy/blob/c0d4894384abb81cf4729984a1eaa4c437d266da/kivy/tests/test_filechooser_unicode.py#L50-L52).

gottadiveintopython:
> It won't work.

Sorry, I didn't know that.

mak8kammerer:
@misl6, could you please review my PR, as it seems that @Julian-O is offline for long time.

Sorry for pinging.

gottadiveintopython:
Seems like `on_play` and `on_stop` events work differently depending on audio providers when `sound.loop = True`. 

- sdl2 ... these events will be fired every iteration
- gstplayer ... nothing will be fired during the loop

mak8kammerer:
> Seems like `on_play` and `on_stop` events work differently depending on audio providers when `sound.loop = True`.
> 
>     * sdl2 ... these events will be fired every iteration
> 
>     * gstplayer ... nothing will be fired during the loop

This is not related with this PR. I think it should be clarified in a separate issue.

gottadiveintopython:
Yeah, I'll open an issue. (I forgot to mention that the issue occurs on the master branch)
The reason I mentioned the issue is that, unless we define when `on_play` and `on_stop` events occur, it may be hard to define when `on_eos` should occur.

mak8kammerer:
Resolved merge conflicts.

Issue #8140, opened 2023-02-18
On macOS, Window size is scaled differently from widget size


probablyfine:
**Software Versions**
* Python: 3.8.9
* OS: macOS 12.6 (Apple silicon)
* Kivy: 2.1.0
* Kivy installation method: pip

**Describe the bug**
On macOS, the units for `Window` size (`kivy.core.window.Window`) seem to be twice as large as the units for widgets, and this appears to be completely independent of screen dpi/scaling: changing the screen scaling doesn't affect this behavior. Using the `kivy.metrics.dp` method also doesn't help with the issue.

**Expected behavior**
The units for Window size should be the same as the units for widget size.

**To Reproduce**
```
from kivy.core.window import Window
from kivy.app import App
from kivy.uix.boxlayout import BoxLayout
from kivy.lang import Builder

Builder.load_string(
"""
<MyWidget>
  size_hint: None, None
  size: 200, 200
  Label:
    canvas.before:
      Color:
        rgb: 0.7, 0.7, 0.7
      Rectangle:
        pos: self.pos
        size: self.size
"""
)

class MyWidget(BoxLayout):
  def __init__(self, **kwargs):
    super(MyWidget, self).__init__(**kwargs)

    # This makes the window twice as large as expected
    Window.size = self.size
    
    # Dividing the dimensions by 2 produces the "correct" size
    # Window.size = (self.size[0]/2, self.size[1]/2)

class MyApp(App):
    def build(self):
        return MyWidget()

if __name__ == '__main__':
    MyApp().run()
```

In the above code, `Window.size` is set to the same size as the main `BoxLayout`'s size, which is set in the `kv` string. This makes the window twice as large as it should be. If you uncomment the other line that first divides each `BoxLayout` dimension by 2, this produces the "correct" size window.

matheussmoreira:
The same is occurring with me, I'll investigate what can be done.

matheussmoreira:
For some reason this bug is not occurring with me anymore. These last days I had to change my machine (macOS version remains the same of this issue's), but maybe this problem may have been resolved by some recent update too!

Julian-O:
@probablyfine: Given the experience of @matheussmoreira, can you please tell us if this problem disappears with the latest versions?

probablyfine:
[...]

PR #8630, opened 2024-02-24
Add `Window.unrotated_size`, simplify, clean-up and fix `Window.size` and `Window._density` logic.

misl6:
Maintainer merge checklist

- Avoid usage of `SDL_GL_GetDrawableSize` which is a strict requirement for https://github.com/kivy/kivy/pull/8534:
   - Instead we're using `SDL_GetWindowSizeInPixels` (https://wiki.libsdl.org/SDL2/SDL_GetWindowSizeInPixels) which is a more generic version, and is also `SDL3` ready, as `SDL_GL_GetDrawableSize` will be removed.
 
 - Avoid DRY issues for `_get_width` and `_get_height`, now the generic code from `_get_size` is used instead.
 
 - Avoid usage for `_get_gl_size()` in `WindowBase`, as this is a SDL2 specific feature. `_get_gl_size()` has been moved to `window_pixel_size` property, which is still SDL2-specific but is not used outside of the `SDL2` implementation.
 
 -  `resize_display_mode` method for SDL2 implementation was left abandoned, and not used by the API, but it actually makes sense when a resize is requested when in fullscreen mode, this has been reworked and is now called by the `resize_window` method, when needed.
 
 - `_get_effective_size` has been removed, and `Window.unrotated_size` public property has been added (and achieves the same result, while fixing issues)
 
 - Fixes an issue that prevented to correctly rotate a window on screens with density > 1.0
 
 - Fixes issue #8140:
     - Setting `Window.size=100,100`, when `_density=1.0`, sets the `system_size` to the same value.
     - Setting `Window.size=100,100`, when `_density=2.0`, sets the `system_size` to `50,50`
  
  No `versionchanged` has been applied to `Window.size` as `Window.size` was already meant to work as now better explained in docs.

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -417,26 +417,6 @@ class WindowBase(EventDispatcher):
     :attr:`modifiers` is an :class:`~kivy.properties.AliasProperty`.
     '''
 
-    def _get_size(self):
-        r = self._rotation
-        w, h = self._size
-        if platform == 'win' or self._density != 1:
-            w, h = self._win._get_gl_size()
-        if self.softinput_mode == 'resize':
-            h -= self.keyboard_height
-        if r in (0, 180):
-            return w, h
-        return h, w
-
-    def _set_size(self, size):
-        if self._size != size:
-            r = self._rotation
-            if r in (0, 180):
-                self._size = size
-            else:
-                self._size = size[1], size[0]
-            self.dispatch('on_pre_resize', *size)
-
     minimum_width = NumericProperty(0)
     '''The minimum width to restrict the window to.
 
@@ -481,9 +461,76 @@ class WindowBase(EventDispatcher):
     and defaults to True.
     '''
 
+    def _get_unrotated_size(self):
+        w, h = self._size
+
+        w *= self._density
+        h *= self._density
+
+        return w, h
+
+    unrotated_size = AliasProperty(
+        _get_unrotated_size, None, bind=("_size", "_density")
+    )
+    '''The unrotated size of the window in drawable pixels. This is the actual
+    pixel size of the drawable area of the window, and takes into account the
+    window's pixel density.
+
+    The pixel density is the ratio of the physical pixels to the size of the
+    window in window coordinates. For example, on a Retina HiDPI display, the
+    pixel density is typically 2.0, meaning that the window's drawable size in
+    pixels is twice the size in window coordinates.
+
+    .. versionadded:: 3.0.0
+
+    :attr:`unrotated_size` is a
+    read-only :class:`~kivy.properties.AliasProperty`.
+
+    '''
+
+    def _get_size(self):
+        w, h = self._get_unrotated_size()
+
+        r = self._rotation
+
+        if r not in (0, 180):
+            h, w = w, h
+
+        return w, h
+
+    def _set_size(self, size):
+        r = self._rotation
+        new_w, new_h = size
+
+        if r not in (0, 180):
+            new_h, new_w = new_w, new_h
+
+        if (
+            new_w / self._density == self._size[0]
+            and new_h / self._density == self._size[1]
+        ):
+            # The size is the same, no need to resize anything
+            return
+
+        self._size = (new_w / self._density, new_h / self._density)
+        self.dispatch('on_pre_resize', *self._size)
+
     size = AliasProperty(_get_size, _set_size, bind=('_size', '_rotation'))
-    '''Get the rotated size of the window. If :attr:`rotation` is set, then the
-    size will change to reflect the rotation.
+    '''The rotated size of the window in drawable pixels. This is the actual
+    pixel size of the drawable area of the window, and takes into account the
+    window's :attr:`rotation` and pixel density.
+
+    The pixel density is the ratio of the physical pixels to the size of the
+    window in window coordinates. For example, on a Retina HiDPI display, the
+    pixel density is typically 2.0, meaning that the window's drawable size in
+    pixels is twice the size in window coordinates.
+
+    Consider using :attr:`system_size` if you want to programmatically set the
+    window size, as is more portable, and will give a nicer UX across different
+    OSes and displays.
+
+    This property differs from :attr:`unrotated_size` as it also accounts for
+    the window's rotation.
 
     .. versionadded:: 1.0.9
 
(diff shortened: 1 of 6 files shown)


Xovaryu:
What exactly would the code be such that `? = [299, 199]` yields a window size of actually 299, 199 without rounding errors, that could also be used with `Widget.size` after this PR?

misl6:
> What exactly would the code be such that ? = [299, 199] yields a window size of actually [299, 199] on a system with 200% scaling without rounding errors, that could also be used with Widget.size after this PR?

I'm not sure I understood what you mean with "What exactly would the code be such that ? = [299, 199]", can you please make sure that  I can understand it?

Xovaryu:
> I'm not sure I understood what you mean with "What exactly would the code be such that ? = [299, 199]", can you please make sure that I can understand it?

As I see it setting `Window.size` always multiplies the passed values with `._density`. I understand that this is now the normal way Kivy handles system scaling, but it does have 3 problems.
1. Resolutions can't be set accurately. This is a bit of a nitpick, but nevertheless a resolution like [299, 199] can never be set with `Windows.size`, as attempting to do so will after the multiplication and rounding steps make a window resolution of [298, 198] instead.
2. This may cause compounding rounding errors when the window size is manipulated back and forth with fixed values like this code that manifests the issue on my scaling 225% system in Kivy 2.3. The PR may remove the need for putting that factor in manually or programmatically, and fix self-assignment, but the rounding issues do remain.
```python
Window.size = ((Window.size[0]+24)/2.25, Window.size[1]/2.25)
EventLoop.idle()
Window.size = ((Window.size[0]-24)/2.25, Window.size[1]/2.25)
EventLoop.idle()
print(Window.size)
```
3. There is a schism between `Widget.size` and `Window.size`. That was the actual issue in https://github.com/kivy/kivy/issues/8140 because `Widget.size` is an accurate physical pixel size. If `Widget.size` is reported back as [4, 4] for instance, that is exactly 4 times 4 pixels on screen, so a total of 16 physical pixels on the screen associated with that widget. But on a system with a scaling of 200% if the `Window.size` would report [4, 4] then this would be with that factor included, so the physical pixel size would actually be [2, 2] and that's 4 pixels. This inherently makes working with widget and window sizes together at once pretty unintuitive and iffy like the rounding errors outlined above. This PR hence doesn't address that mentioned issue at all as far as I see, but rather fixes a tangential issue.

Hence I am asking for a piece of code that explicitly bypasses all the application of scaling. Something like `Window.pixel_size = [299, 199]` that will then set this precise value without ever doing any multiplication or division on it ever.

misl6:
Let's open a discussion regarding this topic.

Window coordinates (position, size) and drawable size are two different beasts and cannot be comparable (and not due to Kivy's fault).

I will take macOS APIs to better explain this matter, and I will use macOS terminology, but AFAIK, this is pretty similar across all the platforms (and we also have to conform to our window provider APIs).
 Window coordinates (pos, size) are in `Points`, where a `Point` is one unit in user space, before any transformations on the space.

On Retina displays (HIDPI),  there are 4 pixels for each point, and for this reason, `Points` (again, window pos, window size, pointer location) can be expressed as floating-point values to achieve greater precision. (but consider http://www.openradar.me/11978869)

Here you can find a better and exhaustive explanation from old Apple docs: https://developer.apple.com/library/archive/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Explained/Explained.html#//apple_ref/doc/uid/TP40012302-CH4-SW1  

--

So, going back to us is just not possible in some cases to set a Window size and pos exactly to what the user specified in pixels, as a conversion between points and a backing scale factor (Kivy exposes it via `density`) is strictly needed.

Some additional references, regarding our main window provider (SDL): 
- Set window size: https://wiki.libsdl.org/SDL2/SDL_SetWindowSize.  
> The window size in screen coordinates may differ from the size in pixels, if the window was created with [SDL_WINDOW_ALLOW_HIGHDPI](https://wiki.libsdl.org/SDL2/SDL_WINDOW_ALLOW_HIGHDPI) on a platform with high-dpi support (e.g. iOS or macOS). Use [SDL_GL_GetDrawableSize](https://wiki.libsdl.org/SDL2/SDL_GL_GetDrawableSize)() or [SDL_GetRendererOutputSize](https://wiki.libsdl.org/SDL2/SDL_GetRendererOutputSize)() to get the real client area size in pixels. “
- Get window size in pixels (via backing scale factor): https://wiki.libsdl.org/SDL2/SDL_GetWindowSizeInPixels 

So, a proposal: 
- Update `Window.size` docs to better explain this behaviour, but by fixing the actual code via this PR, so `(Window.size = Window.size) == Window.size`, except for rounding. 
- Update `Window.system_size` docs, to better explain the usage, and encourage users to use this property to set the Window size, by following the Window coordinates logic, and not pixels one.


Do you agree ?

Xovaryu:
> Window coordinates (position, size) and drawable size are two different beasts and cannot be comparable (and not due to Kivy's fault).
...

Yeah, after digging into SDL something like that is already what I suspected.

> So, going back to us is just not possible in some cases to set a Window size and pos exactly to what the user specified in pixels, as a conversion between points and a backing scale factor (Kivy exposes it via density) is strictly needed.

So this is basically a cross-platform regressive issue. On one hand this is a terribly weird hill for precision to go and die on, on the other hand... well it is very far from the weirdest or most unintuitive behavior I've seen.
Now, from testing with other applications I know that it can be done, but the fact that it can be done in theory, and in practice with different frameworks that are not in place unfortunately isn't terribly helpful now is it...
SDL2 seems to support it via hints (https://www.reddit.com/r/GraphicsProgramming/comments/115ntyq/sdl2_windows_resolution_scaling/) but I have little idea how applicable it is and what other cross-platform horrors it may introduce. And then I think a switch to SDL3 may be planned? If so, then that might be another can of worms.

> So, a proposal: 
> 
> * Update `Window.size` docs to better explain this behaviour, but by fixing the actual code via this PR, so `(Window.size = Window.size) == Window.size`, except for rounding.
> 
> * Update `Window.system_size` docs, to better explain the usage, and encourage users to use this property to set the Window size, by following the Window coordinates logic, and not pixels one.
> 
> 
> Do you agree ?

Sure. I mean in an ideal world I would love to see a hybrid solution where users of Kivy can have their cake (scaling with system scale) and eat it too (scaling without system scale), but sometimes if the complexity outweighs the benefits (or there are hard framework issues) a no is the right answer. And fortunately for my use case pixel precision is a very low priority peeve, not a strict requirement, so I'll be fine either way.
So a fix for the self-assignment issue and then better documentation may well be the best course of action.

misl6:
> Now, from testing with other applications I know that it can be done, but the fact that it can be done in theory, and in practice with different frameworks that are not in place unfortunately isn't terribly helpful now is it...
SDL2 seems to support it via hints 

See https://github.com/kivy/kivy/pull/8058 (but this is just a tweak for Windows).



> And then I think a switch to SDL3 may be planned? If so, then that might be another can of worms.

Not officially in the `Kivy 3.0.0.0` plan, as they (SDL) do not have an ETA for SDL3 release, but if SDL3 release happens before `Kivy 3.0.0` reaches Beta, then yes.


Today I will work on the cleanup + docs, so later we can merge it as-is and continue with ANGLE integration, but feel free to propose changes (to docs and code) in future.

PR #8633, opened 2024-02-26
Typo: "kiby" instead of "kivy" in global variable in _text_pango.pyx

Xovaryu:
This is just a typo fix for the line `global kiby_fontfamily_cache`, the kiby instead of kivy being the typo. I don't use Pango myself and it seems a little odd that this typo is in the code when it looks like it might break something? Someone may want to have a look into this.

Maintainer merge checklist


misl6 (approved):
Nice catch.

LGTM. Thank you!

PR #8640, opened 2024-03-04
Typo: Docs grammar issue in loader.py (been/be)

Xovaryu:
Fixed a typo, been/be in this segment for the docs:
```
    The default value is 2 for giving a smooth user experience. You could
    increase the number of workers, then all the images will be loaded faster,
    but the user will not be(en) able to use the application while loading.
    Prior to 1.6.0, the default number was 20, and loading many full-hd images
    was completely blocking the application.
```
The block also seems a little unclear to me, though that may simply because I lack the technical background of what's going on.

Maintainer merge checklist


misl6 (approved):
LGTM. Thank you!

PR #8650, opened 2024-03-18
Expose a `native_handle` property on `WindowBase`

### LINKED ISSUE #8597: Expose a method on `kivy.core.window` to get the window handle
Sometimes, and #8596 is one of these cases, we need to get the native Window handle.

With SDL2 (our main window core provider) that could be done by exposing the info retrieved via `SDL_GetWindowWMInfo`, but the API should be provider-agnostic and be ready for the next decade of Window providers.

DataTriny:
Maintainer merge checklist

Closes #8597 

Retrieving a pointer to the `NSWindow` will also be required for AccessKit integration, but the native code is not present at the moment.

### DIFF
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -507,6 +507,16 @@ cdef class _WindowSDL2Storage:
                 windows_info.hdc = wm_info.info.win.hdc
                 return windows_info
 
+    def get_native_handle(self):
+        window_info = self.get_window_info()
+
+        IF USE_WAYLAND:
+            if isinstance(window_info, WindowInfoWayland):
+                return window_info.surface
+
+        if window_info is not None:
+            return window_info.window
+
     # Transparent Window background
     def is_window_shaped(self):
         return SDL_IsShapedWindow(self.win)
--- a/kivy/core/window/_window_sdl2.pyx
+++ b/kivy/core/window/_window_sdl2.pyx
@@ -512,17 +512,17 @@ cdef class _WindowSDL2Storage:
         window_info = self.get_window_info()
 
         if setupconfig.USE_X11:
-            from kivy.core.window.window_info import WindowInfoX11
+            from .window_info import WindowInfoX11
             if isinstance(window_info, WindowInfoX11):
                 return window_info.window
 
         if setupconfig.USE_WAYLAND:
-            from kivy.core.window.window_info import WindowInfoWayland
+            from .window_info import WindowInfoWayland
             if isinstance(window_info, WindowInfoWayland):
                 return window_info.surface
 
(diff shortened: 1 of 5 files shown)


kivy/core/window/_window_sdl2.pyx
```diff
@@ -507,6 +507,16 @@ cdef class _WindowSDL2Storage:
                 windows_info.hdc = wm_info.info.win.hdc
                 return windows_info
 
+    def get_native_handle(self):
+        window_info = self.get_window_info()
+
+        IF USE_WAYLAND:
```
misl6, line 513: We're trying to avoid `IF/ELSE` and `DEF` on new code (and slowly migrating during changes required for `3.0.0`) as is deprecated.

See: http://docs.cython.org/en/latest/src/userguide/migrating_to_cy30.html#deprecation-of-def-if

misl6 (changes requested):
Hi @DataTriny !

Thank you for taking care of it.

I've left a comment regarding a Cython deprecation, please consider taking care of it so we do not introduce deprecated code.

Regarding missing platforms: I will open a separate issue to target it on a separate PR, but will be quite straightforward as `SDL_GetWindowWMInfo` already provides the info we need.

What do you think about adding a test that exercises this new feature?

kivy/tests/test_window_base.py
```diff
@@ -126,3 +127,19 @@ def test_window_opacity_clamping_negative(self):
...
+        win = self.Window
+        native_handle = win.native_handle
+
+        if setupconfig.USE_X11:
+            self.assertNotEqual(native_handle, 0)
+
+        if setupconfig.USE_WAYLAND:
+            self.assertNotEqual(native_handle, 0)
+
+        if setupconfig.PLATFORM == 'win32':
```
DataTriny, line 144: This check is different from the one in the method being tested, but this is what is used in `test_window_info.py` so I figured maybe your CI expects this exact configuration.

kivy/tests/test_window_base.py
```diff
@@ -126,3 +127,19 @@ def test_window_opacity_clamping_negative(self):
         if self.check_opacity_support():
             self.Window.opacity = -1.5
             self.assertEqual(self.Window.opacity, 0.0)
+
+
+class WindowNativeHandleTest(GraphicUnitTest):
```
DataTriny, line 132: I guess another property of native handles is that they should not change over time, but I really don't know how to exercise that in a test.

misl6: Checking if we're getting a value when we expect one is fine for me ATM.

misl6 (approved):
Let's wait for the CI jobs to complete the run, but LGTM.
Thank you!

Issue #8653, opened 2024-03-21
Pyinstaller 6.5.0 not working with Kivy 2.3.0


ikus060:
**Software Versions**
* Python: 3.10
* OS: Linux
* Kivy: 2.3.0
* Kivy installation method: pip install kivy

**Describe the bug**

When trying to create a package using Pyinstaller, the hooks is failling with this error.  This problem was also repported in this discussion: https://github.com/orgs/kivy/discussions/1

**Expected behavior**

I'm expecting recent version of Pyinstaller to be working with Kivy.

**To Reproduce**

min.py:
```
# Importing required libraries
import kivy
from kivy.app import App
from kivy.uix.label import Label

# Defining the App class
class MinimalApp(App):
    def build(self):
        # Creating a Label widget
        label = Label(text="Hello, Minimal Kivy App!")
        return label

# Running the application
if __name__ == "__main__":
    MinimalApp().run()
```

Run pyinstaller command:
```
pyinstaller min.py
```

Here the full log of this problem:
```
$ pyinstaller min.py
89 INFO: PyInstaller: 6.5.0, contrib hooks: 2024.3
89 INFO: Python: 3.10.12
90 INFO: Platform: Linux-6.6.10-76060610-generic-x86_64-with-glibc2.35
90 INFO: wrote /home/ikus060/workspace/PDSL/minarca.git/min.spec
98 INFO: Extending PYTHONPATH with paths
['/home/ikus060/workspace/PDSL/minarca.git']
141 INFO: Kivy: v2.3.0
141 INFO: Logger: Record log in /home/ikus060/.kivy/logs/kivy_24-03-21_10.txt
[INFO   ] [Logger      ] Record log in /home/ikus060/.kivy/logs/kivy_24-03-21_10.txt
[INFO   ] [Kivy        ] v2.3.0
141 INFO: Kivy: Installed at "/home/ikus060/workspace/PDSL/minarca.git/.venv/lib/python3.10/site-packages/kivy/__init__.py"
[INFO   ] [Kivy        ] Installed at "/home/ikus060/workspace/PDSL/minarca.git/.venv/lib/python3.10/site-packages/kivy/__init__.py"
141 INFO: Python: v3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
[INFO   ] [Python      ] v3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
[...]

PR #8654, opened 2024-03-21
Fix PyInstaller hooks to work with v6.x.x #8653

ikus060:
Maintainer merge checklist

### DIFF
--- a/kivy/tools/packaging/pyinstaller_hooks/__init__.py
+++ b/kivy/tools/packaging/pyinstaller_hooks/__init__.py
@@ -78,7 +78,12 @@ try:
 except ImportError:
     kivy_deps = None
 from kivy.factory import Factory
-from PyInstaller.depend import bindepend
+try:
+    # Pyinstaller >= 6
+    from PyInstaller.depend.bindepend import get_imports
+except ImportError:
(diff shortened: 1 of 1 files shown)


misl6:
Question (from someone not very into Windows development):
- Do we need to support `PyInstaller<6.0.0` ? (Consider Kivy is going to have a major release, so we can break things, if well documented)

akshayaurora:
I would say considering the code to allow this is a few lines,  let the support for older pyinstaller be present for now. Maybe add a deprecated warning.. Can always remove/deprecate the support for older pyinstaller.

ikus060:
@misl6 
> Do we need to support PyInstaller<6.0.0

Since Pyinstaller introduce a couple of breaking chance that might impact some application, I would consider preferable to keep compatibility with Pyinstaller v5 and v6

misl6:
> @misl6
> 
> > Do we need to support PyInstaller<6.0.0
> 
> Since Pyinstaller introduce a couple of breaking chance that might impact some application, I would consider preferable to keep compatibility with Pyinstaller v5 and v6

Makes sense!

misl6 (approved):
LGTM. Thank you!

PR #8667, opened 2024-04-01
`Label`: Prevent overwriting `disabled_color` attribute when `color` is updated

ikus060:
Ensure that when setting rules for color and disabled_color, refrain from overwriting color attributes if the label is disabled, allowing the label to be drawn using the disabled_color value.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/label.py
+++ b/kivy/uix/label.py
@@ -379,6 +379,10 @@ class Label(Widget):
                 self._label.options['outline_color'] = (
                     self.disabled_outline_color if value else
                     self.outline_color)
+            elif self.disabled and name in ('color', 'outline_color'):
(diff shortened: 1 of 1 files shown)


misl6 (commented):
Nice catch!

What about something like:

```python
if self.disabled and name in ('color', 'outline_color'):
    # We should not update .....
    return
```

That will prevent us from triggering an un-needed texture update, and the code (IMHO) will look more readable and Pythonic 🐍❤️

(Please let me know what are your sincere thoughts)

ikus060:
Sure, I will do the change tomorrow...

misl6 (approved):
LGTM. Thank you!

Issue #8669, opened 2024-04-03
AnchorLayout prevents child widgets from controlling their height


nanouasyn:
**Software Versions**
* Python: 3.11.3
* OS: Linux Mint 19.1 Tessa
* Kivy: 2.3.0

**Describe the bug**
When a widget that sets its ``height`` based on its ``width`` is placed in an ``AnchorLayout``, the ``height`` of the widget does not change according to its ``width`` set by the ``AnchorLayout``. ``AnchorLayout`` controls the height of such a widget, ignoring the outright prohibition (``size_hint_y = None``).

**To Reproduce**

Put the ``Image`` widget, which increases its height based on its width and image proportions, inside the ``AnchorLayout``. Set ``size_hint_y`` to ``None`` to prevent ``AnchorLayout`` from controlling the height of the ``Image``.

```python
from kivy.lang import Builder
from kivy.app import runTouchApp

runTouchApp(Builder.load_string("""
AnchorLayout:
    anchor_y: 'center'
    Image:
        source: 'data/logo/kivy-icon-256.png'
        fit_mode: 'fill'
        size_hint_y: None
        height: self.width * self.texture_size[1] / self.texture_size[0]
"""))
```

It is expected that when changing the width of the ``AnchorLayout``, the ``Image`` will change its height. But in fact, the height of the ``Image`` does not change:

https://github.com/kivy/kivy/assets/27895729/09cc4994-f192-4c57-a73b-02e21554a810

**Expected behavior**

``Image`` should change size, filling the window in width and keeping the proportions.

https://github.com/kivy/kivy/assets/27895729/3032f785-775d-441d-8ec1-0e27b104983b

PR #8670, opened 2024-04-03
Fix AnchorLayout children height bug (#8669)

nanouasyn:
Maintainer merge checklist

#8669

``AnchorLayout`` saves the ``size`` of the child widget before recalculating it and putting it back. Setting ``size`` means that ``width`` will be set first, and then ``height`` will be set. If ``width`` handler sets a new ``height`` value, ``AnchorLayout`` will immediately replace it with the height value that it was originally going to set. This is not what it should do, in case it does not control the height of the child widget. It should not set the ``height`` at all if the ``size_hint_y`` value forbids him to change it. Even if it is going to set the same value as it was before, it is quite possible that in the process the value will change and the child widget will get the wrong ``height``.

### DIFF
--- a/kivy/uix/anchorlayout.py
+++ b/kivy/uix/anchorlayout.py
@@ -106,7 +106,7 @@ class AnchorLayout(Layout):
                     ch = shh_max
                 c.height = ch
 
-            cw, ch = self.size
+            cw, ch = c.size
 
             if anchor_x == 'left':
                 x = x + pad_left


misl6:
That makes sense!

This is the kind of issue that may show up again in the future if nothing blocks us from re-introducing the bug, would you consider creating a simple test that exercises the issue?

nanouasyn:
@misl6 I'm sorry. I can't build a lib from sources. Therefore, I can't run the tests. In this particular case, the fix was small and obvious, but writing tests blindly, without a clear understanding how testing works in the project, would be risky. In addition, all layouts are susceptible to such a potential error. Thus, we need to cover them all with tests. But as far as I can see, so far there are no such tests anywhere. Perhaps it is necessary to create an issue about the coverage of all layouts with such tests?

nanouasyn:
@misl6 It seems to me that the test can be like this:

```python
def test_anchorlayout_no_height_control(self):
    from kivy.uix.anchorlayout import AnchorLayout
    
    def set_height_equal_width(widget, value):
        widget.height = widget.width
    
    widget = Widget(size_hint=(1, None))
    widget.bind(width=set_height_equal_width)
    layout = AnchorLayout(size_hint=(None, None), size=(1000, 1000))
    layout.add_widget(widget)

    layout.do_layout()

    assert widget.size == [1000, 1000]
```

nanouasyn:
A new fix. Before this commit, the new positions of child widgets were calculated based on the ``height`` or ``width`` that either the widget had before we started resizing, or that we set ourselves. But if, for example, the ``height`` changed by itself, due to the fact that we changed the ``width``, the calculation of positions was based not on the new ``height``, but on the outdated ``height``. Therefore, it is necessary to reload the actual ``size`` after resizing so that the positioning remains correct in case the size in a dimension that we do not control has changed while resizing in another dimension. My previous commit corrects size errors in such situations, this one corrects positioning errors.

nanouasyn:
@misl6 It's done. Tests have also been added.

Issue #8678, opened 2024-04-12
`ConfigParser.write` takes 1 positional argument but 2 were given 


marinelay:
**Software Versions**
* Python: 3.9.18
* OS: Ubuntu 22.04
* Kivy: 2.3.0
* Kivy installation method: pip install

**Describe the bug**

https://github.com/kivy/kivy/blob/c0d4894384abb81cf4729984a1eaa4c437d266da/kivy/__init__.py#L482-L487

`Config` is object of `ConfigParser` and `ConfigParser.write` method does not require any positional arguments.
Thus, it always raises `TypeError: write() takes 1 positional argument but 2 were given` at line 484.
I believe setting `self.filename` as `kivy_config_fn` is more accurate like this:
```python
Config.filename = kivy_config_fn
Config.write()
```
I wonder it is the right behavior.

**Expected behavior**
```python
Config.filename = kivy_config_fn
Config.write()
```

**To Reproduce**
```python
from kivy.config import Config

with open("abc", 'w') as fd:
        Config.write(fd)
```

**Code and Logs and screenshots**
https://github.com/kivy/kivy/blob/c0d4894384abb81cf4729984a1eaa4c437d266da/kivy/__init__.py#L482-L487

misl6:
Nice catch!

I can confirm the issue is reproducible with `python examples/demo/touchtracer/main.py -m monitor --save`
I guess no one used the `--save` option for the last  ~12 years? 😅

Would you like to make a PR for it?

misl6:
Fixed via #8681

PR #8681, opened 2024-04-15
Fix `ConfigParser.write` error with  when `--save` is supplied as an argument

marinelay:
Fix issue #8678 

Maintainer merge checklist

### DIFF
--- a/kivy/__init__.py
+++ b/kivy/__init__.py
@@ -481,7 +481,8 @@ if not environ.get('KIVY_DOC_INCLUDE'):
     if need_save and 'KIVY_NO_CONFIG' not in environ:
         try:
             with open(kivy_config_fn, 'w') as fd:
-                Config.write(fd)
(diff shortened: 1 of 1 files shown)


kivy/__init__.py
```diff
@@ -481,7 +481,8 @@ def _patch_mod_deps_win(dep_mod, mod_name):
     if need_save and 'KIVY_NO_CONFIG' not in environ:
         try:
             with open(kivy_config_fn, 'w') as fd:
```
misl6, line 483: The file gets opened with the proper permissions and encoding into the `write` method, so no need to open it.

misl6 (changes requested):

kivy/__init__.py
marinelay: I'm sorry for my mistake! I've fixed it.

misl6 (approved):
LGTM. Thank you!

PR #8682, opened 2024-04-15
Fix `xsel` import when clipboard is empty, and add timeout failsafe

clayote:
My Kivy installation was hanging on import, and it turned out to be because `xsel` wasn't sending anything through its pipe -- because the selection was empty. It looked like `xsel` only got called to make sure it was *there*, so I added a `--version` argument, which will return something with 0 exit code if `xsel` is available.

I also added a timeout to the subprocess, because it seems bad that kivy just hangs when the subprocess does. If `xsel` is broken, Kivy should just use another clipboard provider, rather than locking up.

I removed the bare `except` because it did nothing.

### DIFF
--- a/kivy/core/clipboard/clipboard_xsel.py
+++ b/kivy/core/clipboard/clipboard_xsel.py
@@ -13,7 +13,7 @@ if platform != 'linux':
 
 try:
     import subprocess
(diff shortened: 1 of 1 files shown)


misl6 (approved):
LGTM. Thank you!

PR #8686, opened 2024-04-17
Describe graphics stack more accurately in README.md

clayote:
Kivy *supports* OpenGL ES 2.0, but it's mostly *built on* SDL these days. I believe my build goes through Vulkan.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ whilst making your code reusable and deployable: Innovative user interfaces made
 easy.
 
 Kivy is written in Python and [Cython](https://cython.org/) and is built on
-[OpenGL ES 2.0](https://www.khronos.org/opengles/). It supports various input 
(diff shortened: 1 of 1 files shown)


misl6 (changes requested):
Hi @clayote !

SDL for Kivy is just one (even if it's the default on most platforms) of the window provider implementations.

Then, we get an OpenGL / OpenGLES context (natively, or via ANGLE), and OpenGL/OpenGLES is still used under the hood. (see https://github.com/kivy/kivy/tree/master/kivy/graphics)

Feel free to describe the stack more accurately, but Kivy is not built on top of SDL, and OpenGL is still used.

clayote:
Oh.

Well, since I misunderstood something so basic, I probably shouldn't be the one to describe it, at the moment...

Issue #8363, opened 2023-08-28
Kivy references deprecated "renpy"


RobertFlatt:
**Software Versions**
* Python:
* OS:
* Kivy:  master
* Kivy installation method:

**Describe the bug**
Kivy.metrics has 2 instances of `autoclass('org.renpy.android.PythonActivity')`
These should be autoclass('org.kivy.android.PythonActivity') as "renpy" is deprecated.

https://github.com/kivy/kivy/blob/master/kivy/metrics.py#L195
https://github.com/kivy/kivy/blob/master/kivy/metrics.py#L257

These result in run time errors on Android.

**Code and Logs and screenshots**
```
08-28 17:03:31.600 29390 29706 I python  :    File "/home/feluk6174/python/llista/.buildozer/android/platform/build-arm64-v8a_armeabi-v7a/build/python-installs/llista/armeabi-v7a/kivy/metrics.py", line 195, in get_dpi
08-28 17:03:31.602 29390 29706 I python  :    File "/home/feluk6174/python/llista/.buildozer/android/platform/build-arm64-v8a_armeabi-v7a/build/python-installs/llista/armeabi-v7a/jnius/reflect.py", line 209, in autoclass
08-28 17:03:31.603 29390 29706 I python  :    File "jnius/jnius_export_func.pxi", line 24, in jnius.jnius.find_javaclass
08-28 17:03:31.605 29390 29706 I python  :    File "jnius/jnius_export_class.pxi", line 266, in jnius.jnius.JavaClass.__init__
08-28 17:03:31.607 29390 29706 I python  :  AttributeError: 'Class' object has no attribute '_JavaClass__cls_storage'
08-28 17:03:31.607 29390 29706 I python  : Python for android ended.
```

OHAProject:
"I have encountered this  issue that prevents me from using my app, and I'm looking for a solution."

HyTurtle:
Hardware is still accessed through renpy (it's not PythonActivity on those mentioned lines), it's not fully depreciated - something else is going on here.

RobertFlatt:
Could be. 
Since this issue I have seen more cases of this message, and it is not a message I usually see.
Don't know what the common denominator is.

These are the other cases:

https://groups.google.com/g/kivy-users/c/5JHi0lCoAlI/m/v49UD507AwAJ
https://discord.com/channels/423249981340778496/712344698559397895/1146462240195756205
https://discord.com/channels/423249981340778496/423250272316293120/1146358878318100541

[...]

PR #8727, opened 2024-05-14
Refactoring some obsolete functions that generate compilation errors on Android (fix issue #8363)

JLucasXavier:
I carried out some refactorings in functions that use the jnius library according to Android references. This should resolve issue #8363


misl6:
Hi, I see we have multiple references to `org.renpy.android.*` out there (and quite sure some code is outdated / unused), but we can change all these references?

PR #8741, opened 2024-05-26
`avplayer` audio provider: Fixes `seek` and `state` on iOS and macOS

tito:
The previous implementation was using playAtTime_, but 100% sure it was not working as playAtTime_ schedules the audio to play at a specific time, not to seek within the current audio file. To seek to a specific position in the audio, we should use the currentTime property of AVAudioPlayer.

In addition, i added a delegate to correctly catch when the sound is stopping when reaching eof, which make the state property correctly work.

Maintainer merge checklist


misl6 (approved):
Tested on macOS, works great.
Updated title.

LGTM. Thank you!

PR #8799, opened 2024-08-08
Add static method `from_bytes` to `SoundSDL2` to initialize an instance from audio sample in memory

sassanh:
Kivy allows loading audio samples from filesystem by filename using the `source` property of `Sound` object, but sometimes the audio samples are generated in the memory, for example it happens when using text to speech engines like Picovoice's Orca or Piper and in environments like RPi running on a SD-card, writing the samples from memory to filesystem in a temporary file and then asking Kivy to instruct SDL to read that file again into memory may cause some unwanted side effects like delay or exhaustion of the SD-card.

`from_bytes` static method added in this pull request, lets the user instantiate `SoundSDL2` with data coming directly from memory.

I considered re-using the `source` property by adding a "mem:" prefix or something like that as a signal for `load` method, but it's a `StringProperty` and only accepts `str`, not `bytes`.

Maintainer merge checklist

### DIFF
--- a/kivy/core/audio_output/audio_sdl2.pyx
+++ b/kivy/core/audio_output/audio_sdl2.pyx
@@ -43,6 +43,8 @@ __all__ = ('SoundSDL2', 'MusicSDL2')
 
 include "../../../kivy/lib/sdl2.pxi"
 include "../../../kivy/graphics/common.pxi"  # For malloc and memcpy (on_pitch)
+from cpython cimport PyBytes_AsString, PyBytes_FromStringAndSize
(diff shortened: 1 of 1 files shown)


misl6:
Hi @sassanh ;

We applied the migration from SDL2 to SDL3, can you please check if this change is still required on the latest `master` and apply the changes accordingly?

Feel free to ping me when ready!

sassanh:
Hi @misl6,

I think the failed test has nothing to do with the changes introduced in this pull request. Please let me know if I'm mistaken.

Issue #8557, opened 2024-01-09
Fails to build with clang 17 and GCC 14 due to -Wincompatible-pointer-types


kanavin:
When building 2.3.0 with clang, the following error occurs:
```
/mnt/b/yoe/master/build/tmp/work/core2-64-yoe-linux-musl/python3-kivy/2.3.0/git/kivy/graphics/cgl_backend/cgl_gl.c:4549:52: error: incompatible function pointer types assigning to 'void (*)(GLuint, GLsizei, const GLchar **, const GLint *)' (aka 'void (*)(unsigned int, int, const char **, const int *)') from 'void (GLuint, GLsizei, const GLchar *const *, const GLint *)' (aka 'void (unsigned int, int, const char *const *, const int *)') [-Wincompatible-function-pointer-types]
 4549 |   __pyx_v_4kivy_8graphics_3cgl_cgl->glShaderSource = glShaderSource;
      |                                                    ^ ~~~~~~~~~~~~~~
```

Full build log:
https://sprunge.us/lq5PKS

misl6:
Hi @kanavin ,

This one could be related: https://github.com/kivy/kivy/pull/8415


Can you please confirm your clang version?

kanavin:
> Hi @kanavin ,
> 
> This one could be related: #8415
> 
> Can you please confirm your clang version?

@kraj I think this was this with 17.x?

kraj:
> > Hi @kanavin ,
> > This one could be related: #8415
> > Can you please confirm your clang version?
> 
> @kraj I think this was this with 17.x?

yes, 17.0.6 to be precise

misl6:
As pointed out by @glaubitz in #8746 , the same issue is happening on GCC 14. (Issue title updated)

glaubitz:
I just tried to have a go at this and tried the following change:

```diff
diff -Nru kivy-2.3.0.orig/kivy/core/window/_window_sdl2.pyx kivy-2.3.0/kivy/core/window/_window_sdl2.pyx
--- kivy-2.3.0.orig/kivy/core/window/_window_sdl2.pyx   2024-01-05 10:40:12.000000000 +0100
+++ kivy-2.3.0/kivy/core/window/_window_sdl2.pyx        2024-06-11 09:59:18.605347161 +0200
@@ -263,7 +263,7 @@
         for joy_i in range(SDL_NumJoysticks()):
             SDL_JoystickOpen(joy_i)
 
[...]

PR #8911, opened 2024-11-04
Revert #8415 and align `glShaderSource` `typedef` in `common_subset.h` with Khronos Headers

cederom:
* Problem and solution reported here https://github.com/kivy/kivy/issues/8557.
* Build is broken with CLANG17 and GCC14 for Kivy 2.3.0+ (including master).
* @danfe (Alexey Dokuchaev) created this patch for FreeBSD port.
* This patch proves danfe's fix and just reverts breaking commit https://github.com/kivy/kivy/pull/8415.
* THIS IS PROOF OF CONCEPT ONLY.

Maintainer merge checklist


misl6:
I will be using the headers from the [KhronosGroup/OpenGL-Registry](https://github.com/KhronosGroup/OpenGL-Registry) as the authoritative source. All headers across all available versions consistently use the `const GLchar *const* string` form, rather than `const GLchar** string`.

This PR not only reverts #8415 but also updates the `glShaderSource` typedef in `common_subset.h`. The previous typedef incorrectly used `const GLchar** string`, which could have caused compile-time issues when `common_subset.h` was used (see [gl_redirect.h](https://github.com/kivy/kivy/blob/6aa77eec1a8e17f869e24ad635003f4d251bd2f4/kivy/include/gl_redirect.h)).

If you encounter any issues with this PR breaking your build, please open a new issue and include the relevant snippets from your OpenGL/OpenGLES headers so we can better diagnose the problem.

misl6 (approved):
LGTM. Thank you!

(See https://github.com/kivy/kivy/pull/8911#issuecomment-2466262722 for further info)

PR #8939, opened 2024-12-04
Fix a crash when loading a lot of images at once

clayote:
I tried using `kivy.core.image.Image` to load a bunch of images at once. I feel like this might not be the greatest idea for my app's sake, but in any case, doing this caused a crash as Kivy's cache tried to unload things while another thread was loading things into it.

The fix is one line. May as well?

Maintainer merge checklist

### DIFF
--- a/kivy/cache.py
+++ b/kivy/cache.py
@@ -193,7 +193,7 @@ class Cache(object):
         import heapq
         time = Clock.get_time()
         heap_list = []
-        for key in Cache._objects[category]:
+        for key in list(Cache._objects[category]):
             obj = Cache._objects[category][key]
             if obj['lastaccess'] == obj['timestamp'] == time:
                 continue


misl6 (approved):
LGTM. Thank you!

kivy/cache.py
```diff
@@ -193,7 +193,7 @@ def _purge_oldest(category, maxpurge=1):
         import heapq
         time = Clock.get_time()
         heap_list = []
-        for key in Cache._objects[category]:
+        for key in list(Cache._objects[category]):
             obj = Cache._objects[category][key]
```
cclauss, line 197: Lines 196 and 197 could have been unified for performance into:
```python
for key, obj in Cache._objects[category].items():
```

This might have solved this issue.

[`ruff rule PLC0206`](https://docs.astral.sh/ruff/rules/dict-index-missing-items)
# dict-index-missing-items (PLC0206)

Derived from the **Pylint** linter.

## What it does
Checks for dictionary iterations that extract the dictionary value
via explicit indexing, instead of using `.items()`.

## Why is this bad?
Iterating over a dictionary with `.items()` is semantically clearer
and more efficient than extracting the value with the key.

## Example
```python
ORCHESTRA = {
    "violin": "strings",
    "oboe": "woodwind",
    "tuba": "brass",
    "gong": "percussion",
}

for instrument in ORCHESTRA:
    print(f"{instrument}: {ORCHESTRA[instrument]}")
```

Use instead:
```python
ORCHESTRA = {
    "violin": "strings",
    "oboe": "woodwind",
    "tuba": "brass",
    "gong": "percussion",
}

for instrument, section in ORCHESTRA.items():
    print(f"{instrument}: {section}")
```

PR #8944, opened 2024-12-08
Python lint: Use `ruff` instead of `flake8`

cclauss:
https://docs.astral.sh/ruff is an extremely fast Python linter and code formatter, written in Rust. With [over 800 linting rules](https://docs.astral.sh/ruff/rules), ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [Black](https://github.com/psf/black), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [pyupgrade](https://pypi.org/project/pyupgrade/), [autoflake](https://pypi.org/project/autoflake/), and more, all while executing tens or hundreds of times faster than any individual tool.

Maintainer merge checklist

### DIFF
--- a/kivy/network/urlrequest.py
+++ b/kivy/network/urlrequest.py
@@ -75,12 +75,11 @@ from kivy.weakmethod import WeakMethod
 try:
     import ssl
 
-    HTTPSConnection = None
     from http.client import HTTPSConnection
 except ImportError:
     # depending the platform, if openssl support wasn't compiled before python,
     # this class is not available.
-    pass
+    HTTPSConnection = None
 
 
 # list to save UrlRequest and prevent GC on un-referenced objects
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,9 +40,9 @@ dev =
     kivy_deps.gstreamer_dev~=0.3.3; sys_platform == "win32"
     kivy_deps.sdl2_dev~=0.8.0; sys_platform == "win32"
     kivy_deps.glew_dev~=0.3.1; sys_platform == "win32"
-    flake8
     pre-commit
     responses
+    ruff>=0.8.3,<1.0.0
 base =
     pillow>=9.5.0,<11
 media =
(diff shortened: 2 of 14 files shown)


cclauss:
@misl6 Your review, please.

kivy/network/urlrequest.py
```diff
@@ -75,7 +75,6 @@ def bug_posted(req, result):
 try:
     import ssl
 
-    HTTPSConnection = None
```
misl6, line 78: Later into the code, `HTTPSConnection` is checked to not being `None` as we support this case.

setup.cfg
```diff
@@ -41,9 +41,9 @@ dev =
     kivy_deps.gstreamer_dev~=0.3.3; sys_platform == "win32"
     kivy_deps.sdl2_dev~=0.8.0; sys_platform == "win32"
     kivy_deps.glew_dev~=0.3.1; sys_platform == "win32"
-    flake8
     pre-commit
     responses
+    ruff
```
misl6, line 46: How about pinning a version (can be relaxed in both files, and not necessarily `~=`) as `cicd-requirements.txt` ?

misl6 (changes requested):
Hi @cclauss !

Thank you for taking care of it! I really appreciate!

Left a couple of comments, let me know when you're ready for another run.

setup.cfg
cclauss: Does this mean `ruff==0.8.3` (done now) or `ruff==0.8.x` or something else.

misl6: `~=` allows `0.8.*`, so is more relaxed than `==`.

I would suggest a `> 0.8.3` and `< 1.0.0` here, as a breakage does not impact any user, but it might just impact CI/CD pipelines and dev tests.

cclauss:
Are other changes needed?

misl6 (approved):
LGTM. Thank you!

Issue #8023, opened 2022-09-30
TouchRippleButtonBehavior on_disabled failed 


yumoqing:
**Software Versions**
* Python: 3.9.13
* OS: mac os big sur
* Kivy: 2.1.0
* Kivy installation method:

code:
```
from kivy.app import App 
from kivy.uix.behaviors.touchripple import TouchRippleButtonBehavior
from kivy.uix.boxlayout import BoxLayout

class TestWidget(TouchRippleButtonBehavior, BoxLayout):
    def __init__(self, **kw):
        super().__init__(**kw)
        self.disabled = True

class TestApp(App):
    def build(self):
        return TestWidget()

if __name__ == '__main__':
    TestApp().run()
```
**Describe the bug**
the code above run in exception:
```
 Traceback (most recent call last):
   File "/Volumes/home/ymq/pydev/github/glib/src/scripts/t.py", line 15, in <module>
     TestApp().run()
   File "/usr/local/lib/python3.9/site-packages/kivy/app.py", line 954, in run
     self._run_prepare()
   File "/usr/local/lib/python3.9/site-packages/kivy/app.py", line 924, in _run_prepare
     root = self.build()
   File "/Volumes/home/ymq/pydev/github/glib/src/scripts/t.py", line 12, in build
     return TestWidget()
   File "/Volumes/home/ymq/pydev/github/glib/src/scripts/t.py", line 8, in __init__
     self.disabled = True
   File "kivy/properties.pyx", line 520, in kivy.properties.Property.__set__
   File "kivy/properties.pyx", line 1662, in kivy.properties.AliasProperty.set
   File "/usr/local/lib/python3.9/site-packages/kivy/uix/widget.py", line 1573, in set_disabled
     self.inc_disabled()
   File "/usr/local/lib/python3.9/site-packages/kivy/uix/widget.py", line 1581, in inc_disabled
     self.property('disabled').dispatch(self)
   File "kivy/properties.pyx", line 608, in kivy.properties.Property.dispatch
   File "kivy/properties.pyx", line 625, in kivy.properties.Property.dispatch
   File "kivy/properties.pyx", line 1672, in kivy.properties.AliasProperty._dispatch
   File "kivy/_event.pyx", line 1307, in kivy._event.EventObservers.dispatch
   File "kivy/_event.pyx", line 1213, in kivy._event.EventObservers._dispatch
[...]

PR #8979, opened 2025-01-14
Fix the `AttributeError` that occurs when `TouchRippleButtonBehavior` is disabled.

gottadiveintopython:
Maintainer merge checklist

(This is not just a bug fix, as `TouchRippleBehavior` now calls `ripple_fade()` instead of `TouchRippleButtonBehavior` when the widget is disabled.)

Fixes: #8023

### DIFF
--- a/kivy/uix/behaviors/touchripple.py
+++ b/kivy/uix/behaviors/touchripple.py
@@ -137,6 +137,7 @@ class TouchRippleBehavior(object):
         super(TouchRippleBehavior, self).__init__(**kwargs)
         self.ripple_pane = CanvasBase()
         self.bind(
+            disabled=self._ripple_fade,
             ripple_color=self._ripple_set_color,
             ripple_pos=self._ripple_set_ellipse,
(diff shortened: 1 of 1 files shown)


kivy/uix/behaviors/touchripple.py
```diff
@@ -138,6 +138,7 @@ def __init__(self, **kwargs):
         self.ripple_pane = CanvasBase()
         self.canvas.add(self.ripple_pane)
         self.bind(
+            disabled=lambda widget, value: widget.ripple_fade() if value else None,
```
misl6, line 141: Minor: Instead of using a lambda, can we move this logic into a method?

gottadiveintopython: Okay, thank you!

gottadiveintopython: Updated the PR. Should I change the `def ripple_fade(self)` to `def ripple_fade(self, *args)` instead of adding the `def _ripple_fade(self, *args)`?

PR #8981, opened 2025-01-21
Fix twisted reactor shutdown when using twisted > 24.3.0

### LINKED ISSUE #8821: Kivy 2.3.0 support reactor shutdown error twisted 24.7.0
**Software Versions** Twisted 24.7.0
* Python: 3.11
* OS: Windows 10
* Kivy: 2.3.0

**Describe the bug**
Upgraded to twisted==24.7.0
Kivy application integrated with twisted reactor fails on application shutdown.
File support.py line 230 method call _mainLoopShutdown(self) fails with missing attribute error in twisted internet._threadedselect.py
Checking the source compared to revision 24.3.0 the previous released version, shows the method has been removed.
Because of the exception the program doesn't shutdown

I rolled back to twisted 24.3.0 all good.

**Expected behavior**
No exception when program stopped.

**To Reproduce**
Example echo server app run then observe problem when it is stopped.

```python
Get the echo server application example from the reactor integration to kivy docs
I did not modify.

Other stuff removed.....

   File "C:\Users\ronmc\Dev\venv\Lib\site-packages\kivy\support.py", line 230, in reactor_stop
     reactor._mainLoopShutdown()
     ^^^^^^^^^^^^^^^^^^^^^^^^^
 AttributeError: 'ThreadedSelectReactor' object has no attribute '_mainLoopShutdown'

```

**Additional context**

rnixx:
Fixes #8821

### DIFF
--- a/kivy/support.py
+++ b/kivy/support.py
@@ -227,7 +227,11 @@ def install_twisted_reactor(**kwargs):
             Logger.info("Support: Stopping twisted threads")
             reactor.threadpool.stop()
         Logger.info("Support: Shutting down twisted reactor")
-        reactor._mainLoopShutdown()
+        # twisted <= 24.3.0
+        if hasattr(reactor, '_mainLoopShutdown'):
+            reactor._mainLoopShutdown()
+        else:
+            reactor._uninstallHandler()
         try:
             reactor.stop()
         except ReactorNotRunning:


misl6 (approved):
Well ...

I do not see from twisted code a cleaner way to do that, in order to avoid long-term breakages.

LGTM for now, maybe, if this is the only way to get it working ( I know nothing about twisted and consequently about the integration with Kivy ) we can propose them a public API?

rnixx:
@misl6 I already got an answer about the issue here https://github.com/twisted/twisted/pull/12207#issuecomment-2608130976. Will try to act on this asap.

PR #8983, opened 2025-01-25
Coreimage fix to add more types than just BytesIO

kuzeyron:
Added changes to the CoreImage so we can use other types too. Example would be to make this [code](https://gist.github.com/kuzeyron/30b5b646b7b1a37109a37e02232e45a8) able to use rawIO instead of BytesIO.
Tested locally on my machine (Linux, pil and sdl2 provider) with v3.0.0.dev0.

Maintainer merge checklist


misl6:
Seen it includes changes to `sdl2`, as the API changed a bit, and we're near to making the switch to `sdl3`, can you ping me back when `sdl3` is in and is rebased on top of it?

kuzeyron:
Sure thing!

akshayaurora (approved):

misl6:
Hi @kuzeyron;

We applied the migration from SDL2 to SDL3, can you please check if this change is still required on the latest `master` and apply the changes accordingly?

Feel free to ping me when ready!

akshayaurora:
@kuzeyron seems it needs updating to support img_sdl3 ?

kuzeyron:
> @kuzeyron seems it needs updating to support img_sdl3 ?

I have to create another PR and close this one. The reason why this one is still open is because I need to use it as a model.

PR #8996, opened 2025-02-16
🐛 fix: KV Parser for f-strings now properly resolves nested attribute names

FilipeMarch:
Previously this would not work as expected, i.e., even if `alice.age` is updated, the `Label` text would remain the same.
```python
from kivy.app import App
from kivy.lang import Builder


class Person(EventDispatcher):
    age = NumericProperty()


kv = """
BoxLayout:
    orientation: 'vertical'
    Button:
        text: 'Make Alice older'
        on_press: app.alice.age += 1
    Label:
        text: f'Alice is {app.alice.age} years old' # the only important line
"""


class MainApp(App):
    alice = Person(age=25)

    def build(self):


MainApp().run()
```

For instance, since the following code works, I don't think the behavior of f-strings in KV should be any different:
```yml
Label:
    text: 'Alice is ' + str(app.alice.age) + ' years old'
```

Previously, the `ParserRuleProperty.get_names_from_expression` method only handled simple attribute accesses by checking if an `ast.Attribute` node's value was an `ast.Name`. This meant that multi-level attribute chains (e.g., `app.some_object.some_property`) were not fully extracted, leading to incomplete dependency tracking for KV f-string expressions.

This commit updates the handling of `ast.Attribute` nodes by:
- Recursively traversing the attribute chain to collect all attribute names.
- Reversing the collected parts to build a complete dotted attribute path.
- Yielding the full attribute (e.g., "root.person.age") for proper binding.
- Adding a return statement to ensure that once the `ast.Attribute` node is processed, no further processing occurs on the same node, avoiding duplicate or incorrect yields.

This change ensures that KV expressions with multi-level attribute references update dynamically as expected.
Maintainer merge checklist

### DIFF
--- a/kivy/lang/parser.py
+++ b/kivy/lang/parser.py
@@ -276,7 +276,7 @@ class ParserRuleProperty(object):
             yield from cls.get_names_from_expression(node.iter.value)
 
         if isinstance(node, ast.Attribute):
-            # Build the full dotted attribute path (e.g. "root.object.property")
+            # Build full dotted attribute path (e.g. "root.object.property")
             attribute_parts = []
             current_node = node
             while isinstance(current_node, ast.Attribute):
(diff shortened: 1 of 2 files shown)


misl6 (approved):
LGTM. Thank you!

PR #9000, opened 2025-02-24
[WIP] Remove deprecated `padding_x` and  `padding_y` in Label

misl6:
Maintainer merge checklist

Ref: #8177 

⚠️ DO NOT MERGE YET. MISSES MIGRATION DOCS AND IMPROVED TESTING ⚠️

### DIFF
--- a/examples/canvas/lines.py
+++ b/examples/canvas/lines.py
@@ -136,7 +136,7 @@ Builder.load_string('''
                     allow_no_selection: False
                     size_hint_x: None
                     width: self.texture_size[0]
-                    padding_x: '5dp'
+                    padding: '5dp', 0, '5dp', 0
                     on_state:
                         if self.state == 'down': root.dashes = []
                         if self.state == 'down': root.dash_length = 1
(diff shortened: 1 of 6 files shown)


gottadiveintopython:
I wonder if we should replace the properties with ones that always raise an exception when accessed, rather than remove them, so that we don't silently break existing code?

```python
import pytest


class PropertyRemovedError(Exception):
    pass


class RemovedProperty:
    def __init__(self, version_removed: str):
        self.version_removed = version_removed

    def __set_name__(self, owner, name):
        self.name = name
        self.owner = owner

    def __get__(self, *args):
        raise PropertyRemovedError(f"Property '{self.owner.__name__}.{self.name}' was removed in version {self.version_removed}")

    __set__ = __delete__ = __get__


class Test(Widget):
    value = RemovedProperty("0.11.1")


def test_set_via_kwargs():
    # Unfortunately, not PropertyRemovedError
    with pytest.raises(TypeError):
        Test(value=1)


def test_set():
    t = Test()
    with pytest.raises(PropertyRemovedError):
        t.value = 1


def test_delete():
    t = Test()
    with pytest.raises(PropertyRemovedError):
        del t.value


def test_read():
    t = Test()
    with pytest.raises(PropertyRemovedError):
        t.value


def test_set_via_kv():
    from textwrap import dedent
    from kivy.lang import Builder

    with pytest.raises(PropertyRemovedError):
        Builder.load_string(dedent("""
            Test:
                value: 1
            """))


if __name__ == "__main__":
    pytest.main([__file__])
```

(I first tried using `AttributeError` instead of a custom error, but `test_set_via_kv` didn't pass)

Issue #8770, opened 2024-07-01
ZeroDivisionError in mouse.py input provider


massimilianodelliubaldini:
**Software Versions**
* Python: 3.10.0
* OS: Windows 10 22H2
* Kivy: 2.3.0
* Kivy installation method: via pip

**Describe the bug**
The `create_hover` function in mouse.py sets values `nx` and `ny` to divide themselves by `win._density`, which may be zero. When this happens, a stack trace is thrown and the app crashes.

**Expected behavior**
Similar fixes to be applied as described by https://github.com/kivy/kivy/issues/7006#issuecomment-900231781 to prevent the error from being thrown. I'm unaware of downstream impacts of not performing the division at all, just here to report the issue. 

**To Reproduce**
Unable to reproduce as it seems to be intermittent. Reporting on behalf of another user with attached screenshot.

**Code and Logs and screenshots**
![image](https://github.com/kivy/kivy/assets/8584296/5905d53e-c70a-49d7-96a5-f4707f2b0d1e)
![image](https://github.com/kivy/kivy/assets/8584296/57589539-602f-40c7-a8b4-61e49f62fa55)

**Additional context**
Let me know if you need any other details. Thank you!

mp-007:
@massimilianodelliubaldini 

Not 100% sure, but I think sometime GetDpiForWindow can return 0 when there is not valid active window (ex: active window is None)

source:
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getdpiforwindow

So, in this case _update_density_and_dpi in window_sld2 will set a _density at 0.
```
    def _update_density_and_dpi(self):
        if platform == 'win':
            from ctypes import windll
            self._density = 1.
            try:
                hwnd = windll.user32.GetActiveWindow()
                self.dpi = float(windll.user32.GetDpiForWindow(hwnd))
                self._density = self.dpi / 96
            except AttributeError:
                pass
        else:
            self._density = self._win._get_gl_size()[0] / self._size[0]
            if self._is_desktop:
                self.dpi = self._density * 96.
```

to fix this, maybe this condition can be made

```
    def _update_density_and_dpi(self):
        if platform == 'win':
            from ctypes import windll
            self._density = 1.
            try:
                hwnd = windll.user32.GetActiveWindow()
                dpi = float(windll.user32.GetDpiForWindow(hwnd))
                if dpi:
                    self.dpi = dpi
                    self._density = self.dpi / 96
            except AttributeError:
                pass
[...]

PR #8865, opened 2024-10-24
Update window_sdl2.py to avoid a density at 0 in windows.

mp-007:
try to fix issue #8770. If GetDpiForWindow return 0 when there is not valid active window (ex: active window is None), _density will be 0.

Maintainer merge checklist
* [ ] Title is descriptive/clear for inclusion in release notes.
* [ ] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [ ] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #9029, opened 2025-04-07
Fallback mechanism for `SDL2` window creation when `GetDpiForWindow` returns `0` ( on Windows)

mp-007:
change branch for PR #8865

try to fix issue https://github.com/kivy/kivy/issues/8770. If GetDpiForWindow return 0 when there is not valid active window (ex: active window is None), _density will be 0.

Maintainer merge checklist

### DIFF
--- a/kivy/core/window/window_sdl2.py
+++ b/kivy/core/window/window_sdl2.py
@@ -404,9 +404,7 @@ class WindowSDL(WindowBase):
             except AttributeError:
                 pass
         else:
-            self._density = (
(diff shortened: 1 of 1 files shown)


misl6 (commented):
Hi @mp-007 ,

From tests looks like it fails to create a Window, does it works on your side locally?

mp-007:
> Hi @mp-007 ,
> 
> From tests looks like it fails to create a Window, does it works on your side locally?

There was an unwanted change in my pull request. Now all the tests are passing.

misl6 (approved):
LGTM. Thank you!

Issue #4179, opened 2016-04-16
kivy-mode keeps indenting


elvis-sik:
I open an empty file in emacs with kivy mode set. I type

```
Label:
```

and press `Enter`. It indents label 4 spaces to the right.

I find this annoying. Is this known/expected behavior?

inclement:
I'm fairly sure this is expected and generally desirable - usually when a line ends with a colon, the next line will be a child widget or property of what you just wrote.

elvis-sik:
I'm sorry, I think I didn't express myself well. The line that gets indented is the one with `Label:` itself. The following line (the one in which the cursor ends up) ends up indented by 8 spaces.

inclement:
Oh, I see. That sounds like a bug to me, but is not happening for me. I think I might have installed kivy-mode via package-install (probably from MELPA), does that version work for you?

elvis-sik:
I'm trying to investigate. Learning how to use emacs's debugger and how to read elisp. I'm using spacemacs. kivy-mode installed in a folder called `elpa` so that's probably where it came from.

elvis-sik:
I'm confused. Maybe someone more experienced can help me. I have evaluated `(debug-on-entry 'kivy-indent-line)`. Then I create a file `foo.kv` and type `Foo:` and press `Enter`. The debugger starts. I eventually get to this [call-stack.txt](https://github.com/kivy/kivy/files/253182/call-stack.txt)

I press `d` until the function `syntax-ppss-flush-cache` returns. Then `Foo:` gets indented 4 spaces. The strange thing is if I call myself the function (by evaluating `(syntax-ppss-flush-cache 1 1)`) it doesn't indent `Foo:`.

misl6:
Closing as can't reproduce (https://github.com/kivy/kivy/pull/9051#issuecomment-2913182673), feel free to re-open if the issue persists.

PR #9051, opened 2025-05-05
Change Emacs `kivy-mode` parent mode to `prog-mode`

colebrooks:
The included Emacs major mode for editing kv lang source code was based on the built in Emacs mode **fundamental-mode**, however, per [the Emacs documentation](https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Major-Modes.html), custom major modes should be derived from one of **text-mode**, **prog-mode**, or **special-mode** if applicable, or **fundamental mode** _only if none of those three are applicable_. prog-mode is intended for buffers that contain programming language source code, so naturally it is appropriate to use prog-mode as the parent of kivy-mode. 

This has the benefit of keeping the same minor modes bound to prog-mode active in kivy-mode buffers, while using fundamental-mode requires Emacs users to bind those minor modes to the kivy-mode-hook in addition to the prog-mode-hook. For example, a popular minor mode for prog-mode is **highlight-indent-guides-mode**. By using prog-mode as the parent for kivy-mode, this will also be activated in kivy-mode, without requiring it to be explicitly bound to kivy-mode.

### DIFF
--- a/kivy/tools/highlight/kivy-mode.el
+++ b/kivy/tools/highlight/kivy-mode.el
@@ -153,7 +153,7 @@
 
 
 ;;;###autoload
-(define-derived-mode kivy-mode fundamental-mode "kivy"
+(define-derived-mode kivy-mode prog-mode "kivy"
   "Simple mode to edit kivy.
 
 \\{kivy-mode-map}"


misl6:
Hi @colebrooks;

Consider me as a Emacs noob, does this change (which makes sense according to docs and your explanation ) may have also fixed https://github.com/kivy/kivy/issues/4179 ?

PR #9054, opened 2025-05-12
Improved animation example

grnlogic:
I have added speed control and reset functionality to the animation example.
- "Fast"/ "Slow" button to change the duration of the animation
- "Reset" button to restore the position and size of the button
- The code comes with clear comments for new functions

This modifies the example in the "Component: Examples" section and does not change the API.


misl6 (changes requested):
Hi @grnlogic,

Thank you for the improvements, however, looks like the label for the change speed button is set to the wrong value at startup, can you take care of it?

Issue #4189, opened 2016-04-18
Ongoing scroll effects are not canceled when scroll_x or scroll_y is set


dessant:
Found by @kived while testing https://github.com/kivy/kivy/pull/4180#issuecomment-211042902.

kived:
Note that the scroll effects also set `scroll_x`/`scroll_y` via a callback. So fixing this would probably require either a `no_cancel` flag to set in the callback, or separate `_scroll_x`/`_scroll_y` properties, which would not cancel while the `scroll_x`/`scroll_y` properties would.

Julian-O:
Is this the same as #4486?

@kived and @dessant: Is this still a problem with Kivy 2.2.1?

dessant:
Hi @Julian-O, I was a core dev a long time ago, but I no longer use Kivy. There are some open issues left that were created by me, and the reported bugs were mostly confirmed at the time by other users, but you will probably need to triage them again.

https://github.com/kivy/kivy/issues/created_by/dessant

Julian-O:
Thanks dessant. I'll try not to bother you again.

PR #9063, opened 2025-05-25
Fix issues with Scrolling conflicts with scroll bars

akshayaurora:
Original Issue.

When scrollbar is already scrolling using either middle mouse button with smooth scrolling or when scrolling with content.
- The active scrolling effect is not cancelled when user clicks on the scroll bars,
-  ScrollView momentarily moves to the position user clicked and the continues earlier scrolling effect.

This fixes that issue,
- Stops the active scrolling and
- Resets scroll position to the place user has touched on the scrollview.

Possibly relevant issues. https://github.com/kivy/kivy/issues/4189

Maintainer merge checklist

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -798,14 +798,24 @@ class ScrollView(StencilView):
             return self.simulate_touch_down(touch)
 
         if in_bar:
+            e = None
             if (ud['in_bar_y'] and not
                     self._touch_in_handle(
                         self._handle_y_pos, self._handle_y_size, touch)):
                 self.scroll_y = (touch.y - self.y) / self.height
+                e = self.effect_y
             elif (ud['in_bar_x'] and not
                     self._touch_in_handle(
                         self._handle_x_pos, self._handle_x_size, touch)):
                 self.scroll_x = (touch.x - self.x) / self.width
+                e = self.effect_x
+
+            if e:
+                velocity = e.velocity
+                e.velocity = 0
+                self._update_effect_bounds()
+                if velocity > 0:
+                    e.trigger_velocity_update()
 
         # no mouse scrolling, so the user is going to drag the scrollview with
         # this touch.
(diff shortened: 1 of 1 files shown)


kivy/uix/scrollview.py
```diff
@@ -1148,7 +1158,7 @@ def update_from_scroll(self, *largs):
         self._bar_color = self.bar_color
         ev()
 
-    def _bind_inactive_bar_color(self, *l):
+    def _bind_inactive_bar_color(self, *args):
```
kuzeyron, line 1161: For readability I prefer having `*_` over `*args` when args isn't used.

kivy/uix/scrollview.py
```diff
@@ -798,14 +798,24 @@ def on_scroll_start(self, touch, check_children=True):
             return self.simulate_touch_down(touch)
 
         if in_bar:
+            e = None
```
kuzeyron, line 801: `e` means what exactly?

kuzeyron (changes requested):

kivy/uix/scrollview.py
akshayaurora: Look at the entire file `e` is in this file used to hold temporary value of effect_x and effect_y in scrollview. Following existing usage.

kivy/uix/scrollview.py
akshayaurora: Same as above this is following existing usage, in this file. While I am ok with using *_  it would stand out from other places of similar usage in the file.

Not in the scope of this change.

misl6 (approved):
LGTM. Thank you!

gottadiveintopython:
The local variable `e` is assigned four times in these changes, but the first three assignments don’t seem to have any effect.

kivy/uix/scrollview.py
```diff
@@ -798,14 +798,24 @@ def on_scroll_start(self, touch, check_children=True):
...
                         self._handle_y_pos, self._handle_y_size, touch)):
                 self.scroll_y = (touch.y - self.y) / self.height
+                e = self.effect_y
             elif (ud['in_bar_x'] and not
                     self._touch_in_handle(
                         self._handle_x_pos, self._handle_x_size, touch)):
                 self.scroll_x = (touch.x - self.x) / self.width
+                e = self.effect_x
+
+            e = self.effect_y if ud['in_bar_y'] else self.effect_x
```
AndiEcker, line 813: like 水戸う納豆齋(Nattōsai Mitō) already mentioned: this line is overwriting ALWAYS the value of the variable `e`. so either the 3 previous assignments of `e` or this line have to be removed

PR #9076, opened 2025-06-06
Add type hints to core providers

gottadiveintopython:
Maintainer merge checklist

Hi, I suppose there's no reason not to have type hints on core providers, given that other global objects like `kivy.clock.Clock` and `kivy.lang.Builder` have one.

Also, is there a reason for `kivy.core.text.Text` to exist? It's not listed in the `__all__`, and the unit tests succeeds without it.

### DIFF
--- a/kivy/core/audio_output/__init__.py
+++ b/kivy/core/audio_output/__init__.py
@@ -72,7 +72,7 @@ class SoundLoader:
         SoundLoader._classes.append(classobj)
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


misl6:
Seems to me that `Text` have been added by mistake (https://github.com/kivy/kivy/pull/3849).
I guess this is a good moment to just clean it up.

gottadiveintopython:
Thanks for the investigation. Updated.

PR #9084, opened 2025-06-12
Allow multiple IO's to be used while handling images in Kivy Core Image implementations

kuzeyron:
Added changes to the CoreImage so we can use other types too.
+ Changed a check in `core/window/__init__` because it was invalid.

Maintainer merge checklist

### DIFF
--- a/kivy/core/image/__init__.py
+++ b/kivy/core/image/__init__.py
@@ -225,11 +225,11 @@ class ImageLoaderBase(object):
         return None
 
     @staticmethod
-    def can_save(fmt, is_bytesio=False):
+    def can_save(fmt, is_bytesio_like=False):
         '''Indicate if the loader can save the Image object
 
         .. versionchanged:: 1.11.0
-            Parameter `fmt` and `is_bytesio` added
+            Parameter `fmt` and `is_bytesio_like` added
         '''
         return False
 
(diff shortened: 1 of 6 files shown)


misl6 (approved):
LGTM. Thank you!

gottadiveintopython:
That `if hasattr(filename, 'read') and callable(getattr(filename, 'read', None)):` check seems redundant. Couldn't it just be `if callable(getattr(filename, 'read', None)):` instead?

kuzeyron:
> That `if hasattr(filename, 'read') and callable(getattr(filename, 'read', None)):` check seems redundant. Couldn't it just be `if callable(getattr(filename, 'read', None)):` instead?

Doesn't work with just strings. It will fail.

gottadiveintopython:
I tested it with the following code, and it worked. Am I overlooking something?

```python
import pytest
from types import SimpleNamespace

obj_without_read = SimpleNamespace()
obj_with_callble_read = SimpleNamespace(read=lambda: None)
obj_with_read = SimpleNamespace(read=None)


def check1(filename):
    return hasattr(filename, 'read') and callable(getattr(filename, 'read', None))


def check2(filename):
    return callable(getattr(filename, 'read', None))


@pytest.mark.parametrize("input", ["a_string", obj_with_read, obj_with_callble_read, obj_without_read])
def test_check1_and_check2_are_semantically_equivalent(input):
    assert check1(input) is check2(input)
```

kuzeyron:
Feel free to address this in a new PR,

PR #9093, opened 2025-06-28
Kivy and its configuration is now aware of venv configurations + keeping the old logic if not

kuzeyron:
Kivy now better supports virtual environments (venvs). It can detect venv configurations and will place its configuration folder within the venv directory when launched through one. This allows for multiple, isolated Kivy builds on the same system without conflicting with the global `~/.kivy` configuration.

### DIFF
--- a/.ci/cicd-requirements.txt
+++ b/.ci/cicd-requirements.txt
@@ -1,4 +1,4 @@
-build~=1.2.1
+build~=1.3.0
 cibuildwheel~=2.23.2
 ruff>=0.8.3,<1.0.0
 twine~=6.1.0


kivy/__init__.py
```diff
@@ -348,16 +348,21 @@ def _patch_mod_deps_win(dep_mod, mod_name):
...
 if not environ.get('KIVY_DOC_INCLUDE'):
     # Configuration management
+    user_home_dir = expanduser('~')
+    kivy_home_dir = None
+
     if 'KIVY_HOME' in environ:
         kivy_home_dir = expanduser(environ['KIVY_HOME'])
-    else:
-        user_home_dir = expanduser('~')
+    elif platform in {'android', 'ios'}:
```
Zen-CODE: The below would be slightly more efficient in preventing 2 checks for "ios" / "android".
```python

    if 'KIVY_HOME' in environ:
    elif platform == 'android':
    elif platform == 'ios':
        user_home_dir = join(user_home_dir, 'Documents')
    elif sys.prefix != sys.base_prefix:
        # Detection if venv being used with the framework
        user_home_dir = sys.path[0]
```

kuzeyron: Okay, that's done but it seems I messed up the PR..

Zen-CODE: It's brought is some changes from a PR I merged this morning (just doc changes). Perhaps re-fetch main, merge that back into your branch locally and re-push?

kuzeyron: If you can assist me with the commands! I don't know the way around the git commands. Whatever I try seems to fail so I usually end up by creating a new PR..

Zen-CODE: From what I can tell, it's something like:
```
git remote add upstream https://github.com/kivy/kivy
git fetch upstream
git checkout stop-effect-in-scroll_to
git merge upstream/master
git rebase upstream/master
git add .
git rebase --continue # or git commit if you used merge
git push origin stop-effect-in-scroll_to
```
Can't really verify that, but hope it works for you?

Issue #9072, opened 2025-06-02
scroll_to method while using overscroll effects


mkolender:
**Software Versions**
* Python: 3.11.2, 3.13.3
* OS: Raspberry Pi OS bookworm, Windows 11 23H2 (OS Build 22631.5335)
* Kivy: v2.3.1
* Kivy installation method: pip

**Describe the bug**
I'm working on an app that supports both touchscreen and physical keyboard for navigating the interface. I'm highlighting buttons selected with keyboard and scrolling a scrollview using scroll_to method. When user scrolls the view with a swipe and the swipe animation isn't over the scroll_to method has no effect. After setting scroll_view.effect_y.velocity to 0 scroll_to seems to be working, but there is still an issue. When the user overscrolls the view scroll_to is not working if the view is still overscrolled. After that the view is sometimes jumping for a split of a second and goes back without scrolling properly to a desired widget.

**Expected behavior**
Scroll_to should be scrolling to desired widget no matter if the view is still moving after swipe-scrolling. It should also work when the view is overscrolled at the moment.

**To Reproduce**
Add a few widgets to a BoxLayout inside of a ScrollView. Scroll the screen by swiping to the top so that the view is overscrolled and run scroll_to for the widget on the bottom of BoxLayout (press up arrow in my example code). The scroll_to method has no effect then. Also the scrollview doesn't scroll when velocity is not equal to 0. Sometimes the view is jumping back and forth when using scroll_to and doesn't scroll properly.

**Code and Logs and screenshots**
```python
from kivy.app import App
from kivy.lang import Builder
from kivy.uix.boxlayout import BoxLayout
from kivy.uix.button import Button
from kivy.core.window import Window
from kivy.properties import BooleanProperty, NumericProperty


kv = """
#:set normal_state_color (48/255, 164/255, 156/255, 1)
#:set down_state_color (26/255, 26/255, 32/255, 1)
#:set highlight_frame_color (1, 1, 1, 1)
#:import Window kivy.core.window.Window

<ListButton>
    background_normal: ""
    background_color: 0, 0, 0, 0
    size_hint_y: None
    height: Window.height / 5.3
    highlighted: False

    canvas.before:
        Color:
            rgba: normal_state_color if self.state == "normal" else down_state_color
        RoundedRectangle:
            pos: self.x, self.y  # Use padding
            size: self.width, self.height
            radius: [self.height / 6]
        Color:
[...]

PR #9103, opened 2025-07-22
Enhance Scrollview.scroll_to to stop animation, prior to scroll_to animation

ElliotGarbus:
This PR is the solution to this issue: https://github.com/kivy/kivy/issues/9072
When the ScrollView is scrolling or over-scrolling, scroll_to() does not work.
The fix adds a halt() method to KineticEffect, and calls the halt() method from scroll_to prior to the animation.

Details:
KineticEffect.halt() stops the movement immediately without momentum.
This method immediately stops any ongoing kinetic scrolling effect by:
- Cancelling any scheduled velocity updates to prevent further movement
- Setting is_manual to False to indicate no manual interaction is in progress

Unlike the stop() method which calculates final velocity for momentum
scrolling, this method provides instant cessation of movement without 
any residual animation.

This is useful when you need to programmatically stop scrolling in 
response to user actions or when the scroll position needs to be fixed
immediately.  The halt() method is called by ScrollView.scroll_to to 
stop scrolling animation prior to moving to the target widget. 

This PR fixes the issue.  

Here is a test case.  WIthout the fix if the buttons are moving under momentum, the scroll_to does not work.  With the fix, scroll_to works as expected.

```python
from kivy.app import App
from kivy.lang import Builder
from kivy.properties import BooleanProperty
from kivy.metrics import dp

"""
Test code for identifying issue.  When the ScrollView is scrolling or over-scrolling, scroll_to() does not work.
The spinner in the top left selects the effect_cls for the ScrollView
The two buttons on the bottom "Scroll to top" and "Scroll to bottom" use, scroll_to() to scroll to the widget.
"""


kv = """
BoxLayout:
    orientation: 'vertical'
    spacing: dp(5)
    Spinner:
        size_hint_y: None
        height: dp(48)
        values: ['DampedScrollEffect', 'OpacityScrollEffect', 'ScrollEffect']
        text: 'DampedScrollEffect'
        on_text: 
            sv.effect_cls = self.text
            sv.rebind()
    SV:
        id: sv
        BoxLayout:
            id: scroll_box
            orientation: 'vertical'
            size_hint_y: None
            height: self.minimum_height
    BoxLayout:
        size_hint_y: None
        height: dp(48)
        Button:
            text: 'Scroll to top'
            on_release: 
                sv.scroll_to(scroll_box.children[-1])
        Button:
            text: 'Scroll to bottom'
            on_release: 
                sv.scroll_to(scroll_box.children[0])
"""


class SV(ScrollView):
    stop_animation = BooleanProperty(False)

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.rebind()

    def rebind(self):
        self.effect_y.bind(velocity=lambda obj, v: self.show_info('velocity', v))
        self.effect_y.bind(value=lambda obj, v: self.show_info('value', v))
        self.effect_y.bind(is_manual=lambda obj, v: self.show_info('is_manual', v))
        self.effect_y.bind(overscroll=lambda obj, v: self.show_info('overscroll', v))
        self.effect_y.bind(displacement=lambda obj, v: self.show_info('displacement', v))

    def on_scroll_y(self, obj, v):
        print(f'scroll_y: {v}')

    def show_info(self, name, v):
        print(f'{name}: {v}')


class EvaluateScrollEffectsApp(App):
    def build(self):

    def on_start(self):
        for i in range(50):
            self.root.ids.scroll_box.add_widget(Button(text=f'Button {i}', size_hint_y=None, height=dp(48)))


EvaluateScrollEffectsApp().run()
```

Maintainer merge checklist

### DIFF
--- a/kivy/effects/kinetic.py
+++ b/kivy/effects/kinetic.py
@@ -196,3 +196,22 @@ class KineticEffect(EventDispatcher):
         self.velocity -= self.velocity * self.friction * dt / self.std_dt
         self.apply_distance(self.velocity * dt)
         self.trigger_velocity_update()
+
+    def halt(self):
+        '''Stop the movement immediately without momentum.
+        This method immediately stops any ongoing kinetic scrolling effect by:
+        - Cancelling any scheduled velocity updates to prevent further movement
+        - Setting is_manual to False to indicate no manual interaction is in progress
+
+        Unlike the stop() method which calculates final velocity for momentum
+        scrolling, this method provides instant cessation of movement without
+        any residual animation.
+
+        This is useful when you need to programmatically stop scrolling in
+        response to user actions or when the scroll position needs to be fixed
+        immediately.  The halt() method is called by ScrollView.scroll_to to
+        stop scrolling animation prior to moving to the target widget.
+        '''
+        self.is_manual = False
+        self.velocity = 0
+        self.trigger_velocity_update.cancel()
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -1065,6 +1065,12 @@ class ScrollView(StencilView):
         sxp = min(1, max(0, self.scroll_x - dsx))
         syp = min(1, max(0, self.scroll_y - dsy))
 
+        # Stop any existing motion before starting new animation
+        if self.effect_x:
+            self.effect_x.halt()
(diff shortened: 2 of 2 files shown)


Zen-CODE:
This works for me +1

PR #9113, opened 2025-08-08
Kivy can now be used with venv for setup

kuzeyron:
I did something wrong with the git commands. The old PR: https://github.com/kivy/kivy/pull/9093

Kivy now better supports virtual environments (venvs). It can detect venv configurations and will place its configuration folder within the venv directory when launched through one. This allows for multiple, isolated Kivy builds on the same system without conflicting with the global ~/.kivy configuration.

### DIFF
--- a/doc/sources/guide/config.rst
+++ b/doc/sources/guide/config.rst
@@ -18,6 +18,9 @@ On desktop, this defaults to::
 
     <HOME_DIRECTORY>/.kivy/config.ini
 
+If you are using venv (Virtual Environment) it should look like this::
+    <PROJECT_DIRECTORY>/.kivy/config.ini
+
 Therefore, if your user is named "tito", the file will be here:
 
 - Windows: ``C:\Users\tito\.kivy\config.ini``
--- a/doc/sources/guide/config.rst
+++ b/doc/sources/guide/config.rst
@@ -21,7 +21,7 @@ On desktop, this defaults to::
 If you are using venv (Virtual Environment) it should look like this::
     <PROJECT_DIRECTORY>/.kivy/config.ini
 
-Therefore, if your user is named "tito", the file will be here:
+If you are not using a virtual environment and your user is named "tito", the file will be here:
 
 - Windows: ``C:\Users\tito\.kivy\config.ini``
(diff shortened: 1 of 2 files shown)


kuzeyron:
Zen, this is the final try :P

Zen-CODE:
It depends how persistent you are? I would try pushing again with changes to `kivy/__init__.py` removed as that will test if you changes there broken anything. If it still fails, we know it's already broken....

kuzeyron:
I'll need to put some time because it worked the first time I made the PR but I forgot to make a backup of what I did..

Zen-CODE:
It will be committed to git. So just make a note of this commit hash, and commit the `__init__.py` from main/master? You can then revert to that hash to get your file back....

kuzeyron:
I couldn't find it. Original PR: https://github.com/kivy/kivy/pull/9093

kuzeyron:
Okay, that solved it.

Zen-CODE (dismissed):
Nice. Thanks for all the effort! :-)

Zen-CODE:
We probably need to update the documentation to explain the new behaviour. I'll try and get there but if you have the energy.. :-)  I think that needs to be added before we can really merge this...

https://kivy.org/doc/stable/guide/config.html

from https://github.com/kivy/kivy/blob/master/doc/sources/guide/config.rst

kuzeyron:
In this PR I did add about "PROJECT_DIRECTORY" for the config-doc. Is it not enough?

Zen-CODE:
The problem is some of docs of it become inaccurate. e.g.
```
Therefore, if your user is named "tito", the file will be here:

Windows: C:\Users\tito\.kivy\config.ini
macOS: /Users/tito/.kivy/config.ini
Linux: /home/tito/.kivy/config.ini
```
It should indicate that's if you are not using a virtual environment e.g.
```
If you are not using a virtual environment and your user is named "tito", the file will be here:
````
I think that would be enough?

kuzeyron:
To me that's fine!

Zen-CODE:
Thanks again ;-)

akshayaurora (approved):

PR #9114, opened 2025-08-10
Use virtual time in animation tests to speed them up

gottadiveintopython:
Maintainer merge checklist

Currently, `test_animations.py` takes about 10 seconds on my machine, which is quite long. I'm sure this happens to others, as the cause is `time.sleep()`. This PR makes it run instantly.


misl6 (approved):
LGTM. Thank you!

Issue #6889, opened 2020-05-21
ScrollView in a RecyclceView not working


ElliotGarbus:
**Software Versions**
* Python: 3.73
* OS: Win10
* Kivy: 1.11.1
* Kivy installation method: 

**Describe the bug**
A ScrollView that scrolls in the x direction is placed in a RecycleView that scrolls in the y direction.  When touching the scrollview, the RecycleView will NOT scroll.

**Expected behavior**
When touching a widget in the ScrollView, I expect vertical movement to result in the RecycleView scrolling.

**To Reproduce**

**Code and Logs and screenshots**
```python
from kivy.app import App
from kivy.lang import Builder
from kivy.uix.boxlayout import BoxLayout
from kivy.properties import NumericProperty, StringProperty


kv = """
<TestBox>
    Button:
        size_hint_x: None
        width: 100
        text: str(root.index)
        font_size: 30
        on_release: print(f'Button: {root.text} pressed, with index {root.index}')
    ScrollView
        scroll_type: ['content']
        #bar_width: dp(20)
        do_scroll: True, False
        Label:
            id: scroll_label
            text: root.text
            font_size: 30
            size_hint: None, None
            size: self.texture_size
    
RootBox:
    orientation: 'vertical'
    RecycleView:
        viewclass: 'TestBox'
        data: root.rv_data_list
        scroll_type: ['bars','content']
        bar_width: dp(20)
        do_scroll: False, True
        RecycleBoxLayout:
            default_size: None, dp(56)
            default_size_hint: 1, None
            size_hint_y: None
            height: self.minimum_height
            orientation: 'vertical'
"""


class TestBox(BoxLayout):
    index = NumericProperty()
    text = StringProperty()


class RootBox(BoxLayout):
    rv_data_list = [{'text': 'This needs to be long enough to scroll in the x direction. ' * 2 + str(n),
                     'index': n + 200} for n in range(40)]


class RVTestApp(App):
    def build(self):
        return Builder.load_string(kv)


RVTestApp().run()

```

**Additional context**

snuq:
Just to elaborate, this is not in any way tied to recycleview, it behaves similarly with a scrollview inside of a scrollview:

```from kivy.app import App
from kivy.lang import Builder
KV = """
<Scroller@ScrollView>:
    size_hint_y: None
    height: 100
    do_scroll_y: False
    Label:
        font_size: 20
        size_hint: None, None
        size: self.texture_size
[...]

Issue #4052, opened 2016-03-01
ScrollView passing through touch events


zeeMonkeez:
It seems like on iOS, touch events in scrollviews are first passed to children to see whether they are captured (say in Settings.app, or links in Safari). If no child captures the event, a scroll is initiated. In [this gist](https://gist.github.com/zeeMonkeez/0c677719f2c2b3317118) I've changed `on_touch_down` to reproduce this behaviour. Should this be added at least as an option to `ScrollView`, so that interactions with children are lagless, yet scrolls can still be initiated.

kived:
If that's true, then if there's a Button in your ScrollView then you can't scroll on top of that button. If the Button filled all the visible space in the ScrollView, it would be impossible to scroll at all.

zeeMonkeez:
I understand the limitation, hence it should not be default behaviour so existing code does not break.
Having said this, having a Button filling the ScrollView is bad UI design IMHO, but even then if the Button's `on_touch_down` returns `False` the event still bubbles up to ScrollView and scrolling is possible. In that case, if `on_touch_move` should also cancel the grab on the Button, and `on_release` should not be fired.

It just seems like this is the default behaviour in iOS: In Settings.app, On/Off switches in scrollable lists always swallow the touch. Larger cells (the buttons leading to the next level) capture the touch immediately, yet on move they lose focus and a scroll is initiated.

Not sure how his is handled on android.

Issue #2986, opened 2015-02-02
Scrollview inside a scrollview's content obscures UI interaction


mixedCase:
Placing a scrollview inside another one will make the contents of the parent one behave as if they were on top, yet it's rendered perfectly on the screen. This can be seen when the content is scrolled.

Demo of the issue here:
https://www.youtube.com/watch?v=qfPB9Idrk_0

In here I'm using an external scrollview for y axis movement, while the inner one is used only for x axis movement of the data. The contents of the inner one work as expected, only the external one's content (in this case, the whole inner scrollview) trigger the bug.
##

mixedCase:
@thopiekar This is also reproducible on Android. I would guess Windows and OS X too but we'd need more testers to confirm where the bug may be.

matham:
Can you please provide some example code to make it easier to test and fix?

mixedCase:
Sure!

https://gist.github.com/mixedCase/f89023fe948c47e8df01

I'm using python2 btw.

darkopevec:
I added size_hint_y: 2 to the second BoxLayout and believe it works fine with https://github.com/kivy/kivy/pull/4653

mixedCase:
Closing because if there's a bot marking issues as stale then there's no point in filing issues no one cares about fixing.

dessant:
@mixedCase, I'm not sure that's the right kind of approach to free software development, please refrain from making such posts on the issue tracker.

dessant:
@mixedCase, one of the posts mentions having success with a pr, it would make sense to test what is the current state of the bug report and update the thread with any changes you have found, or at least confirm that the bug is still there. Alternatively you could also give a shot at fixing the issue.

mixedCase:
I have no interest in pursuing this issue since I'm no longer using Kivy. I left a video showcasing the bug, and a minimal working example; if no one cared in two years and then comes a bot marking old issues as automatically invalid, I'm not sure I should feel bad about "not having the right kind of approach to free software development".

I think I did my part as a bug reporter. If someone has lost interest in fixing bugs I suggest he hands off development to someone else as I did with KivyMD.

dessant:
It was not marked as invalid, but stale. This is a pretty standard feature at bigger projects, signaling that the issue may need to be rechecked. From the moment someone confirms the bug, the issue will not be considered for closure.

PR #9118, opened 2025-08-15
Fixing Orthogonal Nested ScrollViews

ElliotGarbus:
The modifications allow orthoganal nested ScrollViews to work as expected.

Previously, if a `ScrollView` was nested inside another `ScrollView`, the
inner scrollview would consume all touches, preventing the outer scrollview
from scrolling, even when the movement was orthogonal to the inner's scroll
axis.

Now, when two `ScrollView` widgets are nested with perpendicular scrolling
directions (e.g., inner vertical, outer horizontal), touches are routed
according to movement direction:

- **Same-axis movement**: The inner `ScrollView` handles scrolling.
- **Orthogonal movement**: The touch is released to the parent `ScrollView`,
  allowing it to scroll.

This behavior is automatic and requires no configuration changes.
It makes nested scrolling more intuitive.

The changes are limited to the on_scroll_move method.

The direction of the touch is determined only at the beginning of the on touch move, changing orientation(from vertical to horizontal for example) requires releasing the touch and starting a new touch.  Only one `ScrollView` is active at a time.

This PR fixed the following issues:
https://github.com/kivy/kivy/issues/5617
https://github.com/kivy/kivy/issues/6889
https://github.com/kivy/kivy/issues/4052
https://github.com/kivy/kivy/issues/2986

Maintainer merge checklist

### DIFF
--- a/kivy/uix/scrollview.py
+++ b/kivy/uix/scrollview.py
@@ -913,6 +913,27 @@ class ScrollView(StencilView):
                 return
             ud['dx'] += abs(touch.dx)
             ud['dy'] += abs(touch.dy)
+
+            # In the case of nested scroll views, we are in the inner SV.
+            # if user intent is orthogonal to this SV's allowed axis,
+            # yield to parent immediately by returning False, then the parent
+            # will handle the scroll.
+
+            if ud['dx'] > self.scroll_distance and not self.do_scroll_x:
+                if self.do_scroll_y and self.effect_y:
+                    self.effect_y.cancel()
+                if self.do_scroll_x and self.effect_x:
+                    self.effect_x.cancel()
+                touch.ud[self._get_uid('svavoid')] = True
+                return False
+            if ud['dy'] > self.scroll_distance and not self.do_scroll_y:
+                if self.do_scroll_x and self.effect_x:
+                    self.effect_x.cancel()
+                if self.do_scroll_y and self.effect_y:
+                    self.effect_y.cancel()
+                touch.ud[self._get_uid('svavoid')] = True
+                return False
+
             if ((ud['dx'] > self.scroll_distance and self.do_scroll_x) or
                     (ud['dy'] > self.scroll_distance and self.do_scroll_y)):
                 ud['mode'] = 'scroll'


ElliotGarbus:
Here is a test program I used to interactively test the behavior of nested scroll views.

```python
"""
Test file for demonstrating nested ScrollViews with orthogonal scrolling.
This test shows how the updated ScrollView implementation allows proper
nested scrolling when the scroll directions are perpendicular to each other.

Layout:
- Top level: Horizontal BoxLayout
- Left side: Vertical ScrollView containing multiple horizontal ScrollViews
- Right side: Horizontal ScrollView containing multiple vertical ScrollViews
"""

from kivy.app import App
from kivy.metrics import dp


class NestedScrollViewTest(App):
    def build(self):
        
        # Main horizontal layout
        main_layout = BoxLayout(orientation='horizontal', spacing=10, padding=10)
        
        # Left side: Vertical ScrollView with horizontal ScrollViews inside
        left_vertical_sv = ScrollView(
            do_scroll_x=False,  # Only vertical scrolling
            do_scroll_y=True,
            scroll_type=['bars', 'content'],
            bar_width=dp(8),
            bar_color=[0.3, 0.6, 1.0, 0.8]
        )
        
        # Container for horizontal ScrollViews
        left_container = BoxLayout(
            orientation='vertical',
            spacing='20dp',
            size_hint_y=None
        )
        left_container.bind(minimum_height=left_container.setter('height'))
        
        # Add multiple horizontal ScrollViews
        for i in range(8):
            # Create a horizontal ScrollView
            horizontal_sv = ScrollView(
                do_scroll_x=True,   # Only horizontal scrolling
                do_scroll_y=False,
                scroll_type=['bars', 'content'],
                size_hint_y=None,
                height=dp(120),
                bar_width=dp(6),
                bar_color=[1.0, 0.5, 0.3, 0.8]
            )
            
            # Create content for horizontal ScrollView
            horizontal_content = BoxLayout(
                orientation='horizontal',
                spacing=dp(10),
                size_hint_x=None
            )
            horizontal_content.bind(minimum_width=horizontal_content.setter('width'))
            
            # Add a label to identify this row
            label = Label(
                text=f'Row {i+1} - Horizontal Scroll',
                size_hint_x=None,
                width=dp(200),
                height=dp(30),
                color=[1, 1, 1, 1],
                bold=True
            )
            horizontal_content.add_widget(label)
            
            # Add buttons to make content wider than the ScrollView
            for j in range(15):
                btn = Button(
                    text=f'Btn {j+1}',
                    size_hint_x=None,
                    width=dp(100),
                    height=dp(80),
                    background_color=[0.2 + (i * 0.1) % 0.8, 0.3, 0.7, 1]
                )
                horizontal_content.add_widget(btn)
            
            horizontal_sv.add_widget(horizontal_content)
            left_container.add_widget(horizontal_sv)
        
        left_vertical_sv.add_widget(left_container)
        
        # Right side: Horizontal ScrollView with vertical ScrollViews inside
        right_horizontal_sv = ScrollView(
            do_scroll_x=True,   # Only horizontal scrolling
            do_scroll_y=False,
            scroll_type=['bars', 'content'],
            bar_width=dp(8),
            bar_color=[0.3, 1.0, 0.6, 0.8]
        )
        
        # Container for vertical ScrollViews
        right_container = BoxLayout(
            orientation='horizontal',
            spacing='20dp',
            size_hint_x=None
        )
        right_container.bind(minimum_width=right_container.setter('width'))
        
        # Add multiple vertical ScrollViews
        for i in range(6):
            # Create a vertical ScrollView
            vertical_sv = ScrollView(
                do_scroll_x=False,  # Only vertical scrolling
                do_scroll_y=True,
                scroll_type=['bars', 'content'],
                size_hint_x=None,
                width=dp(200),
                bar_width=dp(6),
                bar_color=[1.0, 0.3, 0.5, 0.8]
            )
            
            # Create content for vertical ScrollView
            vertical_content = GridLayout(
                cols=1,
                spacing=dp(10),
                size_hint_y=None,
                height=0
            )
            vertical_content.bind(minimum_height=vertical_content.setter('height'))
            
            # Add a label to identify this column
            label = Label(
                text=f'Column {i+1}\nVertical Scroll',
                size_hint_y=None,
                height=dp(40),
                color=[1, 1, 1, 1],
                bold=True
            )
            vertical_content.add_widget(label)
            
            # Add buttons to make content taller than the ScrollView
            for j in range(20):
                btn = Button(
                    text=f'Item {j+1}',
                    size_hint_y=None,
                    height=dp(60),
                    background_color=[0.7, 0.3, 0.2 + (i * 0.1) % 0.8, 1]
                )
                vertical_content.add_widget(btn)
            
            vertical_sv.add_widget(vertical_content)
            right_container.add_widget(vertical_sv)
        
        right_horizontal_sv.add_widget(right_container)
        
        # Add both sides to main layout
        main_layout.add_widget(left_vertical_sv)
        main_layout.add_widget(right_horizontal_sv)
        
        return main_layout


if __name__ == '__main__':
    NestedScrollViewTest().run()
```

ElliotGarbus:
I'm converting this to a draft, I need to make changes to the mouse wheel handling.

ElliotGarbus:
Mouse wheel handing has been added.

ElliotGarbus:
I've found an issue when multiple scrollviews are visible - as in the testcase above.  The touch grabs were not being properly released.  I've got a fix and will update the PR in a day or so.  Thanks for your patience.

ElliotGarbus:
The issue with grabs has been fixed.  Here is a brief summary.  I added some additional comments to the code that relates to these changes.  
The core problem:  There are two nested scrollviews in the right-hand side and left-hand side of a Boxlayout.  Aggressively using right-hand side nested scrollview, and then very quickly moving to the other side would not result in the proper grab handing.  A number of changes were made to address this issue:

- **Prevent off-viewport grabs**
  - `on_touch_down`: returns False immediately when the touch is outside this `ScrollView` (`collide_point` check) so siblings/parent can handle it.

- **Safe redispatch gating**
  - `simulate_touch_down`: blocks redispatch if another widget currently owns the touch (`grab_current` not in {None, self}) or if this `ScrollView` isn’t in `touch.grab_list`.
  - Introduced a per-touch override flag `svforce` (keyed by this SV’s uid) so intentional redispatch paths can bypass the guard.

- **Intentional redispatch paths use override**
  - `_change_touch_mode`: wraps `simulate_touch_down` with `svforce` so the timeout-based redispatch to children still works.
  - `on_scroll_stop` (click passthrough when mode is `unknown`): wraps `simulate_touch_down` with `svforce` for single-tap forwarding.

- **Prompt cleanup on release**
  - `on_touch_up`: if this `ScrollView` previously handled the touch (its uid key exists in `touch.ud`), it now dispatches `on_scroll_stop`, ungrabs if still holding, preserves focus behavior, and returns. This prevents lingering grabs that block siblings.

ElliotGarbus:
I'm marking this as a draft again.  I found some new corner cases causing incorrect behavior.  I am making progress on a new implementation that uses a special widget for nesting ScrollViews, NestedScrollVIewManager that will result in a more maintainable solution.  I'll leave this as a draft for now.  If I prefer the new implementation I will close this PR.  If not I will come back and work to fix these corner cases.  

In my proposed implementation for nested scrollviews they will be under a NestedScrollViewManager.  It will look something like this:
```
NestedScrollViewManager:
    ScrollView:  # the outer scrollview
        ...attributes and layout
        ...
        ScrollView:  # one or more scrollviews deeper in the widget tree
```
My objective is to deliver a solution that is more maintainable both for ScrollView and the NestedScrollViewManager.

DexerBR:
> I'm marking this as a draft again. I found some new corner cases causing incorrect behavior. I am making progress on a new implementation that uses a special widget for nesting ScrollViews, NestedScrollVIewManager that will result in a more maintainable solution. I'll leave this as a draft for now. If I prefer the new implementation I will close this PR. If not I will come back and work to fix these corner cases.
> 
> In my proposed implementation for nested scrollviews they will be under a NestedScrollViewManager. It will look something like this:
> 
> ```
> NestedScrollViewManager:
>     ScrollView:  # the outer scrollview
>         ...attributes and layout
>         ...
>         ScrollView:  # one or more scrollviews deeper in the widget tree
> ```
> 
> My objective is to deliver a solution that is more maintainable both for ScrollView and the NestedScrollViewManager.


@ElliotGarbus Even with an alternative implementation, I believe this PR is necessary to address issues that make implementing this orthogonal sliding behavior (even without nested scrollviews) complicated:

ElliotGarbus:
@DexerBR I agree on the need.  
In my implementation, the ScrollView widget will be simplified as it no longer directly supports nesting.  Adding the NestedScrollViewManager becomes a bigger PR with a greater impact.  I hope to be done in less than 2 weeks.  

My current status: I have simplified the ScrollView, refactored the long methods with helper functions and added significant implementation comments.  I will be starting on the NestedScrollViewManager later today.  When I have something working I'll share it on discord - and then create a new PR.   I'll look forward to your feedback.

If for some reason NestedScrollViewManager seems like the wrong direction, I'll come back to this PR.  As I see it today, adding the NestedScrollViewManger will allow new capabilities and improve the maintainability of the code.

DexerBR:
> @DexerBR I agree on the need. In my implementation, the ScrollView widget will be simplified as it no longer directly supports nesting. Adding the NestedScrollViewManager becomes a bigger PR with a greater impact. I hope to be done in less than 2 weeks.
> 
> My current status: I have simplified the ScrollView, refactored the long methods with helper functions and added significant implementation comments. I will be starting on the NestedScrollViewManager later today. When I have something working I'll share it on discord - and then create a new PR. I'll look forward to your feedback.
> 
> If for some reason NestedScrollViewManager seems like the wrong direction, I'll come back to this PR. As I see it today, adding the NestedScrollViewManger will allow new capabilities and improve the maintainability of the code.

All right 👍

Considering this hypothetical layout, how would the NestedScrollViewManager behave?
```kvlang
FloatLayout:
    BoxLayout:
        ScrollView:
            BoxLayout:
                Button:
                ScrollView:
                    BoxLayout:
                    BoxLayout:
                Button:
                ScrollView:
                    BoxLayout:
                        Button:
                        BoxLayout:
                            ScrollView:
                            Button:
        BoxLayout:
        Button:
    ScrollView:
        BoxLayout:
```

ElliotGarbus:
To use the `NestedScrollViewManager`, we would modify the widget tree.  The `NestedScrollViewManager`, is a `RelativeLayout` that captures and routes touches/events to it's child scrollviews.  My objective is to create a system that enables new capabilities and is also more maintainable.

```
FloatLayout:
    BoxLayout:
        NestedScrollViewManager:
            ScrollView:  # this is the outer SV
                BoxLayout:
                    Button:
                    ScrollView:  # an inner SV
                        BoxLayout:
                        BoxLayout:
                    Button:
                    ScrollView:  # an inner SV
                        BoxLayout:
                            Button:
                            BoxLayout:
                                ScrollView:  # I'll ignore this... I'm not contemplating triple nesting...
                                Button:
        BoxLayout:
        Button:
    ScrollView:  # this is not a nested scrollview
        BoxLayout:
```
Here are my initial thoughts on the implementation, I'm confident these details will change.  

The `NestedScrollViewManager` will grab the touch and ungrab the touch as required, and issue on_scroll_start, move and start events as required to the appropriate underlying scrollview (outer or inner).   

The `NestedScrollViewManager` and the `ScrollView` will use `touch.ud` (user data dictionary) to maintain per-touch state across the touch lifecycle (down -> move -> up). This is critical for:
    1. Nested ScrollView coordination (avoiding conflicts)
    2. Gesture detection (scroll vs click )
    3. State persistence across method calls

This information is used to control the behavior of the `ScrollView` and message routing based on the state of the touch.  

 `touch.ud` Keys for Nested Scroll Behavior

| Key              | Purpose                                                                 |
|------------------|-------------------------------------------------------------------------|
| `mode`     | Indicates whether the touch has been classified pending, scrolling, replay. |
| `nested_managed` | Marks that this touch is being managed by a `NestedScrollViewManager`. Prevents confusion with normal `ScrollView` touches. |
| `nested_role`    | Tracks the current ownership/interpretation of the touch within nested scrolling (e.g. `"inner"`, `"outer"`, `"inner-rejected"`). |
| `scroll_owner`   | Identifies the specific scrollview widget currently handling the scroll.       |


The `NestedScrollViewMagager` always knows the outer scrollview - it is it's child.  The inner scrollview is "discovered" if the on_scroll_start message to children returns True.  This way there is no additional overhead to track inner scrollviews or requirement for the developer to tag an inner scrollview.

Having knowledge of the inner vs outer scrollview will simplify the behavior and make the code more maintainable.  
It will required the creation of the `NestedScrollViewManager` widget, and an update to the `ScrollView`.
I'll know much more after I have something working.

Happy to answer any questions.

ElliotGarbus:
I'm closing this pull request I have a more complete implementation, PR coming soon.

PR #9119, opened 2025-08-17
Fixed SettingNumeric textinput validation

takkasila:
- Fixed SettingNumeric textinput validation.
- Added unittest for the fix.

Maintainer merge checklist

### DIFF
--- a/kivy/tests/test_uix_settings.py
+++ b/kivy/tests/test_uix_settings.py
@@ -3,7 +3,9 @@ import json
 import pytest
 
 from kivy.config import ConfigParser
-from kivy.uix.settings import Settings
+from kivy.uix.settings import Settings, SettingNumeric
+from kivy.uix.textinput import TextInput
+import unittest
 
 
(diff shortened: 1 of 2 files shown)


kivy/tests/test_uix_settings.py
```diff
@@ -31,3 +33,73 @@ def test_settings_create_json_panel_errors():
...
+                    "section": "Section", 
+                    "key": "Key"
+                }
+            ]"""
+        )
+
+
+    @pytest.fixture(autouse=True, scope="function")
+    def setup(self):
+        self.settingNumeric = SettingNumericTest(self.panel)
```
Zen-CODE, line 79: I am concerned here. You are actually testing a subclass of `SettingNumeric`, not an instance of `SettingNumeric` itself. You've added a bunch of properties which could interfere with it's behaviour. Even if not now, it could in future versions. Rather build the test to test a pure `SettingNumeric` instance.

PR #9133, opened 2025-08-29
Add a fixture and utility functions for using `pytest` to test a Kivy app

clayote:
I found stdlib's `unittest` unpleasant to work with, so I converted all my `GraphicUnitTest`s to use `pytest` instead. This code came in handy, and may be generally useful.

The `kivy_init` fixture does the same stuff as `setUp()` and `tearDown()` in `GraphicUnitTest`. It is up to the user to make a fixture for their app in particular.

`advance_frames` is the same as in `GraphicUnitTest`.

`idle_until` runs the event loop until some predicate holds true. Useful for letting needed events fire before you start to test.

Maintainer merge checklist

### DIFF
--- a/kivy/tests/common.py
+++ b/kivy/tests/common.py
@@ -21,6 +21,7 @@ import sys
 from functools import partial
 import os
 import threading
+from typing import Callable, Optional
 from kivy.graphics.cgl import cgl_get_backend_name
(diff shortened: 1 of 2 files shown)


misl6 (approved):
LGTM. Thank you!

Issue #9134, opened 2025-08-29
`.kivy` is created wherever a kivy application is run.


DexerBR:
**Software Versions**
* Python: 3.11.9
* OS: WIndows
* Kivy: 3.0.0.dev0
* Kivy installation method: pip install -e .

**Describe the bug**
When runing a kivy app outsize the "root" dir, a `.kivy` folder is created

**Expected behavior**
The `.kivy` folder should only be created in a specific directory (possibly in the "root" directory?)

**To Reproduce**
`python examples\widgets\bubble_test.py` > will create a `.kivy` inside `examples\widget`

**Code and Logs and screenshots**
N/A

**Additional context**
N/A

kuzeyron:
It's not a bug: https://github.com/kivy/kivy/pull/9113

DexerBR:
> It's not a bug: [#9113](https://github.com/kivy/kivy/pull/9113)

From the PR description:

> Kivy now better supports virtual environments (venvs). It can detect venv configurations and will place its configuration folder within the venv directory when launched through one. This allows for multiple, isolated Kivy builds on the same system without conflicting with the global ~/.kivy configuration.

@kuzeyron It turns out that it is not being placed inside the virtual environment directory, but in the execution directory.

DexerBR:
@kuzeyron I believe that `user_home_dir = dirname(sys.prefix)` produces a more predictable and expected result than `user_home_dir = sys.path[0]`. I tested it, and it works as expected.

kuzeyron:
Great! Can you make the PR?

DexerBR:
> Great! Can you make the PR?

I noticed you opened the PR, thanks!

kuzeyron:
Is it okay if we close this ticket as complete?

kuzeyron:
Closing because of https://github.com/kivy/kivy/pull/9137

PR #9137, opened 2025-09-05
Fix of venv-config

kuzeyron:
Solves https://github.com/kivy/kivy/issues/9134 where .kivy is being created in nested folders of venv.

### DIFF
--- a/kivy/__init__.py
+++ b/kivy/__init__.py
@@ -359,7 +359,7 @@ if not environ.get('KIVY_DOC_INCLUDE'):
         user_home_dir = join(user_home_dir, 'Documents')
     elif sys.prefix != sys.base_prefix:
         # Detection if venv being used with the framework
-        user_home_dir = sys.path[0]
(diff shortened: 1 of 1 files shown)


AndreMiras:
Thanks @kuzeyron for the fix.
@DexerBR since you're more in context, could you give this a try and let us know if that works for you too

AndreMiras:
I wanted to ask for a test, but I see a lot is happening at module top level, making it hard to properly unit test.
That thing probably deserve some refactoring.
Any change look good to me, merging

AndreMiras (approved):
Thanks for the follow up

DexerBR:
> Thanks @kuzeyron for the fix. @DexerBR since you're more in context, could you give this a try and let us know if that works for you too

After this PR, the behavior is more predictable: the `.kivy` folder is now created in the same directory as the virtual environment. For example:

```
project/
├─ .venv/
└─ .kivy/
```

PR #9138, opened 2025-09-08
:construction_worker: Fix cmake install error in macOS CI

AndreMiras:
Resolves brew install failure when cmake is already installed from a different tap (local/pinned vs homebrew/core). The workflow now uninstalls any existing cmake installation before installing from homebrew/core to avoid tap conflicts.
The error was:
```
Error: cmake was installed from the local/pinned tap
but you are trying to install it from the homebrew/core tap.
Formulae with the same name from different taps cannot be installed at the same time.
To install this version, you must first uninstall the existing formula:
  brew uninstall cmake
Then you can install the desired version:
  brew install cmake
```

Also use explicit macOS version 14 rather than latest.

Maintainer merge checklist

### DIFF
--- a/.github/workflows/osx_wheels_app.yml
+++ b/.github/workflows/osx_wheels_app.yml
@@ -18,7 +18,7 @@ env:
 jobs:
   kivy_examples_create:
     # we need examples wheel for tests, but only windows actually uploads kivy-examples to pypi/server
-    runs-on: macos-latest
+    runs-on: macos-14
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python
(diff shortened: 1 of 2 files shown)


misl6:
@AndreMiras why pinning a specific macOS version?

AndreMiras:
> @AndreMiras why pinning a specific macOS version?

I only see your comment now.
I thought for multiple reasons. I thought explicit was better than implicit, we don't have new OS versions that often and most likely when we do the CI would break until we fix it for the new version or decide to downgrade on the previous one explicitly. Also we had comment saying latest is 14 which would no longer be valid as latest changes but then we may forget these comments.
So in general I tend to like it better when it's explicit and that we bump versions as we support them. Same thing for Python versions, requirements, Docker base images, at least in the CI. Specially when working with limited resources.
Do you prefer it with latest?

PR #9140, opened 2025-09-09
Skip the copy of logo folder on `Android` and `iOS` to avoid warnings in logs

kuzeyron:
…because p4a uses something else so this is not needed and is annoying for users because it warns about needed permissions.


misl6:
It will very likely fail also on iOS (due to sandboxing), and not needed. Better to skip on all the mobile platforms.

kuzeyron:
Done

PR #9154, opened 2025-09-18
Add `glReadPixels_inplace` helper

gottadiveintopython:
Maintainer merge checklist

Hi, I want to use Kivy in a way similar to Manim — that is, to write animations in Kivy and generate video files from them. However, the features Kivy currently provides make this workflow somewhat inconvenient, so I’d like to add a new one.

### Why Kivy is currently inconvenient for this purpose

Kivy already has a Python API that captures the current frame: the `glReadPixels()`. However, each call to this function allocates and frees a memory buffer large enough to hold the screen’s pixels, which is inefficient.
https://github.com/kivy/kivy/blob/91dc72d7cc979f37e2b1344bf9edd5db7957bbd1/kivy/graphics/opengl.pyx#L1154-L1185

### Proposed functionality

This PR introduces `glReadPixels_inplace()`, which writes pixel data directly into a preallocated memory buffer.

```python
from kivy.graphics.opengl import glReadPixels_inplace, GL_RGB, GL_UNSIGNED_BYTE

out_buf = bytearray(screen_width * screen_height * 3)
glReadPixels_inplace(0, 0, screen_width, screen_height, GL_RGB, GL_UNSIGNED_BYTE, out_buf)
```


kivy/graphics/opengl.pyx
```diff
@@ -1151,6 +1151,14 @@ def glPolygonOffset(GLfloat factor, GLfloat units):
     '''
     cgl.glPolygonOffset(factor, units)
 
+def glReadPixels_inplace(GLint x, GLint y, GLsizei width, GLsizei height, GLenum format,
+                         GLenum type, unsigned char[::1] out_buf):
+    '''See: `glReadnPixels() on Kronos website
+    <http://www.khronos.org/opengles/sdk/docs/man/xhtml/glReadPixels.xml>`_
```
misl6: Looks good, can we better document it here?

gottadiveintopython:
I just realized that `glReadnPixels` isn’t available in OpenGL ES 2.0 — it was introduced in OpenGL ES 3.2. Would it be better to implement `glReadPixels_inplace` without relying on it?
https://registry.khronos.org/OpenGL/api/GLES2/gl2.h
https://registry.khronos.org/OpenGL/api/GLES3/gl32.h

kuzeyron (dismissed):

misl6:
> I just realized that `glReadnPixels` isn’t available in OpenGL ES 2.0 — it was introduced in OpenGL ES 3.2. Would it be better to implement `glReadPixels_inplace` without relying on it? https://registry.khronos.org/OpenGL/api/GLES2/gl2.h https://registry.khronos.org/OpenGL/api/GLES3/gl32.h

We can add a guard about the availability on OpenGL ES 3.2 (and above), and if not available we can decide to use another (slower I guess) implementation.

gottadiveintopython:
I don't think I can fix the unit test failures on Windows because I don't have a Windows machine.

Also, here are the reasons why I implemented things the way I did:

- I didn't use the pytest parametrization feature because [it doesn't work with unittest-based tests](https://docs.pytest.org/en/stable/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses).
- I split `glReadPixels_inplace` into two functions instead of using a single function that checks the availability of `glReadnPixels` inside of it, so that both implementations can be tested.

```python
# If we do this, we cannot test both paths, right?
def glReadPixels_inplace(...):
    if cgl.glReadnPixels == NULL:
        ...
    else:
        ...
```

However, I'm not sure this was the right choice because it makes the implementation uglier. Let me hear your opinion.

gottadiveintopython:
@misl6 Can I have your opinion?

The current implementation checks for the availability of `glReadnPixels` at import time.
Should this check instead be performed at call time, like so?

```python
def glReadPixels_inplace(...):
    if cgl.glReadnPixels == NULL:
        # Implementation that does not rely on glReadnPixels
    else:
        # Implementation that relies on glReadnPixels
```

misl6:
> @misl6 Can I have your opinion?
> 
> The current implementation checks for the availability of `glReadnPixels` at import time. Should this check instead be performed at call time, like so?
> 
> ```python
> def glReadPixels_inplace(...):
>     if cgl.glReadnPixels == NULL:
>         # Implementation that does not rely on glReadnPixels
>     else:
>         # Implementation that relies on glReadnPixels
> ```

Checked the current implementation in this PR, and I do not see any potential drawbacks if kept during import time.
Maybe, just explain with a few comments what you're doing in that chunk of code.

Style side: At call time looks cleaner, but is also less efficient. But I prefer efficiency, if well documented.

gottadiveintopython:
@misl6 Is it okay to skip unit tests on Windows? I’m unable to fix the errors.

misl6:
> @misl6 Is it okay to skip unit tests on Windows? I’m unable to fix the errors.

Can you rebase on top of latest master so we can check what's failing? Actually I think it might be related to something that might be already been fixed.

gottadiveintopython:
@misl6 The same error also occurs when testing `glReadPixels`, as shown in the CI tests for the latest commit. Therefore, I believe the issue lies in the Windows CI setup rather than in the changes I made.

misl6 (approved):
LGTM. Thank you!

misl6:
@gottadiveintopython , may I ask you to post here the stack trace of the failing Windows CI, so anyone going back in the git blame and checking why the execution of the tests are disabled on Windows can have a better overview?

gottadiveintopython:
```
================================== FAILURES ===================================
____________ Test_glReadPixels_inplace.test_a_buffer_of_exact_size ____________

self = <kivy.tests.test_graphics.Test_glReadPixels_inplace testMethod=test_a_buffer_of_exact_size>

    @requires_graphics
    def test_a_buffer_of_exact_size(self):
        from kivy.graphics.opengl import GL_RGB, GL_UNSIGNED_BYTE, glReadPixels_inplace
        self.Window.clearcolor = (1, 0, 0, 1)
        self.advance_frames(1)
        buf = bytearray(b'ABCDEF')
        glReadPixels_inplace(0, 0, 1, 1, GL_RGB, GL_UNSIGNED_BYTE, memoryview(buf)[:3])
>       assert buf == b'\xff\x00\x00DEF'
E       AssertionError: assert bytearray(b'ABCDEF') == b'\xff\x00\x00DEF'
E         
E         At index 0 diff: 65 != 255
E         
E         Full diff:
E         - (b'\xff\x00\x00DEF')
E         + bytearray(b'ABCDEF')

kivy\tests\test_graphics.py:1166: AssertionError
______________ Test_glReadPixels_inplace.test_a_buffer_too_large ______________

self = <kivy.tests.test_graphics.Test_glReadPixels_inplace testMethod=test_a_buffer_too_large>

    @requires_graphics
    def test_a_buffer_too_large(self):
        from kivy.graphics.opengl import GL_RGB, GL_UNSIGNED_BYTE, glReadPixels_inplace
        self.Window.clearcolor = (1, 0, 0, 1)
        self.advance_frames(1)
        buf = bytearray(b'ABCDEF')
        glReadPixels_inplace(0, 0, 1, 1, GL_RGB, GL_UNSIGNED_BYTE, buf)
>       assert buf == b'\xff\x00\x00DEF'
E       AssertionError: assert bytearray(b'ABCDEF') == b'\xff\x00\x00DEF'
E         
E         At index 0 diff: 65 != 255
E         
E         Full diff:
E         - (b'\xff\x00\x00DEF')
E         + bytearray(b'ABCDEF')

kivy\tests\test_graphics.py:1175: AssertionError
```

As shown, nothing is written to the output buffer, and  `kivy.graphics.glReadPixels` seems to have the same issue:

```python
class Test_glReadPixels(GraphicUnitTest):
    @requires_graphics
    def test_glReadPixels(self):
        from kivy.graphics.opengl import GL_RGB, GL_UNSIGNED_BYTE, glReadPixels
        self.Window.clearcolor = (1, 0, 0, 1)
        self.advance_frames(1)
        pixels = glReadPixels(0, 0, 1, 1, GL_RGB, GL_UNSIGNED_BYTE)
        assert pixels == b"\xff\x00\x00"
```

```
================================== FAILURES ===================================
_____________________ Test_glReadPixels.test_glReadPixels _____________________

self = <kivy.tests.test_graphics.Test_glReadPixels testMethod=test_glReadPixels>

    @requires_graphics
    def test_glReadPixels(self):
        from kivy.graphics.opengl import GL_RGB, GL_UNSIGNED_BYTE, glReadPixels
        self.Window.clearcolor = (1, 0, 0, 1)
        self.advance_frames(1)
        pixels = glReadPixels(0, 0, 1, 1, GL_RGB, GL_UNSIGNED_BYTE)
>       assert pixels == b"\xff\x00\x00"
E       AssertionError: assert b'\x00\x00\x00' == b'\xff\x00\x00'
E         
E         At index 0 diff: b'\x00' != b'\xff'
E         
E         Full diff:
E         - b'\xff\x00\x00'
E         ?     ^^
E         + b'\x00\x00\x00'
E         ?     ^^

kivy\tests\test_graphics.py:1159: AssertionError
```

PR #9168, opened 2025-09-30
Document `Builder` attributes

clayote:
I thought I had to keep track of what kv-lang files I'd loaded on my own, but `Builder` does that. I'm documenting it so other people don't think they have to roll their own file tracking, either. And why not document the other attributes while I'm at it? They are, apparently, public.

Maintainer merge checklist


misl6:
Seems we have a PEP8 failure, can you take a look at it?

misl6 (approved):
LGTM. Thank you!

PR #9177, opened 2025-10-15
Fix ZeroDivisionError when showing hidden window on Windows

g-k-s-03:
### Summary
This PR fixes a ZeroDivisionError that occurs when a hidden window is shown on Windows. The fix calls `_update_density_and_dpi()` after `Window.show()`.

### Issue
#9170 

### How to test
```python
from kivy.core.window import Window
Window.hide()
Window.show()  # Should not raise ZeroDivisionError
assert Window._density > 0

PR #9189, opened 2025-10-24
Fix `ZeroDivisionError` when showing hidden window on Windows

g-k-s-03:
closes #9177
This PR fixes the indentation and improves clarity in the show() method of the SDL3 window provider (kivy/core/window/window_sdl3.py).
The update ensures the comment and function call for updating density and DPI are properly indented within the if self._is_desktop block.
Additionally, this fix prevents potential ZeroDivisionError when the window starts hidden.

### DIFF
--- a/kivy/core/window/window_sdl3.py
+++ b/kivy/core/window/window_sdl3.py
@@ -402,8 +402,8 @@ class WindowSDL(WindowBase):
     def show(self):
         if self._is_desktop:
             self._win.show()
-            # Update density and DPI now that window is visible
-            # This fixes ZeroDivisionError when window starts hidden
+            # Update density and DPI now that window is visible.
+            # This fixes ZeroDivisionError when window starts hidden.
             self._update_density_and_dpi()
         else:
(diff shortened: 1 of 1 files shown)


kivy/core/window/window_sdl3.py
```diff
@@ -401,7 +401,10 @@ def hide(self):
 
     def show(self):
         if self._is_desktop:
-            self._win.show_window()
+            self._win.show()
```
misl6, line 404: Hi, can you explain why `self._win.show()` instead of `self._win.show_window()` ?

g-k-s-03: Hi @misl6,

I use self._win.show() instead of self._win.show_window() because show() is the higher-level method that not only makes the window visible but also updates internal states like density and DPI. This ensures the window renders correctly and prevents potential ZeroDivisionError when starting with a hidden window.

g-k-s-03: @misl6, if you want me to change it, I can change it

PR #9194, opened 2025-10-31
Add initial support for mobile (iOS, Android) wheels

psychowasp:
This pull request introduces significant improvements for iOS platform support in the Kivy build system, primarily by integrating iOS framework handling with `cibuildwheel` and updating build scripts and logic to streamline cross-platform compilation. The changes automate iOS dependency installation, refactor framework detection and linking, and modernize platform-specific build flags. iOS builds now use the standard build pipeline rather than relying on the legacy `kivy-ios` environment.

**iOS Build Integration and Framework Handling**
* Added `.ci/ios_frameworks.sh` script to automate downloading and unpacking required iOS frameworks and dependencies, including SDL3 and ANGLE, for use in builds. (`.ci/ios_frameworks.sh`)
* Updated `pyproject.toml` to run the new iOS framework setup script before building with `cibuildwheel`, enabling automated framework preparation for iOS builds. (`pyproject.toml`)

**Platform Detection and Build Logic Refactoring**
* Refactored platform detection in `setup.py` to rely on `cibuildwheel` for iOS builds, removing legacy `kivy-ios` environment checks and adding logic to collect and organize iOS framework paths and headers. (`setup.py`) [[1]](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L162-R171) [[2]](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R211-R280)
* Updated build flag determination functions (`determine_angle_flags`, `determine_gl_flags`, `determine_sdl3`) to use the new framework organization and paths for iOS, ensuring correct linking and inclusion of headers for ANGLE, EGL, GLESv2, and SDL3 frameworks. (`setup.py`) [[1]](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L669-R786) [[2]](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L691-R823) [[3]](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R891-R893)

**Platform-Specific Compilation and Linking Flags**
* Modernized iOS-specific build flags for core image, AVFoundation, and other subsystems to use the correct frameworks and headers, including new dependencies required for App Store support and improved compatibility with Xcode 16 and macOS 14.5+. (`setup.py`) [[1]](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R1139-R1148) [[2]](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R1160-R1197)

**Mobile Platform Compatibility**
* Improved sleep function handling in `kivy/clock.py` for mobile platforms by using `time.sleep` on Android and iOS, enhancing cross-platform reliability. (`kivy/clock.py`)

**General Codebase Maintenance**
* Minor cleanups and reordering of imports in `setup.py` to improve readability and maintainability. (`setup.py`)

These changes collectively make the iOS build process more robust, automated, and maintainable, aligning it with modern Python packaging and CI practices.

### DIFF
--- a/kivy/uix/widget.py
+++ b/kivy/uix/widget.py
@@ -1573,7 +1573,6 @@ class Widget(WidgetBase):
                 self.inc_disabled()
             else:
                 self.dec_disabled()
-            return True
 
     def inc_disabled(self, count=1):
         self._disabled_count += count


psychowasp:
i added this ios_frameworks.sh to .ci 
which cibuildwheel will run before starting to build wheel and downloads SDL / Angle frameworks..

not sure howto deal with the paths in both .sh and setup.py when it comes to the dist/Frameworks..

@misl6 but let me know what you think so far, and what i need to change

how its been tested so far [ios_setup_py.yml](https://github.com/kivy-school/kivy-mobile-builder/blob/main/.github/workflows/ios_setup_py.yml)

setup.py.bak
```diff
@@ -0,0 +1,1333 @@
+#
```
kuzeyron: I don't think this file should be included

psychowasp: Nope it shouldnt but lets just call this a draft for now..

Right now i hat no way of testing if my changes effects building on other platforms.
So atm it was also to trigger the tests, but i see they fail with all PRs atm..

.ci/ios_frameworks.sh
```diff
@@ -0,0 +1,19 @@
...
+    SDL_FW="$FWS_DIR/SDL3.xcframework/$PLATFORM/SDL3.framework"
+    cp -rf $SDL_FW/Headers $SDL_FW/Headers/SDL3
+done
+
+
+ANGLE_VERSION="chromium-7151_rev1"
+
+#wget -O angle-iphoneall-universal.tar.gz "https://github.com/kivy/angle-builder/releases/download/$ANGLE_VERSION/angle-iphoneall-universal.tar.gz"
+curl -LO "https://github.com/kivy/angle-builder/releases/download/$ANGLE_VERSION/angle-iphoneall-universal.tar.gz"
+tar -xzvf angle-iphoneall-universal.tar.gz -C $FWS_DIR
```
misl6: What about moving all this part to `./tools/build_ios_dependencies.sh` ?

We can also considering adding there the build scripts (that I expect are really similar to the macOS ones) for SDL3.

... or move `./tools/build_macos_dependencies.sh` to `./tools/build_apple_dependencies.sh` and produce fat frameworks with the required platforms.

kivy/clock.py
```diff
@@ -502,6 +502,10 @@ def _usleep(microseconds, obj=None):
                 obj, ctypes.byref(delay), 0,
                 ctypes.c_void_p(), ctypes.c_void_p(), False)
             _kernel32.WaitForSingleObject(obj, 0xffffffff)
+    elif platform in ['android', 'ios']:
+        # on mobile platforms, just use time.sleep
+        def _usleep(microseconds, obj=None):
+            time.sleep(microseconds / 1000000.)
```
misl6, line 508: Does this one also works for legacy `kivy-ios` builds? (short-term talking)

.ci/ios_frameworks.sh
psychowasp: You mean make xcframework containing both macOS / iOS / iOS-Sim ?

kivy/clock.py
psychowasp: i assume kivy-ios triggeres platform == “ios” as always..
and when running it on kivy-ios the whole try fails and falls back to the usleep in exception..

so i assume it will just pick the one under ios now and not fall back to the except one, which is just the same..

problem just was whatever didnt work with ctypes before, works now and no longer triggers the except when running ios.. 

only difference now is we no longer need to fake platform as ios
cibuildwheel will set it as so, and same for the sys_platform in dependencies, it no longer triggered as “darwin” but “ios”

.ci/ios_frameworks.sh
psychowasp: IMG_5019

cant speak for kivy desktop but, just keep embed it into the wheel as before, no need to change that..

but for ios i still think its better that the frameworks are embeded by xcode, as we current do..

both kivy-ios and psproject has the ability to add required frameworks..

most important is just that there is example some wheel to download that contains the frameworks, like we do with psproject / anaconda.org/kivyschool.

https://anaconda.org/kivyschool/kivy-sdl3-angle/files/manage

both for ios embeded but also if working on kivy for ios.. 

in current state of psproject, mac platform is also included, and it just take advantage that kivy supplies universal build, and fills out the gap for both intel/arm64

im just telling it only to include the frameworks when ios is picked, and not when mac (since its embeded into the wheels)..

How both briefcase / psproject deals with site-packages works like this
briefcase got 2 set of site-packages
* site_packages.iphoneos
* site_packages.iphonesimulator

and wheels are always a set of 3 
xxx-ios_13_0_arm64_iphoneos
xxx-ios_13_0_arm64_iphonesimulator
xxx-ios_13_0_x86_64_iphonesimulator

when pip installing you add parameter
platform=ios_13_0_arm64_iphoneos

if pip installing into site_packages.iphoneos

platform=ios_13_0_arm64_iphonesimulator

if pip installing into site_packages.iphonesimulator

if currently working on arm64 mac.. 

psproject just checks the host arch and picks between the 2 simulator wheels that way.. 

inside xcode you just got this run script that checks if you currently building for iphoneos or iphonesimulator, and will copy the site_packages.xxx to app/site_packages 

ofc psproject is expanded to both check if iphoneos/iphonesimulator else macOS
so it has a site_packages.macos 

i just choosed todo 
site_packages/iphoneos
site_packages/Iphonesimulator
site_packages/macos

psychowasp: > What about moving all this part to `./tools/build_ios_dependencies.sh` ?
> 
> We can also considering adding there the build scripts (that I expect are really similar to the macOS ones) for SDL3.
> 
> ... or move `./tools/build_macos_dependencies.sh` to `./tools/build_apple_dependencies.sh` and produce fat frameworks with the required platforms.

Whatever works for you guys 

most important is that the right path to the .sh is stated here https://github.com/kivy-school/kivy/blob/65b3b92e0b57e497894f790658629b62aa817ca6/pyproject.toml#L16

atm its not 100 % currect, because should just say .ci/ios_frameworks.sh

but since i been testing it external, it wanted kivy/ first..

i assume same could be done with
```toml
[tool.cibuildwheel.macos]
before-all= ./tools/build_macos_dependencies.sh
```

so u got like 
```toml
[tool.cibuildwheel.ios]
before-all= ./tools/build_ios_dependencies.sh
```

but would be smarter that build_depencies (if building) hat its own run / own release of wheel set and before-all script just need to download the binary for each build.. 

and the rest of us still got some link to downloading the depencies.. 
that example was a true nightmare when trying to support PySwiftKit + kivy…
because i hat to use kivy-ios, i hat to BUILD EVERYTHING EVERY RELEASE OF PYTHON PACKAGE….

psychowasp: [https://github.com/kivy-school/kivy/blob/52e867e31ca595ad6a1fcba1b304fc8cb1d3e098/pyproject.toml#L16](https://github.com/kivy-school/kivy/blob/52e867e31ca595ad6a1fcba1b304fc8cb1d3e098/pyproject.toml#L16)

[https://github.com/kivy-school/kivy/blob/52e867e31ca595ad6a1fcba1b304fc8cb1d3e098/.ci/ios_frameworks.sh#L3](https://github.com/kivy-school/kivy/blob/52e867e31ca595ad6a1fcba1b304fc8cb1d3e098/.ci/ios_frameworks.sh#L3)

[https://github.com/kivy-school/kivy/actions/runs/19409890707/workflow](https://github.com/kivy-school/kivy/actions/runs/19409890707/workflow)

mhsmith:
@freakboy3742: FYI

.ci/ios_frameworks.sh
misl6: Let's move it to tools, and if we can merge it to `build_macos_dependencies` to avoid almost duplicate code (via a unification of iOS and macOS deps build logic) that does similar things even better (otherwise, we will look into it later).

kivy/clock.py
misl6: Got it!

tools/cibuildwheel-before-all-macos.sh
```diff
@@ -0,0 +1,3 @@
+
+# consider prebuild sdl frameworks somewhere else and this script for cibuildwheel before-all just downloads it
+# and puts it where setup.py wants it to be.
```
misl6, line 3: AFAIK `cibuildwheel before-all` step could also perform build operations, can't  the current `tools/build_macos_dependencies.sh` be used? (which builds almost everything from scratch except for ANGLE, which takes a lot to complete, and requires additional setup for devs if the build is performed locally).

Actually, we would prefer the same for iOS SDL3 frameworks, to keep SDL3 versioning full control in the Kivy repo.

Would it possible to move the SDL3 deps build for iOS here instead of downloading it from https://pypi.anaconda.org/kivyschool/simple ?

psychowasp: https://pypi.anaconda.org/kivyschool/simple is also only temporary as long, we didnt have access to a prebuild set of Angle / SDL3..

but should it not more be like install_macos_dependencies.sh vs build ?
Same for ios.. 

but dont know where you guys going to build/host them?.. 

but isnt that something you guys can just edit in another PR afterwards ? 

Should just be the ios.sh that needs changes to your likings.. 

I just kept anaconda.org/kivyschool which all components is based on the versions used on kivy-ios, until there is some official deps from kivy gibhub.

Because its abit the chicken vs the egg now and which comes first xD.

psychowasp:
i personal like the way beeware has the 
* OpenSSL
* XZ
* libFFI
* Bzip

as seperate releases, and always easy to get same builds as the pythons.

same would be nice when working for stuff for kivy or kivy itself..

just get the sdl3/angle deps directly from kivy github (kivy-deps repo), which matches either master branch or a release.

freakboy3742:
> i personal like the way beeware has the
> 
> * OpenSSL
> * XZ
> * libFFI
> * Bzip
> 
> as seperate releases, and always easy to get same builds as the pythons.
> 
> same would be nice when working for stuff for kivy or kivy itself..

FWIW: The aim is that those four in particular *won't* be BeeWare's responsibility for much longer - responsibility will be passed upstream to CPython. At the CPython core team sprint in September, there was some discussion about a unified approach to maintaining binary dependency assets. Nothing has come of that *yet*, but I believe "for the 3.15 release" was the broad timeframe being suggested.

misl6:
> i personal like the way beeware has the
> 
> * OpenSSL
> * XZ
> * libFFI
> * Bzip
> 
> as seperate releases, and always easy to get same builds as the pythons.
> 
> same would be nice when working for stuff for kivy or kivy itself..
> 
> just get the sdl3/angle deps directly from kivy github (kivy-deps repo), which matches either master branch or a release.

For experience, maintaining separate releases for "non-standard" dependencies is quite painful and adds an additional burden to maintain, for which we do not have enough workforce. (See dependencies for Windows, which we discussed multiple times to be migrated here).

A separate release means:
- A separate repo (generic like kivy-sdk-packager is not always a good idea, and kivy-sdk-packager is a valid example)
- More CI/CD pipelines to mantain
- Un-necessary releases of the dependencies, that does not always fit into a release of the "real" product (See how many SDL3 releases we already made for Windows, with Kivy 3.x.x which is nowhere near to a release).

So, except to ANGLE, which is a different beast, we strongly encourage to keep the dependencies build internal. But yes, it can be changed in a later PR, but if you already have the build scripts, it would be nice to include it since day 0 :)

psychowasp:
Well both ways got its pro and cons..


but if SDL3+rest have to be build each time you have to make a wheel release, then its also important they are bundled with it, so they are assets in the release or embeded into the kivy build..

What worked with kivy-ios toolchain, i dont think will work well with new pep730 standards.

In the end goal of pep730 is about moving away from having to explicit build for ios local. And have same benefits of wheel install like on example macos .. 

And whatever building is required should happen in the cloud.. 

I just know for PSProject and Briefcase to run to current kivy ios wheels, we just need to pip install them, and download the same SDL3+Rest of frameworks as kivy was build with.. 

so for those 2 project makers, its important that ios wheel is present and if external frameworks is needed, they should be easy to fetch as prebuild frameworks from upstream, and is matching kivy build.. 


But sure is best to have it all ready from "0 day" but i think its a tricky one atm..

again just so we all clear about what will happen now xD

this PR will break kivy-ios and the recipe to build it.. 

i hat hoped the current kivy tests would hat made it more clear if any of my changes hat bricked other platforms.. 
but seems like i said ealier its only after builds the tests fails..

i also tried not to affect Android side, but in the end that should also be the next thing todo, implement Android Wheel.. and that properly also going to change how you need to think in terms of how p4a needs to work in the future.. 

Right now im just waiting for example meson-python to make release of current master, then Numpy + some others gets native ios wheel support out of the box.. 

so only reason we still would need recipes/building at all, is those few cases which still have to be fixed completely upstream..

misl6:
Ok, so, I'm performing some tests, and I've been able to run a modified version of kivy-ios generated project with Python 3.14.2 (self-built) + your produced wheels.

It works, but the startup time is quite long. Is happening the same on your side?

FYI: Will work on `pyobjus` so we can have an official build that supports iOS + I will move the `ios` "package" code out of `kivy-ios` (that will cease to exists as-is).

For the binary dependencies (SDL3, ANGLE), I'm still thinking about a viable solution for us. How psproject and beeware make sure to select the right release for a specific Kivy release ATM?

psychowasp:
I have tried explained it on discord now..

just ask there and we can always write a final summary here..

psychowasp:
*** issue fixed *** bad sync of kivy-master -> kivy-school/kivy-master

Issue #8926, opened 2024-11-14
Can't scroll horizontally on scrollview nested inside another scrollview


mz3r0:
**Software Versions**
* Python: 3.12.0
* OS: Windows 10 Pro Version 22H2 Build 19045.5131
* Kivy: 2.3.0
* Kivy installation method: pip

**Description**

In the widget hierarchy I have an outer ScrollView and an inner ScrollView and I want to scroll horizontally in the parent, as long as the mouse isn't hovering on the inner one. I set do_scroll_y as False on the outer SV, and do_scroll_x as False on the inner. To my experience so far, setting these only affect the scroll bar and don't help me achieve the desired result.

Note that horizontal scrolling works only when I drag and move. If I set `scroll_type: ['bars']`, to disable swiping gestures on my app (since it's intended for Desktop use), the only way to scroll is by dragging the scroll bar, but that's not what I want.

**Expected behavior**
I want to use the scroll wheel normally (that is, without holding shift like I do on web pages for horizontal scrolling) and see the outer SV scroll horizontally. I want to use the **wheel** for that.

**To Reproduce**

```python
from kivy.lang import Builder
import pandas as pd
import numpy as np

from kivy.uix.recycleview.views import RecycleDataViewBehavior
from kivy.uix.label import Label
from kivy.uix.recycleview import RecycleView
from kivy.app import App

from kivy.core.window import Window

KV = """
Screen:
    name: 'main'

    ScrollView:
        do_scroll_y: False
        
        scroll_wheel_distance: dp(100)
        bar_width: dp(10)
        bar_color: 146/255, 146/255, 146/255, 1
        bar_inactive_color: 146/255, 146/255, 146/255, 1
        
        BoxLayout:
            orientation: 'horizontal'
            padding: [30]
            spacing: 30
            size_hint_x: None
            width: self.minimum_width
            
            BoxLayout:
                orientation: 'vertical'
                size_hint_x: None
                width: 600
                height: self.minimum_height
                
                MyRecycleView:
                
            BoxLayout:
                orientation: 'vertical'
                size_hint_x: None
                width: 400
                
            BoxLayout:
                orientation: 'vertical'
                size_hint_x: None
                width: 400
                
            BoxLayout:
                orientation: 'vertical'
                size_hint_x: None
                width: 400

<MyTableCell>:
    text_size: self.size
    adaptive_height: True
    halign: "center"

<MyRecycleView>:
    do_scroll_x: False
    viewclass: "MyTableCell"
[...]

PR #9195, opened 2025-10-31
ScrollView Nesting Support and update

ElliotGarbus:
This commit adds support for arbitrary levels of nesting.  All configurations of nesting are supporting including Orthogonal, parallel and mixed nesting to arbitrary depth levels.
 
The changes include a major refactoring of ScrollView to improve readability and maintainability.
This PR fixed the following issues:
https://github.com/kivy/kivy/issues/5617
https://github.com/kivy/kivy/issues/6889
https://github.com/kivy/kivy/issues/4052
https://github.com/kivy/kivy/issues/2986
https://github.com/kivy/kivy/issues/8926

In _event.pyx, I have deprecated these calls: dispatch_children() and dispatch_generic(). They were used in ScrollView, there are not used anywhere else in Kivy. They are also not really documented. An argument could be made for just deleting them.  They had been used previously to recursively issue the on_scroll events. 

3 examples are added to the examples directory:
- An example of orthogonally nested recycleviews
- An example of orthogonally nested ScrollViews
- An example of parallel nested ScrollViews.

A change was also made the to the state-machine that determines touch or scroll. If the users is dwelling on a label (or any widget that does not consume the touch) past the time threshold, the touch is turned to a scroll.  Allowing users to dwell on text without the "losing the touch" after the timeout.

The on_scroll_start/move/stop have all been updated to reflect the scroll motion.  The previous implementation mirrored the on_touch_methods.

Nested ScrollViews
------------------
An application can have multiple levels of nested ScrollViews.  The
ScrollView class manages the hierarchy of nested ScrollViews.

It is important to set the :attr:`do_scroll_x` and :attr:`do_scroll_y`
properties to the correct values for the nested ScrollViews.
The nested scrolling behavior, described below, is only applied if the
:attr:`do_scroll_x` and :attr:`do_scroll_y` properties are set to the
appropriate values.

To achieve orthogonal scrolling, for example, the outer
ScrollView should have :attr:`do_scroll_x` set to False and the inner
ScrollView should have :attr:`do_scroll_y` set to True, or vice versa.

For Parallel scrolling, both ScrollViews should have the same value for
:attr:`do_scroll_x` and :attr:`do_scroll_y`.  This can allow vertical in
vertical,  horizontal in horizontal, or XY in XY scrolling.

For mixed scrolling, the outer ScrollView should have :attr:`do_scroll_x`
and :attr:`do_scroll_y` set to True, and the inner ScrollView should have
either :attr:`do_scroll_x` or :attr:`do_scroll_y` set to False.

**Orthogonal Scrolling** (outer and inner scroll in different directions):
    - Touch: Each ScrollView scrolls only in its direction
    - Wheel: Innermost ScrollView scrolls if it supports that axis
    - Example: Vertical outer + Horizontal inner

**Parallel Scrolling** (outer and inner scroll in the same direction):
    - Touch: Web-style boundary delegation (see below)
    - Wheel: Innermost ScrollView scrolls, no delegation to outer
    - Scrollbar: Scroll does not affect other ScrollView
    - Example: Vertical outer + Vertical inner

**Mixed Scrolling** (outer scrolls XY, inner one axis, or vice versa):
    - Shared axis: Web-style boundary delegation
    - Exclusive axes: Immediate delegation or inner-only scrolling
    - Wheel: Innermost ScrollView scrolls if it supports the axis
    - Example: XY outer + Horizontal inner

Web-Style Boundary Delegation
----------------------------

For parallel and shared-axis scrolling, ScrollView uses web-style
delegation:

- If a touch starts at the inner boundary and moves out, it delegates
      scrolling to the outer ScrollView right away.
- If a touch starts at the inner boundary and moves in, only the inner
      ScrollView scrolls.
- If a touch starts not at a boundary, only the inner ScrollView
      scrolls, even if the gesture later reaches a boundary. Delegation to
      the outer never occurs mid-gesture in this case.
- A new touch starting at the boundary is required to delegate to the
      outer.

Disable this behavior by setting :attr:`parallel_delegation` to False.

Wheel Behavior in Nested ScrollViews
------------------------------------

When the mouse scroll wheel or a trackpad is used, ScrollView always
applies web-style delegation:

  - Only the innermost ScrollView under the pointer/cursor will handle
    the wheel event.
  - If that ScrollView cannot scroll further along the wheel axis, the
    event is not propagated to outer ScrollViews. This matches standard
    browser and OS behavior.
  - Outer ScrollViews never respond to wheel events unless the pointer is
    over their scrollbars or unless there are no deeper nested
    ScrollViews.

This prevents "scroll hijacking" and provides an intuitive nested scroll
experience.

**Examples:**
  - In a vertical ScrollView containing a horizontal ScrollView, using the
    wheel over the horizontal ScrollView means only the vertical (outer)
    ScrollView scrolls, if its direction matches the wheel.
  - For two nested vertical ScrollViews, the innermost one under the
    pointer will scroll with the wheel until it cannot scroll further;
    wheel events are not delegated to the parent.

Wheel behavior is always active and is NOT affected by the
:attr:`parallel_delegation` or :attr:`delegate_to_outer` properties.
Those only control touch and touchpad gesture behavior.

 Three new properties are added 
-----------------------------------
**slow_device_support = BooleanProperty(False)**
Enable slow device support for scroll gesture detection.
    
On very slow devices, at least 3 frames are needed to accumulate
velocity data for scroll effects to work properly. When enabled,
after the scroll timeout expires, the gesture handoff will be delayed
until at least 3 frames have rendered, ensuring sufficient velocity
data accumulation.

Background: This 3-frame delay was added in 2014 to support low
 performance  android devices.  The property makes this additional
delay optional.
   
This addresses issues #1464 and #1499 for low-performance devices.
Disable this on modern hardware to improve touch responsiveness.

**parallel_delegation = BooleanProperty(True)**
Controls boundary delegation behavior for parallel nested ScrollViews.
    
When True (default, web-style):
- Touch starting at inner boundary, moving away → delegates to outer
- Touch starting not at boundary → scrolls inner only, never delegates
    
When False:
- No delegation, only touched ScrollView scrolls

**delegate_to_outer = BooleanProperty(True)**
Controls whether touch scroll gestures delegate to outer ScrollViews.
    
When True (default):
- Orthogonal: Cross-axis gestures immediately delegate to outer
 (e.g., horizontal swipe in vertical-only ScrollView)
- Parallel: At boundaries, delegates to outer 
 (respects parallel_delegation)
- Arbitrary depth: Continues searching up hierarchy for 
capable ScrollView
    
When False:
- No delegation to outer ScrollViews
- Only the directly touched ScrollView handles the gesture
    
NOTE: Mouse wheel events are NOT affected by this property. Wheel scrolling
always uses web-style behavior: scroll the innermost ScrollView under cursor
that can handle the direction. 
    
Example use cases:
- Set False to lock touch scrolling to a specific nested level
- Set False to prevent inner touch scroll from affecting outer scroll
    

Maintainer merge checklist

### DIFF
--- a/examples/widgets/scrollview_nested_orthogonal.py
+++ b/examples/widgets/scrollview_nested_orthogonal.py
@@ -113,7 +113,7 @@ class HorizontalScrollRow(BoxLayout):
         label = Label(
             text=f'Row {self.row_index + 1} - Horizontal Scroll',
             size_hint_x=None,
-            width=200,
+            width='200dp',
             color=[1, 1, 1, 1],
             bold=True
         )
@@ -124,7 +124,7 @@ class HorizontalScrollRow(BoxLayout):
             btn = Button(
                 text=f'Btn {j + 1}',
                 size_hint_x=None,
-                width=100,
+                width='100dp',
                 background_color=[0.2 + (self.row_index * 0.1) % 0.8,
                                   0.3, 0.7, 1]
             )
@@ -143,7 +143,7 @@ class VerticalScrollColumn(BoxLayout):
         label = Label(
             text=f'Column {self.column_index + 1}\nVertical Scroll',
             size_hint_y=None,
-            height=40,
+            height='40dp',
             color=[1, 1, 1, 1],
             bold=True
         )
@@ -154,9 +154,9 @@ class VerticalScrollColumn(BoxLayout):
             btn = Button(
                 text=f'Item {j + 1}',
                 size_hint_y=None,
-                height=60,
-                background_color=[0.7, 0.3, 0.2 +
-                                  (self.column_index * 0.1) % 0.8, 1]
+                height='60dp',
+                background_color=[0.7, 0.3,
+                                  0.2 + (self.column_index * 0.1) % 0.8, 1]
             )
             content.add_widget(btn)
 


ElliotGarbus:
A few additional comments:
The behavior of the on_scroll events has been updated.  The on_scroll_start, on_scroll_move, and on_scroll_start originally simply mirrored the on_touch_* events and were used to implement recursive issuing of events.  The events now fire based on movement.

These events had also relied on a non-standard event dispatch mechanism that silently returned if a widget did not support the event.  I have deprecated these calls: dispatch_children() and dispatch_generic().  They were used in ScrollView, there are not used anywhere else in Kivy.  They are also not really documented.  An argument could be made for just deleting them.

I used the following deprecation message: 'Deprecated in 3.0.0, Will be removed after two releases')

ElliotGarbus:
A video of 12 scrolling configurations in an XY scrollview.
https://github.com/user-attachments/assets/ec417493-5bb1-4cfb-b098-dbc9d75744c5

ElliotGarbus:
FWIW The failures are not related to this PR.

ElliotGarbus:
Examples added to exampled directory.

examples/widgets/scrollview_nested_orthogonal.py
```diff
@@ -0,0 +1,186 @@
+"""
```
kuzeyron, line 1: With this code the fourth scroller on the left side locks itself after randomly dragging them around. I don't see this behavior with the rest on Linux and Android.

ElliotGarbus: Any other detail on how you were testing?  Mouse, Touchpad, TouchScreen? 
Are you using the left and right scroll views simultaneously?
Using the touchscreen on my Windows laptop I can not reproduce the issue.

Any details would be helpful.

ElliotGarbus: Fixed and pushed.

Samael-TLB:
> In _event.pyx, I have deprecated these calls: dispatch_children() and dispatch_generic().

@ElliotGarbus I would request you to not make them deprecated. These will be used in an upcoming pr that I'm doing to revamp event bubbling in Kivy. Atleast dispatch_children(), should be kept as its an handy and efficient iteration on children done for an event. If removed, then same logic would need to be implemented in python and might not be that efficient in large widget trees. In widget, touch events the same logic in python is being used which can be replaced with this for a cleaner and efficient iteration.

ElliotGarbus:
@Samael-TLB It is easy enough to rollback the deprecation - but before I do, lets have a discussion on what generic dispatch does.  It is far from efficient.  Generic dispatch is a simple wrapper around dispatch that checks that the dispatched event is supported prior to making a call to dispatch.  The method dispatch_children() calls generic_dispatch().  This was used to walk the widget tree and recursively send on_scroll_* events down the widget tree to other scrollviews.  This is quite inefficient.  It's core value it that it creates the illusion of sending an event to a widget that does not support that event.  Is this what you want?

Samael-TLB:
> @Samael-TLB It is easy enough to rollback the deprecation - but before I do, lets have a discussion on what generic dispatch does. It is far from efficient. Generic dispatch is a simple wrapper around dispatch that checks that the dispatched event is supported prior to making a call to dispatch. The method dispatch_children() calls generic_dispatch(). This was used to walk the widget tree and recursively send on_scroll_* events down the widget tree to other scrollviews. This is quite inefficient. It's core value it that it creates the illusion of sending an event to a widget that does not support that event. Is this what you want?

After discussion with @ElliotGarbus, we have decided to keep the deprecated warning for dispatch children for now and based on further fixation of the pipeline, we might reintroduce an appropriate function that can dispatch to children without needing to replicate the logic in each Event Dispatcher

kuzeyron:
LGTM

ElliotGarbus:
@kuzeyron 
The deprecation messages read  'Deprecated in 3.0.0, Will be removed after two releases'
The documentation that adds new properties is tagged:     " .. versionadded:: NEXT_VERSION"

Let me know if or when either of these should be updated.

PR #9201, opened 2025-11-09
Fix mix_init return value

Lauszus:
Currently the return value is not used, so it did not cause any issues

Maintainer merge checklist


kivy/core/audio_output/audio_sdl3.pyx
```diff
@@ -59,7 +59,7 @@ cdef mix_init():
 
     # avoid next call
     if mix_is_init != 0:
-        return
+        return mix_is_init > 0 ? 1 : 0;
```
misl6: This is not a valid Cython syntax unfortunately ☺️

Lauszus: Hehe fixed it now.

misl6 (approved):
LGTM. Thank you!

PR #9203, opened 2025-11-09
Bump `Cython` max version to `3.2.0`

misl6:
Maintainer merge checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [kivy]
 cython_min=0.29.1
-cython_max=3.0.12
+cython_max=3.2.0
 cython_exclude=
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -1,6 +1,6 @@
 [kivy]
 cython_min=0.29.1
-cython_max=3.0.12
+cython_max=3.2.0
 cython_exclude=
 python_versions=3.9 - 3.13
```
kuzeyron, line 5: Shouldn't this be set for 3.14 too?

misl6: Yes, nice catch! Seems we missed to increase it previously. Will do in another PR

kuzeyron (approved):

Issue #9208, opened 2025-11-12
Pytest v9+: Marks applied to fixtures have no effect


cclauss:
Related to:
* #9204

Our CI tests are failing due to a critical change in pytest v9's behavior: you can no longer apply pytest marks (including custom marks like [`@async_run`](https://github.com/search?q=repo%3Akivy%2Fkivy%20async_run)) directly to fixtures. This raises a PytestRemovedIn9Warning and errors out collection of:

[test_app.py](https://github.com/kivy/kivy/blob/6f322647b24e4215cd056cc6d3fe7b90c2f6e29a/kivy/tests/test_app.py) (see lines like 56)
[test_uix_dropdown.py](https://github.com/kivy/kivy/blob/6f322647b24e4215cd056cc6d3fe7b90c2f6e29a/kivy/tests/test_uix_dropdown.py) (see line 42)
[test_uix_modal.py](https://github.com/kivy/kivy/blob/6f322647b24e4215cd056cc6d3fe7b90c2f6e29a/kivy/tests/test_uix_modal.py) (see line 46)
[test_uix_videoplayer.py](https://github.com/kivy/kivy/blob/6f322647b24e4215cd056cc6d3fe7b90c2f6e29a/kivy/tests/test_uix_videoplayer.py) (see lines 19, 36, 51, 69, 85, 96)

See the deprecation notice for details:
https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function

PR #9209, opened 2025-11-12
Temporarily avoid pytest v9

cclauss:
Workaround until compatibility with pytest v9 is fixed.
* #9208
* https://docs.pytest.org/en/stable/changelog.html
```diff
[options.extras_require]
dev =
-   pytest_asyncio!=0.11.0
+   pytest_asyncio==1.2.0  # See: kivy/kivy#9208
```

Maintainer merge checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,9 +29,9 @@ install_requires =
 [options.extras_require]
 tuio = oscpy
 dev =
-    pytest>=3.6,<3.9  # See: kivy/kivy#9208
+    pytest>=3.6
     pytest-cov
-    pytest_asyncio>=1.2.0
+    pytest_asyncio==1.2.0  # See: kivy/kivy#9208
     pytest-timeout
     pytest-benchmark
     pyinstaller


misl6:
I'm looking into it, do you have any other suggestions? (E.g. you have investigated on how to migrate our code to be compatibile with Pytest 9)?

misl6:
Lost some time investigating about it, I'm now getting a:

```
E               pytest.PytestRemovedIn9Warning: 'test_basic_app[kivy_app0]' requested an async fixture 'kivy_app', with no plugin or hook that handled it. This is usually an error, as pytest does not natively support it. This will turn into an error in pytest 9.
E               See: https://docs.pytest.org/en/stable/deprecations.html#sync-test-depending-on-async-fixture
```

So seems pytest is not finding out that the fixture is managed via pytest-async even if it is.

If I move the logic of async_run to the single test / fixture, that error disappears, but instead I get a "AttributeError: 'NoneType' object has no attribute 'text'", and that make me think about if this tests ever worked :/

cclauss:
Sorry.  I got distracted on another project.  I should have time to look into your suggestion tomorrow.  Thanks for digging in.

misl6:
Closing as fixed on master

Issue #9231, opened 2025-12-30
_trigger_update_hex method from ColorPicker raised ValueError when user enter non color hexadecimal characters like ?;


javixav:
**Software Versions**
* Python: 3
* OS: Windows 10
* Kivy: 2.2.1
* Kivy installation method: pip

**Describe the bug**

The Widget ColorPicker from kivy has a ColorPicker_Input widget (TextInput) that allows user to enter a hex color value like "#f8f5f4", but when user writes non hexadecimal characters, it raises a ValueError and stop the app.

**Expected behavior**
When user writes non hexadecimal characters, it should allow or filter non hexadecimal characters

**To Reproduce**
In the ColorPicker TextInput just delete the # and write a character like ; ??&& and even letters [gG-zZ]

**Code and Logs and screenshots**
```python
logs : ValueError: invalid literal for int() with base 16: ';'
```

**A way to resolve this**
Add to you Class ColorPicker.
```
    def _check_text(self, text):
        import re
        regex_pattern = r"^[a-fA-F0-9#]+$"
        if not re.match(regex_pattern, text):
            return
        else:
            self._trigger_update_hex(text)
```
and in the kv add : 
```
ColorPicker_Input:
    size_hint_y: None
    height: '28sp'
    mroot: root
    text: str(root.hex_color)
    on_text: root._check_text(args[1])
```

ElliotGarbus:
Taking a closer look at this widget, there is a similar problem all of the inputs.  There are 2 types of inputs.  A hex input, and an input that takes a value from 0 to 255.  The hex input will throw an exception on invalid inputs, the 0-255 input takes invalid values and writes the log.  In both cases a preferred solution is to create an input_filter for the underlying TextInput and only accept valid values.  This would not allow the user to enter invalid values.

ElliotGarbus:
@javixav I have created a pull request that fixes your issue and overall makes the inputs more robust.  This ended up being a little more complex than I had expected.  The fix improves the usability.

- Add ColorHexInput and ColorNumericInput classes with specialized input filtering
- ColorHexInput: Automatically prepends '#', limits to 8 hex digits, restores valid value on focus loss
- ColorNumericInput: Restricts input to 0-255, prevents leading zeros
- Use Clock.create_trigger for deferred color updates to handle paste operations correctly

PR #9234, opened 2026-01-03
ColorPicker: Make TextInput based inputs more robust

ElliotGarbus:
This PR fixes: #9231 

The changes create two new classes derived from TextInput that filter the input to ensure only valid inputs are accepted.  There are additional changes to ensure that pasting content into the text boxes, or loss of focus update the color.

Maintainer merge checklist

### DIFF
--- a/kivy/uix/colorpicker.py
+++ b/kivy/uix/colorpicker.py
@@ -37,6 +37,7 @@ __all__ = ('ColorPicker', 'ColorWheel')
 
 from math import cos, sin, pi, sqrt, atan
 from colorsys import rgb_to_hsv, hsv_to_rgb
+from string import hexdigits
 
 from kivy.clock import Clock
 from kivy.graphics import Mesh, InstructionGroup, Color
@@ -63,9 +64,7 @@ class ColorHexInput(TextInput):
         if undo:
             return substring
 
-        hex_chars = '0123456789abcdefABCDEF'
         current_text = self.text
-
         # Already at max length (9 = '#' + 8 hex chars)
         if len(current_text) >= 9:
             return ''
@@ -76,7 +75,7 @@ class ColorHexInput(TextInput):
                 # Only allow '#' at the beginning
                 if not current_text and not result:
                     result = char
-            elif char in hex_chars:
+            elif char in hexdigits:
                 # Auto-add '#' if first character is a hex digit
                 if not current_text and not result:
                     result = '#'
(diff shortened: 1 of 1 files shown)


kivy/uix/colorpicker.py
```diff
@@ -431,15 +536,26 @@ def _update_clr(self, dt):
         # to prevent interaction between hsv/rgba, we work internally using rgba
         mode, clr_idx, text = self._upd_clr_list
         try:
-            text = min(255.0, max(0.0, float(text)))
+            # Treat empty string as 0
+            if not text or text.strip() == '':
```
kuzeyron: I think `if not text.strip():` would be enough. Unless text is not a string..

kivy/uix/colorpicker.py
```diff
@@ -46,10 +46,110 @@ def on_color(instance, value):
...
+
+        # Check for leading zeros in the pasted content
+        # '01', '001', etc. should be rejected
+        if len(result) > 1 and result[0] == '0':
+            return ''
+
+        # Check if current text is '0' and we're trying to add more digits
+        # This would create a leading zero: '0' + '5' = '05'
+        if current_text == '0' and result:
+            return ''
```
kuzeyron: Personally I would put these together with the help of any()

kivy/uix/colorpicker.py
```diff
@@ -46,10 +46,110 @@ def on_color(instance, value):
...
+    # Leading zeros are rejected.
+
+    def _filter_numeric(self, substring, undo):
+        """Filter input to allow only integers 0-255."""
+        # Allow undo operations to proceed without filtering
+        if undo:
+            return substring
+
+        current_text = self.text
+        result = ''.join(c for c in substring if c in '0123456789')
```
kuzeyron: With string there is a way to check for digits. `str.isdigit()`

kivy/uix/colorpicker.py
ElliotGarbus: Good catch!  That was a left over that I fixed elsewhere.

kivy/uix/colorpicker.py
ElliotGarbus: got it

kivy/uix/colorpicker.py
ElliotGarbus: I actually simplified things a little more... I was thinking about how to handle a paste of digits that included invalid chars, and decided it is best to not allow it, rather than filter out the bad chars.

ElliotGarbus:
@kuzeyron Thanks for the quick feedback.  I have implemented the recommended changes.  I also made a few additional enhancements:
- Use string.hexdigits constant instead of hardcoded hex characters
- Added additional leading zero case in ColorNumericInput:
  * Prevent inserting "0" at cursor position 0 when text exists
 - Simplified pasting logic
 - Removed try except block

kuzeyron (approved):

kuzeyron:
LGTM!

PR #9245, opened 2026-01-21
Fix pytest 9 compatibility

mgrossu:
Fix pytest 9 compatibility when building for openSUSE by adding asyncio_mode = "auto" to pyproject.toml and updating the common test, because pytest 9 changed its default asyncio handling which caused tests to fail


misl6:
Hi @mgrossu , this seems to not have fixed our issue with pytest 9 (see tests runs), do you have any other suggestions?

misl6:
Closing as already fixed on master

Issue #8571, opened 2024-01-17
Provide (internally or a via an external dependency) a module that provides OS-specific folders


misl6:
A PR (https://github.com/kivy/kivy/pull/8153) tried to fix our data folder destination, which looks wrong on some platforms, but has been kept unmerged as was breaking the API.

The idea is to offer a module (and later use it internally) for our users to access OS-specific folders (User data, User config, User downloads, ...)

- Is **platformdirs** project (https://github.com/platformdirs/platformdirs) enough?
- If not, better to integrate the features in `platformdirs` or to write our version of the tool?

ElliotGarbus:
I have not used platformdirs but it looks very good. 
On Windows I've been using https://gist.github.com/mkropat/7550097
and then creating my own constructs.  plafromdirs looks much better.

Here what i have been adding to the App class to support Windows and MacOS:
```python
    def get_application_config(self, defaultpath='%(appdir)s/%(appname)s.ini'):
        if platform == 'macosx':  # mac will not write into app folder
            s = '~/.%(appname)s.ini'
        else:
            s = defaultpath
        return super().get_application_config(defaultpath=s)

    @property
    def download_directory(self):
        if platform == 'win':
            return get_path(FOLDERID.Downloads)
        elif platform == 'macosx':
            return str(Path().home() / 'Downloads')
        else:
            raise ValueError(f'Platform {platform} not defined')

    @property
    def library_directory(self):  # This is a user data dir, user data was called the library in the app
        if platform == 'win':
            doc_root = get_path(FOLDERID.Documents)
            doc_path = Path(doc_root) / 'Customer Name Here' / 'App Name Here'
            doc_path.mkdir(parents=True, exist_ok=True)
            return str(doc_path)
        elif platform == 'macosx':
            doc_path = Path().home() / 'Documents' / 'Customer Name Here' / 'App Name Here'
            doc_path.mkdir(parents=True, exist_ok=True)
            return str(doc_path)
        else:
            raise ValueError(f'Platform {platform} not defined')
```

ElliotGarbus:
OBSOLETE:  I'm changing direction based on feedback, and experiments.  See: https://github.com/kivy/kivy/issues/8571#issuecomment-3814713520

Here is a proposal for adding this feature:

## Problem Statement

[...]

PR #9246, opened 2026-01-25
update user_data_dir, create user_cache_dir, Move Kivy configuration to app-specific directories for desktop OS, 

ElliotGarbus:
This PR is a response to #8571 
The scope of the changes:
The `user_data_dir` is update for Linux, correcting the path.
The `user_cache_data_dir` is created for all platforms.
On Desktop OS, the .kivy directory is now in a app specific path, rather that the global ` ~/.kivy/`

Kivy now isolates configuration, logs, and user modules per application instead of sharing ~/.kivy/ globally. This improves security, enables per-app settings, and follows platform conventions for data storage.
Note that while in development, if under a venv, the `.kivy' dir will be located in the main project dir.

Details:
Config and Clock initialization is deferred until App.__init__() to access the app name for directory paths. A lazy proxy pattern with callbacks ensures backward compatibility for modules accessing Config at import time.

Key changes:
- ConfigProxy and ClockProxy defer initialization until App.__init__()
- Config.on_config_ready() callback mechanism for early module access
- App-specific paths: user_data_dir/<AppName>/.kivy/config.ini
- New App.user_cache_dir property for temporary/cache data
- Fixed Linux XDG compliance (use XDG_DATA_HOME, not XDG_CONFIG_HOME)
- Default logger level set to INFO before Config loads

Breaking changes:
- Desktop apps: config.ini moves from ~/.kivy/ to app-specific location
- No automatic migration; users can set KIVY_HOME for compatibility
- Modules importing Config at top-level must use callbacks to know when the config is initialized

Migration guide updated with detailed instructions for users and framework developers.

Maintainer merge checklist

### DIFF
--- a/doc/sources/migration.rst
+++ b/doc/sources/migration.rst
@@ -8,6 +8,94 @@ Introduction
 
 Kivy 3.x.x introduces several changes and improvements compared to Kivy 2.x.x. This guide will help you migrate your existing Kivy 2.x.x codebase to Kivy 3.x.x.
 
+Configuration System Changes
+-----------------------------
+
+*App-Specific Configuration Directories*
+
+In Kivy 3.x.x, the configuration system has been redesigned to use app-specific directories for desktop platforms, 
+similar to mobile platforms. This ensures better isolation between applications and follows platform conventions more closely.
+
+**Key Changes:**
+
+1. **Config File Location**: The ``config.ini`` file is now located in ``user_data_dir/.kivy/config.ini`` instead of a global ``~/.kivy/config.ini``.
+
+2. **App-Specific Paths**: ``user_data_dir`` now returns an app-specific path:
+   
+   - Windows: ``%APPDATA%\<appname>``
+   - macOS: ``~/Library/Application Support/<appname>``
+   - Linux: ``~/.local/share/<appname>`` (XDG compliant)
+   - iOS: ``~/Documents/<appname>`` (unchanged)
+   - Android: App-specific data directory (unchanged)
+
+3. **Cache Directory Support**: A new ``user_cache_dir`` property is available on ``App``:
+   
+   - Windows: ``%LOCALAPPDATA%\<appname>\Cache``
+   - macOS: ``~/Library/Caches/<appname>``
+   - Linux: ``~/.cache/<appname>`` (XDG compliant)
+   - iOS: ``~/Library/Caches/<appname>``
+   - Android: App-specific cache directory
+
+4. **Deferred Config Loading**: Config is now loaded during ``App.__init__()`` instead of at module import time. This ensures the config file is loaded from the correct app-specific location.
+
+5. **KIVY_HOME Override**: The ``KIVY_HOME`` environment variable can still be used to override the default paths for backward compatibility.
+
+**Migration Steps:**
+
+For most applications, no changes are required. The config file will be created in the new location automatically on first run.
+
+If you need to migrate existing configuration files:
+
+.. code-block:: python
+
+    import os
+    import shutil
+    from kivy.app import App
+    
+    class MyApp(App):
+        def __init__(self, **kwargs):
+            super().__init__(**kwargs)
+            
+            # Optional: Migrate old config to new location
+            old_config = os.path.expanduser('~/.kivy/config.ini')
+            new_config = os.path.join(self.user_data_dir, '.kivy', 'config.ini')
+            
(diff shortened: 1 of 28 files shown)


ElliotGarbus:
A few notes for reviewers.  Config and Clock have side effects on import.  In order to use the name of the app as part of the path, I created proxy's for these objects.  The Config object has another complexity in that a number of files pull parameters out of the Config on startup.  To support this behavior I added a callback to the config, that calls the registered callback method after the Config has been loaded.

ElliotGarbus:
@misl6 I have a strange issue here.  I'm showing a PEP8 failure on setup.py, but I have not modified setup.py.  Let me know if you want me to do the pep8 fix.

ElliotGarbus:
Most of the code in this PR is the create a path that includes the App.name for desktop apps.  This allows putting the `.kivy` directory in the correct location.  This is not an issue for mobile apps.

An alternative approach is to set  KIVY_HOME prior to importing kivy.  I have an example of this in the code below. The downsize of alt approach is the need to provide the name of the app prior to importing kivy.  The positive is the simplicity.  Open to feedback.

```python
from os import environ
import sys
from sys import platform as _sys_platform
from pathlib import Path

def _get_platform():  # from kivy.utils
    # existence of environ variables set during Python initialization
    kivy_build = environ.get('KIVY_BUILD', '')
    if kivy_build in {'android', 'ios'}:
        return kivy_build
    elif 'P4A_BOOTSTRAP' in environ:
        return 'android'
    elif 'ANDROID_ARGUMENT' in environ:
        # We used to use this method to detect android platform,
        # leaving it here to be backwards compatible with `pydroid3`
        # and similar tools outside kivy's ecosystem
        return 'android'
    elif _sys_platform in ('win32', 'cygwin'):
        return 'win'
    elif _sys_platform == 'darwin':
        return 'macosx'
    elif _sys_platform.startswith('linux'):
        return 'linux'
    elif _sys_platform.startswith('freebsd'):
        return 'linux'
    return 'unknown'

def set_app_name(app_name):
    """
    Set the app name for desktop platforms
    This is used to set the correct path for KIVY_HOME
    setting the location of the .kivy dir where config.ini and the
    log dir, module dir, and icon dir are located.
    """
    platform = _get_platform()
    if environ.get('KIVY_HOME', None) or platform in ('ios', 'android'):
        return
    if 'kivy.config' in sys.modules:
        raise RuntimeError(
            "Config has already been imported. "
            "set_app_name must execute before kivy.config to ensure "
            "proper initialization order."
        )
    name = app_name.lower()
    name = name[:-3] if name.endswith('app') else name
    if platform == 'win':
        kivy_home = Path(environ['APPDATA']) / f'{name}/.kivy'
    elif platform == 'macosx':
        kivy_home = Path(f'~/Library/Application Support/{name}/.kivy')
    elif platform == 'linux':
        p = Path(environ.get('XDG_DATA_HOME', '~/.local/share'))
        kivy_home = p / f'{name}/.kivy'
    else:
        raise ValueError(f'Unknown platform: {platform}')
    kivy_home = kivy_home.expanduser()
    kivy_home.mkdir(parents=True, exist_ok=True)
    environ['KIVY_HOME'] = str(kivy_home)
```
And an example using the file above:
```python
from home_path import set_app_name # must import prior to kivy
set_app_name('EnvPathCheckApp')

import os
from kivy.app import App
from kivy.lang import Builder


kv = """
AnchorLayout:
    Button:
        text: 'user_data_dir'
        size_hint: None, None
        size: '200dp', '48dp'
        on_release: print(app.user_data_dir)
"""

class EnvPathCheckApp(App):
    def build(self):

    def on_start(self):
        print(f'{self.user_data_dir=}')
        print(f'{os.environ['KIVY_HOME']=}')

EnvPathCheckApp().run()
```
Output:
```
self.user_data_dir='C:\\Users\\ellio\\AppData\\Roaming\\envpathcheck'
os.environ['KIVY_HOME']='C:\\Users\\ellio\\AppData\\Roaming\\envpathcheck\\.kivy'
```

ElliotGarbus:
The approach of creating proxy's for Config, Clock, Window just added to much complexity.  I'm closing this PR.
I will split this into two PRs in the near future

1. update user_data_dir and create user_cache_dir
2. create a utility to set the KIVY_HOME env var with the proper path. https://github.com/kivy/kivy/pull/9246#issuecomment-3802223879

PR #9248, opened 2026-01-25
Feature: Hover manager and hover behavior

### LINKED ISSUE #9236: Re-integrate `hover` package into Kivy
As per discussion in https://github.com/kivy/kivy/pull/9187#issuecomment-3539695729, this is my plan to re-integrate my `hover` package into Kivy framework:

1. Revert @DexerBR commits made by [the request 9187](https://github.com/kivy/kivy/pull/9187)
2. Remove `"Experimental"` notice from `kivy.eventmanager` module, and from `WindowBase`, `Widget` and `MotionEvent` classes which was introduced in [request 7658](https://github.com/kivy/kivy/pull/7658)
3. Integrate my `hover` package https://github.com/pythonic64/hover into Kivy

I've added the module level [documentation](https://github.com/pythonic64/hover/pull/5/files) to hover package and made some minor code fixes, all marked as part of [0.2.0 release](https://github.com/pythonic64/hover/milestone/1?closed=1). Hover package itself will be moved into Kivy-Garden and remain available to for `Kivy>=2.1.0`.

pythonic64:
This request re-introduces hover manager and hover behavior into Kivy framework. 

Classes included in the request are taken from my https://github.com/pythonic64/hover package and adjusted for Kivy. They are the following:

1. `HoverManager` in `kivy.eventmanagers.hover` module.
2. `HoverBehavior` in `kivy.uix.behaviors.hover` module.
3. `HoverCollideBehavior` in `kivy.uix.behaviors.hover` module.

Class `MotionCollideBehavior` is renamed to `HoverCollideBehavior` and its code is changed to do collision check only for hover events. Previous behavior was not generally correct, because an event with any `etype` or any `type_id` would be checked for collision before being dispatched to the child widgets.

An example showcasing how manager and behavior work is available at `examples/widgets/hover_behavior.py`.

Maintainer merge checklist:


abedhammoud:
I have used this this hover library for couple of years, its well written and has added a much needed functionality to my desktop app.

kivy/uix/behaviors/hover.py
```diff
@@ -0,0 +1,261 @@
...
+        :meth:`collide_point` and dispatch `on_hover_enter`, `on_hover_update`
+        or `on_hover_leave` events.
+
+        :Parameters:
+            `etype`: `str`
+                Event type, one of "begin", "update" or "end"
+            `me`: :class:`~kivy.input.motionevent.MotionEvent`
+                Hover motion event
+        """
+        if etype == 'update' or etype == 'begin':
```
misl6, line 211:
```suggestion
        if etype in {'update' , 'begin'}:
```

kivy/uix/behaviors/hover.py
```diff
@@ -0,0 +1,261 @@
...
+            return super().on_motion(etype, me)
+        if self.hover_mode == 'default':
+            if super().on_motion(etype, me):
+                return True
+            return self.dispatch('on_hover_event', etype, me)
+        prev_mode = me.dispatch_mode
+        if self.hover_mode == 'self':
+            me.dispatch_mode = MODE_DONT_DISPATCH
+        accepted = super().on_motion(etype, me)
+        accepted = self.dispatch('on_hover_event', etype, me) or accepted
```
misl6, line 194:
```suggestion
        accepted = accepted or self.dispatch('on_hover_event', etype, me)
```

kivy/uix/behaviors/hover.py
```diff
@@ -0,0 +1,261 @@
...
+HoverBehavior
+=============
+
+.. versionadded:: 3.0.0
+
+.. note::
+    :class:`HoverBehavior` is part of the `hover` package which can be found in
+    the GitHub `repository <https://github.com/pythonic64/hover>`_. You can
+    install the `hover` package using the instructions from the GitHub page, if
+    you want to use `Kivy>=2.1.0,<3.0.0`.
```
misl6, line 11: “Is part” might be misleading. Please also keep in mind that once Kivy 3.0.0 is released, we will no longer provide support for versions below 3.0.0. Therefore, it’s best to encourage users to migrate to Kivy 3.0.0 as soon as possible.

kivy/eventmanager/hover.py
```diff
@@ -0,0 +1,228 @@
...
+        if self._clock_event:
+            self._clock_event.cancel()
+            self._clock_event = None
+
+    def dispatch(self, etype, me):
+        me_grab_list = me.grab_list[:]
+        del me.grab_list[:]
+        accepted = self._dispatch_to_widgets(etype, me)
+        self._events[me.uid].insert(0, (me, me.grab_list[:]))
+        self._event_times[me.uid] = Clock.get_time() if Clock else 0
```
misl6, line 150: Just curious, why we can't assume that Clock is available?

kivy/eventmanager/hover.py
```diff
@@ -0,0 +1,228 @@
...
+        )
+        self._events = defaultdict(list)  # me.uid -> [(me, me.grab_list[:]),]
+        self._event_times = {}  # me.uid -> Clock.get_time()
+        self._clock_event = None
+
+    def start(self):
+        if self.event_repeat_timeout >= 0:
+            global Clock
+            if not Clock:
+                from kivy.clock import Clock
```
misl6, line 128: Why not importing Clock on top?

misl6 (commented):
I’ve left a few minor suggestions, please let me know what you think.
It’s not mandatory, and we can address it in a later PR, but what about adding some tests?

kivy/uix/behaviors/hover.py
pythonic64: Are you sure that creating a set just to check for `etype` value is faster then checking individually for both cases?
To add, `etype` "update" is most likely to happen for hover events.

kivy/uix/behaviors/hover.py
pythonic64: I think that you misread this line of code, because in this case we want both `super().on_motion(etype, me)` and `self.dispatch('on_hover_event', etype, me)` functions to be called and then return `true` if any of those functions returned `true`.

kivy/eventmanager/hover.py
pythonic64: Singleton `Clock` is an optional dependency for the hover manager and if `event_repeat_timeout` is lesser that `0`, then `Clock` will not be used.

kivy/uix/behaviors/hover.py
pythonic64: Can you suggest a better wording for this case? Maybe "... is also a part of the `hover` package..."?

I want to developers to be aware that they don't have to go through upgrading to Kivy `3.x.x`, which is a breaking change upgrade, if they just want to use `hover` package with Kivy `2.x.x`.

kivy/eventmanager/hover.py
pythonic64: Singleton `Clock` is an optional dependency for the `HoverManager` and importing a singleton on top which may not be used is not a good practice. Moreover:

1. Importing `Clock` slows down the import of the `HoverManager` as `kivy.clock` as to import singleton `Config`, other classes and then create a clock instance.
2. Using late imports for optional dependencies is already a practice in the Kivy framework. Some location are: https://github.com/kivy/kivy/blob/2.3.1/kivy/input/providers/mouse.py#L126, http://github.com/kivy/kivy/blob/2.3.1/kivy/core/window/__init__.py#L638, https://github.com/kivy/kivy/blob/2.3.1/kivy/uix/textinput.py#L2044

pythonic64:
@misl6 I won't have the time in the foreseeable future to write the tests, but I can find some time to review the pull request if someone else is willing to write them.

That being said, I can confirm that the feature is working correctly for one large desktop app, and a few other developers reported no issues when using the feature in their projects.

abedhammoud:
I confirm I have been using this feature for nearly two years and it works flawlessly. Thank you.

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

kivy/uix/behaviors/hover.py
```diff
@@ -0,0 +1,261 @@
...
+:class:`HoverCollideBehavior` is meant to be used with
+:class:`~kivy.uix.stencilview.StencilView` or its subclasses so that hover
+events (events with :attr:`~kivy.input.motionevent.MotionEvent.type_id` set to
+"hover") don't get handled when their position is outside the view's bounding
+box.
+
+Example of using :class:`HoverCollideBehavior` with
+:class:`~kivy.uix.recycleview.RecycleView`::
+
+    from kivy.uix.behaviors import HoverCollideBehavior
```
gottadiveintopython, line 102: Wouldn't this line be `from kivy.uix.behaviors.hover import HoverCollideBehavior` ?

pythonic64: That line works because class `HoverCollideBehavior` is imported in the `kivy/uix/behaviors/__init__.py` file.

https://github.com/kivy/kivy/pull/9248/changes#diff-27a3947581670e65212f248246da70dfc24fb9051fe687fb6448ee21843fa3bdR104

gottadiveintopython: Sorry, I didn't notice that. Thanks.

pythonic64: Never mind. Thank you for the review.

kivy/eventmanager/hover.py
```diff
@@ -0,0 +1,228 @@
...
+    event_repeat_timeout = 1 / 30.0
+    """Minimum wait time to repeat existing static hover events and it defaults
+    to `1/30.0` seconds. Negative value will turn off the feature.
+
+    To change the default value use `event_repeat_timeout` keyword while making
+    a manager instance or set it directly after the instance is made. Changing
+    the value after the manager has started will have no effect.
+    """
+
+    def __init__(self, **kwargs):
```
gottadiveintopython: Is there a specific reason to accept arbitrary keyword arguments?
Wouldn't it be clearer to define the keyword arguments explicitly?

```python
    def __init__(self, *, event_repeat_timeout=...):
```

pythonic64: You're right. We can use `event_repeat_timeout` as an argument with a default value of `1/30.0`.
```python
  def __init__(self, event_repeat_timeout=1/30.0):
```

Use of `**kwargs` is not needed by the manager class nor by its super class.

gottadiveintopython:
If no one else is willing to, I'll add the unit tests. Would you prefer that I add them to this PR or open a separate PR?

pythonic64:
@gottadiveintopython Can you make a separate pull request and tag me on it once done?

pythonic64:
@misl6 Can you reply on comment threads above?

misl6 (approved):
LGTM. Thank you!

Sorry for the late reply, and thank you @gottadiveintopython for providing the tests!

PR #9255, opened 2026-01-29
Add `user_cache_dir` property and fix Linux `user_data_dir`

ElliotGarbus:
Resolves Issue #8571 

This PR adds features for improving application directory management:

For desktop and mobile platforms user_data_dir and user_cache_dir are now supported by kivy on all platfroms.
On desktop platforms a new environment variable KIVY_DESKTOP_PATH_ID sets user-facing, human-readable application titles for directories

Details:

1. Add App.user_cache_dir Property
   - Provides standard location for temporary/cache data
   - Follows platform conventions (XDG on Linux, Caches on macOS/iOS)
   - System may purge this data at any time
   - Platform paths:
     * Windows: %APPDATA%\<AppName>\Cache
     * macOS: ~/Library/Caches/<AppName>
     * Linux: ~/.cache/<AppName> (respects $XDG_CACHE_HOME)
     * Android: Context.getCacheDir()
     * iOS: ~/Library/Caches/<AppName>

2. Fix Linux user_data_dir XDG Compliance
   - Corrected to use XDG_DATA_HOME instead of XDG_CONFIG_HOME
   - Now follows XDG Base Directory specification
   - Uses ~/.local/share/<AppName> by default

3. Introduce KIVY_DESKTOP_PATH_ID Environment Variable
   - Sets user-facing, human-readable application titles for directories
   - Makes it easy for end users to identify app directories when browsing
   - Primary use case: Show "My Photo Editor" instead of "photoeditorapp"
   - Applies to KIVY_HOME, user_data_dir, and user_cache_dir on desktop
   - Highest priority (overrides KIVY_HOME and venv detection)
   - Mobile platforms (iOS/Android) unaffected - use App.name as before
   - Platform paths with KIVY_DESKTOP_PATH_ID='My Photo Editor':
     * Windows: %APPDATA%\My_Photo_Editor\.kivy
     * macOS: ~/Library/Application Support/My_Photo_Editor/.kivy
     * Linux: ~/.local/share/My_Photo_Editor/.kivy

- Documentation updates
  * Add "Application Storage Directories" section to App docstring
  * Recommend platformdirs for additional desktop OS-specific directories
  * Document KIVY_DESKTOP_PATH_ID in environment.rst
  * Update app-specific configuration guidance in config.rst

- Add examples/desktop_path_id/ directory
  * set_desktop_path_id.py: Utility function with validation
  * demo_app.py: Working demonstration
  * README.txt: Complete usage documentation

Breaking Changes: None
- Existing apps continue to work unchanged
- KIVY_HOME environment variable still respected
- Virtual environment detection still works
- New features are opt-in via KIVY_DESKTOP_PATH_ID

Maintainer merge checklist

### DIFF
--- a/examples/desktop_path_id/demo_app.py
+++ b/examples/desktop_path_id/demo_app.py
@@ -10,6 +10,7 @@ import os
 os.environ['KIVY_DESKTOP_PATH_ID'] = 'My Photo Editor'
 
 from textwrap import dedent
+from kivy import kivy_home_dir
 from kivy.app import App
 from kivy.lang import Builder
 from kivy.properties import StringProperty
@@ -54,7 +55,7 @@ class DesktopPathIdDemoApp(App):
         path_id = os.environ.get('KIVY_DESKTOP_PATH_ID', 'Not set')
 
         # KIVY_HOME is constructed as user_data_dir/.kivy
-        kivy_home = os.path.join(self.user_data_dir, '.kivy')
+        kivy_home = kivy_home_dir
         config_path = os.path.join(kivy_home, 'config.ini')
         logs_path = os.path.join(kivy_home, 'logs')
 
@@ -85,7 +86,7 @@ class DesktopPathIdDemoApp(App):
             """).strip()
 
     def print_paths(self):
-        kivy_home = os.path.join(self.user_data_dir, '.kivy')
+        kivy_home = kivy_home_dir
         config_file = os.path.join(kivy_home, 'config.ini')
         logs_directory = os.path.join(kivy_home, 'logs')
 


kivy/utils.py
```diff
@@ -41,6 +41,66 @@ def difference(set1, set2):
...
+
+    .. versionadded:: 3.0.0
+    '''
+    if not path_id or not path_id.strip():
+        raise ValueError("path_id cannot be empty")
+
+    # Remove control characters (0x00-0x1F, 0x7F-0x9F)
+    safe_id = sub(r'[\x00-\x1F\x7F-\x9F]', '', path_id)
+    # Replace invalid filesystem chars with underscore
+    safe_id = sub(r'[/\\:*?"<>|]', '_', safe_id)
```
kuzeyron, line 103: What do you think if it didn't use underscore and is just putting them together so they become camel-cased(CompanyAppName)?

ElliotGarbus: There are a number of languages that do not have a concept of upper-case.  Languages like Chinese, Thai, Korean, Arabic, Hebrew... If a developer wants to use CapCase, it will be preserved by the normalizer.

ElliotGarbus: It does look like I could make the normalizer more robust by also limiting string length, and normalizing Unicode.  I'll do some more research and update later today.  

The key benefit of KIVY_DESKTOP_PATH_ID is that if puts the .kivy dir in an app specific directory, rather that the global `~/.kivy`, I'll update the docs to make this clear.

ElliotGarbus: PR Updated

kivy/__init__.py
```diff
@@ -351,7 +351,20 @@ def _patch_mod_deps_win(dep_mod, mod_name):
...
+        from kivy.utils import normalize_path_id
+        path_id = normalize_path_id(environ['KIVY_DESKTOP_PATH_ID'])
+
+        if platform == 'win':
+            base_dir = environ['APPDATA']
+        elif platform == 'macosx':
+            base_dir = expanduser('~/Library/Application Support')
+        else:  # linux
+            base_dir = expanduser(environ.get('XDG_DATA_HOME', '~/.local/share'))
+
```
kuzeyron: Venv is important having because when we are trying out PR's or the setup is used with venv. I don't want this to bloat my system with configurations just because I had to try a PR.

ElliotGarbus: The support for venv is still there, it is just lower priority.  My thinking is if you have specified this path element it should be used in development.  Are you recommending a change?

```py
    # Configuration management
    # Check for KIVY_DESKTOP_PATH_ID first on desktop platforms
    if platform in ('win', 'macosx', 'linux') and 'KIVY_DESKTOP_PATH_ID' in environ:
        from kivy.utils import normalize_path_id
        path_id = normalize_path_id(environ['KIVY_DESKTOP_PATH_ID'])

        if platform == 'win':
            base_dir = environ['APPDATA']
        elif platform == 'macosx':
            base_dir = expanduser('~/Library/Application Support')
        else:  # linux
            base_dir = expanduser(environ.get('XDG_DATA_HOME', '~/.local/share'))

        kivy_home_dir = join(base_dir, path_id, '.kivy')
    elif 'KIVY_HOME' in environ:
    elif platform == 'android':
    elif platform == 'ios':
        user_home_dir = join(user_home_dir, 'Documents')
    elif sys.prefix != sys.base_prefix:
        # Detection if venv being used with the framework
        user_home_dir = dirname(sys.prefix)
```

kuzeyron: Ah, it wasn't visible to me in the reviewer mode.
`if platform in ('win', 'macosx', 'linux') and 'KIVY_DESKTOP_PATH_ID' in environ:` can you change it to a set instead? That's what most linters prefers (`if platform in {'win', 'macosx', 'linux'} and 'KIVY_DESKTOP_PATH_ID' in environ:`)

ElliotGarbus: Of course!  I should get to this later today.

ElliotGarbus: Changes committed.

kivy/app.py
```diff
@@ -852,27 +908,64 @@ def directory(self):
 
     def _get_user_data_dir(self):
         # Determine and return the user_data_dir.
-        data_dir = ""
+        # Check for KIVY_DESKTOP_PATH_ID on desktop platforms
+        desktop_path_id = os.environ.get('KIVY_DESKTOP_PATH_ID')
+        if platform in ('win', 'macosx', 'linux') and desktop_path_id:
```
kuzeyron: Also this one

kivy/utils.py
```diff
@@ -11,291 +11,417 @@
...
     if isinstance(s, (list, tuple)):
-        s = [x / 255. for x in s]
+        s = [x / 255.0 for x in s]
         if len(s) == 3:
             s.append(1)
         return s
     if isinstance(s, (int, float)):
-        s = [s / 255.]
-        s.extend(x / 255. for x in args)
+        s = [s / 255.0]
```
kuzeyron: These changes aren't required as `.0 and 0.` just works.

kivy/utils.py
```diff
@@ -11,291 +11,417 @@
...
         return [random(), random(), random(), alpha]
 
 
 def is_color_transparent(c):
-    '''Return True if the alpha channel is 0.'''
+    """Return True if the alpha channel is 0."""
     if len(c) < 4:
         return False
-    if float(c[3]) == 0.:
+    if float(c[3]) == 0.0:
```
kuzeyron: Same here

kivy/utils.py
```diff
@@ -11,291 +11,417 @@
...
-    'thistle': '#d8bfd8',
-    'tomato': '#ff6347',
-    'turquoise': '#40e0d0',
-    'violet': '#ee82ee',
-    'wheat': '#f5deb3',
-    'white': '#ffffff',
-    'whitesmoke': '#f5f5f5',
-    'yellow': '#ffff00',
-    'yellowgreen': '#9acd32',
+    "aliceblue": "#f0f8ff",
```
kuzeyron: This is an overkill.

kivy/utils.py
```diff
@@ -429,29 +563,29 @@ def format_bytes_to_human(size, precision=2):
 def _get_platform():
     # On Android sys.platform returns 'linux2', so prefer to check the
     # existence of environ variables set during Python initialization
-    kivy_build = environ.get('KIVY_BUILD', '')
-    if kivy_build in {'android', 'ios'}:
+    kivy_build = environ.get("KIVY_BUILD", "")
```
kuzeyron: Same here..

kivy/utils.py
ElliotGarbus: It looks like inadvertently ran ruff with the format option... and it made these changes... I'll revert and fix.  A bit frustrating.  I'll get this cleaned up later today.

FWIW I'm running ruff check to check for pep8 compliance.  I believe I'm using it correctly and as configured for kivy.  It is not flagging the tuple vs set use.

kivy/__init__.py
ElliotGarbus: @kuzeyron I've made the changes to use a set rather than a tuple in the code that I changed.  FWIW, there are 17 instances of using in with a tuple in `kivy.__init__.py' in code that I have not touched.

ElliotGarbus:
All comments addressed with last update.

kivy/app.py
```diff
@@ -852,27 +908,64 @@ def directory(self):
...
         else:  # _platform == 'linux' or anything else...:
-            data_dir = os.environ.get('XDG_CONFIG_HOME', '~/.config')
-            data_dir = expanduser(join(data_dir, self.name))
+            data_dir = join(os.environ.get('XDG_DATA_HOME', '~/.local/share'),
+                           app_id)
+
+        data_dir = expanduser(data_dir)
         if not exists(data_dir):
-            os.mkdir(data_dir)
+            os.makedirs(data_dir, exist_ok=True)
```
kuzeyron: The if statement is useless because exist_ok

kivy/__init__.py
```diff
@@ -361,16 +373,27 @@ def _patch_mod_deps_win(dep_mod, mod_name):
...
+            'See https://kivy.org/doc/stable/guide/environment.html'
+        )
+
     kivy_home_dir = kivy_home_dir or join(user_home_dir, '.kivy')
     kivy_config_fn = join(kivy_home_dir, 'config.ini')
     kivy_usermodules_dir = join(kivy_home_dir, 'mods')
     icon_dir = join(kivy_home_dir, 'icon')
 
     if 'KIVY_NO_CONFIG' not in environ:
         if not exists(kivy_home_dir):
```
kuzeyron: Same here

kivy/app.py
```diff
@@ -852,27 +908,64 @@ def directory(self):
...
+            cache_dir = join(os.environ['LOCALAPPDATA'], app_id, 'Cache')
+        elif platform == 'macosx':
+            cache_dir = f'~/Library/Caches/{app_id}'
+        else:  # _platform == 'linux' or anything else...:
+            cache_dir = join(os.environ.get('XDG_CACHE_HOME', '~/.cache'),
+                            app_id)
+
+        cache_dir = expanduser(cache_dir)
+        if not exists(cache_dir):
+            os.makedirs(cache_dir, exist_ok=True)
```
kuzeyron: Same here

kuzeyron:
If you strip away `if not exists(cache_dir):` and only leave ` os.makedirs(cache_dir, exist_ok=True)` this PR is ready for testing.

kivy/app.py
ElliotGarbus: Thanks for the review.

ElliotGarbus:
> If you strip away `if not exists(cache_dir):` and only leave ` os.makedirs(cache_dir, exist_ok=True)` this PR is ready for testing.

Done!  Thanks again for the review.

kuzeyron:
Okay, just tried this PR with the example code of yours. Commenting out `os.environ['KIVY_DESKTOP_PATH_ID'] = ...` doesn't place the kivy configuration where the venv are. The fail-safe part works but it should only happen if it fails to parse `KIVY_DESKTOP_PATH_ID`. I wasn't clear enough earlier with that point, sorry about that.

ElliotGarbus:
> Okay, just tried this PR with the example code of yours. Commenting out `os.environ['KIVY_DESKTOP_PATH_ID'] = ...` doesn't place the kivy configuration where the venv are. The fail-safe part works but it should only happen if it fails to parse `KIVY_DESKTOP_PATH_ID`. I wasn't clear enough earlier with that point, sorry about that.

The code does place the kivy config in the project directory if `KIVY_DESKTOP_PATH_ID`. is not set.  There was an issue with the example not properly reporting the path when the `KIVY_DESKTOP_PATH_ID` was not set.  This has been fixed.

kuzeyron:
Any way for the config not to create empty folders? It's creating the folder `/home/kuzeyron/.local/share/desktoppathiddemo/` and the folder is empty when using venv.

ElliotGarbus:
> Any way for the config not to create empty folders? It's creating the folder `/home/kuzeyron/.local/share/desktoppathiddemo/` and the folder is empty when using venv.

I'm away from my development machine right now, but I will confirm later today.  I suspect that directory is being created when the test program accesses `App.user_data_dir`, and it is not being created for the config.

ElliotGarbus:
@kuzeyron I've confirmed the empty directory you are seeing is from accessing `App.user_data_dir`. I could enhance the example to have it remove the created files and directories on exit, or add a "clean up" button that removes them, or just leave it as it is.

What do you think?

kuzeyron:
I'll get back to you tomorrow. I need to fully focus on your changes

ElliotGarbus:
@kuzeyron FWIW I did some experiments where I delete the directories created on exit.  The log files on Windows are not freed up until the processes closed.  I may just add a comment that it is safe to manually delete these directories.

kivy/app.py
```diff
@@ -852,27 +908,62 @@ def directory(self):
...
+            cache_dir = join(os.environ['LOCALAPPDATA'], app_id, 'Cache')
+        elif platform == 'macosx':
+            cache_dir = f'~/Library/Caches/{app_id}'
+        else:  # _platform == 'linux' or anything else...:
+            cache_dir = join(os.environ.get('XDG_CACHE_HOME', '~/.cache'),
+                            app_id)
+
+        cache_dir = expanduser(cache_dir)
+        os.makedirs(cache_dir, exist_ok=True)
+        return cache_dir
```
kuzeyron, line 965: I guess we could use the same checks here if it's venv and so on

ElliotGarbus: That is a change in behavior.  `App.user_data_dir` is already present in Kivy version 1.7 and has this behavior.  
It is not intuitive to request a directory have have it default to a different directory.  

A possible option would be to create a new environment var, perhaps `KIVY_VENV_DIRS` that would put the .kivy dir, user_*_dir under a a project directory, `kivy_dev`.  Something like:
```
MyProject
    /kivy_dev
        /.kivy
        /user_data
        /user_cache
```
The default behavior for App.user_*_dir would remain if the KIVY_VENV_DIRS was not set.  KIVY_VENV_DIRS would be higher priority that KIVY_HOME.  If the KIVY_DESKTOP_PATH_ID was set, and KIVY_VENV_DIRS was set, the path_id would be used as part of the project dir paths.  The `user_data` and `user_cache` dirs would only be created on access (as they are now).

I offer this as an option to meet the spirit of your request.  I am not advocating for it.  I prefer that App.user_*_dir maintain the current behavior.  If the primary concern is a clean-up after the example runs, I can solve that problem.

kuzeyron: I have this in mind when it exits the application it might not always work for it to clean itself if segfault is happening. Leaving behind empty folders is a no go as the user might changes the names. Even though on Linux they could be placed in `/tmp`. Then it wouldn't bother at all if it leaves behind empty folders.

ElliotGarbus: I'm not sure I understand your comment, are you talking specifically about the example?

To clean up the example, I could add directions for the user to delete the 2 directories that have been created, and create a small utility program the does the cleanup for the original settings.

kuzeyron: I'm talking about kivy/app.py

ElliotGarbus: I apologize, I just don't understand your comment.  Can you expand?

The App.user_data_dir and App.cache_data_dir only create a directory if they are accessed.  

The behavior of App.user_data_dir and App.user_cache_dir is to be persistent from run to run.   user_data_dir is persistent and preserved by the OS.  The user_cache_dir is persistent but can be cleared by OS, is not guaranteed to be backed up.   It is for data that can be recreated, but is used to enhance performance from run to run.  

If losing the file would break the user’s data → user_data_dir.
If losing the file only slows you down → user_cache_dir.

kuzeyron: > only create a directory if they are accessed

On my end it creates these folders without doing anything extra.

If cache dir would be placed in `/tmp` it would solve the issue with empty folders as the system wipes out `/tmp` on shutdown.

And then we have data dir that might need some more love to its setup

ElliotGarbus: Here is a small test:
```python
import os
# os.environ['KIVY_DESKTOP_PATH_ID'] = 'My Test to Evaluate Kivy_Desktop_Path_ID'

from kivy.app import App
from kivy.lang import Builder
from kivy import kivy_home_dir, kivy_config_fn

kv = """
AnchorLayout:
    Button:
        size_hint: None, None
        size: dp(200), dp(48)
        text: 'Test Access user_data_dir'
        on_release: app.access_user_data_dir()
"""

class ConfigOnlyCheckApp(App):
    def build(self):

    def on_start(self):
        print(f"kivy_home_dir: {kivy_home_dir}")
        print(f"config_path: {kivy_config_fn}")

    def access_user_data_dir(self):
        print(f"user_data_dir: {self.user_data_dir}")


if __name__ == '__main__':
    ConfigOnlyCheckApp().run()
```
Running this test we see the following behaviors:

With `KIVY_DESKTOP_PATH_ID` not set, commented out as above
- The user_data_dir is not created, it is created when the button is pressed
- The .kivy file is placed at the project path (if running under a venv)
- The user_cache_dir is not created

With `KIVY_DESKTOP_PATH_ID` set:
- The user_data_dir is created, the ,kivy dir is located in the user_data_dir
- The user_cache_dir is NOT created

The behavior of the cache dir and the tmp dir are not the same.  
Cache directories:
- Persist across application runs (expected behavior)
- Cleared by OS only when storage is critically low
- Users/apps can manually clear them
- Expected to survive reboots

Temp directories (%TEMP%, /tmp):
- May be cleared on every reboot
- More aggressively cleaned by the OS
- Not expected to persist across sessions

kuzeyron: alright, I'll get back at this tomorrow

kuzeyron:
@misl6 Any final comment to add?

misl6:
> @misl6 Any final comment to add?

Well, just an awesome work!

Thank you @ElliotGarbus and @kuzeyron !

PR #9256, opened 2026-01-29
Add example and documentation for app-specific KIVY_HOME for desktop platforms

ElliotGarbus:
Provides a ready-to-use implementation and example for setting KIVY_HOME to an app-specific directory before importing Kivy. This ensures each application has isolated configuration, logs, and modules rather than sharing ~/.kivy on desktop platforms.
This is related to issue #8571 

Changes:
- Add examples/kivy_home/ with set_kivy_home.py utility and demo app
- Update environment.rst to document app-specific KIVY_HOME configuration
- Update config.rst to reference the example
- Update app.py module docstring with KIVY_HOME guidance

Platform-specific paths when using app-specific KIVY_HOME:
- Windows: %APPDATA%\<appname>\.kivy
- macOS: ~/Library/Application Support/<appname>/.kivy
- Linux: ~/.local/share/<appname>/.kivy (XDG compliant)

The example includes:
- Platform detection and path generation
- App name normalization (lowercase, strip 'app' suffix)
- Runtime check to ensure KIVY_HOME is set before kivy.config import
- Complete demo app showing configuration paths
- Documentation explaining when and how to use it

Maintainer merge checklist

### DIFF
--- a/examples/kivy_home/set_kivy_home.py
+++ b/examples/kivy_home/set_kivy_home.py
@@ -1,0 +1,118 @@
+"""
+Set KIVY_HOME for App-Specific Configuration
+=============================================
+
+This module provides a utility to set KIVY_HOME to an app-specific directory
+before importing Kivy. This ensures each application has its own isolated
+configuration, logs, and cache directories.
+
+Usage:
+    from set_kivy_home import set_app_name  # BEFORE importing kivy
+    set_app_name('MyApp')
+
+    from kivy.app import App  # Now safe to import
+
+Platform Paths:
+    Windows: %APPDATA%/<appname>/.kivy
+    macOS: ~/Library/Application Support/<appname>/.kivy
+    Linux: $XDG_DATA_HOME/<appname>/.kivy (typically ~/.local/share/<appname>/.kivy)
+
+Note:
+    This must be called BEFORE importing any kivy modules, otherwise
+    RuntimeError will be raised.
+"""
+
+from os import environ
+import sys
+from sys import platform as _sys_platform
+from pathlib import Path
+
+
+def _get_platform():
+    """Get the current platform (from kivy.utils)."""
+    # On Android sys.platform returns 'linux2', so prefer to check the
+    # existence of environ variables set during Python initialization
+    kivy_build = environ.get('KIVY_BUILD', '')
+    if kivy_build in {'android', 'ios'}:
+        return kivy_build
+    elif 'P4A_BOOTSTRAP' in environ:
+        return 'android'
+    elif 'ANDROID_ARGUMENT' in environ:
+        # We used to use this method to detect android platform,
+        # leaving it here to be backwards compatible with `pydroid3`
+        # and similar tools outside kivy's ecosystem
+        return 'android'
+    elif _sys_platform in ('win32', 'cygwin'):
+        return 'win'
+    elif _sys_platform == 'darwin':
+        return 'macosx'
+    elif _sys_platform.startswith('linux'):
+        return 'linux'
+    elif _sys_platform.startswith('freebsd'):
+        return 'linux'
+    return 'unknown'
+
+
+def set_app_name(app_name):
+    """
+    Set the app name for desktop platforms to configure KIVY_HOME.
+
+    This function sets KIVY_HOME to an app-specific directory based on
+    platform conventions. It must be called BEFORE importing any Kivy modules.
+
+    Args:
+        app_name (str): The name of your application. The name will be
+            normalized (lowercased, 'app' suffix stripped) to match
+            App.user_data_dir behavior.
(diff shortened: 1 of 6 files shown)


ElliotGarbus:
Note: Pep8 failures are not related to this PR.

examples/kivy_home/set_kivy_home.py
```diff
@@ -0,0 +1,118 @@
...
+        raise RuntimeError(
+            "Config has already been imported. "
+            "set_app_name must execute before kivy.config to ensure "
+            "proper initialization order."
+        )
+
+    # Normalize app name (lowercase, strip 'app' suffix)
+    # This matches the behavior of App.user_data_dir
+    name = app_name.lower()
+    name = name[:-3] if name.endswith('app') else name
```
kuzeyron, line 100: In my book it would be nicer if it look for the app title before it uses class name of the app.

examples/kivy_home/set_kivy_home.py
```diff
@@ -0,0 +1,118 @@
+"""
```
kuzeyron, line 1: Maybe this file should be directly used from Kivy when it reads the app title?

examples/kivy_home/set_kivy_home.py
ElliotGarbus: The desire is the put the `.kivy` directory under the `user_data_dir`.  

Looking here: https://github.com/kivy/kivy/blob/22426a44a31e5e0fca07471e6cf51f5bf21275f8/kivy/app.py#L853
https://github.com/kivy/kivy/blob/22426a44a31e5e0fca07471e6cf51f5bf21275f8/kivy/app.py#L915
We see the App.user_data_dir is based on the class name, not the title.  

I could change App.user_data_dir to be based on the title and fallback to the class name, and use the same behavior here.  What do you think?

kuzeyron: That would be great

examples/kivy_home/set_kivy_home.py
ElliotGarbus: @kuzeyron Thanks for your feedback.

The problem is that config.py reads the config file on import, and creates the Config object (and reads from `KIVY_HOME/config.ini`).  The env var KIVY_HOME is read in `kivy.__init__`.   

I had attempted to fix this by creating a Config proxy, with that had a separate initialize phase and deferred the initialize phase until App was instanced.  This required changes to about 30 files before I abandoned the approach for this much simpler one. Clock, Window, behaviors, inputs and widgets rely on data coming from `config.ini`.  

I had another idea to create a registry in `~/.kivy`.  The registry would associate the App.name with the `config.ini` path.  If the app did have a registry entry the first time it started, it would write the registry and restart.  The registry would hold the path to `config.ini`.  This seemed a bit odd. 

I also considered making `config.ini` read-only and putting it `kivy/data` next to `style.kv`  I think given the way most use Config this would be reasonable.  This would require changing how the default kivy settings panel works.  I wondered if that was important for non-conventional platforms - so I opted against that direction.

I concluded setting KIVY_HOME prior to startup was a better compromise.   I'm very open to input.

FWIW, In Android KIVY_HOME is set the P4A bootstrap.

examples/kivy_home/set_kivy_home.py
ElliotGarbus: I have an open PR: https://github.com/kivy/kivy/pull/9255
I will make the changes there.  I will mark this PR, and 9255 as draft.
I will also need to take a closer look at the mobile platforms to ensure consitency.

ElliotGarbus: Basing user_data_dir on title introduces an ordering hazard. The title attribute is commonly set later in the application lifecycle (often in build()), while user_data_dir may be accessed before title is set. 

If title is defined after user_data_dir is first used, the resulting directory becomes order-dependent. Using App.name avoids this issue because it is stable and available at App instantiation time.  

The title is designed to be used dynamically and user_data_dir is generated on first use.  

An alternative that meets the spirit of your request could be satisfied by creating a new env var DESKTOP_PATH_ID.  This could be set with a string that is more useful than the app name, not burdened with the title behaviors, and be use create KIVY_HOME, and the user_data_dir and user_cache_dir paths. App.name could be used a fallback.

What do you think, should I update the PR to use DESKTOP_PATH_ID?

kuzeyron: Do whatever necessary :)

ElliotGarbus: I have the implementation almost completed.  I will be closing this PR, and integrating the change using the new env var with the PR that updated the user_data_dir.  It ends up being a much cleaner solution!

PR #9263, opened 2026-02-13
`boxshadow`: add `optimize` option

### LINKED ISSUE #8593: BoxShadow bug
The `BoxShadow` instruction has a bug: it does not take into account the relative size of the widget. I'm going to demonstrate it now:

```python
from kivymd.app import App
from kivy.lang import Builder

KV = """
Screen:
    canvas:
        Color:
            rgba: 1, 1, 1, 1
        Rectangle:
            pos: self.pos
            size: self.size


    Button:
        pos_hint: {"center_x": .5, "center_y": .5}
        size_hint: .5, .5
        background_color: 0, 0, 0, 0

        canvas.before:
            Color:
                rgba: 0, 0, 0, .85
            BoxShadow:
                pos: self.pos
                size: self.size
                
                offset: 0, -2
                spread_radius: -4, -4
                border_radius: 14, 14, 14, 14
                blur_radius: 10

        canvas.after:
            Color:
                rgba: 1, 1, 1, 1
            RoundedRectangle:
                pos: self.pos
                size: self.size
                radius: [20,]
"""


class Example(App):
    def build(self):


Example().run()
```

When we run this code, we see what we expected to see:
Снимок экрана 2024-01-27 в 13 51 57
And now let's increase the size of the application window:

https://github.com/kivy/kivy/assets/16930280/fa5d61dc-c392-46ec-af4e-5e54de6ee1c5

We see that the borders of the shadow become sharp, their edges are cut off and the radius of the shadow no longer corresponds to the original one:

Снимок экрана 2024-01-27 в 13 57 30

And this is how, on closer inspection, the shadow looks, the size of the parent widget of which has not changed:

Снимок экрана 2024-01-27 в 14 02 44

T-Dynamos:
Closes #8593 

CC: @DexerBR 

~~We set `optimize` to `False` by default because it affects quality. Although it can change performance, we prioritize *better output quality* over the *performance* gain.~~

I get around ~250 fps with optimization disabled, with vsync off and maxfps 2000. 
and around ~340 fps with optimization enabled.

Test code:

```python
from kivy.app import App
from kivy.lang import Builder
from kivy.clock import Clock
from kivy.properties import BooleanProperty

KV = """
Screen:
    canvas:
        Color:
            rgba: 1, 1, 1, 1
        Rectangle:
            pos: self.pos
            size: self.size


    Button:
        id: button
        pos_hint: {"center_x": .5, "center_y": .5}
        size_hint: .3, .3
        background_color: 0,0,0,0

        canvas.before:
            Color:
                rgba: 0, 0, 0, 0.5

            BoxShadow:
                optimize: app.optimize_enabled
                pos: self.pos
                size: self.size
                offset: 0, 0
                border_radius: [self.size[0]/8]*4
                blur_radius:20

        canvas.after:
            Color:
                rgba: 1, 1,1, 1
            SmoothRoundedRectangle:
                pos: self.pos
                size: self.size
                radius: [self.size[0]/8,]

    Label:
        id: fps_label
        text: "FPS: 0"
        color: 0, 0, 0, 1
        size_hint: None, None
        size: self.texture_size
        pos: 10, self.parent.height - self.height - 10

    ToggleButton:
        id: optimize_toggle
        text: "optimize: ON" if self.state == "down" else "optimize: OFF"
        size_hint: None, None
        size: 160, 36
        pos: 10, 10
        state: "down" if app.optimize_enabled else "normal"
        on_state: app.optimize_enabled = (self.state == "down")
"""


class Example(App):
    optimize_enabled = BooleanProperty(False)

    def change_size_widget(self):
        Animation(size_hint=(0.8, 0.8)).start(self.root.ids.button)

    def build(self):
        root = Builder.load_string(KV)
        Clock.schedule_interval(self._update_fps, 0)
        Clock.schedule_once(lambda *_: self._pulse_size(root.ids.button), 0)
        return root

    def _update_fps(self, _dt):
        label = self.root.ids.fps_label
        label.text = f"FPS: {Clock.get_fps():.1f}"
        label.size = label.texture_size

    def _pulse_size(self, widget):
        grow = Animation(size_hint=(0.85, 0.85), duration=0.8, t="in_out_sine")
        shrink = Animation(size_hint=(0.3, 0.3), duration=0.8, t="in_out_sine")
        anim = grow + shrink
        anim.bind(on_complete=lambda *_: self._pulse_size(widget))
        anim.start(widget)


Example().run()
```

Screen recording drops fps a bit.

https://github.com/user-attachments/assets/67628f93-12f7-4b04-bb0f-97e15a64c4ed

Maintainer merge checklist

### DIFF
--- a/kivy/graphics/boxshadow.pxd
+++ b/kivy/graphics/boxshadow.pxd
@@ -14,6 +14,7 @@ cdef class BoxShadow(InstructionGroup):
     cdef tuple _border_radius
     cdef tuple _spread_radius
     cdef bint _optimize
+    cdef tuple _optimize_fbo_size
     cdef VertexInstruction _fbo_rect
     cdef VertexInstruction _texture_container
     cdef Scale _fbo_scale
--- a/kivy/graphics/boxshadow.pyx
+++ b/kivy/graphics/boxshadow.pyx
@@ -65,8 +65,6 @@ from kivy.graphics.gl_instructions import ClearBuffers, ClearColor
 from kivy.graphics.instructions cimport InstructionGroup
 
 
-cdef tuple FBO_BASE_SIZE = (100, 100)
-
 cdef str SHADOW_fs = """
 #ifdef GL_ES
     precision highp float;
@@ -178,10 +176,13 @@ cdef class BoxShadow(InstructionGroup):
         `border_radius`: list | tuple, defaults to ``(0.0, 0.0, 0.0, 0.0)``.
             Specifies the radii used for the rounded corners clockwise:
             top-left, top-right, bottom-right, bottom-left.
-        `optimize`: bool, defaults to ``False``.
+        `optimize`: bool, defaults to ``True``.
             When enabled, uses a fixed size FBO and scales it to the
             target size for better performance. When disabled, rebuilds the FBO
             to match the target size for higher quality.
+        `optimize_fbo_size`: list | tuple, defaults to ``(100.0, 100.0)``.
+            Defines the fixed FBO size used when :attr:`optimize` is ``True``.
+            Larger values improve quality but increase memory and startup cost.
     '''
 
     def __init__(self, *args, **kwargs):
@@ -193,7 +194,8 @@ cdef class BoxShadow(InstructionGroup):
         blur_radius = kwargs.get("blur_radius", 15.0)
         spread_radius = kwargs.get("spread_radius", (0.0, 0.0))
         border_radius = kwargs.get("border_radius", (0.0, 0.0, 0.0, 0.0))
-        optimize = kwargs.get("optimize", False)
+        optimize = kwargs.get("optimize", True)
+        optimize_fbo_size = kwargs.get("optimize_fbo_size", (100.0, 100.0))
 
         self._inset = self._check_bool(inset)
         self._pos = self._check_iter("pos", pos)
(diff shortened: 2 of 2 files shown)


DexerBR:
@T-Dynamos I took a look at the PR and your approach is interesting! The optimization flag is a nice addition.

However, I was wondering if you could also include a property called `optimize_fbo_size` that allows users to control the resolution of the optimized FBO, instead of having it hardcoded to (100, 100).

This would give users more flexibility to fine-tune the quality/performance trade-off.

Usage example:
```python
BoxShadow:
    optimize: True
    optimize_fbo_size: 150, 150  # custom optimized resolution
    size: self.parent.size
    blur_radius: 20
```

The property would default to (100, 100) for backward compatibility and would only be used when `optimize: True`.

What do you think?

DexerBR:
Regarding setting `optimize` to `False` by default, I'm a bit concerned because in the tests I ran during the development of this feature, creating/recreating FBOs and textures in Kivy is a relatively expensive operation.

For example, consider an app that starts with 25 buttons, each generating a shadow texture of 2000x2000 pixels. That's 25 FBO creation operations at startup, which can cause noticeable stuttering.

Other problematic scenarios:
- ScrollView with multiple BoxShadows being resized during scroll (constant FBO recreation)
- Animations that change the size of elements with shadows (FBO rebuild every frame)
- Responsive layouts that frequently recalculate sizes

With `optimize: True` by default:
- FBOs created once with a fixed size
- Only scaling (cheap operation) instead of texture recreation
- Predictable and consistent performance

Perhaps the best approach would be:
1. `optimize: True` by default (performance first)
2. Users who need maximum quality can explicitly disable it
3. `optimize_fbo_size` allows fine-tuning when needed

T-Dynamos:
@DexerBR Yes I agree, setting `optimize` to true by default would be an ideal option. I forgot to account for performance degradation caused by multiple fbos :). And yes that idea of `optimize_fbo_size` sounds good. I will implement it soon.

PR #9270, opened 2026-02-19
Build SDL3 dependencies during wheel build also for iOS to centralize deps handling

misl6:
Maintainer merge checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,6 @@ from collections import OrderedDict
 from os import walk, environ, makedirs
 from os.path import join, dirname, exists, basename, isdir
 import os
-import shutil
 from copy import deepcopy
 from kivy.utils import pi_version
 import sys
@@ -757,13 +756,6 @@ def determine_base_flags():
         'library_dirs': [] + libs,
         'extra_link_args': [],
         'extra_compile_args': []}
-    # if c_options['use_ios']:
-    #     sysroot = environ.get('IOSSDKROOT', environ.get('SDKROOT'))
-    #     if not sysroot:
-    #         raise Exception('IOSSDKROOT is not set')
-    #     flags['include_dirs'] += [sysroot]
-    #     flags['extra_compile_args'] += ['-isysroot', sysroot]
-    #     flags['extra_link_args'] += ['-isysroot', sysroot]
     if platform.startswith('freebsd'):
         flags['include_dirs'] += [join(
             environ.get('LOCALBASE', '/usr/local'), 'include')]
@@ -1434,6 +1426,7 @@ if isdir(binary_deps_path):
             binary_deps.append(
                 join(root.replace(binary_deps_path, 'binary_deps'), fname))
 
+
 def glob_paths(*patterns, excludes=('.pyc', )):
     files = []
     base = Path(join(src_path, 'kivy'))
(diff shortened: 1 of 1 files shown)


psychowasp:
how are the SDL3 components going to be handled afterward ?, if they are build when ios wheel is created 

i mean afterwards in a xcode project? (not only kivy-ios but other xcode generator project as well?)

misl6:
> how are the SDL3 components going to be handled afterward ?, if they are build when ios wheel is created
> 
> i mean afterwards in a xcode project? (not only kivy-ios but other xcode generator project as well?)

I guess other Xcode generators will need to add a specific implementation to handle these cases. I talked to the PEP 730 author, and he also suggested to have a `.frameworks` (or something similar, as it happens for macOS) folder inside the wheel, where these frameworks could be placed.

The Xcode generators will need to look-up for valid frameworks (in all the folders, unless we define a standard) and move them to a proper folder + link to XCode project.

psychowasp:
In other words the generated frameworks will be bundled with the kivy ios wheel ?

PSProject / kivyschool.com/kivy-on-ios
is based on all is done in pyproject.toml, with the project.requirements

and psproject, always uses the toml file, for info / pip installing the requirements, soo i guess the backend for using kivy in psproject, just needs to look in requirements and pip download the wheel based on what kivy version requirements has. And i just need to move the frameworks and add them to xcode ?

but the PEP730 suggested runscript, only handles .so files, so i assume the wheel is not going to drop them in site-packages by default , But frameworks just bundled along the wheel, for manual copy/move ?

psychowasp:
tried the PR but I see the part that does the .framework part isn't implemented yet ?

but I assume you just building it like this now ?

```
export CIBW_BEFORE_ALL_IOS="./tools/cibuildwheel-before-all-ios.sh"
export KIVY_DEPS_ROOT="$(pwd)/ios-kivy-dependencies"
export KIVY_SPLIT_EXAMPLES="1"

cibuildwheel --platform ios --archs "arm64_iphoneos x86_64_iphonesimulator arm64_iphonesimulator" --output-dir "$OUTPUT_DIR"
```

but this could also have been specified in the pyproject.toml

```toml
[tool.cibuildwheel.ios]
before-all = "./tools/cibuildwheel-before-all-ios.sh"
environment = { KIVY_DEPS_ROOT = "$(pwd)/ios-kivy-dependencies", KIVY_SPLIT_EXAMPLES = "1" }
```

and will only trigger when cibuildwheel --platform ios

because now I only need to run

cibuildwheel --platform ios --archs "arm64_iphoneos x86_64_iphonesimulator arm64_iphonesimulator" --output-dir "$OUTPUT_DIR"

psychowasp:
tried it now

added as post-process after pip install of requirements is done in psproject code
IMG_5175

xcode project support folder where SDL frameworks was dated back to 2025 
IMG_5176

Perfetto!!!!

misl6:
> tried it now
> 
> added as post-process after pip install of requirements is done in psproject code IMG_5175
> 
> xcode project support folder where SDL frameworks was dated back to 2025 IMG_5176
> 
> Perfetto!!!!

Nice!! Will merge in few seconds, so we can continue enhancing the support for iOS (and update SDL3 frameworks as well)

PR #9272, opened 2026-02-27
fix: replace 61 bare except clauses with except Exception

haosenwang1018:
Replace bare `except:` clauses with `except Exception:` for PEP 8 compliance.


kuzeyron:
Thank you!

Issue #8573, opened 2024-01-20
Add an `SVG` parser + renderer core provider


misl6:
Our current SVG implementation is lightweight and does not need any extra dependency, but is buggy and incomplete.
That's the pain point of trying to maintain a custom implementation.

For Kivy`3.0.0` the proposal is to move the SVG parser+renderer to the "core provider" concept, to open the opportunity of having multiple implementations available, and ease the integration with external tools.

Having a fully working SVG parser+renderer means that our users can start using SVG icons in their apps, but also that we can abandon `.png` files for icons in our default theme (and themes?)

baseplate-admin:
Is this up for discussion?

Technically we have 2 options

1. Using `skia` to render svg
2. Using `usvg` to render svg.

Since `kivy` has shown interest in `skia` i think maybe that's the better way forward

misl6:
When referring to the "core provider", we're talking about an abstraction that uses one of the available implementations to achieve the result.

So, we can have both, but certainly, most of the time investment made in Skia for text layouting will also help here.

ElliotGarbus:
I'm about to get started on an SVG provider.  Given that Skia has not yet been implemented into the main branch, I'm going to start with resvg_py.  The goal will be to go from SVG to the ImageData used in the image provider.  After Skia is integrated into Kivy an additional provider could be created.

@baseplate-admin I see you are the maintainer for resvg_py.  I'd welcome your thoughts.

baseplate-admin:
Hi, Thank you for mentioning me.

> The goal will be to go from SVG to the ImageData used in the image provider.

Okay so whats the general flow for this, can we convert svg-to-png in memory and then convert that to ImageData?
`resvg-py` was originally intended to be used as a high level wrapper around `resvg`, that means currently the module only supports this operation
* `svg` -> `png`


When i created the library, i didn't see any interest of exposing lower level bindings.

---

With that being said, i am open to (and willing to contribute to `kivy`) adding any functionalities the community desires :)

ElliotGarbus:
@baseplate-admin 

My current thought:
resvg_py: svg_to_bytes(svg_path=..., width=..., height=...) → PNG bytes
PIL: Image.open(io.BytesIO(png_bytes)).convert('RGBA') → raw RGBA pixels
Kivy: ImageData(w, h, 'rgba', pixels) or add_mipmap(level, w, h, pixels, 0)

If resvg-py could deliver RGBA bytes that would skip the conversion. 

[...]

PR #9273, opened 2026-03-02
Added an SVG image provider, tests and example.

ElliotGarbus:
Adds an SVG image provider.  

I beilve this statisfies the issue: #8573 
_"Having a fully working SVG parser+renderer means that our users can start using SVG icons in their apps, but also that we can abandon .png files for icons in our default theme (and themes?)"_

The Image widget and canvas instructions can now use svg files.
The provider is for `thorvg-python`, A ctypes API for thorvg, with additional functions for getting Pillow Image (https://pypi.org/project/thorvg-python/).   ThorVG is is a production-ready vector graphics engine.  Details: https://github.com/thorvg/thorvg

The idea of the PR is to be able to use svg icons and SVGs for buttons and other images in Kivy widgets.  It rasterizes the svg into an image.  The image is rendered once.  It will not re-render on size changes.  The user has the option to set the mipmap to improve image quality when scaling down the images.  This is implemented as an image provider so the textures are cached. 

To use the provider, thorvg-python must be installed. 
>pip install thorvg-python

A few items that deserve focus:
I have added a new config section `svg`, with one entry 'default size'  this is used to set a minimum size for rendered svg if the native size is below the default size.  This is of course documented, but deserves some thought.  Is this the correct place for this?  Should this be another option in graphics?  I'm open.

The image provider URI as also been expanded to support optional parameters, and i have created a size parameter to support setting the min rendering size in the URI.

This is the first of 3 PR's I expect to create based on ThorVG.
1) SVG support for Images - This PR
2) SVG Vector support - An SVG widget that will support rendering on upscaling (in buckets) and have limited features to manipulate the DOM.  Set current color, show/hide by id... This will be implemented as an SVG provider and an SVGWidget that relies on the provider.  This will require the creation of a new provider class.
3) Lottie Support - A LottieWidget and an Lottie animation provider.

## Stack

This PR is part of set of PRs that use ThorVG:

- **#9297** — `kivy.lib.thorvg` Cython wrapper + desktop CI + bundled ThorVG v1.0.4 *(base of the stack — merge first)*
- **#9273** — SVG image-loader migration (`img_thorvg_svg`)
- **#9284** — SVG widget migration (`svg_thorvg`) + zero-copy render *(this PR)*
- **#9293** — Lottie provider migration (`lottie_thorvg`) + zero-copy per-frame blit

All three consumer PRs are rebased on `feature/thorvg-cython`, so their diffs currently include the wrapper commits; those will drop out automatically once #9297 merges. The three consumer PRs are independent of each other and can land in any order after #9297.

Maintainer merge checklist


kivy/core/image/img_thorvg_svg.py
```diff
@@ -0,0 +1,258 @@
...
+If the package is not available the loader is silently skipped and Kivy's
+other image providers remain unaffected.
+'''
+
+__all__ = ('ImageLoaderThorvgSvg', )
+
+
+
+try:
+    import thorvg_python as tvg
```
kuzeyron: If the package is relatively small it might be worth adding it to https://github.com/kivy/kivy/blob/master/setup.cfg#L18 but then again is the question about pillow.. Could these operations be done without Pillow?

ElliotGarbus: The thorvg-python v1.0.1 installed size: ~2.4 MB  
pillow is installed as part of the kivy["base"] install.  I was relying on pillow for the conversion to RGBA.  I changed the code to pull the data directly from the buffer, bypassing pillow.

Should I add thorvg-python to the "base" install?

kuzeyron: That would be great I think so it doesn't complicate it for the users. I wonder if the kivy-sdk can include this for the deps being used with pip installment. Might be worth asking @misl6 about that as he knows more about it than I do.

ElliotGarbus: @kuzeyron Looking at thorvg-python, I only see wheels for Windows, Linux, MacOS.  
https://pypi.org/project/thorvg-python/1.1.1/#files

ThorVG (https://github.com/thorvg/thorvg) the underlying C++ lib has binaries for IOS and Android.  Should I ask the thorvg-python maintainer to add support for the mobile platforms or look to create a PR to thorvg-python?  

Is there some additional integration required for the Kivy IOS/Android tool chains?

kuzeyron: > Is there some additional integration required for the Kivy IOS/Android tool chains?

I can only think of recipes.

> Should I ask the thorvg-python maintainer to add suppor..

Absolutely, that would ease the pain when it comes to Android and iOS.

ElliotGarbus:
@kuzeyron Doing a little more investigation I learned the, Apple App Store/IOS requires a statically linked library so the ctypes approach will not work for IOS.  Using Cython to expose the API is the desired approach.  Once I go down this path I might as well use the same Cython wrapper for all platforms.

My path going forward will be to build out the 3 elements (Image provider for SVG, SvgWidget (and provider), LottieWidget (and provider).  I will use thorvg-python for my development work, but ultimately replace thorvg-python with "kivy-thorvg" (or "thorvg-kivy")? , The ThorVG C++ library and a new Cython wrapper that supports all platforms.  

 Then with consultation with @misl6  we can decide how and where to package the "kivy-thorvg" binary, integrate it into the install process.  
 
 I would think the cleanest way forward is for me to submit 2 additional PRs (SvgWidget, LottieWidget and providers) using thorvg-python, and then do follow up PRs to replace the optionally loaded thorvg-python with "kivy-thorvg".
 
 @kuzeyron I look forward to your feedback.

EDIT: Apple App Store/IOS no longer requires a statically linked library for IOS - This is out of date.  I will continue to use thorvg-python development.  The solution can be a kivy-ios recipe for the existing ctypes wrapper - or a new wrapper for all.  TBD later.

ElliotGarbus:
updated to remove workarounds after thorvg fixes,  Made changes to match updated ThorVG api update

misl6:
Hi everyone!

Did not had the time to reply in the past month, but I've followed the progress via notifications.
Thank you @ElliotGarbus for taking care of it!

Even if as per now the `thorvg-python` is not yet available for Android and iOS (even Kivy is not officially building wheels yet), we can consider it for inclusion as-is, without creating any extra package, at least at that stage.

If in future, we want to reduce the footprint of an app using the SVG feature we might think of creating a kivy-specific package that uses throvg and exposes (and builds) only the features we actually need.

Will look into the PR soon and leave a review.

kivy/tests/test_thorvg_svg_provider.py
```diff
@@ -0,0 +1,302 @@
...
+        assert self.re.match('image.png') is None
+
+    def test_wrong_prefix_does_not_match(self):
+        assert self.re.match('@provider:pil(image.png)') is None
+
+    def test_missing_provider_name_does_not_match(self):
+        assert self.re.match('@image_provider:(image.png)') is None
+
+    def test_missing_path_does_not_match(self):
+        assert self.re.match('@image_provider:pil') is None
```
misl6: Should not be in generic test_image_provider.py?

kivy/core/image/img_thorvg_svg.py
```diff
@@ -0,0 +1,270 @@
...
+property directly::
+
+    Image(source='icons/house.svg', mipmap=True)
+
+Requirements
+------------
+
+``thorvg-python`` must be installed::
+
+    pip install thorvg-python
```
misl6, line 63: We can actually add an `svg` option, and than add here the `thorvg-python` dependency. (`pip install kivy[svg]`, obviously `pip install kivy[full]`, will also include `svg`).

See: https://github.com/kivy/kivy/blob/255863d342059dcd14f4f1e563cc3fc56571a231/setup.cfg#L29

If in future we change our mind about the `thorvg-python` dependency and we decide to create our own, that option will continue to work, and we do not break anything on users side.

kivy/tests/test_thorvg_svg_provider.py
ElliotGarbus: Good catch!  I was so focused on testing what I changed I didn't give it a thought!

kivy/core/image/img_thorvg_svg.py
ElliotGarbus: @misl6 My current thought is to NOT use thorvg-python and create a minimal cython wrapper for ThorVG that only supports the APIs used by Kivy in the SVG and Lottie PRs.  

While the thorvg-python maintainer has been quite responsive, it appears to be a single person project not formally part of the ThorVG project.   The wrapper is rather simple - but I'm not sure we want to create this dependency.

ThorVG (C++) has an active community lots of contributors, some commercial users and sponsors. 

My proposal is create a minimal cython wrapper for ThorVG and integrate this within the kivy build system.  If there is support for this idea, I'm sure I can compete this rather quickly and update the SVG/Lottie PRs to utilize the new backend.

I welcome your feedback.

kivy/tests/test_thorvg_svg_provider.py
ElliotGarbus: Updated tests

kivy/core/image/img_thorvg_svg.py
ElliotGarbus: I have created a cython wrapped version of ThorVG, and integrated it into kivy as described here: https://github.com/kivy/kivy/pull/9297

I have updated this PR to use the cython wrapped ThorVG

ElliotGarbus:
Updated to use kivy.lib.thorvg

kivy/core/image/img_thorvg_svg.py
ElliotGarbus: I have added thorvg to kivy.lib.thorvg  See:https://github.com/kivy/kivy/pull/9297

misl6 (approved):
LGTM. Thank you!

PR #9275, opened 2026-03-04
`Clock`: Enhance and fix `@triggered` decorator (instance isolation, debouncing, lazy initialization) and add deterministic pytest fixture for manual clock advancement

DexerBR:
### Summary
This PR refactors the `@triggered` decorator to improve its behavior, especially when used inside classes (instance methods). It fixes a long-standing issue where different instances of the same class shared the same trigger state.

### Key Features
- **Instance Isolation**: Methods now have independent triggers for each object, preventing one widget from canceling another's schedule.
- **Debouncing**: Added a new `debounce=True` option to reset the timer on every call.
- **Lazy Initialization**: Triggers are now instantiated only upon their first call. This eliminates the overhead of creating Clock objects for thousands of widgets at once, saving memory and CPU.
- Added `.is_triggered` property and improved `.cancel()` method support.

### Migration Note
The legacy shared behavior (where all instances shared one trigger) can be restored by simply adding `@classmethod` above the `@triggered` decorator.

### Documentation
- Updated `clock.py` with detailed technical comments on the Descriptor pattern.
- Updated `migration.rst` with examples for new behaviors and decorator stacking conventions.

### Tests
- Added extensive tests for the `@triggered` decorator, including instance isolation, debouncing, and lazy initialization.
- **kivy_clock_advance (new fixture)**: Introduced a deterministic way to test time-based logic. It allows "advancing the clock" virtually, eliminating flaky test failures caused by CPU load in CI environments. Instead of real-time `sleep` or other methods based on real time, tests now can jump instantly to a specific point in the future.

Maintainer merge checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,4 @@ kivy/core/clipboard/_clipboard_sdl3.c
 kivy/graphics/egl_backend/egl_angle.c
 kivy/core/image/_img_sdl3.cpp
 wheelhouse/
+/.vscode
--- a/doc/sources/migration.rst
+++ b/doc/sources/migration.rst
@@ -875,6 +875,68 @@ You no longer need to manually clean up groups:
 +---------------------------+---------------------------+----------------------------------------+
 
 
+=====
+Clock
+=====
+
+*Improved @triggered Decorator Behavior, Instance Isolation and Debouncing*
+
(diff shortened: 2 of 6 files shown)


kuzeyron (approved):

kuzeyron:
LGTM! Tested locally and everything is running.

Issue #8575, opened 2024-01-20
Add `TextEditor` core provider, rewrite `TextInput` UI element on top of it


misl6:
Similar to https://github.com/kivy/kivy/issues/8574, handling text input internationally is a painful task, and sometimes we do not even have a proper API to handle it, as text input is something strictly tied with the OS-backed TextEdit widget. (This is not just Kivy's case, but also other frameworks are on the same boat).

The current status:

- The window provider (SDL in most cases) does not have (and is not required to?) all the events and methods to access the complete feature set that the (mostly mobile) platforms offer.

- We currently only support basic text input methods (we receive keystrokes from the keyboard and change the TextInput widget text property accordingly). This may be fine for languages that do not require an Input Method Editor ([IME](https://learn.microsoft.com/en-gb/windows/apps/design/input/input-method-editors)) to input text.

- The window provider (SDL in that case) is not able to handle multiple TextInput fields at the same time. That breaks the keyboard suggestions for us.

- The current implementation does not allow us to handle autofills (E.g. for password managers)

- The current implementation does not allow us to handle accessibility features on textinputs.

We have a POC https://github.com/kivy/kivy/pull/8226, whose major intention was to test the feasibility of using OS-backed hidden widgets to get the events. That worked quite fine. We need to find out how to manage it for Desktop platforms, and write a unified API + keep the support for the legacy (keystrokes) version.

**Please note** that this issue is partially on hold until https://github.com/kivy/kivy/issues/8574 is assessed, since `kivy.ui.textinput` will use `TextLayout` core provider (for layouting, rendering, getting selection rectangles, getting glyph at cursor pos, ...) and `TextEditor` core provider for handling the textedit events.

Julian-O:
`textinput.py` stands at 4,050 lines of code. `TextInput` is a single class with 3,600 lines of code. 

This would be an excellent opportunity to break it down into much, much smaller parts with subclassing, mixins, delegation, etc. and improve unit-test coverage.

Issue #8574, opened 2024-01-20
Add `TextLayout` core provider, remove / migrate old text provider implementation


misl6:
RTL, BiDi, justification, line height, letter spacing, emojis, font fallback, ellipses, line breaking, max lines ...
Oh, what a mess.

Handling text and layouting it correctly, taking care of the needs of all the different scripts and languages from all around the world is not an easy task. But is something we certainly need to have to grow as a framework, and support all the communities around the world.

Over the years, we stratified a pile of workarounds on top of our `Text` core provider, docs, and unofficial guides with different levels of success. But, was not straightforward, not well integrated, and our users were expecting these features just out-of-the box.

The `TextLayout` core provider will offer an abstraction that can be used as a bridge between the `TextLayout` implementation and the widgets that need to lay out a text.

A `TextLayout` implementation can decide (and advertise) to not conform to all the methods. This will make it possible for our users to keep using non-fully compliant implementations if the fully compliant one is too huge for their use case, or not available on their target platform.

ATM (will update this issue as things evolve) , we're investigating using `Skia/SkParagraph` as the default `TextLayout` implementation.

bambier:
I was walking around creating simple text on picture using [Skia](https://pypi.org/project/skia-python/). It doesn't seems like it fix the RTL languages problems by it self! Actually the thing that fixes the problem is [Harfbazz](https://github.com/harfbuzz).
There is [python binding lib for Harfbazz](https://github.com/harfbuzz/uharfbuzz) that let us to find offsets for words. 
If we want to migrate to new surface just because of RTL language problems, isn't it beter to use Harfbazz insted? In any case we have to use it to fix these problem!

Issue #4401, opened 2016-06-16
Allow text to be selectable/copyable


ryandoherty:
In our application (and I assume many others), having text be selectable/copyable would be a huge benefit to us and our users. 

The benefit of this would be making Kivy apps feel more native to each platform, easier customer support (customers can copy/paste, no more screenshots) and easier development (sometimes I need to copy/paste values displayed in our application).

Understandably the behavior may not be the same for all platforms. For desktop the normal click+drag would work, but for mobile it may be a 'long press' that copies the text to the clipboard or pops up a dialog. I tested out 3 mobile apps on Android, all either copied text immediately on long press or popped up a custom menu which had 'Copy Text' as an option. And on iOS some apps don't even have the ability to select text, but some do with a long press.

tshirtman:
I don't think it should be part of the default Label class, but mixin Label + ButtonBehavior if you want the "mobile" behavior, or using a themed and disabled TextInput if you want the desktop behavior, could be a way to achieve that, you could even make two implementations and decide which to return at runtime, depending on kivy.utils.platform, it could be a useful addition to kivy if you find it works nicely :).

ryandoherty:
I was thinking of it being an attribute that can be set on the Label class, or maybe a new class for text that a developer wants or assumes would need to be selectable. For some elements having selectable text may not be possible or wanted (especially UI like buttons or links, this may really screw things up). 

Using a themed and disabled TextInput seems like an abuse of TextInput and could cause problems if you have lots of text or need some styling or layout that is available via Label. 

For most applications and webpages, the default assumption of users is if something is text, they can select it and copy it. I'm thinking that the same expectations exist for developers, so either making Labels selectable/copyable by default or via an attribute would be the simplest to understand and learn.

Kovak:
Just did a quick check of several google applications on Android, the majority of basic text (equivalent of Kivy's label) does not default to selectable in any form.

This includes:
1. The entire text of an article in Play Newsstand.
2. Elements in gmail that are not the body of an email.
3. Descriptions for Albums, Artists in Play Music

Perhaps a more minimally styled TextInput is the most appropriate solution for application text that wants to be selectable.

ryandoherty:
Can you style text in a TextInput? Text that's in a Label and styled is a high candidate for needing to be selectable and copyable. I really think re-using TextInput isn't the best decision and is being suggested because it's currently (just barely) possible to do so.

[...]

PR #9278, opened 2026-03-13
Add selectable HTML widgets with link handling and docs

davidguigui29:
## Summary
- add new kivy.uix.selectable_html module with:
  - SelectableBodyTextInput (readonly selection/copy behavior)
  - SelectableMarkupLabel
  - SelectableHtmlLabel
  - HTML->text/markup helpers and link open/copy helpers
- register new widgets in kivy/factory_registers.py
- add runnable example examples/widgets/selectable_html.py
- add minimal API page doc/sources/api-kivy.uix.selectable_html.rst and include this file in git tracking

## Validation
- python3 -m py_compile kivy/uix/selectable_html.py examples/widgets/selectable_html.py
- local run check: python3 examples/widgets/selectable_html.py (window provider initializes and app main loop starts)


kivy/factory_registers.py
```diff
@@ -157,6 +157,9 @@
 r('TabbedPanel', module='kivy.uix.tabbedpanel')
 r('TabbedPanelHeader', module='kivy.uix.tabbedpanel')
 r('TextInput', module='kivy.uix.textinput')
+r('SelectableBodyTextInput', module='kivy.uix.selectable_html')
+r('SelectableMarkupLabel', module='kivy.uix.selectable_html')
+r('SelectableHtmlLabel', module='kivy.uix.selectable_html')
```
kuzeyron, line 162: Would be neater if you manually import this.

.gitignore
```diff
@@ -14,6 +14,7 @@ build/*
 doc/build
 doc/autobuild.py-done*
 doc/sources/api-*.rst
+!doc/sources/api-kivy.uix.selectable_html.rst
```
kuzeyron, line 17: It should be ignored like the rest.

kuzeyron (changes requested):
Couple of suggestions from my end.

ElliotGarbus:
Took a look at `kivy/uix/selectable_html.py` in detail. A few things I think are worth resolving before merge, roughly in order of how blocking I think they are:

## 1. Regex-based HTML parsing

`html_to_text` and `html_to_kivy_markup_with_links` parse HTML entirely through chained regex substitutions — there's no real parser involved, not even stdlib `html.parser.HTMLParser`. The anchor-tag regex in particular:

```python
r'(?is)<a\b[^>]*href=(["\'])(?P<href>.*?)\1[^>]*>(?P<label>.*?)</a>'
```

won't handle nested tags inside the label correctly, and non-greedy matching against real-world malformed/deeply-nested HTML is a well-known source of subtle corruption. This tends to work fine in demos and then falls over on real content. Given the whole point of the widget is rendering arbitrary HTML, I think this needs to be rebuilt on a real parser (or at minimum get a much stronger test suite that throws adversarial/malformed markup at it) before it's core.

## 2. Scope — this looks purpose-built for an email client, not a general widget

`normalized_sender_html` specifically unwraps `<iframe srcdoc="...">` content, and `SelectableHtmlLabel`'s fallback string is `"HTML message content unavailable"`. That's a pretty specific origin showing through in what's pitched as a general-purpose widget. Is this a webmail-viewer component that should live as a separate package / kivy-garden widget? I'd lean toward core not carrying app-specific baggage like srcdoc-unwrapping.

## 3. No scheme allowlist before `webbrowser.open`

`resolve_link_url` only special-cases `https?|mailto|tel` and `//`-relative URLs — anything else (`javascript:`, `file:`, a registered custom URI scheme) falls through unchanged and goes straight to `webbrowser.open()`. Combined with `auto_open_links` defaulting to `True`, that means links in untrusted HTML content can be auto-opened with no scheme restriction. I'd want an explicit allowlist (http/https, maybe mailto) and would suggest defaulting `auto_open_links` to `False`.

## 4. Heavy reliance on TextInput's private internals - that are likely to change

`SelectableBodyTextInput` and the bubble-patching logic reach into `_bubble`, `_handle_left`/`_handle_right`/`_handle_middle`, `but_copy`, `but_selectall`, and override `_handle_pressed`/`_handle_released`/`_show_cut_copy_paste` — all private, undocumented `TextInput` implementation details. This is wrapped in a lot of blanket `except Exception: pass` guards, which reads like the fragility is already known and being defensively swallowed rather than surfaced. 

`TextInput` is scheduled to get a complete rearchitected as part of the Kivy 3.0 effort.  see: #8574 and #8575 
The use of `TextInput`'s selection internals will likely break this widget instead of raising a clear error. 

For consideration the new `TextInput` may needs new public hooks to enable something like this widget.

## 5. Test coverage

Given items 1 and 3 are correctness/security concerns rather than style, I'd treat test coverage as a blocking ask rather than a nice-to-have.

---

Net: the selection/copy UX approach is clever, but I wouldn't merge as-is — the regex HTML parser and default-on external link opening are the two I'd want addressed first.

ElliotGarbus:
Stepping back, I think most of the above collapses into one root cause: this bundles two mostly-unrelated problems into one widget.

html_to_kivy_markup, html_to_kivy_markup_with_links, html_to_text, resolve_link_url, and open_external_link are all pure functions with no dependency on the selection machinery. Ship those as a standalone utility module and link handling already works for free — Label(markup=True, text=html_to_kivy_markup(raw)) gets you rendering plus clickable links via Label's existing on_ref_press, no new widget class required.

The actual hard problem is selectable text on a Label, which Kivy has no native support for. That's what forces everything else in this PR — the hidden invisible TextInput overlay, the pixel-syncing across three widget instances, and the reaching into TextInput's private _bubble/_handle_left/but_copy internals wrapped in blanket except Exception: pass. None of that is actually about HTML — it's a "make Label selectable" problem. And it's not a new problem: #4401 has been asking for selectable/copyable text since 2016, entirely independent of HTML rendering.

## Proposed split:

Land the HTML→markup/text utilities (with the parser and link-scheme concerns above still addressed) as a small, easily-tested module. This is the broadly reusable, low-risk part.

Drop SelectableBodyTextInput/SelectableMarkupLabel/SelectableHtmlLabel from this PR. If selectable rich text is wanted in core, it deserves its own design discussion against #4401 — possibly as a real feature on Label/TextInput rather than a widget that fakes it by gluing two widgets together with private-internal coupling.  KivyMD's allow_selection/allow_copy in Label is a possible reference implementation.

PR #9281, opened 2026-03-18
Add unit tests for `kivy.uix.behaviors.hover` module

gottadiveintopython:
Maintainer merge checklist

This PR should be merged after #9248.


misl6:
@gottadiveintopython , can you please rebase on top of master ? Thank you again!

gottadiveintopython:
I have no idea why segmentation faults are occurring in the macOS CI tests.

misl6:
> I have no idea why segmentation faults are occurring in the macOS CI tests.

I'm wondering if we have a leak somewhere (in tests or kivy itself), can you try to gradually increase / reduce the tests and use module level fixtures like https://github.com/kivy/kivy/blob/master/kivy/tests/test_thorvg_svg_provider.py ?

FYI: I cannot reproduce it locally, but it might be we're seeing it due to some memory limitations on GitHub runners.

gottadiveintopython:
> use module level fixtures

The `kivy_init` fixture is defined with `scope="function"`, which means all fixtures that depend on it also have to be defined with `scope="function"`. (A fixture cannot outlive a fixture it depends on.) So none of the remaining fixtures can be widened to module scope.

gottadiveintopython:
The `kivy_init` fixture alone doesn't cause any errors, as shown by the latest commit's CI tests. However, when it is used together with the `advance_frames` helper, a segmentation fault occurs, as shown by the CI tests for the second latest commit.

They were added in #9133, but this pull request is the first time they have been used in Kivy's unit tests, which is probably why the issue hasn't been discovered until now.

gottadiveintopython:
@misl6 I probably don't understand the unit test setup very well, but considering that the base class of `GraphicUnitTest` is defined as follows:

https://github.com/kivy/kivy/blob/176f96f6430e2061dfbe73aef0d6afe9d45a258b/kivy/tests/common.py#L32-L36
https://github.com/kivy/kivy/blob/176f96f6430e2061dfbe73aef0d6afe9d45a258b/kivy/tests/common.py#L146

I did the same for the `kivy_init` fixture. In other words, I avoided the segmentation faults by skipping the unit tests on macOS CI. Let me know if this approach is undesirable.

### Note

The following code alone causes a segmentation fault, as shown by the CI tests for `5d65b5f`:

```python
from kivy.tests.common import GraphicUnitTest

class Test(GraphicUnitTest):
    def test_advance_frames(self):
        self.advance_frames(1)
```

Likewise, the following code alone without the changes introduced by this PR also causes a segmentation fault, as shown by the CI tests for `195cbbd`:

```python
from kivy.tests.common import advance_frames

def test_kivy_init(kivy_init):
    advance_frames(1)
```

PR #9293, opened 2026-04-05
Add Lottie animation provider and LottieWidget

ElliotGarbus:
Adds support for Lottie animations via a new core provider and UIX widget.

- `kivy/core/lottie/` — `LottieBase` abstract provider with a ThorVG backend (`lottie_thorvg.py`) and a null fallback. Exposes frame-accurate playback, markers, named slots (colour/opacity overrides), and optional quality control.
- `kivy/uix/lottie.py` — `LottieWidget` backed by the provider. API mirrors `kivy.uix.video.Video`: `source`, `state` (play/pause/stop), `position`, `progress`, `duration`, `loaded`, `loop`, `volume`. Slot overrides via `apply_slot()` / `clear_slot()`.
- `kivy/tests/test_lottie_provider.py` and `test_uix_lottie.py` — unit and graphic tests covering core behaviour.

This work is part 3 of a 3 part plan for vector graphics:

- Add an svg image provider, done see: https://github.com/kivy/kivy/pull/9273
- Add an svg widget that allows some DOM manipulation and scales up and down. https://github.com/kivy/kivy/pull/9284
- Add a Lottie provider, backend and widget. This PR

## Stack

This PR is part of set of PRs that use ThorVG:

- **#9297** — `kivy.lib.thorvg` Cython wrapper + desktop CI + bundled ThorVG v1.0.4 *(base of the stack — merge first)*
- **#9273** — SVG image-loader migration (`img_thorvg_svg`)
- **#9284** — SVG widget migration (`svg_thorvg`) + zero-copy render *(this PR)*
- **#9293** — Lottie provider migration (`lottie_thorvg`) + zero-copy per-frame blit

All three consumer PRs are rebased on `feature/thorvg-cython`, so their diffs currently include the wrapper commits; those will drop out automatically once #9297 merges. The three consumer PRs are independent of each other and can land in any order after #9297.

Maintainer merge checklist
* [x] Title is descriptive/clear for inclusion in release notes.
* [x] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [x] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #9284, opened 2026-03-19
SvgWidget, AsyncSvgWidget with SVG Provider backed by ThorVG 

ElliotGarbus:
## Summary

- Adds `kivy.core.svg` — a new SVG core provider with a ThorVG backend (`svg_thorvg.py`) that rasterizes SVGs to Kivy textures with supersampling, power-of-2 bucket snapping, and async loading support.
- Adds `kivy.uix.svg` — `SvgWidget` and `AsyncSvgWidget` with fit modes, layer overrides, `current_color` support.
- Registers both widgets in the factory and adds a canvas style rule in `style.kv`.
- Adds four interactive example apps: `svg_layers.py`, `svg_scale.py`, `svg_download.py`, and `icon_toggle_button.py`.

This work is part 2 of a 3 part plan for vector graphics:
1) Add an svg image provider, done see: https://github.com/kivy/kivy/pull/9273
2) Add an svg widget that allows some DOM manipulation and scales up and down.  This PR
3) Add a Lottie provider, backend and widget.  TBD

This also relates to the kivy3.0 plan: https://github.com/kivy/kivy/issues/8573

Open Issue: Do we want to deprecate the existing SVG support that was implemented as a graphics instruction.  This never left the "highly experimental state", or simply delete it.  https://kivy.org/doc/stable/api-kivy.graphics.svg.html#module-kivy.graphics.svg

## Stack

This PR is part of set of PRs that use ThorVG:

- **#9297** — `kivy.lib.thorvg` Cython wrapper + desktop CI + bundled ThorVG v1.0.4 *(base of the stack — merge first)*
- **#9273** — SVG image-loader migration (`img_thorvg_svg`)
- **#9284** — SVG widget migration (`svg_thorvg`) + zero-copy render *(this PR)*
- **#9293** — Lottie provider migration (`lottie_thorvg`) + zero-copy per-frame blit

All three consumer PRs are rebased on `feature/thorvg-cython`, so their diffs currently include the wrapper commits; those will drop out automatically once #9297 merges. The three consumer PRs are independent of each other and can land in any order after #9297.

Maintainer merge checklist
* [x] Title is descriptive/clear for inclusion in release notes.
* [x] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [x] Added to the milestone version it was merged into.
* [x] **Unittests** are included in PR.
[...]

PR #9297, opened 2026-04-20
Add internal ThorVG Cython wrapper (kivy.lib.thorvg)

ElliotGarbus:
Introduce kivy/lib/thorvg/ as an internal Cython binding to ThorVG v1.0.4, in preparation for replacing the external `thorvg-python` dependency used by the unmerged SVG / SVG-image / Lottie feature branches. ThorVG is linked statically on all desktop platforms; no separate kivy-deps wheel is shipped.

Wrapper surface (minimal - only what the three providers need):

* Paint / Picture with ownership-aware __dealloc__ and set_opacity / id
* SwCanvas with zero-copy Python buffer protocol for GPU upload (Texture.blit_buffer consumes this directly - see bench below)
* Accessor for SVG element-id traversal using ThorVG 1.0.4's native accessible-mode API (removes the XML-fallback the feature branches previously carried)
* LottieAnimation with frame / segment / marker / slot / quality controls
* Engine initialised exactly once on module import (SW, threads=0)

Build wiring (setup.py):

* c_options['use_thorvg'], auto-enabled if headers+lib are discoverable
* determine_thorvg_flags() with KIVY_THORVG_{INCLUDE,LIB}_DIR env-var overrides and KIVY_DEPS_ROOT/dist fallback
* TVG_STATIC defined, C++ runtime linked on darwin / unix; compiled as C++ (language='c++')

Tests + benchmarks (skip-if-unavailable, no SDL3/GStreamer needed):

* kivy/tests/test_thorvg_wrapper.py - 23 unit tests covering engine version, canvas / picture / accessor / animation lifecycle, buffer allocation, ownership transfer on Canvas.add, and callback exception safety
* kivy/tests/bench_thorvg_wrapper.py - 6 micro-benchmarks; crucially a matched pair (memoryview vs bytes) that makes the ~45x zero-copy win of the buffer protocol visible and regression-guards against providers falling back to bytes(canvas) on the hot path

Lightweight CI gate (.github/workflows/test_thorvg_wrapper.yml) + helpers (.ci/thorvg_wrapper_*.py): builds ThorVG v1.0.4 statically with Meson/Ninja, compiles only the wrapper extension, runs the unit tests and benchmarks on Ubuntu / macOS / Windows. Bootstraps empty kivy / kivy.lib / kivy.logger stubs so tests can import kivy.lib.thorvg without a full pip install -e .

## Stack

This PR is part of set of PRs that use ThorVG:

- **#9297** — `kivy.lib.thorvg` Cython wrapper + desktop CI + bundled ThorVG v1.0.4 *(base of the stack — merge first)*
- **#9273** — SVG image-loader migration (`img_thorvg_svg`)
- **#9284** — SVG widget migration (`svg_thorvg`) + zero-copy render *(this PR)*
- **#9293** — Lottie provider migration (`lottie_thorvg`) + zero-copy per-frame blit

All three consumer PRs are rebased on `feature/thorvg-cython`, so their diffs currently include the wrapper commits; those will drop out automatically once #9297 merges. The three consumer PRs are independent of each other and can land in any order after #9297.

Maintainer merge checklist


setup.py
```diff
@@ -857,6 +858,62 @@ def determine_angle_flags():
     return flags
 
 
+def determine_thorvg_flags():
+    """Flags to build and statically link the internal kivy.lib.thorvg
+    Cython wrapper against ThorVG v1.0.4.
```
misl6: Let's avoid to talk about a specific ThorVG version, version is managed in the build tools (tools/...)

setup.py
```diff
@@ -857,6 +858,62 @@ def determine_angle_flags():
...
+
+    Header + static library lookup order:
+
+    1. ``KIVY_THORVG_INCLUDE_DIR`` / ``KIVY_THORVG_LIB_DIR`` environment
+       variables (set by mobile recipes - p4a, kivy-ios - and by users with
```
misl6, line 871: Let's consider `kivy-ios` as something deprecated, how about adding the custom script to build thorvg in build_ios_dependencies.sh ?

setup.py
```diff
@@ -1334,6 +1391,47 @@ def determine_sdl3():
...
+        exists(join(inc, 'thorvg_capi.h'))
+        for inc in thorvg_flags['include_dirs']
+    )
+    if header_found:
+        print('ThorVG headers detected - building kivy.lib.thorvg wrapper')
+        c_options['use_thorvg'] = True
+    else:
+        print(
+            'ThorVG headers not found; skipping kivy.lib.thorvg wrapper. '
+            'Set KIVY_THORVG_INCLUDE_DIR / KIVY_THORVG_LIB_DIR or '
```
misl6: Let's also include KIVY_DEPS_ROOT in the error explaination

misl6 (commented):
Nice!!!

Left some initial comments on the build scripts, as we need to think about a plan for iOS (and Android soon, as we will leave the old "recipe" approach to the past)

setup.py
ElliotGarbus: Got it.

setup.py
ElliotGarbus: Will do.  I'll be starting on build for mobile later today.

setup.py
ElliotGarbus: Will do!

ElliotGarbus:
> Nice!!!
> 
> Left some initial comments on the build scripts, as we need to think about a plan for iOS (and Android soon, as we will leave the old "recipe" approach to the past)

I was going to start on Android and IOS later today.  Any guidance on the proper approach would be appreciated.  Should I create a .xcFramework for IOS?  What about Android?

misl6:
> > Nice!!!
> > Left some initial comments on the build scripts, as we need to think about a plan for iOS (and Android soon, as we will leave the old "recipe" approach to the past)
> 
> I was going to start on Android and IOS later today. Any guidance on the proper approach would be appreciated. Should I create a .xcFramework for IOS? What about Android?

Yes, for iOS an .xcframework is great (and you can think about linking it dynamically like SDL3 and ANGLE).

For other platforms (mainly for Linux and macOS) is statically linking it something strictly required or we can think about dynamic linking? (delocate for macOS and its counterpart for Linux will take care of including it properly in the wheel).

For Android, better ask @T-Dynamos. What we can do here on kivy repository to get ready for an Android wheel so we do not have to maintain it separately in python-for-android?

ElliotGarbus:
@misl6 
Thanks for the feedback. Proceeding as follows:

**iOS** — one PR against kivy/kivy:
- extend `tools/build_ios_dependencies.sh` to emit `ThorVG.xcframework`, matching SDL3/ANGLE. 

**Desktop plan — dynamic on Linux + macOS, static only on Windows**
Before any of the four open PRs merge (#9297 wrapper, #9273, #9284, #9293), I'll rework #9297; 

- **Linux**: integrate ThorVG into tools/build_linux_dependencies.sh as a shared library in dist/lib; auditwheel bundles it into the wheel.
- **macOS**: integrate ThorVG into tools/build_macos_dependencies.sh. Build per-arch via Meson, lipo into universal2, package as ThorVG.framework in dist/Frameworks; delocate bundles it.

- **Windows**: stays static. (open to feedback)

**Android** — deferring until @T-Dynamos weighs in happy to draft a design doc or just
get guidance and implement, whichever is more useful.

T-Dynamos:
p4a already has a `libthorvg` PR here: https://github.com/kivy/python-for-android/pull/3288 .

I can update kivy recipe to include `libthorvg` in dependencies and dynamically link it.

Also latest p4a supports building wheels, we can write minimal script which generates a kivy wheel.

For example: https://github.com/T-Dynamos/p4a-wheels/blob/main/recipebuild.py (only ~100 lines)

P.S.: I already host kivy wheels: https://anshdadwal.is-a.dev/p4a-wheels/p4a/kivy/ (not a production ready environment)

setup.py
ElliotGarbus: done

setup.py
ElliotGarbus: done

ElliotGarbus:
Thanks @T-Dynamos — really appreciate the offer, and I want to align before we converge on next steps. Android packaging isn't an area I have any expertise in, so I'd rather defer to you on the right approach than propose one.

Since this work targets the Kivy 3.0 release rather than a near-term patch, do we want to update the p4a recipe or does it make more sense to aim straight at a PEP 738 wheel? I'm not sure what the right tradeoff is given p4a's modernization trajectory.

Happy to follow your lead on whatever direction makes sense.

ElliotGarbus:
@T-Dynamos 
Quick follow-up on my earlier question, since [#9301](https://github.com/kivy/kivy/pull/9301) just landed — that PR adds an iOS wheel build directly in kivy/kivy via cibuildwheel --platform ios, with tools/build_ios_dependencies.sh cross-compiling ThorVG into a KivyThorVG.xcframework that gets bundled into the iOS wheel. No kivy-ios recipe involved.

I'm wondering whether the structurally similar approach might make sense for Android: a cibuildwheel --platform android job in kivy/kivy, with tools/build_android_dependencies.sh cross-compiling libthorvg per-ABI for bundling into the kivy Android wheel. (But Android isn't iOS - so I'm asking)

Would that direction make sense to you, or are there Android-specific reasons you'd want a different shape (your [recipebuild.py](https://github.com/T-Dynamos/p4a-wheels/blob/main/recipebuild.py) pipeline, or keeping more of the build on the p4a side)? 

Happy to follow your lead on any of this.

T-Dynamos:
One thing worth clarifying first is that on Android we are currently dynamically linking against the ThorVG wheel (to minimize app size if some other dependency also needs ThorVG) and expecting that dependency to be provided separately.

I am not fully sure that using cibuildwheel gives much benefit.

On Android, Kivy does not really work outside the `p4a` runtime environment, so the actual integration point is still `p4a`.

My main concern is mostly duplicated maintenance. If we add Android wheel builds using `cibuildwheel`, then we will probably end up maintaining native build logic in two different places:

* `p4a` recipe 
* Android wheel build scripts in this repo

So personally I would lean more toward keeping Android build flow centered around `p4a`, unless there is some strong reason to support standalone Android wheels outside the `p4a` ecosystem.

ElliotGarbus:
Thanks @T-Dynamos for the detailed response. 

I want to flag that I'm reading some tension between this proposal and @misl6's earlier comment about "leaving the recipe approach to the past" and building the Android wheel in this repo. I don't have the expertise to adjudicate, and I think the two of you have different — both reasonable — reads on the destination architecture. 

@misl6, can you weigh in on direction so we know which path to take? Happy to follow whichever you confirm.

@T-Dynamos Kind of nit - but you called out "if some other dependency also needs ThorVG", I want to be clear that I did not build a complete wrapper for ThorVG, but a limited wrapper that only meets the needs of the Image, SVG and Lottie providers.  I did this with the thought of minimizing maintenance.

setup.py
ElliotGarbus: Done

ElliotGarbus:
@T-Dynamos
I’m planning to extend tools/build_thorvg.sh (which already builds for macOS/Linux/Windows/iOS) to support Android NDK builds. The script would produce a libKivyThorVG.so per ABI (since the wrapper is Kivy-specific rather than a full ThorVG binding).

If it works for you, your recipe could call into it rather than duplicating the ThorVG build logic.

Does that sound reasonable?

T-Dynamos:
@ElliotGarbus 

I will recommended you to not build thorvg for android in that script just leave room for env variables -  if build type is android expect them.
After this PR is merged I will enable kivy recipe to work with libthorvg in p4a.
After that I will send one more PR in this repo which enables building android wheel.

(*reply from Discord*)

ElliotGarbus:
@T-Dynamos Thank you!

misl6:
Hi everyone, and thank you for the constant feedback that has allowed me to join the discussion even a few weeks later.

I will review this PR as-is, and then iterate over the iOS one once this is merged (and other related ones to follow).

Regarding the Android discussion: I think we should align with Python standards, ( in this case PEP 738 ) and follow the official path.

PEP 738 ("Adding Android as a supported platform") was accepted for CPython 3.13, establishing Android as a Tier 3 supported platform with proper packaging conventions, standard wheel tags on PyPI, and integration with tools like `cibuildwheel`. Following this path means being open to the broader Python ecosystem rather than staying in a separate niche that strictly requires our own tooling.

python-for-android will not cease to exist, but its role should evolve toward helping users package their app together with the bootstrapper in a fast and convenient way — not building what can be produced and verified separately in a standard, trusted workflow.

This approach is what can help Kivy grow and keep going not only in the near future, but also ten years from now.

But, let's discuss this in a separate thread ! 😀

misl6 (approved):
LGTM. Thank you!

Might I ask you in separate PR to add some references to the installation docs?

(I usually prefer to follow what the CI does, but others might love a reference of extra build deps (meson, etc) in the docs)

PR #9186, opened 2025-10-21
Use `httpbingo.org` instead of `httpbin.org` for urlrequest tests

misl6:
Maintainer merge checklist
* [ ] Title is descriptive/clear for inclusion in release notes.
* [ ] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [ ] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #9139, opened 2025-09-08
:white_check_mark: Improve urlrequest test reliability

AndreMiras:
UrlRequest.is_finished can flip true in the worker thread before Kivy posts the terminal callback to the Clock.
The tests stopped on is_finished and sometimes asserted while the last queued event was still "progress", leading to flaky failures.

The error was:
```
FAILED kivy/tests/test_urlrequest/test_urlrequest_urllib.py::test_auth_header - AssertionError: assert 'progress' in ('success', 'redirect')
FAILED kivy/tests/test_urlrequest/test_urlrequest_urllib.py::test_auth_auto - AssertionError: assert 'progress' in ('success', 'redirect')
FAILED kivy/tests/test_urlrequest/test_urlrequest_urllib.py::test_ca_file[http] - AssertionError: assert 'progress' in ('success', 'redirect')
FAILED kivy/tests/test_urlrequest/test_urlrequest_urllib.py::test_ca_file[https] - AssertionError: assert 'progress' in ('success', 'redirect')
```

Maintainer merge checklist
* [x] Title is descriptive/clear for inclusion in release notes.
* [x] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [x] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #9304, opened 2026-05-06
Tests: replace httpbin with local pytest-httpserver for UrlRequest

ElliotGarbus:
## Summary

The `UrlRequestUrllib` tests have been intermittently flaky in CI. Two
prior PRs identified the two distinct failure modes but neither landed:

- [#9139 — :white_check_mark: Improve urlrequest test reliability](https://github.com/kivy/kivy/pull/9139)
  (DRAFT) diagnosed the `is_finished` race: `UrlRequest.is_finished`
  can be observed `True` before the Clock-dispatched terminal callback
  has actually delivered every queued event, so the queue's last entry
  could still be `'progress'` when the test ran its assertions.
- [#9186 — Use httpbingo.org instead of httpbin.org](https://github.com/kivy/kivy/pull/9186)
  (OPEN) tried to swap the external service. As the author noted in
  the comments, this didn't really fix the issue — the tests were
  still tied to a third-party host.

This PR closes out both threads by making the suite self-contained and
race-free.

### Changes

- Replace all `httpbin.org` URLs with routes registered on a local
  `pytest-httpserver` instance. Tests no longer depend on the public
  internet at all.
- Add a `httpsserver` fixture (in a new `kivy/tests/test_urlrequest/conftest.py`)
  backed by a `trustme`-issued CA, so `test_ca_file[https]` validates a
  real certificate against `UrlRequest`'s `ca_file=` parameter.
- Replace `wait_request_is_finished` with `wait_for_terminal_event`,
  which waits on the test callback queue's last entry being one of
  `success`/`redirect`/`error`/`failure`. This is what the assertions
  actually depend on, so it removes the race that #9139 spotted.
- Drop the `nonetwork`/`NONETWORK` skip markers and the `errno == 11001`
  skip dance — there is no external network to fail to.
- Add `@pytest.mark.timeout(30)` per test so a wedged worker thread
  cannot hang CI.
- Add `pytest-httpserver` and `trustme` to the `dev` extras in
  `setup.cfg`.
- Tests: expand UrlRequest urllib coverage to all callbacks and key options

The `UrlRequestRequests` test file is untouched: it already uses
`responses` to mock at the requests-library layer.

Closes #9186.
Supersedes #9139.

Maintainer merge checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,8 +34,8 @@ dev =
     pytest_asyncio!=0.11.0
     pytest-timeout
     pytest-benchmark
-    pytest-httpserver
-    trustme
+    pytest-httpserver==1.1.5
+    trustme==1.2.1
     pyinstaller
     sphinx~=6.2.1
     sphinxcontrib-jquery~=4.1
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,6 +34,8 @@ dev =
     pytest_asyncio!=0.11.0
     pytest-timeout
     pytest-benchmark
+    pytest-httpserver
+    trustme
     pyinstaller
     sphinx~=6.2.1
     sphinxcontrib-jquery~=4.1
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ dev =
     pytest_asyncio!=0.11.0
     pytest-timeout
     pytest-benchmark
-    pytest-httpserver==1.1.5
+    pytest-httpserver==1.1.3
     trustme==1.2.1
     pyinstaller
     sphinx~=6.2.1
--- a/kivy/tests/test_urlrequest/conftest.py
+++ b/kivy/tests/test_urlrequest/conftest.py
@@ -9,13 +9,49 @@ via the ``ca_file=`` parameter.
 
 Two server fixtures are exposed:
 
-* ``httpserver`` (provided by ``pytest-httpserver``): plain HTTP.
+* ``httpserver`` (provided by ``pytest-httpserver``): plain HTTP, bound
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -34,6 +34,8 @@ dev =
     pytest_asyncio!=0.11.0
     pytest-timeout
     pytest-benchmark
+    pytest-httpserver
+    trustme
```
misl6: Can I ask you (mainly for starting trusting our SBOM) to pin a specific version?

The dependency bot will take care of submitting update PRs.

For the other dependencies apart from these two, do not worry, we will take care of them in future.

misl6 (changes requested):

setup.cfg
ElliotGarbus: Got it.  I will complete shortly.

ElliotGarbus: Pinned `pytest-httpserver==1.1.3` and `trustme==1.2.1` in b5ba412. Future updates can flow through Dependabot.

ElliotGarbus: pinned to 1.1.3 because 1.1.4+ require Python 3.10... I'll post an issue on dropping Python 3.9 support

ElliotGarbus:
FYI on the `pytest-httpserver==1.1.3` pin: 1.1.4+ require Python `>=3.10`, but Kivy still supports 3.9. Filed #9305 to track dropping 3.9 (EOL since Oct 2025). Once that lands, this pin can move to the latest release.

ElliotGarbus:
@misl6 Changes made, Ready for review

misl6 (approved):
LGTM. Thank you!

PR #9312, opened 2026-05-15
[WIP] Optimize `glReadPixels` helper

gottadiveintopython:
Maintainer merge checklist

The current implementation of `glReadPixels` performs two memory allocations: one for internal use and one for the return value. Removing the former improves performance. Here is the code I used to benchmark it:

```python
from functools import partial

Config.set("graphics", "width", 800)
Config.set("graphics", "height", 600)
from kivy.app import App
from kivy.lang import Builder


KV = '''
Button:
    text: "timeit glReadPixels"
    font_size: 50
    on_release: app.timeit_glReadPixels()
'''


class TestApp(App):
    def build(self):

    def timeit_glReadPixels(self):
        from timeit import timeit
        from kivy.graphics.opengl import (
            glReadPixels, GL_RGBA, GL_RGB, GL_UNSIGNED_BYTE,
        )

        f = partial(glReadPixels, 0, 0, *self.root.size, GL_RGBA, GL_UNSIGNED_BYTE)
        print("GL_RGBA:", timeit(f, number=1000))
        f = partial(glReadPixels, 0, 0, *self.root.size, GL_RGB, GL_UNSIGNED_BYTE)
        print("GL_RGB: ", timeit(f, number=1000))


if __name__ == "__main__":
    TestApp().run()
```

Results on my Linux desktop:

- The `GL_RGB` format becomes 18% faster (from 5.7 seconds to 4.7 seconds).
- The `GL_RGBA` format becomes 87% faster (from 3.6 seconds to 0.5 seconds).

### DIFF
--- a/kivy/graphics/opengl.pyx
+++ b/kivy/graphics/opengl.pyx
@@ -1158,31 +1158,26 @@ def glReadPixels(GLint x, GLint y, GLsizei width, GLsizei height, GLenum format,
 
     We support only GL_RGB/GL_RGBA as a format and GL_UNSIGNED_BYTE as a
     type.
+
+    .. versionchanged:: 3.0.0
+       The return value is now a :class:`bytearray` instead of a :class:`bytes` object.
     '''
     assert format in (GL_RGB, GL_RGBA)
     assert type == GL_UNSIGNED_BYTE
 
-    cdef object py_pixels = None
     cdef long size
-    cdef char *data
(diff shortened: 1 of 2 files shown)


gottadiveintopython:
Since a more efficient API was added in #9154, the motivation for this PR has diminished. I'd rather not change the existing API and risk accidentally introducing a bug.

PR #9314, opened 2026-05-15
Video: AVFoundation provider (macOS+IOS) + thumbnails, buffering, options, cross-platform demo

ElliotGarbus:
`VideoBase` has been extended with three new capabilities — thumbnails, a buffering signal, and provider-specific options — and a native video provider for macOS and IOS built on AVFoundation is included as the first implementation. 

---

### Core (`kivy/core/video/__init__.py`, `kivy/uix/video.py`)

- **`VideoBase.generate_thumbnail(source, time, size)`** — class-method API for extracting a still frame from any video file. Providers that support it return a `Texture`; others return `None`.
- **`VideoBase.buffering`** / **`on_buffering`** — minimal boolean signal covering both the initial pre-playback wait and mid-stream rebuffering stalls. Mapped to `Video.buffering` on the widget.
- **`VideoBase.options`** / **`Video.options`** (`DictProperty(allownone=True)`) — opaque provider-specific dict forwarded at load time. Providers document their supported keys; unknown keys are warned and ignored.
- **`_fps` added to `VideoBase.__slots__`** — fixes a latent `AttributeError` on instantiation.
- **`Texture.__init__` gains `nofree=False`** — allows external systems (e.g. a native video provider) to own the GL texture lifetime and prevent Kivy from queueing a redundant `glDeleteTextures`.

### AVFoundation provider (new — macOS)

- **Zero-copy pixel pipeline**: `IOSurface → ANGLE EGL pbuffer → GL_TEXTURE_2D`. Requires Kivy 3.0's ANGLE/Metal GL backend. If the zero-copy probe fails at runtime the player dispatches `on_eos` with a logged diagnostic rather than silently falling back.
- **Buffering signal** driven by `AVPlayer.timeControlStatus == .waitingToPlayAtSpecifiedRate`.
- **Thumbnail generation** via `AVAssetImageGenerator`; BGRA→RGBA swizzle performed in Cython so `glTexImage2D` sees a valid format on ANGLE.
- **`options` dict** — supported keys:
  - `automatically_waits_to_minimize_stalling` (`bool`, default `True`) — forwarded directly to `AVPlayer`.
- **Provider registration** — `avfoundation` is inserted as the first candidate for `video` on Apple platforms in `kivy/core/__init__.py`.

### Build (`setup.py`)

- Detects `use_avfoundation` and compiles `video_avfoundation.pyx` + `video_avfoundation_implem.mm` with ANGLE link flags (hard dependency on Kivy 3.0's ANGLE backend for Apple platforms).

### Tests

- `kivy/tests/test_video.py` — cross-platform tests for `VideoBase`: `options` round-trip, `buffering` default + `on_buffering` dispatch.
- `kivy/tests/test_video_avfoundation.py` — AVFoundation-specific tests: `automatically_waits_to_minimize_stalling` round-trip, unknown option tolerance, thumbnail generation.

### Examples

- `examples/widgets/video/video_demo.py` — cross-platform demo (H.264/MP4 only) with a 2×2 clip-card home grid, thumbnail previews, buffering overlay, live-scrub seek bar (`SeekSlider` with pause-on-drag + keyframe seek during drag + precise seek on release), and a live-state readout panel. Remote clips cached to `App.user_cache_dir` via `UrlRequest`.

---

Maintainer merge checklist


ElliotGarbus:
I'm making this a WIP.  I am adding IOS to this PR.  Please feel free to review the other changes.

kivy/core/video/__init__.py
```diff
@@ -74,11 +105,13 @@ def __init__(self, **kwargs):
         self._texture = None
         self._volume = 1.
         self._state = ''
+        self._buffering = False
 
         self._autoplay = kwargs.get('autoplay')
         self._async = kwargs.get('async')
         self.eos = kwargs.get('eos')
         self._fps = kwargs.get('fps')
+        self._options = dict(kwargs.get('options') or {})
```
kuzeyron: `self._options = dict(kwargs.get('options', {}))`

ElliotGarbus: I'll get that cleaned up later today... that's a bit embarrassing.  Thanks for the feedback!

kivy/core/video/__init__.py
```diff
@@ -155,6 +197,65 @@ def _get_state(self):
...
 
+    def _get_options(self):
+        return self._options
+
+    options = property(lambda self: self._get_options(),
+                       doc='Get the provider-specific options dict that was '
+                           'passed to the provider at construction time. '
+                           'Read-only at the base class level; providers may '
+                           'mutate it during their lifetime. '
+                           '\n\n.. versionadded:: 3.0.0')
```
kuzeyron, line 208: I think you are missing a space here

kuzeyron: I'm not sure if it's correct by having `versionadded`

ElliotGarbus: I'll get it figured out... 
@kuzeyron Great to see you hear - hope your feeling better!

kuzeyron: One day at a time, I'm getting better and better :) Today is the first day to actually go through everything. And hopefully I can start coding again after a long break.

BTW, I think `versionadded` is a value that's being used with the doc-page so it might be correct

kivy/core/video/__init__.py
ElliotGarbus: Looking a little higher in the code, the `or {}` was intentional.
On line 98
`kwargs.setdefault('options', None)`
This puts the key 'options' in the kwargs dict with a value of None if options is not passed in.
Therefore `kwargs.setdefault('options', {})` will never return the empty dict.
`self._options = dict(kwargs.get('options') or {})` does return the empty dict if options is None. 

Looking at the code I do realize the `kwargs.setdefault('options', None)` is redundant.  I could drop this -but it nicely documents the default value for options.   Even if I remove the setdefault(), We will want to keep the `or {}`.  This will catch the case that a user passes in `options=None'

Let me know if you want me to remove the `kwargs.setdefault('options', None)` on line 98.  I would recommend it stays.

kivy/core/video/__init__.py
ElliotGarbus: I have confirmed this is correct.

Glad to hear your getting better.  Wishing you a speedy recovery to 100%!

kivy/core/video/__init__.py
kuzeyron: Can you lookup what the intentions are for the "options"? Would be extremely weird if it's not a dict when it's called from other instances. And change it from None to {} with setdefault. I have only looked at https://kivy.org/doc/master/api-kivy.core.video.html#module-kivy.core.video but I can't see anything related to "options". Just so you wouldn't have to convert it again to a dict format.

Edit: Forgot to check what you had added to the show. Defaulting to {} would be way better.

ElliotGarbus: Done

kivy/core/video/__init__.py
```diff
@@ -74,11 +105,13 @@ def __init__(self, **kwargs):
         self._texture = None
         self._volume = 1.
         self._state = ''
+        self._buffering = False
 
         self._autoplay = kwargs.get('autoplay')
         self._async = kwargs.get('async')
         self.eos = kwargs.get('eos')
         self._fps = kwargs.get('fps')
+        self._options = dict(kwargs['options'])
```
kuzeyron: Without dict() would be nicer as you are now expecting dicts being passed through. You might want to check the type of `_options` somewhere in this code so it throws errors if someone is feeding it with other types than dicts.

ElliotGarbus: That was there to make a copy of the dict - but I know it is only read by the provider, so I'll remove it.  The options attribute is a Kivy DictProperty with allownone=True.

ElliotGarbus: done

kuzeyron: Great! Thanks ^

kivy/core/video/__init__.py
```diff
@@ -44,6 +58,16 @@ class VideoBase(EventDispatcher):
             providers).
         `autoplay`: bool, defaults to False
             Auto play the video on init.
+        `options`: dict, defaults to None
```
kuzeyron: ``options`: dict, defaults to {} or DictProperty()`

kivy/uix/video.py
```diff
@@ -194,11 +294,22 @@ def _do_video_load(self, *largs):
             # Check if filename is not url
             if '://' not in filename:
                 filename = resource_find(filename)
-            self._video = CoreVideo(filename=filename, **self.options)
+            # Forward both the dict (for providers that introspect
+            # self._options) and the unpacked kwargs (back-compat with
+            # apps that put well-known kwargs like 'eos' in options).
+            # ``options`` accepts None as a legal value -- normalize it
+            # to {} here so the rest of the stack only sees a dict.
+            options = self.options if self.options is not None else {}
```
kuzeyron, line 302: Maybe something like this: `options = self.options if isinstance(self.options, (dict, DictProperty)) else {}`. Otherwise we are having mixups between dict and None

Edit: Also having a message if it's empty?! Maybe warn if it resets to {}

kivy/core/video/__init__.py
ElliotGarbus: With the change we made options now defaults to {}.  I'll make the change in the comment.

You may recall that in my original code:
`kwargs.setdefault('options', None)`
We changed that to:
`kwargs.setdefault('options', {})`

kivy/uix/video.py
ElliotGarbus: I don't think I understand... 
DictProperty is the descriptor/property declaration type, not the expected runtime value of self.options. At runtime self.options should be either None or a dict, so `options = self.options if self.options is not None else {}` intentionally distinguishes those two cases.  Therefore `options` is always a dict. If self.options is None options is set to {}. 

If self.options is None options will be set to an empty dict.  This is the desired behavior.
If self.options is not none, it is a dict so options will be a dict.

Am I missing something?

kuzeyron: If options are mixed in each scripts it's going to confuse other devs. Having them to behave the same is what I'm trying to tell you. Otherwise different names that describes it better would be great. In this case I assume `options`are all of the same value/s (addressed or copied).

Regarding the DictProperty there are cases where it's passed on directly. Supporting that would be great. Some of Kivy's properties comes with limitations. At this moment I don't remember which limitations there are. Some cases with DictProperty you have to re-convert its content so it becomes a valid dictionary. I don't know why but it can happen.

In the end, these are what comes to mind. Whatever you do is up to you. I'm just a little worried if there are mixups and confuses other devs in the future.

ElliotGarbus: Got it. I appreciate the objective.  There is an opportunity to improve the comment. 
```
# ``options`` accepts None as a legal value -- normalize it
# to {} here so the rest of the stack only sees a dict.
options = self.options if self.options is not None else {}
```
I'll be more precise, I'll change "rest of the stack" to "kivy.core.video"

I could also change the DictProperty to not allow None, but it seems desirable to be able to say Video(options=None).  I'll give this some thought.

As always thanks for the feedback!

ElliotGarbus: Done

kivy/tests/test_video_avfoundation.py
```diff
@@ -0,0 +1,150 @@
...
+import os
+import sys
+import unittest
+from os.path import abspath, join
+
+import pytest
+
+
+pytestmark = pytest.mark.skipif(
+    sys.platform != 'darwin',
```
misl6: Let's include `ios` from today, might be useful when we will start testing it :)

kivy/core/video/video_avfoundation_implem.mm
```diff
@@ -0,0 +1,957 @@
...
+#include "video_avfoundation_implem.h"
+
+#include <string.h>
+#include <stdlib.h>
+
+#include <EGL/egl.h>
+#include <EGL/eglext.h>
+#if TARGET_OS_IPHONE || TARGET_IPHONE_SIMULATOR
+#  include <OpenGLES/ES2/gl.h>
+#  include <OpenGLES/ES2/glext.h>
```
misl6: This looks from OpenGLES.framework (deprecated) and not from ANGLE, can't we just rely on ANGLE ones?

misl6 (commented):
Left two minor comments.

Awesome work as usual!

kivy/tests/test_video_avfoundation.py
ElliotGarbus: Done

kivy/core/video/video_avfoundation_implem.mm
ElliotGarbus: Done!  Good Catch.

misl6 (approved):
LGTM. Thank you!

PR #9320, opened 2026-05-26
fix(setup.py): honour KIVY_SDL3_PATH on macOS host (fixes #9316)

### LINKED ISSUE #9316: setup.py: KIVY_SDL3_PATH ignored on macOS host (use_sdl3 auto-detect gated on sys.platform == 'ios')
## Summary

When building Kivy 3.0 from source on a macOS host with `KIVY_SDL3_PATH`
set, `setup.py`'s `use_sdl3` auto-detection does not fire, and the
darwin `use_osx_frameworks` branch instead probes
`/Library/Frameworks/SDL3.framework` — silently ignoring the
user-supplied `KIVY_SDL3_PATH`. This blocks any cross-build pipeline
that vendors SDL3 outside `/Library/Frameworks/`, including kivy-ios.

## Where

`kivy/setup.py` — the `use_sdl3` auto-enable block (around line 340)
and the SDL3 path-resolution logic that downstream consumes
`use_osx_frameworks` vs `KIVY_SDL3_PATH`.

## Current behavior

The auto-enable for `use_sdl3` is gated on `sys.platform == 'ios'`,
which is only true when CPython itself reports `ios` (PEP 730 iOS host
/ cibuildwheel iOS). When cross-compiling from a macOS host
(`sys.platform == 'darwin'`) with `KIVY_SDL3_PATH` set — e.g. via
kivy-ios's `KivyRecipe.get_recipe_env()` — the auto-enable does not
fire, `use_sdl3` stays `False`, and the SDL2 path is selected.

Even when the user forces `USE_SDL3=1` via the `USE_<key>` env-var
override hook (around setup.py lines 340–345), `use_osx_frameworks`
(which defaults `True` on darwin) probes
`/Library/Frameworks/SDL3.framework` and short-circuits the
`KIVY_SDL3_PATH` consumption.

## Symptom

On a macOS host cross-build with `KIVY_SDL3_PATH` pointing at a
non-system SDL3 stack:

- `use_sdl3` stays `False`.
- SDL3 include dirs from `KIVY_SDL3_PATH` are not added.
- Builds either fail to find `SDL3/SDL.h`, or silently pick up a
  `/Library/Frameworks/SDL3.framework` that doesn't match the
  intended SDL3 vendor location.

Current workaround in kivy-ios (carried in a local recipe override):

```python
def get_recipe_env(self, plat):
    env = super().get_recipe_env(plat)
    env["USE_SDL3"] = "1"
    env["USE_OSX_FRAMEWORKS"] = "0"
    return env
```

## Suggested fix

Make `KIVY_SDL3_PATH` the controlling signal:

1. If `KIVY_SDL3_PATH` is set in the environment, auto-enable
   `use_sdl3 = True` regardless of `sys.platform`.
2. If `KIVY_SDL3_PATH` is set, skip the
   `/Library/Frameworks/SDL3.framework` probe — or, equivalently,
   give `KIVY_SDL3_PATH` precedence over `use_osx_frameworks`.

After this lands, callers can drop the `USE_SDL3=1` /
`USE_OSX_FRAMEWORKS=0` env-var workarounds entirely. Benefits any
cross-build pipeline that vendors SDL3 outside the system framework
location — kivy-ios, cibuildwheel iOS host pipelines, custom
toolchains, Homebrew packagers, etc.

## Reproduction

Minimal repro from a macOS host:

```bash
git clone https://github.com/kivy/kivy
cd kivy
git checkout master  # any post-SDL3 branch
KIVY_SDL3_PATH=/some/non-system/sdl3/include python setup.py build_ext --dry-run
# Observe: use_sdl3 stays False; KIVY_SDL3_PATH is not consumed.
```

End-to-end repro via kivy-ios is described in
kivy/kivy-ios#998 (item 5).

## Context

Discovered while bringing up Kivy 3.0 iOS support via kivy-ios. The
broader catch-up tracking issue is kivy/kivy-ios#998 — items 1–4 are
kivy-ios-owned recipe gaps; this issue is the upstream-Kivy half of
item 5.

ElliotGarbus:
**Problem**

When building Kivy for iOS on a macOS host with `KIVY_SDL3_PATH` set (e.g. via kivy-ios), the build system ignores the supplied path and falls back to probing `/Library/Frameworks/SDL3.framework` instead. This causes builds to either pick up the wrong SDL3 or fail to find it entirely.

**Root cause**

`use_sdl3` auto-detection was gated on `sys.platform == 'ios'`, which is only true when CPython itself runs on iOS. When cross-compiling from a macOS host (`sys.platform == 'darwin'`), the auto-detection never fires, and the macOS framework probe runs unconditionally — silently ignoring `KIVY_SDL3_PATH`.

**Fix**

If `KIVY_SDL3_PATH` is set, treat it as the authoritative SDL3 location: auto-enable `use_sdl3` and skip the framework probe. Setting `KIVY_SDL3_PATH` is now sufficient on its own — callers no longer need the `USE_SDL3=1` / `USE_OSX_FRAMEWORKS=0` workaround.

**Who is affected**

Any macOS host build where SDL3 is not in `/Library/Frameworks/` — including iOS cross-builds via kivy-ios, cibuildwheel iOS pipelines, Homebrew installs (`/opt/homebrew/`), and custom SDL3 builds.

Fixes #9316. Related: kivy/kivy-ios#998 (item 5).

---

Maintainer merge checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -517,6 +517,7 @@ elif platform == 'android':
 # If KIVY_SDL3_PATH is explicitly provided by the caller (e.g. a kivy-ios
 # cross-build running on a macOS host), treat it as an unconditional request
 # to use SDL3 via that path rather than system frameworks.
+# Regression: https://github.com/kivy/kivy/issues/9316
 if c_options['use_sdl3'] is None and environ.get('KIVY_SDL3_PATH'):
     print('KIVY_SDL3_PATH set, enabling SDL3 and bypassing framework probe.')
     c_options['use_sdl3'] = True
@@ -590,8 +591,9 @@ if c_options['use_sdl3'] or can_autodetect_sdl3:
     if c_options['use_osx_frameworks'] and platform == 'darwin' \
             and not environ.get('KIVY_SDL3_PATH'):
         # check the existence of frameworks
-        # (skipped when KIVY_SDL3_PATH is set — path-based config takes
-        # precedence over the system/vendored framework probe)
+        # Guard: skipped when KIVY_SDL3_PATH is set — path-based config takes
+        # precedence over the system/vendored framework probe.
+        # Regression: https://github.com/kivy/kivy/issues/9316
         if KIVY_DEPS_ROOT:
             default_sdl3_frameworks_search_path = join(
                 KIVY_DEPS_ROOT, "dist", "Frameworks"
--- a/setup.py
+++ b/setup.py
@@ -514,6 +514,14 @@ elif platform == 'android':
     c_options['use_android'] = True
     c_options['use_sdl3'] = True
 
+# If KIVY_SDL3_PATH is explicitly provided by the caller (e.g. a kivy-ios
+# cross-build running on a macOS host), treat it as an unconditional request
+# to use SDL3 via that path rather than system frameworks.
+if c_options['use_sdl3'] is None and environ.get('KIVY_SDL3_PATH'):
+    print('KIVY_SDL3_PATH set, enabling SDL3 and bypassing framework probe.')
+    c_options['use_sdl3'] = True
+    c_options['use_osx_frameworks'] = False
+
 # detect gstreamer, only on desktop
 # works if we forced the options or in autodetection
 if platform not in ('ios', 'android') and (c_options['use_gstreamer']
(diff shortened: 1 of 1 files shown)


kuzeyron (approved):

kuzeyron:
@misl6 This PR looks fine to me. But the Mac part is usually your thing :)

misl6 (approved):
Agree and approve the fix for the macOS only case.

For iOS, that would not be the case as during cibuildwheel build phases, `platform` is equal to `ios`, not `darwin`.

PR #9325, opened 2026-06-07
Modernize build: PEP 621 metadata

ElliotGarbus:
Working on kivy-ios I saw kivy-garden and other dependencies "leaking" into ios apps.  I fixed this using an exclude list in kivy-ios.  I used this as an opportunity to update the kivy build system.

- Migrate project metadata, dependencies and extras from setup.py/setup.cfg into a PEP 621 [project] table in pyproject.toml; slim setup.py to the C-extension build only (version stays dynamic).
- Split the Kivy-examples wheel into packaging/kivy_examples/ with its own pyproject.toml + setup.py; repoint the osx/manylinux/windows CI builds to it and drop the KIVY_BUILD_EXAMPLES path from setup.py.
- Retire setup.cfg: move [coverage:run] -> [tool.coverage.run] and [kivy] -> [tool.kivy]; update cython_cfg.py and doc conf.py to read [tool.kivy] from pyproject.toml (tomllib with tomli fallback).

Maintainer merge checklist
* [ ] Title is descriptive/clear for inclusion in release notes.
* [ ] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [ ] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #9326, opened 2026-06-11
:arrow_up: Complete upgrade with fixes for the use of Sphinx

kuzeyron:
Changes:
- Upgrading Sphinx to 9.1.0 from 6.2.1
- Removed jQuery as it's not used with Sphinx any longer
- Fixed a few bugs with the layout as the search engine threw +99 errors
- The layout is now up-to-date and follows Sphinx [own](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/layout.html) layout a little bit more
- Translated the jQuery code into vanilla JavaScript
- Removed finally block from kivy/uix/codeinput.py as it was the only file that had warnings with Sphinx
- Fixed the rendering issue where it's rendered as `from_kivy.app`
- Updated the CI to comply with Sphinx==9.1.0 with help from @ElliotGarbus


ElliotGarbus:
@kuzeyron  A few things worth double-checking:

**`!important` in `fresh.css`** — The new `.highlight .w` rule uses `text-decoration: none !important` and `border-bottom: none !important`. This usually signals a specificity conflict somewhere. Could you trace where the competing rule lives and see if the cascade can be fixed more cleanly, rather than overriding it? Happy to help dig into it if needed!

**`codeinput.py` change** — This one stood out a little since it's a non-doc file in a docs-infrastructure PR.  What was the `from_kivy.app` rendering bug? 

**Dangling jQuery references** — Related to point one — it'd be great to do a quick grep for `jquery` across all templates and static files just to make sure there are no other leftover references to the deleted JS files. Shouldn't take long and ensures nothing is broken.

Edit: deleted item on jquery... looks clean, well done!

kuzeyron:
> @kuzeyron  A few things worth double-checking:
> 
> **`!important` in `fresh.css`** — The new `.highlight .w` rule uses `text-decoration: none !important` and `border-bottom: none !important`. This usually signals a specificity conflict somewhere. Could you trace where the competing rule lives and see if the cascade can be fixed more cleanly, rather than overriding it? Happy to help dig into it if needed!
> 
> **`codeinput.py` change** — This one stood out a little since it's a non-doc file in a docs-infrastructure PR.  What was the `from_kivy.app` rendering bug? 
> 
> **Dangling jQuery references** — Related to point one — it'd be great to do a quick grep for `jquery` across all templates and static files just to make sure there are no other leftover references to the deleted JS files. Shouldn't take long and ensures nothing is broken.
> 
> Edit: deleted item on jquery... looks clean, well done!
> 

The rendering issue: https://kivy.org/doc/master/tutorials/pong.html
And it's the same on all other documents on master. Try changing to stable and it's gone.
And highlight.w doesn't seem to be used anywhere else. I have checked a lot of pages if there are anything that looks weird thanks to the changes but it doesn't. EDIT: Found one issue where the color is black instead of whiteish. I'll look into that in the evening when I get back.

EDIT: [9.0.4](https://github.com/sphinx-doc/sphinx/blob/v9.0.4/pyproject.toml) targets "Pygments>=2.17" and [6.2.1](https://github.com/sphinx-doc/sphinx/blob/v6.2.1/pyproject.toml) is targeting "Pygments>=2.13". So something must have changed last time Sphinx was upgraded to ~6.2.1 and wasn't fixed on our side.

ElliotGarbus:
The code_input.py fix looks good - but it seems out of scope for this pr.

kuzeyron:
> The code_input.py fix looks good - but it seems out of scope for this pr.

It was Sphinx that threw warnings about that part so I thought why not?! If there would have been more files with warnings, I would have made another PR to cover it. But it was just one file.

kuzeyron:
Screenshot_20260614_124010
Seems fine to me now so it's ready for full review now ^

ElliotGarbus:
Just a thought on the Sphinx version situation — since Sphinx is only used to build the docs and isn't a runtime dependency of Kivy itself, could we decouple it from the general `dev` extra and give it its own `docs` extra?

The Python 3.12+ requirement would be enforced on the CI runner for the `gen_docs` job rather than in the config itself, so the change is straightforward. If #9325 lands first, it would go in `pyproject.toml`:

```toml
[project.optional-dependencies]
dev = [
    # general dev tools
]
docs = [
    "sphinx==9.1.0",  # requires Python 3.12+ on the doc-build runner
]
```

Otherwise it could go in `setup.cfg` in the meantime:

```ini
[options.extras_require]
dev =
    # general dev tools
    ...
docs =
    sphinx==9.1.0  # requires Python 3.12+ on the doc-build runner
```

That way the CI docs job can install `kivy[docs]` on a Python 3.12+ runner and use Sphinx 9.1.0, while the regular test matrix continues supporting 3.11+ for Kivy itself. The two concerns are fully independent — the Python floor on Sphinx is really just a constraint on the doc-building environment, not on what Kivy supports at runtime.

This would avoid having to wait for the broader Python 3.11 deprecation decision just to unblock the Sphinx upgrade. This would also pair nicely with #9325, which migrates everything to `pyproject.toml` — if that lands first, the change would go there instead of `setup.cfg`.

ElliotGarbus:
One more thing worth noting — the change would also need to touch the CI side. Looking at `.ci/ubuntu_ci.sh`, the `gen_docs` job installs via `install_kivy`, which runs:

```bash
python3 -m pip install -e "$(pwd)[full,dev]"
```

followed by `generate_docs`, which simply calls `make html` in the `doc/` directory. Sphinx comes entirely from the `dev` extra through that install — there's no separate `doc-requirements.txt` step in the CI path (`doc/doc-requirements.txt` is currently just a placeholder comment anyway).

So splitting Sphinx into a `docs` extra would also mean updating the `gen_docs` job to call `install_kivy full,docs` on a Python 3.12+ runner, rather than `full,dev`. The rest of the `dev` tools would stay on the existing runner matrix as-is.

kuzeyron:
> One more thing worth noting — the change would also need to touch the CI side. Looking at `.ci/ubuntu_ci.sh`, the `gen_docs` job installs via `install_kivy`, which runs:
> 
> ```bash
> python3 -m pip install -e "$(pwd)[full,dev]"
> ```
> 
> followed by `generate_docs`, which simply calls `make html` in the `doc/` directory. Sphinx comes entirely from the `dev` extra through that install — there's no separate `doc-requirements.txt` step in the CI path (`doc/doc-requirements.txt` is currently just a placeholder comment anyway).
> 
> So splitting Sphinx into a `docs` extra would also mean updating the `gen_docs` job to call `install_kivy full,docs` on a Python 3.12+ runner, rather than `full,dev`. The rest of the `dev` tools would stay on the existing runner matrix as-is.

I think it would be great if we wait with that PR because of the version tool that is above the search box on the doc pages. If it works as intended the other PR could be applied and you might have to mess with that thing that is creating `versions.json`. Server and local builds aren't the same when it comes to `versions.json`. I have only been using a hack to fake the same results on my end.

ElliotGarbus:
Tested locally — clean build with Sphinx 9.0.4, the from kivy.app rendering fix looks great, dark mode and code highlighting both look good.

Would you be open to also bumping to 9.1.0 and moving Sphinx into a separate docs extra? That way the CI gen_docs job can run on a Python 3.12+ runner and we'd land on the latest Sphinx rather than one version behind. Happy to help with the CI side if useful.  

Whenever you get a chance to add that in, I'm happy to re-review and approve.

kuzeyron:
> Tested locally — clean build with Sphinx 9.0.4, the from kivy.app rendering fix looks great, dark mode and code highlighting both look good.
> 
> Would you be open to also bumping to 9.1.0 and moving Sphinx into a separate docs extra? That way the CI gen_docs job can run on a Python 3.12+ runner and we'd land on the latest Sphinx rather than one version behind. Happy to help with the CI side if useful.
> 
> Whenever you get a chance to add that in, I'm happy to re-review and approve.

> Would you be open to also bumping to 9.1.0 and moving Sphinx into a separate docs extra?

Absolutely. But I know almost nothing about the CI-code. Something I never really have had to do before so help with that would be great! You can DM me on Discord when you have the time!

.github/workflows/test_ubuntu_python.yml
```diff
@@ -83,14 +83,10 @@ jobs:
         source .ci/ubuntu_ci.sh
         install_ubuntu_build_deps
         ./tools/build_linux_dependencies.sh
-    - name: Setup env
-      run: |
-        source .ci/ubuntu_ci.sh
-        prepare_env_for_unittest
```
ElliotGarbus, line 89: Removing the prepare_env_for_unittest step from gen_docs — confirmed correct. I checked this from a few angles and it's the right call. The removal is isolated to gen_docs so the real build/test jobs are unaffected.
One suggestion: a short comment in the workflow noting the docs job intentionally skips the unittest env setup would help stop someone re-adding it later.

I don't consider this a blocking issue but wanted to call it to your attention.

doc/README.md
```diff
@@ -12,37 +12,15 @@ If you intend on editing and contributing documentation, assure the kivy source
...
-
-  ``pip install -r doc-requirements.txt``
-  
-
 Building the documentation
 --------------------------
 
+Install the documentation dependencies (Sphinx requires Python 3.12+)::
+
+  ``pip install -e ".[full,docs]"``
```
ElliotGarbus: The CI gen_docs job installs dev,docs, but the README tells contributors `pip install -e ".[full,docs]"`. I went down a rabbit hole on whether full is actually needed for docs, and it isn't: core_select_lib short-circuits on KIVY_DOC, so the image/video/audio backends (pillow, ffpyplayer, etc. — what full adds) are never imported during a doc build, and none of them generate API pages anyway. ThorVG is only pulled in lazily via the core providers, so it's skipped too. The minimal correct install is really just `.[docs]`. 

Might be worth dropping full from the README so it matches CI — either way builds fine, just a consistency thing.

ElliotGarbus:
Did a local build with Sphinx 9.1.0 (Python 3.13) and reviewed the changes — looks great. Full clean build succeeded, the from kivy.app rendering fix holds across all pages, and CI is green including gen_docs. Nice work.

I left two comments on the code.  I don't consider either of them blocking, let me know what you think.

ElliotGarbus (approved):
Well done!

misl6:
FYI: Merged #9325 , ping me when the needed changes are in place :)

kuzeyron:
> FYI: Merged #9325 , ping me when the needed changes are in place :)

Ping @misl6 ^

misl6 (approved):
LGTM. Thank you!

PR #9331, opened 2026-06-16
feat: introduce kivy.mobile — platform-neutral bridge for mobile window/display geometry

ElliotGarbus:
As we look to evolve the way mobile app are built in Kivy 3.0 there is an opportunity to move some of the code that exists in kivy-ios and p4a into kivy.  This keeps the build tools "collectors of files" not owners of runtimes.

Kivy currently relies on packaging tools — kivy-ios and python-for-android — to provide runtime information that Kivy's own layout and metrics systems need. When a Kivy app runs on iOS, Kivy calls ios.get_dpi(), ios.get_scale(), and ios.get_kheight() — functions written by the build tool, not by Kivy. The interface is implicit, untested in Kivy's own CI, and means that runtime behaviour Kivy depends on is owned by a packaging recipe rather than by Kivy itself.

The right model is for build tools to be collectors: their job is to gather frameworks, libraries, and packages and assemble an app bundle. Runtime behaviour belongs in Kivy. 

This PR takes the first step. kivy.mobile moves the mobile runtime interface — and its iOS implementation — into Kivy core. kivy.mobile._platform.ios owns what was previously delegated to ios.py: DPI, display scale, safe-area insets, and live keyboard height via a NSNotificationCenter observer. metrics.py and core/window now import from kivy.mobile rather than bare import ios. The API contract is declared, documented, and tested in Kivy CI without requiring a device.  

A follow-up PR will fill in the Android details after related changes to support Android are made, specifically pulling some content from p4a to a separate kivy-android support module. 

## Summary

- Introduces `kivy.mobile`, a new Kivy-owned namespace that centralises mobile runtime geometry (DPI, scale, safe area, keyboard height) currently scattered across packaging-side shims (`ios.py`, bare `import android`).
- iOS is fully implemented via ctypes / Objective-C runtime (no new compiled dependency). Android is an explicit, placeholder that will be filled in after some changes to p4a.
- `Window.safe_area` (`DictProperty`) is added, refreshed on startup and `on_rotate`. `metrics.py` and `core/window` are updated to call `kivy.mobile` instead of bare platform imports.

### New files

- `kivy/mobile/__init__.py` — Public API, dispatches by `kivy.utils.platform`. Raises `ImportError` on desktop with a clear message.
- `kivy/mobile/_platform/ios.py` — Full UIKit implementation: DPI (nativeScale × base), scale, density, safe area (safeAreaInsets), live keyboard height via NSNotificationCenter observer, subscribe_keyboard_height.
- `kivy/mobile/_platform/android.py` — Explicit placeholder with UserWarning at import and TODO pointing at jnius / android module implementation path.
- `kivy/tests/test_mobile.py` — Desktop-runnable tests that load `_platform` modules directly by file path, bypassing the mobile-only guard.

## Design decisions

- **mobile-only** — `kivy.mobile` raises `ImportError` on desktop. Guard with `if platform in ('ios', 'android'):`.
- **No DPI lookup table** — the legacy per-model table topped out at iPhone 13 / 2021 iPad. `nativeScale × base_dpi` (163 ppi iPhone, 132 ppi iPad via `UIDevice.model`) works for all devices past and future.
- **`get_kheight()` removed** — internal convention, never documented as public API. 
- **Android placeholder** — jnius and the `android` p4a module are available today; the placeholder exists only because elements of the p4a Android module needs to be packaged as a separate kivy-android wheel.


misl6 (approved):
LGTM. Thank you!

PR #9334, opened 2026-06-22
[build wheel linux] fix ci by using pipx to install meson

damascene:
Make sure pip is installed for build to pass


ElliotGarbus:
Thanks for catching this and putting up a fix! 

Great find, especially since the `manylinux_wheels` job only runs on schedule/tag/`[build wheel]`, so this is easy to miss.

Rather than bootstrapping pip, let's use `pipx` here — it's the right tool for this. The manylinux images ship it and set `PIPX_BIN_DIR=/usr/local/bin` (already on `PATH`), so it installs meson in an isolated env and drops it straight onto `PATH` for the ThorVG build, with no pip-bootstrapping needed. Let's also pin the version (my mistake) for reproducible wheel builds (`--upgrade` resolves to `1.11.1` today anyway):

`pipx install "meson==1.11.1"`

That single line can replace all three. To confirm the wheel build goes green, include `[build wheel linux]` in the PR title and push a commit (editing the title alone won't trigger it, since the workflow doesn't listen for `edited` events) — or add `[build wheel linux]` to a commit message. Thanks again.

@misl6 could you approve the workflow runs on this PR when you get a chance? It looks like they're all sitting in "awaiting approval" since this is @damascene's first contribution, so no CI has actually run yet.
One thing to note for the wheel build specifically: the `manylinux_wheels` job is gated on `[build wheel linux]` being in the PR title, and it reads that from the event that queued the run. So to actually exercise this fix end-to-end, the order needs to be: add `[build wheel linux]` to the PR title → push a commit (which queues a fresh run) → then approve that run. Otherwise the standard checks will run but the wheel job will just show as skipped.
Thanks!

kuzeyron (approved):

kuzeyron:
LGTM! Merging.

PR #9337, opened 2026-07-01
:bug: Fix docs version chooser after Sphinx upgrade (#9326)

ElliotGarbus:
The jQuery to vanilla-JS refactor in #9326 broke the sidebar version chooser. Docs are served under "/doc/<version>/...", but the new code used `pathname.split('/').filter(Boolean)[0]`, which points at the "doc" segment instead of the version. As a result no version was ever preselected, and switching versions rewrote the wrong segment (dropping the "/doc/" prefix). It also appended a trailing slash, producing URLs like "/doc/master/tutorials/pong.html/" that 404 on the server.

Anchor on the "doc" path segment to locate the version, and rebuild the path from the raw split so the rest of the URL is preserved exactly with no spurious trailing slash.

Maintainer merge checklist

### DIFF
--- a/doc/sources/.static/kivy.js
+++ b/doc/sources/.static/kivy.js
@@ -313,13 +313,22 @@ document.addEventListener("DOMContentLoaded", function() {
     var selector = document.getElementById("version_selector");
 
     if (selector) {
-        selector.addEventListener('change', function() {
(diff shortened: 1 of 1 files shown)


kuzeyron (approved):

kuzeyron:
LGTM! Thank you!

PR #9339, opened 2026-07-04
Control statements for kv-lang if/elif/else, for, slot, and factory

mtctn:
# Control statements for kv-lang

> Targets Kivy **3.0.0**.

I'm quite happy to share this feature which I hope will be welcomed in the Kivy project.
I've been working on it for a long time, and I admit that a few sessions using Claude Fable 5 helped me finalize it.

This PR message goes into details, to share the whys and hows of decisions I believe are right and that weren't obvious at first.

## Why

kv-lang is great at describing a *fixed* tree of widgets, but most UIs aren't fixed: things appear and disappear, and lists grow and shrink. kv has no real answer for either, so everyone improvises - and tends to improvise differently each time.

Take hiding a widget. There's no "hide" in kv, so in practice you reach for several properties at once and hope the combination is good enough:

```yaml
Label:
    opacity: 0 if hide else 1         # make it invisible...
    disabled: hide                    # ...and stop it handling touches...
    height: 0 if hide else dp(40)     # ...and try to reclaim its space
    size_hint_y: None
```

None of those removes the widget; each has its own side effects; and the recipe varies from project to project (if not from file to file).

Rendering a list has the same problem from the other direction - you leave kv and do it in Python, or use a RecycleView for half a dozen widgets, and every contributor picks a different pattern.

The two things people actually want here are just `if` and `for`. This PR brings these Pythonic control structures into kv, declaratively and reactively, so the common case has one boring, obvious shape:

```yaml

if visible:
    Label:
        text: 'now you see me'

for item in self.items:
    Label:
        text: item.title
```

This is not a RecycleView replacement: `for` builds a real widget per item, so RecycleView is still the right tool for large, scrolling, virtualised lists. The target is the everyday handful-of-widgets case where RecycleView is overkill.

## A todo list, in kv and nothing else

Add, list, empty state, delete - the whole thing, with no Python beyond loading the string:

```yaml
BoxLayout:
    orientation: 'vertical'
    padding: dp(8)
    spacing: dp(8)
    todos: []

    BoxLayout:                              # input row
        size_hint_y: None
        height: dp(40)
        spacing: dp(8)
        TextInput:
            id: field
            hint_text: 'What needs doing?'
            multiline: False
            on_text_validate:
                root.todos = root.todos + [self.text]
                self.text = ''
        Button:
            text: 'Add'
            size_hint_x: None
            width: dp(80)
            on_release: field.dispatch('on_text_validate')

    if not root.todos:
        Label:
            text: 'Nothing to do...'

    for i, todo in enumerate(root.todos):   # one row per todo
        BoxLayout:
            size_hint_y: None
            height: dp(36)
            Label:
                text: todo
            Button:
                text: 'x'
                size_hint_x: None
                width: dp(40)
                on_release: root.todos = root.todos[:i] + root.todos[i + 1:]
```

Reassigning `todos` is all it takes - the `if` and `for` blocks re-render themselves.

A fully working demo is included in `examples/kv/app_todo.kv`.

## What it adds

Four reactive control statements at child-widget position, with the same binding semantics as ordinary kv property expressions:

- **`if` / `elif` / `else`** - mount a block while its condition holds. A block is *whatever fits in an ordinary rule*: besides children it may set properties, declare a `canvas`, and bind handlers, all applied to the host widget while the branch is active and removed when it leaves.
- **`for ... in ...`** - one copy of the body per item, reconciled as the iterable changes. It uses native Python comprehension syntax, so tuple targets and trailing `if` filters work (`for i, x in enumerate(items) if x.visible:`). 
- **`slot`** - a named insertion point in a class rule that the usage site can fill, with a fallback.
- **`factory <expr>`** - build one child widget whose class is chosen by an expression (a class object, or a name resolved through the `Factory`).

**Opt-in and non-breaking.** This is purely additive. Existing kv keeps parsing and behaving exactly as before - control statements are new keywords in positions that used to be syntax errors, so nothing that parses today changes meaning. A rule with no control statements takes an unchanged fast path, so code that doesn't use the feature pays nothing for it, in behaviour or performance.

### Setting properties conditionally, without repeating yourself

An `if`/`else` block can set the host widget's own properties, so a group of properties that all hinge on one condition can state it once instead of on every line:

```yaml
# today - the condition is repeated on every property:
<StatusLabel@Label>:
    error: False
    color: (1, 0, 0, 1) if self.error else (1, 1, 1, 1)
    bold: self.error
    font_size: '18sp' if self.error else '14sp'
```

```yaml
# with if / else - say it once:
<StatusLabel@Label>:
    error: False
    if self.error:
        color: 1, 0, 0, 1
        bold: True
        font_size: '18sp'
    else:
        color: 1, 1, 1, 1
        bold: False
        font_size: '14sp'
```

### Slots

A slot is a named hole a container class leaves for its callers to fill, with fallback content for when they don't. It lets you build a reusable layout shell once and drop different content into it at each call site, without subclassing:

```yaml
# a reusable frame with a header slot and a default (body) slot
<Card@BoxLayout>:
    orientation: 'vertical'
    slot header:
        Label:
            text: 'Untitled'         # shown when the caller fills nothing
    slot:                            # default slot: plain children land here

# call site - fill the header, drop the body content straight in
Card:
    slot header:
        Label:
            text: 'Inbox'
            bold: True
    Label:
        text: 'You have 3 new messages'
```

A definition can also hand values to whatever fills it: a property line in a `slot` definition is a *slot-scoped local*, computed in the defining rule's context and read by fallback and fill content through the reserved `slot` name:

```yaml
<Card@BoxLayout>:
    unread: 0
    slot header:
        badge: f'({root.unread})' if root.unread else ''
        Label:
            text: 'Untitled ' + slot.badge     # fallback reads it

Card:
    slot header:
        Label:
            text: 'Inbox ' + slot.badge        # so does the fill
```

### A widget chosen at runtime (`factory`)

`factory <expr>` builds one child whose class comes from an expression - a class object, or a name resolved through the `Factory`. The body is applied to the instance, and the widget is rebuilt when the class expression changes. Combined with `for`, this is a whole form rendered from data:

```yaml
<Form@BoxLayout>:
    schema: []
    for field in self.schema:
        factory field['cls']:        # e.g. 'TextInput', 'CheckBox', ...
            hint_text: field.get('label', '')
```

### Iteration-local values and reactive ids

A property line inside a `for` is a *local*: one reactive value per iteration, shared by everything in that iteration's body (so the `[]` below is one list, not three). A widget inside the block may carry an `id`, which the iteration's own content can reach:

```yaml
<Cart@BoxLayout>:
    items: []
    for item in self.items:
        line_total: item.qty * item.price     # iteration-local, reactive
        CheckBox:
            id: pick                          # iteration-local id
        Label:
            text: '%s = %s' % (item.name, line_total)
        Label:
            text: 'picked' if pick.activated else ''
```

In an `if`, an id is reachable across the whole rule and is `None` while the branch is inactive - so a sibling can react to the widget coming and going:

```yaml
<Editor@BoxLayout>:
    editing: False
    if self.editing:
        TextInput:
            id: field
    Button:
        disabled: field is None      # True until the TextInput exists
```

### Graphics control flow (inside canvas)

`if` and `for` also work *inside* a `canvas` block, where they generate graphics instructions instead of widgets - data-driven drawing that reacts to its inputs, the same way it reacts for widgets:

```yaml
<Chart@Widget>:
    series: []
    canvas:
        Color:
            rgba: 1, 1, 1, 1
        for line in self.series:      # one Line per series, reactive
            key: line.uid             # keyed: reconcile instead of rebuild
            Line:
                points: line.points
        if self.selected:             # extra graphics only when selected
            Color:
                rgba: 1, 0, 0, 1
            Rectangle:
                pos: self.pos
                size: self.size
```

The block's instructions sit at its position among the surrounding ones and are rebuilt when the condition or iterable changes. A canvas `for` without a `key:` rebuilds its instructions wholesale; with a `key:` it reconciles per iteration (keeping, moving or rebuilding instruction groups by key), so textured/stateful instructions (`Rectangle(source=...)`, `Mesh`, `Fbo`) and large series are not re-uploaded on every change.

## Design

The guiding choice is that control statements are not a second engine bolted onto kv - they reuse the language's existing reactivity. An `if` chain compiles to a single *selector* expression (the index of the active branch) and a `for` to a single *iterator* expression (the list of loop-value tuples), both ordinary `ParserRuleProperty` objects. Binding, rebinding and teardown ride the standard watched-keys path, so the same things that make `text: self.value` reactive make the conditions reactive.

At runtime each control statement is a small node that owns a contiguous span of its parent's children and rebuilds that span when its expression changes. Nodes activate in a deferred phase of `_apply_rule`, after properties and handlers are resolved, so ids and values are already in place; they build their subtrees by re-entering `_apply_rule`, so canvas, handlers and nested control statements come for free. Rebuilds are synchronous (after `self.open = True` the children exist immediately), and an update that fires while another apply or rebuild is in flight is queued and run right after it, so a rebuild can never re-enter an in-flight rule or observe a half-reconciled children list. Widgets without control statements take a fast path - one boolean test per child.

### Reactive properties

**Golden rule: scoped values.** A property line binds to the *nearest enclosing scope*: the host widget in a rule-level `if`, the iteration in a `for` (including in an `if` nested inside the `for`, where it becomes a *conditional* local), the slot in a `slot` definition. An iteration-local is a named, reactive intermediate computed once per iteration and *shared* by everything in that iteration's body. The name is visible only inside the block.

Properties in an active `if` / `elif` / `else` branch **add ordinary kv bindings while active and drop them when inactive**, and they compose exactly like bindings from stacked rules: coexisting drivers settle by reactivity order (the last dependency to change wins), and kv never auto-reverts a property.

So the idiomatic way to drive a property by a condition is `if` / `elif` / `else`, which keeps exactly one binding live and always leaves the property with a defined value:

```yaml
<StatusLabel@Label>:
    error: False
    if self.error:
        color: 1, 0, 0, 1
        bold: True
    else:
        color: 1, 1, 1, 1
        bold: False
```

The `else` branch is how a value comes back - there is no implicit reversion to an earlier state. A property set in a branch with nothing to fall back to (no `else`, no unconditional binding) keeps the branch's last value after it leaves: kv's one-way binding, applied to structure.

Mixing an unconditional binding with a conditional override of the *same* property - or two independent conditional blocks that drive the same property at once - is allowed, but is a power-user move with the usual parallel-binding caveat: while more than one is active their bindings are all live and resolve by reactivity order, just like several rules setting one property:

```yaml
color: self.theme_color     # always bound
if self.urgent:
    color: self.urgent_color     # bound while urgent
# when both are bound, the last reactive update wins
```

Two properties of the model matter:

- **Inactive branches hold no bindings at all.** A branch's expressions are bound and evaluated only when it activates - synchronously, so the value is correct on the first frame - which keeps the property *hot path* untouched: a widget with no active branch performs exactly as before.
- **Teardown is surgical.** Binding a branch records exactly the handler entries it created (`_bind_prop` / `_unbind_captured`), so deactivating it unbinds only its own callbacks and never disturbs another binding's, whether base or sibling.

Handlers and canvas in a branch follow the same mount/teardown discipline: a branch's handlers are `fbind`'d on activation and `unbind_uid`'d on teardown, and its `canvas` is built into one `InstructionGroup` added to the host canvas and removed atomically when the branch leaves.

### Iteration scopes: locals and loop targets

Canvas and handlers are accepted only in `if`/`elif`/`else` blocks outside any `for` - an `if` nested in a `for` follows the for-body rules (nesting context semantics). A property line in a `for` body is not a host property but an **iteration-local** (see below); canvas and handlers, whose per-iteration meaning would be ambiguous, are rejected. A property line in a `slot` definition is a **slot-scoped local** (see below).

A property declared in a `for` body is a value local to each iteration, not a property of the host widget. Each iteration owns a hidden scope object (an `EventDispatcher`); the local is a real reactive property on it, evaluated once per iteration, so a `[]` is a single shared list (every reference in the iteration sees the same object), and it re-evaluates when its dependencies change. References to the name are rewritten at parse time to attribute access on the scope object (`__kvscope_N.name`), so reactivity rides the same watched-keys path as everything else - no new binding engine, and the name is naturally scoped to the block. Locals may reference the loop targets and earlier locals.

The **loop targets live on the same scope object**, as reactive properties. Every reference to a loop variable - in children, locals, handlers, nested headers - is rewritten to scope attribute access, which is what makes keyed reconciliation cheap: when a kept key comes back with a *replaced* item, the runtime just assigns the new values to the scope properties and the existing bindings re-fire. No rebuild, no re-apply pass - the same dispatch that drives `text: self.value` drives "this row now shows another task".

By the nearest-scope rule, a property line in an `if` nested inside a `for` is a *conditional* iteration-local: declared on the scope up front (`None` until first activation), bound while the branch is active, keeping its last value when the branch leaves - the same add/drop-bindings model as conditional host properties, applied to the scope. A local and an `id` cannot share a name: unlike Python shadowing, both writers would stay live (the id writes at mount, the local re-evaluates on every dependency change), so the clash is a parse-time error.

### Slot-scoped locals

A property line in a `slot` definition declares a local on a per-slot scope object, evaluated in the *defining* rule's context. Content reads it through the reserved `slot` name (`slot.badge`), which is injected into the content's evaluation context - deliberately *not* parse-time rewriting: a fill is written in another rule (often another file), whose parser cannot know the slot's local names. Dotted access through an injected dispatcher gives fills reactivity through the ordinary watched-keys path with zero cross-rule knowledge. In nested slots the innermost `slot` wins.

This is deliberately **not** textual inlining: inlining `bucket: []` would substitute a fresh `[]` at each reference, giving three different lists instead of one shared value. Backing the local with a property is what makes the value single and shared.

### Reactive ids

`id` is allowed on a widget inside an `if`, a `for`, or an explicit `slot` block. Because that widget comes and goes, the id can't be a static entry in `root.ids` (it would blink in and out and silently break anything bound to it). Instead it is held as a reactive property on a scope object:

- In an **`if`** the id lives on a rule-level scope, reachable across the whole rule, and is `None` while its branch is inactive - so an expression like `disabled: field is None` reacts to the widget mounting and unmounting. The runtime resets it to `None` when the branch leaves - with a compare-and-set (only the widget that owns the value clears it), so two branches of one chain, or two complementary chains (`if cond:` / `if not cond:`), may share an id: it always points at the widget most recently mounted under that name.
- In a **`for`** the id lives on the iteration scope, reachable only by that iteration's own content, and dies with the iteration.
- In a **`slot` block** (a definition's fallback or a fill) the id lives on the rule-level scope of the rule *providing* that content.

The id'd widget is parse-time marked with its `(scope_key, name)`; at build time the runtime sets `scope.name = widget` instead of touching `root.ids`, and references are rewritten to scope attribute access (so they bind reactively). Scope properties are created `allownone=True, rebind=True`. `id` is still rejected directly on a control statement, inside `factory` blocks, and on plain children implicitly routed into a slot - whether those are routed depends on class rules the parser cannot see from the call site; an explicit `slot:` fill block is the supported spelling.

### Building a widget by class (`factory`)

`factory <expr>` is the single-widget analogue of the span nodes: a `FactoryNode` (an `ExpressionNode`) owns a one-widget span and rebuilds it when the class expression changes. The expression yields a class object or a string resolved through `Factory.get`; the block body is applied to the instance as an ordinary rule (so its properties, children, canvas and handlers all work, and unknown properties auto-create as elsewhere in kv). A constant class name is built once with no binding. `factory` is rejected inside `canvas`.

### Control flow inside canvas

A canvas is an ordered container just like `widget.children`, so the same "reactive span of a container" idea applies - the only differences are the four container ops (build / mount / unmount / teardown). The expression machinery is factored into a mixin (`_ExpressionDriven`) shared by the widget nodes and the canvas nodes, so both react through the identical bound-expression path.

Each canvas `if`/`for` owns one `InstructionGroup` slotted into the enclosing canvas at the block's position; rebuilding clears and refills that group, so instruction positions never need recomputing and content after the block stays put. Nesting composes (a nested block adds its own group to its parent's group). A canvas `for` without a `key:` rebuilds the group wholesale - fine for cheap, stateless drawing; with a `key:` it reconciles per iteration (each iteration owns a sub-group kept, moved or rebuilt by key), so textured/stateful instructions (`Rectangle(source=...)`, `Mesh`, `Fbo`) and large series are not destroyed and re-uploaded on every change - mirroring the widget `for`. Like the widget nodes, the canvas nodes live with their host widget (no module registry), so they are collected with it.

### Other decisions

- **Keyed `for` reconciliation: same key, same widgets.** A kept key keeps (and if needed moves) its widgets; a kept key whose item was replaced gets the new loop values re-dispatched through its scope properties, so live widget state (text mid-edit, focus, selection) survives item updates. Only new keys build and only vanished keys destroy. Keys behave like dict keys (hashable, compared by equality) and must be unique and *stable*. Without `key:` the position in the iterable is the key. In canvas, a kept iteration with changed values is rebuilt instead - instructions hold no user state worth preserving.
- **Slot precedence: most derived wins.** The first rule declaring a slot defines the insertion point and fallback; later same-name blocks are fills (base fallback < subclass fill < instance fill), evaluated in the providing rule's context. Fills may re-expose a slot under a fresh name (forwarding through wrappers), and a slot may live inside an `if`. A filled slot never builds its fallback. Filling a name no class rule ever declared degrades to defining a new insertion point in place - and logs a warning, since an instance-level definition can never be filled (it is almost always a typo).

## Rejected ideas

- **A layered "single live driver" property stack.** An earlier version kept, per `(widget, property)`, a stack of candidate bindings - an unconditional *base* plus one *layer* per active branch - bound only the top layer, gave layers document-order priority (a later block beat an earlier one), and on deactivation *re-asserted* the base, rebinding it live; a companion path recorded a base set in one rule so an `if` in another rule could restore it. Dropped: it is a second precedence model bolted alongside kv's existing "stacked bindings, last-writer-per-dependency wins", and the re-assert step is an implicit revert no other kv binding performs (kv is strictly one-way). The chosen model - a branch adds and removes ordinary bindings - needs none of that machinery, keeps one precedence rule (reactivity order, as everywhere else), and makes reversion explicit via `else`.

- **One global precedence stack across all rules.** A further step would fold cross-rule precedence and branch precedence into a single priority-ordered stack per property, with `-prop:` meaning "clear lower layers". Rejected: applied universally it changes the meaning of existing control-free multi-rule kv (whose bindings coexist today); applied only to control-using widgets it reintroduces the inconsistency it was meant to remove. Cross-rule precedence is therefore left exactly as kv has it today.

- **Reverting a branch-only property to its default.** Considered snapping a property that is set only inside a branch back to its property default when the branch leaves. Rejected as un-kv: one-way bindings never revert. The value is kept; use `else` to set it back deliberately.

- **`for` / `else` for empty states.** Jinja renders the `else` block when the iterable is empty; Python's `for ... else` means something else entirely (run after normal completion), so overloading the keyword would confuse more than it helps. The golden path is a paired `if not x:` block next to the loop - one duplicated condition on the same bound property.

- **`let` for locals.** Considered a dedicated keyword for scoped locals instead of bare property lines. Rejected for now: the nearest-scope rule ("a property line binds to the nearest enclosing scope") is one sentence and covers host properties, iteration-locals and slot-locals uniformly. `let name:` is a syntax error today, so it can still be added later if rule-level locals prove needed.

- **Last-wins for local/id name clashes.** Considered Python-style shadowing. Rejected: unlike Python, neither writer stops - the id writes at mount, the local re-evaluates on each dependency change, so "last" would mean reactivity order and the name would flap between a widget and a value. Every program the clash rejects is broken, so it is a parse-time error.

## Follow-up ideas (out of scope here)

Things this implementation surfaced that stand on their own:

- **AST-based watched keys.** kv's dependency extraction is a regex (`lang_keyvalue`), which is why `field.text.strip()` binds nothing (the dependency must *end* the dotted chain) and why iterating a dict needs the `dict.items(x)` spelling. The parser now runs `ast.parse` over every expression anyway (for scope rewriting); deriving watched keys from the AST would fix both gotchas for *all* of kv, not just control headers. The highest-value follow-up.
- **Mount/unmount transitions.** Nodes have clean mount/unmount seams; `enter:` / `leave:` animation blocks would be a localized extension of `IfNode` / `ForNode`.
- **Scoped slots inside `for`.** Chaining an iteration scope into a slot scope would lift the no-slot-in-for restriction and give the caller-provided-row-template pattern (`List` shells with `slot row:`).
- **Inspector integration.** Nodes are introspectable on their host (`_kv_control_nodes`, spans, active branch, iteration keys); the inspector module could visualize live control flow with little work.

## Restrictions (enforced with targeted error messages)

- No control statements at the top level of a kv file.
- A property line binds to the nearest enclosing scope (host in a rule-level `if`, iteration anywhere under a `for`, slot in a `slot` definition). Handlers and `canvas` are accepted only in `if`/`elif`/`else` outside any `for`; `for` bodies and `slot` blocks take neither. A one-off conditional value can still use an expression (`text: 'a' if cond else 'b'`).
- Inside a `canvas` block only `if`/`for` are allowed (not `slot`/`factory`); their bodies are graphics instructions, and a control statement may not be nested under an individual instruction.
- `id` is allowed on widgets inside `if`/`for` and explicit `slot` blocks (reactive); it is rejected directly on a control statement, on plain children implicitly routed into a slot, and inside `factory` blocks. A local and an `id` cannot share a name.
- `key` is a reserved name inside `for` bodies (an iteration-local cannot be called `key`); keys behave like dict keys and must be unique.
- Loop targets form their own scope and may shadow any name (including `self`/`root`/`app` and the metric helpers); a child widget's own `self`/`args` still win inside that widget. No walrus in headers, no `async for`, single `for` clause.
- No slot definition inside `for`; fills must be direct children of the instance; a `slot:` fill block can't be mixed with plain children. `factory` requires a class expression and is not allowed inside `canvas`.

## Caveats

- The children a control statement builds are managed. Manually *adding* children is safe (appends land after the managed content); manually removing or reordering them (`clear_widgets()` included) declares a second, competing owner of the same list - nothing raises and nothing leaks, but the insertion position of managed content becomes undefined from then on. Kivy-style soft behavior, documented in the language guide.

## Testing

- **60 parser tests** (`test_lang_parser_control.py`): syntax, compiled output, full-body `if` parsing, `factory` parsing, iteration-local and slot-local extraction, conditional locals (if-in-for), reference rewriting, reactive-id scoping (if / for / slot), the local/id name-clash error, and every error path.
- **101 runtime tests** (`test_lang_control_runtime.py`): branch switching; keyed reconciliation - identity across reorders/removals/appends *and* same-key value re-dispatch through kept widgets; nesting both ways; per-item reactivity; slot fallback / fill / inheritance / forwarding / scoped locals / fill ids / unknown-name warning; handler- and `rulectx`-leak regressions; re-entrant rebuild convergence; the full-body `if` - conditional properties as add/remove bindings (one-way, no revert; `else` for graceful reversion; parallel blocks resolve by reactivity order; base coexists across rules), conditional canvas mount/unmount and binding, conditional handlers; iteration-locals (shared identity, reactivity, local-to-local, conditional locals' lifecycle); `factory` (by name and by class, reactive rebuild, inside `for`, document position, unknown-class error context, GC); reactive ids (`if` nullable/reachable-outside, `for` iteration-local, complementary chains sharing an id via compare-and-set reset, no `root.ids` leak, GC); `if`/`for` inside canvas - reactive instruction build/teardown, document position, z-order across branches, nesting, delayed bindings, keyed reconciliation, no instruction-handler leaks; kv-located header errors (non-iterable); and that a widget using control statements is garbage-collected once dropped.
- Full suite green apart from environment-specific collection errors for optional native modules (system tray SDL3, thorvg) present before this change.

Maintainer merge checklist
* [X] **Unittests** are included in PR.
* [X] Properly documented, including `versionadded`, `versionchanged` as needed.

### DIFF
--- a/doc/sources/guide/lang.rst
+++ b/doc/sources/guide/lang.rst
@@ -419,6 +419,144 @@ This class, created just by the declaration of this rule, inherits from the
 Button class and allows us to change default values and create bindings for all
 its instances without adding any new code on the Python side.
 
+Control statements
+------------------
+
+.. versionadded:: 3.0.0
+
+By default a rule builds a *fixed* tree of widgets. Control statements let that
+tree change at runtime -- widgets that appear and disappear, and lists that grow
+and shrink -- described directly in kv and kept up to date automatically. At any
+child position a rule may contain an ``if`` / ``elif`` / ``else``, ``for``,
+``slot`` or ``factory`` block. Each block is *reactive*: whenever a property
+used in its header changes, the block rebuilds the content it manages. There is
+nothing to wire up -- reassign the property and the tree follows.
+
+Conditional content
+~~~~~~~~~~~~~~~~~~~~
+
+Use ``if`` / ``elif`` / ``else`` to mount widgets only while a condition holds:
+
+.. code-block:: kv
+
+    <LoginBox@BoxLayout>:
+        logged_in: False
+        Label:
+            text: 'Account'
+        if self.logged_in:
+            Button:
+                text: 'Log out'
+        else:
+            Button:
+                text: 'Log in'
+
+Toggling ``logged_in`` swaps one button for the other. Unlike a widget hidden
+with ``opacity`` or ``disabled``, an inactive branch's widgets are not in the
+tree at all -- no ghost touch target and no leftover layout space. A branch may
+also set the host widget's own properties, declare a ``canvas`` or bind event
+handlers while it is active. Branch properties are ordinary one-way bindings
+added while the branch is active: leaving a branch never reverts a value, so
+give the chain an ``else`` when something must come back.
+
+Repeated content
+~~~~~~~~~~~~~~~~
+
+Use ``for`` to build one copy of the body per item, kept in sync as the iterable
+changes:
+
+.. code-block:: kv
+
+    <FruitList@BoxLayout>:
+        orientation: 'vertical'
+        fruits: ['apple', 'pear', 'plum']
+        for fruit in self.fruits:
+            Label:
+                text: fruit
+
+Reassigning ``fruits`` re-renders the list -- no ``add_widget`` loop and no
+manual bookkeeping. The header accepts Python's *comprehension* syntax (what is
+valid inside a list comprehension, not statement syntax), so tuple targets and
+a trailing ``if`` filter work, and the loop variables are available in the
+body's expressions:
+
+.. code-block:: kv
+
+    for i, fruit in enumerate(self.fruits) if fruit != 'pear':
+        Label:
+            text: f'{i}: {fruit}'
+
+There is no ``for`` / ``else``; the idiomatic empty state is a paired
+``if not self.fruits:`` block next to the loop.
+
+Add a ``key:`` line to give each item a stable identity: same key, same
+widgets. Existing widgets (and their live state, such as text being edited)
+are kept and moved rather than rebuilt when the list changes, and an item
+replaced under a stable key refreshes the kept widgets through their bindings:
+
+.. code-block:: kv
+
+    for fruit in self.fruits:
+        key: fruit
+        TextInput:
+            text: fruit
+
+Slots and factory
+~~~~~~~~~~~~~~~~~
+
+Two further blocks round out the set.
+
+``slot`` declares a named, fillable hole in a reusable container rule, with
+fallback content that the call site can override -- the same idea as Vue slots
+or the web-components ``<slot>``. Build the shell once and drop different
+content into it at each call site, without subclassing:
+
+.. code-block:: kv
+
+    # a reusable frame with a header slot and a default (body) slot
+    <Card@BoxLayout>:
+        orientation: 'vertical'
+        slot header:
+            Label:                       # fallback, shown if the caller fills nothing
+                text: 'Untitled'
+        slot:                            # default slot: plain children land here
+
+    # call site -- fill the header, drop the body content straight in
+    Card:
+        slot header:
+            Label:
+                text: 'Inbox'
+                bold: True
+        Label:                           # plain child, routed into the default slot
+            text: 'You have 3 new messages'
+
+A slot definition may also declare *slot-scoped locals* -- reactive values
+computed by the defining rule and read by fallback and fill content through
+the reserved ``slot`` name (``text: slot.title``) -- so a shell can hand data
+to whatever content the caller drops in.
+
+``factory <expr>`` builds a single child whose class is chosen by an expression
+(a class object, or a name resolved through the :class:`~kivy.factory.Factory`),
+rebuilt when the class changes. Because the expression is ordinary Python,
+picking a class at runtime needs no extra machinery:
+
+.. code-block:: kv
+
+    <MessageList@BoxLayout>:
+        orientation: 'vertical'
+        messages: []
+        for message in self.messages:
+            factory 'StrongLabel' if message.unread else 'Label':
+                text: message.text
+
+Learn more
+~~~~~~~~~~
+
+Control statements come with a handful of rules -- for example they cannot
+appear at the top level of a file, and inside a ``canvas`` block only ``if`` and
+``for`` are allowed. For the full description, including reactive ids,
+iteration-local values and graphics control flow inside ``canvas``, see the
+:ref:`control statements <kv_control_statements>` reference.
+
 Re-using styles in multiple widgets
 -----------------------------------
 
--- a/doc/sources/sphinxext/kivy_pygments_theme.py
+++ b/doc/sources/sphinxext/kivy_pygments_theme.py
@@ -12,7 +12,10 @@ class KivyStyle(Style):
     styles = {
         # No corresponding class for the following:
         #Text:                     "", # class:  ''
-        Whitespace:                "underline #ffffff",      # class: 'w'
+        # Leave whitespace unstyled: recent Pygments emits Whitespace ('w')
+        # tokens for interior spaces too, so any decoration here (an underline
+        # used to) shows up under every space in a code block.
+        Whitespace:                "",                       # class: 'w'
         Error:                     "#FF0000 border:#FF0000", # class: 'err'
         Other:                     "#FF0000",                # class 'x'
 
--- a/examples/kv/app_todo.kv
+++ b/examples/kv/app_todo.kv
@@ -1,0 +1,137 @@
+#:import uuid uuid
+
+# A tiny todo app written entirely in kv, showcasing the control statements:
+# slot (with a slot-scoped local and ids in fills), if/else, for (with key:
+# and an iteration-local) and factory.
+# Run it with:  python3 kvrun.py app_todo.kv
+#
+# State is a {uid: task} dict. A kv dict literal is a DictProperty, which
+# dispatches on in-place key ops -- so one task can be updated/added/removed
+# with `tasks[uid] = ...` / `del tasks[uid]`, no whole-container copy.
+# Note the dict.items(x) / dict.values(x) form: passing the property as a bare
+# argument lets kv bind it (x.items() would hide `x` behind the call and the
+# binding would never fire) -- the usual kv binding gotchas apply to control
+# headers too.
+
+# the two presentations the factory switches between per task
+<TodoLabel@Label>:
+    task: None
+    halign: 'left'
+    valign: 'middle'
+    text_size: self.size
+    text: root.task['text'] if root.task else ''
+
+<DoneLabel@TodoLabel>:
+    markup: True
+    color: .5, .5, .5, 1
+    text: ('[s]' + root.task['text'] + '[/s]') if root.task else ''   # struck through
+
+# --- a reusable framed panel with a customizable header (slot) ---
+<Panel@BoxLayout>:
+    # declared here (not just on the instance) so kv infers a DictProperty,
+    # which dispatches on in-place key ops; an instance-only `tasks:` would be
+    # a plain ObjectProperty and setitem would not re-render
+    tasks: {}
+    orientation: 'vertical'
+    padding: dp(12)
+    spacing: dp(8)
+    canvas.before:
+        Color:
+            rgba: .12, .12, .14, 1
+        Rectangle:
+            pos: self.pos
+            size: self.size
+    slot header:
+        # slot-scoped local: computed by Panel, readable by any header fill
+        # (and the fallback below) as `slot.summary`
+        summary: f'{len(root.tasks)} task(s)'
+        Label:                                  # fallback, overridden below
+            text: slot.summary
+            size_hint_y: None
+            height: dp(32)
+    slot:                                        # default slot: body lands here
+
+
+# --- the app ---
+Panel:
+    tasks: {uuid.uuid4(): {'text': 'Buy milk', 'done': False}, uuid.uuid4(): {'text': 'Write a kv example', 'done': False}, uuid.uuid4(): {'text': 'Tick me off', 'done': False}}
+    # reactive aggregate; dict.values(self.tasks) keeps `self.tasks` bindable
+    done: len([t for t in dict.values(self.tasks) if t['done']])
+
+    slot header:                                 # fill the panel's header slot
+        Label:
+            # the fill reads the slot-scoped local computed by Panel
+            text: f'My tasks - {slot.summary}, {root.done} done'
+            font_size: '20sp'
+            size_hint_y: None
+            height: dp(32)
+
+    slot:                                        # explicit default fill: ids are
+        BoxLayout:                               # allowed on explicit slot content
+            orientation: 'vertical'
+            spacing: dp(6)
+
+            BoxLayout:                           # input row
+                size_hint_y: None
+                height: dp(40)
+                spacing: dp(8)
+                TextInput:
+                    id: field                    # reactive id on this rule's scope
+                    hint_text: 'Add a task and press Enter'
+                    multiline: False
+                    focus: True
+                    on_text_validate:
+                        if field.text.strip(): root.tasks[uuid.uuid4()] = {'text': field.text, 'done': False}
+                        field.text = ''
+                        field.focus = True
+                Button:
+                    text: 'Add'
+                    size_hint_x: None
+                    width: dp(70)
+                    # `field` is a reactive id (None until the TextInput
+                    # exists), hence the trailing guard. And as everywhere in
+                    # kv, `field.text` must end the dotted chain for kv to
+                    # bind it -- a bare `field.text.strip()` would never
+                    # re-evaluate (see the kv gotchas), so keep the parens.
+                    disabled: not (field.text or '').strip() if field else True
+                    on_release: field.dispatch('on_text_validate')
+
+            if not root.tasks:                   # empty state / hint (if / else)
+                Label:
+                    text: 'No tasks yet. Add one above.'
+                    size_hint_y: None
+                    height: dp(28)
+            else:
+                Label:
+                    text: 'Tick to complete, press x to delete.'
+                    size_hint_y: None
+                    height: dp(28)
+
+            for uid, task in dict.items(root.tasks):     # one row per task
+                key: uid                         # same key => same widgets; a
+                                                 # replaced task re-dispatches
+                                                 # through the kept row's bindings
+                kind: 'DoneLabel' if task['done'] else 'TodoLabel'   # iteration-local
+                BoxLayout:
+                    size_hint_y: None
+                    height: dp(36)
+                    spacing: dp(8)
+                    CheckBox:
+                        size_hint_x: None
+                        width: dp(40)
+                        activated: task['done']          # note: 'activated', not 'active'
+                        on_activated:
+                            # in-place update of one task -- the DictProperty
+                            # dispatches, the row is kept (stable key) and its
+                            # bindings receive the new task value
+                            root.tasks[uid] = dict(task, done=self.activated)
+                    factory kind:                # done vs todo presentation
+                        task: task
+                    Button:
+                        text: 'x'
+                        size_hint_x: None
+                        width: dp(36)
+                        on_release:
+                            del root.tasks[uid]
+
+            Widget:                              # spacer: keep content at the top
--- a/kivy/extras/highlight.py
+++ b/kivy/extras/highlight.py
@@ -3,7 +3,7 @@
 from pygments.lexer import RegexLexer, bygroups, using
 from pygments.lexers.agile import PythonLexer
 from pygments import highlight
-from pygments.token import Comment, Text, Name, Punctuation, Operator
+from pygments.token import Comment, Text, Name, Punctuation, Operator, Keyword
 from pygments.formatters import get_formatter_by_name
 import sys
 
@@ -22,6 +22,19 @@ class KivyLexer(RegexLexer):
                 bygroups(Punctuation, Text, Name.Class, Text, Operator),
                 'classList'),
             (r'[A-Za-z][A-Za-z0-9]*$', Name.Attribute),
+            # control statements: keyword + a Python-highlighted header, e.g.
+            # "if self.x:", "for item in self.items:", "factory expr:".
+            # These must precede the widget-definition rule below, which would
+            # otherwise paint the whole header as a single Name.Class token.
+            (r'(if|elif|for|factory)(\s+)(.+?)(\s*)(:)(\s*)$',
+                bygroups(Keyword, Text, using(PythonLexer), Text,
+                         Punctuation, Text)),
+            (r'(else)(\s*)(:)(\s*)$',
+                bygroups(Keyword, Text, Punctuation, Text)),
+            (r'(slot)(\s+)([A-Za-z_][A-Za-z0-9_]*)(\s*)(:)(\s*)$',
+                bygroups(Keyword, Text, Name, Text, Punctuation, Text)),
+            (r'(slot)(\s*)(:)(\s*)$',
+                bygroups(Keyword, Text, Punctuation, Text)),
             (r'(.*?)(\s*)(:)(\s*)$',
                 bygroups(Name.Class, Text, Punctuation, Text)),
             (r'(.*?)(\s*)(:)(\s*)(.*?)$',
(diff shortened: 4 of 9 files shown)


ElliotGarbus:
This is strong and considered work. 

I'll be upfront about two things: my standing, and the practical situation, because both matter for how you should weigh what follows.

**The practical reality.** I'm a contributor, not a maintainer — I can't merge this, and I don't want to imply there's a review process queued up to adjudicate it. A language change of this size realistically needs the maintainer who owns kv's core and maintainer bandwidth is thin right now — I wouldn't count on quick review.  I'd rather flag that now than have you refine ids and exception semantics for months against a review that may not come soon. You should get to make decisions about your own work with the real picture.  I will add a few maintainers to the review list and they can weigh in on their own - though realistically I wouldn't count on quick attention..

**My read, as one contributor's opinion.** The thing I keep coming back to is complexity budget, and I'll state it as a value rather than a fact. Making ids survive control flow requires three id-visibility tiers with three lifetimes, reactive nullable ids, parse-time reference rewriting, and a compare-and-set sharing rule. Each piece is well-designed — and each is permanent surface that every kv user has to learn and every maintainer has to keep correct. My honest reservation is that kv earns its keep by staying small enough to hold in your head, and this is a large, hard-to-reverse expansion to deliver expressiveness Python already provides (you can build widgets in a loop today). Reasonable people weight that differently; you may be right that the declarative version is worth it. But that's the axis I'd want decided deliberately, and it doesn't hinge on any single unresolved detail.

And I want to ground that rather than wave at it. Here's a `DataInputs` widget from something I built recently —  data-driven, with set_data/get_data/clear:

```python
class DataInputs(BoxLayout):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        for text in ['Teacher', 'School', 'Class', 'Date', 'Contacts']:
            self.add_widget(LabeledInput(text=text))

    def get_data(self):
        return {c.ids.lbl.text: c.ids.ti.text for c in self.children}

... # set_data and clear are similar... 
```

Two things I take from it. The construction loop is the *smallest, most trivial* part — the class exists for the behavior (`get_data`/`set_data`/`clear`), and the loop is four incidental lines a `for` keyword would absorb while getting in the way of the rest. And identity is carried by *data*: each `LabeledInput` owns its ids at its own scope, and the label text is literally the dict key. Well-structured kv rarely needs to *name* a loop-generated widget at all. The elaborate id apparatus in this PR is solving, carefully, a problem that idiomatic component structure mostly sidesteps.

So my honest assessment of the value: for those of us comfortable dropping to Python, this pattern is trivial enough that in years of building I've never wanted grammar for it. I'll be fair that the residual value isn't zero — it's real for people who build primarily *in kv* and don't reach for Python. But my read is that slice is narrow, and I don't think it justifies four keywords plus a scope model as permanent language surface.

**The one place I'd actively point elsewhere.** The `hide` case that motivates some of this doesn't want control flow at all. Hiding wants collapse-and-restore with state preserved, not teardown-and-rebuild — which points to a filtering `visible` property (the CSS `display: none` analog) rather than structural `if`. It means teaching the core layouts to honor visibility instead of iterating raw `self.children`. A property on `Widget` plus work in a handful of core layouts — a much smaller conceptual surface than four keywords and a scope model. 

**One coupled technical note, if it does move forward.** The reactive-`None` ids *increase* the exception surface rather than being orthogonal to it: any reference that isn't None-guarded (`text: field.text`, not just `field is None`) evaluates against `None` while its branch is inactive, so mid-evaluation raises become common, not rare — which makes the expression-error handling on the shared apply path load-bearing, and worth *showing* exercised rather than describing. The "points at whichever mounted last" sharing rule similarly leans on a reconciliation order I'd want specified and shown stable if a raise interrupts a pass mid-way. The ids model and the exception model are really one question.

To be clear about what this is: not a rejection — I can't reject it and I'm speaking only for myself. It's a straight read from someone who spent real time with the design. The work is good; my honest opinion is that the language cost is more than I'd want kv to take on for value this narrow, and that the hide case has a smaller home worth pursuing on its own. Happy to help with that piece if it's useful.

mtctn:
Thank you @ElliotGarbus for your constructive feedback.

I'll try to clarify some points.

**About the feature itself and the value added (regardless of the implementation and architecture).** In my opinion one strength of kv is that it is much more *expressive* than direct Python counterpart, allowing to declare, instantiate, and bind widgets and properties in a much more concise way than in Python.

When it comes to defining a clear separation of concerns in a project between what is defined in Python and what is defined in kv, two models can hold:

- either define all the widget tree on Python side, and use kv as a kind of stylesheet at component level (think CSS), current kv-lang is perfect for that, but it forces you to use less expressive structures to glue all the widgets and properties on Python side
- or use Python mainly for the business logic, and hand all the templating/interface definition to kv. This model takes full benefits of kv expressiveness, but is limited by the lack of control structures

The third way, which is the solution I currently turn to in many cases, is to define anything I can on kv side, and revert to Python when kv can not express what I want anymore, but it leads to inconsistent codebases and is a bit frustrating. 

To me your DataInput example is good to advocate for the need of standardized control structures. To implement such cases you are forced to add a lot of verbose boilerplate: Python class, getters, setters, reactive bindings, widget tree management.

All that can become clean, readable, and reactive-for-free if you want your list input to be reactive:

```
<DataInputs@BoxLayout>:
    for text in ['Teacher', 'School', 'Class', 'Date', 'Contacts']:
        LabeledInput:
            text: text
```

Zen of Python says: "There should be one-- and preferably only one --obvious way to do it." To me Kivy lacks obvious ways for these kinds of use cases, so that solution is standard and not per project.

Having control structure in a project's templating engine is the standard in most frameworks (Django, Jinja2, React, Vue to name only the ones I use most), so I'd bet that the slice of developers wanting to use such structures is not that narrow.

**About the architecture.**
First totally agree that introducing a `visible` property honored by layouts is a very good step towards filling the gap. Just a nuance: show/hide (visible) and create/tear-down (my if) are two different features, both of interest depending on the use case.
 
As you mentioned on discord, another possible architecture for control structures is the QML way with iteration/conditionals as reactive components (Conditional, Repeater). Having a `visible` property would allow to define invisible anchors in a Layout's children as the point of insertion (and managing widget) of controlled children. I didn't go in this direction since it appeared to me more invasive (requires updating the contract of all layouts), but perhaps less burden from a maintainer's point of view.

I tried in my proposal to find a good balance between feature-rich and contained API surface, but I am completely open to discuss this, for instance reactive ids could be dropped for a first version.

About reactive ids: agree with your comment. I implemented at some point a two-step resolution on reactive updates to ensure that the id would never flip to None mid-update when switching between branches but it think I lost it during a refactor and then forgot this idea. Definitively need to implement this so that when if/elif/else or parallel ifs define the same id, it never goes to `None` mid-flight.
The "points at whichever mounted last" decision is to keep things simple and be coherent with legacy kv design that allows redefinition of the same id multiple times without guards (but, I agree, has a quite predictable resolution order when no control structure are defined).

Issue #8625, opened 2024-02-20
issue with textinput keyboard_suggestions or InputType on android when using backspace


Dro12u:
**Software Versions**
* Python: 3.10.4
* OS: windows10/ ubuntu 22.04
* Kivy: v2.3.0
* tested on the latest version of android

**Describe the bug**
basically what happens is when i have inputtype and/or keyboard_suggestions set, the backspace does some weird behavior shown in the video attached. this doesn't happen on windows only on phone/tablet 

**Expected behavior**
it would be nice to have the keyboard_suggestions feature working properly for my users.

**To Reproduce**
issue happens when i have my textinput set like this:
         self.text_input = TextInput(input_type='text',hint_text=f"{self.instructions_message}", keyboard_suggestions=True, multiline=True,  size_hint=(1, 0.8) )
or this:
         self.text_input = TextInput(input_type='text', hint_text=f"{self.instructions_message}", multiline=True,  size_hint=(1, 0.8) )

**Code and Logs and screenshots**

https://github.com/kivy/kivy/assets/149750354/7abec06b-8510-4ae6-bba2-219dd1b7cf38

```

**Additional context**
tried to download latest version of sdl 
tried to use develop instead of master for p4a.branch in my .spec file 
the only way to have this behavior fixed is by fully removing the inputtype and the keyboard_suggestions from the textinput

misl6:
I do not think we have a specific one, even if this issue already showed up somewhere, so I will keep this open.
However, you may find it interesting to check out [this recap](https://github.com/kivy/kivy/issues/8575).

PR #9344, opened 2026-07-15
`Component: TextInput` Fix TextInput suggestions bugs

### LINKED ISSUE #9167: Autocompletion makes TextInput working abnormally after deleting characters or changing suggestions
**Software Versions**

* Python: 3.11.5
* OS: Android 12.0
* Kivy: 2.3.0/2.3.1
* Kivy installation method: Buildozer (all settings default)

**Describe the bug**

On Android, when entering text into the `TextInput`widget, with autocompletion enabled in your keyboard, after a few selections and/or deletions of suggested words or simply deleting previously entered words or characters, the text input becomes random and chaotic: typed characters appear not in the position of the cursor, and the user can't either write or delete what they need. The whole TextInput field turns into complete gibberish.

This problem instantly disappears if you just close (unfocus) and then open (focus) the field again. Then you can again type normally - until the bug appears next time, which is almost inevitable if you write a relatively long text and use autocompletion actively.

This behavior is only present when autocompletion is on. If you disable autocompletion in the keyboard settings or make `input_type = 'null'` in your TextInput properties, the issue is absent. So this is somehow connected with the way Kivy (SDL?) processes autocompletion, especially when you press backspace or use a different autocompleted suggestion than you typed manually.

**Expected behavior**

`TextInput` should allow the user to type what they want correctly at all times, using autocompletion if it's on.

**To Reproduce**

There are various ways how this bug appears, the easiest way is typing some word incorrectly and then click on the suggested word which would be correct, and do this a few times. For example: type "I wanr", then click on "want" in the suggested words. Repeat 2-3 times. Then you will see that the input is broken: letters appear in wrong positions, and when you click backspace, they are deleted in wrong positions too. You actually can't type anything anymore.

I observe this bug in the simplest possible app with TextInput using the default, non-edited `buildozer.spec` file, generated as is.

Here is my code:

```python
from kivy.app import App

class MyApp(App):
    def build(self):
        layout = BoxLayout()
        text_input = TextInput(multiline=True, input_type="text")
        layout.add_widget(text_input)
        return layout

if __name__ == '__main__':
    MyApp().run()
```
As you can see from the screenshot below, instead of "want", "wat r" appears:

Image

After this point, you can't type anything properly anymore.

See my video on YouTube where I reproduce this scenario using the aforementioned code, where I type "wanr/want" 2 times. In the end, the whole text becomes totally broken in multiple unexpected positions: [https://youtube.com/shorts/uayo6GTt5ag?feature=share](https://youtube.com/shorts/uayo6GTt5ag?feature=share).

(If you can't access the YouTube link, use this alternative OneDrive link: [SVID_20250929_092344_1.mp4](https://1drv.ms/v/c/e77e42402671b083/EQJtFwqQTPBPo0pvUtkyGNwBvX34RdqthEDCntIb4ihx7Q?e=sty1TT).)

**Additional context**

My device is Huawei P20, but I received reports about this issue from numerous users from various countries, since it appears in my app on Google Play Store. 

I've observed this issue since 2022 and very surprised it hasn't been addressed for such a long time, even after a few Kivy updates. I think it's a major issue: you can't work with text normally in Kivy apps, at least with autocompletion. A lot of my users are confused, many of them deleted the app because they just can't work. I look forward when it will be finally fixed.

Novfensec:
I mean literally so simple fix it seems to be.
Works for me!! I hope it works for all. Just create an inherited textinput and override that func and see it working!!!!

Maintainer merge checklist

Closes #9167 #8625 

https://github.com/user-attachments/assets/a370b640-c365-4071-aa65-d762b496974a

### DIFF
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -3058,6 +3058,9 @@ class TextInput(FocusBehavior, Widget):
         Window.unbind(on_textedit=self.window_on_textedit)
 
     def window_on_textedit(self, window, ime_input):
+        if not self._lines:
+            self._refresh_text(self.text)
+
         if self._ime_composition:
             ci = self.cursor_index()
             len_ime = len(self._ime_composition)
--- a/kivy/uix/textinput.py
+++ b/kivy/uix/textinput.py
@@ -3061,24 +3061,50 @@ class TextInput(FocusBehavior, Widget):
         if not self._lines:
             self._refresh_text(self.text)
 
-        if self._ime_composition:
-            ci = self.cursor_index()
-            len_ime = len(self._ime_composition)
-
-            self._selection_from = max(0, ci - len_ime)
-            self._selection_to = ci
-            self._selection = True
-
+        if self._selection:
             self.delete_selection()
 
-        if ime_input:
-            if self._selection:
+        current_ci = self.cursor_index()
+        cursor_restored = False
+
+        if self._ime_composition:
+            ci = getattr(
+                self, "_ime_absolute_cursor", self.cursor_index(self._ime_cursor)
+            )
+            start_ci = max(0, ci - len(self._ime_composition))
+
+            if self.text[start_ci:ci] == self._ime_composition:
+                self._selection_from = start_ci
+                self._selection_to = ci
+                self._selection = True
                 self.delete_selection()
 
+                if current_ci >= ci:
+                    current_ci -= len(self._ime_composition)
+                elif current_ci > start_ci:
+                    current_ci = start_ci
+                cursor_restored = True
+
+            else:
+                start_ci_curr = max(0, current_ci - len(self._ime_composition))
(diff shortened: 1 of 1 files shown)


Novfensec:
Should I fix tests also now. Please let this assigned to someone else I can't handle tests!!!

kuzeyron:
That would be great.

Novfensec:
I see a new issue if using periods the period is not removed from text history

Novfensec:
Definitely works with suggestion now without any prior bugs from my side. I request fellow devs to test the changes proposed themselves and point out bugs still remaining!!

kuzeyron:
Okay, I just tested this on one of my Android devices. The bug is still there when you select and copy some text and then when you paste it goes nuts.

rodrufo:
Hi guys, i`m om my first project using kivy and facing this bug. No solution?

PR #9345, opened 2026-07-18
kivy.mobile Android backend for Window geometry, metrics scaling, and keyboard input type

ElliotGarbus:
## Summary

Implements the Android backend for `kivy.mobile` - the platform-neutral mobile
window/display geometry API introduced in #9331 - replacing the placeholder in
`kivy/mobile/_platform/android.py`.

All values are read at runtime by reflecting against the running
`PythonActivity.mActivity` via `jnius` (a standalone Kivy-org package present in
every Kivy Android build). **No compiled extension and no python-for-android
changes are required.**

### Implemented

Tier-1 (cross-platform API):
- `get_dpi()` - physical screen DPI (`densityDpi`); matches `Metrics.dpi`
- `get_scale()` / `get_density()` - display scale factor (`scaledDensity`)
- `get_keyboard_height()` - soft-keyboard (IME) height in px, 0 when hidden
- `get_safe_area()` - system-bar insets unioned with the display cutout
- `subscribe_keyboard_height(callback)` - keyboard-height change notifications,
  driven by a lazily-scheduled Kivy `Clock` poll (callback runs on the Kivy main
  thread, so it can update Kivy properties directly)

Tier-2 (Android extras):
- `get_display_cutout()` - cutout bounding rects, or `None`
- `get_system_bar_insets()` - status-bar and nav-bar insets, separated

### Implementation notes

- Lengths are returned in **pixels** (Kivy's Android layout coordinate system;
  density is folded into `kivy.metrics.Metrics`, not window coordinates).
- `WindowInsets` / `DisplayCutout` reads are marshalled onto the Android UI
  thread via `Activity.runOnUiThread` (Kivy/SDL runs on a separate thread);
  `DisplayMetrics` is read directly since it is thread-safe.
- All getters degrade gracefully (fallback values / `None`) on error.

### API level / compatibility

- `get_dpi()`, `get_scale()`, `get_density()` use `DisplayMetrics` and work on
  **any** Android API level.
- The inset-based reads (`get_keyboard_height`, `get_safe_area`,
  `get_system_bar_insets`) use `WindowInsets.getInsets(type)`, added in
  **API 30 (Android 11)**. `android.view.WindowInsets$Type` is resolved lazily,
  so on API < 30 the module still imports and the DisplayMetrics reads keep
  working; the inset getters return zeros/`None` and emit a one-time warning.
- Method resolution is via reflection, so the build/compile API level is
  irrelevant - only the runtime device API matters.

## Test plan / validation

Validated on a physical **Pixel 8a** using a Kivy probe app that displayed each
API's live value and cross-checked it against `kivy.metrics.Metrics`:

- `get_dpi()` / `get_scale()` / `get_density()` match `Metrics`.
- `get_keyboard_height()` reports correctly (0 hidden, non-zero shown), and
  `subscribe_keyboard_height` fires on show/hide.
- `get_safe_area()`, `get_display_cutout()`, and `get_system_bar_insets()`
  validated in both portrait and landscape.

The logic was first validated via a byte-identical standalone prototype on the
device, then transplanted here. An integration build (this change compiled into
an APK with python-for-android + SDL3) confirmed Kivy packages `kivy.mobile`
correctly and imports the backend.

### DIFF
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -685,7 +685,7 @@ class WindowBase(EventDispatcher):
 
     def _refresh_safe_area(self, *args):
         """Update Window.safe_area from kivy.mobile.get_safe_area()."""
-        if platform not in ('ios', 'android'):
+        if platform not in {'ios', 'android'}:
             return
         from kivy.mobile import get_safe_area
         self.safe_area = get_safe_area()
@@ -709,7 +709,7 @@ class WindowBase(EventDispatcher):
         return 0
 
     def _get_kheight(self):
-        if platform in ('android', 'ios'):
+        if platform in {'android', 'ios'}:
             from kivy.mobile import get_keyboard_height
             return get_keyboard_height()
         return self._get_kivy_vkheight()
@@ -1196,7 +1196,7 @@ class WindowBase(EventDispatcher):
         # next frame so the first value isn't stuck at all-zeros.
         Clock.schedule_once(self._refresh_safe_area, 0)
 
-        if platform in ('ios', 'android'):
+        if platform in {'ios', 'android'}:
             from kivy.mobile import subscribe_keyboard_height
             # subscribe_keyboard_height fires on every IME height change; route
             # it through trigger_keyboard_height for parity with iOS. That is a
--- a/kivy/core/window/__init__.py
+++ b/kivy/core/window/__init__.py
@@ -35,7 +35,6 @@ from kivy.core.image import ImageLoader
 
 # late import
 VKeyboard = None
-android = None
 Animation = None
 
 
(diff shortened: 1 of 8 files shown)


ElliotGarbus:
## Keyboard input type now goes through SDL3 text-input properties
I moved the on-screen keyboard's input-type configuration off the
python-for-android `android` module. Instead of
`PythonActivity.changeKeyboard(...)`, `show_keyboard()` in
`_window_sdl3.pyx` now uses SDL3's `SDL_StartTextInputWithProperties`:
- A coarse, cross-platform `SDL_TextInputType`
  (`SDL_PROP_TEXTINPUT_TYPE_NUMBER`) drives the iOS keyboard.
- On Android I additionally set
  `SDL_PROP_TEXTINPUT_ANDROID_INPUTTYPE_NUMBER` with the full
  `android.text.InputType` value, so we keep the finer-grained mappings
  (`url` / `tel` / `datetime`, plus best-effort no-suggestions) that the
  coarse SDL enum can't express.

Why this layer: input type is a text-input/windowing concern, not device
geometry, so it doesn't belong in `kivy.mobile`. Routing it through SDL
keeps the mapping in the SDL3 window backend and removes the **last
keyboard dependency** on the p4a `android` Python module — no more
`changeKeyboard` or  jnius, involvement. It also matches
SDL3's intent: the SDL3 bootstrap stubs out `changeKeyboard` precisely
because this path exists.

Note: this requires **SDL 3.2+** (where `SDL_StartTextInputWithProperties`
and `SDL_TextInputType` landed). The needed bindings were added to
`kivy/lib/sdl3.pxi`.

## Follow-up PR: the app-lifecycle `android` calls that should move

This PR is deliberately scoped to window/display geometry + keyboard, so
it does **not** touch the remaining direct uses of the p4a `android`
module. Those are app-lifecycle / back-key concerns, not geometry:

- `kivy/app.py:1110` — `mActivity.finishAndRemoveTask()`
- `kivy/app.py:1133` — `mActivity.moveTaskToBack(True)`
- `kivy/core/window/__init__.py:2090` — `mActivity.moveTaskToBack(True)`
- `kivy/core/window/window_sdl3.py:252` — `mActivity.finishAndRemoveTask()`

These should move behind `kivy.mobile._platform.android` in a dedicated
follow-up — they're genuinely Android-only actions (no iOS analogue), so
they stay in the Android backend rather than on the neutral `kivy.mobile`
surface. 

Consolidating the platform-specific code into one clear location
(`kivy.mobile`) is the larger win here: it breaks Kivy's built-in dependency
on the p4a `android` bootstrap module and confines the Java/jnius coupling to
a single, well-defined seam — which in turn makes it far simpler to support
alternative bootstraps (e.g. kivyforge alongside p4a), since each only has to
satisfy that one contract rather than scattered `from android import ...`
call sites across the codebase. I've kept this out of the current PR to avoid
conflating lifecycle / back-key behavior changes with the geometry + keyboard
work here.  I will follow up with this additional PR shortly.

kuzeyron (dismissed):

kuzeyron (approved):

kuzeyron:
LGTM! Thank you!

PR #9350, opened 2026-07-22
Accept pathlib.Path in path-accepting APIs

ElliotGarbus:
## Summary

Kivy's path-accepting APIs historically required a `str`. This PR lets them also accept any `os.PathLike` (e.g. `pathlib.Path`), by coercing `Path -> str` via `os.fspath()` at a small set of central choke points. All internals (string parsing, cache keys, Cython `.encode()`) stay string-based, so provider `.pyx` code is untouched.

Coercion only fires for `isinstance(x, os.PathLike)`. A `str` is never `os.PathLike` (even a path-shaped one), so plain strings — including URIs like `atlas://`, `http://`, `data:` — and `BytesIO` inputs pass through unchanged.

## Changes

- **`kivy/utils.py`**: new `path_to_str()` helper (coerces `os.PathLike` -> `str`; leaves `str`/`bytes`/`None`/file-like/URIs unchanged).
- **`kivy/properties.pyx`**: `StringProperty.convert()` now coerces `os.PathLike` -> `str`. This one change makes every path-bearing `StringProperty` accept `Path` (`Image.source`, `font_name`, `App.kv_file`, the `background_*`/atlas image sources, `Window.icon`, ...). A non-path, non-`str` value still raises `ValueError`.
- **`kivy/resources.py`**: coerce in `resource_find`, `resource_add_path`, `resource_remove_path` (cache keys / path-list membership stay consistent).
- **core loaders**: `Image`/`ImageLoader` (`load`/`save`/`filename`), `SoundLoader.load`, `VideoBase.filename`/`generate_thumbnail`, `LabelBase.resolve_font_name`/`register`, `SvgLoader.load`, Lottie filename.
- **lang / atlas / storage**: `Builder.load_file`/`unload_file`/`load_string`, `Atlas`, `JsonStore`, `DictStore`.
- **app / config**: `App.load_kv`/`get_application_config`; `Config.read` now accepts `os.PathLike` while still rejecting multiple filenames.
- Docstrings updated with `.. versionchanged:: 3.0.0` notes.
- New tests: `kivy/tests/test_pathlib_support.py`, parametrized over `str` vs `Path`.

## Note for reviewers: global `StringProperty` coercion

[...]

PR #9351, opened 2026-07-22
Normalize resource cache keys so path spellings share one entry

ElliotGarbus:
## Summary

`resource_find` (`kv.resourcefind`) and `Loader.image` (`kv.loader`) keyed their result caches on the **raw input string**. As a result, different spellings of the *same* file each created a separate cache entry and re-resolved / re-fetched it:

- separator variants: `'a/b.png'` vs `'a\\b.png'`
- redundant segments: `'a/./b.png'`, `'a/../a/b.png'`
- case differences on case-insensitive platforms (Windows)
- `pathlib.Path` inputs (which coerce to the OS-native separator form)

This was a pre-existing dedup/efficiency miss (not a correctness bug — `resource_find` → `_resolve_path` → `abspath` already resolves every spelling to the same file). This PR closes it so all spellings share **one** cache entry.

## Changes

- **`kivy/resources.py`**: new `_cache_key()` helper. `resource_find` now keys `kv.resourcefind` on `os.path.normcase(os.path.normpath(filename))` for filesystem paths. URIs (`atlas://`, `data:`, and the `@image_provider:...` scheme) are keyed **verbatim** so `normpath` never collapses their `//` / mangles scheme separators.
- **`kivy/loader.py`**: new `_loader_cache_key()` helper applying the same normalization. Crucially, the async loader now separates the **key** (normalized — used for the cache entry *and* the in-flight client-identity match) from the **source** (the raw filename that is actually loaded). URLs, `atlas://` and case-sensitive paths are therefore never rewritten. The queue dict, `_q_done` tuples and `_client` tuples now carry the normalized key.
- **Docstrings**: `.. versionchanged:: 3.0.0` notes on `resource_find` and `Loader.image`.
- **New tests**: `kivy/tests/test_resource_cache_key.py` — cross-spelling dedup (verifies a single filesystem resolution / no duplicate in-flight load across `'/'`, native-sep, `Path`, and redundant-segment spellings), platform-aware case folding, and URI/URL passthrough for both caches.

## Stacking

This is the follow-up promised in #9350 (which documented the spelling-duality nuance and left it to a separate PR to avoid conflating "accept `Path`" with a change to cache-keying semantics).

**It is stacked on #9350** and its diff currently includes that PR's commit (`Accept pathlib.Path in path-accepting APIs`). Please merge #9350 first; this PR should then reduce to the single cache-key commit. The `Path`-dedup assertions here rely on the coercion added by #9350.

## Test plan

- [x] `pytest kivy/tests/test_resource_cache_key.py` — 16 passed locally (incl. the in-flight loader dedup test where an image provider was available)
- [x] `pytest kivy/tests/test_pathlib_support.py -k "resource or config or store or builder"` — no regression
- [x] `ruff check` on touched files
- [ ] CI matrix (Windows/macOS/Linux × py3.11–3.14) for compiled-provider + cross-platform coverage


kuzeyron (dismissed):

kuzeyron (approved):

kuzeyron:
LGTM! Thank you!

PR #9352, opened 2026-07-25
kivy.mobile: pull the Android Activity from the bootstrap instead of a hardcoded PythonActivity

ElliotGarbus:
Kivy's Android backend reflects a class name it hardcodes,
`org.kivy.android.PythonActivity`. That makes Kivy depend on a
python-for-android implementation detail, and it silently breaks any build whose
activity is named something else — including p4a's own `--activity-class-name`.

This inverts it. Kivy names no activity class: it imports a top-level
`_kivy_bootstrap` module, supplied by whoever built the APK, and pulls the
current Activity from it.

## The contract

| member | required | meaning |
| --- | --- | --- |
| `get_activity()` | yes | the current `android.app.Activity`, or `None` where none exists |
| `get_context()` | no | a `Context` for processes that never have an Activity |
| `remove_presplash()` | no | dismiss the boot splash; Kivy calls it after the first frame |

The module is imported on first use and `get_activity()` is called live on every
access, so nothing goes stale when Android recreates the Activity on rotation,
configuration change or process death — the reason this is a call into the
bootstrap rather than a value registered once.

`None` is a legitimate answer, since a background service has no Activity. A
*missing module* while an Android runtime is present raises
`ActivityProviderMissing` rather than degrading to plausible-looking defaults,
because it can only mean the APK was built by something that does not support
Kivy 3; the geometry getters deliberately re-raise it instead of swallowing it
with their fallbacks. Off-device, where jnius is absent, those getters keep
returning their documented defaults, so desktop imports and the test suite are
unaffected.

Kivy pulls rather than the bootstrap pushing a provider at startup. p4a runs the
app's `main.py` as the process entry point, so it has no Python of its own to
register from without a new C-level seam in every bootstrap; and registering
would mean importing Kivy before the app could set its `KIVY_*` environment and
config.

## What changed

| file | change |
| --- | --- |
| `kivy/mobile/_platform/android.py` | the contract: discovery, `get_activity`, `get_app_context`, `remove_presplash`, plus `run_on_ui_thread`, `get_files_dir`, `get_cache_dir` |
| `kivy/mobile/__init__.py` | Tier 3 public API (`ActivityProviderMissing`, `get_activity`, `get_app_context`) and its documentation |
| `kivy/app.py` | `user_data_dir` / `user_cache_dir` from the Context instead of `PythonActivity.mActivity` |
| `kivy/core/clipboard/clipboard_android.py` | dropped its private `_Runnable`, its own `run_on_ui_thread`, and stashing the clipboard on `PythonActivity` |
| `kivy/base.py` | `remove_android_splash` goes through the bootstrap hook |
| `kivy/tests/test_mobile.py` | fake-bootstrap harness and contract tests |

`remove_presplash()` is optional because dismissing a splash belongs to the
bootstrap: p4a overlays a `View` and removes it, while a bootstrap using the
Android 12 system splash has no equivalent method to offer. Making it a
module-level hook restores the capability negotiation Kivy 2.x got from
`try: import android`, without Kivy naming a bootstrap or a Java method. An
absent hook is a no-op, not a warning.

After this, `grep -rn "PythonActivity\|mActivity" kivy/` matches nothing outside
the test harness.

## Tests

`kivy/tests/test_mobile.py` gains a fake `_kivy_bootstrap` harness in place of
the fake `autoclass`, and contract tests covering: that Kivy holds no bootstrap
class name; that a missing module, or one without `get_activity`, raises; that
the failure is not masked by the getters' defaults; that the Activity is pulled
live rather than cached; that `None` is accepted for a process without one; that
an optional `get_context()` is preferred when present; that storage paths come
from the Context; that `run_on_ui_thread` posts to the Activity; and that the
presplash hook is optional. A regression guard asserts that no bootstrap method
name reappears in the backend source.

## Verified on device

An x86_64 emulator running an APK built by python-for-android's `sdl3` bootstrap
from this branch (NDK r28c, SDL3 3.4.2, CPython 3.14.2). Each check prints one
marker:

```
CONTRACT: ACTIVITY_OK package=org.kivy.contract3
CONTRACT: PULLED_FROM_BOOTSTRAP_OK kivy.get_activity() is the bootstrap's Activity
CONTRACT: NO_CLASS_NAME_IN_KIVY_OK kivy asked _kivy_bootstrap; build chose org.kivy.android.PythonActivity
CONTRACT: APP_CONTEXT_OK package=org.kivy.contract3
CONTRACT: GEOMETRY_OK dpi=160.0 scale=1.00 density=1.00 fontscale=1.00 keyboard=0 ...
CONTRACT: STORAGE_OK user_data_dir=/data/user/0/org.kivy.contract3/files
CONTRACT: CLIPBOARD_OK provider=ClipboardAndroid round-trip exact
CONTRACT: PRESPLASH_OK bootstrap implements remove_presplash
CONTRACT: ALL_OK 8/8
```

`PULLED_FROM_BOOTSTRAP` asserts that `kivy.mobile.get_activity()` and
`_kivy_bootstrap.get_activity()` return the same Java object.
`NO_CLASS_NAME_IN_KIVY` asserts that the live Activity's Java class equals the
build's `android.config.ACTIVITY_CLASS_NAME` — the build's choice, not Kivy's.
`CLIPBOARD` exercises the coupling that started this.

## Bootstrap side

This needs a bootstrap that ships `_kivy_bootstrap`. The python-for-android
implementation is kivy/python-for-android#3356 — one new 64-line module in the
`android` recipe plus documentation for bootstrap authors, with no changes to
`start.c`, to any bootstrap, or to the build. Until it merges, testing on p4a
needs that branch. A second, Gradle- and wheel-based Android builder implements
the same contract against an unmodified Kivy, differing only in the ways the
contract allows: its own activity constant, and no `remove_presplash()` because
it uses the androidx core-splashscreen system splash.

Kivy 2.3.1 is unaffected — it does not import `kivy.mobile`, and p4a keeps its
existing `android` module for it.

---

Maintainer merge checklist

### DIFF
--- a/kivy/core/clipboard/clipboard_android.py
+++ b/kivy/core/clipboard/clipboard_android.py
@@ -7,61 +7,31 @@ Android implementation of Clipboard provider, using Pyjnius.
 
 __all__ = ('ClipboardAndroid', )
 
+import time
+
 from kivy import Logger
 from kivy.core.clipboard import ClipboardBase
-from jnius import autoclass, cast, PythonJavaClass, java_method
+from kivy.mobile._platform.android import get_app_context, run_on_ui_thread
+from jnius import autoclass
 
 AndroidString = autoclass('java.lang.String')
-PythonActivity = autoclass('org.kivy.android.PythonActivity')
 Context = autoclass('android.content.Context')
 VER = autoclass('android.os.Build$VERSION')
 sdk = VER.SDK_INT
 
-
-class _Runnable(PythonJavaClass):
-    '''Schedule a Python call onto the Android UI thread (fire-and-forget).
-
-    A tiny local replacement for python-for-android's ``android.runnable`` so
-    this provider depends only on pyjnius, not the p4a ``android`` module.
-    '''
-
-    __javainterfaces__ = ['java/lang/Runnable']
-    __javacontext__ = 'app'
-    # Keep strong references until the JVM has run each proxy, otherwise the
-    # pyjnius object may be collected before ``run()`` fires.
-    _refs = []
-
-    def __init__(self, func, args, kwargs):
-        super().__init__()
-        self._func = func
-        self._args = args
-        self._kwargs = kwargs
-        _Runnable._refs.append(self)
-
-    @java_method('()V')
-    def run(self):
-        try:
-            self._func(*self._args, **self._kwargs)
-        finally:
-            _Runnable._refs.remove(self)
-
-
-def run_on_ui_thread(f):
-    '''Decorator running *f* on the UI thread via ``Activity.runOnUiThread``.'''
-    def wrapper(*args, **kwargs):
-        PythonActivity.mActivity.runOnUiThread(_Runnable(f, args, kwargs))
-    return wrapper
+# The ClipboardManager, fetched once on the UI thread.  Process-wide rather than
+# per-instance because it comes from the Application context, which outlives any
+# one Activity.
+_clipboard = None
 
 
 class ClipboardAndroid(ClipboardBase):
 
     def __init__(self):
         super(ClipboardAndroid, self).__init__()
-        self._clipboard = None
         self._data = dict()
         self._data['text/plain'] = None
         self._data['application/data'] = None
-        PythonActivity._clipboard = None
 
     def get(self, mimetype='text/plain'):
         return self._get(mimetype).encode('utf-8')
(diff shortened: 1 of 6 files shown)


kivy/core/clipboard/clipboard_android.py
```diff
@@ -72,46 +42,57 @@ def put(self, data, mimetype='text/plain'):
...
     def _get(self, mimetype='text/plain'):
-        clippy = PythonActivity._clipboard
+        clippy = _clipboard
         data = ''
         if sdk < 11:
             data = clippy.getText()
         else:
-            ClipDescription = autoclass('android.content.ClipDescription')
             primary_clip = clippy.getPrimaryClip()
             if primary_clip:
```
kuzeyron, line 84: This if-statement is not needed because you have try/except next to it

ElliotGarbus: Done.

ElliotGarbus:
Note to self:  After this is merged, update android_clipboard.py to use threading.Event with a timeout instead of polling

kuzeyron (approved):

PR #9355, opened 2026-07-27
Build ThorVG for Android without a python-for-android recipe

ElliotGarbus:
## Summary

ThorVG's Android integration currently depends on an out-of-tree python-for-android recipe: it isn't built by CI, is versioned separately from the pin in `tools/build_thorvg.sh`, and is an undocumented unconditional runtime dependency. This PR replaces that with the same approach Kivy already uses for iOS: ThorVG becomes part of Kivy's own build (via `tools/build_thorvg.sh` + `setup.py`), not a separate mobile dependency.

- `tools/build_thorvg.sh`: new `THORVG_ANDROID=<abi>` mode that cross-compiles ThorVG with the NDK's unified LLVM toolchain, targeting API 21 (matching `ANDROID_API_LEVEL` in `pyproject.toml`). Always static — Kivy's Android wheel build has no wheel-repair step, so the archive links directly into `kivy.lib.thorvg._thorvg.so` (same rationale as the existing Windows static path). Installs to `<prefix>/libs/<abi>/libthorvg.a`, dropping into the same `dist/libs/<abi>/` tree the Android SDL3 libraries already use.
- `tools/meson-cross-android-{arm64-v8a,x86_64}.ini`: new Meson cross files for the two ABIs Kivy's Android wheel targets.
- `setup.py`: adds a `ThorVG` entry to `plat_options['android']['libraries'][abi]` alongside the existing SDL3 entries, and a dedicated `platform == 'android'` branch in `determine_thorvg_flags()` (`TVG_STATIC=1`, matching Windows).

Only `arm64-v8a` and `x86_64` are wired up (the ABIs Kivy's Android wheel already targets). 

## Test plan

- [x] Cross-compiled `libthorvg.a` for both `arm64-v8a` and `x86_64` via `tools/build_thorvg.sh` and verified the object code architecture (`ELF64 AArch64` / `x86_64`).
- [x] Built a full Kivy Android wheel (`arm64-v8a`) via `cibuildwheel` linking against the new static archive; confirmed `use_thorvg = 1` and a clean link with `-Wl,--no-undefined` (no missing symbols).
- [x] Deployed to a physical Pixel 8a (Android 16) via `kivyforge`. Used `svg-explorer` example and visually confirmed `img_thorvg_svg` renders correctly on-device — flat shapes (star, heart) and a gradient fill (hexagon) all render as expected.
- [x] CI (wheel builds / lint) on this PR.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -295,7 +295,16 @@ if platform == 'android':
             'SDL3_ttf': {
                 'path': join(lib_path, 'libSDL3_ttf.so'),
                 'headers': join(root, 'dist', 'include', 'SDL3_ttf'),
-            }
+            },
+            # Built by tools/build_thorvg.sh (THORVG_ANDROID=<abi>), not a
+            # python-for-android recipe - a static archive, unlike the .so
+            # siblings above, since Kivy's Android wheel pipeline has no
+            # wheel-repair step to bundle a separate libthorvg.so (see
+            # [tool.cibuildwheel.android] in pyproject.toml).
+            'ThorVG': {
+                'path': join(lib_path, 'libthorvg.a'),
+                'headers': join(root, 'dist', 'include', 'thorvg-1'),
+            },
         }
 
     android_data['abis'] = android_abis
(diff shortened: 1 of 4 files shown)


T-Dynamos:
@ElliotGarbus Do ensure p4a side dynamic linking approach still works and p4a side thorvg recipe should not be removed as it's required by recipes other than kivy.
Dynamic linking is done to reduce apk size.

ElliotGarbus:
@T-Dynamos Thanks for flagging this — to confirm: this PR doesn't touch python-for-android or its thorvg recipe at all. It adds a second, independent build path inside Kivy's own tools/build_thorvg.sh/setup.py (mirroring the existing iOS/Windows approach) so Kivy's Android wheel can build ThorVG without depending on the out-of-tree p4a recipe. 

The p4a recipe and its dynamic-linking setup are untouched and remain available for any other recipe that needs it — removing it was never actually part of this PR (I speculated about that as a possible future follow-up; I've dropped that line since it's out of scope here).

The PR does statically link ThorVG, let me know if that is a concern.

kuzeyron (approved):

PR #9368, opened 2026-08-23
Kuzeyron clipboard wayland

akshayaurora:
Maintainer merge checklist

### DIFF
--- a/kivy/core/clipboard/clipboard_wayland.py
+++ b/kivy/core/clipboard/clipboard_wayland.py
@@ -1,69 +1,70 @@
 '''
-Clipboard Wayland: An implementation of the Clipboard using wl-clipboard.
+Clipboard Wayland: an implementation of the Clipboard using wl-clipboard.
 '''
 
-__all__ = ('ClipboardWayland',)
-
-from kivy.utils import platform
-
-
-if platform != 'linux':
-    raise SystemError('Unsupported platform for wayland-clipboard')
-
+__all__ = ('ClipboardWayland', )
 
 from shutil import which
+from kivy.utils import platform
+from kivy.core.clipboard import ClipboardBase
+
+if platform != 'linux':
+    raise SystemError('unsupported platform for wayland clipboard')
+
+
+def _wayland_session_available():
+    import os
+    # libwayland does not require WAYLAND_DISPLAY: it defaults to wayland-0,
+    # and WAYLAND_SOCKET is an already-connected fd from the compositor.
+    if os.environ.get('WAYLAND_SOCKET'):
+        return True
+    # WAYLAND_DISPLAY is a socket *name* (wayland-0), not a path, unless it
+    # starts with / (absolute socket path, libwayland >= 1.15).
+    display = os.environ.get('WAYLAND_DISPLAY') or 'wayland-0'
+    if display.startswith('/'):
+        socket_path = display
+    else:
+        runtime_dir = os.environ.get('XDG_RUNTIME_DIR')
+        if not runtime_dir:
+            return False
+        socket_path = os.path.join(runtime_dir, display)
+    return os.path.exists(socket_path)
+
+
+if not _wayland_session_available():
+    raise SystemError('not running under Wayland')
 
 
 wl_copy, wl_paste = which('wl-copy'), which('wl-paste')
 
-if not any((wl_copy, wl_paste)):
+if not (wl_copy and wl_paste):
     raise SystemError('wl-clipboard is not installed')
 
 
-from subprocess import check_output
-from re import search
-from kivy.core.clipboard import ClipboardBase
-from kivy.logger import Logger
-
-
-def _io(cmd, encoding: str = None):
-    encoding = encoding or 'utf-8'
-
-    return check_output(
-        cmd, encoding=encoding, timeout=1
-    )
-
-
-info = search(r"^([\w-]+)\s+([\d\.]+)", _io([wl_paste, '--version']))
-Logger.info(f'{info.group(1)}: v{info.group(2)}')
+def _io(cmd, data=None):
+    from subprocess import CalledProcessError, DEVNULL, check_output
+    try:
+        return check_output(cmd, input=data, stderr=DEVNULL)
+    except CalledProcessError:
+        return b''
 
 
 class ClipboardWayland(ClipboardBase):
-    def get(self, mimetype: str = None):
-        mimetype = mimetype or 'text/plain'
+    def get(self, mimetype='text/plain'):
+        return _io([wl_paste, '--no-newline'])
 
-        return _io(
-            [wl_paste, '--type', mimetype]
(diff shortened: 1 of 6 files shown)


kivy/core/clipboard/clipboard_wayland.py
```diff
@@ -0,0 +1,71 @@
...
+    raise SystemError(
+        'wl-clipboard not available or not running under Wayland')
+
+
+def _io(cmd, data=None):
+    from subprocess import CalledProcessError, DEVNULL, check_output
+    try:
+        return check_output(cmd, input=data, stderr=DEVNULL)
+    except CalledProcessError:
+        return b''
```
kuzeyron, line 49: A popup when it fails would be great. Confusing if there are no other signs that it failed

kivy/core/clipboard/clipboard_wayland.py
```diff
@@ -0,0 +1,71 @@
...
+        socket_path = display
+    else:
+        runtime_dir = os.environ.get('XDG_RUNTIME_DIR')
+        if not runtime_dir:
+            return False
+        socket_path = os.path.join(runtime_dir, display)
+    return os.path.exists(socket_path)
+
+
+def _wl_clipboard_commands():
```
kuzeyron: Kinda useless don't you think?

kivy/core/clipboard/clipboard_wayland.py
```diff
@@ -0,0 +1,71 @@
...
+__all__ = ('ClipboardWayland', )
+
+from kivy.utils import platform
+from kivy.core.clipboard import ClipboardBase
+
+if platform != 'linux':
+    raise SystemError('unsupported platform for wayland clipboard')
+
+
+def _wayland_session_available():
```
kuzeyron, line 15: I will have to test this to confirm if this works even if I run the app through X11.

kuzeyron: It worked on my machine.

kivy/core/clipboard/clipboard_wayland.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def _wl_clipboard_commands():
+    from shutil import which
+    return which('wl-copy'), which('wl-paste')
+
+
+wl_copy, wl_paste = _wl_clipboard_commands()
+
+if not (wl_copy and wl_paste and _wayland_session_available()):
```
kuzeyron: I would check for `_wayland_session_available()` before `which()`.

kivy/core/clipboard/clipboard_wayland.py
akshayaurora: No for Popup,  maybe be better to have a Raise so user can handle the issue?

kuzeyron: I had in mind if it's in production state. Those usually don't have a terminal in use.

akshayaurora: That's the point,  as a framework kivy should not make these decisions. raising a error that can be caught by the user "programmer" would allow them to handle these errors as they see fit.

kuzeyron: But this is in the core section and it rides through other widgets. Having a visual error is simply easier when it comes to bug reports. Kinda what most programs do. And having a visual error makes it better so no one have to implement it manually.

akshayaurora: So for people that would like a popup, they can catch the exception and show a popup in their apps before packaging them.

Those would would like not to show a popup and maybe silently fallback to using a different provider or do something else should be able to ...

kuzeyron: I'm going to be stubborn and stand with my statement.

kivy/core/clipboard/clipboard_wayland.py
```diff
@@ -0,0 +1,67 @@
...
+        if not runtime_dir:
+            return False
+        socket_path = os.path.join(runtime_dir, display)
+    return os.path.exists(socket_path)
+
+
+wl_copy, wl_paste = which('wl-copy'), which('wl-paste')
+
+
+if not (_wayland_session_available() and wl_copy and wl_paste):
```
kuzeyron: Check for `_wayland_session_available()` way before  `wl_copy, wl_paste = which('wl-copy'), which('wl-paste') to save time.`

kivy/core/clipboard/clipboard_wayland.py
akshayaurora: Thinking more about it, in other providers::

They never raise or show a popup. Empty clipboard is returned as empty data.

**xclip / xsel** (`ClipboardExternalBase`) — the closest analogue, also subprocess:

```15:18:kivy/core/clipboard/_clipboard_ext.py
    def get(self, mimetype='text/plain'):
        p = self._clip('out', 'clipboard')
        data, _ = p.communicate()
        return data
```

`Popen.communicate()` does not look at the exit code. Empty or failed paste just comes back as empty bytes. `put()` is the same: no check, no log, no raise.

**GTK3** — `wait_for_text()` is `None` when empty → return `''`.

**SDL3** — `_has_text()` is false → return `''`.

**Windows** — `GetClipboardData` is null → close clipboard, return `''`.

**macOS** — `stringForType_` is missing → return `""`.

**Dummy** — missing key → `None`, and `paste()` turns that into `''`.

None of them inspect exception text for `""`. Failures at **import** time skip the provider (`SystemError` / `ImportError`); failures at **get/put** time are silent empty data.

So returning `b''` on `CalledProcessError` is the same contract.

kuzeyron: It's:
```python
    except CalledProcessError:
        return b''
```
Users running the app in production state would be confused if it's empty or just suddenly crashes.
And for xsel/xclip it would follow the same step and have a visual sign if it fails. I would implement a basic error popup with https://github.com/kivy/kivy/blob/master/kivy/data/style.kv so other modules can also use it.

kivy/core/clipboard/clipboard_wayland.py
akshayaurora: I think it should be resolved now.

kivy/core/clipboard/clipboard_wayland.py
akshayaurora: Then we are adding a feature which is separate from the scope of this change no?

That change would make sense as a separate PR that then makes behavior consistent on all providers.

kuzeyron: That's fine with me. One of us will have to implement it and cover the other clipboards too

kuzeyron (approved):

kuzeyron:
LGTM! Thanks for the assistance with the PR. One of us will have to cover when a clipboard fails in another PR.

PR #9370, opened 2026-09-01
Drop the repeated numpoints operand in multistroke

darkdi:
Three places pick the point count with `numpoints and numpoints or self.numpoints`. The first two operands are the same name, and `x and x` is `x` for every value, so the expression already means `numpoints or self.numpoints`.

Nothing changes for any input, including `numpoints=0`, which falls back to the attribute in both forms.


kuzeyron (approved):

kuzeyron:
LGTM! Thank you!

PR #9132, opened 2025-08-28
[WIP] Add native support for fallback fonts and emoji detection in Kivy (`SDL3` text provider); add helper class `SystemEmojiFontsFinder`.

DexerBR:
This PR introduces **fallback font support** for the **SDL3 text provider**, allowing alternative fonts to be used automatically when the desired character is not available in the primary font.  

It also adds **native emoji support** through the use and exposure of the new helper class `SystemEmojiFontsFinder`, which detects available emoji fonts on the system. This class is also available for external use, enabling users to manually register their own fonts, including system emoji fonts, using `LabelBase.register`.  

## Notes

It’s important to note that while this support is now available, it is **not guaranteed**, since it still depends on emoji fonts being present on the host system. A more reliable long-term solution would be to bundle one or more emoji-capable fonts with Kivy itself. For now, handling these edge cases remains the responsibility of the user.

### Test code:
```python
import os

os.environ["KIVY_TEXT"] = "sdl3"

from kivy.app import App
from kivy.core.text import LabelBase
from kivy.core.text.system_emoji_fonts import SystemEmojiFontsFinder
from kivy.uix.boxlayout import BoxLayout
from kivy.uix.label import Label

fallback_options = [
    r"C:\Windows\Fonts\seguiemj.ttf",
    r"NotoColorEmoji.ttf",
]
LabelBase.register(
    name="MyFont1",
    fn_regular=r"C:\WINDOWS\FONTS\MTCORSVA.ttf",
    fallback_fonts=fallback_options,
)

available_emoji_fonts = SystemEmojiFontsFinder.get_available_fonts()
LabelBase.register(
    name="MyFont2",
    fn_regular=r"C:\WINDOWS\FONTS\ROCK.ttf",
    fallback_fonts=available_emoji_fonts,
)


class TestApp(App):
    def build(self):
        layout = BoxLayout(orientation="vertical", padding=20, spacing=10)

        # using custom font
        label1 = Label(
            text="Code with ❤️ Build with 🔥 Ship with 🚀\nDreams → Reality 💫",
            font_name="MyFont1",
            bold=True,
            font_size="24sp",
            markup=True,
        )
        label2 = Label(
            text="Code with ❤️ Build with 🔥 Ship with 🚀\nDreams → Reality 💫",
            font_name="MyFont2",
            bold=True,
            font_size="24sp",
        )

        # native emoji support
        label3 = Label(
            text="Code with ❤️ Build with 🔥 Ship with 🚀\nDreams → Reality 💫",
[...]

PR #9266, opened 2026-02-16
Centralize provider registry for all core providers

ElliotGarbus:
Implements a centralized provider configuration registry in kivy/core/__init__.py as the single source of truth for all Kivy core provider configurations, replacing the duplicated provider lists in kivy/__init__.py and individual provider modules.

For details see: https://github.com/kivy/kivy/pull/9224#issuecomment-3644917784

Changes:
- Add PROVIDER_CONFIGS dict in kivy/core/__init__.py containing all 8 provider categories (window, text, video, audio_output, image, camera, spelling, clipboard)
- Add helper functions: get_provider_options(), get_provider_modules(), get_all_categories()
- Update kivy/__init__.py to dynamically build kivy_options from the registry
- Refactor all 8 provider modules to use get_provider_modules() while maintaining platform-specific filtering logic
- Add comprehensive test suite validating registry completeness, synchronization, and usage

Benefits:
- Eliminates sync issues between kivy_options and provider module configurations
- Single source of truth for provider lists and priority ordering
- Easier to add new providers (update one location)
- Platform-specific logic remains decentralized in provider modules for flexibility

Notes:
- Removed 'sdl' from window providers (file doesn't exist, was obsolete entry)
- Removed 'gif' from image providers (provider was already removed)
- video_android intentionally excluded from registry (used directly, not env-configurable)

Maintainer merge checklist
* [ ] Title is descriptive/clear for inclusion in release notes.
* [ ] Applied a `Component: xxx` label.
* [ ] Applied the `api-deprecation` or `api-break` label.
* [ ] Applied the `release-highlight` label to be highlighted in release notes.
* [ ] Added to the milestone version it was merged into.
* [ ] **Unittests** are included in PR.
* [ ] Properly documented, including `versionadded`, `versionchanged` as needed.

PR #9372, opened 2026-09-09
Add opt-in SDL3 font fallback with cache isolation and regression tests

tshirtman:
SDL3 labels currently render missing glyphs as tofu even when applications can supply another font. Add ordered fallback font files through `LabelBase.register(..., fallback_fonts=[...])`, using SDL3_ttf's native fallback API.

Continues #9132 with the original author's commit rebased onto master. System emoji discovery is opt-in through `SystemEmojiFontsFinder`; default font behavior stays consistent across machines.

The update fixes cache collisions between fallback chains, stale fallbacks after font changes, font handle ownership, and PathLike registration. It preserves the existing font registry format and adds a portable example and native rendering regression tests.

Validation on `eb783393f`: current PR CI passes Ubuntu Python 3.13/3.14 (tests and benchmarks), Windows Python 3.11–3.14 (tests and benchmarks), all four macOS matrix jobs including Intel and ARM, documentation generation, PEP8, and Linux source-distribution tests. Platform wheel builds were skipped by the existing workflow conditions.

The first CI run exposed a provider-registry test that mistook the new discovery helper for a provider. Added the helper to its existing exclusions and reran CI successfully. Local validation also includes 89 targeted tests and repository-wide Ruff; the native fallback tests use bundled fonts and SDL3_ttf 3.2.2.

Review considerations:
- Keep fallback opt-in to avoid system-dependent default text metrics.
- Validate tight bounds, mixed baselines, wrapping and TextInput cursor/selection with complex scripts; glyph fallback alone does not implement complete Unicode layout.
- Fallbacks are font files, not style-aware font families. Color font formats and outline/hinting behavior need platform coverage.
- The 64-entry native cache can hold multiple font handles per entry. Long chains may warrant a separate handle budget or shared ownership.


examples/text/sdl3_font_fallback.py
```diff
@@ -0,0 +1,36 @@
+"""Demonstrate opt-in SDL3 fallback using available system emoji fonts."""
+import os
+
+os.environ["KIVY_TEXT"] = "sdl3"
```
kuzeyron, line 4: You might like this: https://github.com/kivy/kivy/pull/9266

PR #9374, opened 2026-09-11
Fix AliasProperty dispatch inconsistencies (#6901)

### LINKED ISSUE #6901: Event-Dispatching on AliasProperty looks inconsistent
**Software Versions**
* Python: 3.8.1
* OS:  Linux Mint 19.3 Tricia
* Kivy: v2.0.0rc2, git-1d16c82, 20200525
* Kivy installation method: `pipenv install -e .`

**Describe the bug**

```python
from kivy.properties import NumericProperty, AliasProperty

class Person(EventDispatcher):
    age = NumericProperty(0)
    is_adult = AliasProperty(
        getter=lambda self: self.age >= 20,
        setter=None,
        bind=('age', ),
        cache=True,
    )

    def on_is_adult(self, __, v):
        print(f"`is_adult` was set to {v}.")


print("---- p1 ----")
p1 = Person(age=0)
p1.is_adult  # A
p1.age = 10
print(p1.is_adult)

print("---- p2 ----")
p2 = Person(age=0)
p2.age = 10
print(p2.is_adult)

print("---- p3 ----")
p3 = Person(age=10)
print(p3.is_adult)
```

```
---- p1 ----
False
---- p2 ----
`is_adult` was set to False.
False
---- p3 ----
`is_adult` was set to False.
False
```

The difference between `p1` and `p2` is the line `A`, which just looks up the `is_adult` property. And surprisingly, it makes a difference between thier output.

ElliotGarbus:
Fixes #6901

## Problem

Two related bugs in `AliasProperty`'s dispatch logic (`trigger_change()` / `set()`):

1. **Inconsistent initial-value dispatch.** Whether the *first* established value of an `AliasProperty` was dispatched depended on unrelated implementation details - e.g. whether the property had already been read once, or whether an `on_<name>` handler existed on the class (which forces an early read during `__init__` via auto-`fbind`). This produced surprising, order-dependent behavior, illustrated at length in the issue thread (the `Person`/`is_adult` examples).

2. **Missed dispatches when the getter returns `None`.** The alias property's internally tracked value (`ps.value`) was only kept in sync when `cache=True`. With `cache=False`, it was never updated at all, so `trigger_change()` permanently compared new getter results against the untouched `None` placeholder. Any time the getter legitimately returned `None` (or, for `cache=True`, the very first real value happened to be `None`), the comparison `None != None` incorrectly evaluated as "no change" and the dispatch was silently dropped. See the `Rect`/`aspect_ratio` and `Test`/`y` (Discord) examples in the issue.

## Fix

- `trigger_change()`: the very first time an alias property's value is established (tracked via the existing `alias_initial` flag), it now always dispatches exactly once, regardless of read/bind ordering.
- `trigger_change()` / `set()`: the tracked value (`ps.value`) and `alias_initial` are now kept in sync unconditionally, not just when `cache=True`. This is what actually fixes the `None`-handling bug - `use_cache` now only affects whether `get()` returns a cached value or recomputes on every read; it no longer has any bearing on dispatch correctness.

This matches the "always dispatch" direction the issue thread converged on, and (verified empirically) is not a behavior change for the overwhelmingly common case where the getter's value is never `None` - the old buggy `None`-placeholder comparison already caused a dispatch on first establishment by accident in that case.

## Testing

- Added 4 regression tests to `kivy/tests/test_properties.py` covering both issues (initial-value dispatch with/without priming-by-read, and `None`-getter dispatches for both `cache=True` and `cache=False`).
- Full `kivy/tests/test_properties.py` suite passes (68/68), plus `test_buttonbehavior.py` / `test_togglebuttonbehavior.py` / `test_widget.py` / `test_uix_widget.py` (139 passed, 1 unrelated skip) to check for regressions in real-world `AliasProperty` usage (`ButtonBehavior.pressed`, `ToggleButtonBehavior.activated`).
- Added a migration guide entry (`doc/sources/migration.rst`) documenting the behavior change and its (minimal) migration impact.

## Migration impact

This is primarily a bug fix. Code that binds to an `AliasProperty` and relies on constructor-time dependency changes may now observe one additional initial dispatch call in scenarios that previously (inconsistently) suppressed it - see the migration guide entry for details.


kuzeyron (approved):

kuzeyron:
LGTM! Thank you!

PR #9384, opened 2026-09-21
WIP setup: make sure the thorvg modules gets a RPATH on Unix systems

ydirson:
When building thorvg using build_linux_dependencies.sh, the matching cython module would not be linked with a RPATH set, which prevents the module from working, and causes pytest to fail as early as test collection stage.

Since RPATH is indeed set for ios and darwin, this uses the same location to set RPATH for Unix systems.  I've chosen to rework a bit the existing tests to get some visible grouping for the reader. I'm not happy with the result, let's discuss this.

Maintainer merge checklist


ydirson:
Obsoleted by #9385

